PR #78, opened 2014-08-01
Asynchronous IO performance enhancements

d-cameron:
Bug fix: Removed 2s idle thread delay when closing asynchronous SAM writer.

Asynchronous IO is currently only performed when writing records. For my use case of reading a SAM/BAM and writing a subset of records to multiple files, asynchronous parsing of the input bam on a background thread improved throughput (and CPU usage) by ~30%.

In addition to the above, basic performance profiling indicates that the performance of BlockingQueue<T> of AbstractAsyncWriter.queue can be improved by batching of records before handing over to the consumer thread (this was done in the above async reader code)

Should I incorporate an enhancement adding input parsing async capability and/or output buffering into this pull request?

### DIFF
--- a/src/java/htsjdk/samtools/util/AbstractAsyncWriter.java
+++ b/src/java/htsjdk/samtools/util/AbstractAsyncWriter.java
@@ -63,8 +63,12 @@ public abstract class AbstractAsyncWriter<T> implements Closeable {
         checkAndRethrow();
 
         if (!this.isClosed.getAndSet(true)) {
-            try { this.writer.join(); }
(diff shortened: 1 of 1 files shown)


d-cameron:
I've created a pull request with two read performance enhancements:
* Separate thread for BAM decompression - this improves read throughput by ~80% (121s vs 227s for my test BAM).
* AsyncBufferIterator - this class allows callers to wrap the input iterator in another thread thereby decoupling the downstream processing from the input parsing. For use cases where the per record processing time is comparable to the input parsing time, this can further improve throughput (I see a ~50% improvement for the SV read extraction stage of GRIDSS).

PR #149, opened 2015-01-04
Added input validation functions that can deal with both file and URL in...

iliat:
...puts, in preparation for allowing GA4GH resources as inputs into Picard tools.


iliat:
@nh13 A part from the older discarded PR that I still need to get in to support Urls as INPUTs for Picard tools. I will send the Picard side of this change to you as well.

jacarey:
Looks good.

PR #159, opened 2015-01-19
Set compression level for GZIP files

nnordborg:
I have added IOUtil.setDefaultCompressionLevel() to make it possible to change the compression level for GZIP files that are opened for writing. The code is similar to what already exists in BlockCompressedOutputStream and is intended to be used in the same way (eg. in picard.cmdline.CommandLineProgram).


vdauwera:
@nnordborg Same question as for your PR to picard -- are you still pursuing this actively or can we close this PR for now?

nnordborg:
We deployed it in our server environment. It works but seems like it didn't made as much difference as some was hoping for.

nh13:
Could you squash these commits and then we can include nonetheless?

nnordborg:
I am not so familiar with git... not really sure what you mean with 'squash the commits'. I guess I need to do some 'rebase' command, but whatever I try I can't get any other response than  'Rebase did nothing. HEAD was already up-to-date'. 

Also, I am not sure why there are 6 commits listed above. Only the last one was committed in the 'compression-level' branch.

nh13:
I rebased the commits for you (used ```git rebase -i```).  Tracking this here: https://github.com/samtools/htsjdk/pull/271

PR #160, opened 2015-01-22
Expose BinaryTagCodec

vadimzalunin:
I'd like to be able to use this codec in CRAM. Is there a particular reason for this class to be package-private?


nh13:
it was there before CRAM.  :+1:

PR #166, opened 2015-02-11
make BAMFileSpan public

vadimzalunin:
BAMIndex class is public and BAMIndex.getSpanOverlapping() method is public but it returns package-private BAMFileSpan.


vadimzalunin:
my fault, forgot to check if it compiles :)

Issue #161, opened 2015-02-06
add getBackingResourceName() to SAMReader


lbergelson:
There is currently no way to get information about the datasource backing the SAMReader.  This makes it difficult to return useful error messages when there is any sort of exception.  

It should return a `String` which sufficiently describes the resource  enough that a human could identify which file, url, database, etc was the problem.

nh13:
Would ```SamReader.Type``` help?  Could you give a concrete example? Since you are in the same room, I would be happy to discuss in person.

mattsooknah:
Should be resolved by https://github.com/samtools/htsjdk/pull/168.

PR #168, opened 2015-02-11
adding getResourceDescription() to SamReader

lbergelson:
this is to address #161 

I also fixed a completely incorrect comment for SamReader.type()

For some reason there was a transient build failure relating to `RelativeIso8601DateTest`.  I reran the build without problem, but there seems to be some edge case where one of these tests is failing at some random time.  (I clicked rerun without recording more details thinking I could still access the old build.  I can't figure out how to get at it now though.)


mattsooknah:
Looks good. I will merge.

PR #170, opened 2015-02-12
Simple sbt leaving ant build in place. Used only for maven publishing.

jacarey:
This includes the minimal set of changes to allow for an sbt build for maven publishing.

The .idea files can mostly be ignored. They are just for building in intellij with the sbt build.


geoffjentry:
+:100:

PR #172, opened 2015-02-12
Moving ReaderImplementation to SamReader

lbergelson:
it was in the deprecated `SAMFileReader` which is confusing especially since it depends on `PrimitiveSamReader` which is in `SamReader`

updated a few out of date documentation strings

### DIFF
--- a/src/java/htsjdk/samtools/BAMFileReader.java
+++ b/src/java/htsjdk/samtools/BAMFileReader.java
@@ -43,7 +43,7 @@ import java.util.NoSuchElementException;
 /**
  * Class for reading and querying BAM files.
  */
-class BAMFileReader extends SAMFileReader.ReaderImplementation {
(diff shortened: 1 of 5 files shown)


mattsooknah:
@nh13 says that this would be a good opportunity to add docs to the methods in PrimitiveSamReader and ReaderImplementation to explain what they do.

lbergelson:
A lot of them look like they're hacks to help transition from SAMFileReader to using SamReader directly.  

I can label them, or I can try and do some refactoring.

mattsooknah:
I think labeling them would be enough, for now.

nh13:
@mattsooknah I think you should be able to add in the docs I described in about 5-10 min.  Could you add those descriptions and update the PR?

mattsooknah:
Alright, I've added some more documentation and rebased. Just waiting on Travis now.

nh13:
:+1:

PR #173, opened 2015-02-19
removing SAMRecord.setAlignmentEnd()

lbergelson:
It throws unsupported operation exception.
This is useless and confusing.  It it's there for backwards compatibility (which is probably even worse than removing it because programs fail at run time) it should be deprecated and then removed after whatever delay is required.


mattsooknah:
Agreed , let's just get rid of it outright - no need to deprecate.

PR #174, opened 2015-02-20
Remove the auto dependency on scala

jacarey:


mattsooknah:
:+1:

superbobry:
On a related note: should `testng` be a test-only dependency?

PR #179, opened 2015-02-23
Un-deprecating GenotypeBuilder.log10PError

lbergelson:
it is still in use and has no replacement

### DIFF
--- a/src/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
+++ b/src/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
@@ -230,18 +230,14 @@ public final class GenotypeBuilder {
         return this;
     }
 
-    /**
-     * Adaptor interface from the pLog10Error system.
-     *
-     * Will be retired when
+    /**  Set the GQ with a log10PError value
      *
      * @param pLog10Error
      * @return
      */
-    @Deprecated
     public GenotypeBuilder log10PError(final double pLog10Error) {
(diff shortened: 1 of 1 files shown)


lbergelson:
@mattsooknah you seem to be the person to look at these

mattsooknah:
Looks good. Although there's something Zen about the statement "Will be retired when"

nh13:
Where is this being used?

mattsooknah:
It's used in GATK (GATKVariantContextUtils).

lbergelson:
GATK/hellbender

nh13:
Move it there and delete it from HTSJDK. 

N

Thumb typed for added typos

akiezun:
that sounds like a suboptimal solution here. Let's just remove the
@Deprecated

nh13:
If it is only used in the GATK and it has been deprecated we should move it out. 

N

Thumb typed for added typos

nh13:
Flip-flop, if the plan is to add more utility tools into htsjdk then lets remove the deprecation.

mattsooknah:
It appears to have been deprecated when it was brought over to HTSJDK, two years ago. It converts from log10(p(calling error)) to the corresponding Phred score, by just multiplying by -10. 

Thus I would infer that it was supposed to be retired when GATK could report the likelihood directly as a Phred score. Was this ever planned to happen?

If Phred-scaling is supposed to be the realm of HTSJDK, then it should stay there; otherwise we should move it to GATKVariantContextUtils, I think.

PR #182, opened 2015-02-24
creating a Locatable interface

lbergelson:
Renaming htsjdk.samtools.util.Interval -> NamedInterval which now extends the minimal Interval
extracting out a minimal genomic interval class from and naming that Interval
SAMRecord implements HasGenomicInterval
Feature implements HasGenomicInterval
renaming classes that are based on NamedInterval
renaming htsjdk.tribble.index.interval.Interval to FileInterval

This is motivated by the need to have a common interface to work with tribble `Feature` and `SAMRecord` for location based caching purposes. 
It's also a first step into integrating the the many interval classes that htsjdk, picard,  and GATK have and pushing functionality into a common base class.  See https://github.com/broadinstitute/hellbender/issues/100   and https://github.com/broadinstitute/hellbender/issues/159

@droazen please review

### DIFF
--- a/src/java/htsjdk/samtools/SAMRecord.java
+++ b/src/java/htsjdk/samtools/SAMRecord.java
@@ -25,6 +25,8 @@ package htsjdk.samtools;
 
 
 import htsjdk.samtools.util.CoordMath;
+import htsjdk.samtools.util.Locatable;
+import htsjdk.samtools.util.SimpleInterval;
 import htsjdk.samtools.util.StringUtil;
 
 import java.lang.reflect.Array;
@@ -80,7 +82,7 @@ import java.util.List;
  * @author alecw@broadinstitute.org
  * @author mishali.naik@intel.com
  */
-public class SAMRecord implements Cloneable
+public class SAMRecord implements Cloneable, Locatable
 {
     /**
      * Alignment score for a good alignment, but where computing a Phred-score is not feasible. 
@@ -1177,6 +1179,48 @@ public class SAMRecord implements Cloneable
      */
     protected SAMBinaryTagAndValue getBinaryAttributes() {
         return mAttributes;
+    }
+
+    /**
+     * Interval covering the aligned, unclipped, bases in this read as reported by getAlignmentStart() and getAlignmentEnd()
+     * @return the aligned interval or null if getReadUnmappedFlag()
+     */
+    public SimpleInterval getLocation() {
+        if (getReadUnmappedFlag()) {
+            return null;
+        } else {
+            return new SimpleInterval(getReferenceName(), getAlignmentStart(), getAlignmentEnd());
+        }
+    }
+
+    /**
+     * @return reference name, null if this is unmapped
+     */
+    @Override
+    public String getContig() {
+        if( getReadUnmappedFlag()) {
+            return null;
+        } else {
+            return getReferenceName();
+        }
+    }
+
+    /**
+     * an alias of {@link #getAlignmentStart()
+     * @return 1-based inclusive leftmost position of the clipped sequence, or 0 if there is no position.
+     */
+    @Override
+    public int getStart() {
+        return getAlignmentStart();
+    }
+
+    /**
+     * an alias of {@link #getAlignmentEnd()}
+     * @return 1-based inclusive rightmost position of the clipped sequence, or 0 read if unmapped.
+     */
+    @Override
+    public int getEnd() {
+        return getAlignmentEnd();
     }
 
     /**
--- a/src/java/htsjdk/samtools/util/Interval.java
+++ b/src/java/htsjdk/samtools/util/Interval.java
@@ -32,10 +32,7 @@ import java.util.Collection;
  *
  * @author Tim Fennell
  */
-public class Interval implements Comparable<Interval>, Cloneable {
-    private final String sequence;
-    private final int start;
-    private final int end;
+public class Interval extends SimpleInterval implements Comparable<Interval>, Cloneable, Locatable {
     private final boolean negativeStrand;
     private final String name;
 
@@ -63,24 +60,18 @@ public class Interval implements Comparable<Interval>, Cloneable {
      *
      */
     public Interval(final String sequence, final int start, final int end, final boolean negative, final String name) {
-        this.sequence = sequence;
-        this.start = start;
-        this.end = end;
+        super(sequence, start, end);
         this.negativeStrand = negative;
         this.name = name;
-        if (this.end < this.start-1) {
-            throw new IllegalArgumentException("start must be less than or equal to end!");
-        }
     }
 
-    /** Gets the name of the sequence on which the interval resides. */
-    public String getSequence() { return sequence; }
+    /** Gets the name of the sequence on which the interval resides.
+     * This is a simple alias of getContig()
+     * @deprecated use getContig() instead
+     */
+    @Deprecated
+    public String getSequence() { return getContig(); }
 
-    /** Gets the 1-based start position of the interval on the sequence. */
-    public int getStart() { return start; }
-
-    /** Gets the 1-based closed-ended end position of the interval on the sequence. */
-    public int getEnd() { return end; }
 
     /** Returns true if the interval is on the negative strand, otherwise false. */
     public boolean isNegativeStrand() { return this.negativeStrand; }
@@ -93,8 +84,8 @@ public class Interval implements Comparable<Interval>, Cloneable {
 
     /** Returns true if this interval overlaps the other interval, otherwise false. */
     public boolean intersects(final Interval other) {
-        return  (this.getSequence().equals(other.getSequence()) &&
-                 CoordMath.overlaps(this.start, this.end, other.start, other.end));
+        return  (this.getContig().equals(other.getContig()) &&
+                 CoordMath.overlaps(this.getStart(), this.getEnd(), other.getStart(), other.getEnd()));
     }
 
     public int getIntersectionLength(final Interval other) {
@@ -108,9 +99,9 @@ public class Interval implements Comparable<Interval>, Cloneable {
     /** Returns a new Interval that represents the intersection between the two intervals. */
     public Interval intersect(final Interval that) {
         if (!intersects(that)) throw new IllegalArgumentException(that + " does not intersect " + this);
-        return new Interval(this.sequence,
-                            Math.max(this.start, that.start),
-                            Math.min(this.end, that.end),
+        return new Interval(this.getContig(),
+                            Math.max(this.getStart(), that.getStart()),
+                            Math.min(this.getEnd(), that.getEnd()),
                             this.negativeStrand,
                             this.name + " intersection " + that.name);
     }
@@ -118,16 +109,16 @@ public class Interval implements Comparable<Interval>, Cloneable {
 
     /** Returns true if this interval overlaps the other interval, otherwise false. */
     public boolean abuts(final Interval other) {
-        return this.getSequence().equals(other.getSequence()) &&
-               (this.start == other.end + 1 || other.start == this.end + 1);
+        return this.getContig().equals(other.getContig()) &&
+               (this.getStart() == other.getEnd() + 1 || other.getStart() == this.getEnd() + 1);
     }
 
     /** Gets the length of this interval. */
-    public int length() { return this.end - this.start + 1; }
+    public int length() { return this.getEnd() - this.getStart() + 1; }
 
     /** Returns a new interval that is padded by the amount of bases specified on either side. */
     public Interval pad(final int left, final int right) {
-        return new Interval(this.sequence, this.start-left, this.end+right, this.negativeStrand, this.name);
+        return new Interval(this.getContig(), this.getStart()-left, this.getEnd()+right, this.negativeStrand, this.name);
     }
 
     /** Counts the total number of bases a collection of intervals. */
@@ -148,13 +139,13 @@ public class Interval implements Comparable<Interval>, Cloneable {
     public int compareTo(final Interval that) {
         if (that == null) return -1; // nulls last
 
-        int result = this.sequence.compareTo(that.sequence);
+        int result = this.getContig().compareTo(that.getContig());
         if (result == 0) {
-            if (this.start == that.start) {
-                result = this.end - that.end;
+            if (this.getStart() == that.getStart()) {
+                result = this.getEnd() - that.getEnd();
             }
             else {
-                result = this.start - that.start;
+                result = this.getStart() - that.getStart();
             }
         }
 
@@ -173,14 +164,14 @@ public class Interval implements Comparable<Interval>, Cloneable {
 
     @Override
     public int hashCode() {
-        int result = sequence.hashCode();
-        result = 31 * result + start;
-        result = 31 * result + end;
+        int result = getContig().hashCode();
+        result = 31 * result + getStart();
+        result = 31 * result + getEnd();
         return result;
     }
 
     public String toString() {
-        return getSequence() + ":" + start + "-" + end + "\t" + (negativeStrand ? '-' : '+') + "\t" + ((null == name) ? '.' : name);
+        return getContig() + ":" + getStart() + "-" + getEnd() + "\t" + (negativeStrand ? '-' : '+') + "\t" + ((null == name) ? '.' : name);
     }
 
     @Override
@@ -188,4 +179,5 @@ public class Interval implements Comparable<Interval>, Cloneable {
         try { return (Interval) super.clone(); }
         catch (CloneNotSupportedException cnse) { throw new SAMException("That's unpossible", cnse); }
     }
+
 }
--- a/src/java/htsjdk/samtools/util/Locatable.java
+++ b/src/java/htsjdk/samtools/util/Locatable.java
@@ -1,0 +1,25 @@
+package htsjdk.samtools.util;
+
+/**
+ * Any class that has a single logical mapping onto the genome should implement Locatable
+ * positions should be reported as 1-based and closed at both ends
+ *
+ */
+public interface Locatable {
+
+    /**
+     * Gets the contig name for the contig this is mapped to.  May return null if there is no unique mapping.
+     * @return name of the contig this is mapped to, potentially null
+     */
+    String getContig();
+
+    /**
+     * @return 1-based start position, undefined if getContig() == null
+     */
+    int getStart();
+
+    /**
+     * @return 1-based closed-ended position, undefined if getContig() == null
+     */
+    int getEnd();
+}
--- a/src/java/htsjdk/samtools/util/SimpleInterval.java
+++ b/src/java/htsjdk/samtools/util/SimpleInterval.java
@@ -1,0 +1,82 @@
+package htsjdk.samtools.util;
+
+
+/**
+ * Minimal immutable class representing a 1-based closed ended genomic interval
+ * SimpleInterval does not allow null contig names.  It cannot represent an unmapped Locatable.
+ *
+ *@warning 0 length intervals are allowed, which are represented as end = start-1
+ */
+public class SimpleInterval implements Locatable {
+
+    private final int start;
+    private final int end;
+    private final String contig;
+
+    /**
+     * Create a new immutable 1-based interval of the form [start, end]
+     * @param contig the name of the contig, must not be null
+     * @param start  1-based inclusive start position
+     * @param end  1-based inclusive end position
+     */
+    public SimpleInterval(final String contig, final int start, final int end){
+        if(contig == null){
+            throw new IllegalArgumentException("contig cannot be null");
+        }
+        if (start <= 0 ){
+            throw new IllegalArgumentException("SimpleInterval is 1 based, so start must be >= 1, start:%d" + start);
+        }
+        if (end < start -1){
+            throw new IllegalArgumentException( String.format("end must be >= start -1 . start:%d end:%d", start, end));
+        }
+        this.contig = contig;
+        this.start = start;
+        this.end = end;
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        final SimpleInterval that = (SimpleInterval) o;
+
+        if (end != that.end) return false;
+        if (start != that.start) return false;
+        if (!contig.equals(that.contig)) return false;
+
+        return true;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = start;
+        result = 31 * result + end;
+        result = 31 * result + contig.hashCode();
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return String.format("%s:%s-%s", contig, start, end);
+    }
+
+    /**
+     * @return name of the contig this is mapped to
+     */
+    public final String getContig(){
+        return contig;
+    }
+
+    /** Gets the 1-based start position of the interval on the contig. */
+    public final int getStart(){
+        return start;
+    }
+
+    /**
+     * @return the 1-based closed-ended end position of the interval on the contig.
+     * @warning in the case of a 0-length interval getEnd() == getStart - 1 */
+    public final int getEnd(){
+        return end;
+    }
+}
--- a/src/java/htsjdk/tribble/Feature.java
+++ b/src/java/htsjdk/tribble/Feature.java
@@ -24,24 +24,18 @@
 package htsjdk.tribble;
 
 
+import htsjdk.samtools.util.Locatable;
+
 /**
- * Represents a locus on a reference sequence.   The coordinate conventions for start and end are implementation
- * dependent, no specific contact is specified here.
+ * Represents a locus on a reference sequence.   All Features are expected to return 1-based closed-ended intervals.
  */
-public interface Feature {
+public interface Feature extends Locatable {
 
     /**
      * Return the features reference sequence name, e.g chromosome or contig
+     * @deprecated use getContig() instead
      */
+    @Deprecated
     public String getChr();
 
-    /**
-     * Return the start position
-     */
-    public int getStart();
-
-    /**
-     * Return the end position
-     */
-    public int getEnd();
 }
--- a/src/java/htsjdk/tribble/bed/BEDFeature.java
+++ b/src/java/htsjdk/tribble/bed/BEDFeature.java
@@ -31,6 +31,10 @@ import java.awt.*;
 /**
  * @author jrobinso
  * @date Dec 24, 2009
+ *
+ * BED feature start and end positions must adhere to the Feature interval specifications.
+ * This is different than the 0-based representation in a BED file.  This conversion is handled by {@link BEDCodec}.
+ * Anyone writing a bed file should be aware of this difference.
  */
 public interface BEDFeature extends Feature {
     Strand getStrand();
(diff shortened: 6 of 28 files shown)


lbergelson:
After continued discussion it seems like changing this interface to Locatable  and pushing the getContig(), getStart(), getEnd() calls into this interface would be a good idea.

mattsooknah:
Would it be difficult to have more things use the base Interval instead of requiring a NamedInterval?

It seems like many things don't really care about the name, and it's awkward to keep creating new NamedIntervals with ```name=null```.

lbergelson:
Ideally everything that can would be transitioned to `Interval`.  It's some work.  Methods have to be pulled up to `Interval` and then overridden again in `NamedInterval` in a number of cases.  We have a very similar issue in hellbender.

nh13:
What's the impact on Picard?  We have to make sure changes in htsjdk to support internal development effort for (currently) non-production software does not impact production software.

mattsooknah:
```Interval``` is used all over the place in Picard. If it's just renaming ```Interval``` -> ```NamedInterval```, then it's a lot of code change but very little effort. But if we wanted to do it right (as Louis described above), then it could get pretty messy...

**Comment by lbergelson** [comment]
After getting feedback I've minimized the changes.  This should cause no issues with dependant code.

There is now a new `Locatable` interface which defines a single method ``SimpleInterval getInterval()`
Introduced the `SimpleInterval` which is just a triplet of contig,start,stop.  

`Feature`,`Interval`, and `SAMRecord` now implement `Locatable` for a unified (if slightly awkward) way of handling them.

@mattsooknah @nh13 @droazen Please re-review when you get a change.

File: src/java/htsjdk/samtools/SAMRecord.java [hunk 1]
```diff
@@ -1179,6 +1181,15 @@ protected SAMBinaryTagAndValue getBinaryAttributes() {
         return mAttributes;
     }
 
+    @Override
```
nh13: Comment about what this concretely returns.  Does the clipping affect mapping, is it the span over the genome, what happens for unmapped reads etc.?

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1179,6 +1181,15 @@ protected SAMBinaryTagAndValue getBinaryAttributes() {
...
+    public SimpleInterval getInterval() {
+        if (mReferenceName != null && mAlignmentStart > 0){
```
nh13: What happens to unmapped reads with a reference and start but no alignment end?  I.e. they are given the start coordinate of their mapped mate.

src/java/htsjdk/samtools/util/Interval.java
```diff
@@ -63,24 +60,15 @@ public Interval(final String sequence, final int start, final int end) {
...
-
-    /** Gets the 1-based start position of the interval on the sequence. */
-    public int getStart() { return start; }
-
-    /** Gets the 1-based closed-ended end position of the interval on the sequence. */
-    public int getEnd() { return end; }
+    /** Gets the name of the sequence on which the interval resides.
+     * This is a simple alias of getContig()
+     */
+    public String getSequence() { return getContig(); }
```
nh13: should we deprecate since it seems we want to use ```getContig```?

src/java/htsjdk/samtools/util/SimpleInterval.java
```diff
@@ -0,0 +1,45 @@
...
+        this.contig = contig;
+        this.start = start;
+        this.end = end;
+    }
+
+    @Override
+    public final SimpleInterval getInterval(){
+        return this;
+    }
+
```
nh13: remember when I said I would stop being nitpicky? I lied, extra newline.

src/java/htsjdk/samtools/util/SimpleInterval.java
```diff
@@ -0,0 +1,45 @@
...
+    /** Gets the 1-based start position of the interval on the contig. */
+    public final String getContig(){
+        return contig;
+    }
+    /** Gets the 1-based closed end position of the interval on the contig. */
+    public final int getStart(){
+        return start;
+    }
+    public final int getEnd(){
```
nh13: comment, and add some newlines

src/java/htsjdk/tribble/bed/SimpleBEDFeature.java
```diff
@@ -136,4 +137,9 @@ public void setLink(String link) {
     public java.util.List<FullBEDFeature.Exon> getExons() {
         return emptyExonList;
     }
+
+    @Override
```
nh13: ```SimpleInterval``` is 1-based, are all the the classes that implement ```SimpleInterval``` as well?

src/java/htsjdk/samtools/util/SimpleInterval.java
```diff
@@ -0,0 +1,45 @@
+package htsjdk.samtools.util;
+
+
+/**
+ * minimal immutable class representing a 1-based closed ended genomic interval
```
nh13: thank-you for the clear definition

src/java/htsjdk/samtools/util/Locatable.java
```diff
@@ -0,0 +1,9 @@
+package htsjdk.samtools.util;
+
+public interface Locatable {
+    /**
+     * gets a {@link SimpleInterval} matching the mapping of this locatable to the reference genome
+     * @return a SimpleInterval or null if this is unmapped
+     */
+    SimpleInterval getInterval();
```
nh13: Maybe call this ```getSimpleInterval``` since we have a class called ```Interval``` which presumably could be the inferred return type?

src/java/htsjdk/tribble/BasicFeature.java
```diff
@@ -24,6 +24,8 @@
  * OTHER DEALINGS IN THE SOFTWARE.
  */
 
+import htsjdk.samtools.util.SimpleInterval;
+
 /**
  * Trivial feature -- all it holds is the key information.
  *
```
nh13: wouldn't it be nice if BasicFeature told you if it was 1-based closed ended, or is it supposed to be undefined?  If the former, document, if the latter, update the ```getInterval``` function.

src/java/htsjdk/tribble/bed/SimpleBEDFeature.java
```diff
@@ -23,6 +23,7 @@
  */
 package htsjdk.tribble.bed;
 
+import htsjdk.samtools.util.SimpleInterval;
 import htsjdk.tribble.annotation.Strand;
 
 import java.awt.*;
```
nh13: Comment about if a BED is 1-based closed ended (see ```getInterval```).

src/java/htsjdk/tribble/dbsnp/OldDbSNPFeature.java
```diff
@@ -23,6 +23,7 @@
  */
 package htsjdk.tribble.dbsnp;
 
+import htsjdk.samtools.util.SimpleInterval;
 import htsjdk.tribble.Feature;
 import htsjdk.tribble.annotation.Strand;
```
nh13: Comment about if a BED is 1-based closed ended (see ```getInterval```).

src/java/htsjdk/tribble/gelitext/GeliTextFeature.java
```diff
@@ -23,6 +23,7 @@
  */
 package htsjdk.tribble.gelitext;
 
+import htsjdk.samtools.util.SimpleInterval;
 import htsjdk.tribble.Feature;
```
nh13: Comment about if a BED is 1-based closed ended (see ```getInterval```).

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -25,6 +25,7 @@
 
 package htsjdk.variant.variantcontext;
```
nh13: Comment about if a BED is 1-based closed ended (see ```getInterval```).

src/java/htsjdk/samtools/SAMRecord.java
lbergelson: Changing to getReadUnmappedFlag() instead of this check.

src/java/htsjdk/samtools/util/SimpleInterval.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+public class SimpleInterval implements Locatable {
+
+    private final int start;
+    private final int end;
+    private final String contig;
+
+    public SimpleInterval(final String contig, final int start, final int end){
```
droazen: Add javadoc to constructor

src/java/htsjdk/samtools/util/SimpleInterval.java
```diff
@@ -0,0 +1,45 @@
...
+        if(contig == null){
+            throw new IllegalArgumentException("contig cannot be null");
```
droazen: Can the current implementation of `SimpleInterval` represent an unmapped interval? If so, how?

src/java/htsjdk/tribble/bed/SimpleBEDFeature.java
droazen: I think that internally we should always convert to 1-based closed, and only convert to other representations when writing formats that require it (like BED files).

src/java/htsjdk/samtools/util/SimpleInterval.java
nh13: What is an unmapped interval?

src/java/htsjdk/tribble/bed/SimpleBEDFeature.java
nh13: Agreed.  My intuition is that some of the classes that now implement ```getInterval``` are not 1-based closed, and so those method need to be updated.

mattsooknah:
If we're going to represent everything as 1-based closed, then presumably zero-length intervals will not be allowed, since they can't be represented. So, e.g. if we read in a BED file with a zero-length interval, we should just ignore it.

(currently, Picard tests are throwing an error on this because they assumes you can represent a zero-length BED feature as an Interval with end = start-1 ... ack!)

yfarjoun:
picard interval lists are 1 based closed...zero sized intervals are allowed
and possible. the way to do that is to have the end position be one _less_
than the start position.

mattsooknah:
Yes, what I'm saying is that seems like is a bug, not a feature...

yfarjoun:
it is true that GATK intervals cannot represent this...this is one of the
differences between the two engines and it should be resolved by the
community.

yfarjoun:
Oh, no...it's a feature!!

yfarjoun:
for example, the current ICE exome target contains zero-sized intervals.

mattsooknah:
why, though?

nh13:
Agreed, lets keep it in.  There are target captures that have zero intervals (this is public, no internal leakage please).

nh13:
insertions

mattsooknah:
Ah, I see. In that case, ```SimpleInterval``` will need to be changed to not throw an exception if end = start -1

src/java/htsjdk/samtools/util/SimpleInterval.java
droazen: I know it sounds like an oxymoron, but we actually have this concept in the GATK. Eg., if the user passes in `-L unmapped`, the GATK will create an interval object (a `GenomeLoc`, in our case) to represent this "unmapped" interval, and the engine, when given this unmapped interval object, will then set up a traversal over the unmapped reads in the input file(s). 

If a `SamRecord` can be unmapped, yet implements `Locatable`, doesn't that imply that "Locatables" in general need some way of indicating the absence of a location (ie., something like an `isUnmapped()` method)?

lbergelson:
If 0 length intervals are necessary we need to change to an open ended
interval.  Having an interval with end < start is crazy.

src/java/htsjdk/samtools/util/SimpleInterval.java
nh13: It's not an absence of location, it's a location not in the reference.  ```Locatable``` is relative to the reference implicitly, so you can say that it is "not in the reference".  I would not use "unmapped" here since that is confusing to a process (i.e. mapping).  I would ```inLocation``` or ```inReference``` or some positive term (I don't like these terms).

droazen: It is the absence of a *known* location, isn't it?

nh13: Locatable implies you can know.

droazen:
Given the sheer volume of code in the GATK and elsewhere that assumes 1-based closed intervals, I don't think that switching to an open-ended representation is a viable option, whatever the benefits from the standpoint of mathematical purity.

src/java/htsjdk/samtools/util/SimpleInterval.java
droazen: Now we are getting philosophical... to be clear, I am fine with Louis's solution of having `SAMRecord.getInterval()` return null for unmapped reads rather than providing a way for `SimpleInterval` itself to represent the "location unknown" state. I just wanted to bring up the old GATK approach of having `isUnmapped()` built into the `GenomeLoc` class to make sure that we don't want that here.

src/java/htsjdk/samtools/util/Locatable.java
droazen: How about `getLocation()` for consistency with the interface name?

lbergelson:
@droazen updated

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1180,6 +1182,33 @@ protected SAMBinaryTagAndValue getBinaryAttributes() {
...
+        if (getReadUnmappedFlag()) {
+            return null;
+        } else {
+            return new SimpleInterval(mReferenceName, mAlignmentStart, mAlignmentEnd);
+        }
+    }
+
+    @Override
+    public String getContig() {
+        return null;
```
droazen: Why is this not returning `mReferenceName`?

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1180,6 +1182,33 @@ protected SAMBinaryTagAndValue getBinaryAttributes() {
...
+    }
+
+    @Override
+    public int getStart() {
+        return 0;
+    }
+
+    @Override
+    public int getEnd() {
+        return 0;
```
droazen: Why are these not returning `mAlignmentStart` / mAlignmentEnd` ?

src/java/htsjdk/samtools/SAMRecord.java
lbergelson: because it's completely broken... huh, not sure how I managed that one.

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1180,6 +1182,40 @@ protected SAMBinaryTagAndValue getBinaryAttributes() {
...
+        if (getReadUnmappedFlag()) {
+            return null;
+        } else {
+            return new SimpleInterval(mReferenceName, mAlignmentStart, mAlignmentEnd);
+        }
+    }
+
+    /**
+     * @return get
+     */
```
droazen: Delete this incomplete javadoc

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1180,6 +1182,40 @@ protected SAMBinaryTagAndValue getBinaryAttributes() {
...
+    @Override
+    public String getContig() {
+        if( getReadUnmappedFlag()) {
+            return null;
+        } else {
+            return getReferenceName();
```
droazen: This seems fine given the disclaimer in the comments for this class that `Only if the readUnmappedFlag is false can the reference name/index and alignment start be interpreted as indicating an actual alignment position.`

src/java/htsjdk/samtools/util/SimpleInterval.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Create a new immutable 1-based interval of the form [start, end]
+     * @param contig the name of the contig
+     * @param start  1-based inclusive start position
+     * @param end  1-based inclusive end position
+     */
+    public SimpleInterval(final String contig, final int start, final int end){
+        if(contig == null){
+            throw new IllegalArgumentException("contig cannot be null");
```
droazen: `Locatable` interface documentation allows null contig names, though:

     /**
      * Gets the contig name for the contig this is mapped to.  May return null if there is no unique mapping.
      * @return name of the contig this is mapped to, potentially null
      */
     String getContig();

src/java/htsjdk/tribble/BasicFeature.java
```diff
@@ -24,31 +24,20 @@
...
- * Trivial feature -- all it holds is the key information.
- *
- * @author Mark DePristo
- * @since 5/3/12
+ * A simple concrete Feature.
  */
-public class BasicFeature implements Feature {
-    private final String chr;
-    private final int start, stop;
+public class BasicFeature extends SimpleInterval implements Feature {
```
droazen: Since this is a `Feature`, it shouldn't extend `SimpleInterval`

src/java/htsjdk/samtools/util/SimpleInterval.java
lbergelson: Yes, because some `Locatable` might be unmapped.  A `SimpleInterval` is always mapped or it is meaningless.  We can relax that for simple interval if you think it's weird.

src/java/htsjdk/tribble/Feature.java
```diff
@@ -24,24 +24,18 @@
...
  */
-public interface Feature {
+public interface Feature extends Locatable {
 
     /**
      * Return the features reference sequence name, e.g chromosome or contig
+     * @deprecated use getContig() instead
      */
+    @Deprecated
     public String getChr();
```
droazen: Create separate ticket to remove calls to the now-deprecated `getChr()` in the GATK as well

src/java/htsjdk/tribble/dbsnp/OldDbSNPFeature.java
```diff
@@ -82,10 +85,16 @@
      * the required getting and setter methods
      */
 
+    @Deprecated
     public String getChr() {
```
droazen: Should be @Override

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1605,14 +1606,32 @@ private final Genotype fullyDecodeGenotypes(final Genotype g, final VCFHeader he
     // tribble integration routines -- not for public consumption
     //
     // ---------------------------------------------------------------------------------------------------------
+    @Deprecated
```
droazen: @Override as well

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1605,14 +1606,32 @@ private final Genotype fullyDecodeGenotypes(final Genotype g, final VCFHeader he
...
+    }
+
+    @Override
+    public String getContig() {
         return contig;
     }
 
+    /**
+     * @return 1-based start position of the Variant,
+     * INDELS events usually start on the first unaltered reference base before the INDEL
```
droazen: INDELS -> INDEL

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1605,14 +1606,32 @@ private final Genotype fullyDecodeGenotypes(final Genotype g, final VCFHeader he
...
+        return getContig();
+    }
+
+    @Override
+    public String getContig() {
         return contig;
     }
 
+    /**
+     * @return 1-based start position of the Variant,
```
droazen: 1-based -> 1-based inclusive

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1605,14 +1606,32 @@ private final Genotype fullyDecodeGenotypes(final Genotype g, final VCFHeader he
...
+     * @warning be aware that the start position of the VariantContext is defined in terms of the start position specified in the
+     * underlying vcf file, VariantContexts representing the same biological event may have different start positions depending on the
+     * specifics of the vcf file they are derived from
+     */
     public int getStart() {
         return (int)start;
     }
 
+    /**
+     * @return 1-based closed end position of the Variant, if the END info field is specified that value is returned
```
droazen: should note: otherwise the end is the start + reference allele length - 1

src/java/htsjdk/tribble/Feature.java
lbergelson: see https://github.com/broadinstitute/hellbender/issues/259

src/tests/java/htsjdk/samtools/util/SimpleIntervalTest.java
```diff
@@ -0,0 +1,27 @@
+package htsjdk.samtools.util;
+
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class SimpleIntervalTest {
+
+    @Test
+    public void testZeroLengthInterval(){
+        SimpleInterval interval = new SimpleInterval("1",100,99);
```
droazen: Call `getStart()` and `getEnd()` in this test as well

droazen:
:+1: new version looks good -- just a few minor comments here and there

src/java/htsjdk/samtools/util/SimpleInterval.java
droazen: No, that's fine -- but you should explicitly mention this in the class documentation

src/java/htsjdk/tribble/BasicFeature.java
lbergelson: aaaaah

lbergelson: changed.  these extra 12 lines of autogenerated code are on your head!

droazen: I can live with it :)

src/java/htsjdk/tribble/dbsnp/OldDbSNPFeature.java
lbergelson: done

src/java/htsjdk/variant/variantcontext/VariantContext.java
lbergelson: done

src/java/htsjdk/variant/variantcontext/VariantContext.java
lbergelson: done

src/java/htsjdk/variant/variantcontext/VariantContext.java
lbergelson: done

src/java/htsjdk/variant/variantcontext/VariantContext.java
lbergelson: should we remove the stuff about indels?

src/tests/java/htsjdk/samtools/util/SimpleIntervalTest.java
lbergelson: done

src/java/htsjdk/samtools/util/SimpleInterval.java
lbergelson: done

lbergelson:
@nh13 could you take a look at this again?

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1180,6 +1182,48 @@ protected SAMBinaryTagAndValue getBinaryAttributes() {
...
 
     /**
+     * Interval covering the aligned, unclipped, bases in this read as reported by getAlignmentStart() and getAlignmentEnd()
+     * @return the aligned interval or null if getReadUnmappedFlag()
+     */
+    public SimpleInterval getLocation() {
+        if (getReadUnmappedFlag()) {
+            return null;
+        } else {
+            return new SimpleInterval(mReferenceName, mAlignmentStart, mAlignmentEnd);
```
nh13: This should probably use ```getAlignmentEnd``` instead of ```mAlignmentEnd```, since the later may be unset for mapped reads until the former is called.  It is probably better practice to use the getters in these calls anyhow.

src/java/htsjdk/tribble/bed/BEDFeature.java
```diff
@@ -31,6 +31,10 @@
 /**
  * @author jrobinso
  * @date Dec 24, 2009
+ *
+ * BED feature start and end positions must adhere to the Feature interval specifications
+ * This is different than the 0-based representation in a BED file.  This conversion is handled by {@link BEDCodec}
```
nh13: periods in lines 35/36.

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1605,14 +1606,33 @@ private final Genotype fullyDecodeGenotypes(final Genotype g, final VCFHeader he
...
+        return getContig();
+    }
+
+    @Override
+    public String getContig() {
         return contig;
     }
 
+    /**
+     * @return 1-based inclusive start position of the Variant,
```
nh13: dangling comma

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1605,14 +1606,33 @@ private final Genotype fullyDecodeGenotypes(final Genotype g, final VCFHeader he
...
     public int getStart() {
         return (int)start;
     }
 
+    /**
+     * @return 1-based closed end position of the Variant, if the END info field is specified that value is returned, otherwise the end is
+     * the start + reference allele length - 1
+     * insertions end on the reference base before the insertion event
+     * deletions end on the final deleted reference base
+     */
```
nh13: The formatting of this comment is confusing (no periods, newlines, capitalization).

nh13:
There are a few minor comments, please fix, and then :+1:

src/java/htsjdk/samtools/SAMRecord.java
lbergelson: done

src/java/htsjdk/tribble/bed/BEDFeature.java
lbergelson: done

src/java/htsjdk/variant/variantcontext/VariantContext.java
lbergelson: done

src/java/htsjdk/variant/variantcontext/VariantContext.java
lbergelson: improved the comment

lbergelson:
Squashed commits

I also added `@Deprecated` to the other 2 members of the `gelitext` package since the codec is now deprecated and no else uses those classes

lbergelson:
@nh13 I think it's ready to merge unless there are any other comments.

nh13:
@mattsooknah can you verify that this version will work with our Picard?

mattsooknah:
All tests pass. S'good to go.

lbergelson:
@nh13 At the request of @droazen I've added an equals and hashcode method to SimpleInterval.  Should be a non breaking change.  Going to merge on a thumbs up and travis completion.

lbergelson:
Also a SimpleInterval.toString().  Getting crazy over here.

lbergelson:
@nh13 Good to merge?

nh13:
Nice work!  Go for it.

PR #185, opened 2015-02-27
indexing tabix writer for bgzf compressed files

shenkers:
It is not possible to generate a tabix indexed BGZF compressed feature file, ie

http://sourceforge.net/p/samtools/mailman/message/33486490/

I created the TabixIndexingFeatureWriter to manage this process. It currently does not do anything to enforce output file structure (ie headers). To use this class, the caller instantiates a writer, adds features to it, and the index is written to disk at the conventional location when the writer is closed. 

It doesn't seem like HTSJDK has a general interface for writing line-oriented features (something symmetric with the LineReader interface, except for writing), so I also created an interface LineEncoder, so the user can specify how different feature formats are written. 

Also, I created a toplevel class for PositionalOutputStream (which is currently an inner class of IndexingVariantContextWriter), to reduce code duplication by creating another inner class.

Finally, I created two simple tests to show that a TabixIndexingFeatureWriter can be used to generate indexes for either uncompressed or BGZF compressed BED feature files.

### DIFF
--- a/src/java/htsjdk/samtools/util/PositionalOutputStream.java
+++ b/src/java/htsjdk/samtools/util/PositionalOutputStream.java
@@ -1,0 +1,66 @@
+/*
+ * The MIT License
+ *
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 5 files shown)


nh13:
It looks like this fails on three of the four JDKs (oraclejdk7, openjdk6, openjdk7).  It's likely you have some Java 8 code.  Could you fix and make sure the Travis CI builds pass?  Also, I would rebase and squash your commits.  Thanks for making this useful pull request.

geoffjentry:
+1 for Java 8

shenkers:
Sure, I'll do that... I saw there was an issue with Travis, but I was
having trouble finding a descriptive error message. I did initially have a
java 8 dependency, I guess I didn't clean up well enough (I'll try and stay
out of the htsjdk/java8 controversy ;) ). I'm happy if it's useful, HTSJDK
is a great resource, and it's nice to give back a little!

shenkers:
I removed a dangling java 8 import and squashed things down. Let me know if there's anything else.

vdauwera:
It looks like we have two solutions for the same problem here. Can we get a show of hands for what's preferred between #185 and #217?

vdauwera:
So #217 has been merged; as I understand it this becomes redundant (unless I'm missing something)

shenkers:
Yes, #217 looks good to me

vdauwera:
Thanks @shenkers -- sorry you ended up doing work that didn't get merged in, but please do contribute again.

PR #193, opened 2015-03-09
Allow comment lines in liftover chain files.

gbggrant:
Ignore the id field (which is not used and is duplicated in some old chain files)


vdauwera:
@gbggrant What's the status of this PR?

PR #195, opened 2015-03-13
paired flag in addUnmappedPair

vadimzalunin:
the reads created in addUnmappedPair method should have paired flag on


nh13:
can you squash the commits?  afterwards :100: / :100:

PR #198, opened 2015-03-18
Use RuntimeIOException instead of RuntimeException where appropriate

Lenbok:
Many places in the code directly throw RuntimeException, especially under circumstances where a more specialized exception type would be informative. In particular, cases where a checked IOException is being replaced with RuntimeException now throw RuntimeIOException.

(I haven't included any of the uses in the CRAM code at this stage since it looks to be under fairly rapid development)


lbergelson:
I support this plan.

Lenbok:
Sweet! Is there someone who should be notified in general to help get things merged in? I ask as I have another pull request #202 that had a positive comment but no further action, and another #201 that has no comments, all of these are 6 weeks old. We'd like to help more.

geoffjentry:
Hi there - thanks for this! We'll take a look at the other two soon.

PR #199, opened 2015-03-18
deleting SimpleInterval because it's too controversial

lbergelson:
We wanted to standardize interval types based on a single simple interval class.  It's too controversial though.  We're going to remove `SimpleInterval` and push it elsewhere instead.


yfarjoun:
:-( I'm so sad that this couldn't get resoved at the right level.

Y.

droazen:
This is a temporary retreat -- when we get `SimpleInterval` to where we want it to be, we may make another push for unification in htsjdk. We just can't deal with the unification issue at the same time we're trying to do basic things in hellbender like test for overlap :)

We will be creating tickets to address the issues raised in our conversations yesterday (like queries involving zero-length intervals), so your dreams of a mathematically-pure future are not dead!

yfarjoun:
:-)

mattsooknah:
Good to merge :shipit:

PR #201, opened 2015-03-18
Consistent header order

Lenbok:
The HashMap implementation has changed in Java8 so the iteration order is different with respect to earlier Java versions. This results in SAM header attribute ordering changing, as they are currently being inserted into a regular HashMap. This PR swaps to using a LinkedHashMap so that the attributes are now maintained in insertion-order.


geoffjentry:
Hi @Lenbok - it looks like this PR conflicts with master now (sorry, that's our fault for being pokey). Can you fix that and squash it down to a single commit?

Lenbok:
@geoffjentry Cool, I have now rebased/squashed to current master.

geoffjentry:
Thanks!

PR #202, opened 2015-03-18
Allow SamReaderFactory to be directly supplied with a ReferenceSource

Lenbok:
instance (rather than just a File to be used to construct a
default ReferenceSource).

This enables cross-file sharing of a single ResourceSource instance, as
well as the possibility of that instance being a subclass of
ResourceSource with an alternative method of locating reference
sequence data.


vadimzalunin:
I like where this is going and perhaps we should promote/abstract ReferenceSource class to make more generic and useful outside of cram.

PR #208, opened 2015-03-23
Fixed SamInputResource/InputResource to handle both FTP and HTTP URLS an...

tfenne:
...d not just assume all URLs are HTTP.


nh13:
:+1:

lbergelson:
It looks like [SeekableStreamFactory](https://github.com/samtools/htsjdk/blob/a30bc12df2aeb3b4312a9c236c6a639025d5b596/src/java/htsjdk/samtools/seekablestream/SeekableStreamFactory.java) has been gradually reimplemented here.

tfenne:
Good point @lbergelson  - I've updated to use SeekableStreamFactory.  Thanks!

PR #211, opened 2015-03-30
Patch for hdfs

zong0jie:
I just add functions so that htsjdk.jar can read and write files (sam/bam/bai, metrics, and IndexedFastaSequenceFile) on Hadoop hdfs.
The hdfs configure file is src/java/hdfs.config.properties.

I run several tests and find it works well. I don't know whether my work is helpful, if you think this work is make sense please contact me by zongjie@novelbio.com.


akiezun:
who can review this? may we trouble @tomwhite for an opinion?

tomwhite:
I had a quick look, and my first (high-level) question is: would it be possible to change the interfaces in htsjdk to use InputStream and OutputStream rather than (or in addition to) File? This would make it possible to use the interfaces with HDFS, without so many invasive changes to htsjdk.

zong0jie:
so lots of my work is to change the FileInputStream to InputStream and OutputStream and use IOUtils.java to get the stream.

zong0jie:
I don't know whether somebody can consider the suggestion that put all the instantiate of File and InputStream and OutputStream into IOUtil.java, just like a factory.
Such as using:
File file = IOUtil.getFile("fileName");
InputStream is = IOUtil.getInputStream(file);
instead of:
File file = new File("fileName");
FileInputStream = new FileInputStream(file);

So that we can implement the InputStream and can operate other file systems such as hdfs.
Without this change, it may hard to make the package compatible with hdfs because here are so many code to change.

tomwhite:
Rather than subclassing `java.io.File`, it would be preferable to use the new Java file support in Java 7 (`java.nio.file`), since this supports pluggable filesystems (not just Hadoop). Hadoop support is provided by https://github.com/damiencarol/jsr203-hadoop.

I've submitted a PR for the reference package here: https://github.com/samtools/htsjdk/pull/308.

droazen:
Thanks for submitting this for consideration, but now that we've migrated to Java 8 we've decided to go with the `java.nio.file.Path` approach in https://github.com/samtools/htsjdk/pull/308 for HDFS support. Closing.

PR #212, opened 2015-04-01
A downsampling iterator extracted from Picard's DownsampleSam program.

tfenne:


src/java/htsjdk/samtools/DownsamplingIterator.java
```diff
@@ -0,0 +1,128 @@
+package htsjdk.samtools;
```
nh13: Could you please add the license and year?

nh13:
:+1: after the license.

PR #215, opened 2015-04-10
Fix addMetaDataLine so sequence dictionary not duplicated

abaumann:
There was an issue in addMetaDataLine where it was calling
loadMetaDataMaps after adding a new header line, which repopulated
all metadata again.  This was mostly not a problem since the
majority of the data structures were maps, so the keys/values were
just replaced, however contigs were stored in a list, so the list
effectively got duplicated.  Instead the functionality inside
loadMetaDataMaps got moved into a function to handle one line
at a time, which now gets called by addMetaDataLine for the new
line, and in loadMetaDataMaps inside a for-loop for each
metadata entry.  Only the constructor makes use of loadMetaDataMaps.
Unit test added to VCFHeaderUnitTest.java to ensure the exception
is no longer thrown.

Resolves: 214

### DIFF
--- a/src/java/htsjdk/variant/vcf/VCFHeader.java
+++ b/src/java/htsjdk/variant/vcf/VCFHeader.java
@@ -174,8 +174,7 @@ public class VCFHeader {
      * existing header line is replaced.
      */
     public void addMetaDataLine(final VCFHeaderLine headerLine) {
-        mMetaData.add(headerLine);
-        loadMetaDataMaps();
+        addMetadataHeaderLine(headerLine);
     }
 
     /**
@@ -185,27 +184,27 @@ public class VCFHeader {
         return Collections.unmodifiableList(contigMetaData);
     }
 
-	/**
-	 * Returns the contigs in this VCF file as a SAMSequenceDictionary. Returns null if contigs lines are
-	 * not present in the header. Throws SAMException if one or more contig lines do not have length
-	 * information.
-	 */
-	public SAMSequenceDictionary getSequenceDictionary() {
-		final List<VCFContigHeaderLine> contigHeaderLines = this.getContigLines();
-		if (contigHeaderLines.isEmpty()) return null;
+    /**
+     * Returns the contigs in this VCF file as a SAMSequenceDictionary. Returns null if contigs lines are
+     * not present in the header. Throws SAMException if one or more contig lines do not have length
+     * information.
+     */
+    public SAMSequenceDictionary getSequenceDictionary() {
+        final List<VCFContigHeaderLine> contigHeaderLines = this.getContigLines();
+        if (contigHeaderLines.isEmpty()) return null;
 
-		final List<SAMSequenceRecord> sequenceRecords = new ArrayList<SAMSequenceRecord>(contigHeaderLines.size());
-		for (final VCFContigHeaderLine contigHeaderLine : contigHeaderLines) {
-			sequenceRecords.add(contigHeaderLine.getSAMSequenceRecord());
-		}
+        final List<SAMSequenceRecord> sequenceRecords = new ArrayList<SAMSequenceRecord>(contigHeaderLines.size());
+        for (final VCFContigHeaderLine contigHeaderLine : contigHeaderLines) {
+            sequenceRecords.add(contigHeaderLine.getSAMSequenceRecord());
+        }
 
-		return new SAMSequenceDictionary(sequenceRecords);
-	}
+        return new SAMSequenceDictionary(sequenceRecords);
+    }
 
-	/**
-	 * Completely replaces the contig records in this header with those in the given SAMSequenceDictionary.
-	 */
-	public void setSequenceDictionary(final SAMSequenceDictionary dictionary) {
+    /**
+     * Completely replaces the contig records in this header with those in the given SAMSequenceDictionary.
+     */
+    public void setSequenceDictionary(final SAMSequenceDictionary dictionary) {
         this.contigMetaData.clear();
 
         // Also need to remove contig record lines from mMetaData
@@ -223,9 +222,9 @@ public class VCFHeader {
         this.mMetaData.addAll(contigMetaData);
     }
 
-	public VariantContextComparator getVCFRecordComparator() {
-		return new VariantContextComparator(this.getContigLines());
-	}
+    public VariantContextComparator getVCFRecordComparator() {
+        return new VariantContextComparator(this.getContigLines());
+    }
 
     /**
      * @return all of the VCF FILTER lines in their original file order, or an empty list if none were present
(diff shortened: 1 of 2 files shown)


src/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -273,29 +272,36 @@ public void loadVCFVersion() {
...
+                            + " automatically adding a corresponding PL field to your VCF header");
+                }
+                addMetaDataLine(new VCFFormatHeaderLine(VCFConstants.GENOTYPE_PL_KEY, VCFHeaderLineCount.G, VCFHeaderLineType.Integer, "Normalized, Phred-scaled likelihoods for genotypes as defined in the VCF specification"));
             }
-            addMetaDataLine(new VCFFormatHeaderLine(VCFConstants.GENOTYPE_PL_KEY, VCFHeaderLineCount.G, VCFHeaderLineType.Integer, "Normalized, Phred-scaled likelihoods for genotypes as defined in the VCF specification"));
+        } else if ( line instanceof VCFFilterHeaderLine ) {
+            final VCFFilterHeaderLine filterLine = (VCFFilterHeaderLine)line;
+            mFilterMetaData.put(filterLine.getID(), filterLine);
+        } else if ( line instanceof VCFContigHeaderLine ) {
+            contigMetaData.add((VCFContigHeaderLine)line);
```
droazen: Shouldn't you be checking here that you're not adding a duplicate contig declaration to `contigMetaData`?

src/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -273,29 +272,36 @@ public void loadVCFVersion() {
...
+                    System.err.println("Found " + VCFConstants.GENOTYPE_LIKELIHOODS_KEY + " format, but no "
+                            + VCFConstants.GENOTYPE_PL_KEY + " field.  We now only manage PL fields internally"
+                            + " automatically adding a corresponding PL field to your VCF header");
+                }
+                addMetaDataLine(new VCFFormatHeaderLine(VCFConstants.GENOTYPE_PL_KEY, VCFHeaderLineCount.G, VCFHeaderLineType.Integer, "Normalized, Phred-scaled likelihoods for genotypes as defined in the VCF specification"));
             }
-            addMetaDataLine(new VCFFormatHeaderLine(VCFConstants.GENOTYPE_PL_KEY, VCFHeaderLineCount.G, VCFHeaderLineType.Integer, "Normalized, Phred-scaled likelihoods for genotypes as defined in the VCF specification"));
+        } else if ( line instanceof VCFFilterHeaderLine ) {
+            final VCFFilterHeaderLine filterLine = (VCFFilterHeaderLine)line;
+            mFilterMetaData.put(filterLine.getID(), filterLine);
```
droazen: It's a bit inconsistent that for info and format lines, we call `addMetaDataMapBinding()` and therefore preserve the existing value when adding a line with an existing ID, whereas for filter and "other" lines we always replace the existing value in the map with new values having the same ID (since for those lines you call `Map.put()` directly instead of `addMetaDataMapBinding()`). Can you think of a good reason for this inconsistency? Perhaps we should make the behavior consistent here -- either always replace existing values with the same key, or never replace them.

src/tests/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
```diff
@@ -131,6 +133,20 @@ public void testVCFHeaderSampleRenamingSitesOnlyVCF() throws Exception {
...
+
+        VCFFileReader reader = new VCFFileReader(input, false);
+        VCFHeader header = reader.getFileHeader();
+
+        VCFInfoHeaderLine newInfoField = new VCFInfoHeaderLine("test", VCFHeaderLineCount.UNBOUNDED, VCFHeaderLineType.String, "test info field");
+        header.addMetaDataLine(newInfoField);
+
+        // getting the sequence dictionary was failing due to duplicating contigs in issue #214,
+        // as long as we have no exception then we pass this test
+        header.getSequenceDictionary();
```
droazen: Add an explicit assertion that the number of contig lines in the header (`header.getContigLines().size()`) is the same before and after the call to `addMetaDataLine()`

droazen:
Review complete -- back to @abaumann

src/java/htsjdk/variant/vcf/VCFHeader.java
abaumann: These are both good valid points.  I don't know what the intention was initially around these choices, but I agree it is probably the most consistent choice is do what addMetaDataMapBinding does and don't replace but show a message if we are in debug mode.  I'd also be ok with replacing them, but presumably there was a reason for how addMetaDataMapBinding works...

src/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -273,41 +272,64 @@ public void loadVCFVersion() {
...
-     * Add line to map, issuing warnings about duplicates
+     * Add line to map at a given key issuing warnings about duplicates
      *
-     * @param map
-     * @param line
-     * @param <T>
+     * @param map a map from each key to the associated VCFCompoundHeaderLine
+     * @param key the key to insert this line at
+     * @param line the line to insert at this key
+     * @param <T> a type of vcf header line that extends VCFCompoundHeaderLine
```
droazen: Remove mention of `VCFCompoundHeaderLine` from the docs, since this method now takes anything that extends `VCFHeaderLine` itself.

src/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -273,41 +272,64 @@ public void loadVCFVersion() {
...
             }
-            addMetaDataLine(new VCFFormatHeaderLine(VCFConstants.GENOTYPE_PL_KEY, VCFHeaderLineCount.G, VCFHeaderLineType.Integer, "Normalized, Phred-scaled likelihoods for genotypes as defined in the VCF specification"));
         }
+
+        contigMetaData.add(line);
     }
 
     /**
-     * Add line to map, issuing warnings about duplicates
+     * Add line to map at a given key issuing warnings about duplicates
```
droazen: Mention in method description that the line will not be added if the key already exists.

src/tests/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
```diff
@@ -131,6 +134,64 @@ public void testVCFHeaderSampleRenamingSitesOnlyVCF() throws Exception {
...
+        VCFFileReader reader = new VCFFileReader(input, false);
+        VCFHeader header = reader.getFileHeader();
+
+        final int numContigLinesBefore = header.getContigLines().size();
+
+        VCFInfoHeaderLine newInfoField = new VCFInfoHeaderLine("test", VCFHeaderLineCount.UNBOUNDED, VCFHeaderLineType.String, "test info field");
+        header.addMetaDataLine(newInfoField);
+
+        // getting the sequence dictionary was failing due to duplicating contigs in issue #214,
+        // we expect this to now throw an exception
```
droazen: we expect it to NOT throw an exception :)

src/tests/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
```diff
@@ -131,6 +134,64 @@ public void testVCFHeaderSampleRenamingSitesOnlyVCF() throws Exception {
...
+        VCFHeaderLine newHeaderLine = new VCFHeaderLine("key", "value");
+        // add this new header line
+        header.addMetaDataLine(newHeaderLine);
+
+        final int numHeaderLinesBefore = header.getOtherHeaderLines().size();
+        // readd the same header line
+        header.addMetaDataLine(newHeaderLine);
+        final int numHeaderLinesAfter = header.getOtherHeaderLines().size();
+
+        // assert that we have the same number of contig lines before and after
```
droazen: not contig lines,  "other" header lines

src/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -273,41 +272,64 @@ public void loadVCFVersion() {
...
-            } else if ( line instanceof VCFContigHeaderLine ) {
-                contigMetaData.add((VCFContigHeaderLine)line);
-            } else {
-                mOtherMetaData.put(line.getKey(), line);
+            addMetadataHeaderLine(line);
+        }
+    }
+
+    /**
+     * add a single header line to the appropriate map/list depending on its type
```
droazen: Mention in docs here that lines with pre-existing keys will be discarded and not added to the header

droazen:
Looks good -- just a few minor documentation issues. Feel free to rebase and merge once these are addressed and tests pass.

src/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -273,41 +272,64 @@ public void loadVCFVersion() {
...
+            final VCFFormatHeaderLine formatLine = (VCFFormatHeaderLine)line;
+            addMetaDataMapBinding(mFormatMetaData, formatLine.getID(), formatLine);
+
+            if ( hasFormatLine(VCFConstants.GENOTYPE_LIKELIHOODS_KEY) && ! hasFormatLine(VCFConstants.GENOTYPE_PL_KEY) ) {
+                if ( GeneralUtils.DEBUG_MODE_ENABLED ) {
+                    System.err.println("Found " + VCFConstants.GENOTYPE_LIKELIHOODS_KEY + " format, but no "
+                            + VCFConstants.GENOTYPE_PL_KEY + " field.  We now only manage PL fields internally"
+                            + " automatically adding a corresponding PL field to your VCF header");
+                }
+                addMetaDataLine(new VCFFormatHeaderLine(VCFConstants.GENOTYPE_PL_KEY, VCFHeaderLineCount.G, VCFHeaderLineType.Integer, "Normalized, Phred-scaled likelihoods for genotypes as defined in the VCF specification"));
```
droazen: Just noticed this -- why is this calling `addMetaDataLine()` instead of `addMetadataHeaderLine()`?

src/tests/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
```diff
@@ -131,6 +134,64 @@ public void testVCFHeaderSampleRenamingSitesOnlyVCF() throws Exception {
         final VCFHeader header = (VCFHeader) codec.readHeader(vcfIterator).getHeaderValue();
     }
 
+    @Test
+    public void testVCFHeaderAddMetaDataLine() {
```
droazen: This should really be called something like `testVCFHeaderAddMetaDataLineDoesNotDuplicateContigs()`, since it's not actually testing whether the new line was added correctly.

src/java/htsjdk/variant/vcf/VCFHeader.java
abaumann: addMetadataHeaderLine is lazy and likes to delegate :)

src/tests/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
abaumann: now not throw an exception now

Issue #203, opened 2015-03-18
write tests for VariantContext validation


akiezun:
htsjdk 1.129
none of the validation methods on VariantContext is tested:
extraStrictValidation
validateReferenceBases
validateRSIDs(rsIDs);
validateAlternateAlleles();
validateChromosomeCounts();

magicDGS:
This could be close, no?

PR #216, opened 2015-04-10
Add unit tests for validation methods in VariantContext

abaumann:
There will be unit tests that test the functionality of:
extraStrictValidation
validateReferenceBases
validateRSIDs(rsIDs);
validateAlternateAlleles();
validateChromosomeCounts();

The extraStrictValidation method calls all the other methods,
so the test for this makes use of the data providers for the
other tests - I couldn't think of a better way to do this,
and someone could change what extraStrictValidation does, so
it seemed safest to provide a test for it.  Each method has
a pass and a fail test for all the conditions I could think of.

Resolves: #203


vdauwera:
@abaumann Is this still in progress or is it ready to be reviewed?

abaumann:
There's a pull request that just needs to be reviewed

nh13:
:+1: squash and merge

akiezun:
@abaumann rebase squash and merge

abaumann:
I believe this should be good to go now

Issue #184, opened 2015-02-27
IndexFactory.createTabixIndex() is not able to handle block-compressed files


droazen:
Currently, if you pass a block-compressed file into `IndexFactory.createTabixIndex()`, the file never gets decompressed internally, causing the underlying codec to blow up with errors like the following:

        Exception in thread "main" htsjdk.tribble.TribbleException$InvalidHeader: Your input file has a malformed header: We never saw the required CHROM header line (starting with one #) for the input VCF file
	at htsjdk.variant.vcf.VCFCodec.readActualHeader(VCFCodec.java:115)
	at htsjdk.tribble.AsciiFeatureCodec.readHeader(AsciiFeatureCodec.java:88)
	at htsjdk.tribble.AsciiFeatureCodec.readHeader(AsciiFeatureCodec.java:41)
	at htsjdk.tribble.index.IndexFactory$FeatureIterator.readHeader(IndexFactory.java:413)
	at htsjdk.tribble.index.IndexFactory$FeatureIterator.<init>(IndexFactory.java:401)
	at htsjdk.tribble.index.IndexFactory.createTabixIndex(IndexFactory.java:327)

I see two possible solutions, and wanted to solicit feedback on which would be preferable:

1. Create an overload of `createTabixIndex()` that accepts a stream rather than a File (this places the burden of setting up the stream correctly on the client)

2. Have the existing `FeatureIterator.initStream()` check for block-compressed input via `AbstractFeatureReader.hasBlockCompressedExtension()`, and wrap the stream returned within a `BlockCompressedInputStream` as needed.

Thoughts?

jmthibault79:
I like option 2, so calling it just works.  What are the downsides to this solution?

I noticed that there are two `FeatureIterator` classes with similar signatures, so we may want to improve this confusing situation as well.

jmthibault79:
See also #185 !

mattsooknah:
I also prefer option 2 - it should be the factory's job to handle this sort of thing.

vdauwera:
Folks, it looks like we have two solutions for the same problem here. Can we get a show of hands for what's preferred between #185 and #217?

vdauwera:
Umm, looks like I'm late to the party, #217 has been merged. Never mind.

PR #217, opened 2015-04-13
Modify IndexFactory.createTabixIndex() for bgzip input

abaumann:
Modify the internal FeatureIterator.initStream method inside
IndexFactory to accept both block-compressed and non-compressed
input files. This method leaves the constructor as is and does
the work without extra work for the user.  The method will
check that this is proper bgzip data, and throw exception if
not.  The unit test checks that the same result comes from calling
this method with a VCF and the same bgzipped VCF.

Resolves: #184


nh13:
@jmthibault79 or @mattsooknah or @droazen, can you take a look at this?

src/java/htsjdk/tribble/index/IndexFactory.java
```diff
@@ -420,8 +419,32 @@ private FeatureCodecHeader readHeader() {
...
+                    final int bufferSize = Math.max(Defaults.BUFFER_SIZE, BlockCompressedStreamConstants.MAX_COMPRESSED_BLOCK_SIZE);
+                    final BufferedInputStream bufferedStream = new BufferedInputStream(fileStream, bufferSize);
+
+                    if (!BlockCompressedInputStream.isValidFile(bufferedStream)) {
+                        throw new TribbleException.MalformedFeatureFile("Input file is not in valid block compressed format.", inputFile.getAbsolutePath());
+                    }
+
+                    // if we got here, the file is valid, make a SeekableStream for the BlockCompressedInputStream to read from
+                    final SeekableStream seekableStream =
+                            SeekableStreamFactory.getInstance().getBufferedStream(
```
jmthibault79: I would save a local reference to the `SeekableStreamFactory` instance here, to make it a little more readable.

But that's just personal preference, ignore if you disagree.

jmthibault79:
I like this approach.  :+1: from me.

abaumann:
slight modification to reuse SeekableStreamFactory.getInstance(), if build passes, shall I merge?

jmthibault79:
Yes go ahead.

PR #220, opened 2015-04-17
remove default implementation of FeatureCodec.canDecode 

akiezun:
remove default implementation of FeatureCodec.canDecode to enforce implementation. Remove 2 obsolete and deprecated codecs. Required as part of https://github.com/broadinstitute/hellbender/issues/230


src/java/htsjdk/tribble/AbstractFeatureCodec.java
```diff
@@ -26,7 +26,10 @@
 import java.io.IOException;
 
 /**
- * Simple basic class providing much of the basic functionality of codecs
+ * Simple basic class providing much of the basic functionality of codecs.
+ * Every concrete subclass must implement {@link FeatureCodec#canDecode(String)} to indicate whether it can decode the file.
+ * Note that that method is the only way that the right codec for a file is identified and that <bold>only one</bold> codec
```
droazen: Typo: that that -> that

src/java/htsjdk/tribble/AbstractFeatureCodec.java
```diff
@@ -26,7 +26,10 @@
...
+ * is allowed to identify itself as being able to decode any given file.
```
droazen: Add a comment that checking the file's extension is a perfectly acceptable implementation of `canDecode()`, and that file contents only rarely need to be checked (this should probably go in the javadoc for `canDecode()` itself in the `FeatureCodec` interface rather than here).

.idea/modules.xml
```diff
@@ -6,4 +6,4 @@
       <module fileurl="file://$PROJECT_DIR$/.idea/modules/htsjdk-build.iml" filepath="$PROJECT_DIR$/.idea/modules/htsjdk-build.iml" />
     </modules>
   </component>
-</project>
\ No newline at end of file
+</project>
```
droazen: Revert unrelated changes to modules.xml and .gitignore

akiezun:
oops not ready yet

src/java/htsjdk/tribble/gelitext/GeliTextCodec.java
```diff
@@ -1,112 +0,0 @@
-/*
```
nh13: is this used in IGV?  This needs to be checked before you can remove things in tribble.

akiezun: will change the request to only remove the default impl. and keep the codecs

PR #221, opened 2015-04-17
remove default implementation of FeatureCodec.canDecode 

akiezun:
remove default implementation of FeatureCodec.canDecode to enforce implementation in concrete classes

### DIFF
--- a/src/java/htsjdk/tribble/AbstractFeatureCodec.java
+++ b/src/java/htsjdk/tribble/AbstractFeatureCodec.java
@@ -27,6 +27,9 @@ import java.io.IOException;
 
 /**
  * Simple basic class providing much of the basic functionality of codecs
+ * Every concrete subclass must implement {@link FeatureCodec#canDecode(String)} to indicate whether it can decode the file.
+ * Note that that method is the only way that the right codec for a file is identified and that <bold>only one</bold> codec
+ * is allowed to identify itself as being able to decode any given file.
  */
 public abstract class AbstractFeatureCodec<FEATURE_TYPE extends Feature, SOURCE> implements FeatureCodec<FEATURE_TYPE, SOURCE> {
     private final Class<FEATURE_TYPE> myClass;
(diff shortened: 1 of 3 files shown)


src/java/htsjdk/tribble/dbsnp/OldDbSNPCodec.java
```diff
@@ -89,6 +89,11 @@ public OldDbSNPFeature decode(String line) {
     }
 
     @Override
+    public boolean canDecode(String path){
+	return path.toLowerCase().endsWith(".dbsnp") || path.toLowerCase().endsWith(".rod") || path.toLowerCase().endsWith(".snp");
```
droazen: Do we really want ".rod" and ".snp" as recognized file extensions for this codec? They are way too generic and non-standard. Just use ".dbsnp"

nh13: How will this affect current users?  Does this require them to change all their files to ".dbsnp"?

droazen: Only if they are using a toolkit that relies on `canDecode()` for the selection of codecs for Tribble-supported file formats. Hellbender is the only such toolkit, as far as I know, so they will only be required to use the ".dbsnp" extension if they want to read a dbsnp file in hellbender. Tools that don't try to do fancy auto-selection of codecs will be unaffected.

vdauwera: Um, not sure if this is relevant or not but we just recently deleted OldDbSNPFeature entirely (on the premise that no one uses it anymore).

akiezun:
@droazen please review

src/java/htsjdk/tribble/dbsnp/OldDbSNPCodec.java
droazen: This codec should either be deleted (based on @vdauwera 's comment above), or not claim generic extensions such as `.rod` and `.snp` in its `canDecode()` method.

droazen:
Looks good -- one comment to address involving `OldDbSNPCodec`, then merge at will.

nh13:
@akiezun looks like there is a merge conflict, but after that, merge it!

PR #223, opened 2015-04-24
Yf geometrical downsampling

yfarjoun:
- Moved PgIdGenerator from Picard-public to here 
- Fixed a small bug in SAMRecordSetBuilder (didn't add the NM tag which then fails validation)

Needed for a new tool in Picard (Position-based downsampling)

### DIFF
--- a/src/java/htsjdk/samtools/SAMRecordSetBuilder.java
+++ b/src/java/htsjdk/samtools/SAMRecordSetBuilder.java
@@ -27,6 +27,7 @@ import htsjdk.samtools.DuplicateScoringStrategy.ScoringStrategy;
 import htsjdk.samtools.util.CloseableIterator;
 import htsjdk.samtools.util.CoordMath;
 import htsjdk.samtools.util.RuntimeIOException;
+import htsjdk.samtools.util.SequenceUtil;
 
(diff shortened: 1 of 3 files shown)


src/java/htsjdk/samtools/SAMRecordSetBuilder.java
```diff
@@ -346,6 +347,7 @@ public void addPair(final String name, final int contig, final int start1, final
         end2.setAlignmentStart(start2);
         end2.setReadNegativeStrandFlag(true);
         end2.setCigarString(readLength + "M");
+        end2.setAttribute(ReservedTagConstants.NM,0);
```
nh13: I don't think this is the only place where ```NM``` is not set (see the other ```addPair``` methods).  Do you want to add it to ```createReadNoFlag```?

yfarjoun: impossible. :-/

yfarjoun: OK. fixed it. look again?

yfarjoun:
Got verbal :+1: from @nh13 

will merge once tests pass.

PR #228, opened 2015-04-26
Update README.md with license info

### LINKED ISSUE #176: Explicit declaration of license is missing
The README doesn't state a license, and there is also no LICENSE file. It seems from at least some of the source code that the code is under the MIT license. I'd recommend making that explicit by putting a LICENSE file in the root folder of the repo, and/or declaring the license under which the code is released explicitly in the README.

vdauwera:
Fixes #176


nh13:
:+1:

PR #230, opened 2015-04-28
Added Genotype Filters to JEXL Map

johnwallace123:
In order to use the "FT" flag in JEXL expressions, I thought it would be a good idea to add them to the JEXL map.  I needed to use this flag in GATK's GenotypeConcordance tool to set genotypes that fail genotype-level filters to no-call.

I propose the following:

- Add the meta-tag 'isPassFT' which is true when g.isFiltered() is false.  (NOTE: I believe isFiltered is already available at the variant level; I'm also open to using "isFailFT" if that's more intuitive)
- Add the VCFConstants.GENOTYPE_FILTER_KEY (FT) to the JEXL map.  If the FT key is not present or is "PASS" (and therefore not present in g.getFilters()), manually set this entry to PASS.

### DIFF
--- a/src/java/htsjdk/variant/variantcontext/JEXLMap.java
+++ b/src/java/htsjdk/variant/variantcontext/JEXLMap.java
@@ -102,6 +102,8 @@ class JEXLMap implements Map<JexlVCMatchExp, Boolean> {
                 infoMap.put("isNoCall", g.isNoCall()? "1" : "0");
                 infoMap.put("isMixed", g.isMixed()? "1" : "0");
                 infoMap.put("isAvailable", g.isAvailable()? "1" : "0");
+                infoMap.put("isPassFT", g.isFiltered()? "0" : "1");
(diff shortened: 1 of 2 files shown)


nh13:
@droazen can you review this one?

droazen:
@johnwallace123 The proposal seems reasonable, but could you please add two unit tests for this feature (one for `isPassFT` and one `FT`)?

johnwallace123:
I'd be happy to add the tests, but I wasn't sure where they should go; I didn't see a unit test for the JEXL map directly.  Should I put these tests in the VariantContextUnitTest?

Thanks for your patience with my first journey into the belly of htsjdk :).

droazen:
@johnwallace123 Please add the tests to `VariantJEXLContextUnitTest`. Thanks for the contribution! :)

johnwallace123:
I attempted to test everything that I could think of; please let me know if it is acceptable for you, or if you'd like something different (more test cases and/or split into multiple tests).

droazen:
Looks great -- merging!

PR #232, opened 2015-04-29
Adding the ability to output the SAM flag field in various formats,

nh13:
similar to the 'samtools view -X' option.  Use the FLAG_FIELDS_OUTPUT
within the Defaults class to set this.

[Review on Reviewable](https://reviewable.io/reviews/samtools/htsjdk/232)

### DIFF
--- a/src/main/java/htsjdk/samtools/SamFlagField.java
+++ b/src/main/java/htsjdk/samtools/SamFlagField.java
@@ -1,0 +1,203 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Nils Homer
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package htsjdk.samtools;
+
+/**
+ * This determines how flag fields are represented in the SAM file.
+ *
+ * In a string FLAG, each character represents one bit with
+ * p=0x1 (paired), P=0x2 (properly paired), u=0x4 (unmapped),
+ * U=0x8 (mate unmapped), r=0x10 (reverse), R=0x20 (mate reverse)
+ * 1=0x40 (first), 2=0x80 (second), s=0x100 (not primary), 
+ * x=0x200 (failure), d=0x400 (duplicate), and S=0x800 (secondary).
+ * This was inspired by 'samtools view -X'.
+ *
+ * We also output a character when the following bits *are not* set:
+ * m=0x4 (mapped), M=0x8 (mate mapped), f=0x10 (forward), F=0x20 
+ * (mate forward).
+ * 
+ * @author nhomer
+ */
+public enum SamFlagField {
+    NONE {
+        @Override
+        public String format(final int flag) {
+            throw new SAMFormatException("NONE not allowed for the SamFlagField when writing the SAM flag field.");
+        }
+        @Override
+        protected int parseWithoutValidation(final String flag) {
+            throw new SAMFormatException("NONE not allowed for the SamFlagField when reading the SAM flag field.");
+        } 
+    },
+    DECIMAL {
+        @Override
+        public String format(final int flag) {
+            return Integer.toString(flag);
+        }
+        /** Throws NumberFormatException if it can't parse the flag **/
+        @Override
+        protected int parseWithoutValidation(final String flag) {
+            return Integer.parseInt(flag);
+        }
+    },
+    HEXADECIMAL {
+        @Override
+        public String format(final int flag) {
+            return String.format("%#x", flag);
+        }
+        @Override
+        protected int parseWithoutValidation(final String flag) {
+            return Integer.valueOf(flag.substring(2), 16);
+        }
+    },
+    OCTAL {
+        @Override
+        public String format(final int flag) {
+            return String.format("%#o", flag);
+        }
+        @Override
+        protected int parseWithoutValidation(final String flag) {
+            return Integer.valueOf(flag, 8);
+        }
+    },
+    STRING {
+        /*
+        It is important that the first character of a string does not start with a digit, so we can
+        determine which format given an input flag value.  See of.
+         */
+
+        @Override
+        public String format(final int flag) {
+            // Adapted from the the implementation here:
+            //   https://github.com/jmarshall/cansam/blob/master/lib/alignment.cpp
+            final StringBuilder value = new StringBuilder();
+
+            if ((flag & SAMFlag.READ_UNMAPPED.flag) != 0)                   value.append('u');
+            else                                                            value.append('m');
+
+            if ((flag & SAMFlag.READ_REVERSE_STRAND.flag) != 0)             value.append('r');
+            else if ((flag & SAMFlag.READ_UNMAPPED.flag) == 0)              value.append('f');
+
+            if ((flag & SAMFlag.MATE_UNMAPPED.flag) != 0)                   value.append('U');
+            else if ((flag & SAMFlag.READ_PAIRED.flag) != 0)                value.append('M');
+
+            if ((flag & SAMFlag.MATE_REVERSE_STRAND.flag) != 0)             value.append('R');
+            else if ((flag & SAMFlag.READ_PAIRED.flag) != 0)                value.append('F');
+
+            if ((flag & SAMFlag.READ_PAIRED.flag) != 0)                     value.append('p');
+            if ((flag & SAMFlag.PROPER_PAIR.flag) != 0)                     value.append('P');
+            if ((flag & SAMFlag.FIRST_OF_PAIR.flag) != 0)                   value.append('1');
+            if ((flag & SAMFlag.SECOND_OF_PAIR.flag) != 0)                  value.append('2');
+
+            if ((flag & SAMFlag.NOT_PRIMARY_ALIGNMENT.flag) != 0)           value.append('s');
+            if ((flag & SAMFlag.SUPPLEMENTARY_ALIGNMENT.flag) != 0)         value.append('S');
+            if ((flag & SAMFlag.READ_FAILS_VENDOR_QUALITY_CHECK.flag) != 0) value.append('x');
+            if ((flag & SAMFlag.DUPLICATE_READ.flag) != 0)                  value.append('d');
+
+            return value.toString();
+        }
+
+        @Override
+        protected int parseWithoutValidation(final String flag) {
+            SamFlagField.validate(flag, STRING);
+
+            // Adapted from the the implementation here:
+            //   https://github.com/jmarshall/cansam/blob/master/lib/alignment.cpp
+
+            int value = 0;
+
+            for (int i = 0; i < flag.length(); i++) {
+                switch (flag.charAt(i)) {
+                    case 'p':  value |= SAMFlag.READ_PAIRED.flag;  break;
+                    case 'P':  value |= SAMFlag.PROPER_PAIR.flag;  break;
+                    case 'u':  value |= SAMFlag.READ_UNMAPPED.flag;  break;
+                    case 'U':  value |= SAMFlag.MATE_UNMAPPED.flag;  break;
+                    case 'r':  value |= SAMFlag.READ_REVERSE_STRAND.flag;  break;
+                    case 'R':  value |= SAMFlag.MATE_REVERSE_STRAND.flag;  break;
+                    case '1':  value |= SAMFlag.FIRST_OF_PAIR.flag;  break;
+                    case '2':  value |= SAMFlag.SECOND_OF_PAIR.flag;  break;
+                    case 's':  value |= SAMFlag.NOT_PRIMARY_ALIGNMENT.flag;  break;
+                    case 'x':  value |= SAMFlag.READ_FAILS_VENDOR_QUALITY_CHECK.flag;  break;
+                    case 'd':  value |= SAMFlag.DUPLICATE_READ.flag;  break;
+                    case 'S':  value |= SAMFlag.SUPPLEMENTARY_ALIGNMENT.flag;  break;
+                    case 'f':
+                    case 'F':
+                    case 'm':
+                    case 'M':
+                    case '_':
+                        break;
+                    default:
+                        throw new SAMFormatException("Unknown flag character '" + flag.charAt(i) + "' in flag '" + flag + "'");
+                }
+            }
+
+            return value;
+        }
+    };
+
+    /** Returns the string associated with this flag field. */
+    abstract public String format(final int flag);
+
+    /** Parses the flag.  Validates that the flag is of the correct type. */
+    public final int parse(final String flag) {
+        return parse(flag, true);
+    }
+
+    /** Infers the format from the flag string and parses the flag. */
+    public static int parseDefault(final String flag) {
+        return SamFlagField.of(flag).parse(flag, false);
+    }
+
+    /** Performs the actual parsing based on the radix.  No validation that the flag is of the correct radix
+     * should be performed.
+     */
+    abstract protected int parseWithoutValidation(final String flag);
+
+    /** Parses the flag.  Performs optional validation that the flag is of the correct type. */
+    private int parse(final String flag, final boolean withValidation) {
+        if (withValidation) SamFlagField.validate(flag, this);
+        return parseWithoutValidation(flag);
+    }
+
+    /**
+     * Returns the type of flag field for this string.  This does not guarantee it is of the flag field,
+     * as it only checks the first two characters.
+     */
+    public static SamFlagField of(final String s) {
+        if (s.isEmpty()) throw new SAMFormatException("Could not determine flag field type; saw an empty flag field");
+        else if (s.startsWith("0x")) return HEXADECIMAL;
+        else if (s.startsWith("0X")) return HEXADECIMAL;
+        else if (s.startsWith("0") && s.length() > 1) return OCTAL;
+        else if (Character.isDigit(s.charAt(0))) return DECIMAL;
+        else return STRING;
+    }
+
+    private static void validate(final String flag, final SamFlagField expectedField) {
+        final SamFlagField actualField = SamFlagField.of(flag);
+        if (actualField != expectedField) {
+            throw new SAMFormatException(expectedField.name() + " sam flag must start with [1-9] but found '" + flag + "' (" + actualField.name() + ")");
+        }
+    }
+}
--- a/src/test/java/htsjdk/samtools/SamFlagFieldTest.java
+++ b/src/test/java/htsjdk/samtools/SamFlagFieldTest.java
@@ -1,0 +1,150 @@
+package htsjdk.samtools;
+
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+/**
+ * @author nhomer
+ */
+public class SamFlagFieldTest {
+
+    @Test
+    public void testAllFlags() {
+        int flagAsInteger = 0;
+        for (final SAMFlag samFlag : SAMFlag.values()) {
+            flagAsInteger |= samFlag.flag;
+        }
+        final String flagAsString = SamFlagField.STRING.format(flagAsInteger);
+
+        Assert.assertEquals(flagAsString, "urURpP12sSxd");
+        Assert.assertEquals(SamFlagField.STRING.parse(flagAsString), flagAsInteger);
+    }
+
+    @Test
+    public void testAllFlagsReverseOrder() {
+        int flagAsInteger = 0;
+        for (final SAMFlag samFlag : SAMFlag.values()) {
+            flagAsInteger |= samFlag.flag;
+        }
+        final String flagAsString = new StringBuilder("urURpP12sSxd").reverse().toString();
+
+        Assert.assertEquals(flagAsInteger, SamFlagField.STRING.parse(flagAsString));
+        Assert.assertEquals(SamFlagField.STRING.parse(flagAsString), flagAsInteger);
+    }
+
+    @Test
+    public void testForwardStrandFlags() {
+        final int flagAsInteger = SamFlagField.STRING.parse("f");
+        final String flagAsString = SamFlagField.STRING.format(flagAsInteger);
+
+        Assert.assertEquals(flagAsString, "mf");
+        Assert.assertEquals(SamFlagField.STRING.parse(flagAsString), flagAsInteger);
+    }
+
+    @Test
+    public void testPairedForwardStrandFlags() {
+        final int flagAsInteger = SamFlagField.STRING.parse("mfMFp");
+        final String flagAsString = SamFlagField.STRING.format(flagAsInteger);
+
+        Assert.assertEquals(flagAsString, "mfMFp");
+        Assert.assertEquals(SamFlagField.STRING.parse(flagAsString), flagAsInteger);
+    }
+
+    @Test
+    public void testMappedFlags() {
+        final int flagAsInteger = SamFlagField.STRING.parse("m");
+        final String flagAsString = SamFlagField.STRING.format(flagAsInteger);
+
+        Assert.assertEquals(flagAsString, "mf");
+        Assert.assertEquals(SamFlagField.STRING.parse(flagAsString), flagAsInteger);
+    }
+
+    @Test
+    public void testPairedMappedFlags() {
+        final int flagAsInteger = SamFlagField.STRING.parse("pmM");
+        final String flagAsString = SamFlagField.STRING.format(flagAsInteger);
+
+        Assert.assertEquals(flagAsString, "mfMFp");
+        Assert.assertEquals(SamFlagField.STRING.parse(flagAsString), flagAsInteger);
+    }
+
+    @Test
+    public void testMateMappedNotOnFragmentFlags() {
+        final int flagAsInteger = SAMFlag.MATE_UNMAPPED.flag;
+        final String flagAsString = SamFlagField.STRING.format(flagAsInteger);
+
+        Assert.assertEquals(flagAsString, "mfU");
+        Assert.assertEquals(SamFlagField.STRING.parse(flagAsString), flagAsInteger);
+    }
+
+    @Test
+    public void testMateMappedOnlyOnPairsFlags() {
+        final int flagAsInteger = SAMFlag.MATE_UNMAPPED.flag | SAMFlag.READ_PAIRED.flag;
+        final String flagAsString = SamFlagField.STRING.format(flagAsInteger);
+
+        Assert.assertEquals(flagAsString, "mfUFp");
+        Assert.assertEquals(SamFlagField.STRING.parse(flagAsString), flagAsInteger);
+    }
+
+    @Test
+    public void testMateForwardStrandNotOnFragmentFlags() {
+        final int flagAsInteger = 0;
+        final String flagAsString = SamFlagField.STRING.format(flagAsInteger);
+
+        Assert.assertEquals(flagAsString, "mf");
+        Assert.assertEquals(SamFlagField.STRING.parse(flagAsString), flagAsInteger);
+    }
+
+    @Test
+    public void testMateForwardStrandOnlyOnPairsFlags() {
+        final int flagAsInteger = SAMFlag.READ_PAIRED.flag;
+        final String flagAsString = SamFlagField.STRING.format(flagAsInteger);
+
+        Assert.assertEquals(flagAsString, "mfMFp");
+        Assert.assertEquals(SamFlagField.STRING.parse(flagAsString), flagAsInteger);
+    }
+
+    @Test
+    public void testFlagTypesParsing() {
+        Assert.assertEquals(SamFlagField.of("0"), SamFlagField.DECIMAL);
+        Assert.assertEquals(SamFlagField.of("1234"), SamFlagField.DECIMAL);
+        Assert.assertEquals(SamFlagField.of("0xDOESNOTMATTER"), SamFlagField.HEXADECIMAL);
+        Assert.assertEquals(SamFlagField.of("0x"), SamFlagField.HEXADECIMAL);
+        Assert.assertEquals(SamFlagField.of("0[^x]DOESNOTMATTER"), SamFlagField.OCTAL);
+        Assert.assertEquals(SamFlagField.of("0a"), SamFlagField.OCTAL);
+        Assert.assertEquals(SamFlagField.of("DOESNOTMATTER"), SamFlagField.STRING);
+    }
+
+    @Test
+    public void testFlagTypesFormatting() {
+
+        Assert.assertEquals(SamFlagField.DECIMAL.format(1), "1");
+        Assert.assertEquals(SamFlagField.DECIMAL.format(124), "124");
+
+        Assert.assertEquals(SamFlagField.HEXADECIMAL.format(1), "0x1");
+        Assert.assertEquals(SamFlagField.HEXADECIMAL.format(9), "0x9");
+        Assert.assertEquals(SamFlagField.HEXADECIMAL.format(10), "0xa");
+        Assert.assertEquals(SamFlagField.HEXADECIMAL.format(16), "0x10");
+
+        Assert.assertEquals(SamFlagField.OCTAL.format(1), "01");
+        Assert.assertEquals(SamFlagField.OCTAL.format(124), "0174");
+
+        Assert.assertEquals(SamFlagField.STRING.format(337), "mrMFp1s");
+    }
+    
+    @Test(expectedExceptions = SAMFormatException.class)
+    public void testIllegalStringFlagCharacter(){
+        SamFlagField.STRING.parse("HELLO WORLD");
+    }
+
+    @Test(expectedExceptions = SAMFormatException.class)
+    public void testIllegalHexadecimalFlagCharacter(){
+        SamFlagField.HEXADECIMAL.parse("HELLO WORLD");
+    }
+
+    @Test(expectedExceptions = SAMFormatException.class)
+    public void testIllegalStringFlagCharacterExclamation(){
+        SamFlagField.STRING.parse("pmMr!F1s");
+    }
+}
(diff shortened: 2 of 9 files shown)


nh13:
@jmarshall any objections to adding this functionality?  I really like it in ```samtools```.

src/java/htsjdk/samtools/SAMTextWriter.java
```diff
@@ -93,7 +154,19 @@ public void writeAlignment(final SAMRecord alignment) {
         try {
             out.write(alignment.getReadName());
             out.write(FIELD_SEPARATOR);
-            out.write(Integer.toString(alignment.getFlags()));
+            switch (this.flagFieldsOutput) {
```
mcovarr: Warning: quite possibly more trouble than it's worth, but this could be converted to polymorphism in the enum instances:

```
public enum FlagFieldsOutput {
  ...
  HEXADECIMAL {
    @Override
    public String format(int flags) {
      return String.format("0x%x", alignment.getFlags());
    }
    public abstract String format(int flags);
  },
  ...
}
```

Then this code becomes:

```
  out.write(flagFieldsOutput.format(alignment.getFlags()));
```

src/java/htsjdk/samtools/SAMFileWriterFactory.java
```diff
@@ -142,6 +143,15 @@ public SAMFileWriterFactory setTempDirectory(final File tmpDir) {
     }
 
     /**
+     * Set the flag output format only when writing tet.
```
vdauwera: Writing "tet"? Typo or am I missing something?

mcovarr:
:+1:

src/java/htsjdk/samtools/SAMTextWriter.java
```diff
@@ -45,13 +45,70 @@
...
+        STRING {
+            private final String flag2CharTable = "pPuUrR12sxdS\0\0\0\0"; // when the bit is set
+            private final String notFlag2CharTable = "\0\0mMfF\0\0\0\0\0\0\0\0\0\0"; // when the bit is not set
+
+            private String getStringFlags(final int flag) {
+                String s = "";
+                for (int i = 0; i < 16; ++i) {
+                    if ((flag & 1 << i) != 0) {
+                        if ('\0' != flag2CharTable.charAt(i)) s += flag2CharTable.charAt(i);
+                        else if ('\0' != notFlag2CharTable.charAt(i)) s += notFlag2CharTable.charAt(i);
```
jmarshall: This is on the wrong `if`

jmarshall:
My concern with this functionality has always been that [it is underspecified](http://sourceforge.net/p/samtools/mailman/message/23461794/).  I have long since given up on the canonical representation thing, but preventing the possibility of an empty field for a flag value of 0 remains IMHO vital.

I see that (modulo the bug I added a line note for, and contrary to the outdated letter descriptions in the _In a string FLAG_ comment added) you have prevented an empty field with `mMfF` flags.

I like this functionality too, which is why I [implemented it in my own C++ BAM IO library](https://github.com/jmarshall/cansam/blob/master/tools/samcat.1#L95-116) several years ago (see jmarshall/cansam@804e5190858e31b896cd550fdc525146f5593a2b).  There are two differences between the flag letters that cansam implements and that you propose here:

* For quality control failure, old-samtools used `f` (which we want to repurpose for _forward_), I used `q` and you are proposing `x`.  I'd like to suggest you use `q` too.

* You've added `mM` as the opposite of `uU`.  Note that you'll always print one of `f` and `r`, so if those are read as _mapped to forward/reverse_ then `m` is redundant.  So it seems to me that `mM` are clutter and a waste of a letter, but YMMV.

I support adding similar functionality in htslib, though it too would have to break with old-samtools and use `f` for _forward_.  So I would support this htsjdk change if you change `x` to `q`, as then all three implementations can be compatible.  (I wouldn't much want to emit `mM` in cansam or htslib, but I would probably accept them on input.  Though it would be a shame if the next flag we add to the spec could naturally be abbreviated as `m`!)

As a quality of implementation matter, while it is temptingly easy to just loop over the bits from 0 to 15 and print out bit letters, this leads to unnecessarily crappy output.  For an unmapped read, you will print `uf` meaning _unmapped on the forward strand?!_; for single-ended reads, you will print stuff like `mrMF` even though `MF` is meaningless for this read.  See [cansam's sam_format()](https://github.com/jmarshall/cansam/blob/master/lib/alignment.cpp#L653-659) for an approach that preserves information (e.g. `ur` for unmapped but stupidly set reverse bit) but suppresses meaningless information (e.g. single ended `f` instead of `fF`).

nh13:
@jmarshall perfect, this is just what I was looking for.  Thank-you!

jmarshall:
I may be having second thoughts about `q` vs. `x`… @nh13 what were your reasons for selecting `x` for bit 0x200?

nh13:
@jmarshall I must have pulled it from an older version of samtools or just made it up.  I am not sure which I like better.

The major use of this specification, is for bioinformaticians to be able to quickly interpret the flag field as a string, so I would only go as far as to support writing this to output.  I would not want this to be part of the SAM spec either, though it could be a best-practice for a "string" output.

Next, ```mM``` and ```fF``` do not have a 1:1 relationship, since we could have a set of reads that are unmapped but also have the ```0x10``` or ```0x20``` bit set.  If I recall correctly, we updated the spec to allow for this recently.

jmarshall:
> I must have pulled q vs. x from an older version of samtools or just made it up. I am not sure which I like better.

I picked `q` as **q**uality failure, but to the extent that this flag might become a general-purpose filtered-out flag, `x` would be preferable — see samtools/hts-specs#85.

> Next, mM and fF do not have a 1:1 relationship, since we could have a set of reads that are unmapped but also have the 0x10 or 0x20 bit set. If I recall correctly, we updated the spec to allow for this recently.

We updated the spec merely so that idiotic bwa output would be well-defined.  This does not mean that unmapped but with 0x10 set reads are a meaningful or important use-case, and in fact we discouraged them (see samtools/hts-specs#68).  But whatevs.

nh13:
@vdauwera any update?

vdauwera:
@nh13 None from my end. As far as I know this is between you and @jmarshall ...

nh13:
@mcovarr one last thumbs up?  I just added some more comments for the non-set bits.

jmarshall:
FWIW :+1: from me with `x` for (quality) failure.

**Except** that it still contains the same bug I pointed out previously.  Have you tested this?  Did it ever actually print out any of `mMfF`?

src/java/htsjdk/samtools/SAMTextWriter.java
mcovarr: This doesn't look right to me either, and given the way the two flag tables are initialized the `else` appears to be unreachable.

src/java/htsjdk/samtools/SAMTextWriter.java
```diff
@@ -45,13 +45,76 @@
...
+            private final String notFlag2CharTable = "\0\0mMfF\0\0\0\0\0\0\0\0\0\0"; // when the bit is not set
+
+            private String getStringFlags(final int flag) {
+                String s = "";
+                for (int i = 0; i < 16; ++i) {
+                    if ((flag & 1 << i) != 0) { // the bit is set
+                        if ('\0' != flag2CharTable.charAt(i)) s += flag2CharTable.charAt(i);
+                    }
+                    else if ('\0' != notFlag2CharTable.charAt(i)) { // the bit is not set
+                        if ('\0' != notFlag2CharTable.charAt(i)) s += notFlag2CharTable.charAt(i);
```
mcovarr: This needlessly performs the same `'\0' != notFlag2CharTable.charAt(i)` test twice.

src/java/htsjdk/samtools/SAMTextWriter.java
```diff
@@ -45,13 +45,76 @@
...
+
+            private String getStringFlags(final int flag) {
+                String s = "";
+                for (int i = 0; i < 16; ++i) {
+                    if ((flag & 1 << i) != 0) { // the bit is set
+                        if ('\0' != flag2CharTable.charAt(i)) s += flag2CharTable.charAt(i);
+                    }
+                    else if ('\0' != notFlag2CharTable.charAt(i)) { // the bit is not set
+                        s += notFlag2CharTable.charAt(i);
+                     }
```
jmarshall: Hooray!  :smile:   But we might as well go for 100% clarity:

```java
if ((flag & 1 << i) != 0) {
    if ('\0' != flag2CharTable.charAt(i)) s += flag2CharTable.charAt(i);
}
else {
    if ('\0' != notFlag2CharTable.charAt(i)) s += notFlag2CharTable.charAt(i);
}
```

nh13:
@jmarshall I was thinking of adding some tests, and the easiest way to do that would be to be able to *read* the string flag field.  This seems dangerous if there is no consensus convention.  What are your thoughts?

jmarshall:
I only asked about tests to the extent that if you were testing it you would probably notice that the previous `mMfF` code was not working.  But that bug is fixed now…

I thought you did not support implementing _reading_ this? :stuck_out_tongue:

nh13:
I wanted to write a few tests, and that lead to trying to support reading and writing: https://github.com/samtools/htsjdk/tree/nh_flag_fields_format_now_with_tests

nh13:
I pushed the reading and writing to hear (eek).  I feel torn about whether to think about putting support in Picard for this.

jmarshall:
Okay, with my serious hat on… I support reading symbolic flags too, which is why I implemented it in my own C++ BAM IO library several years ago (see jmarshall/cansam@804e5190858e31b896cd550fdc525146f5593a2b).

IMHO the necessary minimal consensus would be:

1. that the FLAG field remains unambiguously parseable;
2. that we agree on the values of the letters `pPuUrR12sxdSmMfF`;
3. that we don't make additional requirements on the letters' ordering etc.

(1) implies that when writing a symbolic flag, implementations must not start the output with a digit.  Your code will always write either `u` or `m` before it writes `1` or `2`, so that is alright.  (Apart from can't-start-with-digit and can't-be-empty, I don't think we need to have any constraints about letter ordering.)

I think it is a serious mistake to consult `SAM_FLAG_FIELD_FORMAT` when reading.  Your code will interpret `12` as either unmapped/mate-unmapped (decimal) or proper-paired/reversed (hex) or first/second (symbolic) depending on how `SAM_FLAG_FIELD_FORMAT` is set.  This violates (1) and is **not** a feature.

It also turns using different formats for the FLAG column on different lines into an error.  Far better to recognise that decimal/hex/symbolic FLAG data is naturally self-identifying and parse accordingly: starts with `0x` ⇒ hex; else starts with `0` ⇒ octal; else starts with digit ⇒ decimal; otherwise ⇒ symbolic.  (You might prefer to omit octal, but there's no reason not to parse it, no?)

src/java/htsjdk/samtools/SamFlagField.java
```diff
@@ -0,0 +1,102 @@
...
+        private int parseStringFlags(final String flag) {
+            int ret = 0;
+            for (int i = 0; i < flag.length(); i++) {
+                final int idx = flag2CharTable.indexOf(flag.charAt(i));
+                if (-1 != idx) {
+                    ret = ret | (1 << idx);
+                }
+            }
+            return ret;
+        }
```
jmarshall: When parsing, should unknown letters be ignored (as here) or produce an exception?

akiezun:
assigning to @nh13 to make a move

nh13:
@jmarshall I am off the opinion we should not mix flag formats within a file, you ok with this further restriction?

jmarshall:
>  I am off the opinion we should not mix flag formats within a file

Why?

But let's start with the important bit.

Do you agree that SAM files should be self-contained?  In particular, do you agree that someone seeing `12` in a FLAG field in an alignment record should be safe in assuming that this means unmapped/mate-unmapped?

nh13:
@vadimzalunin opening this again and restoring the branch.

nh13:
@jmarshall I am going to try one more time to push this PR through.  I have updated the implementation to respect 1-3 as you suggest, as I think you are very correct about these, especially about automatically recognizing the flag format per record.  Thank-you for the thoughtful review.

My only concern is where we put the definition of the letters that represent the flag values.   I suppose we could add to columns to 1.4.2, `set letter` and `unset letter`.

src/java/htsjdk/samtools/SamFlagField.java
```diff
@@ -0,0 +1,130 @@
...
+        @Override
+        public String format(final int flag) {
+            throw new SAMException("NONE not allowed for the SamFlagField when writing the SAM flag field.");
+        }
+        @Override
+        public int parse(final String flag) {
+            throw new SAMException("NONE not allowed for the SamFlagField when reading the SAM flag field.");
+        } 
+    },
+    DEFAULT {
```
akiezun: It's a bit strange to have an constant named default. I'd rather call it `DECIMAL` and have the default use be `DECIMAL`

src/java/htsjdk/samtools/SamFlagField.java
```diff
@@ -0,0 +1,130 @@
...
+        public String format(final int flag) {
+            return String.format("%#x", flag);
+        }
+        @Override
+        public int parse(final String flag) {
+            if (flag.startsWith("0x")) return Integer.valueOf(flag.substring(2), 16);
+            else return Integer.valueOf(flag, 16); // this really should not be supported, it should always have 0x..
+        }
+    },
+    OCTAL {
```
akiezun: if we have decimal, hexadecimal and octal, it seems logical to add binary (and flags are naturally binary)

src/java/htsjdk/samtools/SAMTextWriter.java
```diff
@@ -80,8 +76,42 @@ public Writer getWriter() {
...
+
+    /**
+     * Constructs a SAMTextWriter that writes to an OutputStream.  The OutputStream
+     * is wrapped in an AsciiWriter, which can be retrieved with getWriter().
+     * @param stream Need not be buffered because this class provides buffering.
+     */
+    public SAMTextWriter(final OutputStream stream, final SamFlagField samFlagFieldOutput) {
         this.file = null;
         this.out = new AsciiWriter(stream);
+        this.samFlagFieldOutput = samFlagFieldOutput;
```
akiezun: `samFlagFieldOutput` must not be null as per code in `writeAlignment` - check it in all constructors and blow up early

src/java/htsjdk/samtools/SAMLineParser.java
```diff
@@ -432,6 +453,10 @@ private RuntimeException reportFatalErrorParsingLine(final String reason) {
         return new SAMFormatException(makeErrorString(reason));
     }
 
+    private RuntimeException reportFatalErrorParsingLine(final String reason, final Throwable throwable) {
```
akiezun: why have this as a method and not just throw at the call site?

src/java/htsjdk/samtools/SAMLineParser.java
```diff
@@ -170,6 +177,21 @@ private int parseInt(final String s, final String fieldName) {
         }
         return ret;
     }
+    
+    private int parseFlag(final String s, final String fieldName) {
+        final int ret;
+        final SamFlagField samFlagField = this.samFlagField.orElse(SamFlagField.getSamFlagField(s));
+        try {
+            ret = samFlagField.parse(s);
```
akiezun: I'd find it easier to read like this (ie eliminate the useless temp declarations outside of the `try` block)
```
try{
   final SamFlagField samFlagField = this.samFlagField.orElse(SamFlagField.getSamFlagField(s));
   return samFlagField.parse(s);
} catch ...
```

src/java/htsjdk/samtools/SAMLineParser.java
```diff
@@ -160,6 +159,14 @@ public ValidationStringency getValidationStringency() {
         return this.validationStringency;
     }
 
+    /**
+     * Sets the expected SAM flag type expected for all records.
```
akiezun: Document that `samFlagField` can be `null`

src/java/htsjdk/samtools/SAMFileWriterFactory.java
```diff
@@ -178,6 +179,15 @@ public SAMFileWriterFactory setTempDirectory(final File tmpDir) {
     }
 
     /**
+     * Set the flag output format only when writing text.
+     * Default value: [[htsjdk.samtools.SAMTextWriter.samFlagFieldOutput.DEFAULT]]
+     */
+    public SAMFileWriterFactory setSamFlagFieldOutput(final SamFlagField samFlagFieldOutput) {
```
akiezun: I think passing null here will result in a blow up later on. If that's true then check non null here and blow up early. Otherwise, if null is OK, then document that null is OK.

src/java/htsjdk/samtools/SamFlagField.java
```diff
@@ -0,0 +1,130 @@
...
+
+    // For testing
+    public String getFlag2CharTable() { return null; }
+    public String getNotFlag2CharTable() { return null; }
+
+    /**
+     * Returns the type of flag field for this string.  This does not guarantee it is of the flag field,
+     * as it only checks the first two characters.
+     */
+    public static SamFlagField getSamFlagField(final String s) {
```
akiezun: i"d rename this to `create` or `of`. Having calls like
```
  SamFlagField sff = SamFlagField.getSamFlagField(s);
```
just reads bad to me.

nh13:
@akiezun thanks for the review.  Should I consider your review complete as I am still assigned, or are you going to comment further?

akiezun:
my review is complete. Maybe others will add more as part of the review party.

src/java/htsjdk/samtools/SAMFileWriterFactory.java
nh13: will throw a `IllegalArgumentException`; thanks for pointing this out.

src/java/htsjdk/samtools/SAMLineParser.java
nh13: It can't be `null`.  I'll do a check here.

src/java/htsjdk/samtools/SAMLineParser.java
nh13: I will change it to:
```
    private int parseFlag(final String s, final String fieldName) {
        final SamFlagField samFlagField = this.samFlagField.orElse(SamFlagField.getSamFlagField(s));
        try {
            return samFlagField.parse(s);
        } catch (NumberFormatException e) {
            throw reportFatalErrorParsingLine("Non-numeric value in " + fieldName + " column");
        } catch (SAMFormatException e) {
            throw reportFatalErrorParsingLine("Error in " + fieldName + " column: " + e.getMessage(), e);
        }
    }
```
The reason is for me is to clearly identify which method call may causing the exception to be thrown.

src/java/htsjdk/samtools/SAMLineParser.java
nh13: Following the convention at line 460.  It avoids me having to write ```SAMFormatException(makeErrorString(...), e)``` at every call site.

src/java/htsjdk/samtools/SAMTextWriter.java
nh13: I will do it, but the other objects are not being checked for null here, so it seems like overkill.

src/java/htsjdk/samtools/SamFlagField.java
nh13: I will leave that as a future improvement, as the functionality I am trying to re-create was in `samtools` originally, and did not have binary.  Also, this may confuse the auto-format recognition code depending on the binary output format.

src/java/htsjdk/samtools/SamFlagField.java
nh13: done.

src/java/htsjdk/samtools/SamFlagField.java
nh13: Done

src/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -206,6 +206,7 @@ public SamReaderFactory validationStringency(final ValidationStringency validati
             this.validationStringency = validationStringency;
             return this;
         }
+
```
yfarjoun: extra space?

src/tests/java/htsjdk/samtools/SAMTextWriterTest.java
```diff
@@ -56,7 +51,35 @@ public void testNullHeader() throws Exception {
...
+
     private void doTest(final SAMRecordSetBuilder recordSetBuilder) throws Exception{
+        doTest(recordSetBuilder, SamFlagField.DECIMAL);
+    }
+
+    private void doTest(final SamFlagField samFlagField) throws Exception {
+        doTest(getSAMReader(true, SAMFileHeader.SortOrder.coordinate), samFlagField);
+    }
+
+    private void doTest(final SAMRecordSetBuilder recordSetBuilder, final SamFlagField samFlagField) throws Exception{
```
yfarjoun: ) {

src/tests/java/htsjdk/samtools/SAMTextWriterTest.java
```diff
@@ -56,7 +51,35 @@ public void testNullHeader() throws Exception {
...
+    public void testBasicOctalFlag() throws Exception {
+        doTest(SamFlagField.OCTAL);
+    }
+
+    @Test
+    public void testBasicStringFlag() throws Exception {
+        doTest(SamFlagField.STRING);
+    }
+
     private void doTest(final SAMRecordSetBuilder recordSetBuilder) throws Exception{
```
yfarjoun: ) {

src/java/htsjdk/samtools/SamFlagField.java
```diff
@@ -0,0 +1,130 @@
+package htsjdk.samtools;
```
yfarjoun: license

yfarjoun:
Please respond to new comments. Could you please not merge since we have a critical bug in the latest release. (wait until 2.2.1 is out, thanks!)

src/java/htsjdk/samtools/SamReaderFactory.java
nh13: :/

nh13:
@yfarjoun responded to all your comments (looks good).  I would like @jmarshall 's input before this gets merged.

droazen:
Re-assigning to @nh13 -- since two weeks have elapsed and there's no reply from @jmarshall, and all comments have been addressed, can we merge this?

nh13:
@droazen please do not re-assign.  Since this functionality affects both https://github.com/samtools/samtools and htsjdk, I would like @jmarshall to review.

droazen:
@nh13 In that case, could you contact @jmarshall through a means other than an at-mention to ask for a review? This has been sitting here for a long time.

nh13:
I have been emailing him and he has been agreed to review soon. Please be patient. I don't see this blocking others.

jmarshall:
@droazen: he has.  Relax, dude!  Also 10 days ago ≠ two weeks.

droazen:
@nh13 @jmarshall Ok, thanks, just making sure that this wasn't stalled! Sorry for the annoyance, but (as you may know) we've instituted a biweekly review of the 5 oldest pull requests in htsjdk in an effort to reduce the PR backlog (https://github.com/samtools/htsjdk/issues/548), and as part of this effort we have the somewhat unenviable job of contacting PR authors to inquire about status, and closing PRs when there's no reply after a reasonable amount of time has elapsed.

src/java/htsjdk/samtools/Defaults.java
```diff
@@ -87,7 +90,7 @@
      * A mask (pattern) to use when building EBI reference service URL for a
      * given MD5 checksum. Must contain one and only one string placeholder.
      */
-    public static final String EBI_REFERENCE_SEVICE_URL_MASK;
+    public static final String EBI_REFERENCE_SERVICE_URL_MASK;
```
jmarshall: As an outside observer, not sure how this unrelated change got in here…

src/java/htsjdk/samtools/SamFlagField.java
```diff
@@ -0,0 +1,154 @@
...
+    },
+    HEXADECIMAL {
+        @Override
+        public String format(final int flag) {
+            return String.format("%#x", flag);
+        }
+        @Override
+        public int parse(final String flag) {
+            if (flag.startsWith("0x")) return Integer.valueOf(flag.substring(2), 16);
+            else return Integer.valueOf(flag, 16); // this really should not be supported, it should always have 0x..
```
jmarshall: :+1: × :100: to this comment!

jmarshall:
With one minor and one major caveat, this is a :+1: from me.

**1.** Writing the flags as a string still suffers from the issue I noted last April (see _As a quality of implementation matter…_ above), but this is a minor issue that is partly a matter of taste.

**2.** The major problem is that using `SAMLineParser.withSamFlagField()` just has the effect of asking for your data to be misinterpreted.  Using it to set a fixed parsing flag radix means that a flag column of `12` will mean something different depending on what you set the radix to, and that SAM records will not be self-contained in this regard — see my previous rounds of comments for why I think this matters so much.

I think what you want this to be used for is to say "all records in this file must have HEX flags", or "…STRING flags" or etc.  Then IMHO you should implement it so that after `withSamFlagField(HEXADECIMAL)`, `parseFlag()` accepts only flag strings starting with `0[Xx]` and throws an exception otherwise.  Similarly after `withSamFlagField(STRING)`, `parseFlag()` checks that the flag string starts with a non-digit and throws otherwise.  Etc.

For example, in the current implementation (if I understand Java correctly!), `DECIMAL.parse()` throws if it sees a hex or symbolic flag (though doesn't but should if it sees octal, if you're going to support that).  The `else` in `HEXADECIMAL.parse()` should be changed to throw.  `STRING.parse()` should throw if the first character is a digit.

Then setting a radix with `withSamFlagField()` will have the effect of "all flags must be this", rather than "I will interpret the text of all flags as this, whether they are or not".

src/java/htsjdk/samtools/SamFlagField.java
```diff
@@ -0,0 +1,154 @@
...
+        public int parse(final String flag) { return Integer.parseInt(flag); }
+    },
+    HEXADECIMAL {
+        @Override
+        public String format(final int flag) {
+            return String.format("%#x", flag);
+        }
+        @Override
+        public int parse(final String flag) {
+            if (flag.startsWith("0x")) return Integer.valueOf(flag.substring(2), 16);
```
jmarshall: or `"0X"`

src/java/htsjdk/samtools/Defaults.java
nh13: I found a type, fixing it.

src/java/htsjdk/samtools/SamFlagField.java
nh13: good, I will change it then.

src/java/htsjdk/samtools/SamFlagField.java
nh13: done.

nh13:
@jmarshall I made all the changes you suggested, thank-you.  I think the major difference between the implementation I made and https://github.com/jmarshall/cansam/blob/master/lib/alignment.cpp is the former only outputs the paired flags (ex. mate reverse, mate unmapped) if the "read paired" flag is set, while the latter will always output paired flags (ex. `FIRST_IN_PAIR`), even if the `PAIRED` is not set.

Thanks for the 👍  and then I will squash and rebase.

yfarjoun:
Hey @jmarshall we're here at the review party and would love if you could give this a final pass. Sorry to bug you, Thanks!

droazen:
@nh13 @jmarshall We're at our bi-weekly htsjdk review party, and this ticket has once again come up as the oldest unresolved PR. Does @nh13 have the green light to squash and merge this?

yfarjoun:
@nh13 Given that @jmarshall is not responding to this. I would like to try to resolve the last issue. Given that the output should be descriptive rather than interpretive, why not emit all set flags even when the "paired" is not set? This seems especially important when debugging for example if the bam is in a strange state e.g. MATE_UNMAPPED is set but read isn't PAIRED...it seems to me that just like the numerical output doesn't change the output based on the meaning, neither should the human-readable output...but I''d love to hear your reasoning.

nh13:
@yfarjoun to be honest, I can't recall fully my reasoning, but likely I am guarding against outputting flags that don't make sense together.  To get this pushed through, I can make it output all flags like @jmarshall's implementation.  This needs a rebase, so I will do that first, and then add in the second commit as you suggest.  Once approved, I will squash and do a final rebase.

nh13:
@yfarjoun back to you.

yfarjoun:
tests are failing....I'm not sure if you can see the output of Travis...so here is the salient bit:

```code

Gradle suite > Gradle test > htsjdk.samtools.SamFlagFieldTest.testMappedFlags FAILED
    java.lang.AssertionError: expected [mf] but found [mfMF]
        at org.testng.Assert.fail(Assert.java:94)
        at org.testng.Assert.failNotEquals(Assert.java:496)
        at org.testng.Assert.assertEquals(Assert.java:125)
        at org.testng.Assert.assertEquals(Assert.java:178)
        at org.testng.Assert.assertEquals(Assert.java:188)
        at htsjdk.samtools.SamFlagFieldTest.testMappedFlags(SamFlagFieldTest.java:55)
Gradle suite > Gradle test > htsjdk.samtools.SamFlagFieldTest.testForwardStrandFlags FAILED
    java.lang.AssertionError: expected [mf] but found [mfMF]
        at org.testng.Assert.fail(Assert.java:94)
        at org.testng.Assert.failNotEquals(Assert.java:496)
        at org.testng.Assert.assertEquals(Assert.java:125)
        at org.testng.Assert.assertEquals(Assert.java:178)
        at org.testng.Assert.assertEquals(Assert.java:188)
        at htsjdk.samtools.SamFlagFieldTest.testForwardStrandFlags(SamFlagFieldTest.java:39)
Gradle suite > Gradle test > htsjdk.samtools.SamFlagFieldTest.testMateMappedNotOnFragmentFlags FAILED
    java.lang.AssertionError: expected [mf] but found [mfUF]
        at org.testng.Assert.fail(Assert.java:94)
        at org.testng.Assert.failNotEquals(Assert.java:496)
        at org.testng.Assert.assertEquals(Assert.java:125)
        at org.testng.Assert.assertEquals(Assert.java:178)
        at org.testng.Assert.assertEquals(Assert.java:188)
        at htsjdk.samtools.SamFlagFieldTest.testMateMappedNotOnFragmentFlags(SamFlagFieldTest.java:71)
Gradle suite > Gradle test > htsjdk.samtools.SamFlagFieldTest.testMateForwardStrandNotOnFragmentFlags FAILED
    java.lang.AssertionError: expected [mf] but found [mfMF]
        at org.testng.Assert.fail(Assert.java:94)
        at org.testng.Assert.failNotEquals(Assert.java:496)
        at org.testng.Assert.assertEquals(Assert.java:125)
        at org.testng.Assert.assertEquals(Assert.java:178)
        at org.testng.Assert.assertEquals(Assert.java:188)
        at htsjdk.samtools.SamFlagFieldTest.testMateForwardStrandNotOnFragmentFlags(SamFlagFieldTest.java:87)
```

nh13:
I couldn't get the single test Gradle build working since there was a cram test error (see my other PR).

nh13:
See issue https://github.com/samtools/htsjdk/issues/621 for what I am getting when trying to execute a single test on my system.  I am relying on travis to run the tests for now :/.

jmarshall:
This time around I have only reviewed _SamFlagField.java_'s `format()` and `parseWithoutValidation()`.

I did not expect Nils to rewrite `format()` to follow my Cansam code, but since this has come to pass I am (unsurprisingly) happy with the algorithm.  I agree with @yfarjoun that a numeric→string→numeric round trip needs to preserve even strange flag combinations.

One caveat: I thought we had agreed to use `x` for READ_FAILS_VENDOR_QUALITY_CHECK as in your previous versions (e.g., [fe5ca92](https://github.com/samtools/htsjdk/blob/fe5ca9240176616f118db1d3dede5ba7eeb1ed32/src/java/htsjdk/samtools/SamFlagField.java#L89))?

yfarjoun:
@jmarshall from the cpp code
```cpp
 case 'q':  value |= QUALITY_FAILED;  break;
```
do we want the htsjdk and the samtools version to be different?

jmarshall:
@yfarjoun: Reread the comments on this PR from [last May](#issuecomment-98075683).

yfarjoun:
got it. thanks.

yfarjoun:
@nh13: can you switch back to using `x`? AFAICT this is the last gating issue and then we can be done with this PR.

nh13:
@yfarjoun reverted to using `x` (see https://github.com/samtools/htsjdk/pull/232/commits/72caa68f554ee2885d899230de8f4699593928b7#diff-7081faf5e0aab98ed981739dcedf82e2R116), squashed and rebased.

src/main/java/htsjdk/samtools/SamFlagField.java
```diff
@@ -0,0 +1,203 @@
...
+            else if ((flag & SAMFlag.READ_PAIRED.flag) != 0)                value.append('F');
+
+            if ((flag & SAMFlag.READ_PAIRED.flag) != 0)                     value.append('p');
+            if ((flag & SAMFlag.PROPER_PAIR.flag) != 0)                     value.append('P');
+            if ((flag & SAMFlag.FIRST_OF_PAIR.flag) != 0)                   value.append('1');
+            if ((flag & SAMFlag.SECOND_OF_PAIR.flag) != 0)                  value.append('2');
+
+            if ((flag & SAMFlag.NOT_PRIMARY_ALIGNMENT.flag) != 0)           value.append('s');
+            if ((flag & SAMFlag.SUPPLEMENTARY_ALIGNMENT.flag) != 0)         value.append('S');
+            if ((flag & SAMFlag.READ_FAILS_VENDOR_QUALITY_CHECK.flag) != 0) value.append('x');
```
nh13: reverted to `x`

src/main/java/htsjdk/samtools/SamFlagField.java
```diff
@@ -0,0 +1,203 @@
...
+                    case 'p':  value |= SAMFlag.READ_PAIRED.flag;  break;
+                    case 'P':  value |= SAMFlag.PROPER_PAIR.flag;  break;
+                    case 'u':  value |= SAMFlag.READ_UNMAPPED.flag;  break;
+                    case 'U':  value |= SAMFlag.MATE_UNMAPPED.flag;  break;
+                    case 'r':  value |= SAMFlag.READ_REVERSE_STRAND.flag;  break;
+                    case 'R':  value |= SAMFlag.MATE_REVERSE_STRAND.flag;  break;
+                    case '1':  value |= SAMFlag.FIRST_OF_PAIR.flag;  break;
+                    case '2':  value |= SAMFlag.SECOND_OF_PAIR.flag;  break;
+                    case 's':  value |= SAMFlag.NOT_PRIMARY_ALIGNMENT.flag;  break;
+                    case 'q':  value |= SAMFlag.READ_FAILS_VENDOR_QUALITY_CHECK.flag;  break;
```
yfarjoun: should this be x as well?

src/main/java/htsjdk/samtools/SamFlagField.java
yfarjoun: this implies that there's no round-trip test... 🙈

yfarjoun: wrong place.

src/main/java/htsjdk/samtools/SamFlagField.java
yfarjoun: this implies that there's no round-trip test... 🙈

src/test/java/htsjdk/samtools/SamFlagFieldTest.java
```diff
@@ -0,0 +1,140 @@
...
+        Assert.assertEquals(flagAsString, "urURpP12sSxd");
+    }
+
+    @Test
+    public void testAllFlagsReverseOrder() {
+        int flagAsInteger = 0;
+        for (final SAMFlag samFlag : SAMFlag.values()) {
+            flagAsInteger |= samFlag.flag;
+        }
+        final String flagAsString = new StringBuilder("urURpP12sSqd").reverse().toString();
```
yfarjoun: q?

yfarjoun: this shouldn't pass should it?

nh13: I need to learn this gradle business if I am to test locally.  Off to learn gradle.

nh13:
@yfarjoun added round-trip tests and fixed up the `x`.

yfarjoun:
👍 though we should probably wait until SRA stops failing. 🚒

yfarjoun:
tests pass, @nh13, I think we can merge this!

Issue #183, opened 2015-02-26
Support VN:1.5


jeremiahsavage:
I've used biobambam to merge some files, and it ouputs
`@HD VN:1.5 SO:coordinate`
using https://github.com/gt1/libmaus/blob/master/src/libmaus/bambam/BamCatHeader.hpp#L116

According to the spec ( http://samtools.github.io/hts-specs/SAMv1.pdf ), 1.5 is the current version, but picard ValidateSamFile gives an error, which the diff below eliminates.

```
diff --git a/src/java/htsjdk/samtools/SAMFileHeader.java b/src/java/htsjdk/samtools/SAMFileHeader.java
index 2e237c4..ac361b1 100644
--- a/src/java/htsjdk/samtools/SAMFileHeader.java
+++ b/src/java/htsjdk/samtools/SAMFileHeader.java
@@ -46,9 +46,9 @@ public class SAMFileHeader extends AbstractSAMHeaderRecord
     public static final String VERSION_TAG = "VN";
     public static final String SORT_ORDER_TAG = "SO";
     public static final String GROUP_ORDER_TAG = "GO";
-    public static final String CURRENT_VERSION = "1.4";
+    public static final String CURRENT_VERSION = "1.5";
     public static final Set<String> ACCEPTABLE_VERSIONS =
-            new HashSet<String>(Arrays.asList("1.0", "1.3", "1.4"));
+            new HashSet<String>(Arrays.asList("1.0", "1.3", "1.4", "1.5"));

     /**
      * These tags are of known type, so don't need a type field in the text representation.
```

vdauwera:
@nh13 Do I understand correctly that this is just a matter of adding "1.5" to the list of acceptable versions? Or is there any actual dev/validation work required?

yfarjoun:
It would be good to know what the differences are between 1.4 and 1.5 and
make sure (i.e. test?) that we actually support 1.5 before we do this...

no?

vdauwera:
You tell me, I don't have a sense of how much changes between spec versions. Would be awfully convenient if there was an official release corresponding to each version of the spec, with exotic things like, say, release notes spelling out the important differences.

jmarshall:
That would indeed be convenient (and is samtools/hts-specs#47), but like many convenient things it is temporarily thwarted by the existence of day jobs.

[...]

PR #233, opened 2015-04-30
#183 - Changes to accommodate v1.5 of the SAM file specification

nenewell:
Changes to SamFileHeader.java, ValidateSamFileTest.java, and test files in SamFileHeaderMergerTest to accommodate v1.5 of the SAM file spec.


src/tests/java/htsjdk/samtools/ValidateSamFileTest.java
```diff
@@ -437,7 +437,7 @@ public void testHeaderVersion(final String version, final boolean expectValid) t
                 {"1.0", true},
                 {"1.3", true},
                 {"1.4", true},
-                {"1.5", false},
+                {"1.5", true}
```
nh13: We probably want to add ```{"1.6", false}``` to the list to make sure we fail on version 1.6.

nh13:
Thanks @nenewell.  I would suggest you squash the two commits and modify the commit so it only changes the two ```.java``` files, and not the ```.idea``` and ```*iml``` files.

nenewell:
Thank you - I’ve updated the pull request.

nenewell:
I see this has failed a Travis build. A recent build of a pull request of mine for hellbender failed due to a security issue because I am an external source - this was resolved by lbergelson (#478).

lbergelson:
I don't think this is due to the same issue.  It looks like it might be a transient issue with downloading something from an FTP site.  Rerunning to see if it goes away.

lbergelson:
Rerun solved the issue.

nenewell:
Thank you.

vdauwera:
Does anything more need to be done on this? @geoffjentry, could you opine?

geoffjentry:
This might just need to be rebased on top of current htsjdk master but I'm getting test failures in picard when I use this htsjdk

nenewell:
I will rebase.

nenewell:
samRoundTrip test is failing. I'll fix once I've updated my system with ant. I don't understand why the build passed earlier test runs...

nenewell:
The samRoundTrip test failure is fixed, and "VN:1.4" has been replaced with "VN1.5" in the headers of 21 other test files, even though these do not cause htsjdk test failures, in case they may cause test failures in the tool builds. But testLazyInstance failed in one of the four Travis runs - a "lazy" date object seems to have gotten the time wrong by 130 seconds...

geoffjentry:
I'm still seeing test failures with picard when using this htsjdk

nenewell:
@geoffjentry, could you please send a log file with the errors to nenewell@comcast.net? 
It would be great if there were a way for me to build the tools together with htsjdk, so you folks wouldn't have to be bothered with this.

geoffjentry:
@nenewell You should be able to do it, I'm talking about the publicly available picard tools (https://github.com/broadinstitute/picard).

If you clone that repo and then have your htsjdk sitting in that top level dir you should be able to run 'ant test'

nenewell:
Thanks - I will do that.

geoffjentry:
@nenewell Great. Let me know if you encounter difficulties

nenewell:
The picard test failures were caused by the presence of the old version number in the VN fields of test files - I've updated those and submitted the changes as picard pull request #211. So picard should be OK - I still get six test failures, but they must be related to my setup and not the SAM file version update because I also get them when I use unmodified htsjdk and picard code from the Broad trunk. I can't access non-picard tool development, but any version-specific test files there will also need to be updated...

nenewell:
The current state on this, to the best of my understanding, is that the necessary htsjdk updates are handled by pull request 233, and the necessary picard updates are handled by picard pull request 211 (which fails Travis because htsjdk changes are not yet merged). I don't know if any other tools need updating - if so, I'd be happy to do that, but I don't know if I would be able to access them. Please let me know if you'd like me to rebase this pull request and picard request 211, or if there's anything else I can do on this...Thank you

jmarshall:
Is the change to `ACCEPTABLE_VERSIONS` at least likely to be merged any time soon?  Fortunately samtools does not yet generate VN:1.5 either, but it's been in the spec for two years and probably other tools do…

nh13:
@nenewell looks like #211 includes a patch for hdfs, which goes beyond just fixing tests for the new version in picard.

nenewell:
The link in my post of May 19 points to htsjdk #211 (not picard 211) due to git’s default behavior –  I didn’t yet know that prepending ‘#’ to a number automatically produces a link to that request in the current repo.

nh13:
Agreed it was my mistake as I deleted the comment after I made it. 

Thumb typed for added typos

akiezun:
@geoffjentry can you review and decide what @nenewell needs to do here?

nh13:
Note: Impact to production systems need to be considered and tested before this is rolled out.  How will projects that have mixes of 1.4 versus 1.5 work etc?  I am sure others may face the same problem.  This is an internal Broad question to answer, so lets take it offline, but it unit tests to verify our desired behavior in the case seem necessary.

nh13:
Closing this and moving the PR discussion to #285

PR #235, opened 2015-05-04
Mark SAMRecord and VariantContext (and dependent classes) as Serializable

droazen:
These classes can be trivially serialized, with the exception of the
SAMFileSource field in SAMRecord, which is now marked as transient.

Added tests to prove that serialization/deserialization work.


droazen:
For @lbergelson to review, please

lbergelson:
I think we should update doc for the attributes to say "If you put in a thing that isn't serializable, then this won't be serializable"

droazen:
@akiezun should weigh in as well

lbergelson:
It looks good to me.  Massive helper methods for tiny change.  We need someone to run picard private on this.

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -175,7 +177,7 @@
      */
     private ValidationStringency mValidationStringency = ValidationStringency.SILENT;
 
-    private SAMFileSource mFileSource;
+    private transient SAMFileSource mFileSource;
```
lbergelson: I'd add a line comment here explaining why this is transient.

mattsooknah:
I will run picard-private on this.

lbergelson:
thanks!

src/tests/java/htsjdk/variant/VariantBaseTest.java
```diff
@@ -119,4 +123,121 @@ public static SAMSequenceDictionary createArtificialSequenceDictionary() {
...
+        Assert.assertEquals(actual.getLikelihoods(), expected.getLikelihoods(), "Genotype getLikelihoods");
+        Assert.assertTrue(Arrays.equals(actual.getPL(), expected.getPL()));
+
+        Assert.assertEquals(actual.getGQ(), expected.getGQ(), "Genotype phredScaledQual");
+        assertAttributesEquals(actual.getExtendedAttributes(), expected.getExtendedAttributes());
+        Assert.assertEquals(actual.isPhased(), expected.isPhased(), "Genotype isPhased");
+        Assert.assertEquals(actual.getPloidy(), expected.getPloidy(), "Genotype getPloidy");
+    }
+
+    private static void assertAttributesEquals(final Map<String, Object> actual, Map<String, Object> expected) {
```
akiezun: this seems non-trivial. can you add a comment to explain the logic here?

src/tests/java/htsjdk/variant/VariantBaseTest.java
```diff
@@ -119,4 +123,121 @@ public static SAMSequenceDictionary createArtificialSequenceDictionary() {
...
+        // now expectedKeys contains only the keys found in expected but not in actual,
+        // and they must all be null
+        for ( final String missingExpected : expectedKeys ) {
+            final Object value = expected.get(missingExpected);
+            Assert.assertTrue(isMissing(value), "Attribute " + missingExpected + " missing in one but not in other" );
+        }
+    }
+
+    private static void assertAttributeEquals(final String key, final Object actual, final Object expected) {
+        if ( expected instanceof Double ) {
```
akiezun: do we need it for `float` too?

src/tests/java/htsjdk/variant/VariantBaseTest.java
```diff
@@ -119,4 +123,121 @@ public static SAMSequenceDictionary createArtificialSequenceDictionary() {
...
+            // must be very tolerant because doubles are being rounded to 2 sig figs
+            assertEqualsDoubleSmart(actual, (Double) expected, 1e-2);
+        }
+        else {
+            Assert.assertEquals(actual, expected, "Attribute " + key);
+        }
+    }
+
+    private static boolean isMissing(final Object value) {
```
akiezun: can you comment the logic here?

src/tests/java/htsjdk/variant/variantcontext/VariantContextUnitTest.java
```diff
@@ -989,4 +993,18 @@ public void testSetAttribute() {
...
+    @Test
+    public void testSerialization() throws Exception {
+        final VariantContext initialVC = new VariantContextBuilder("serialization_test", "chr1", 1, 1, Arrays.asList(Aref, C, T)).genotypes(GenotypeBuilder.create("AA", Arrays.asList(Aref, Aref))).make();
+
+        final ByteArrayOutputStream byteArrayStream = new ByteArrayOutputStream();
+        final ObjectOutputStream out = new ObjectOutputStream(byteArrayStream);
+        out.writeObject(initialVC);
+        out.close();
+
+        final ObjectInputStream in = new ObjectInputStream(new ByteArrayInputStream(byteArrayStream.toByteArray()));
```
akiezun: does `in` need to be closed?

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+
+public class SAMRecordUnitTest {
+
+    @DataProvider(name = "serializationTestData")
+    public Object[][] getSerializationTestData() {
+        return new Object[][] {
+                { new File("testdata/htsjdk/samtools/serialization_test.sam") },
+                { new File("testdata/htsjdk/samtools/serialization_test.bam") }
```
akiezun: CRAM?

src/tests/java/htsjdk/variant/VariantBaseTest.java
lbergelson: This stuff is all just lifted out of hellbender.  I was just kind of assuming it's functioning as expected...

src/java/htsjdk/samtools/SAMRecord.java
droazen: Done

src/tests/java/htsjdk/variant/VariantBaseTest.java
droazen: Will add a comment anyway.

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -41,7 +42,9 @@
  *
  * @author depristo
  */
-public final class CommonInfo {
+public final class CommonInfo implements Serializable {
```
akiezun: what about the attributes in CommonInfo?

The declaration is `private Map<String, Object>`
serialization will break on this - Object can be anything.

src/tests/java/htsjdk/variant/VariantBaseTest.java
droazen: Not for the purposes of the test suite, which is where this code lives.

src/tests/java/htsjdk/variant/VariantBaseTest.java
droazen: Done

src/tests/java/htsjdk/variant/variantcontext/VariantContextUnitTest.java
droazen: Not really, but will close it anyway.

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
droazen: For serialization purposes we just need to test `SAMRecord` and its one subclass `BAMRecord`

src/java/htsjdk/variant/variantcontext/CommonInfo.java
lbergelson: The options are to either change it to `Map<String, Serializable>` which could potentially be a breaking change for people using serializable attributes, or to add a doc string saying "these must be serializable or serialization will fail"  which will not break any ones code, but means that serialization might fail.  We were leaning towards the "document and force it onto the user" policy since it seems controversial to potentially break anyone's code.

src/java/htsjdk/variant/variantcontext/GenotypesContext.java
```diff
@@ -39,7 +40,9 @@
 /**
  * Represents an ordered collection of Genotype objects
  */
-public class GenotypesContext implements List<Genotype> {
+public class GenotypesContext implements List<Genotype>, Serializable {
```
akiezun: this class has a subclass LazyGenotypesContext that has a field `Object unparsedGenotypeData;` - this is not serializable

It also has a field of type LazyParser that has 2 implementations:
- one is a member inner class of AbstractVCFCodec (and serializing it would pull in all of AbstractVCFCodec and its superclasses)
- the other is a BCF2LazyGenotypesDecoder which refers to a lot of stuff which would need to be serializable

akiezun:
i'm done reviewing: 2 issues with serializing fields of type Object and one with a subclass that pulls in a ton of stuff.

src/java/htsjdk/variant/variantcontext/CommonInfo.java
droazen: Yes, I've added docs to the `VariantContextBuilder` attribute setting methods warning that attribute values must be serializable or else serialization will fail. Unfortunately we're probably not at liberty to change the actual type from `Object` to `Serializable`, or to throw if a value if not an instance of `Serializable`, as these would be breaking changes for client code.

src/java/htsjdk/variant/variantcontext/GenotypesContext.java
droazen: `unparsedGenotypeData` is serializable in practice, since it's always of type `String` or `byte[]`

droazen: Will look into the `LazyParser` issue

droazen: Have modified the tests so that they now detect the issue with `LazyParser` not being `Serializable`. Will work on a fix. Nice catch!

droazen:
All comments addressed -- @akiezun and @lbergelson please re-review.

I've worked around the `LazyGenotypesContext` issue via custom serialization, and modified tests to use real VCF/BCF data with lazily-parsed genotypes. Confirmed that these lazy genotypes are now getting serialized/deserialized correctly.

droazen:
@mattsooknah Would you mind re-testing the latest version of this against picard(-private)? Fixed an important bug in my last push. Thanks!

akiezun:
@droazen this LGTM

mattsooknah:
All tests pass on picard public + private. LGTM

PR #238, opened 2015-05-05
fixed an overflow bug in the output when indexing large bam

yfarjoun:
this fixes a small bug in the output of the BamIndexer which presents itself on large BAMs.,


jacarey:
I'm surprised they weren't longs to begin with.

:+1:

yfarjoun:
I think that the problem is that it isn't using the ProgressLogger...which would be the right way to do it....out of scope!

PR #239, opened 2015-05-06
Make VCFHeader Serializable

droazen:
This was missed in the previous pull request to make SAM/Variant
records Serializable. We want to be able to serialize VCFHeaders
just as we do SAMFileHeaders

### DIFF
--- a/src/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
+++ b/src/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
@@ -249,16 +249,34 @@ public abstract class VCFCompoundHeaderLine extends VCFHeaderLine implements VCF
     }
 
     /**
-     * returns true if we're equal to another compounder header line
+     * returns true if we're equal to another compound header line
      * @param o a compound header line
      * @return true if equal
      */
-    public boolean equals(Object o) {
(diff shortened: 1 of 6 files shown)


droazen:
For @lbergelson to review please

lbergelson:
@droazen Looks good to me.
@mattsooknah Can you run whatever needs to be done to test this for picard?

droazen:
Strange, the serialization test passes for me locally but fails on travis. Will investigate

lbergelson:
What's the failure say?

droazen:
It says the header metadata mismatches, but the expected/actual output is identical, and the same assert passes locally for me in both Java 7 and 8.

lbergelson:
Is it failing only on a specific jvm, or in all of them on travis?

droazen:
All of them -- am experimenting now to try to narrow down the problem.

droazen:
Believe I've fixed it -- the various classes extending `VCFHeaderLine` overrode `equals()` but not `hashCode()`, violating contract, and their `equals()` methods did not properly chain up the class hierarchy. @lbergelson re-review at your leisure.

mattsooknah:
:+1: in terms of picard public/private compatibility.

lbergelson:
@mattsooknah Thanks again!

lbergelson:
@droazen :+1:

PR #240, opened 2015-05-06
Added a couple of methods to IntervaList to centralize "padding" done…

tfenne:
… right.

### DIFF
--- a/src/java/htsjdk/samtools/util/IntervalList.java
+++ b/src/java/htsjdk/samtools/util/IntervalList.java
@@ -98,6 +98,26 @@ public class IntervalList implements Iterable<Interval> {
         this.header.setSortOrder(SAMFileHeader.SortOrder.coordinate);
     }
 
+    /** Returns a new IntervalList where each interval is padded by the specified amount of bases. */
+    public IntervalList padded(final int before, final int after) {
+        if (before < 0 || after < 0) throw new IllegalArgumentException("Padding values must be >= 0.");
+        final IntervalList padded = new IntervalList(this.getHeader().clone());
+        final SAMSequenceDictionary dict = padded.getHeader().getSequenceDictionary();
+        for (final Interval i : this) {
+            final SAMSequenceRecord seq = dict.getSequence(i.getContig());
+            final int start = Math.max(1, i.getStart() - before);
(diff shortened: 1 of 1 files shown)


tfenne:
Hey @yfarjoun - I got tired of writing and re-writing code to pad interval lists, so I added methods akin to sorted() and uniqued() to IntervalList itself.  Can you take a quick look please?  Once this is merged I have a similar small PR for Picard to use this instead of it's own code in IntevalListTools

src/java/htsjdk/samtools/util/IntervalList.java
```diff
@@ -98,6 +98,25 @@ public void sort() {
         this.header.setSortOrder(SAMFileHeader.SortOrder.coordinate);
     }
 
+    /** Returns a new IntervalList where each interval is padded by the specified amount of bases. */
+    public IntervalList padded(final int before, final int after) {
+        if (before < 0 || after < 0) throw new IllegalArgumentException("Padding values must be >= 0.");
+        final IntervalList padded = new IntervalList(this.getHeader());
```
yfarjoun: clone the header to avoid changing the old header

src/java/htsjdk/samtools/util/IntervalList.java
```diff
@@ -98,6 +98,25 @@ public void sort() {
...
+        final SAMSequenceDictionary dict = padded.getHeader().getSequenceDictionary();
+        for (final Interval i : this) {
+            final int start = Math.max(1, i.getStart()-before);
+            final int end   = Math.min(dict.getSequence(i.getContig()).getSequenceLength(), i.getEnd()+after);
```
yfarjoun: Since we do not check that the contig of an .added() interval is in the dictionary... should you check here to avoid a NPE?

yfarjoun:
review done. :+1: pending two small requests

tfenne:
Just pushed an update that address your two concerns @yfarjoun .  I dislike adding code to null check the sequence record - we really should just assert that intervals are valid when they are added.

yfarjoun:
I agree with your disdain, and was going to make that change...but didn't get around to it. Feel free to submit a pull request doing that :-) 

:+1: for this PR once test pass, feel free to merge.

Issue #229, opened 2015-04-27
Increase number of likelihoods allowed in PL field


vdauwera:
Needed to fix https://github.com/broadinstitute/gsa-unstable/issues/937

There is a bug occurring in GATK (reported independently several times already) due to large numbers (hundreds) of alternative alleles in the processed loci in some users' GVCF files. Assuming there are N alleles in diploid organism, then VCF file stores N * (N - 1) / 2 likelihoods in the PL field. However, htsjdk reads only first 10,000 likelihoods:

LINE 755-766 in htsjdk.variant.vcf.AbstractVCFCoded.java:

    private final String[] INT_DECODE_ARRAY = new String[10000];
    private final int[] decodeInts(final String string) {
    final int nValues = ParsingUtils.split(string, INT_DECODE_ARRAY, ',');
    final int[] values = new int[nValues];
    try {
    for ( int i = 0; i < nValues; i++ )
    values[i] = Integer.valueOf(INT_DECODE_ARRAY[i]);
    } catch (final NumberFormatException e) {
    return null;
    }
    return values;
    }

Since GATK assumes that there are N * (N - 1) / 2 likelihoods, then it eventually jumps out of the 10,000 array.

Unless there is a good reason for hardcoding a 10K limit, we'd like to relax this to allow considering larger numbers of alternate alleles, which becomes more likely to happen now that we're calling indels in very large cohorts.

nh13:
@vdauwera gsa-unstable is not public, do you want to link that here?

vdauwera:
I'm including the link for the internal team's convenience. Unless it's frowned upon to do this, I don't see the harm?

nh13:
It doesn't allow the community to comment or be involved when the details are hidden. 

N

Thumb typed for added typos

vdauwera:
Well that's why I copied the issue description almost verbatim. There are no details hidden/missing. 

If people want to see the original bug reports, see the support forum here: http://gatkforums.broadinstitute.org/discussion/comment/20470#Comment_20470

nh13:
Perfect thank you!

Thumb typed for added typos

vdauwera:
@ronlevine is taking this one via fork

ronlevine:
Do you have the original input variant from the user?

vdauwera:
Yep, see filepath in https://github.com/broadinstitute/gsa-unstable/issues/856

ronlevine:
[...]

PR #241, opened 2015-05-07
Increase size of decodeInts array

ronlevine:
Solution for https://github.com/samtools/htsjdk/issues/229.

### DIFF
--- a/src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
+++ b/src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
@@ -39,6 +39,7 @@ import htsjdk.variant.variantcontext.GenotypeLikelihoods;
 import htsjdk.variant.variantcontext.LazyGenotypesContext;
 import htsjdk.variant.variantcontext.VariantContext;
 import htsjdk.variant.variantcontext.VariantContextBuilder;
+import org.apache.commons.lang3.StringUtils;
 
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
@@ -752,13 +753,15 @@ public abstract class AbstractVCFCodec extends AsciiFeatureCodec<VariantContext>
         return new LazyGenotypesContext.LazyData(genotypes, header.getSampleNamesInOrder(), header.getSampleNameToOffset());
     }
 
-    private final String[] INT_DECODE_ARRAY = new String[10000];
     private final int[] decodeInts(final String string) {
-        final int nValues = ParsingUtils.split(string, INT_DECODE_ARRAY, ',');
+        // find number of CSV items to set array size
+        int numItems = StringUtils.countMatches(string, ",") + 1;
+        String[] intDecodeArray = new String[numItems];
+        final int nValues = ParsingUtils.split(string, intDecodeArray, ',');
         final int[] values = new int[nValues];
         try {
             for ( int i = 0; i < nValues; i++ )
-                values[i] = Integer.valueOf(INT_DECODE_ARRAY[i]);
+                values[i] = Integer.valueOf(intDecodeArray[i]);
         } catch (final NumberFormatException e) {
             return null;
         }
--- a/pom.xml
+++ b/pom.xml
@@ -118,6 +118,11 @@
         <artifactId>commons-jexl</artifactId>
         <version>2.1.1</version>
       </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.4</version>
+        </dependency>
       <dependency>
         <groupId>commons-logging</groupId>
         <artifactId>commons-logging</artifactId>
@@ -145,6 +150,10 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-jexl</artifactId>
     </dependency>
+      <dependency>
+          <groupId>org.apache.commons</groupId>
+          <artifactId>commons-lang3</artifactId>
+      </dependency>
     <dependency>
       <groupId>commons-logging</groupId>
       <artifactId>commons-logging</artifactId>


src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -752,9 +755,12 @@ private final static boolean isVCFStream(final InputStream stream, final String
         return new LazyGenotypesContext.LazyData(genotypes, header.getSampleNamesInOrder(), header.getSampleNameToOffset());
     }
 
-    private final String[] INT_DECODE_ARRAY = new String[10000];
+    private final int LARGE_SIZE = 10000;
+    private final String[] INT_DECODE_ARRAY = new String[2*LARGE_SIZE];
     private final int[] decodeInts(final String string) {
         final int nValues = ParsingUtils.split(string, INT_DECODE_ARRAY, ',');
+        if ( nValues > LARGE_SIZE )
+            log.warn("Decoding " + nValues + " integers in " + name);
```
vdauwera: I think this warning needs to be more clear, eg add something like "This warning is associated with decoding genotype annotations that have a very large number of values, typically due to the presence of a very large number of alternate alleles." so anyone who encounters it in the future has some idea of what's happening.

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -752,9 +755,12 @@ private final static boolean isVCFStream(final InputStream stream, final String
         return new LazyGenotypesContext.LazyData(genotypes, header.getSampleNamesInOrder(), header.getSampleNameToOffset());
     }
 
-    private final String[] INT_DECODE_ARRAY = new String[10000];
+    private final int LARGE_SIZE = 10000;
+    private final String[] INT_DECODE_ARRAY = new String[2*LARGE_SIZE];
```
vdauwera: Please add some comments in the code clarifying why we are putting this in.

ronlevine: I am not sure how specific I should get with this one. I don't  think there should be any mention of PL. I was thinking about something like "For warning if there is a very large number of genotype fields due to an excessive alternate alleles".

vdauwera:
This change pushes the limit for number of alleles up to ~200. That's enough to resolve this particular case, but I wonder if it will be general enough in the long run. So here are a few (admittedly code-naive) questions/thoughts:

1) What would be the performance hit of increasing the number further? Do we have an easy way to quantify this?

2) Is there a way to guesstimate the order of magnitude needed based on the length of the string that we are asking ParsingUtils.split to parse?

3) What would be the performance hit of counting commas in the string (since the number of separators +1 gives us the number of elements we need to read in)?

4) If none of this is a good option, is there a way for tools (like GATK) that utilize htsjdk to override the htsjdk's hardcoded value, so that we can have a user-customizable setting for this?

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
vdauwera: Agreed, it shouldn't specifically mention PLs. Other genotype fields may cause the same issue.

ronlevine:
1) Should not be a performance hit, it would increase the memory footprint
2) I was looking into that but have yet to find a great answer
3) I was thinking the same yesterday. It's scalable and is worth try.

ronlevine:
@vdauwera You can take a look at my implementation of 3).

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -752,13 +753,19 @@ private final static boolean isVCFStream(final InputStream stream, final String
         return new LazyGenotypesContext.LazyData(genotypes, header.getSampleNamesInOrder(), header.getSampleNameToOffset());
     }
 
-    private final String[] INT_DECODE_ARRAY = new String[10000];
+
     private final int[] decodeInts(final String string) {
-        final int nValues = ParsingUtils.split(string, INT_DECODE_ARRAY, ',');
+        // find number of CSV items to set array size
+        int numItems = string.split(",").length;
```
vdauwera: Might it be slightly faster to do `int numItems = StringUtils.countMatches(string, ",") +1;`?

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -752,13 +753,19 @@ private final static boolean isVCFStream(final InputStream stream, final String
...
+        String[] intDecodeArray = new String[numItems];
+        final int nValues = ParsingUtils.split(string, intDecodeArray, ',');
+        if ( nValues != numItems ){
```
vdauwera: This seems redundant -- I'm not sure it's possible for the condition to not be met unless something goes grievously wrong with one of the steps above...?

ronlevine: This is defensive programming, or in this case paranoid programming. It will be removed.

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
ronlevine: Yes, both would have to parse the entire string but countMatches() avoids the construction and copy into the array of strings. I am having trouble accessing org.apache.commons.lang.StringUtils (which contains countMatches()) instead of com.sun.deploy.util.StringUtils. Looks like a classpath problem. Looked in pom.xml and found org.apache.commons dependency...

vdauwera: Naive question -- can you not just slap `import org.apache.commons.lang.StringUtils;` in the import statements block?

ronlevine: Unfortunately, no. I wish maven was that smart and would import it. In a vanilla build system, you could do that and just add it to the classpath. I might've found the answer, the pom file does not have the artifactId for the jar containing this functionality, only commons-jexl. It looks like I need commons-lang (https://commons.apache.org/proper/commons-lang/)

ronlevine: The pom file modification worked! Back in business...

vdauwera: Yay :)

jacarey: I should note that the pom is very stale and we don't use it for builds at all. It will likely be removed. In general we try to avoid importing new libraries to htsjdk to keep it lean as a policy.

ronlevine: So, you are saying I should not use org.apache.commons.lang.StringUtils.countMatches()?

ronlevine: How is it typically built?

ronlevine:
@vdauwera Addressed comments, ready for review.

vdauwera:
@geoffjentry @jacarey Would one of you please review this, taking into account the discussion in #229, and let us know what would be the Right Thing To Do in your view?

geoffjentry:
I think I need to be convinced that the commons lib needs to happen here

vdauwera:
@geoffjentry I'm happy to take any solution that is equivalent to doing 
    
     int numItems = StringUtils.countMatches(string, ",") + 1;

that doesn't use the new lib. I have no clue what is better in terms of performance. (I'm a doctor, dammit, not a programmer -- although I play one on tv)

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
jacarey: Ant.

For pushing signed releases to oss we use sbt. To future confusion we will likely be removing the pom.

ronlevine:
What is the downside of including the lib? HTSJDK is already using the highly specialized commons-jexl. The upside is that it would greatly save development time and minimize coding errors in the future.

ronlevine:
I found apache's implementation of countMatches(). We could *clone* this code:
http://www.java2s.com/Code/Java/Data-Type/CountshowmanytimesthesubstringappearsinthelargerString.htm

ronlevine:
@vdauwera I can *clone* apache's countMatches() and not use the entire apache commons lib, but I was concerned about the license agreement.

vdauwera:
Yeah we want to make sure we respect the license. I don't know what the apache license entails. 

Meanwhile though our user johnwallace opened a PR here: https://github.com/samtools/htsjdk/pull/246

Can you take a look and see if it complements or supersedes yours?

geoffjentry:
@ronlevine I didn't say that I was dead set against it, just that I needed to be convinced beyond "it makes this one thing easier to write". As you pointed out we do take in external deps but our bar for doing so is intentionally high.

ronlevine:
@geoffjentry I understand. Note that HTSJDK StringUtil.isBlank() was previously cloned from the commons lib:
http://www.java2s.com/Tutorial/Java/0040__Data-Type/ChecksifaStringiswhitespaceemptyornull.htm
We could do the same for countMatches(), but as a mentioned in a previous comment, I am concerned about the Apache license agreement.

ronlevine:
@vdauwera https://github.com/samtools/htsjdk/pull/246 could work after a small fix and unit testing.

geoffjentry:
@ronlevine Hi Ron - we talked about it internally and are ok w/ going ahead w/ the commons lib. We need it to be buildable via the standard ant build though before we can proceed with merging your request.

ronlevine:
@geoffjentry Is there anything for me to do on my end in terms of ant? I did not see anything in build.xml. I will also plan on removing the HTSJDK's *clone* of StringUtils.isBlank().

geoffjentry:
@ronlevine the jar needs to be physically present in htsjdk/lib

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -752,13 +753,15 @@ private final static boolean isVCFStream(final InputStream stream, final String
         return new LazyGenotypesContext.LazyData(genotypes, header.getSampleNamesInOrder(), header.getSampleNameToOffset());
     }
 
-    private final String[] INT_DECODE_ARRAY = new String[10000];
     private final int[] decodeInts(final String string) {
```
bpow: This method could be static if it does not use the INT_DECODE_ARRAY field anymore

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -752,13 +753,15 @@ private final static boolean isVCFStream(final InputStream stream, final String
...
-        final int nValues = ParsingUtils.split(string, INT_DECODE_ARRAY, ',');
+        // find number of CSV items to set array size
+        int numItems = StringUtils.countMatches(string, ",") + 1;
+        String[] intDecodeArray = new String[numItems];
+        final int nValues = ParsingUtils.split(string, intDecodeArray, ',');
```
bpow: If I understand, the reason for avoiding `String.split` was poor performance that was fixed in JDK7. Since htsjdk support for JDK6 is planned to be dropped in 5 months, is it worth it to add an additional dependency now that may be moot in that time?

(It would take some benchmarking to confirm my suspicions).

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
ronlevine: Why should it be a static utility/helper method? It's private and not general purpose?

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
ronlevine: This is from HTSJDK's own class, ParsingUtils, not Apache commons. It also does not use regexes.

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
bpow: I prefer to make methods that do not depend on any instance variables as static, in part to implicitly convey that fact.

At least in theory, invocation of static methods could be slightly faster because they obviate a null check for the instance. I'm not sure how the JVM implements this in practice.

ronlevine:
Superseded by https://github.com/samtools/htsjdk/pull/248.

PR #242, opened 2015-05-07
N on N should be a match, not a mismatch

geoffjentry:


yfarjoun:
hmmm. do you know why the exception was put in in the first time?

jacarey:
:+1:  
This is what I get for rebasing the CRAM code base.

geoffjentry:
@yfarjoun Because it's what samtools does. @eitanbanks and @ktibbett decided that the spec was ambiguous and that we'd be going this way instead in order to match GATK (See PO-1190 for the whole conversation)

eitanbanks:
Just wanted to point out that this has nothing to do with the GATK.  There are 2 methods in SequenceUtil for calculating the value of the NM tag: one (the original) counts N-on-N as a match and the other (copied from samtools) treats it as a mismatch.  The spec is unclear as to which is right so we decided to go with the original Picard version (esp. as the samtools version is so incredibly ugly).

yfarjoun:
Thanks all for the edification!

Issue #196, opened 2015-03-16
Bug when parsing spliced BAM record with very large intron


s-andrews:
I've had a bug report against our software which appears to be an underlying bug in the htsjdk parser for BAM files.  The read which fails to parse is a spliced read with a huge intron in it (226Mbp).  When parsing the cigar string htsjdk corrupts this value (it looks like it overflows so I get a negative value), but if I use samtools to convert the bam file to sam, then htsjdk reads it just fine.

I've put the example files I used (only 1 read in them) up at:

http://ftp2.babraham.ac.uk/ftpusr68/

Some code which illustrates the problem is below.  For the sam file I get:
Cigar is 109M226643263N25M17S Start=10002 End=226653398

..but for the equivalent BAM file I get:
Ignoring SAM validation error: ERROR: Record 1, Read name GR_10257_idx_5/2, bin field of BAM record does not equal value computed based on alignment start and end, and length of sequence to which read is aligned
Cigar is 109M-41792193N25M17S Start=10002 End=-41782058

This is using htsjdk-1.129.jar

import java.io.File;

import htsjdk.samtools.SAMRecord;
import htsjdk.samtools.SamReader;
import htsjdk.samtools.SamReaderFactory;
import htsjdk.samtools.ValidationStringency;

public class PicardCigarBug {
	
	public static void main (String [] args) {
		
		File smallBugBam = new File("C:/Users/andrewss/Desktop/Bug/small.bam");
//		File smallBugBam = new File("C:/Users/andrewss/Desktop/Bug/small.sam");
		
		SamReaderFactory samFactory = SamReaderFactory.makeDefault().validationStringency(ValidationStringency.LENIENT);
		SamReader inputSam = samFactory.open(smallBugBam); 

		
		for (SAMRecord samRecord : inputSam) {
			System.err.println("Cigar is "+samRecord.getCigarString()+" Start="+samRecord.getAlignmentStart()+" End="+samRecord.getAlignmentEnd());
		}
	}
}

vdauwera:
Oops, didn't mean to close this -- wrong button. 

@s-andrews I agree this looks like an overflow. Forgive the naive question, but does it make biological sense to want to try to handle such a large intron? To be honest my instinct would be to treat this as a bad input and filter out this read.

[...]

PR #244, opened 2015-05-08
Issue #196 - fixes a bug which causes an overflow in the CIGAR when a…

nenewell:
… BAM file is parsed that contains a read that spans a very large intron. The bug is caused by the use of an arithmetic (>>) rather than a logical (>>>) right shift operator in BinaryCigarCodec.binaryCigarToCigarElement(). This fix substitutes the logical for the arithmetic operator, and implements a test which reads the sample BAM file (containing a single record) that exposed the bug and compares the indexing bin field with the computed indexing bin field. If the two fields are different, an error is thrown.


vdauwera:
Looks good, just checking with the user who reported the issue that we can use his bam for the test.

nenewell:
Rebased.

vdauwera:
#196

vdauwera:
Ack, user hasn't responded to my request for permission to include the test file in the codebase. Emailed him directly. I'm probably being overly careful but it's better than the opposite...

nenewell:
Agreed.

vdauwera:
Got confirmation that it's cool. Merging.

nenewell:
Thank you – I’m on the board!

PR #245, opened 2015-05-09
Added checks to adding an interval to an IntervalList

yfarjoun:
Contig must be in header. Note that one could still forcefully modify the underlying dictionary, so the protection is only partial.

Removed protection code from the "padding" function since now contig should be in header.


yfarjoun:
@tfenne I added the contig checking to IntervalList and changed your code back to be less paranoid.

Care to take a glance?

tfenne:
Looks good to me! Thanks for doing this.

PR #246, opened 2015-05-11
Fix for Issue #229 (Unlimited PL)

johnwallace123:
The following code should allow for the creation of arbitrarily-sized output arrays (now vectors) from the ParsingUtils.split() function, which will allow for the resolution of Issue #229 with minor modifications to the AbstractVCFCodec.

I have not yet included unit tests for the new split() functionality in ParsingUtils, but I will if it is determined that this is a good direction to go.

The specifics of what I did:
- Added a split() function in ParsingUtils that takes a vector
as output to allow for theoretically unlimited tokens in a
delimited string
- Modified the AbstractVCFCodec to utilize the above function.

### DIFF
--- a/src/java/htsjdk/tribble/util/ParsingUtils.java
+++ b/src/java/htsjdk/tribble/util/ParsingUtils.java
@@ -38,6 +38,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Vector;
 import java.util.WeakHashMap;
 
 /**
@@ -181,6 +182,54 @@ public class ParsingUtils {
         return split(aString, tokens, delim, false);
     }
 
+    /**
+     * Split the string into tokens given by the delimiter.  Will use a 
+     * Vector<String> to allow for dynamically resizing of the tokens array.
+     * Note that with careful pre-allocation of the tokens Vector, this should
+     * be no less efficient than the split(String, String[] char) function.
+     * Also, with this function, we do not need a split(String, Vector<String>, char, bool)
+     * function, as there is no maximum number of tokens.
+     * 
+     * @param sSrting the string to split
+     * @param tokens  a Vector<String> to hold the parsed tokens
+     * @param delim   character that delimits tokens
+     * @return the number of tkens parsed
+     */
+    public static int split(String aString, Vector<String> tokens, char delim) {
+        
+    	int nTokens = 0;
+        int start = 0;
+        int end = aString.indexOf(delim);
+        
+        // Empty out our answer, but preserve the memory allocation
+        tokens.clear();
+        
+        while(end >= 0){
+        	// Only add non-empty tokens
+        	if(start != end){
+        		tokens.add(aString.substring(start, end));
+        		start = end+1;
+        	}
+        	start = end+1;
+        	
+    		// If this condition is true, we likely ended with a token, so
+        	// manually set end = -1 to break the loop
+    		if(start >= aString.length()){
+    			end = -1;
+    		} else {
+    			end = aString.indexOf(delim, start);
+    		}
+
+        }
+        
+        // Add the final element (if non-empty)
+        if(start < aString.length()){
+        	tokens.add(aString.substring(start));
+        }
+    
+        return tokens.size();
+    }
+    
     /**
      * Split the string into tokens separated by the given delimiter.  Profiling has
      * revealed that the standard string.split() method typically takes > 1/2
--- a/src/java/htsjdk/tribble/util/ParsingUtils.java
+++ b/src/java/htsjdk/tribble/util/ParsingUtils.java
@@ -38,7 +38,6 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.Vector;
 import java.util.WeakHashMap;
 
 /**
@@ -195,21 +194,18 @@ public class ParsingUtils {
      * @param delim   character that delimits tokens
      * @return the number of tkens parsed
      */
-    public static int split(String aString, Vector<String> tokens, char delim) {
+    public static int split(String aString, List<String> tokens, char delim) {
         
-    	int nTokens = 0;
-        int start = 0;
+    	int start = 0;
         int end = aString.indexOf(delim);
         
         // Empty out our answer, but preserve the memory allocation
         tokens.clear();
         
         while(end >= 0){
-        	// Only add non-empty tokens
-        	if(start != end){
-        		tokens.add(aString.substring(start, end));
-        		start = end+1;
-        	}
+        	
+        	// add the string, even when empty (start == end)
+        	tokens.add(aString.substring(start, end));
         	start = end+1;
         	
     		// If this condition is true, we likely ended with a token, so
@@ -222,9 +218,12 @@ public class ParsingUtils {
 
         }
         
-        // Add the final element (if non-empty)
+        // Add the final element (if start == string.length(), then we ended
+        // with a delimiter)
         if(start < aString.length()){
         	tokens.add(aString.substring(start));
+        }else{
+        	tokens.add("");
         }
     
         return tokens.size();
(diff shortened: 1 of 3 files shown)


src/java/htsjdk/tribble/util/ParsingUtils.java
```diff
@@ -182,6 +183,54 @@ public static int split(String aString, String[] tokens, char delim) {
...
+        
+        // Add the final element (if non-empty)
+        if(start < aString.length()){
+        	tokens.add(aString.substring(start));
+        }
+    
+        return tokens.size();
+    }
+    
+    /**
```
ronlevine: Is this utility needed?
Why not just do:
String str;
String[] tokens = str.split(",");
int numTokens = tokens.length;

src/java/htsjdk/tribble/util/ParsingUtils.java
```diff
@@ -182,6 +183,54 @@ public static int split(String aString, String[] tokens, char delim) {
...
+        int start = 0;
+        int end = aString.indexOf(delim);
+        
+        // Empty out our answer, but preserve the memory allocation
+        tokens.clear();
+        
+        while(end >= 0){
+        	// Only add non-empty tokens
+        	if(start != end){
+        		tokens.add(aString.substring(start, end));
```
ronlevine: For Vector, add appends to the end of the vector. So, for this case, the first add will create element 10001;

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -752,13 +753,13 @@ private final static boolean isVCFStream(final InputStream stream, final String
...
-    private final String[] INT_DECODE_ARRAY = new String[10000];
+    private final Vector<String> INT_DECODE_VECTOR = new Vector<String>(10000);
     private final int[] decodeInts(final String string) {
-        final int nValues = ParsingUtils.split(string, INT_DECODE_ARRAY, ',');
+        final int nValues = ParsingUtils.split(string, INT_DECODE_VECTOR, ',');
         final int[] values = new int[nValues];
         try {
             for ( int i = 0; i < nValues; i++ )
-                values[i] = Integer.valueOf(INT_DECODE_ARRAY[i]);
+                values[i] = Integer.valueOf(INT_DECODE_VECTOR.elementAt(i));
```
ronlevine: All of *values* will be incorrect until the initial size of INT_DECODE_VECTOR is set  *blank* (zero).

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -752,13 +753,13 @@ private final static boolean isVCFStream(final InputStream stream, final String
         return new LazyGenotypesContext.LazyData(genotypes, header.getSampleNamesInOrder(), header.getSampleNameToOffset());
     }
 
-    private final String[] INT_DECODE_ARRAY = new String[10000];
+    private final Vector<String> INT_DECODE_VECTOR = new Vector<String>(10000);
```
ronlevine: Remove 10000 so that the split() method does not append to the end.

src/java/htsjdk/tribble/util/ParsingUtils.java
ronlevine: Please unit test.

bpow:
It should probably be noted that `String.split` is being avoided because some benchmarking on JDK 6 shows very poor performance for this method.

The `String.split` in OpenJDK 6 always creates a regex, but versions 7 and beyond have a [fastpath](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/7u40-b43/java/lang/String.java#String.split%28java.lang.String%2Cint%29) when splitting is performed on a single character. Presumably the same technique is used in the Oracle JDK versions 7 and above.

Whenever support for JDK version 6 is dropped, it would be reasonable to clean up this workaround, and that cleanup would be facilitated with an appropriate comment somewhere.

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
bpow: Is the allocation of the String[]/Vector<String> really a bottleneck here, or would it be better to avoid having this field in the class?

A Vector<String> could be allocated within the decodeInts method, in which case the method could be made static.

Furthermore, why not use an ArrayList instead of a Vector in that case, and avoid the synchronization overhead?

johnwallace123: I imagine that with potentially millions, it's possible that the allocation could get large, especially if we start with a capacity of 0.  I attempted to make as little additional overhead as possible.

Good point on the ArrayList vs. Vector; most recent update incorporates this change.  Can you tell I come from C++-land :)?

johnwallace123:
I've made some modifications and added unit tests.  I wasn't sure about the corner cases (i.e., collapsing tokens / starting and ending with tokens), so I used some guidance from the unit tests already in place for split().  Please let me know if the behavior of split() should change, and I can fix accordingly.

geoffjentry:
@johnwallace123 and @ronlevine - we looked at this and #241 and unsurprisingly they're very similar. We're leaning to this one but if either of you (or anyone else) can make a strong argument in favor of #241 we're still able to be swayed.

ronlevine:
For https://github.com/samtools/htsjdk/pull/241, the main advantage is less code and it relies on previously used/tested routines. This disadvantages are that it requires the Apache commons library and it should be slower (requires an extra pass (O(n)) to size the array). That being said, I would go with https://github.com/samtools/htsjdk/pull/246. I am not sure how much of advantage pre-allocating a list of 10,000 strings. To save memory, something on the order of 100 or so might be more reasonable. I was also thinking that the List should be converted back to a data structure where element access is O(1) instead of O(n).

bpow:
ArrayList element access is O(1).

I agree that it doesn't seem like there needs to be a pre-allocated list of strings (or array of strings, whatever the collection type) around. Using mutable parameters in order to perform a return from a method (which is what `ParsingUtils.split` does with its second parameter) is a bit of a code smell, especially when the actual return of the method just ends up being the length of the List (which the List object knows).

Pre-allocation in java is frequently a premature optimization that doesn't necessarily help, and means that i) the method has side effects and ii) the method is not thread-safe when called on the same object twice from two different threads.

I would vote for a solution that would add a `public static final List<String> split(String parent, char delim)` to ParsingUtils that basically does what split does now but constructs an ArrayList rather than filling a pre-allocated array or List. This would be more idiomatic java and should still be performant.

I would be happy to add a third pull request to the mix if it would help.

geoffjentry:
@bpow @ronlevine @johnwallace123 It appears to already be happening but since we're pretty resource constrained at the moment could we ask you three to hash out a single solution? FWIW I personally like the direction this conversation is leading.

ronlevine:
@bpow I think it would be simpler to stay will this pull request because it's easier to follow the history. You can give your pseudo code implementation of **public static final List<String> split(String parent, char delim)** as code line note or comment.

bpow:
@ronlevine Sorry about the extra pull request (at #248), but I hadn't seen your suggestion to stay in this thread before putting it together. I could try to rebase my thoughts off the pull request here, but I think things would end up getting squashed into one commit at some point anyway.

ronlevine:
@bpow Let's stay with https://github.com/samtools/htsjdk/pull/248 (this conversation already has a link to it). It looks good at first glance, I will take a closer look soon.

johnwallace123:
@ronlevine @bpow, I like the look of #248 (couple minor comments added); it definitely seems more Java-esqe, and if quick benchmarking shows that preallocation is unnecessary, then it should probably be avoided.  I think it's important to keep the guess at initial capacity (~string length / 2) in the initialization of the ArrayList to avoid copying large arrays when the container grows.

Thanks to both of you for helping us all come to a better solution than I think any of us could get to by ourselves!

johnwallace123:
Superseded by #248

PR #249, opened 2015-05-13
Fix catastrophic bug in VCFHeader that caused newly-added header lines to be silently dropped

droazen:
There was a nasty bug in VCFHeader that caused newly-added header lines
to only be added to type-specific lookup tables and not to the master list
of header lines. This meant that new lines got silently dropped when writing
a modified VCFHeader back out via VCFWriter.

Fixed the issue and added comprehensive unit tests for VCFHeader.addMetaDataLine().

Also attempted to improve documentation and method naming a bit in an
effort to avoid future bugs like this. Eg., there was a method "loadVCFVersion()"
that actually REMOVED the vcf version lines!

### DIFF
--- a/src/java/htsjdk/variant/vcf/VCFHeader.java
+++ b/src/java/htsjdk/variant/vcf/VCFHeader.java
@@ -48,15 +48,11 @@ import java.util.TreeSet;
 
 
 /**
- * NOTE: This class allows duplicate entries in the metadata & stores header lines in
- * lots of places. The original author noted that this should be cleaned up at some point
- * in the future (jgentry - 5/2013)
+ * A class to represent a VCF header
  *
  * @author aaron
- *         <p/>
- *         Class VCFHeader
- *         <p/>
- *         A class representing the VCF header
+ * NOTE: This class stores header lines in lots of places. The original author noted that this should
+ * be cleaned up at some point in the future (jgentry - 5/2013)
  */
 public class VCFHeader implements Serializable {
     public static final long serialVersionUID = 1L;
@@ -110,14 +106,15 @@ public class VCFHeader implements Serializable {
     }
 
     /**
-     * create a VCF header, given a list of meta data and auxillary tags
+     * create a VCF header, given a list of meta data and auxiliary tags
      *
      * @param metaData     the meta data associated with this header
      */
     public VCFHeader(final Set<VCFHeaderLine> metaData) {
         mMetaData.addAll(metaData);
-        loadVCFVersion();
-        loadMetaDataMaps();
+        removeVCFVersionLines(mMetaData);
+        createLookupEntriesForAllHeaderLines();
+        checkForDeprecatedGenotypeLikelihoodsKey();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


droazen:
@jacarey This is a blocking issue for the GATK 3.4 release and needs priority attention. We'd be grateful if you could push the fixed version out to maven central after this gets merged.

droazen:
For @abaumann to review, please.

abaumann:
It looks like there is a lot of code involved in maintaining both mMetaData and each of the type specific meta data maps.  From what I can tell, mMetaData contains the union of each of those type specific maps.  Should we choose one or the other instead of maintaining both a master and type specific?  Either a master map of maps that includes type key, or just the type specific maps to avoid needing to maintain both?

abaumann:
Also, why do we strip off the VCF version lines and set it to always be VCF 4.1?

vdauwera:
I would support refactoring this whole class but in another ticket since we are looking for a quick fix to unblock the GATK release here.

abaumann:
Do we need a unit test to address the VCFWriter issue specifically: (the "This meant that new lines got silently dropped when writing a modified VCFHeader back out via VCFWriter."), or are the tests around checking the master always contains the added lines sufficient?

abaumann:
In response to the need to address quickly, all looks good if none of the other points I raised lead to any needed changes.

droazen:
@abaumann I agree that the existing design of the `VCFHeader` class is extremely poor, and basically an invitation for bugs of this sort to be introduced, but fixing the structural issues would be a complete re-write, more or less, which is always a bad idea just before a release. I prefer for now to confine ourselves to a targeted fix for the bug in question, plus improved tests, docs, and method names to reduce the danger of keeping the existing version of this class around.

droazen:
@abaumann The version line(s) are stripped at construction, and then replaced on-demand with a VCF 4.1 line by certain of the accessor methods. This is more terrible, terrible design -- `VCFWriter` should be handling versioning of the output header, not this class. But I am just preserving the existing behavior for now -- all I did was rename `loadVCFVersion()` to `removeVCFVersionLines()` to reflect what was actually going on :)

jacarey:
Just tag me when this is merged and ready for release... Unfortunately there is no easy way for use to release just the htsjdk so we'll need to do a full picard public release to both github and maven central.

droazen:
@abaumann A `VCFWriter` test to supplement the new `VCFHeader` tests is a good idea -- I will add that now.

droazen:
@abaumann I've added the new `VariantContextWriter` test you requested, and created tickets to eventually address the design problems in the `VCFHeader` class (in a future branch). Let me know if we're good to merge!

abaumann:
Looks good to me, and I agree with the changes to rename and better doc as an appropriate quick solution (glad to see loadVCFVersion renamed given what it was doing as well!), will merge once Travis finally decides to show up for work

abaumann:
@jacarey Merged

vdauwera:
Thanks all for getting this fixed quickly!

jacarey:
Ok starting the release. I'll do the maven releases first to get those done

jacarey:
@droazen HTSJDK version 1.132 pushed to maven central. I should be updated and indexed shortly. 

I'll work on the github and picard public release now, but I think this you guys should be all set with htsjdk.

droazen:
@jacarey Thanks!

PR #252, opened 2015-05-18
Performance enhancement

d-cameron:
Performance optimisation for VCF indexing, bam/sam async writing, and retrieving small subsequences from a reference.

### DIFF
--- a/src/java/htsjdk/samtools/reference/IndexedFastaSequenceFile.java
+++ b/src/java/htsjdk/samtools/reference/IndexedFastaSequenceFile.java
@@ -191,8 +191,8 @@ public class IndexedFastaSequenceFile extends AbstractFastaSequenceFile implemen
 
         long startOffset = ((start-1)/basesPerLine)*bytesPerLine + (start-1)%basesPerLine;
 
-        // Allocate a 128K buffer for reading in sequence data.
-        ByteBuffer channelBuffer = ByteBuffer.allocate(Defaults.NON_ZERO_BUFFER_SIZE);
+        // Allocate a buffer for reading in sequence data.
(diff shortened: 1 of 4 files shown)


geoffjentry:
Hi @d-cameron - can you explain a bit how you decided on `(length % basesPerLine + 2) * bytesPerLine)` to be the smaller buffer size?

d-cameron:
The length of the sequence in the reference file depends on where the newlines occur relative to the start/end position of the sequence. The buffer size chosen ensures that the buffer has enough room to hold all lines that contain at least one base of the subsequence. Technically, since the file is read starting at the offset of the first base, not at the start of the line of the first base, (length % basesPerLine + 1) * bytesPerLine) would be still be sufficient to read the entire subsequence from a single buffer.

geoffjentry:
@d-cameron Gotcha. Sounds good. Can you do a squash/rebase to a single commit and then we can merge this in? Thanks!

d-cameron:
Each commit stands as an individual change and from a version history perspective, 3 different commits for 3 different changes is clearer. Some of my previous pull requests have used the same commit-per-change format and were accepted and this one can be merge in its current state. Should I switch to a single commit format (or make multiple single-commit pull requests) in the future? Is a single commit now the preferred pull request format for htsjdk?

tfenne:
Just my 2c, but this makes total sense to me @d-cameron.  I think folks mostly use the branch-per-feature model in htsjdk, so generally one PR = one feature = one squashed commit.  But if it's three separate features on one branch I don't see an issue.

geoffjentry:
Agree that they're three atomic things but disagree that they're individually each of a scope worthy of a separate commit. OTOH I don't care enough to worry too much here since at least it's reasonable and not 15 commits saying "foo" and "checkpoint" :) Thanks for submitting this!

PR #260, opened 2015-05-28
Rhl allele rep span dels

ronlevine:
Partial solution for https://github.com/broadinstitute/gsa-unstable/issues/998

Summary of code changes:
**Allele.java**
* ctor - created the SPAN_DEL allele
* create() - added case statement for * (spanning deletion)
* added final before method arguments

**AbstractVCFCodec.java**
* changed length() == 0 to isEmpthy()

**VCFConstants.java**
* added SPANNING_DELETION_ALLELE

**AlleleUnitTest.java**
* added testCreatingSpanningDeletionAlleles, testBadNoCallAllele and testBadSpanningDeletionAllele
* In testCreatingSpanningDeletionAlleles, I am not sure if the the spanning deletions should be processed like a no-call,  having no bases (see *Allele.Allele(final byte[] bases, final boolean isRef)*)

### DIFF
--- a/src/java/htsjdk/variant/variantcontext/Allele.java
+++ b/src/java/htsjdk/variant/variantcontext/Allele.java
@@ -129,6 +129,9 @@ public class Allele implements Comparable<Allele>, Serializable {
     /** A generic static NO_CALL allele for use */
     public final static String NO_CALL_STRING = ".";
 
+    /** A generic static SPAN_DEL allele for use */
+    public final static String SPAN_DEL_STRING = "*";
+
     // no public way to create an allele
     protected Allele(final byte[] bases, final boolean isRef) {
         // null alleles are no longer allowed
@@ -155,7 +158,7 @@ public class Allele implements Comparable<Allele>, Serializable {
         this.isRef = isRef;
         this.bases = bases;
 
-        if ( ! acceptableAlleleBases(bases) )
+        if ( ! acceptableAlleleBases(bases, isRef) )
             throw new IllegalArgumentException("Unexpected base in allele bases \'" + new String(bases)+"\'");
     }
 
@@ -190,7 +193,7 @@ public class Allele implements Comparable<Allele>, Serializable {
     private final static Allele ALT_T = new Allele("T", false);
     private final static Allele REF_N = new Allele("N", true);
     private final static Allele ALT_N = new Allele("N", false);
-    private final static Allele SPAN_DEL = new Allele(VCFConstants.SPANNING_DELETION_ALLELE, false);
+    public final static Allele SPAN_DEL = new Allele(SPAN_DEL_STRING, false);
     public final static Allele NO_CALL = new Allele(NO_CALL_STRING, false);
 
     // ---------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


src/java/htsjdk/variant/variantcontext/Allele.java
```diff
@@ -307,8 +323,8 @@ public static boolean acceptableAlleleBases(byte[] bases, boolean allowNsAsAccep
             switch (base) {
                 case 'A': case 'C': case 'G': case 'T':  case 'a': case 'c': case 'g': case 't': 
                     break;
-                case 'N' : case 'n' :
-                    if (allowNsAsAcceptable)
+                case 'N' : case 'n' : case '*':
+                    if (isReferenceAllele)
                         break;
```
eitanbanks: This needs to be changed.
1) It turns out that the spec was changed and 'N's are allowed in both the REF and ALT fields, so they should be treated just like other bases above.
2) '*'s are acceptable only in the ALT field (so NOT if isReferenceAllele is true).  But they are also only allowed as the ONLY character (so this check needs to occur outside of the for loop).  E.g.
if ( wouldBeStarAllele(bases) )
             return true;
and wouldBeStarAllele() would check that it's a single base == to '*'.

eitanbanks: Markdown killed my stars.  :(

src/java/htsjdk/variant/vcf/VCFConstants.java
```diff
@@ -93,6 +94,9 @@
     public static final char DELETION_ALLELE_v3 = 'D';
     public static final char INSERTION_ALLELE_v3 = 'I';
 
+    // spanning deletion allele
+    public static final String SPANNING_DELETION_ALLELE = "*";
+
```
eitanbanks: I think this is better defined as a char instead of a String (since you'll need it for the wouldBeStarAllele() test).

src/java/htsjdk/variant/variantcontext/Allele.java
```diff
@@ -189,6 +190,7 @@ protected Allele(final Allele allele, final boolean ignoreRefState) {
     private final static Allele ALT_T = new Allele("T", false);
     private final static Allele REF_N = new Allele("N", true);
     private final static Allele ALT_N = new Allele("N", false);
+    private final static Allele SPAN_DEL = new Allele(VCFConstants.SPANNING_DELETION_ALLELE, false);
     public final static Allele NO_CALL = new Allele(NO_CALL_STRING, false);
```
eitanbanks: I think this should be public, not private.  It's okay for others to use.  And you will want to do so in CombineGVCFs and GenotypeGVCFs.

src/tests/java/htsjdk/variant/variantcontext/AlleleUnitTest.java
```diff
@@ -59,7 +59,9 @@ public void before() {
         ATIns = Allele.create("AT");
         ATCIns = Allele.create("ATC");
 
-        NoCall = Allele.create(".");
+        NoCall = Allele.create(Allele.NO_CALL_STRING);
+
+        SpandDel = Allele.create(htsjdk.variant.vcf.VCFConstants.SPANNING_DELETION_ALLELE);
     }
```
eitanbanks: Use Allele.SPAN_DEL once it's public.

src/java/htsjdk/variant/variantcontext/Allele.java
vdauwera: Wrap them in backticks to rescue them

ronlevine:
@eitanbanks Post review comments incorporated. Ready for review

eitanbanks:
Looks awesome!

PR #271, opened 2015-06-22
Added IOUtil.setDefaultCompressionLevel() to set the compression level

nh13:
that GZIP file are opened with. Support for this already existed, but
the level was hardcoded to use Defaults.COMPRESSION_LEVEL (=5).

### DIFF
--- a/src/java/htsjdk/samtools/util/BlockCompressedOutputStream.java
+++ b/src/java/htsjdk/samtools/util/BlockCompressedOutputStream.java
@@ -98,14 +98,14 @@ public class BlockCompressedOutputStream
     private final byte[] singleByteArray = new byte[1];
 
     /**
-     * Uses default compression level, which is 5 unless changed by setDefaultCompressionLevel
+     * Uses default compression level, which is 5 unless changed by setCompressionLevel
      */
     public BlockCompressedOutputStream(final String filename) {
         this(filename, defaultCompressionLevel);
     }
 
     /**
-     * Uses default compression level, which is 5 unless changed by setDefaultCompressionLevel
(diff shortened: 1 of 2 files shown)


nh13:
@nnordborg

tfenne:
Instead of doing this in IOUtil - wouldn't it be better to make Defaults.COMPRESSION_LEVEL settable?  Or is the goal to have divergent defaults for SAM/BAM and other GZIP files?

nh13:
Making ``` Defaults.COMPRESSION_LEVEL``` settable would be tough since it make sense to be ```static final``` in ```Defaults.java```.   Nonetheless, it could be better named to be specific to gzip files, which is the intention in the modification to ```IOUtil.java```.  This is similar, as @nnordborg said, to the ```setDefaultCompressionLevel``` method in ```BlockCompressedOutputStream.java```.  I wouldn't be opposed to it being a method in the hacky class ```CustomGzipOutputStream```.  What do you think @tfenne ?

tfenne:
This still feels to me like an odd way to tackle this, but I won't object.  Personally I'd prefer to simply remove "final" from the declaration of COMPRESSION_LEVEL in Defaults (it doesn't have to be final).  Short of that, this works.

I might also suggest changing the name of the property and method to just "compressionLevel" not "defaultCompressionLevel" as none of the methods in the class take in a compression level parameter, so it's more than just a default.

yfarjoun:
@tfenne I updated the names of the variable, setters and getters...good to go now?

tfenne:
:+1:

yfarjoun:
rebased and squashed.

PR #273, opened 2015-06-24
API VariantContextUtils.initializeMatchExps fixed to use interfaces

akiezun:
This is a fix for the API. The method signature should use interfaces not concrete classes to allow more flexibility in client code. This is a backwards compatible change - callers are unaffected and it's a static method so no one can override it.

### DIFF
--- a/src/java/htsjdk/variant/variantcontext/VariantContextUtils.java
+++ b/src/java/htsjdk/variant/variantcontext/VariantContextUtils.java
@@ -241,7 +241,17 @@ public class VariantContextUtils {
         return VariantContextUtils.initializeMatchExps(map);
     }
 
-    public static List<JexlVCMatchExp> initializeMatchExps(ArrayList<String> names, ArrayList<String> exps) {
+    /**
+     * Method for creating JexlVCMatchExp from input walker arguments names and exps.  These two lists contain
+     * the name associated with each JEXL expression. initializeMatchExps will parse each expression and return
+     * a list of JexlVCMatchExp, in order, that correspond to the names and exps.  These are suitable input to
+     * match() below.
+     *
+     * @param names names
+     * @param exps  expressions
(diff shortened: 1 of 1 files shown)


tfenne:
Why not go the whole way and make the two parameters Collection<String> instead of List<String>?

akiezun:
done

lbergelson:
Is that Ok?  Don't these need to be ordered?

tfenne:
@lbergelson the two collections need to have a matched iteration order, but that's achievable with other collections too, e.g. LinkedHashSet.

lbergelson:
If it requires ordering I think accepting collection is a scary idea.  It's unfortunate that java doesn't have a way to specify an ordered, but as a user I would much rather have to convert my LinkedHashSet to a List than expect to be able to pass a HashSet in and then have bizarre bugs later.

akiezun:
i'm happy to do whatever as long as it's at least `List`. Someone in charge please let me know. @nh13 maybe, are you the decider?

yfarjoun:
why is this failing?

yfarjoun:
FWIW I agree with @lbergelson that if it's a collection we shouldn't be assuming anything about the sorting since that's not in the collection contract. I'd go with List.

akiezun:
If I change it back to List, can someone promise to review it? This 1 liner PR with no change in semantics has been in review for 26 days.

yfarjoun:
I'll +1 it if it passes travis.

akiezun:
ok back to List and i added javadoc while i was there. I think the travis failures were some travis problem not me.

nh13:
There are some weird errors unrelated to this commit (ftp access).  :+1:

yfarjoun:
you do the merging @akiezun

akiezun:
thanks guys

Issue #274, opened 2015-07-01
Genotype.getAnyAttribute incorrectly handles converting int[] to List<Integer>


dfsurdyk:
Both getAD() and getPL() return an int[], and doing ```Arrays.asList(int[])``` will return a ```List<int[]>``` instead of ```List<Integer>```.

https://github.com/samtools/htsjdk/blob/master/src/java/htsjdk/variant/variantcontext/Genotype.java#L544

nh13:
```Arrays.asList(int[])``` **will** return `` List<Integer>```, so I am closing this issue.

nh13:
Wow, I am **so** wrong.

dfsurdyk:
@lbergelson - Any progress on merging this with master?  Am I missing something in the PR?

yfarjoun:
My 2ct: I think we shoudl get rid of GetAnyAttribute and implement (as for the info field) getAttributeAsArray{String,Int,Float}

tfenne:
Fixed by #275 (commit b2638e03b). Genotype.getAnyAttribute() now correctly builds a List<Integer> from the AD/PL int[]. Closing as resolved.

PR #275, opened 2015-07-02
Fix for incorrect return value in Genotype.getAnyAttribute()

dfsurdyk:
* The Genotype.getAnyAttribute() incorrectly handles converting int[] to
  List<Integer>.  Both getAD() and getPL() return an int[], and doing
  Arrays.asList(int[]) will return a List<int[]> instead of List<Integer>.

See: https://github.com/samtools/htsjdk/issues/274

### DIFF
--- a/src/java/htsjdk/variant/variantcontext/Genotype.java
+++ b/src/java/htsjdk/variant/variantcontext/Genotype.java
@@ -541,9 +541,13 @@ public abstract class Genotype implements Comparable<Genotype>, Serializable {
         } else if (key.equals(VCFConstants.GENOTYPE_QUALITY_KEY)) {
             return getGQ();
         } else if (key.equals(VCFConstants.GENOTYPE_ALLELE_DEPTHS)) {
-            return Arrays.asList(getAD());
+            List<Integer> intList = new ArrayList<Integer>();
+            for(int i : getAD()) intList.add(i);
+            return intList;
         } else if (key.equals(VCFConstants.GENOTYPE_PL_KEY)) {
-            return Arrays.asList(getPL());
+            List<Integer> intList = new ArrayList<Integer>();
+            for(int i : getPL()) intList.add(i);
+            return intList;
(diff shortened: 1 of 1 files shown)


nh13:
What version of java are you using?

nh13:
And we're back to being open for review...

src/java/htsjdk/variant/variantcontext/Genotype.java
```diff
@@ -541,9 +541,13 @@ public Object getAnyAttribute(final String key) {
         } else if (key.equals(VCFConstants.GENOTYPE_QUALITY_KEY)) {
             return getGQ();
         } else if (key.equals(VCFConstants.GENOTYPE_ALLELE_DEPTHS)) {
-            return Arrays.asList(getAD());
+            List<Integer> intList = new ArrayList<Integer>();
```
jacarey: Initialize list to size of the array.

dfsurdyk:
```
java version "1.7.0_67"
Java(TM) SE Runtime Environment (build 1.7.0_67-b01)
Java HotSpot(TM) 64-Bit Server VM (build 24.65-b04, mixed mode)
```


The following code:
```java
public static void main(String[] args) {
        int[] foo = new int[] { 1, 2, 3 };
        Object bar = Arrays.asList(foo);
        System.out.println(bar);

        List<Integer> baz = (List<Integer>) bar;
        for (int i : baz)
            System.out.println("i = " + i);
}
```

outputs:
```log
[[I@10be6858]
Exception in thread "main" java.lang.ClassCastException: [I cannot be cast to java.lang.Integer
	at Test.main(Test.java:15)
```

nh13:
Agreed that this is not intended behavior, and thanks for the fix.  One minor optimization from @jacarey.

dfsurdyk:
Fixed the array sizing.

nh13:
See commit: b2638e03bf07ea66dd73064f7e04189f0b6338bd
I accidentally pushed to master.

PR #279, opened 2015-07-16
adding .intervals as an acceptable extension for an intervalList. 

yfarjoun:
Given that IntervalListTools emits scattered.intervals files, this change will allow us to read the outputted files


gbggrant:
Excellent!  :+1:

nh13:
No

PR #280, opened 2015-07-16
Fix 2 bugs in MetricsBase

lbergelson:
I split this into 2 commits, the first should be uncontroversial, the second I want to make sure no one objects too.

Commit 1:
   Fix a bug in the implementation of `hashCode()` in `MetricsBase` it was crashing if any values were `null`.

Commit 2: 
   Fix what I consider a bug in the `equals` method of `MetricBase`
   Currently, the equality is exact equality with respect to floating point numbers.  However, when written to disk a MetricsBase has limited floating point precision, so writing something to disk and reading it back in will sometimes not equal the original file. I've altered the `equality` and `hashCode` methods to compare by using the same representation of floating point numbers as is written to disk.  This is technically backwards compatibility breaking, but I think the current behavior is erroneous.  The downside is that the equality code is complicated now.  
  The simpler and much more guaranteed way to check this would be to actually format the whole thing as an string and then compare string equality.  This would guarantee that equals and hashcode will always be in synch with the on disk representation, but will likely come at a performance cost. 

I've also deleted the `equals` specialization for `MetricsBase`s rather than having 2 complex implementations of equals.  This introduces another backwards incompatible change, since this method made no check for actual class which would allow comparing any `MetricBase` subclass to any other.  However, this equals method was broken.  It was an asymmetric equals method.  It checked the fields in one class against the same fields in the other class, but did not make any checks that those fields were all the fields in the second class.  I.e. 

```
	public class A extends MetricBase {
		public int a = 1;
	}
	public class B extends A{
		public int b = 1;
	}

	@Test
	public void testAB(){
		A a = new A();
		B b = new B();
		Assert.assertTrue(a.equals(b));  //true
		Assert.assertTrue(b.equals(a)); // illegal argument exception
	}
```

After my changes both ways will evaluate to false.  

@yfarjoun @nh13 @tfenne Let me know what you think.

### DIFF
--- a/src/java/htsjdk/samtools/metrics/MetricBase.java
+++ b/src/java/htsjdk/samtools/metrics/MetricBase.java
@@ -26,7 +26,6 @@ package htsjdk.samtools.metrics;
 
 import htsjdk.samtools.SAMException;
 import htsjdk.samtools.util.FormatUtil;
-import htsjdk.samtools.util.Objects;
 
 import java.lang.reflect.Field;
 
@@ -38,7 +37,8 @@ import java.lang.reflect.Field;
 public class MetricBase {
     /**
      * An equals method that checks equality by asserting that the classes are of the exact
-     * same type and that all public fields are equal.
+     * same type and that all public fields are equivalent.  Equivalence is checked by formatting
+     * the field as it would be written to disk and then checking String equality.
      *
      * @param o an instance to compare to
      * @return true if they are equal, false otherwise
@@ -46,7 +46,8 @@ public class MetricBase {
     public boolean equals(final Object o) {
         if (o == null) return false;
         if (o.getClass() != getClass()) return false;
-
+        
+        final FormatUtil formatter = new FormatUtil();
         // Loop through all the fields and check that they are either
         // null in both objects or equal in both objects
         for (final Field f : getClass().getFields()) {
@@ -58,12 +59,13 @@ public class MetricBase {
                     if (rhs == null) {
                         // keep going
                     }
-                    else if (rhs != null) {
+                    else {
                         return false;
                     }
                 }
                 else {
-                    if (lhs.equals(rhs)) {
+                    if (formatter.format(lhs).equals(formatter.format(rhs))) //compare based on the serialized representation
+                    {
                         // keep going
                     }
                     else {
@@ -80,11 +82,19 @@ public class MetricBase {
         return true;
     }
 
+    /**
+     * Computes a hashcode by formatting each field into its on disk representation 
+     * and summing the hashcodes of all the fields.
+     */
     public int hashCode() {
         int result = 0;
+        FormatUtil formatter = new FormatUtil();
         for (final Field f : getClass().getFields()) {
             try {
-                result = 31 * result + f.get(this).hashCode();
+                Object value = f.get(this);
+                value = formatter.format(value); //format the value the way it will be written to disk
+                final int fieldHash = value != null ? value.hashCode() : 0;
+                result = 31 * result + fieldHash;
             } catch (IllegalAccessException e) {
                 throw new SAMException("Could not read field " + f.getName() + " from a " + getClass().getSimpleName());
             }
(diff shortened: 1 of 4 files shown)


src/java/htsjdk/samtools/metrics/MetricBase.java
```diff
@@ -92,6 +108,18 @@ public int hashCode() {
...
+    static Object formatIfFloatingPoint(final FormatUtil formatter, final Object value) {
+        if(isFloatingPointType(value)){
+            return formatter.format(value);
+        } else {
+            return value;
+        }
+    }
+
+    static boolean isFloatingPointType(Object value){
+        return (value instanceof Double || value instanceof Float );
```
nh13: extra space

nh13:
First commit: :+1: 
Second commit: I do not think comparing metrics is currently underperformant, so I would prefer a simpler and more intuitive comparison method (compare strings).  I agree that this should fix the asymmetric nature of the ```equals``` method.

nh13:
@lbergelson what else needs to be done to complete this fix?

src/java/htsjdk/samtools/metrics/MetricBase.java
```diff
@@ -84,7 +85,10 @@ public int hashCode() {
         int result = 0;
         for (final Field f : getClass().getFields()) {
             try {
-                result = 31 * result + f.get(this).hashCode();
+                Object value = f.get(this);
+                value = new FormatUtil().format(value); //format the value the way it will be written to disk
```
tfenne: Why do this?  It's seems totally unnecessary to me and makes the hashCode() method much more inefficient.

lbergelson: If equality is based on the formatted value, hashCode will be inconsistent with equals if it is not also based on the formatted value.  I.e. `hashCode(1.0000000001) ~= hashCode(1) even though 1.000000001 == 1.

I believe that simplicity should take preference over performance here since these are not very performance critical classes.  The earlier version had explicit checks to only perform the formatting if it was decimal value that could be formatted, but that seemed confusing and brittle.

lbergelson: @tfenne Does that make sense or do you disagree?

lbergelson:
@nh13 Ok if I squash and merge this?  I haven't gotten a response from @tfenne yet.

nh13:
Lets give it a few more days please.

lbergelson:
Ok.  No problem.

src/java/htsjdk/samtools/metrics/MetricBase.java
tfenne: @lbergelson I see what you're saying.  And I see the logic of your argument.  Something still bothers me about using String formatted values in equals() and hashCode().

It's definitely simpler than, say, rounding floating point values for comparison, or doing an "approximate equality" check on floating point numbers.  And you're not wrong that the equals() and hashCode() methods are probably not performance sensitive on these classes.

Can I suggest a couple of minor changes before merging:
1) On line 89 extract the call to new FormatUtil() and move it outside the loop
2) Add/correct the documentation on equals() and hashCode() to explain in context why they are comparing String formatted versions of things.

lbergelson: I agree, it doesn't seem ideal, but the alternative possibly more performant way was was pretty gross and brittle. 


Will do.  Any objection to either 
+ creating a default formatter singleton 

or

+  rewrite FormatUtil to just have static methods, possibly breaking any client code the extends FormatUtils.  (which would cleaner but maybe unpalatable)

tfenne: I wouldn't do either of those things.  The various java.text.Format classes that FormatUtil relies on underneath do not provide any guarantees regarding thread-safety, and I know for a fact that at least the DateFormat classes are non-thread-safe, so wouldn't be surprised to find the same thing with Decimal/NumberFormat.

lbergelson: That's horrifying... why does `DateFormat` have state?

We should add a comment to `FormatUtil` explaining that then.  It looks like it's just needlessly allocating objects.

lbergelson:
@tfenne Are you alright with the state of this now?  I also think it's weird to have the String comparison, but I think it's the best choice out of several awkward ones.

lbergelson:
Thanks for the review btw.  It's always super helpful to to have input from the person who wrote something.  I could not for the life of me figure out why the Formatter needed to be instantiated to be used.

tfenne:
:+1:

PR #281, opened 2015-07-17
Refactor ProgressLogger for extensibility

lbergelson:
Pulled out an AbstractProgressLoger and refactored project logger to implement that.
This allows other implementations of AbstractProgressLogger that easily substitute a different logging system other than the Picard built in, without having to reimplement the existing logic.


nh13:
re-running travis before merging

PR #282, opened 2015-07-20
moving back to travis non-containerized build

### LINKED ISSUE #276: Builds are failing a lot because of failed FTP tests
All of the tests that make FTP connections seem to be failing at a very high rate since I switched us to travis container builds.  I'm talking with travis to figure out what is going on, but we might want to switch back to non-container build until it's resolved.

lbergelson:
We're having constant failures since I moved us to the travis sudoless build.  Until travis fixes whatever is going on on their end I think we should revert it.  Sorry for the hassle.  Should fix #276 but we go back to having slow dispatch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: java
-sudo: false
+sudo: true
 jdk:
   - oraclejdk7
   - oraclejdk8


nh13:
:+1:

PR #283, opened 2015-07-22
enum describing SAM FLAG (for gui/ menu / etc... )

lindenb:
This pull request contains a new enum htsjdk.samtools.SAMFlag describing the available SAM flags. It can be used in a GUI, HTML form, menu... to display the available flags...

SAMFlag contains a static method returning the java.util.Set for a given integer

```java
public static Set<SAMFlag> getFlags(int flag)
```

I've added a small shortcut in src/java/htsjdk/samtools/SAMRecord.java
```java
   public final Set<SAMFlag> getSAMFlags() {
        return SAMFlag.getFlags( this.getFlags() );
    }
```

I've added a test under src/tests/java/htsjdk/samtools/SAMFlagTest.java

### DIFF
--- a/src/java/htsjdk/samtools/SAMFlag.java
+++ b/src/java/htsjdk/samtools/SAMFlag.java
@@ -1,0 +1,110 @@
+/*
+ * The MIT License
+ *
+ * Author: Pierre Lindenbaum PhD @yokofakun
+ *  Institut du Thorax - Nantes - France
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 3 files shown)


nh13:
It seems like this duplicates some functionality in ```SAMRecord.java```.  Perhaps you could find a way to reconcile the two?

src/java/htsjdk/samtools/SAMFlag.java
```diff
@@ -0,0 +1,92 @@
...
+package htsjdk.samtools;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * SAM flags as enum, to be used in GUI, menu, etc...
+ */
+public enum SAMFlag {
+    READ_PAIRED(0x1),
```
lbergelson: Should these be renamed to reflect the more recent nomenclature? 

I.E.  `READ_PAIRED` ->  `MULTIPLE_READS_IN_TEMPLATE`

lindenb:
@nh13  I've removed the `final int` declarations in SAmRecord and I now use the values from SAMFLag.
@lbergelson I didn't know the new spec. I renamed the whole enum and added a human description for each item.

src/java/htsjdk/samtools/SAMFlag.java
droazen: Not sure that htsjdk is ready for the more modern terminology, since the newer names are inconsistent with the rest of the codebase. Couldn't this lead to confusion/error?

src/java/htsjdk/samtools/SAMFlag.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /** @return a human description for this SAMFlag */
+    public String getDescription() {
+        return this.description;
+    }
+
+    /** @return the SAMFlag for the value 'flg' or null if it was not found */
+    public static SAMFlag valueOf(int flg) {
+        for (SAMFlag f : values()) {
```
jacarey: flg -> flag

jacarey:
:+1:

nh13:
@lindenb could you change the names of each enum value back to the previous names (ex. ```MULTIPLE_READS_IN_TEMPLATE``` -> ```READ_PAIRED_FLAG```.  We are not ready to modify all f the various methods in ```SAMRecord``` to the new terminology in the spec, given all the other code that relies on them.

lindenb:
@nh13 done.

PR #284, opened 2015-07-23
seq dict

lindenb:
added a few functions to SAMSequenceDictionary

* md5() to get a MD5 checksum of the dictionary ( to quickly find a dictionary in a `Map<String,SAMSequenceDictionary> ` )
* toString
* countBases return the number of bases seen so far if we're at position `chrom:pos`
* fraction : same as prevous, but return the ratio with the reference length

### DIFF
--- a/src/java/htsjdk/samtools/SAMSequenceDictionary.java
+++ b/src/java/htsjdk/samtools/SAMSequenceDictionary.java
@@ -24,6 +24,8 @@
 package htsjdk.samtools;
 
 import java.io.Serializable;
+import java.math.BigInteger;
+import java.security.MessageDigest;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -31,12 +33,21 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+import javax.xml.bind.annotation.XmlTransient;
+
 /**
  * Collection of SAMSequenceRecords.
  */
+@XmlRootElement(name="References")
 public class SAMSequenceDictionary implements Serializable {
     public static final long serialVersionUID = 1L;
 
+    /* xml Serialization , for `m_sequence` we use the field instead of the
+    getter because the later wraps the list into an unmodifiable List 
+    see http://tech.joshuacummings.com/2010/10/problems-with-defensive-collection.html */
+    @XmlElement(name="Reference")
     private List<SAMSequenceRecord> mSequences = new ArrayList<SAMSequenceRecord>();
     private final Map<String, SAMSequenceRecord> mSequenceMap = new HashMap<String, SAMSequenceRecord>();
 
@@ -48,6 +59,7 @@ public class SAMSequenceDictionary implements Serializable {
         setSequences(list);
     }
 
+    @XmlTransient //we use the field instead of getter/setter
     public List<SAMSequenceRecord> getSequences() {
         return Collections.unmodifiableList(mSequences);
     }
@@ -58,6 +70,7 @@ public class SAMSequenceDictionary implements Serializable {
 
     /**
      * Replaces the existing list of SAMSequenceRecords with the given list.
+     * Reset the aliases
      *
      * @param list This value is used directly, rather than being copied.
      */
@@ -105,6 +118,9 @@ public class SAMSequenceDictionary implements Serializable {
         return record.getSequenceIndex();
     }
 
+    /**
+     * @return number of SAMSequenceRecord(s) in this dictionary
+     */
     public int size() {
         return mSequences.size();
     }
@@ -119,7 +135,10 @@ public class SAMSequenceDictionary implements Serializable {
         }
         return len;
     }
-
+    
+    /**
+     * @return true is the dictionary is empty
+     */
     public boolean isEmpty() {
         return mSequences.isEmpty();
     }
@@ -129,6 +148,7 @@ public class SAMSequenceDictionary implements Serializable {
      * Non-comprehensive {@link #equals(Object)}-assertion: instead of calling {@link SAMSequenceRecord#equals(Object)} on constituent
      * {@link SAMSequenceRecord}s in this dictionary against its pair in the target dictionary, in order,  call 
      * {@link SAMSequenceRecord#isSameSequence(SAMSequenceRecord)}.
+     * Aliases are ignored.
      *
      * @throws AssertionError When the dictionaries are not the same, with some human-readable information as to why
      */
@@ -150,7 +170,8 @@ public class SAMSequenceDictionary implements Serializable {
         if (thatSequences.hasNext())
             throw new AssertionError(String.format(DICT_MISMATCH_TEMPLATE, thatSequences.next() + " is present in only one dictionary"));
     }
-    
+
+    /** returns true if the two dictionaries are the same, aliases are NOT considered */
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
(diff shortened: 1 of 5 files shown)


src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +165,72 @@ public boolean equals(Object o) {
         return true;
     }
 
+    /** return a MD5 sum for ths dictionary, the checksum is not re-computed each time this
+     * method is called. <pre>md5( (seq1.name+seq1.length) + ' '+(seq2.name+seq2.length) + ' '+...)</pre>
+     * @return a MD5 checksum for this dictionary or the empty string if it is empty
+     */
+    public String md5() {
+        if(isEmpty()) return "";
```
nh13: @lindenb the coding style and elsewhere needs to be consistent with htsjdk.  It makes it easier for folks to maintain and read the code.  In this case, ```if(``` -> ```if (```.  Thanks for updating the code in this pull requests and future ones too, as I really appreciate the contributions you are making.  I would recommend using intelliJ's or your favorite IDE's code re-format function.

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +165,72 @@ public boolean equals(Object o) {
...
+        try {
+            MessageDigest md5 = MessageDigest.getInstance("MD5");
```
nh13: IntelliJ would show a warning here, since this variable could be final.

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +165,72 @@ public boolean equals(Object o) {
...
+            md5.reset();
+            for(SAMSequenceRecord ssr : mSequences)  {
```
nh13: final

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +165,72 @@ public boolean equals(Object o) {
...
+                if(ssr.getSequenceIndex()>0)  md5.update((byte)' ');
```
nh13: missing space and extra space.

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +165,72 @@ public boolean equals(Object o) {
...
+                md5.update(ssr.getSequenceName().getBytes());
+                md5.update(String.valueOf(ssr.getSequenceLength()).getBytes());
+            }
+            String  hash = new BigInteger(1, md5.digest()).toString(16);
```
nh13: ```String  hash``` -> ```String hash```.  I'll stop now but you get the point

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +165,72 @@ public boolean equals(Object o) {
...
+     */
+    public String md5() {
+        if(isEmpty()) return "";
+        try {
+            MessageDigest md5 = MessageDigest.getInstance("MD5");
+            md5.reset();
+            for(SAMSequenceRecord ssr : mSequences)  {
+                if(ssr.getSequenceIndex()>0)  md5.update((byte)' ');
+                md5.update(ssr.getSequenceName().getBytes());
+                md5.update(String.valueOf(ssr.getSequenceLength()).getBytes());
```
nh13: Perhaps we also use the ```MD5_TAG``` as well if it is available?  This would really help in the case that the sequence names and lengths are the same, but the bases differ.  It happens, seriously.

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +165,72 @@ public boolean equals(Object o) {
         return true;
     }
 
+    /** return a MD5 sum for ths dictionary, the checksum is not re-computed each time this
```
nh13: I don't see how it is not re-computed each time.  In the case that it doesn't re-compute, it perhaps should when new sequences are set.

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +165,72 @@ public boolean equals(Object o) {
...
+    }
+    
     @Override
     public int hashCode() {
         return mSequences.hashCode();
     }
+    
+    @Override
+    public String toString() {
+        return mSequences.toString();
```
nh13: So this would return the class name + the hashcode of the array?  Perhaps it is more useful to return the md5?  What's the use case for this?

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +165,72 @@ public boolean equals(Object o) {
...
+    /** 
+     * @param sequenceIndex contig id
+     * @param position position in the contig
+     * @returns the number of bases seen from the first sequence at position 0 to the sequenceIndex/position or -1 on error*/
+    public long countBases(final int sequenceIndex,final int position)  {
+        if (sequenceIndex < 0 || position <0 || sequenceIndex >= mSequences.size() ) {
+            return -1L;
+        }
+        long countBases = 0L; /* number of bases seen so far */
+        for(int i = 0 ; i < sequenceIndex ; i++) countBases += getSequence(i).getSequenceLength();
```
nh13: probably better to use ```for (final SAMSequenceRecord sequence : mSequeces) ...```.

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +165,72 @@ public boolean equals(Object o) {
...
+        
+        return countBases + position;
+    }
+    
+    /** shortcut to <pre>countBases(getSequenceIndex(sequenceName), position);</pre> */
+    public long  countBases(String sequenceName,final int position)  {
```
nh13: perhaps ```getTotalBases``` instead?

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +165,72 @@ public boolean equals(Object o) {
...
+        return countBases(getSequenceIndex(sequenceName), position);
+    }    
+    
+    /**
```
nh13: I would be hesitant to include these utility functions here as there hasn't been a great need for them and they are relatively easy to compute by client code.

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +165,72 @@ public boolean equals(Object o) {
...
+        return mSequences.toString();
+    }
+    
+    /** 
+     * @param sequenceIndex contig id
+     * @param position position in the contig
+     * @returns the number of bases seen from the first sequence at position 0 to the sequenceIndex/position or -1 on error*/
+    public long countBases(final int sequenceIndex,final int position)  {
+        if (sequenceIndex < 0 || position <0 || sequenceIndex >= mSequences.size() ) {
+            return -1L;
```
nh13: I would rather throw an exception here, rather than relying on a return code.

lindenb:
@nh13  Thank you for taking the time to review this ! I'll make the changes later but if you think it's not worth it, do feel free to close this pull request.

While I'm here. Would you accept the following changes : 

* adding JAXB annotations (@XmlRootElement ... ) to SamSequenceRecord and SAMSequenceDictionary , so we can serialize things as XML/JSON (following the CA4GH schema ?)

* adding a method `void addAlias` to help resolving user's query with alternate contig names (chr1, 1, 01 ). Something like:

```java
public void addAlias (String baseName,String newName)
  {
  if(baseName.equals(newName)) return;
  SAMSequenceRecord ssr = getSequence(baseName);
  if( ssr == null ) throw new IllegalArgumentException("no sequence with that basename");
  if(  getSequence(newName) != null) throw new IllegalArgumentException("newName already used");
  this.mSequenceMap.put(newName,ssr);
  }
```

lindenb:
* I've updated my 'format' settings for eclipse (hoping it's ok...)
* in SamSequenceDictionary:
   * added getTotalBases (removed `fraction` as you suggested)
   * added addSequenceAlias : adds a new name to the dictionary map without adding a new sequence
   * added a toString verbose
   * added JaxbElement for XML serialization

* created  SamSequenceDictionaryTest to test SamSequenceDictionary XML serialization and aliases
* added JAXB annotations to AbstractSAMHeaderRecord, SamSequenceDictionary, SAMSequenceRecord
* in SAMSequenceRecord now using the predefined attributes instead of plain string

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +176,124 @@ public boolean equals(Object o) {
...
+     * @param originalName
+     *            existing contig name
+     * @param altName
+     *            new contig name
+     * @return the contig associated to the 'originalName/altName'
+     */
+    public SAMSequenceRecord addSequenceAlias(final String originalName,
+            final String altName) {
+        if (originalName == null)
+            throw new IllegalArgumentException("original name cannot be null");
```
nh13: either put this one one line, or use ```{...}```

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +176,124 @@ public boolean equals(Object o) {
         return true;
     }
 
+    /**
```
nh13: Probably a good idea to update the comment in ```setSequences``` to say that aliases will not be overwritten.

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +176,124 @@ public boolean equals(Object o) {
...
+        final SAMSequenceRecord altSeqRecord = getSequence(altName);
+        if (altSeqRecord != null) {
+            // alias was already set to the same record
+            if (altSeqRecord.equals(originalSeqRecord))
+                return originalSeqRecord;
+            // alias was already set to another record
+            throw new IllegalArgumentException("Alias " + altName
+                    + " was already set to " + altSeqRecord.getSequenceName());
+        }
+        mSequenceMap.put(altName, originalSeqRecord);
```
nh13: Oh no!  Now ```size()``` wont return the correct number of records!

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +176,124 @@ public boolean equals(Object o) {
...
+     * @return a MD5 checksum for this dictionary or the empty string if it is
+     *         empty
+     */
+    public String md5() {
+        if (isEmpty())
+            return "";
+        try {
+            final MessageDigest md5 = MessageDigest.getInstance("MD5");
+            md5.reset();
+            for (final SAMSequenceRecord ssr : mSequences) {
```
nh13: better variable name than ```ssr``` please.

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +176,124 @@ public boolean equals(Object o) {
...
     @Override
     public int hashCode() {
         return mSequences.hashCode();
     }
-}
\ No newline at end of file
+
+    @Override
+    public String toString() {
+        return "SAMSequenceDictionary:( sequences:"+ size()+
```
nh13: if we had an alias and one contig, the size would be two!

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +176,124 @@ public boolean equals(Object o) {
...
+     * @returns the number of bases seen from the first sequence at position 0
+     *          to the sequenceIndex/position or -1 on error
+     */
+    public long getTotalBases(final int sequenceIndex, final int position) {
+        if (sequenceIndex < 0 || position < 0
+                || sequenceIndex >= mSequences.size()) {
+            throw new IllegalArgumentException("getTotalBases("+sequenceIndex+","+position);
+        }
+        long countBases = 0L; /* number of bases seen so far */
+        for (int i = 0; i < sequenceIndex; i++)
```
nh13: ```{ ... }```

src/java/htsjdk/samtools/SAMSequenceDictionary.java
lindenb: I put a new key in `mSequenceMap` but size() gets its values from `mSequences` isn't it ?

```java
    public int size() {
        return mSequences.size();
    }
```

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -163,8 +176,124 @@ public boolean equals(Object o) {
...
+
+    /**
+     * @param sequenceIndex
+     *            contig id
+     * @param position
+     *            position in the contig
+     * @returns the number of bases seen from the first sequence at position 0
+     *          to the sequenceIndex/position or -1 on error
+     */
+    public long getTotalBases(final int sequenceIndex, final int position) {
```
nh13: Is it really appropriate to put utility functions like this here?  Could folks not compute this themselves?  I don't want to dump a ton of utility functions into these library classes.  Can you make a case for its usefulness for others?

nh13:
Minor comments, looking great.  Make sure to format the code similar to the rest of the library's code.  @lindenb

lindenb:
removed 'getTotalBases' at the end, fixed indent , added some comments.
' hope it's ok.

yfarjoun:
:+1: thanks!

PR #285, opened 2015-07-23
Changes to accommodate v1.5 of the SAM file specification

nh13:
See #233


src/java/htsjdk/samtools/SAMTag.java
```diff
@@ -27,6 +27,47 @@
...
+    QT,
+    Q2,
+    R2,
+    RG,
+    RT,
+    SA,
+    SM,
+    TC,
+    U2,
+    UQ
```
nh13: It is much easier to keep this updated via visual inspection with one tag per row.  Sorry for the verbosity.

src/java/htsjdk/samtools/SAMTag.java
```diff
@@ -27,6 +27,53 @@
...
+    RG,
+    RT,
+    S2, // for backwards compatibility
+    SA,
+    SM,
+    SQ, // for backwards compatibility
+    TC,
+    U2,
+    UQ
 }
```
ktibbett: Are these supposed to be in alphabetical order?
LB > H*
MD < MF
OC < OP, OQ
Letter+number should always go either before or after Letter+letter; it's a mix here

ktibbett:
Would it be useful to have one test that reads an older-version file successfully?

Assuming there are no newly required tags in 1.5 (i.e. that a 1.4 file is a valid 1.5 file) then this looks fine.

image

src/java/htsjdk/samtools/SAMTag.java
nh13: Same order as the spec.

PR #288, opened 2015-07-27
Fixed Liftover to handle strand correctly. Previously the...

tfenne:
…implemenation would ignore the strand of the interval passed in, and report lifted intervals as either + (mostly) or - if the chain mapped to the opposite strand.  Now correctly uses the the combination of the input interval's strand *and* the liftover chain information to determine the returned strand.

### DIFF
--- a/src/java/htsjdk/samtools/liftover/LiftOver.java
+++ b/src/java/htsjdk/samtools/liftover/LiftOver.java
@@ -116,7 +116,7 @@ public class LiftOver {
             return null;
         }
 
-        return createToInterval(interval.getName(), targetIntersection);
+        return createToInterval(interval.getName(), interval.isNegativeStrand(), targetIntersection);
     }
 
     public List<PartialLiftover> diagnosticLiftover(final Interval interval) {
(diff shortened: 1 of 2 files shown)


tfenne:
@alecw Could I ask you to take a look at this please since I think you wrote most of the liftover code?  It's only about ~3 lines of change.

src/java/htsjdk/samtools/liftover/LiftOver.java
```diff
@@ -156,8 +156,8 @@ private static Interval createToInterval(final String intervalName, final Target
             toEnd = negativeEnd;
         }
         // Convert to 1-based, inclusive.
-        return new Interval(targetIntersection.chain.toSequenceName, toStart+1, toEnd, targetIntersection.chain.toNegativeStrand,
-                intervalName);
+        final boolean strand = targetIntersection.chain.toOppositeStrand ? !sourceNegativeStrand : sourceNegativeStrand;
```
alecw: Hi Tim,

(Please ignore smell of burning dust from this portion of my memory ;-)

One tiny niggle: could you change the name of this variable from 'strand' to 'negativeStrand?'

Other than that, :+1: 

-Alec

tfenne:
Done!  Thanks @alecw

PR #289, opened 2015-07-27
Fix bug when reading fasta with sequence dictionary and trailing space.

alecw:


nh13:
@alecw what was thrown before the fix?

alecw:
[testng] FAILED: testTrailingWhitespaceWithPreexistingSequenceDictionary
   [testng] htsjdk.samtools.SAMException: Format exception reading FASTA testdata/htsjdk/samtools/reference/reference_with_trailing_whitespace.fasta.  Expected > but saw chr(32) at start of sequence with index 1
   [testng] 	at htsjdk.samtools.reference.FastaSequenceFile.readSequenceName(FastaSequenceFile.java:94)
   [testng] 	at htsjdk.samtools.reference.FastaSequenceFile.nextSequence(FastaSequenceFile.java:67)
   [testng] 	at htsjdk.samtools.reference.FastaSequenceFileTest.testTrailingWhitespaceWithPreexistingSequenceDictionary(FastaSequenceFileTest.java:79)
   [testng] ... Removed 21 stack frames

src/java/htsjdk/samtools/reference/FastaSequenceFile.java
```diff
@@ -145,6 +145,9 @@ private String readSequenceName() {
                     bases = tmp;
             }
         }
+        // When length is known, make sure there is no trailing whitespace that hasn't been traversed.
+        skipToEoln();
```
nh13: Why not put this in the if block on line 139 so it only applies to the 'length is known' case?  I am not sure it matters but I am being defensive.

alecw: Will do.  I agree that is clearer, and I think defensiveness is warranted.

alecw:
OK, code review changes pushed.

Issue #175, opened 2015-02-20
Publish "unstable" build on each push to master?


mattsooknah:
(related to https://github.com/samtools/htsjdk/pull/115)

There's been some interest in having an "unstable" jar published to Maven on each push to master, in addition to the "stable" release every ~2 weeks. 

This would be useful for larger projects that depend on HTSJDK where a few members of the project want to contribute features to HTSJDK, without everyone having to worry about syncing their local versions of HTSJDK before building the other project.

I'm not sure on the best way to do this. We'd also need to make it very clear that it's for experimental use only. In the short term, a fork might be an easier solution, but let's keep this on the radar.

@lbergelson @akiezun @droazen are interested in this.

lbergelson:
I am indeed interested in this.  I haven't gotten a chance to look into it yet much though.

nh13:
Have a suggestion when I get back in next week. Thanks for tracking this!

N

Thumb typed for added typos

lbergelson:
@nh13 What was your suggestion for this?

nh13:
Responded in person, posting for posterity.

vdauwera:
I am also interested in this. What needs to happen to make this happen *soon*?

kshakir:
I'll touch base on Monday. Setup is on the same order of magnitude as the existing Maven Central releases. One can publish pre-releases and/or snapshots. As htsjdk if foss, the repo can be sonatype, or use our artifactory.broadinstute.org. Search "artifactory" on the dsde wiki for a little more info for now.

vdauwera:
Looking fwd to this happening! We are in need of this right now for the issue referenced above.

droazen:
Any word on this?

kshakir:
I pinged GG last week with the following:

> I have more info on publishing HTSJDK *pre-*releases. Happy to chat when you have time, and with more folks if you want to schedule a chat of involved parties. I assume because HTSJDK is OSS, you'll keep using the same host Sonatype to host these new jars. Next steps: we'll need to know 1) Who / what username currently pushes releases (assuming they will keep using their account) and 2) What versioning you want for the pre-release jars: -SNAPSHOT always downloading the latest, or git labeled versions where the consumer has to explicitly lookup the latest

Once the above is resolved, that person just needs to point the build at the right `-SNAPSHOT` or `release` repo: http://central.sonatype.org/pages/ossrh-guide.html#accessing-repositories

kshakir:
[...]

PR #294, opened 2015-08-02
Enabling -SNAPSHOT versioning including git-hashes.

kshakir:
For backwards compatibility for whomever is releasing, default to non-snapshot.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ import sbt.Package.ManifestAttributes
 
 name := "htsjdk"
 
-version := "1.137"
+val buildVersion = "1.137"
 
 organization := "com.github.samtools"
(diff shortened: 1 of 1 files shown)


kshakir:
This is the next step in #175. If desired, the same changes can be applied to Picard.

Left the publish endpoints alone inside the build.sbt. Will change them on the Travis command line, along with providing the rest of the publish credentials.

lbergelson:
Looks good to me but I don't have the password needed to test it.  Can @jacarey take a look at this.  (it will relieve pressure from me asking for releases all the time.)

vdauwera:
Oooooh :+1: (not a code review, just an expression of happiness)

Issue #296, opened 2015-08-03
SamReaderFactory does not pass discovered index files into CRAMFileReader


droazen:
If a *.bai index is found for a cram file via the `SamFiles.findIndex()` call in `SamReaderFactory.open()`, that index file does not propagate to the `CRAMFileReader` returned from `open()` (unlike the `BAMFileReader` case, where the index is passed in to the reader).

droazen:
This was resolved in https://github.com/samtools/htsjdk/pull/299

PR #299, opened 2015-08-10
Hooking up the index file to CRAM files in SamReaderFactory

edwardkw:
This fixes issue #296.


src/tests/java/htsjdk/samtools/SAMFileReaderTest.java
```diff
@@ -53,6 +53,23 @@ public void variousFormatReaderTest(final String inputFile) {
...
+    // tests for CRAM indexing
+
+    @Test(dataProvider = "")
+    public void CRAMIndexTest(final String inputFile) {
+        final File input = new File(TEST_DATA_DIR, inputFile);
+        final SamReader reader = SamReaderFactory.makeDefault().open(input);
+        Assert.assertTrue(reader.hasIndex());
+    }
+
+    @DataProvider(name = "")
```
droazen: You need to give your `DataProvider` a name here, and use that name in the `@Test` annotation above.

src/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -268,6 +268,9 @@ public SamReader open(final SamInputResource resource) {
                             bufferedStream.close();
                             bufferedStream = null;
                         }
+
+                        bufferedStream = IOUtil.maybeBufferInputStream(indexMaybe.asUnbufferedInputStream());
```
droazen: You should always close the previous `bufferedStream` connected to `sourceFile` here before creating your new input stream on the index.

droazen: You also need to handle the case where there is no index (ie., `indexMaybe` == null)

droazen: Actually, instead of opening an input stream on the index here, recommend switching to the constructor:

        public CRAMFileReader(final File cramFile, final File indexFile, final ReferenceSource referenceSource)

as I'm not sure I trust the constructors that take a stream, since the javadoc for those constructors contradicts itself (in one place the `inputStream` is described as `CRAM stream to read` and in another place that same parameter is described as `index file to be used for random access`).

vadimzalunin:
it appears I've been working on the same issue, please see this branch: 
https://github.com/samtools/htsjdk/tree/CRAI_index
I meant to create a PR yesterday but somehow screwed up the master instead :o 
and then I saw this PR, so not sure now. 

Could you please check if CRAI_index branch achieves your goal as well?

edwardkw:
@vadimzalunin I saw your changes, and yes we're working on the same issue :)
Your changes were helpful for me to look at, as well. 
Would you like to be the one who fixes this issue, or would you prefer to leave it to me?

vadimzalunin:
I would prefer my version because GATK CRAM patch is based on it. This is a long shot.

edwardkw:
I will leave it to you then. I will close this PR.

droazen:
Hi Vadim,

We actually need this patch from @edwardkw in today, as there's going to be an htsjdk release and GATK4 needs this change. Since Edward's change appears to be compatible with your branch, I hope you don't object to our merging this now while we wait for your upcoming PR.

vadimzalunin:
no problem, i'll rebase mine branch afterwards

droazen:
:+1:

PR #302, opened 2015-08-12
Simple exception addition

edwardkw:
[Hellbender #793](https://github.com/broadinstitute/hellbender/issues/793)


src/java/htsjdk/samtools/CRAMIterator.java
```diff
@@ -156,6 +156,9 @@ private void nextContainer() throws IOException, IllegalArgumentException,
             final SAMSequenceRecord sequence = cramHeader.getSamFileHeader()
                     .getSequence(container.sequenceId);
             refs = referenceSource.getReferenceBases(sequence, true);
+            if (refs == null) {
+                throw new RuntimeException(String.format("Contig %s not found in the reference file.", sequence.getSequenceName()));
```
lbergelson: Can we make this a more specific exception?  Maybe `CRAMReferenceException` extending from `SAMException`

droazen: This class uses `RuntimeException` everywhere else -- it would be inconsistent to use a more specific exception here.

droazen:
Looks good -- can you add a unit test that triggers this error?

src/java/htsjdk/samtools/CRAMIterator.java
droazen: Took a second look at the class, and I take it back -- it is using `RuntimeException`, but only for generic errors like "not implemented". `CramFileReader` uses `SAMException` in a couple places, so I think there's precedence for just using that (or a new subclass of that) here.

droazen:
@edwardkw still need that unit test before this can be merged

.idea/modules/htsjdk.iml
```diff
@@ -39,5 +39,6 @@
         <SOURCES />
       </library>
     </orderEntry>
+    <orderEntry type="library" name="lib" level="project" />
```
droazen: Revert change to `htsjdk.iml`

src/tests/htsjdk/samtools/CRAMExceptionTest.java
```diff
@@ -0,0 +1,38 @@
+package htsjdk.samtools;
```
droazen: Can you move `CRAMEdgeCasesTest` back to its original location in `src/tests/java/htsjdk/samtools`, move the test case in this file (`CRAMExceptionTest`) into `CRAMEdgeCasesTest`, and then delete this file?

src/tests/htsjdk/samtools/CRAMExceptionTest.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public class CRAMExceptionTest {
+
+    @BeforeTest
+    public void beforeTest() {
+        Log.setGlobalLogLevel(Log.LogLevel.ERROR);
+    }
+
+    @Test(expectedExceptions = CRAMException.class)
+    public void testRefNull() throws IOException {
```
droazen: Test method should have a more accurate name like `testContigNotFoundInRef()`

testdata/htsjdk/samtools/cram/test.dict
```diff
@@ -1,2 +1,3 @@
 @HD	VN:1.4	SO:unsorted
 @SQ	SN:Sheila	LN:20	M5:7ddd8a4b4f2c1dec43476a738b1a9b72	UR:file:/Users/edwardk/Documents/htsjdk/testdata/htsjdk/samtools/cram/auxf.fa
+@SQ	SN:chr1	LN:8	M5:cc0af3a4fedb18378b4b57b98068e69f
```
droazen: As discussed in person, the reference here should not have the contig Sheila in order to trigger your new check (your test case was not actually being run in travis due to the test class being in the wrong place)

src/tests/htsjdk/samtools/CRAMExceptionTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+
+/**
+ * Testing
+ */
+public class CRAMExceptionTest {
```
droazen: Can you delete this file in git, now that it's empty?

src/tests/java/htsjdk/samtools/CRAMEdgeCasesTest.java
```diff
@@ -1,15 +1,16 @@
-package htsjdk.samtools;
+package java.htsjdk.samtools;
```
droazen: Package name is wrong here, which is why travis is failing -- should be just `htsjdk.samtools`

src/tests/java/htsjdk/samtools/CRAMEdgeCasesTest.java
```diff
@@ -33,6 +34,25 @@ public void testUnsorted() throws IOException {
...
+        final ReferenceSource refSource = new ReferenceSource(refFile);
+        final CRAMIterator iterator = new CRAMIterator(new FileInputStream(CRAMFile), refSource);
+        try {
+            while (iterator.hasNext()) {
+                iterator.next();
+            }
+        }
+        // the iterator is wrapping our exception; ensuring that it was actually a CRAMexception
+        catch (RuntimeEOFException e) {
+            Assert.assertEquals(e.getCause().getClass(), CRAMException.class);
```
droazen: You also need to keep track of whether you got any exception at all, and assert that you did at the end of this test (since you can't use the built-in `expectedExceptions` here due to the wrapping.

src/tests/java/htsjdk/samtools/CRAMEdgeCasesTest.java
```diff
@@ -33,6 +34,25 @@ public void testUnsorted() throws IOException {
         testRecords(records, records.iterator().next().getReadBases());
     }
 
+    // int test for CRAMException
+    // testing for a contig found in the reads but not in the reference
+    @Test
+    public void testContigNotFoundInRef() throws IOException {
+        final File CRAMFile = new File("testdata/htsjdk/samtools/cram/test.cram");
+        final File refFile = new File("testdata/htsjdk/samtools/cram/test.fa");
```
droazen: Rename these test files so as to make it clear that they belong to this test -- eg., `testContigNotInRef.cram` and `testContigNotInRef.fa`

PR #305, opened 2015-08-13
allow lazy symlinking of bams

lbergelson:
This is to enable a feature request from the GATK forums. 
https://github.com/broadinstitute/hellbender/issues/781

Allow htsjdk to find a bai file at the actual path of a symlinked bam without having to symlink the bai as well.  

So when looking for an index:
*  Look for the bai or bam.bai at the given file path
*  If not file was found, find the canonical path of the bam and then look there for the bam or bai

This is going to have a minor conflict with @vadimzalunin pull request so whichever goes in last will have to rebase on the other one.

[Review on Reviewable](https://reviewable.io/reviews/samtools/htsjdk/305)

### DIFF
--- a/src/java/htsjdk/samtools/SamFiles.java
+++ b/src/java/htsjdk/samtools/SamFiles.java
@@ -4,6 +4,7 @@ import htsjdk.samtools.cram.CRAIIndex;
 import htsjdk.samtools.cram.build.CramIO;
 
 import java.io.File;
+import java.io.IOException;
 
 /**
  * @author mccowan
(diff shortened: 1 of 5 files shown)


src/java/htsjdk/samtools/SamFiles.java
```diff
@@ -1,18 +1,33 @@
...
+     * Finds the index file associated with the provided SAM file.  The index file must exist and be reachable to be found.
+     *
+     * If the file is a symlink and the index cannot be found, try to unsymlink the file and look for the bai in the actual file path.
      *
      * @return The index for the provided SAM, or null if one was not found.
      */
-    public static File findIndex(final File samFile) {
-        // If input is foo.bam, look for foo.bai
+    public static File findIndex(final File samFile){
+        final File indexFile = lookForIndex(samFile); //try to find the index
```
yfarjoun: chanelling Nils...
spaces between ) and { and between if and (

lbergelson: lets choose a linter and use that automatically :)

yfarjoun:
sure. intellJ, my settings :-)

droazen:
This looks good and can be merged, but must be rebased first. @lbergelson please rebase and merge when you get a chance.

yfarjoun:
this keeps getting pushed down...please rebase again and merge!

lbergelson:
right, I should do this...

tfenne:
@lbergelson I thought I had commented on this before - was there a different version of this PR at some point?  I feel like this is actually a somewhat dangerous change.  At the very least I think there should be some way to enable/disable this behavior in HTSJDK.  It would seem very surprising to me to have a library follow a link, and then browse around the parent directory of the link target for associated files.

lbergelson:
@tfenne I don't remember ever seeing a comment from you before on this.  

This came from a gatk forum feature request.  Someone was frustrated because they kept forgetting to make symlinks for their bai files.  I sympathised since I too have been frustrated by forgetting to make symlinks for my bai files.  

The only thing I was worried about with it was that it would behave differently between systems causing the behavior to change across file systems, but I figured since some filesystems don't support symlinks at all, it is already inconsistent that way.

If you think it's dangerous we can drop it, (I totally forgot about it for several months which should give some indication about my overall level of caring about this particular feature.)  I'm not sure I totally understand your objection though.  It doesn't seem unreasonable to me that a program would try to get the canonical path of a symlink and look around there. 

I don't have very strong expectations about how symlinked files should work though, it seems like there is very inconsistent treatment of them across different software, sometimes I've seen software just refuse to resolve symlinks at all.  

I'm open to either just dropping this pull request, or adding some toggle somewhere.  If your preference would be defaulting to off, I'd rather just drop it though, since I suspect it would be essentially undiscoverable.

tfenne:
@lbergelson Apologies, I must have thought about this when you opened the PR months ago and the forgotten to comment.  My main objection is that I could also see people (me included) getting very confused by this, as it's not what i would expect.

I'd prefer to either
1) Add a warning anytime we resolve an index this way (similar to the way HTSJDK warns about indices older than their BAMs), so it's obvious what's happening, or
2) Not have the feature

The only other option to throw out there would be to have it optional, and have the default behavior be *not* to use the index beside the symlink target, but to tell you if it found it and how to enable it's use.  But that seems cumbersome.  So how about just a warning?

yfarjoun:
I think that option 1) is very reasonable.

lbergelson:
I've added a warning to explain what's going on.  @tfenne Do you think this is sufficient?  

Your third option is a bit awkward, but I can implement that instead if you feel more comfortable with it.

tfenne:
Works for me.  Thanks @lbergelson

lbergelson:
@tfenne Thanks for your input!

PR #308, opened 2015-08-20
Add support for java.nio.file.Path in the reference package.

tomwhite:
This change is to make it possible to read reference sequence files (including indexes and dictionaries) from filesystems other than the local filesystem. The basic idea is to overload constructors and methods that take `java.io.File` to take `java.nio.file.Path`.

In particular, Hadoop can be supported by means of https://github.com/damiencarol/jsr203-hadoop.

Note that this requires Java 7 or above. HTSJDK currently supports Java 6, but only until October 2015, so it should be possible to use this approach after then.

### DIFF
--- a/src/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
+++ b/src/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
@@ -32,14 +32,15 @@ import htsjdk.samtools.util.BufferedLineReader;
 import htsjdk.samtools.util.IOUtil;
 
 import java.io.File;
-import java.io.FileInputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
 
 /**
  * Provide core sequence dictionary functionality required by all fasta file readers.
  * @author Matt Hanna
  */
 abstract class AbstractFastaSequenceFile implements ReferenceSequenceFile {
-    protected final File file;
+    private final Path path;
     protected SAMSequenceDictionary sequenceDictionary;
 
     /**
@@ -47,15 +48,23 @@ abstract class AbstractFastaSequenceFile implements ReferenceSequenceFile {
      * @param file Fasta file to read.  Also acts as a prefix for supporting files.
      */
     AbstractFastaSequenceFile(final File file) {
-        this.file = file;
-        final File dictionary = findSequenceDictionary(file);
+        this(file == null ? null : file.toPath());
+    }
+
+    /**
+     * Finds and loads the sequence file dictionary.
+     * @param path Fasta file to read.  Also acts as a prefix for supporting files.
+     */
+    AbstractFastaSequenceFile(final Path path) {
+        this.path = path;
+        final Path dictionary = findSequenceDictionary(path);
 
         if (dictionary != null) {
             IOUtil.assertFileIsReadable(dictionary);
 
             try {
                 final SAMTextHeaderCodec codec = new SAMTextHeaderCodec();
-                final BufferedLineReader reader = new BufferedLineReader(new FileInputStream(dictionary));
+                final BufferedLineReader reader = new BufferedLineReader(Files.newInputStream(dictionary));
                 final SAMFileHeader header = codec.decode(reader,
                         dictionary.toString());
                 if (header.getSequenceDictionary() != null && header.getSequenceDictionary().size() > 0) {
@@ -70,31 +79,49 @@ abstract class AbstractFastaSequenceFile implements ReferenceSequenceFile {
     }
 
     protected static File findSequenceDictionary(final File file) {
+        if (file == null) {
+            return null;
+        }
+        Path dictionary = findSequenceDictionary(file.toPath());
+        if (dictionary == null) {
+            return null;
+        }
+        return dictionary.toFile();
+    }
+
+    protected static Path findSequenceDictionary(final Path path) {
+        if (path == null) {
+            return null;
+        }
         // Try and locate the dictionary
-        String dictionaryName = file.getAbsolutePath();
-        String dictionaryNameExt = file.getAbsolutePath();
+        Path dictionary = path.toAbsolutePath();
+        Path dictionaryExt = path.toAbsolutePath();
         boolean fileTypeSupported = false;
         for (final String extension : ReferenceSequenceFileFactory.FASTA_EXTENSIONS) {
-            if (dictionaryName.endsWith(extension)) {
-                  dictionaryNameExt = new String(dictionaryName);
-                  dictionaryNameExt += IOUtil.DICT_FILE_EXTENSION;
-                  dictionaryName = dictionaryName.substring(0, dictionaryName.lastIndexOf(extension));
-                  dictionaryName += IOUtil.DICT_FILE_EXTENSION;
-                  fileTypeSupported = true;
-                  break;
+            String filename = dictionary.getFileName().toString();
+            if (filename.endsWith(extension)) {
+                dictionaryExt = dictionary.resolveSibling(filename + IOUtil
+                    .DICT_FILE_EXTENSION);
+                String filenameNoExt = filename.substring(0, filename.lastIndexOf(extension));
+                dictionary = dictionary.resolveSibling(filenameNoExt+ IOUtil.DICT_FILE_EXTENSION);
+                fileTypeSupported = true;
+                break;
             }
         }
         if (!fileTypeSupported)
-            throw new IllegalArgumentException("File is not a supported reference file type: " + file.getAbsolutePath());
+            throw new IllegalArgumentException("File is not a supported reference file type: " + path.toAbsolutePath());
 
-        final File dictionary = new File(dictionaryName);
-        if (dictionary.exists())
+        if (Files.exists(dictionary))
             return dictionary;
         // try without removing the file extension
-        final File dictionaryExt = new File(dictionaryNameExt);
-        if (dictionaryExt.exists())
+        if (Files.exists(dictionaryExt))
             return dictionaryExt;
         else return null;
+    }
+
+    /** Returns the path to the reference file. */
+    protected Path getPath() {
+        return path;
     }
 
     /**
@@ -106,8 +133,13 @@ abstract class AbstractFastaSequenceFile implements ReferenceSequenceFile {
     }
 
     /** Returns the full path to the reference file. */
+    protected String getAbsolutePath() {
+        return path.toAbsolutePath().toString();
+    }
+
+    /** Returns the full path to the reference file. */
     public String toString() {
-        return this.file.getAbsolutePath();
+        return getAbsolutePath();
     }
 
     /** default implementation -- override if index is supported */
@@ -120,7 +152,7 @@ abstract class AbstractFastaSequenceFile implements ReferenceSequenceFile {
 
     /** default implementation -- override if index is supported */
     public ReferenceSequence getSubsequenceAt( String contig, long start, long stop ) {
-        throw new UnsupportedOperationException("Index does not appear to exist for" + file.getAbsolutePath() + ".  samtools faidx can be used to create an index");
+        throw new UnsupportedOperationException("Index does not appear to exist for " + getAbsolutePath() + ".  samtools faidx can be used to create an index");
     }
 
 }
--- a/src/java/htsjdk/samtools/reference/FastaSequenceFile.java
+++ b/src/java/htsjdk/samtools/reference/FastaSequenceFile.java
@@ -32,6 +32,7 @@ import htsjdk.samtools.util.IOUtil;
 import htsjdk.samtools.util.StringUtil;
 
 import java.io.File;
+import java.nio.file.Path;
 
 /**
  * Implementation of ReferenceSequenceFile for reading from FASTA files.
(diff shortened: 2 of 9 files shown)


zong0jie:
I like your idea, and it seems can change very little things to make htsjdk support another filesystems. Still I think create io-streams form htsjdk's IOUtil class is a better choice.
I have done some other works in my branch(I use the branch in production environment) and I don't know whether it is useful for your work.

droazen:
@tomwhite Do we really want to wait until October to merge this PR? Do you feel that this approach has major advantages over the one in https://github.com/samtools/htsjdk/pull/211 that would justify the wait? Can you explain what the advantages are?

Since there are now multiple proposals on the table that address this issue, it would be good to come to a consensus soon on the right approach for htsjdk, as https://github.com/samtools/htsjdk/pull/211 has been open for quite a while, and downstream projects like https://github.com/broadinstitute/hellbender would benefit greatly from this change.

tomwhite:
@droazen The advantage of using java.nio.file is that it's a Java standard, which supports multiple implementations, whereas the approach in https://github.com/samtools/htsjdk/pull/211 is a point solution for Hadoop. The latter involves overriding java.io.File, which was not designed to be overridden. It might work well for Hadoop, but it may not for other file system implementations.

So I think that java.nio.file is the right solution in the long term, but an approach like https://github.com/samtools/htsjdk/pull/211 would be OK to support Hadoop in the short term (i.e. before October).

zong0jie:
@tomwhite @droazen  java.nio,file is one thing, maybe it is better than my solution. 

Another thing is some people just new a java.io.File or a java.io.FileInputStream object in anywhere. And with these code the project can hardly support other filesystems.

So I think the most important thing is to initialize the File object and io-stream in one place, for example in htsjdk.samtools.util.IOUtil. As well as SeekableStream form SeekableStreamFactory. 

If we can obey the rule than it will be easy to do things such as java.nio.file, and most of the code will not be changed.

tomwhite:
I made a minor update to support CRAM files (for https://github.com/HadoopGenomics/Hadoop-BAM/issues/24).

tomwhite:
All tests pass now that htsjdk uses Java 8.

src/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
```diff
@@ -40,6 +42,7 @@
  */
 abstract class AbstractFastaSequenceFile implements ReferenceSequenceFile {
     protected final File file;
+    protected final Path pathFile;
```
lbergelson: I'm a bit worried that this will cause confusing bugs in the future.  It's really unfortunate that `file` is a protected field used by subclasses instead of a property.  I think in this case it might be worth breaking the existing api to avoid having 2 parallel fields holding the same data.

src/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
```diff
@@ -69,7 +73,38 @@
         }
     }
 
+    /**
+     * Finds and loads the sequence file dictionary.
+     * @param file Fasta file to read.  Also acts as a prefix for supporting files.
+     */
+    AbstractFastaSequenceFile(final Path file) {
```
lbergelson: It seems like this is essentially a copy/paste of the other constructor.  Can the file constructor delegate to this one instead?

src/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
```diff
@@ -97,6 +132,35 @@ protected static File findSequenceDictionary(final File file) {
         else return null;
     }
 
+    protected static Path findSequenceDictionary(final Path file) {
```
lbergelson: likewise, can the file version of findSequenceDictionary delegate to this instead?

src/java/htsjdk/samtools/reference/FastaSequenceFile.java
```diff
@@ -53,6 +54,13 @@ public FastaSequenceFile(final File file, final boolean truncateNamesAtWhitespac
         this.in = new FastLineReader(IOUtil.openFileForReading(file));
     }
 
+    /** Constructs a FastaSequenceFile that reads from the specified file. */
```
lbergelson: old constructor should delegate to this one.  Possibly could benefit from adding an `IOUtil.nullFie(File file)` or something like that

src/java/htsjdk/samtools/reference/IndexedFastaSequenceFile.java
```diff
@@ -92,6 +95,31 @@ public IndexedFastaSequenceFile(final File file) throws FileNotFoundException {
         this(file, new FastaSequenceIndex((findRequiredFastaIndexFile(file))));
     }
 
+    /**
+     * Open the given indexed fasta sequence file.  Throw an exception if the file cannot be opened.
+     * @param file The file to open.
```
lbergelson: if the other constructor can delegate to this one that would be great

src/java/htsjdk/samtools/reference/IndexedFastaSequenceFile.java
```diff
@@ -202,10 +251,20 @@ public ReferenceSequence getSubsequenceAt( String contig, long start, long stop
             startOffset += Math.max((int)(startOffset%bytesPerLine - basesPerLine + 1),0);
 
             try {
-                 startOffset += channel.read(channelBuffer,indexEntry.getLocation()+startOffset);
+                if (channel instanceof FileChannel) {
+                    startOffset += ((FileChannel) channel).read(channelBuffer,indexEntry.getLocation()+startOffset);
```
lbergelson: Is this file channel method more performant than the method in the `else`?  I.e. is there any reason to have 2 paths through here instead of just the 1.  The file channel method is obviously more readable, but if it doesn't offer any other advantage I'd probably go with just the second path through the code. 

If you think it's worth have the `if else` in there to handle the file channel case, could you extract this logic into a method.

src/java/htsjdk/samtools/reference/IndexedFastaSequenceFile.java
```diff
@@ -202,10 +251,20 @@ public ReferenceSequence getSubsequenceAt( String contig, long start, long stop
...
+                        channel.position(indexEntry.getLocation()+startOffset);
+                        startOffset += channel.read(channelBuffer);
+                    } finally {
+                        channel.position(oldPos);
+                    }
+                }
             }
             catch(IOException ex) {
-                throw new SAMException("Unable to load " + contig + "(" + start + ", " + stop + ") from " + file);
+                throw new SAMException("Unable to load " + contig + "(" + start + ", " + stop + ") from " + getAbsolutePath());
```
lbergelson: Could you add `ex` as the cause to this exception.  I'm worried there is a case where the finally clause will throw and without the original exception it won't be clear what's going on.

src/java/htsjdk/samtools/reference/ReferenceSequenceFileFactory.java
```diff
@@ -98,4 +100,57 @@ public static ReferenceSequenceFile getReferenceSequenceFile(final File file, fi
...
     }
+
+    /**
+     * Attempts to determine the type of the reference file and return an instance
+     * of ReferenceSequenceFile that is appropriate to read it.  Sequence names
+     * will be truncated at first whitespace, if any.
+     *
+     * @param file the reference sequence file on disk
+     */
+    public static ReferenceSequenceFile getReferenceSequenceFile(final Path file) throws IOException {
```
lbergelson: could we deduplicate the code here by having the old file constructors point to these constructors?

src/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -334,6 +336,26 @@ else if (!file.canRead()) {
     }
 
     /**
+     * Checks that a file is non-null, exists, is not a directory and is readable.  If any
+     * condition is false then a runtime exception is thrown.
+     *
+     * @param file the file to check for readability
+     */
+    public static void assertFileIsReadable(final Path file) {
```
lbergelson: I think the file version of this could delgate to here

lbergelson:
@tomwhite I'm glad we can finally merge this.  I have a few comments though.  If I understand correctly, any `File` can be represented by a `Path`, but every `Path` can't necessarily be translated to a `File`.  So I think we should avoid duplicating any of the logic for files, and instead make the path version of the code the "canonical version", having all the file code point to the new path versions.  It seems like the only complications should be the `protected File file` and the annoyance of having to check for nulls before conversions.  

I think it would be better to break backwards compatibility here, and we can encapsulate the fields to prevent this from happening again.  The alternative is we get into a situation with nearly identical but slightly different fields that will inevitably cause bugs somewhere.  We can provide accessors which will provide a drop in replacement for anyone relying on file.  It seems that file can already be null, so it's fine to return that if we couldn't convert from Path.  @tfenne Do you think think this is an acceptable API change?  We're doing major version bump for the next version anyway, so I think this not unreasonable.

src/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
tfenne: Also could you change the parameter name to `path` please - it's confusing have `Path` objects named `file`.

src/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
tfenne: Also likewise on `file` -> `path`

src/java/htsjdk/samtools/reference/IndexedFastaSequenceFile.java
tfenne: Agree with @lbergelson - in fact doesn't this change the semantics of the operation depending on whether it's a FileChannel or not?  If it's a FileChannel you end up with the channel pointing at the end of the last read, but if it's any other channel it gets reset back to the prior position.  Seems undesirable to have differing behaviour based on the underlying kind of channel

tfenne:
@lbergelson I agree - it's going to be a breaking change any which way this is implemented.  Conceivably someone could have subclassed and accessed the protected `file` attribute - but even if it were via a method access, there will now be time when file is null/not-derivable from `path`, so it could break anyway.

All of that said, I doubt it will have much impact to get rid of the internal `file` reference altogether, make `path` private, and provide an accessor for `path`.

Overall I like the direction @tomwhite is taking things here, but agree that there needs to be some serious de-duplication of code.

src/java/htsjdk/samtools/reference/IndexedFastaSequenceFile.java
tomwhite: FileChannel takes advantage of native code, so I wanted to leave that code path unchanged. I've now split it out into a method as @lbergelson suggested.

@tfenne the semantics are the same for both, since after a positioned read FileChannel does not change the position; see http://docs.oracle.com/javase/7/docs/api/java/nio/channels/FileChannel.html#read(java.nio.ByteBuffer,%20long), where it says "This method does not modify this channel's position."

tomwhite:
Thanks for all the review feedback, @lbergelson and @tfenne. I wanted to be conservative with the addition of Path support, so I originally left the File-based code paths unchanged. I've updated this PR to change that however, so that the amount of code duplication is reduced.

> All of that said, I doubt it will have much impact to get rid of the internal `file` reference altogether, make path private, and provide an accessor for `path`.

Done

src/java/htsjdk/samtools/reference/IndexedFastaSequenceFile.java
```diff
@@ -235,6 +274,29 @@ public ReferenceSequence getSubsequenceAt( String contig, long start, long stop
...
+     * Reads a sequence of bytes from this channel into the given buffer,
+     * starting at the given file position.
+     * @param channel the channel to read from
+     * @param buffer the buffer into which bytes are to be transferred
+     * @param position the position to start reading at
+     * @return the number of bytes read
+     * @throws IOException if an I/O error occurs while reading
+     */
+    private static int readFromPosition(final SeekableByteChannel channel, final ByteBuffer buffer, long position) throws IOException {
+        if (channel instanceof FileChannel) { // special case to take advantage of native code path
```
lbergelson: Thanks for the explanation!

src/java/htsjdk/samtools/reference/ReferenceSequenceFileFactory.java
```diff
@@ -78,24 +79,58 @@ public static ReferenceSequenceFile getReferenceSequenceFile(final File file, fi
      * @param preferIndexed if true attempt to return an indexed reader that supports non-linear traversal, else return the non-indexed reader
      */
     public static ReferenceSequenceFile getReferenceSequenceFile(final File file, final boolean truncateNamesAtWhitespace, final boolean preferIndexed) {
-        final String name = file.getName();
+        return getReferenceSequenceFile(file.toPath(), truncateNamesAtWhitespace, preferIndexed);
```
lbergelson: should we worry about null file dereference?

lbergelson:
Looks good.  Squash and we can merge.

src/java/htsjdk/samtools/reference/ReferenceSequenceFileFactory.java
tomwhite: I didn't, as it preserves the existing behaviour.

tomwhite:
@lbergelson thanks for taking another look. I've squashed this now.

lbergelson:
:+1:

PR #313, opened 2015-08-26
NPE on chunkless query for plaintext file

AlpenDitrix:
Hey, @yfarjoun, as you requested, I've separated bug fix here.

Bug in a nutshell: method `htsjdk.samtools.BinningIndexContext#getChunksOverlapping(int,int)` on occasion returned not an empty list but just `null`

Generally people work with compressed files which queries were handled with help of `htsjdk.tribble.readers.TabixReader`. And he's doing his job in a proper way: if no cnuks were found on specified region, empty iterator was returned:

    421:    for (i = n_off = 0; i < n_bins; ++i) {
    422:        if ((chunks = idx.b.get(bins[i])) != null)
    423:            n_off += chunks.length;
    424:    }
    425:    if (n_off == 0) return EOF_ITERATOR;

But for querying from plain-text VCF file `htsjdk.tribble.TribbleIndexedFeatureReader` is used. So if you tried to do a query on a region, where actually no one chunk is exist, you will obtain this NPE:

	at htsjdk.tribble.index.tabix.TabixIndex.getBlocks(TabixIndex.java:152)
	at htsjdk.tribble.TribbleIndexedFeatureReader.query(TribbleIndexedFeatureReader.java:258)
	at htsjdk.variant.vcf.VCFFileReader.query(VCFFileReader.java:141)
	at htsjdk.tribble.index.tabix.TabixIndexTest.testQueryProvidedItemsAmount(TabixIndexTest.java:121)

The following two strings caused it when `chunks` was supplied with `null` value

        151:    final List<Chunk> chunks = indices[sequenceIndex].getChunksOverlapping(start, end);
        152:    final List<Block> ret = new ArrayList<Block>(chunks.size());

And cause of all these is the next mistake in the method `getChunksOverlapping(int,int)` of `htsjdk.samtools.BinningIndexContext` class

    136:    if (chunkList.isEmpty()) {
    137:        return null;
    138:    } 

And the fix is just a one line.

A tiny description of test is presented in 31bac69 commit, but you can ask more if needed.


droazen:
@AlpenDitrix According to the javadoc for the method in question (`htsjdk.samtools.BinningIndexContext#getChunksOverlapping(int,int)`), it returns a "List of Chunks overlapping the given region.  May return null if there are none." So the method appears to be behaving as designed. With your patch, the method seems inconsistent -- it returns an empty list in the `chunkList.isEmpty()` case, but null in the `overlappingBins == null` case. 

Rather than change the existing contract for this method, perhaps it would be safer/more conservative to fix the client code that's calling it (ie., `htsjdk.tribble.index.tabix.TabixIndex.getBlocks()`) to be able to handle a null return value from `getChunksOverlapping()`. What do you and @yfarjoun think?

yfarjoun:
Given that this is the contract, I don't see this as a bug in the HTS code
either.

I agree with @droazen that an inconsistent contract is not desired.

AlpenDitrix:
My opinion was based on two things:
1) as I think `null`s are from "similar nature" as an exceptions, so they must not be used as a real thing in alghoritms. Because of this `Collections.empty...()` exists; later Java 8 provided `Optional<T>`. So returning a `null` instead of empty collection (or empty array, or empty string) is a bad behavior and breaking of polymorphism and in some books about OOP was said the same. You can google 'Special Case Pattern' to find more stuff about it.
2) I found that TabixReader doesn't have 'return null' lines. That confirmed my opinion that null-instead-list behavior is wrong

So if it's a contract -- then it's a bad contract and that fact that I found that someone written method which occasionally throwing NPE in the middle of computation without catching only confirms that. 
Absolutely different *Special case object* isn't required here because there is no difference between cases of "why we did not get any chunks", so empty collection is suitable solution.


So here's for you to decide: if I hadn't changed your mind and you still want to leave that antipattern contract, I'll revert `getChunksOverlapping` it and add some checks into `getBlock` to prevent that NPE. Otherwise, I'll remove that contract from javadoc (after your answer)

droazen:
I agree that returning null is not ideal from a design standpoint, but consider that we have to be cautious about changing existing method contracts in a library that's used by multiple downstream projects. It's better fix the issue by changing client code and leaving existing method contracts as-is, rather than changing method contracts and risking unintended consequences downstream.

But I'll defer to @yfarjoun to make the final call on this one.

yfarjoun:
I agree @droazen here.

If you'd like to change the contract, I think that it should be opened to
discussion as an issue first to collect user input and look into usages of
said function. for now, I'd ask that you fix the downstream function to
check for null.

Thanks!

AlpenDitrix:
Travis test rerun required. Can't find out the cause in logs, but in my own fork 2/4 tests failed on testFTPDoesExist.
Succeed after rerun 
https://travis-ci.org/AlpenDitrix/htsjdk/builds/77484050
https://travis-ci.org/samtools/htsjdk/builds/77484059

lbergelson:
Failures probably relate to #276.  We're hitting a live ftp server to test our ftp capabilities.  We need to update tests to use a mock ftp server but it's not been a high priority.

yfarjoun:
@droazen  can you re-review this?

src/tests/java/htsjdk/tribble/index/tabix/TabixIndexTest.java
```diff
@@ -50,17 +59,87 @@ public void readWriteTest(final File tabixFile) throws Exception {
...
                 {SMALL_TABIX_FILE},
                 {BIGGER_TABIX_FILE}
         };
     }
 
+    @Test
+    public void testQueryProvidedItemsAmount() {
+        final String VCF = "testdata/htsjdk/tribble/tabix/YRI.trio.2010_07.indel.sites.vcf";
+        // Note that we store only compressed files
+        File plainTextVcfInputFile = new File(VCF);
```
nh13: finals here and below.

src/java/htsjdk/tribble/index/tabix/TabixIndex.java
```diff
@@ -149,11 +149,15 @@ private TabixIndex(final InputStream inputStream, final boolean closeInputStream
...
-        final List<Block> ret = new ArrayList<Block>(chunks.size());
-        for (final Chunk chunk : chunks) {
-            ret.add(new Block(chunk.getChunkStart(), chunk.getChunkEnd() - chunk.getChunkStart()));
+        if (chunks == null) {
+            return Collections.emptyList();
+        } else {
+            final List<Block> ret = new ArrayList<Block>(chunks.size());
+            for (final Chunk chunk : chunks) {
+                ret.add(new Block(chunk.getChunkStart(), chunk.getChunkEnd() - chunk.getChunkStart()));
+            }
```
nh13: your call if you want to convert this to a stream:
```
chunks.stream().map (chunk -> ...)```

**Comment by vadimzalunin** [comment]
I think it looks good. The contract does as i understand is that the method MAY return null if nothing found, so caller must assume null and empty collection are equally acceptable.

**Comment by droazen** [comment]
:+1: the latest version of this looks good, and can be merged after `2.2.1` is out and the remaining comments are addressed

File: src/tests/java/htsjdk/tribble/index/tabix/TabixIndexTest.java [hunk 3]
```diff
@@ -24,13 +24,22 @@
...
+import htsjdk.variant.variantcontext.writer.VariantContextWriterBuilder;
+import htsjdk.variant.vcf.VCFCodec;
+import htsjdk.variant.vcf.VCFFileReader;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
 import java.io.File;
+import java.util.Iterator;
+
+import static org.testng.Assert.assertEquals;
```
yfarjoun: please remove static imports

AlpenDitrix:
Hi, @droazen.
I've added changes regarding to review comments and squashed everything into one commit.

PR #317, opened 2015-09-08
Add SAMRecord deepCopy() method and tests.

cmnbroad:
For Hellbender issue #623.

### DIFF
--- a/src/java/htsjdk/samtools/SAMRecord.java
+++ b/src/java/htsjdk/samtools/SAMRecord.java
@@ -1832,6 +1832,53 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
         return newRecord;
     }
 
+    /**
+     * Returns a deep copy of the SAM record, with the following exceptions:
+     *
+     *  - The header field, which shares the reference with the original record
+     *  - The file source field, which will always always be set to null in the copy
+     *
+     *  Note that some fields, i.e. the cigar elements, alignment blocks, and
+     *  indexing bin, are not explicitly populated in the copy since they are lazily
+     *  generated on demand.
+     *
+     *  Also note that this fails:
+     *
+     *     original.deepCopy().equals(original)
+     *
+     *  due to the fact that SAMBinaryTagAndValue.equals winds up calling object.equals on the
+     *  value field, which uses reference equality.
+     *
+     */
+    public SAMRecord deepCopy() {
+        final SAMRecord newSAM = new SAMRecord(getHeader());
+
+        newSAM.setReadName(getReadName());
+        newSAM.setReadBases(Arrays.copyOf(getReadBases(), getReadLength()));
+        final byte baseQualities[] = getBaseQualities();
+        newSAM.setBaseQualities(Arrays.copyOf(baseQualities, baseQualities.length));
+        newSAM.setReferenceName(getReferenceName());
+        newSAM.setAlignmentStart(getAlignmentStart()); // clears mAlignmentEnd
+        newSAM.setMappingQuality(getMappingQuality());
+        newSAM.setCigarString(getCigarString()); // clears Cigar element and alignmentBlocks
+        newSAM.setFileSource(null);
+
+        newSAM.setFlags(getFlags());
+        newSAM.setMateReferenceName(getMateReferenceName());
+        newSAM.setMateAlignmentStart(getMateAlignmentStart());
+        newSAM.setInferredInsertSize(getInferredInsertSize());
+        newSAM.setReferenceIndex(getReferenceIndex());
+        newSAM.setMateReferenceIndex(getMateReferenceIndex());
+        newSAM.setValidationStringency(getValidationStringency());
+
+        SAMBinaryTagAndValue attributes = getBinaryAttributes();
+        if (null != attributes) {
+            newSAM.setAttributes(attributes.deepCopy());
+        }
+
+        return newSAM;
+    }
+
     /** Simple toString() that gives a little bit of useful info about the read. */
     @Override
     public String toString() {
--- a/src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
+++ b/src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
@@ -30,6 +30,7 @@ import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
 import java.io.*;
+import java.util.Arrays;
 
 public class SAMRecordUnitTest {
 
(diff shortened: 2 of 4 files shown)


src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1832,6 +1832,87 @@ public Object clone() throws CloneNotSupportedException {
...
+     */
+    public SAMRecord deepCopy() {
+        final SAMRecord newSAM = new SAMRecord(getHeader());
+
+        newSAM.setReadName(getReadName());
+        newSAM.setReadBases(Arrays.copyOf(getReadBases(), getReadLength()));
+        final byte baseQualities[] = getBaseQualities();
+        newSAM.setBaseQualities(Arrays.copyOf(baseQualities, baseQualities.length));
+        newSAM.setReferenceName(getReferenceName());
+        newSAM.setAlignmentStart(getAlignmentStart()); // sets mAlignmentEnd
```
droazen: More accurate to say that this clears `mAlignmentEnd`

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1832,6 +1832,87 @@ public Object clone() throws CloneNotSupportedException {
...
+        newSAM.setMappingQuality(getMappingQuality());
+        newSAM.setCigarString(getCigarString()); // clears Cigar element and alignmentBlocks
+        newSAM.setFileSource(null);
+
+        newSAM.setFlags(getFlags());
+        newSAM.setMateReferenceName(getMateReferenceName());
+        newSAM.setMateAlignmentStart(getMateAlignmentStart());
+        newSAM.setInferredInsertSize(getInferredInsertSize());
+        newSAM.setReferenceIndex(getReferenceIndex());
+        newSAM.setMateReferenceName(getMateReferenceName());
```
droazen: `setMateReferenceName()` is called twice here

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1832,6 +1832,87 @@ public Object clone() throws CloneNotSupportedException {
...
+                        newSAM.setAttribute(tag, ((short[]) value).clone());
+                    }
+                }
+                else if (value instanceof int[]) {
+                    if (isUnsignedArrayAttribute(tag)) {
+                        newSAM.setUnsignedArrayAttribute(tag, ((int[]) value).clone());
+                    }
+                    else {
+                        newSAM.setAttribute(tag, ((int[]) value).clone());
+                    }
```
droazen: I think all this branching based on whether the attribute is an unsigned array attribute is unnecessary -- all you get by going through `setUnsignedArrayAttribute()` is a check that the object is an array (which we already know) and that its length is non-zero (which we can assume, since we should assume that the existing `SAMRecord` is valid). I think you can just call directly into `setAttribute(final short tag, final Object value, final boolean isUnsignedArray)`, passing in `isUnsignedArrayAttribute(tag)` for the third parameter.

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1832,6 +1832,87 @@ public Object clone() throws CloneNotSupportedException {
...
+                }
+                else if (value instanceof float[]) {
+                    newSAM.setAttribute(tag, ((float[]) value).clone());
+                }
+                else {
+                    // otherwise, the api limits the remaining possible value types to
+                    // immutable (String or boxed primitive) types
+                    newSAM.setAttribute(tag, value);
+                }
+            attributes = attributes.getNext();
```
droazen: Just noticed that `SAMBinaryTagAndValue` already has a `copy()` method that claims to perform a deep copy, but doesn't. I recommend that you update the docs for `SAMBinaryTagAndValue.copy()` to state that it performs a shallow copy, move the code above into a new `SAMBinaryTagAndValue.deepCopy()` method, and call that from `SAMRecord.deepCopy()`

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -94,4 +95,107 @@ public void testOffsetAtRead(String cigar, int expectedReferencePos, int posInRe
...
+    }
+
+    @Test(dataProvider = "offsetAtReferenceData")
+    public void testDeepCopyOffset(String cigar, int posInReference, int expectedPosInRead, boolean returnLastBaseIfDeleted) {
+        SAMRecord sam = new SAMRecordSetBuilder().addFrag("test", 0, 1, false, false, cigar, null, 2);
+        testDeepCopy(sam);
+    }
+
+    @Test(dataProvider = "serializationTestData")
+    public void testDeepCopySerialization( final File inputFile ) throws Exception {
```
droazen: This test case seems mis-namedm as it's not testing serialization -- shouldn't it just be called testDeepCopyAttributes? (rather than calling a method of that name).

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -94,4 +95,107 @@ public void testOffsetAtRead(String cigar, int expectedReferencePos, int posInRe
...
+    public void testDeepCopyRef(String cigar, int expectedReferencePos, int posInRead) {
+
+        SAMRecord sam = new SAMRecordSetBuilder().addFrag("test", 0, 1, false, false, cigar, null, 2);
+        testDeepCopy(sam);
+    }
+
+    @Test(dataProvider = "offsetAtReferenceData")
+    public void testDeepCopyOffset(String cigar, int posInReference, int expectedPosInRead, boolean returnLastBaseIfDeleted) {
+        SAMRecord sam = new SAMRecordSetBuilder().addFrag("test", 0, 1, false, false, cigar, null, 2);
+        testDeepCopy(sam);
```
droazen: Rather than re-using the existing `DataProviders` and ignoring most of the parameters you get, recommend creating your own `DataProvider` that emits just what you want (ie., a `SAMRecord` or a bam file). Then re-name these test cases to better reflect what they are actually testing.

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -94,4 +95,107 @@ public void testOffsetAtRead(String cigar, int expectedReferencePos, int posInRe
...
+    private SAMRecord testDeepCopy(SAMRecord sam) {
+        SAMRecord deepCopy = sam.deepCopy();
+
+        // force the indexing bins to be computed in order to satisfy equality test
+        sam.setIndexingBin(sam.computeIndexingBin());
+        deepCopy.setIndexingBin(deepCopy.computeIndexingBin());
+        Assert.assertTrue(sam.equals(deepCopy));
+
+        return deepCopy;
+    }
```
droazen: Add a test case that deep-copies a `SAMRecord`, then mutates some non-attributes (ideally both a reference and non-reference field -- eg., a String field and a primitive field), and shows that the original record is unchanged.

droazen:
Review complete -- back to @cmnbroad

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -94,4 +95,214 @@ public void testOffsetAtRead(String cigar, int expectedReferencePos, int posInRe
...
+
+        final SAMRecord deepCopy = testDeepCopy(sam);
+        Assert.assertTrue(Arrays.equals(sam.getBaseQualities(), deepCopy.getBaseQualities()));
+        Assert.assertTrue(sam.getAlignmentStart() == deepCopy.getAlignmentStart());
+
+        // mutate copy and make sure original remains unchanged
+        final byte[] copyBaseQuals = deepCopy.getBaseQualities();
+        for (int i = 0; i < copyBaseQuals.length; i++) {
+            copyBaseQuals[i]++;
+        }
```
droazen: Call `setBaseQualities()` here for good measure instead of just modifying the returned array in-place, just in case `getBaseQualities()` is ever changed to perform a defensive copy (as such a change would silently render this test ineffective).

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -94,4 +95,214 @@ public void testOffsetAtRead(String cigar, int expectedReferencePos, int posInRe
...
+    public Object [][] deepCopyTestData() {
+        return new Object[][]{
+                {"3S9M"},
+                {"4M1I6M"},
+        };
+    }
+
+    @Test(dataProvider = "deepCopyTestData")
+    public void testDeepCopyRef(String cigar) {
+        SAMRecord sam = new SAMRecordSetBuilder().addFrag("test", 0, 1, false, false, cigar, null, 2);
```
droazen: Why not just have the `deepCopyTestData` `DataProvider` build and return these `SAMRecords`, since it appears that each test is creating them in identical fashion from the `Cigar`s

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -94,4 +95,214 @@ public void testOffsetAtRead(String cigar, int expectedReferencePos, int posInRe
...
+        final byte samBytes[] = sam.getByteArrayAttribute("BY");
+        final byte copyBytes[] = deepCopy.getByteArrayAttribute("BY");
+        Assert.assertFalse(copyBytes == samBytes);
+        Assert.assertTrue(Arrays.equals(copyBytes, samBytes));
+
+        // validate mutation independence
+        final byte testByte = -1;
+        Assert.assertTrue(samBytes[2] != testByte);  // ensure initial test condition
+        Assert.assertTrue(copyBytes[2] != testByte); // ensure initial test condition
+        samBytes[2] = testByte;                      // mutate original
```
droazen: Call `sam.setAttribute("BY")` here for good measure (again, just in case `SAMRecord` is ever changed to perform defensive copying of reference return values).

droazen:
:+1: Just a few very minor comments left to address, then this is good to go. Ping me once it's rebased and squashed.

droazen:
:+1: Looks good -- merging!

PR #318, opened 2015-09-08
Fix typo to get htsjdk version the jar

vdanjean:
The Implementation-Version value in the MANIFEST file of htsjdk JAR file was using a wrong (non existent) property, and so was not even expended in the MANIFEST file.


vdanjean:
I do not understand the CI failure. More exactly, I cannot imagine any link between my patch and the CI failure. It is possible to restart the travis-ci check ?

lbergelson:
I restarted it.

droazen:
Tests have passed -- looks good, merging.

PR #319, opened 2015-09-09
Yf more vc filters

yfarjoun:


src/java/htsjdk/variant/variantcontext/filter/CompoundFilter.java
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public CompoundFilter(final boolean requireAll) {
+        super();
+        this.requireAll = requireAll;
+    }
+
+    /**
+     * @param variantContext the record to examine against the sub-filters
+     * @return true if variantContext either passes all the filters (when requireAll==true)
+     * or doesn't fail any of the filters (when requireAll==false)
```
kbergin: I think the wording for the @return is off.. saying that true='passes all' and false='doesn't fail any' is the same thing because of the double negative.

src/tests/java/htsjdk/variant/variantcontext/filter/CompoundFilterTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    static Allele refA = Allele.create("A", true);
+    static Allele G = Allele.create("G", false);
+
+    static VariantContext vc = new VariantContextBuilder("dummy", "chr1", 1, 1, Arrays.asList(refA, G)).make();
+
+    @DataProvider
+    Iterator<Object[]> testCompoundFilterProvider() {
+        final List<Object[]> filters = new ArrayList<Object[]>(10);
+
```
kbergin: Just to make the data provider a little more clear, I think it would be helpful to change the //all pass to a //requireAll=true or a //Passes all filters and then a //requireAll=false or similar below, only because it was initially confusing that there were two sections with the //one fail and //empty comments.

kbergin:
Now that the CI build has passed... Minor comments on documentation that may warrant consideration. But everything else looks good :+1:

lindenb:
minor suggestion: for VariantContextFilter : shouldn't you use/extend the signature of https://docs.oracle.com/javase/8/docs/api/java/util/function/Predicate.html ? 

```java
boolean 	test(T t) 
```

yfarjoun:
That's a nice idea! thanks I'll look into it.

On Wed, Sep 9, 2015 at 10:23 AM, Pierre Lindenbaum <notifications@github.com

yfarjoun:
I can't extend Predicate until htsjdk is in java 8, but I can use 'test'
and when java 8 comes to HTS-JDK we will be able to make it implement
Predicate easily.

Sounds good?

lindenb:
:+1:

PR #321, opened 2015-09-10
Deleting temporary files and directories in tests

yfarjoun:


src/tests/java/htsjdk/samtools/BAMIndexWriterTest.java
```diff
@@ -50,11 +50,14 @@
     public void testWriteText() throws Exception {
         // Compare the text form of the c-generated bai file and a java-generated one
         final File cBaiTxtFile = File.createTempFile("cBai.", ".bai.txt");
+        cBaiTxtFile.deleteOnExit();
```
droazen: Can you create a utility method `createTempFile()` that calls both `File.createTempFile()` and `deleteOnExit()`, and call that method from tests instead of `File.createTempFile()` directly?

droazen:
Back to @yfarjoun

PR #324, opened 2015-09-15
Added support for IUPAC ambiguous codes to bases match comparison method

gbggrant:


eitanbanks:
@yfarjoun grabbing this one since I'm involved in the project.

src/java/htsjdk/samtools/util/SequenceUtil.java
```diff
@@ -48,6 +52,26 @@
...
+        iupacBaseMappings.put((byte) 'R', new HashSet<Byte>(Arrays.asList(A, G)));
+        iupacBaseMappings.put((byte) 'W', new HashSet<Byte>(Arrays.asList(A, T)));
+        iupacBaseMappings.put((byte) 'S', new HashSet<Byte>(Arrays.asList(C, G)));
+        iupacBaseMappings.put((byte) 'Y', new HashSet<Byte>(Arrays.asList(C, T)));
+        iupacBaseMappings.put((byte) 'K', new HashSet<Byte>(Arrays.asList(G, T)));
+        iupacBaseMappings.put((byte) 'V', new HashSet<Byte>(Arrays.asList(A, C, G)));
+        iupacBaseMappings.put((byte) 'H', new HashSet<Byte>(Arrays.asList(A, C, T)));
+        iupacBaseMappings.put((byte) 'D', new HashSet<Byte>(Arrays.asList(A, G, T)));
+        iupacBaseMappings.put((byte) 'B', new HashSet<Byte>(Arrays.asList(C, G, T)));
+        iupacBaseMappings.put((byte) 'N', new HashSet<Byte>(Arrays.asList(A, C, G, T)));
```
eitanbanks: Don't we also want N to match N?

src/java/htsjdk/samtools/util/SequenceUtil.java
```diff
@@ -62,9 +86,35 @@ public static String reverseComplement(final String sequenceData) {
...
+     */
     public static boolean basesEqual(byte lhs, byte rhs) {
         if (lhs == rhs) return true;
+
+        // Convert to UPPER case.
+        if (lhs > 90) lhs -= 32;
+        if (rhs > 90) rhs -= 32;
+        if ((isValidBase(lhs, VALID_BASES_WITH_N_UPPER) && (isValidBase(rhs, VALID_BASES_WITH_N_UPPER)))) {
+            // Both lhs and rhs are [A, C, G, T, N] - NOT iupac ambiguity codes
+            // TODO - is this ever called with '.' rather than 'N'??
```
eitanbanks: How could it be "."?  That's for alleles, not bases.

tfenne:
@gbggrant Did you do any performance testing of this?  It's obviously going to be slower than the old (trivial) implementation.  At a minimum, I think you'd see some benefit from replacing the Map<Byte,Set<Byte>>  with a Map<Byte,Byte[]> since doing a linear scan of a small byte[] will probably be quite a bit faster than checking through a hash table.

src/java/htsjdk/samtools/util/SequenceUtil.java
```diff
@@ -62,9 +86,35 @@ public static String reverseComplement(final String sequenceData) {
...
+            return lhs == rhs;
+        }
+        // If here, one of the two bases is IUPAC ambiguous
+        final Set<Byte> rightSide = iupacBaseMappings.get(rhs);
+        for (byte b : iupacBaseMappings.get(lhs)) {
```
eitanbanks: byte b could be 'final'

src/java/htsjdk/samtools/util/SequenceUtil.java
tfenne: FWIW some illumina fastq files use/used "." instead of "N" for nocall.

eitanbanks:
@tfenne I agree that HashMaps are inefficient and we should change the representation here, but I'm not exactly sure what you were suggesting.  Something like this?
'  static private final byte[][] iupacBaseMappings = new byte[256][];
   iupacBaseMappings['M'] = {A, C};
   etc.
'

tfenne:
@eitanbanks I was suggesting a half-way solution that didn't eliminate the outer map, but replaced the inner Set<Byte> with byte[]s instead.

I think the *best* solution here is more complicated, and would involve bitwise comparisons.  You could define something like the following pseudo-code:

    byte A = 1, C = 2, G = 4, T = 8;
    byte M = A | C;

The runtime comparison could then be done as simply:

    A & M > 0

Then, yes, I think it'd be tempted to create a byte[127] that contained the was filled with 0s for invalid bases and the appropriate bitmask for all valid bases, such that the ultimate comparison would be:

    bases[(int) lhs] & bases[(int) rhs] > 0

src/java/htsjdk/samtools/util/SequenceUtil.java
gbggrant: @tfenne I'll switch to the bit mask representation.  I had ordered the code so that it wouldn't hit the hashmap usage unless it falls through the older handling, but agree that's not enough.

eitanbanks: Actually, it probably is enough @gbggrant.  There are < 100 IUPAC bases in the standard human reference, so we're talking about .000001% of bases that will drop down into the HashMap code.  It's probably fine to show that the performance increase is negligible with a few WGS bams.

tfenne: Not all sequence is human ;)

I didn't go all the way to creating a PR, but here's a gist that gives an implementation using bitwise comparisons: https://gist.github.com/tfenne/cc49c01e663a7bae4957

@gbggrant I grabbed your expanded tests and ran them using this class, and modifying SequenceUtil.basesEqual() to look like this:

    public static boolean basesEqual(final byte lhs, final byte rhs) {
        if (lhs == rhs) return true;
        else return BaseUtil.basesEqual(lhs, rhs);
    }

All the tests pass.  It is, of course, up to you whether to incorporate this or not.

gbggrant:
@eitanbanks @tfenne I have rewritten the code to use Tim's suggested bit mask mapping.  Please re-review.

tfenne:
@gbggrant Looks good to me!

eitanbanks:
If Tim is happy then so am I.  :)

gbggrant:
fyi - this broke unit tests in Picard.  I have reverted the merge and am looking into the problem.

PR #327, opened 2015-09-23
Improve handling of null SAMFileHeaders in alignment records.

cmnbroad:
- This is the first pass of improving null SAMFileHeader support. There are some TODOs inline with questions.
- Once we have agreement on those, the next pass will be to find all the places that call getReferenceIndex/getReferenceName and the mate analogs internally and make sure they do something reasonable with the new sentinel values.
- The existing doc seems to already indicate that the header does not have to be set. We may want to add more detail there, but I've updated the doc for the individual methods, and will clean those up
again once we have resolution on the rest of the contract (ie null value handling, etc.)
- I chose to return a MISSING_HEADER validation error when validating records. Not sure if thats what  we want.
- The existing implementation silently allows the caller to set an invalid reference name (one that doesn't exist in the sequence dictionary), but throws if you try to set an invalid reference index. I've retained that behavior but am wondering if its deliberate/correct.
- I don't see a code path that requires any changes to BAMRecord. Likeiwse there are numerous code paths that assume non null headers that I didn't touch because the records all appear to originate from files and have valid headers (i.e., SAMFileValidator).
- The existing doc says get/setReferenceIndex() is preferred to get/setReferenceName(). Should we change that ?

### DIFF
--- a/src/java/htsjdk/samtools/DefaultSAMRecordFactory.java
+++ b/src/java/htsjdk/samtools/DefaultSAMRecordFactory.java
@@ -18,7 +18,11 @@ public class DefaultSAMRecordFactory implements SAMRecordFactory {
         return new SAMRecord(header);
     }
 
-    /** Create a new BAM Record. */
+    /**
+     * Create a new BAM Record. If the reference sequence index or mate reference sequence index are
+     * any value other than NO_ALIGNMENT_REFERENCE_INDEX, the values must be resolvable against the sequence
+     * dictionary in the header argument.
+     */
     public BAMRecord createBAMRecord (final SAMFileHeader header,
                                       final int referenceSequenceIndex,
                                       final int alignmentStart,
--- a/src/java/htsjdk/samtools/MergingSamRecordIterator.java
+++ b/src/java/htsjdk/samtools/MergingSamRecordIterator.java
@@ -124,6 +124,7 @@ public class MergingSamRecordIterator implements CloseableIterator<SAMRecord> {
         final ComparableSamRecordIterator iterator = this.pq.poll();
         final SAMRecord record = iterator.next();
         addIfNotEmpty(iterator);
+        // this will resolve the reference indices against the new, merged header
         record.setHeader(this.samHeaderMerger.getMergedHeader());
 
         // Fix the read group if needs be
@@ -141,17 +142,6 @@ public class MergingSamRecordIterator implements CloseableIterator<SAMRecord> {
             if (oldGroupId != null) {
                 final String newGroupId = this.samHeaderMerger.getProgramGroupId(iterator.getReader().getFileHeader(), oldGroupId);
                 record.setAttribute(ReservedTagConstants.PROGRAM_GROUP_ID, newGroupId);
-            }
-        }
-
-        // Fix up the sequence indexes if needs be
-        if (this.samHeaderMerger.hasMergedSequenceDictionary()) {
-            if (record.getReferenceIndex() != SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX) {
-                record.setReferenceIndex(this.samHeaderMerger.getMergedSequenceIndex(iterator.getReader().getFileHeader(), record.getReferenceIndex()));
-            }
-
-            if (record.getReadPairedFlag() && record.getMateReferenceIndex() != SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX) {
-                record.setMateReferenceIndex(this.samHeaderMerger.getMergedSequenceIndex(iterator.getReader().getFileHeader(), record.getMateReferenceIndex()));
             }
         }
 
--- a/src/java/htsjdk/samtools/SAMRecord.java
+++ b/src/java/htsjdk/samtools/SAMRecord.java
@@ -40,48 +40,69 @@ import java.util.Set;
 
 /**
  * Java binding for a SAM file record.  c.f. http://samtools.sourceforge.net/SAM1.pdf
- *
+ * <p>
  * The presence of reference name/reference index and alignment start
  * do not necessarily mean that a read is aligned.  Those values may merely be set to force a SAMRecord
  * to appear in a certain place in the sort order.  The readUnmappedFlag must be checked to determine whether
  * or not a read is mapped.  Only if the readUnmappedFlag is false can the reference name/index and alignment start
  * be interpreted as indicating an actual alignment position.
- *
+ * <p>
  * Likewise, presence of mate reference name/index and mate alignment start do not necessarily mean that the
  * mate is aligned.  These may be set for an unaligned mate if the mate has been forced into a particular place
  * in the sort order per the above paragraph.  Only if the mateUnmappedFlag is false can the mate reference name/index
  * and mate alignment start be interpreted as indicating the actual alignment position of the mate.
- *
+ * <p>
  * Note also that there are a number of getters & setters that are linked, i.e. they present different representations
  * of the same underlying data.  In these cases there is typically a representation that is preferred because it
  * ought to be faster than some other representation.  The following are the preferred representations:
- *
- * getReadNameLength() is preferred to getReadName().length()
- * get/setReadBases() is preferred to get/setReadString()
- * get/setBaseQualities() is preferred to get/setBaseQualityString()
- * get/setReferenceIndex() is preferred to get/setReferenceName()
- * get/setMateReferenceIndex() is preferred to get/setMateReferenceName()
- * getCigarLength() is preferred to getCigar().getNumElements()
- * get/setCigar() is preferred to get/setCigarString()
- *
- * Note that setIndexingBin() need not be called when writing SAMRecords.  It will be computed as necessary.  It is only
- * present as an optimization in the event that the value is already known and need not be computed.
- *
- * setHeader() need not be called when writing SAMRecords.  It may be convenient to call it, however, because
- * get/setReferenceIndex() and get/setMateReferenceIndex() must have access to the SAM header, either as an argument
- * or previously passed to setHeader().
- *
+ * </p><ul>
+ * <li>getReadNameLength() is preferred to getReadName().length()</li>
+ * <li>get/setReadBases() is preferred to get/setReadString()</li>
+ * <li>get/setBaseQualities() is preferred to get/setBaseQualityString()</li>
+ * <li>get/setReferenceIndex() is preferred to get/setReferenceName() for records with valid SAMFileHeaders</li>
+ * <li>get/setMateReferenceIndex() is preferred to get/setMateReferenceName() for records with valid SAMFileHeaders</li>
+ * <li>getCigarLength() is preferred to getCigar().getNumElements()</li>
+ * <li>get/setCigar() is preferred to get/setCigarString()</li>
+ * </ul>
+ * <p>
  * setHeader() is called by the SAM reading code, so the get/setReferenceIndex() and get/setMateReferenceIndex()
- * methods will have access to the sequence dictionary.
- *
+ * methods will have access to the sequence dictionary to resolve reference and mate reference names to dictionary
+ * indices.
+ * <p>
+ * setHeader() need not be called explicitly when writing SAMRecords, however the writers require a record
+ * in order to call get/setReferenceIndex() and get/setMateReferenceIndex(). Therefore adding records to a writer
+ * has a side effect: any record that does not have an assigned header at the time it is added to a writer will be
+ * updated and assigned the header associated with the writer.
+ * <p>
  * Some of the get() methods return values that are mutable, due to the limitations of Java.  A caller should
  * never change the value returned by a get() method.  If you want to change the value of some attribute of a
  * SAMRecord, create a new value object and call the appropriate set() method.
- *
+ * </p>
+ * Note that setIndexingBin() need not be called when writing SAMRecords.  It will be computed as necessary.  It is only
+ * present as an optimization in the event that the value is already known and need not be computed.
+ * <p>
  * By default, extensive validation of SAMRecords is done when they are read.  Very limited validation is done when
  * values are set onto SAMRecords.
- */
-/**
+ * <p>
+ * <h3>Notes on Headerless SAMRecords</h3>
+ * <p>
+ * If the header is null, the following SAMRecord methods may throw exceptions:
+ * <ul>
+ * <li>getReferenceIndex</li>
+ * <li>setReferenceIndex</li>
+ * <li>getMateReferenceIndex</li>
+ * <li>setMateReferenceIndex</li>
+ * </ul><p>
+ * Record comparators (i.e. SAMRecordCoordinateComparator and SAMRecordDuplicateComparator) require records with
+ * non-null header values.
+ * <p>
+ * A record with null a header may be validated by the isValid method, but the reference and mate reference indices,
+ * read group, sequence dictionary, and alignment start will not be fully validated unless a header is present.
+ * <p>
+ * Also, SAMTextWriter, BAMFileWriter, and CRAMFileWriter all require records to have a valid header in order to be
+ * written. Any record that does not have a header at the time it is added to the writer will be updated to use the
+ * header associated with the writer.
+ * <p>
  * @author alecw@broadinstitute.org
  * @author mishali.naik@intel.com
  */
@@ -291,8 +312,8 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
     }
 
     private static boolean hasReferenceName(final Integer referenceIndex, final String referenceName) {
-        return (referenceIndex != null && referenceIndex != NO_ALIGNMENT_REFERENCE_INDEX) ||
-                !NO_ALIGNMENT_REFERENCE_NAME.equals(referenceName);
+        return (referenceIndex != null && !referenceIndex.equals(NO_ALIGNMENT_REFERENCE_INDEX)) ||
+                (!NO_ALIGNMENT_REFERENCE_NAME.equals(referenceName));
     }
 
     /**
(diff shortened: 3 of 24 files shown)


droazen:
We discussed this PR in person and agreed on a set of changes to be made before an in-depth review.

cmnbroad:
Summary of behavior with these changes:

SAMRecord:
-setHeader(null) clears the reference indices and sets them to null (leaving them unresolved). Setting the header to non-null attempts to re-resolve the reference names against the current dictionary and updates the indices. This can fail silently in the same way that setReferenceName can fail silently in the existing implementation.
-getReferenceName() is always safe (as before).
-setReferenceName(name) is always safe if the name is non-null (as before). If the record has a header, it will attempt to resolve the name to an index, but this may silently fail and leave the index unresolved.
-setReferenceIndex(NO_ALIGNMENT_INDEX) is always safe. Using any other value will throw if the record has no header or if the value can't be found in the dictionary.
-getReferenceIndex() is unsafe if the record has no header, otherwise it is safe, though if the index is
unresolved and the reference name can't be resolved to an index using the current dictionary, the index will be clobbered and reset to NO_ALIGNMENT_INDEX (as before!!)

Other:
-The SAM/BAM/CRAM writers all now have the side effect that they will  reset a record's header when handed a record that doesn't already have one.
-Note: With these changes, it is not possible to set a SAMRecord's header to null and then directly (without going through a writer) encode it to a stream using the BAMRecordCodec, since setting the
header to null clobbers the reference indices. (This will require a change to the Hellbender read serializer test since it tries to do this).

src/java/htsjdk/samtools/DefaultSAMRecordFactory.java
```diff
@@ -18,7 +18,11 @@ public SAMRecord createSAMRecord(final SAMFileHeader header) {
         return new SAMRecord(header);
     }
 
-    /** Create a new BAM Record. */
+    /**
+     * Create a new BAM Record. If the reference sequence index or mate reference sequence index are
+     * any value other than NO_ALIGNMENT_REFERENCE, the values must be resolvable against the sequence
```
droazen: `NO_ALIGNMENT_REFERENCE_INDEX`

src/java/htsjdk/samtools/MergingSamRecordIterator.java
```diff
@@ -144,17 +145,6 @@ public SAMRecord next() {
...
-        // Fix up the sequence indexes if needs be
-        if (this.samHeaderMerger.hasMergedSequenceDictionary()) {
-            if (record.getReferenceIndex() != SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX) {
-                record.setReferenceIndex(this.samHeaderMerger.getMergedSequenceIndex(iterator.getReader().getFileHeader(), record.getReferenceIndex()));
-            }
-
-            if (record.getReadPairedFlag() && record.getMateReferenceIndex() != SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX) {
-                record.setMateReferenceIndex(this.samHeaderMerger.getMergedSequenceIndex(iterator.getReader().getFileHeader(), record.getMateReferenceIndex()));
-            }
-        }
```
droazen: Are we 100% sure that we can delete this block? I know that the header merger uses per-header translation tables mapping from old to new indices -- does the `record.setHeader(this.samHeaderMerger.getMergedHeader());` line above render this unnecessary?

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -305,119 +308,202 @@ private boolean hasMateReferenceName() {
     }
 
     /**
-     * @return Reference name, or null if record has no reference.
+     * @return Reference name, or NO_ALIGNMENT_REFERENCE_NAME (-1) if the record has no reference name
```
droazen: Value of `NO_ALIGNMENT_REFERENCE_NAME` is `"*"`, not `-1`

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -305,119 +308,202 @@ private boolean hasMateReferenceName() {
...
+            // try to resolve the reference index
+            if (NO_ALIGNMENT_REFERENCE_NAME.equals(mReferenceName)) {
                 mReferenceIndex = NO_ALIGNMENT_REFERENCE_INDEX;
-            } else {
+            }
+            else if (null != mHeader) {
                 mReferenceIndex = mHeader.getSequenceIndex(mReferenceName);
             }
+            else {
+                throw new IllegalArgumentException("A non-null SAMFileHeader is required to resolve the reference index");
```
droazen: This should be an `IllegalStateException`, I think (there are no arguments to this method).

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -305,119 +308,202 @@ private boolean hasMateReferenceName() {
...
+            }
         }
         return mReferenceIndex;
     }
 
     /**
-     * @param referenceIndex Must either equal -1 (indicating no reference), or exist in the sequence dictionary
-     * in the header associated with this record.
+     * Updates the reference index. The record must have a valid SAMFileHeader, and the reference index must appear in
```
droazen: "The record must have a valid SAMFileHeader, unless the referenceIndex parameter equals NO_ALIGNMENT_REFERENCE_INDEX"

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -305,119 +308,202 @@ private boolean hasMateReferenceName() {
...
+            mReferenceIndex = NO_ALIGNMENT_REFERENCE_INDEX;
             mReferenceName = NO_ALIGNMENT_REFERENCE_NAME;
-        } else {
-            try {
-                mReferenceName = mHeader.getSequence(referenceIndex).getSequenceName();
-            } catch (final NullPointerException e) {
-                throw new IllegalArgumentException("Reference index " + referenceIndex + " not found in sequence dictionary.", e);
+        }
+        else if (null == mHeader) {
+            throw new IllegalArgumentException("A non-null SAMFileHeader is required to resolve the reference index");
```
droazen: `IllegalStateException`

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -305,119 +308,202 @@ private boolean hasMateReferenceName() {
...
-    }
+    public String getReferenceName() { return mReferenceName; }
 
-    public void setReferenceName(final String value) {
-        /* String.intern() is surprisingly expensive, so avoid it by looking up in sequence dictionary if possible */
-        if (NO_ALIGNMENT_REFERENCE_NAME.equals(value)) {
+    /**
+     * Sets the reference name for this record. If the record has a valid SAMFileHeader and the reference
+     * name is present in the associated sequence dictionary, the record's reference index will also be
+     * updated with the corresponding sequence index.
```
droazen: "If `referenceName` is `NO_ALIGNMENT_REFERENCE_NAME`, sets the reference index to `NO_ALIGNMENT_REFERENCE_INDEX`

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -305,119 +308,202 @@ private boolean hasMateReferenceName() {
...
     }
 
     /**
-     * @return index of the reference sequence for this read in the sequence dictionary, or -1
-     * if read has no reference sequence set, or if a String reference name is not found in the sequence index..
+     * Returns the reference index for this record.
+     *
+     * If the reference name for this record has previously been resolved against the sequence dictionary, the corresponding
+     * index is returned directly. Otherwise, the record must have a non-null SAMFileHeader that can be used to
+     * resolve the index for the record's current reference name. If the record has a header, and the name does not appear
```
droazen: unless the reference name is `NO_ALIGNMENT_REFERENCE_NAME`

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -305,119 +308,202 @@ private boolean hasMateReferenceName() {
...
+            }
+            else {
+                throw new IllegalArgumentException("Reference index " + referenceIndex + " not found in sequence dictionary.");
             }
         }
     }
 
     /**
-     * @return Mate reference name, or null if one is not assigned.
+     * @return Reference name, or NO_ALIGNMENT_REFERENCE_NAME (-1) if the record has no reference name
```
droazen: `NO_ALIGNMENT_REFERENCE_NAME ("*")`

droazen: "reference name" -> "mate reference name" (at both the beginning and end of this line)

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -305,119 +308,202 @@ private boolean hasMateReferenceName() {
...
      */
     public String getMateReferenceName() {
         return mMateReferenceName;
     }
 
+    /**
+     * Sets the mate reference name for this record. If the record has a valid SAMFileHeader and the mate reference
+     * name is present in the associated sequence dictionary, the record's mate reference index will also be
+     * updated with the corresponding sequence index.
```
droazen: "If mateReferenceName is NO_ALIGNMENT_REFERENCE_NAME, sets the mate reference index to NO_ALIGNMENT_REFERENCE_INDEX"

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -305,119 +308,202 @@ private boolean hasMateReferenceName() {
...
     }
 
     /**
-     * @return index of the reference sequence for this read's mate in the sequence dictionary, or -1
-     * if mate has no reference sequence set.
+     * Returns the mate reference index for this record.
+     *
+     * If the mate reference name for this record has previously been resolved against the sequence dictionary, the
+     * corresponding index is returned directly. Otherwise, the record must have a non-null SAMFileHeader that can be
+     * used to resolve the index for the record's current mate reference name. If the record has a header, and the
```
droazen: unless the mate reference name is NO_ALIGNMENT_REFERENCE_NAME

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -305,119 +308,202 @@ private boolean hasMateReferenceName() {
...
+            // try to resolve the reference index
+            if (NO_ALIGNMENT_REFERENCE_NAME.equals(mMateReferenceName)) {
                 mMateReferenceIndex = NO_ALIGNMENT_REFERENCE_INDEX;
-            } else {
+            }
+            else if (null != mHeader) {
                 mMateReferenceIndex = mHeader.getSequenceIndex(mMateReferenceName);
             }
+            else {
+                throw new IllegalArgumentException("A non-null SAMFileHeader is required to resolve the mate reference index");
```
droazen: `IllegalStateException`

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -305,119 +308,202 @@ private boolean hasMateReferenceName() {
...
     }
 
     /**
-     * @param referenceIndex Must either equal -1 (indicating no reference), or exist in the sequence dictionary
-     * in the header associated with this record.
-     */
-    public void setMateReferenceIndex(final int referenceIndex) {
-        mMateReferenceIndex = referenceIndex;
-        if (mMateReferenceIndex == NO_ALIGNMENT_REFERENCE_INDEX) {
+     * Updates the mate reference index. The record must have a valid SAMFileHeader, and the mate reference index must appear in
```
droazen: "The record must have a valid SAMFileHeader, unless the mateReferenceIndex parameter equals NO_ALIGNMENT_REFERENCE_INDEX"

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -305,119 +308,202 @@ private boolean hasMateReferenceName() {
...
+            mMateReferenceIndex = NO_ALIGNMENT_REFERENCE_INDEX;
             mMateReferenceName = NO_ALIGNMENT_REFERENCE_NAME;
-        } else {
-            try {
-                mMateReferenceName = mHeader.getSequence(referenceIndex).getSequenceName();
-            } catch (final NullPointerException e) {
-                throw new IllegalArgumentException("Reference index " + referenceIndex + " not found in sequence dictionary.", e);
+        }
+        else if (null == mHeader) {
+            throw new IllegalArgumentException("A non-null SAMFileHeader is required to resolve the mate reference index");
```
droazen: `IllegalStateException`

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -684,15 +770,15 @@ protected void initializeCigar(final Cigar cigar) {
      * @return The SAMReadGroupRecord from the SAMFileHeader for this SAMRecord, or null if
      * 1) this record has no RG tag, or 2) the header doesn't contain the read group with
      * the given ID.
```
droazen: "or 3) this record has no SAMFileHeader"

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1337,16 +1423,70 @@ int computeIndexingBin() {
         return GenomicIndexUtil.reg2bin(alignmentStart, alignmentEnd);
     }
 
+    /**
+     *
+     * @return the SAMFileHeader for this record. If the header is null, the following SAMRecord methods may throw
+     * IllegalArgumentExceptions:
```
droazen: `IllegalArgumentExceptions" -> "exceptions" (since I asked you to change several exception classes above)

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1337,16 +1423,70 @@ int computeIndexingBin() {
...
+     *
+     *   getReferenceIndex
+     *   setReferenceIndex
+     *   getMateReferenceIndex
+     *   setMateReferenceIndex
+     *
+     * A SAMRecord with a null header can not be fully validated and will result in SAMValidationErrors if isValid is
+     * called.
```
droazen: Can you copy these warnings on `isValid()` and methods that can throw exceptions into a new paragraph on "Notes on headerless SAMRecords" in the class-level comment at the top of `SAMRecord`?

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1337,16 +1423,70 @@ int computeIndexingBin() {
...
     /**
-     * Setting header into SAMRecord facilitates conversion btw reference sequence names and indices
+     * Sets the SAMFileHeader for this record. Setting the header into SAMRecord facilitates conversion between reference
+     * sequence names and indices.
+     *
+     * If the record has a reference or mate reference name, the corresponding reference and mate reference indices are
+     * resolved and updated using the sequence dictionary in the new header.
+     *
+     * When the SAMFileHeader is set to null, the reference and mate reference indices are cleared. Therefore, calls to
+     * the following SAMRecord methods on records with a null header may throw IllegalArgumentExceptions:
```
droazen: "IllegalArgumentExceptions" -> "exceptions"

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1337,16 +1423,70 @@ int computeIndexingBin() {
...
     public void setHeader(final SAMFileHeader header) {
         this.mHeader = header;
+        if (null == header) {
+            // if the reference and mate reference index values have previously been resolved, set them to unresolved
+            if (null != mReferenceIndex && !mReferenceIndex.equals(NO_ALIGNMENT_REFERENCE_INDEX)) {
+                mReferenceIndex = null;
+            }
+            if (null != mMateReferenceIndex && !mMateReferenceIndex.equals(NO_ALIGNMENT_REFERENCE_INDEX)) {
+                mMateReferenceIndex = null;
+            }
```
droazen: Is there a reason you're not clearing the indices if they're equal to `NO_ALIGNMENT_REFERENCE_INDEX`? I think you should be unconditionally nulling out the indices here for clarity's sake. If you null them out when they are equal to  `NO_ALIGNMENT_REFERENCE_INDEX`, and the corresponding reference name is truly equal to `NO_ALIGNMENT_REFERENCE_NAME`, the index will get reset to `NO_ALIGNMENT_REFERENCE_INDEX` anyway on the next access (even in the headerless case).

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1337,16 +1423,70 @@ int computeIndexingBin() {
...
+        }
+        else {
+            // attempt to resolve the existing reference names and indices against the new sequence dictionary, but
+            // don't throw if the names don't appear in the dictionary
```
droazen: Make this a prominent warning in the method-level docs (that we won't throw if the existing reference names aren't present in the new header)

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1665,7 +1807,12 @@ public int hashCode() {
             */
                 if (firstOnly) return ret;
             }
-            if (getHeader().getSequenceDictionary().size() == 0) {
+            if (getHeader() == null) {
+                if (ret == null) ret = new ArrayList<SAMValidationError>();
+                ret.add(new SAMValidationError(SAMValidationError.Type.MISSING_HEADER, "A non-null SAMHeader is required to resolve the sequence dictionary for a mapped read.", getReadName()));
+                if (firstOnly) return ret;
+            }
```
droazen: Not sure I like this...the whole point of this PR was to make headerless `SAMRecords` an officially supported, valid state for a read to be in. Yet with this change they will fail `isValid()`. Can't we instead say that headerless `SAMRecords` are valid, but are validated to a less strict standard (ie., some checks that require a header are not performed)?

src/java/htsjdk/samtools/SAMRecordCoordinateComparator.java
```diff
@@ -75,10 +75,16 @@ private int compareInts(int i1, int i2) {
      * Less stringent compare method than the regular compare.  If the two records
      * are equal enough that their ordering in a sorted SAM file would be arbitrary,
      * this method returns 0.  If read is paired and unmapped, use the mate mapping to sort.
+     * Record being compared must have non-null SAMFileHedaers.
```
droazen: Record -> Records

droazen: SAMFileHedaers -> SAMFileHeaders

src/java/htsjdk/samtools/SAMRecordCoordinateComparator.java
```diff
@@ -75,10 +75,16 @@ private int compareInts(int i1, int i2) {
...
      *
      * @return negative if samRecord1 < samRecord2,  0 if equal, else positive
      */
     public int fileOrderCompare(final SAMRecord samRecord1, final SAMRecord samRecord2) {
+
+        if (null == samRecord1.getHeader() || null == samRecord2.getHeader()) {
+            throw new IllegalArgumentException("Record must have non-null SAMFileHeaders to be comapred");
```
droazen: Record -> Records
comapred -> compared

src/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
```diff
@@ -200,6 +204,10 @@ private boolean pairedEndAndBothMapped(final SAMRecord record) {
     public int compare(final SAMRecord samRecord1, final SAMRecord samRecord2) {
         int cmp;
 
+        if (null == samRecord1.getHeader() || null == samRecord2.getHeader()) {
+            throw new IllegalArgumentException("Record must have non-null SAMFileHeaders to be comapred");
```
droazen: Record -> Records
comapred -> compared

droazen: Also this check is redundant, since the first thing `compare()` does is to call `fileOrderCompare()`

src/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
```diff
@@ -236,6 +244,10 @@ public int compare(final SAMRecord samRecord1, final SAMRecord samRecord2) {
     private int fileOrderCompare(final SAMRecord samRecord1, final SAMRecord samRecord2, final boolean collapseOrientation, final boolean considerNumberOfEndsMappedAndPairing) {
         int cmp;
 
+        if (null == samRecord1.getHeader() || null == samRecord2.getHeader()) {
+            throw new IllegalArgumentException("Record must have non-null SAMFileHeaders to be comapred");
```
droazen: Record -> Records
comapred -> compared

src/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -894,14 +900,22 @@ public static int getMateUnclippedEnd(final SAMRecord rec) {
...
-                    if (ret == null) ret = new ArrayList<SAMValidationError>();
-                    ret.add(new SAMValidationError(SAMValidationError.Type.CIGAR_MAPS_OFF_REFERENCE,
-                            cigarTypeName + " M operator maps off end of reference", rec.getReadName(), recordNumber));
-                    break;
+            SAMFileHeader samHeader = rec.getHeader();
+            if (null == samHeader) {
+                if (ret == null) ret = new ArrayList<SAMValidationError>();
+                ret.add(new SAMValidationError(SAMValidationError.Type.MISSING_HEADER,
+                        cigarTypeName + " A non-null SAMHeader is required to validate cigar elements for: ", rec.getReadName(), recordNumber));
+            }
```
droazen: Here too, I think a missing header should result in skipped validation steps rather than a validation failure.

src/java/htsjdk/samtools/cram/build/Sam2CramRecordFactory.java
```diff
@@ -96,7 +99,17 @@ public Sam2CramRecordFactory(final byte[] refBases, final SAMFileHeader samFileH
         }
     }
 
+    /**
+     * Create a CramCompressionRecord.
+     *
+     * @param record If the input record does not have an associated SAMFileHeader, it will be updated
+     *               with the header used for the writer in order to allow reference indices to be resolved.
```
droazen: This class is not a "writer" class, is it?

src/java/htsjdk/samtools/util/AbstractProgressLogger.java
```diff
@@ -76,7 +76,7 @@ public synchronized boolean record(final String chrom, final int pos) {
      */
     @Override
     public synchronized boolean record(final SAMRecord rec) {
-        if (rec.getReferenceIndex() == SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX) {
+        if (null == rec.getHeader() || rec.getReferenceIndex() == SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX) {
             return record(null, 0);
```
droazen: Recommend changing this method to use the reference name exclusively. Eg.,

    if ( SAMRecord.NO_ALIGNMENT_REFERENCE_NAME.equals(rec.getReferenceName()) ) {
        return record(null, 0);
    }
    else {
        return record(rec.getReferenceName(), rec.getAlignmentStart());
    }

src/tests/java/htsjdk/samtools/BAMFileWriterTest.java
```diff
@@ -114,10 +118,72 @@ public void testPositive(final String testName, final SAMRecordSetBuilder samRec
...
+        samReader.getFileHeader().setSortOrder(bamReader.getFileHeader().getSortOrder());
+        final CloseableIterator<SAMRecord> it = samReader.iterator();
+        final CloseableIterator<SAMRecord> bamIt = bamReader.iterator();
+        while (it.hasNext()) {
+            Assert.assertTrue(bamIt.hasNext());
+            final SAMRecord samRecord = it.next();
+            final SAMRecord bamRecord = bamIt.next();
+
+            // test only reference names since we'll have lost reference indices due to the fake  null header
+            Assert.assertEquals(bamRecord.getReferenceName(), samRecord.getReferenceName());
```
droazen: Check equality of alignment starts here too

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -90,8 +91,283 @@ public void testOffsetAtReference(String cigar, int posInReference, int expected
...
+        final SAMRecord sam = new SAMRecordSetBuilder().addFrag("test", 0, 1, false, false, cigar, null, 2);
+        Assert.assertEquals(sam.getReferencePositionAtReadPosition(posInRead), expectedReferencePos);
+    }
 
-            SAMRecord sam = new SAMRecordSetBuilder().addFrag("test", 0, 1, false, false, cigar, null, 2);
-            Assert.assertEquals(sam.getReferencePositionAtReadPosition(posInRead), expectedReferencePos);
+    @Test
+    public void testReferenceName() {
+        // NO_ALIGNMENT_NAME
+        SAMRecord sam = new SAMRecordSetBuilder().addFrag("test", 0, 1, false, false, "3S9M", null, 2);
```
droazen: Extract this `new SAMRecordSetBuilder().addFrag("test", 0, 1, false, false, "3S9M", null, 2);` call into a helper method, and call it here and everywhere else this line is duplicated.

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -90,8 +91,283 @@ public void testOffsetAtReference(String cigar, int posInReference, int expected
...
+        sam.setReferenceIndex(3);
+        Assert.assertTrue(sam.getReferenceIndex().equals(3));
+        Assert.assertTrue(sam.getReferenceName().equals("chr4"));
+    }
+
+    @Test(expectedExceptions=IllegalArgumentException.class)
+    public void testInvalidReferenceIndex() {
+        // unresolvable reference
+        final SAMRecord sam = new SAMRecordSetBuilder().addFrag("test", 0, 1, false, false, "3S9M", null, 2);
+        sam.setReferenceIndex(9999);
```
droazen: It's interesting that we throw on an invalid reference index, but not an invalid reference name (when there is a header). Perhaps for consistency's sake we should throw on `setReferenceName()` when a header is present and the reference name is not in the dictionary? Thoughts?

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -90,8 +91,283 @@ public void testOffsetAtReference(String cigar, int posInReference, int expected
...
+    public void testNullHeaderReferenceName() {
+        SAMRecord sam = new SAMRecordSetBuilder().addFrag("test", 0, 1, false, false, "3S9M", null, 2);
+        final SAMFileHeader samHeader = sam.getHeader();
+        Assert.assertTrue(null != samHeader);
+        final String originalRefName = sam.getReferenceName();
+        int originalRefIndex = sam.getReferenceIndex();
+
+        sam.setHeader(null);
+        // setting header to null retains the previously assigned ref name
+        Assert.assertTrue(originalRefName.equals(sam.getReferenceName()));
```
droazen: Since `SAMRecord.mReferenceIndex` is `protected`, and you are in the same package, you should also be able to assert that the `setHeader(null)` call cleared the reference index.

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -90,8 +91,283 @@ public void testOffsetAtReference(String cigar, int posInReference, int expected
...
+        Assert.assertTrue(sam.getReferenceIndex().equals(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX));
+
+        // null header allows any ref name set
+        final String unresolvableRefName = "unresolvable";
+        sam.setReferenceName(unresolvableRefName);
+        Assert.assertTrue(unresolvableRefName.equals(sam.getReferenceName()));
+    }
+
+    @Test
+    public void testNullHeaderReferenceIndex() {
```
droazen: `testNullHeaderReferenceName` and `testNullHeaderReferenceIndex` are not well decomposed -- they both appear to be testing reference name and index-related functionality in equal amounts. Recommend either merging them or better separating the testing of the reference name and index-related methods.

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -90,8 +91,283 @@ public void testOffsetAtReference(String cigar, int posInReference, int expected
...
+        Assert.assertTrue(sam.getReferenceIndex() == SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX);
+    }
+
+    @Test(expectedExceptions=IllegalArgumentException.class)
+    public void testNullHeaderSetReferenceIndex() {
+        final SAMRecord sam = new SAMRecordSetBuilder().addFrag("test", 0, 1, false, false, "3S9M", null, 2);
+        sam.setHeader(null);
+        // setReferenceIndex with null header throws
+        sam.setReferenceIndex(3);
+    }
```
droazen: Add a companion test case that calls `getReferenceIndex()` on a `SAMRecord` with a null header, and expects an exception.

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -90,8 +91,283 @@ public void testOffsetAtReference(String cigar, int posInReference, int expected
...
+
+    @Test(expectedExceptions=IllegalArgumentException.class)
+    public void testNullHeaderForceIndexResolutionFailure() {
+        // force the internal SAMRecord reference index value to null initial state
+        final SAMRecord sam = new SAMRecord(null);
+        sam.setReferenceName("unresolvable");
+        sam.getReferenceIndex();
```
droazen: Add a companion test case that starts with a fully-populated `SAMRecord` with valid contig, nulls out the header, calls `getReferenceIndex()`, and expects an exception.

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -90,8 +91,283 @@ public void testOffsetAtReference(String cigar, int posInReference, int expected
...
+
+    @Test
+    public void testNullHeaderMateReferenceName() {
+        SAMRecord sam = new SAMRecordSetBuilder().addFrag("test", 0, 1, false, false, "3S9M", null, 2);
+        final SAMFileHeader samHeader = sam.getHeader();
+        sam.setMateReferenceName("chr1");
+        final String origMateRefName = sam.getMateReferenceName();
+        final int originalMateRefIndex = sam.getMateReferenceIndex();
+
+        // setting header to null retains ref name and index
```
droazen: It shouldn't retain the index, it should null it out (and you should check that it does, since `mMateReferenceIndex` has protected access and you're in the right package).

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -90,8 +91,283 @@ public void testOffsetAtReference(String cigar, int posInReference, int expected
...
+        Assert.assertTrue(sam.getMateReferenceIndex() == SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX);
+
+        // force the internal SAMRecord reference index value to null initial state
+        sam = new SAMRecord(null);
+        Assert.assertTrue(sam.getMateReferenceIndex() == SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX);
+
+        // force the internal SAMRecord reference index value to null initial state
+        sam = new SAMRecord(null);
+        final String unresolvableRefName = "unresolvable";
+        sam.setMateReferenceName(unresolvableRefName);
```
droazen: Check value of `mMateReferenceIndex` at this point, taking advantage of protected access.

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -90,8 +91,283 @@ public void testOffsetAtReference(String cigar, int posInReference, int expected
...
+        sam.setMateReferenceIndex(3);
+    }
+
+    @Test(expectedExceptions=IllegalArgumentException.class)
+    public void testNullHeaderForceMateIndexResolutionFailure() {
+        // force the internal SAMRecord reference index value to null initial state
+        final SAMRecord sam = new SAMRecord(null);
+        sam.setMateReferenceName("unresolvable");
+        sam.getMateReferenceIndex();
+    }
```
droazen: Add a test that starts with a sam with a header and a valid mate reference name, nulls the header, calls `getMateReferenceIndex()`, and expects an exception.

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -57,27 +57,30 @@
...
  * Some of the get() methods return values that are mutable, due to the limitations of Java.  A caller should
  * never change the value returned by a get() method.  If you want to change the value of some attribute of a
  * SAMRecord, create a new value object and call the appropriate set() method.
  *
+ * Note that setIndexingBin() need not be called when writing SAMRecords.  It will be computed as necessary.  It is only
+ * present as an optimization in the event that the value is already known and need not be computed.
+ *
  * By default, extensive validation of SAMRecords is done when they are read.  Very limited validation is done when
  * values are set onto SAMRecords.
+ *
```
droazen: Add javadoc to the `SAMRecord` constructor documenting behavior when a null header is passed in (can re-use docs from `setHeader()`

droazen:
Review complete -- back to @cmnbroad. Looks very good overall, much closer to what we want than the first version. Mostly minor comments, but a few concerns here and there that need a response.

src/java/htsjdk/samtools/cram/build/Sam2CramRecordFactory.java
cmnbroad: Technically, no, though it only gets used to create records for CRAMFileWriter. Still, its confusing to use the word "writer" here so I'll change it to "factory."

src/java/htsjdk/samtools/SAMRecord.java
cmnbroad: We're using the null value to indicate "unresolved", which is not quite correct in that case, but it seems harmless to do and cleans up the code so I'll change it.

src/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
cmnbroad: True that. Done.

src/java/htsjdk/samtools/MergingSamRecordIterator.java
cmnbroad: I don' think we can leave this code in with the current changes - it assumes the reference indices are stale and relative to the old dictionary, and tries to resolve them by translation. With these changes, the setHeader call above will resolve the indices (by name) against the new dictionary, so they'll already be translated. Treating them as if they're in the old space would be incorrect I think. I should be able to add a test to verify this.

src/java/htsjdk/samtools/SAMRecord.java
cmnbroad: We could go either way (before this change, there was already a validation error enum value for MISSING_HEADER, thoughits unclear to me if there was actually any code path that would produce such an error on a record with a null header without first throwing an NPE). One other option I like is to change the MISSING_HEADER enum value to be a SAMValidationError.Severity.WARNING instead of an ERROR, and only return those for each case where a header is required for validation (including SAMRecord.validateCigar and SAMUtils.validateCigar which can both be called independently of SAMRecord.isValid()).

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
cmnbroad: I agree its not ideal and if I was designing the API from scratch I wouldn't make that choice, but I wouldn't change it at this point either.

cmnbroad:
Back to @droazen. I made all of the changes except for how we handle validation errors; one other option is to make the null-header case a warning rather than a hard error (see comment above). Just wanted to present another option, let me know which you prefer and I'll make the code and doc changes for that.

I also added merging iterator test that forces and tests index remapping, and fixed up some javadoc. I'll also have to make a minor modification in this branch to the deepCopy code to accommodate these changes, but I can't do that until I rebase.

One more thing, I think I mentioned this a while back, but these changes break this [SAMRecordToGATKReadAdapterSerializer test ](https://github.com/broadinstitute/gatk/blob/master/src/test/java/org/broadinstitute/hellbender/engine/spark/SAMRecordToGATKReadAdapterSerializerUnitTest.java#L37) in Hellbender, since it tries to use the BAMRecordCodec to encode/decode records after their header has been set to null. The BAMRecordCodec doesn't serialize the ref names, only the indices, which are clobbered by the setHeader(null). The other cloud mandatory tests all pass.

cmnbroad:
Squashed and rebased in order to integrate deepCopy changes with null headers and test. Added a test to make sure SAMRecordMergingIterator handles index remapping properly. Validation with null headers is no longer a validation error. Ready for review once tests pass.

droazen:
Looks good, except that it appears that a headerless record still can't fully pass validation -- specifically, cigar validation via `SAMUtils.validateCigar()` will fail. This is a problem for the headerless use case, as `SAMRecord.getCigar()` calls `validateCigar()`, so we can't even get the cigar on a headerless record without getting an exception if STRICT validation is on. Can you address?

PR #328, opened 2015-09-23
Handle ambiguous IUPAC chars in reference; change NM calculation

mattsooknah:
* Add support for IUPAC ambiguous base codes. 
* Add a new function, ```basesMatchWithAmbiguity```, that will consider the set relationships between codes when matching a read base to a reference base. 
 * Specifically, if the read base code denotes a subset of the bases represented by the reference base code (e.g. A on read -> R on ref, or M on read -> H on ref), then it's a match.
* Use ```basesMatchWithAmbiguity``` for calculating the NM tag.
* Update documentation and add tests for ```SequenceUtil``` methods.

**Note that this PR changes the NM calculation to be different from samtools. @tfenne @ktibbett @eitanbanks have all OK'd this.**

### DIFF
--- a/src/java/htsjdk/samtools/util/SequenceUtil.java
+++ b/src/java/htsjdk/samtools/util/SequenceUtil.java
@@ -46,8 +46,45 @@ import java.util.regex.Pattern;
 public class SequenceUtil {
     /** Byte typed variables for all normal bases. */
     public static final byte a = 'a', c = 'c', g = 'g', t = 't', n = 'n', A = 'A', C = 'C', G = 'G', T = 'T', N = 'N';
+
     public static final byte[] VALID_BASES_UPPER = new byte[]{A, C, G, T};
     public static final byte[] VALID_BASES_LOWER = new byte[]{a, c, g, t};
+
+    private static final byte A_MASK = 1;
+    private static final byte C_MASK = 2;
+    private static final byte G_MASK = 4;
+    private static final byte T_MASK = 8;
+
+    private static final byte[] bases = new byte[127];
+
+    /*
+     * Definition of IUPAC codes:
+     * http://www.bioinformatics.org/sms2/iupac.html
+     */
+    static {
+        Arrays.fill(bases, (byte) 0);
+        bases[A] = A_MASK;
+        bases[C] = C_MASK;
+        bases[G] = G_MASK;
+        bases[T] = T_MASK;
+        bases['M'] = A_MASK | C_MASK;
+        bases['R'] = A_MASK | G_MASK;
+        bases['W'] = A_MASK | T_MASK;
+        bases['S'] = C_MASK | G_MASK;
+        bases['Y'] = C_MASK | T_MASK;
+        bases['K'] = G_MASK | T_MASK;
+        bases['V'] = A_MASK | C_MASK | G_MASK;
+        bases['H'] = A_MASK | C_MASK | T_MASK;
+        bases['D'] = A_MASK | G_MASK | T_MASK;
+        bases['B'] = C_MASK | G_MASK | T_MASK;
+        bases['N'] = A_MASK | C_MASK | G_MASK | T_MASK;
+        // Also store the bases in lower case
+        for (int i = 'A'; i <= 'Z'; i++) {
+            bases[(byte) i + 32] = bases[(byte) i];
+        }
+        bases['.'] = A_MASK | C_MASK | G_MASK | T_MASK;
+    };
+
 
     /**
      * Calculate the reverse complement of the specified sequence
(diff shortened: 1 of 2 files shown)


src/java/htsjdk/samtools/util/SequenceUtil.java
```diff
@@ -848,7 +922,10 @@ public static String calculateMD5String(final byte[] data, final int offset, fin
     }
 
     /**
-     * A rip off samtools bam_md.c
+     * Calculate MD and NM similarly to Samtools, except that N->N is a match.
+     *
+     * TODO this calculates NM differently than calculateSamNmTag
+     * TODO (specifically, it does not check set relationships between ambiguous bases.
```
mattsooknah: It looks like this method was originally added as part of the CRAM changes. I'm not sure what will be affected by changing it, so I leave it for now and open a separate ticket.

mattsooknah:
@tfenne please review

tfenne:
@mattsooknah I've looked through the code and I think it does a good job of the intended change.

That said, I'm still not clear on why anyone wants to change the way the NM tag is calculated, especially as this will then diverge from samtools.  I can understand the logic of having read bases match to ambiguity codes in the reference, and I wouldn't argue that it is wrong to do so, I'm just not clear on why change it now given that both htsjdk and samtools have been calculating it one way forever.  Is there any background?

mattsooknah:
@tfenne it's a long story, but here's the summary as I understand it:

We were running into issues where GATK (specifically IndelRealigner) converted IUPAC ambiguous codes on the reference to N, and then calculated the NM tag. This resulted in e.g. (N on read -> W on reference) being counted as a match. Subsequently ValidateSam was being called on the indel-realigned BAM, and since Picard/HTSJDK does not convert IUPAC bases to N, the NM values were different, causing GP pipelines to fail when using certain references with ambiguous bases.

There was some discussion about how to make the GATK and Picard calculations consistent. The conclusion was that we should make both programs handle ambiguous codes in a robust way. The latest GATK now calls ```Sequenceutil.calculateSamNmTag```, so they will be consistent as well.

mattsooknah:
@tfenne However, do you think it would be good to solicit more community opinion on this before committing to the change?

yfarjoun:
I think that we should get community opinion on this. It might also spawn a
change in the sam-spec document if we think that there is an ambiguity or
something unclear there...so perhaps we should have a issue open and see
what folks say.

Yossi.

tfenne:
@mattsooknah I totally agree with @yfarjoun .  At a minimum I think it would be good to talk to @jkbonfield and @jmarshall and a) update the SAM specification to be clear on how ambiguity codes should be handled and b) ensure that htsjdk and htslib/samtools stay in sync on the implementation.

mattsooknah:
OK. @ktibbett is away right now, but when she gets back I'll ask her how she wants to move forward, as she was initially interested in this change.

eitanbanks:
@tfenne I'm going back to this ticket (I'm clearing out my backlog of things to do over vacation).  I've updated it so that we do NOT change the behavior of calculateSamNmTag anymore; but I think that the base comparison changes were an improvement - and they do allow us to change back to proper IUPAC matching if/when we get community consensus.  You still okay to review this or should I assign to someone else?

eitanbanks:
Re-pinging you @tfenne.  It's totally cool if you don't have time to review, but please let me know and I'll assign to someone else.
Thanks!

src/java/htsjdk/samtools/util/SequenceUtil.java
```diff
@@ -62,15 +99,27 @@ public static String reverseComplement(final String sequenceData) {
...
+     * a reference sequence, using the reference code as a 'pattern' that the read base must match.
+     *
+     * We take ambiguous codes into account, returning true if the set of possible bases
+     * represented by the read value is a (non-strict) subset of the possible bases represented
+     * by the reference value.
+     *
+     * Since the comparison is directional, make sure to pass read / ref codes in correct order.
+     */
+    public static boolean basesMatchWithAmbiguity(final byte readBase, final byte refBase) {
+        return (bases[readBase] & bases[refBase]) == bases[readBase];
```
tfenne: Something was bugging me about the asymmetry of this method, and I think it's the name.  `basesMatchWithAmbiguity` sounds like it should be reflexive, but the method is not.  Calling this something like `readBaseCanMatchRefBase` or similar that implies a directionality to the comparison might make this less error prone.

src/java/htsjdk/samtools/util/SequenceUtil.java
```diff
@@ -307,6 +357,20 @@ public static String makeSoftClipCigar(final int clipLength) {
         return Integer.toString(clipLength) + "S";
     }
 
+    /**
+     * Helper method to handle the various use cases of base comparison.
+     */
```
tfenne: This could probably use a little doc on the parameters.  Specifically that `matchAmbigious` causes the match to return true when the read base is a subset of the possible reference bases, but not the other way around.  Same comment about name implying reflexiveness.

tfenne:
Looks good to me @eitanbanks .  I'm on board with changing how `NM` is calculated, but only if samtools is changed at the same time so we don't have a stream of "why does Picard differ from samtools" or "why does Picard say my NM tag is invalid" type emails!

eitanbanks:
That's fair.  For now we'll maintain the existing behavior and will make a push if we decide it's worth the effort.
Just addressed your reviewer comments and rebased.  Will merge when tests pass.
Thanks!

PR #336, opened 2015-09-28
fix for CramCompressionRecord.getAlignmentEnd: unmapped reads should …

vadimzalunin:
Sometimes when converting to CRAM slice boundaries extend beyond reference, causing "Slice partially mapped outside of reference" error when reading CRAM. This brunch fixes calculation of alignment end in CramCompressionRecord: unmapped reads should not count towards slice boundaries.


droazen:
:+1:

PR #338, opened 2015-09-29
- SetMateInfoIterator throws an exception when it finds missing

nh13:
mates if desired
- Fixing silly compile warning in ProgressLogger

### DIFF
--- a/src/java/htsjdk/samtools/SamPairUtil.java
+++ b/src/java/htsjdk/samtools/SamPairUtil.java
@@ -25,9 +25,7 @@
 package htsjdk.samtools;
 
 import htsjdk.samtools.util.PeekableIterator;
-import htsjdk.samtools.util.ProgressLogger;
 
-import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
@@ -52,9 +50,9 @@ public class SamPairUtil {
     {
         FR,     // ( 5' --F-->       <--R-- 5'  )  - aka. innie
         RF,     // (   <--R-- 5'   5' --F-->    )  - aka. outie
-        TANDEM; // ( 5' --F-->   5' --F-->  or  (  <--R-- 5'   <--R-- 5'  )
+        TANDEM // ( 5' --F-->   5' --F-->  or  (  <--R-- 5'   <--R-- 5'  )
 
-    };
+    }
 
 
     /**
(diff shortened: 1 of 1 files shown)


nh13:
See: https://github.com/broadinstitute/picard/issues/201

src/java/htsjdk/samtools/SamPairUtil.java
```diff
@@ -426,27 +434,35 @@ else if (record.getSecondOfPairFlag()) {
...
-            // Set mate information on supplemental records
-            if (containsSupplementalRecord) {
-                for (final SAMRecord record : records) {
-                    if (record.getReadPairedFlag() && record.getSupplementaryAlignmentFlag()) {
-                        if (record.getFirstOfPairFlag()) {
-                            SamPairUtil.setMateInformationOnSupplementalAlignment(record, secondPrimaryRecord, this.setMateCigar);
-                        }
-                        else {
-                            SamPairUtil.setMateInformationOnSupplementalAlignment(record, firstPrimaryRecord, this.setMateCigar);
+                //}
```
tfenne: Remove dead `}`

src/java/htsjdk/samtools/SamPairUtil.java
```diff
@@ -426,27 +434,35 @@ else if (record.getSecondOfPairFlag()) {
...
+                                SamPairUtil.setMateInformationOnSupplementalAlignment(record, secondPrimaryRecord, this.setMateCigar);
+                            } else {
+                                SamPairUtil.setMateInformationOnSupplementalAlignment(record, firstPrimaryRecord, this.setMateCigar);
+                            }
+                            this.numMateCigarsAdded++;
                         }
-                        this.numMateCigarsAdded++;
                     }
                 }
+            } else if (!this.ignoreMissingMates) {
```
tfenne: I think this could still fail with an NPE in the (admittedly invalid edge) case where *all* you have is supplemental reads and both the R1 and R2 primary records are missing.

nh13: Where would the NPE happen if both ```firstPrimaryRecord``` and ```secondPrimaryRecord``` were ```null```?

The ```if``` requires both primary R1/R2: ```if (null != firstPrimaryRecord && null != secondPrimaryRecord)```
The ```if``` in the else clause requires primary R1: ```if (null != firstPrimaryRecord && firstPrimaryRecord.getReadPairedFlag())```
The ```else if``` in the else clause requires primary R2: ```else if (null != secondPrimaryRecord && secondPrimaryRecord.getReadPairedFlag())```

Issue #322, opened 2015-09-11
uint32_t is a valid BAM type but blows up CRAM


mp15:
```
ref1_grp2_p003  99  ref1    11  5   10M =   35  34  ACCCGGGGAT  &&&&&&&&&&  ia:i:4294967295 MD:Z:10 NM:i:0  RG:Z:grp2
```

This line from the standard Samtools test suite blows up HTSJDK and thus GATK. Given that Java 8 has explicit support for unsigned arithmetic on the Integer type it might be good to fix this. I am told it originates here:
https://github.com/samtools/htsjdk/blob/3d13016bef282595328d4eb6eed9c71d885a0347/src/java/htsjdk/samtools/cram/structure/ReadTag.java#L380

mp15:
Specifically I suggest you use:
http://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#parseUnsignedInt-java.lang.String
When you do migrate to 8 but in the mean time use a long?

vadimzalunin:
see PR https://github.com/samtools/htsjdk/pull/339, blocked by travis failure irrelevent to the PR, issue https://github.com/samtools/htsjdk/issues/341

droazen:
Fixed in https://github.com/samtools/htsjdk/pull/339 -- closing!

PR #339, opened 2015-09-30
unsigned int is backed by a long value

vadimzalunin:
replaced exception when reading unsigned int in ReadTag with a direct conversion to long

### DIFF
--- a/src/java/htsjdk/samtools/CRAMFileWriter.java
+++ b/src/java/htsjdk/samtools/CRAMFileWriter.java
@@ -379,6 +379,8 @@ public class CRAMFileWriter extends SAMFileWriterImpl {
             outputStream.flush();
             if (indexer != null)
                 indexer.finish();
+        } catch (final RuntimeException re) {
+            throw re;
         } catch (final Exception e) {
             throw new RuntimeException(e);
         }
--- a/src/java/htsjdk/samtools/CRAMIterator.java
+++ b/src/java/htsjdk/samtools/CRAMIterator.java
@@ -153,7 +153,7 @@ public class CRAMIterator implements SAMRecordIterator {
         else
             cramRecords.clear();
 
-        parser.getRecords(container, cramRecords);
+        parser.getRecords(container, cramRecords, validationStringency);
 
         if (container.sequenceId == SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX) {
             refs = new byte[]{};
@@ -253,7 +253,12 @@ public class CRAMIterator implements SAMRecordIterator {
         if (!iterator.hasNext()) {
             try {
                 nextContainer();
-            } catch (final Exception e) {
+            } catch (CRAMException ce) {
+                throw ce;
+            } catch (SAMFormatException se) {
+                throw se;
+            }
+            catch (final Exception e) {
                 throw new RuntimeEOFException(e);
             }
         }
--- a/src/java/htsjdk/samtools/SAMRecord.java
+++ b/src/java/htsjdk/samtools/SAMRecord.java
@@ -966,8 +966,8 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
     /**
      * Get the tag value and attempt to coerce it into the requested type.
      * @param tag The requested tag.
-     * @return The value of a tag, converted into an Integer if possible.
-     * @throws RuntimeException If the value is not an integer type, or will not fit in an Integer.
+     * @return The value of a tag, converted into a signed Integer if possible.
+     * @throws RuntimeException If the value is not an integer type, or will not fit in a signed Integer.
      */
     public Integer getIntegerAttribute(final String tag) {
         final Object val = getAttribute(tag);
@@ -983,6 +983,46 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
             throw new RuntimeException("Value for tag " + tag + " is not in Integer range: " + longVal);
         }
         return (int)longVal;
+    }
+
+    /**
+     * A convenience method that will return a valid unsigned integer as a Long,
+     * or fail with an exception if the tag value is invalid.
+     *
+     * @param tag Two-character tag name.
+     * @return valid unsigned integer associated with the tag, as a Long
+     * @throws {@link htsjdk.samtools.SAMException} if the value is out of range for a 32-bit unsigned value, or not a Number
+     */
+    public Long getUnsignedIntegerAttribute(final String tag) throws SAMException {
+        return getUnsignedIntegerAttribute(SAMTagUtil.getSingleton().makeBinaryTag(tag));
+    }
+
+    /**
+     * A convenience method that will return a valid unsigned integer as a Long,
+     * or fail with an exception if the tag value is invalid.
+     *
+     * @param tag Binary representation of a 2-char String tag as created by SAMTagUtil.
+     * @return valid unsigned integer associated with the tag, as a Long
+     * @throws {@link htsjdk.samtools.SAMException} if the value is out of range for a 32-bit unsigned value, or not a Number
+     */
+    public Long getUnsignedIntegerAttribute(final short tag) throws SAMException {
+        final Object value = getAttribute(tag);
+        if (value == null) {
+            return null;
+        }
+
+        if (value instanceof Number) {
+            final long lValue = ((Number)value).longValue();
+            if (SAMUtils.isValidUnsignedIntegerAttribute(lValue)) {
+                return lValue;
+            } else {
+                throw new SAMException("Unsigned integer value of tag " +
+                        SAMTagUtil.getSingleton().makeStringTag(tag) + " is out of bounds for a 32-bit unsigned integer: " + lValue);
+            }
+        } else {
+            throw new SAMException("Unexpected attribute value data type " + value.getClass() + " for tag " +
+                    SAMTagUtil.getSingleton().makeStringTag(tag));
+        }
     }
 
     /**
@@ -1161,14 +1201,15 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
     /**
      * Set a named attribute onto the SAMRecord.  Passing a null value causes the attribute to be cleared.
      * @param tag two-character tag name.  See http://samtools.sourceforge.net/SAM1.pdf for standard and user-defined tags.
-     * @param value Supported types are String, Char, Integer, Float, byte[], short[]. int[], float[].
+     * @param value Supported types are String, Char, Integer, Float,
+     *              Long (for values that fit into a signed or unsigned 32-bit integer only),
+     *              byte[], short[], int[], float[].
      * If value == null, tag is cleared.
      *
      * Byte and Short are allowed but discouraged.  If written to a SAM file, these will be converted to Integer,
      * whereas if written to BAM, getAttribute() will return as Byte or Short, respectively.
      *
-     * Long with value between 0 and MAX_UINT is allowed for BAM but discouraged.  Attempting to write such a value
-     * to SAM will cause an exception to be thrown.
+     * Long is allowed for values that fit into a signed or unsigned 32-bit integer only, but discouraged.
      *
      * To set unsigned byte[], unsigned short[] or unsigned int[] (which is discouraged because of poor Java language
      * support), setUnsignedArrayAttribute() must be used instead of this method.
@@ -1206,24 +1247,43 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
         setAttribute(tag, value, false);
     }
 
+    /**
+     * Checks if the value is allowed as an attribute value.
+     *
+     * @param value the value to be checked
+     * @return true if the value is valid and false otherwise
+     */
+    protected static boolean isAllowedAttributeValue(final Object value) {
+        if (value instanceof Byte || value instanceof Short || value instanceof Integer ||
+                value instanceof String || value instanceof Character || value instanceof Float ||
+                value instanceof byte[] || value instanceof short[] || value instanceof int[] ||
+                value instanceof float[]) {
+            return true;
+        }
+
+        // A special case for Longs: we require Long values to fit into either a uint32_t or an int32_t,
+        // as that is what the BAM spec allows.
+        if (value instanceof Long) {
+            return SAMUtils.isValidUnsignedIntegerAttribute((Long) value)
+                    || ((Long) value >= Integer.MIN_VALUE && (Long) value <= Integer.MAX_VALUE);
+        }
+        return false;
+    }
+
     protected void setAttribute(final short tag, final Object value, final boolean isUnsignedArray) {
-        if (value != null &&
-                !(value instanceof Byte || value instanceof Short || value instanceof Integer ||
-                        value instanceof String || value instanceof Character || value instanceof Float ||
-                        value instanceof byte[] || value instanceof short[] || value instanceof int[] ||
-                        value instanceof float[])) {
-            throw new SAMException("Attribute type " + value.getClass() + " not supported. Tag: " +
-                    SAMTagUtil.getSingleton().makeStringTag(tag));
-        }
         if (value == null) {
-            if (this.mAttributes != null) this.mAttributes = this.mAttributes.remove(tag);
-        }
-        else {
-            final SAMBinaryTagAndValue tmp;
-            if(!isUnsignedArray) {
-                tmp = new SAMBinaryTagAndValue(tag, value);
+            // setting a tag value to null removes the tag:
+            if (this.mAttributes != null) {
+                this.mAttributes = this.mAttributes.remove(tag);
             }
-            else {
+            return;
+        }
+
+        if (isAllowedAttributeValue(value)) {
+            final SAMBinaryTagAndValue tmp;
+            if (!isUnsignedArray) {
+                tmp = new SAMBinaryTagAndValue(tag, value);
+            } else {
                 if (!value.getClass().isArray() || value instanceof float[]) {
                     throw new SAMException("Attribute type " + value.getClass() +
                             " cannot be encoded as an unsigned array. Tag: " +
@@ -1231,8 +1291,15 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
                 }
                 tmp = new SAMBinaryTagAndUnsignedArrayValue(tag, value);
             }
-            if (this.mAttributes == null) this.mAttributes = tmp;
-            else this.mAttributes = this.mAttributes.insert(tmp);
+
+            if (this.mAttributes == null) {
+                this.mAttributes = tmp;
+            } else {
+                this.mAttributes = this.mAttributes.insert(tmp);
+            }
+        } else {
+            throw new SAMException("Attribute type " + value.getClass() + " not supported. Tag: " +
+                    SAMTagUtil.getSingleton().makeStringTag(tag));
         }
     }
 
--- a/src/java/htsjdk/samtools/cram/encoding/reader/CramRecordReader.java
+++ b/src/java/htsjdk/samtools/cram/encoding/reader/CramRecordReader.java
@@ -17,7 +17,9 @@
  */
 package htsjdk.samtools.cram.encoding.reader;
 
+import htsjdk.samtools.SAMFormatException;
 import htsjdk.samtools.SAMRecord;
+import htsjdk.samtools.ValidationStringency;
 import htsjdk.samtools.cram.encoding.readfeatures.BaseQualityScore;
 import htsjdk.samtools.cram.encoding.readfeatures.Bases;
 import htsjdk.samtools.cram.encoding.readfeatures.Deletion;
@@ -38,6 +40,11 @@ import java.util.LinkedList;
 
 public class CramRecordReader extends AbstractReader {
     private CramCompressionRecord prevRecord;
+    private ValidationStringency validationStringency;
+
+    public CramRecordReader(ValidationStringency validationStringency) {
+        this.validationStringency = validationStringency;
+    }
 
     @SuppressWarnings("ConstantConditions")
     public void read(final CramCompressionRecord cramRecord) {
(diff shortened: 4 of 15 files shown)


droazen:
@vadimzalunin I've made some changes in master to make it easier to see which tests are failing in travis, and have taken the liberty of rebasing this branch on top of those changes. It's clear from doing this that the test failure in this branch is closely related to the changes you've made in it, as you should now be able to see in travis:

    [testng] FAILED: test("auxf#values")
    [testng] htsjdk.samtools.util.RuntimeEOFException: htsjdk.samtools.SAMException: Attribute type class java.lang.Long not supported. Tag: I9
    [testng] 	at htsjdk.samtools.CRAMIterator.hasNext(CRAMIterator.java:257)
    [testng] 	at htsjdk.samtools.CRAMComplianceTest.test(CRAMComplianceTest.java:123)
    [testng] Caused by: htsjdk.samtools.SAMException: Attribute type class java.lang.Long not supported. Tag: I9
    [testng] 	at htsjdk.samtools.SAMRecord.setAttribute(SAMRecord.java:1210)
    [testng] 	at htsjdk.samtools.SAMRecord.setAttribute(SAMRecord.java:1201)
    [testng] 	at htsjdk.samtools.SAMRecord.setAttribute(SAMRecord.java:1177)
    [testng] 	at htsjdk.samtools.cram.build.Cram2SamRecordFactory.create(Cram2SamRecordFactory.java:93)
    [testng] 	at htsjdk.samtools.CRAMIterator.nextContainer(CRAMIterator.java:191)
    [testng] 	at htsjdk.samtools.CRAMIterator.hasNext(CRAMIterator.java:255)
    [testng] 	... 22 more
    [testng] ... Removed 21 stack frames
    [testng] 
    [testng] ===============================================
    [testng]     Ant test
    [testng]     Tests run: 22443, Failures: 1, Skips: 0
    [testng] ===============================================

droazen:
Back to @vadimzalunin for test fixes

mp15:
Looks like this only failing on the incorrect test testUnsignedIntegerSAM which is expecting an exception to be thrown (it shouldn't be with this fix in place). Maybe update the test?

vadimzalunin:
@mp15 done

droazen:
@vadimzalunin Could you please squash this branch into a single commit?

vadimzalunin:
@droazen squashed

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1203,7 +1203,7 @@ protected void setAttribute(final short tag, final Object value) {
 
     protected void setAttribute(final short tag, final Object value, final boolean isUnsignedArray) {
         if (value != null &&
-                !(value instanceof Byte || value instanceof Short || value instanceof Integer ||
+                !(value instanceof Byte || value instanceof Short || value instanceof Integer || value instanceof Long ||
```
droazen: With this change to allow `Long` values directly in `SAMRecord` attributes, what protection do we have against a client setting an attribute outside the range of `uint32_t` that bams can handle? The htsjdk BAM codec will fail given a `Long` value that exceeds `BinaryCodec.MAX_UINT`.

Also, if you set such an attribute using `setAttribute()` and then call `getIntegerAttribute()` you'll get a `RuntimeException` (even though the attribute may be marked as of type i/I, like the `ia:i:4294967295` attribute in http://gatkforums.broadinstitute.org/discussion/comment/24038/#Comment_24038).

Can you comment on these concerns? I don't want to block this important fix, but I want to be sure we're not doing something here that will cause problems later on.

droazen:
Back to @vadimzalunin for comments.

vadimzalunin:
@droazen re-implemented the fix, basically made uint32 policy the same in CRAM as in BAM. Added dedicated test class and a BAM file.

vadimzalunin:
added tests for values over max unsigned int

vadimzalunin:
@droazen please review

mp15:
@vadimzalunin It won't stop an effective review but it looks like you've accidentally left the squashed branch as two commits with same message rather than one?

vadimzalunin:
@mp15  rebased/squashed

src/java/htsjdk/samtools/SAMRecord.java
droazen: I don't think my concerns from last time about allowing `Longs` into `SAMRecord` attributes have been adequately addressed. We need a range check at this `setAttribute()` entry point for the case where `value instanceof Long`, and throw an `IllegalArgumentException` from here if `value < 0 || value > BinaryCodec.MAX_UINT` (plus corresponding unit tests showing that it throws in these cases). I think we should also add a `getUnsignedIntegerAttribute()` accessor to `SAMRecord` that returns a `Long`, since the existing `getIntegerAttribute()` method can't handle the full range of 32-bit unsigned int values.

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1203,7 +1203,7 @@ protected void setAttribute(final short tag, final Object value) {
 
     protected void setAttribute(final short tag, final Object value, final boolean isUnsignedArray) {
         if (value != null &&
-                !(value instanceof Byte || value instanceof Short || value instanceof Integer ||
+                !(value instanceof Byte || value instanceof Short || value instanceof Integer ||value instanceof Long ||
```
droazen: I don't think my concerns from last time about allowing `Longs` into `SAMRecord` attributes have been adequately addressed. We need a range check at this `setAttribute()` entry point for the case where `value instanceof Long`, and throw an `IllegalArgumentException` from here if `value < 0 || value > BinaryCodec.MAX_UINT` (plus corresponding unit tests showing that it throws in these cases). I think we should also add a `getUnsignedIntegerAttribute()` accessor to `SAMRecord` that returns a `Long`, since the existing `getIntegerAttribute()` method can't handle the full range of 32-bit unsigned int values.

src/java/htsjdk/samtools/CRAMIterator.java
```diff
@@ -83,10 +83,10 @@ public void setValidationStringency(
     public CRAMIterator(final InputStream inputStream, final ReferenceSource referenceSource)
             throws IOException {
         if (null == referenceSource) {
-            throw new CRAMException("A reference source is required for CRAM files");
-        }
+            this.referenceSource = new ReferenceSource();
+        } else
+            this.referenceSource = referenceSource;
```
droazen: Add curly braces around else block. Also, falling back to a default-constructed `ReferenceSource` will (I believe) eventually produce a `NullPointerException` deep in the cram parsing code provided that `Defaults.USE_CRAM_REF_DOWNLOAD` is false (which it is by default). I think we should throw here if `referenceSource` is null and `Defaults.USE_CRAM_REF_DOWNLOAD` is false.

src/java/htsjdk/samtools/CRAMIterator.java
```diff
@@ -101,10 +101,10 @@ public CRAMIterator(final InputStream inputStream, final ReferenceSource referen
     public CRAMIterator(final SeekableStream seekableStream, final ReferenceSource referenceSource, final long[] coordinates)
             throws IOException {
         if (null == referenceSource) {
-            throw new CRAMException("A reference source is required for CRAM files");
-        }
+            this.referenceSource = new ReferenceSource();
+        } else
+            this.referenceSource = referenceSource;
```
droazen: Same note above about missing curly braces and relying on a a default-constructed `ReferenceSource` applies here as well.

src/java/htsjdk/samtools/cram/encoding/reader/CramRecordReader.java
```diff
@@ -190,6 +197,7 @@ public void read(final CramCompressionRecord cramRecord) {
             if (prevRecord != null)
                 System.err.printf("Failed at record %d. Here is the previously read record: %s\n", recordCounter,
                         prevRecord.toString());
+            if (e instanceof SAMFormatException) throw (SAMFormatException)e;
```
droazen: Add a separate catch block for `SAMFormatException` instead of calling `instanceof` on the weakly-typed exception object.

src/java/htsjdk/samtools/cram/structure/ReadTag.java
```diff
@@ -374,10 +377,12 @@ public static Object readSingleValue(final byte tagType,
...
                 if (val <= Integer.MAX_VALUE) {
-                    return (int) val;
+                    return (int)val;
                 }
-                throw new RuntimeException(
-                        "Tag value is too large to store as signed integer.");
+                SAMUtils.processValidationError(new SAMValidationError(SAMValidationError.Type.TAG_VALUE_TOO_LARGE,
+                        "Tag value " + val + " too large to store as signed integer.", null), validationStringency);
+                // convert to unsigned int stored in a long
+                return val;
```
droazen: I don't think this is the behavior we want here... When writing a bam, if the attribute value is a `Long` that is > `BinaryTagCodec.MAX_INT` but <= `BinaryTagCodec.MAX_UINT`, we assign it a tag type of "I" (as opposed to "i"). This is consistent with the BAM spec. So when reading attributes back in, if the attribute type is "I" (rather than "i") we should allow all values >= 0 and <= `BinaryTagCodec.MAX_UINT` rather than raising a validation error if the value is too large to be stored in a signed int.

droazen: I realize that you were just trying to replicate the existing behavior here, but the existing behavior seems to simply be wrong (we can't handle "I" attributes properly in htsjdk). For consistency, I think `BinaryTagCodec.readSingleValue()` should also change to allow all values >= 0 and <= `BinaryTagCodec.MAX_UINT`

droazen:
Back to @vadimzalunin for more changes. To summarize what I think should be done here:

-Add extra range checking in `SAMRecord.setAttribute()` when the attribute value is of type `Long` to ensure that all `Long` values are >= 0 and <= `BinaryTagCodec.MAX_UINT`

-Add a `getUnsignedIntegerAttribute()` accessor to SAMRecord that returns a Long and does not blow up if the value is > `Integer.MAX_INT`

-Change both `ReadTag.readSingleValue()` and `BinaryTagCodec.readSingleValue()` to allow all values  >= 0 and <= `BinaryTagCodec.MAX_UINT` for attributes of type "I" (as opposed to "i").

-Add a new unit test that shows that you can successfully round-trip a `Long` attribute that is > `Integer.MAX_INT` and <= `BinaryTagCodec.MAX_UINT` through a SAM, BAM, and CRAM even with strict validation on.

vadimzalunin:
@drozen This started as a tiny CRAM bug fix but gradually turned into policy changer :)

I didn't want to make SAMRecord class aware of limitations in BinaryTagCodec as it is implementation-specific. I guess we just need to promote this check up to the SAMRecord class and make it a top-level policy. 

As for tests passing even on strict validation, this means changing the existing behavior of SAM/BAM.  Fine with me as long as there is no objections from others.

eitanbanks:
@tfenne @nh13 I'd love to hear your thoughts on this if you are willing.

src/java/htsjdk/samtools/CRAMIterator.java
cmnbroad: In this particular case (reading via CRAMIterator), you'll get a CRAMException "Contig  nnn not found in reference file" if no reference is provided :
	at htsjdk.samtools.CRAMIterator.hasNext(CRAMIterator.java:257)
	at htsjdk.samtools.SamReader$AssertingIterator.hasNext(SamReader.java:567)
	at org.broadinstitute.hellbender.utils.iterators.SAMRecordToReadIterator.hasNext(SAMRecordToReadIterator.java:24)

The NPE happens when you *write* a CRAM file with no reference:

java.lang.RuntimeException: java.lang.NullPointerException
	at htsjdk.samtools.CRAMFileWriter.writeAlignment(CRAMFileWriter.java:339)
	at htsjdk.samtools.SAMFileWriterImpl.addAlignment(SAMFileWriterImpl.java:180)

IIRC this second stack represents an exception that has already been caught and re-thrown so it doesn't reflect where the NPE originated (which I think was due to a call to getReferenceBases that returned null). As a side note I'd like to try to narrow the exception handling and propagation in the CRAM code, especially in this file around the hasNext/nextContainer code.

droazen:
@vadimzalunin Unfortunately nothing is a quick fix in htsjdk-land :) If we're going to allow `Long`-valued attributes into `SAMRecord` we need to fully deal with the consequences of that, like ensuring we have proper bounds checking to make sure that we'll never allow values in that would overflow a `uint32_t`. And if in the process we discover that htsjdk doesn't properly support the capital-I typed attributes in the BAM spec, we should fix those problems while they have our attention.

Appreciate your patience, and hope that the requested changes are not too cumbersome.

nh13:
@eitanbanks will take a look shortly

nh13:
I do not understand the purpose of this pull request, namely what bug it supposed to fix.  We had previously agreed that htsjdk did not implement unsigned integers (see ```BinaryTagCodec```) or support the full range of values for ```i``` or ```I```` (see https://github.com/samtools/hts-specs/pull/50).  Could someone summarize what is going on here?

droazen:
@nh13 See the bug report in https://github.com/samtools/htsjdk/issues/322 as well as the following thread on the GATK support forum for background: http://gatkforums.broadinstitute.org/discussion/comment/24038/#Comment_24038

I think with the simple changes outlined above we could bring htsjdk into compliance with the BAM spec and support the full range of values for `I`, and am not sure I see a good argument against making this change. It's certainly surprised several GATK/htsjdk users to find that htsjdk is not fully compliant with the spec on this point.

src/java/htsjdk/samtools/CRAMFileWriter.java
```diff
@@ -379,6 +379,8 @@ protected void finish() {
             outputStream.flush();
             if (indexer != null)
                 indexer.finish();
+        }catch (final RuntimeException re) {
```
nh13: ```} catch```

nh13:
@droazen thanks for the context as it wasn't clear here, it also is timely as htsjdk states that java 7 will be supported only until October 2015, so we should feel free to move to java 8 if we need specific features there.

There seem to be a lot of Cram changes here, are those also necessary given @droazen suggestions, which seem reasonable?

droazen:
@vadimzalunin Ping me once you've made the requested changes and this is ready for final review.

vadimzalunin:
@droazen please review

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1159,6 +1161,32 @@ public Object getAttribute(final short tag) {
     }
 
     /**
+     * A convinience method that will return a valid unsigned integer or fail with an exception if the tag value is invalid.
```
ronlevine: convenience -> convenience

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1159,6 +1161,32 @@ public Object getAttribute(final short tag) {
...
+     * @param tag an unsigned integer represented as a long value
+     * @return valid unsigned integer associated with the tag
+     * @throws {@link htsjdk.samtools.SAMException} if the value is out of range
```
ronlevine: or not a long

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1206,24 +1234,51 @@ protected void setAttribute(final short tag, final Object value) {
...
+        }
+
+        // a special case of unsigned integers:
+        if (value instanceof Long) {
+            return isValidUnsignedInteger((Long) value)
+                    || ((Long) value >= Integer.MIN_VALUE && (Long) value <= Integer.MAX_VALUE);
         }
+        return false;
+    }
+
```
ronlevine: Add javadoc

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1206,24 +1234,51 @@ protected void setAttribute(final short tag, final Object value) {
...
+                value instanceof byte[] || value instanceof short[] || value instanceof int[] ||
+                value instanceof float[]) {
+            return true;
+        }
+
+        // a special case of unsigned integers:
+        if (value instanceof Long) {
+            return isValidUnsignedInteger((Long) value)
+                    || ((Long) value >= Integer.MIN_VALUE && (Long) value <= Integer.MAX_VALUE);
         }
```
ronlevine: Shouldn't a similar check be implemented for Byte, Short, Integer, etc? This isn't a special case, just range check.

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1206,24 +1234,51 @@ protected void setAttribute(final short tag, final Object value) {
...
+        return false;
+    }
+
+    protected void setAttribute(final short tag, final Object value, final boolean isUnsignedArray) {
         if (value == null) {
-            if (this.mAttributes != null) this.mAttributes = this.mAttributes.remove(tag);
+            // setting a tag value to null removes the tag:
```
ronlevine: No need for colon at the end of the comment.

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1206,24 +1234,51 @@ protected void setAttribute(final short tag, final Object value) {
...
+            if (this.mAttributes != null) {
+                this.mAttributes = this.mAttributes.remove(tag);
+            }
+            return;
         }
```
ronlevine: Why was the `if/else` removed in favor of the `return`?

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1159,6 +1161,32 @@ public Object getAttribute(final short tag) {
     }
 
     /**
+     * A convinience method that will return a valid unsigned integer or fail with an exception if the tag value is invalid.
+     * @param tag an unsigned integer represented as a long value
```
ronlevine: `tag` documentation should be the same as setAttribute()

src/java/htsjdk/samtools/SAMRecord.java
droazen: Long is very much a special case, since the bam spec limits us to values that can fit in a `int32_t` or `uint32_t`. Bytes/Short/Integers can fit, but not necessarily Longs.

@ronlevine Have to ask that you abort this new review -- there were a specific set of changes requested in the last review, and if they have been met we should merge this as-is. I am away today on account of the holiday, but will be able to take a look late tonight.

drozen: this is being sent to the wrong email address, sorry.

On Wed, Nov 11, 2015 at 10:53 AM, droazen <notifications@github.com> wrote:

> In src/java/htsjdk/samtools/SAMRecord.java
> <https://github.com/samtools/htsjdk/pull/339#discussion_r44547566>:
>
> > +     *
> > +     * @param value the value to be checked
> > +     * @return true if the value is valid and false otherwise
> > +     */
> > +    public static final boolean isAllowedAttributeValue(final Object value) {
> > +        if (value instanceof Byte || value instanceof Short || value instanceof Integer ||
> > +                value instanceof String || value instanceof Character || value instanceof Float ||
> > +                value instanceof byte[] || value instanceof short[] || value instanceof int[] ||
> > +                value instanceof float[]) {
> > +            return true;
> > +        }
> > +
> > +        // a special case of unsigned integers:
> > +        if (value instanceof Long) {
> > +            return isValidUnsignedInteger((Long) value)
> > +                    || ((Long) value >= Integer.MIN_VALUE && (Long) value <= Integer.MAX_VALUE);
> >          }
>
> Long is very much a special case, since the bam spec limits us to values
> that can fit in a int32_t or uint32_t. Bytes/Short/Integers can fit, but
> not necessarily Longs.
>
> @ronlevine <https://github.com/ronlevine> Have to ask that you abort this
> new review -- there were a specific set of changes requested in the last
> review, and if they have been met we should merge this as-is. I am away
> today on account of the holiday, but will be able to take a look late
> tonight.
>
> —
> <https://github.com/samtools/htsjdk/pull/339/files#r44547566>.
>



-- 
Daniel Rozen

ronlevine: @droazen Will do. It was hard to untangle all the specific changes from the last review since the they were rebased into one commit.

ronlevine: @droazen If there are any changes needed, it would be nice to include in the comment why `Long` is a special case.

vadimzalunin:
Sorry guys, I appreciate this has turned into a large patch. I tried to cover critical points with tests, in particular see src/tests/java/htsjdk/samtools/SAMIntegerTagTest.java
One issue is that SAM is different from BAM and CRAM. 
Another issue is ambiguity in type detection, for example -1 may be a valid signed or invalid unsigned.

vadimzalunin:
@droazen ping

droazen:
@vadimzalunin Am going to take a look now -- while you're waiting, perhaps you could take a look at this serious-sounding CRAM bug we discovered recently? 

https://github.com/samtools/htsjdk/issues/364 

It sounds like the kind of thing that might be a quick fix (ie., a simple off-by-one error)

droazen:
Looks like there are a few additional changes I'd like to see here -- since we're pressed for time, I'll make these changes myself and push them directly into this branch to spare us another code review cycle. While I do this if you could look at https://github.com/samtools/htsjdk/issues/364 it would be much appreciated, as we can't really recommend that anyone use GATK 3.5 with cram for anything other than testing purposes if we have bugs like that open.

droazen:
Ok, I've finished making changes to this branch -- a summary of the notable changes:

-`BinaryTagCodec` and `ReadTag` return an int for `I`-typed attributes when possible, only returning a long for values > `Integer.MAX_VALUE` and <  `BinaryCodec.MAX_UINT`. This is slightly more conservative and less likely to break client code.

-Removed some unrelated cram changes in `CRAMIterator` (allowing null `ReferenceSource`) that conflicted with changes in https://github.com/samtools/htsjdk/pull/368

-Added an overload of `getUnsignedIntegerAttribute()` that takes a `String` (with tests).

-Moved `isValidUnsignedInteger()` to `SAMUtils`

-Improved error messages, documentation, and naming of test cases.

Will rebase and merge now, once tests pass.

vadimzalunin:
@droazen thanks! looks good to me :+1:

PR #342, opened 2015-10-05
Added a pair of methods to manipulate the default options for...

tfenne:
VariantContextWriterBuilder so that it is possible to programmatically set the defaults for all future instances.

### DIFF
--- a/src/java/htsjdk/variant/vcf/VCFRecordCodec.java
+++ b/src/java/htsjdk/variant/vcf/VCFRecordCodec.java
@@ -16,20 +16,20 @@ import java.io.PrintStream;
  * with SortingCollection ONLY.
  */
 public class VCFRecordCodec implements SortingCollection.Codec<VariantContext> {
-
-	final VCFCodec vcfDecoder = new VCFCodec();
-
-	final VCFEncoder vcfEncoder;
-
+	private final VCFCodec vcfDecoder = new VCFCodec();
+	private final VCFEncoder vcfEncoder;
 	private PrintStream outputStream = null;
-
 	private BufferedReader inputReader = null;
 
 	public VCFRecordCodec(final VCFHeader header) {
-		this.vcfEncoder = new VCFEncoder(header, false, false);
+		this(header, false);
+    }
+
+	public VCFRecordCodec(final VCFHeader header, final boolean allowMissingFieldsInHeader) {
+		this.vcfEncoder = new VCFEncoder(header, allowMissingFieldsInHeader, false);
 		// Explicitly set the version because it's not available in the header itself.
 		this.vcfDecoder.setVCFHeader(header, VCFHeaderVersion.VCF4_2);
-    }
+	}
 
 	@Override
 	public void setOutputStream(final OutputStream stream) {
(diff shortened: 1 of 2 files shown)


tfenne:
This is to be paired with a change to Picard that causes CommandLineProgram to set `Options.ALLOW_MISSING_FIELDS_IN_HEADER` when ValidationStringency is `LENIENT` or `SILENT` so that it is possible to process (and fix) VCFs with incomplete headers more easily.

lbergelson:
This seems dangerous to mutate the default options of something statically. I can imagine this causing confusion  in tests and multithreaded programs.

tfenne:
@lbergelson Why do you think it's dangerous?  I'm open to alternative suggestions.  This is the exact same way we allow CLPs to set default options for SAM/BAM writing (e.g. for CREATE_INDEX).

lbergelson:
It seems like it could introduce confusing ordering issues / race conditions if static state is being modified.  We've been hitting issues with mutable statics where we didn't expect them in hellbender, because spark/dataflow run multiple jobs in the same jvm.

lbergelson:
I hadn't seen `setDefaultCreateIndexWhileWriting()` before.  That seems a bit scary to me as well.  In practice this may not be a big problem since most code won't call these methods, but it seems like if it was used more we would end up seeing errors.  I.e.  I suspect the reason that we haven't seen tests failures due to editing static options is because we don't have any tests that set them.

lbergelson:
Instead of changing these statically is it not possible to just pass in an options parameter when you create the writer?

tfenne:
Well, my goal is to provide a mechanism to change the defaults - so that it can be done in one place centrally (e.g. in Picard's CommandLineProgram) and the same mechanism then work for all VCFWriters in all programs.  What you suggest would require modifying every program or class that uses VariantContextWriterBuilder.

lbergelson:
Could you add it to `Defaults.java` and read it from the config file?  Or is it important for your use case that you be able to change it repeatedly in the same JVM?

src/java/htsjdk/variant/vcf/VCFRecordCodec.java
```diff
@@ -16,20 +16,20 @@
...
 
 	public VCFRecordCodec(final VCFHeader header) {
-		this.vcfEncoder = new VCFEncoder(header, false, false);
+		this(header, false);
+    }
+
+	public VCFRecordCodec(final VCFHeader header, final boolean allowMissingFieldsInHeader) {
+		this.vcfEncoder = new VCFEncoder(header, allowMissingFieldsInHeader, false);
 		// Explicitly set the version because it's not available in the header itself.
 		this.vcfDecoder.setVCFHeader(header, VCFHeaderVersion.VCF4_2);
```
yfarjoun: Should this always point to the newest version somehow? It seems ungood to add yet another place where VCF4_2 is hardcoded.

yfarjoun:
I agree with @lbergelson that this can be problematic when the same JVM is used for different tools at the same time. The different tools might set different defaults and thereby cause problems that are hard to debug. 
However, given that this is a problem that needs to be taken care of for SAMWriter as well, the solution there (when it is designed and implemented) would transfer to here. 

All this is to say that I give this my :+1: pending my comment.

tfenne:
Thanks @yfarjoun - I understand the concerns about multiple tools running in the same JVM.  I'm not sure exactly what environment you guys are running into this with, but I would expect that only container-level tools would ever call this code (i.e. command line programs that effectively own the JVM they're in, configuration code in a server app, etc.).

If you end up with a situation where there are multiple tools running in the same JVM, that might come into conflict, the usual solution would be to have the container use independent classloaders with child-first loading, so that each program that expects it's own environment actually gets it's own copy of the classes and libraries it loads.

PR #343, opened 2015-10-05
Fix bugs in .crai handling.

cmnbroad:
- Fix CRAIEntry compareTo implementation (fixes failures due to improper sorting of CRAIEntry lists in CRAIIndex when processing slices).
- Unify AbstractBAMFileIndex constructors to check for the BAM magic number so that CRAMFileReader properly falls back to openCraiFileAsBaiStream when processing a CRAM file with a CRAI index.
- Throw when attempting to create a CRAM index iterator when there is no accompanying reference (fixes NPE).


vadimzalunin:
:+1:

src/tests/java/htsjdk/samtools/SAMFileReaderTest.java
```diff
@@ -56,9 +57,10 @@ public void variousFormatReaderTest(final String inputFile) {
     // tests for CRAM indexing
 
     @Test(dataProvider = "SmallCRAMTest")
-    public void CRAMIndexTest(final String inputFile) {
+    public void CRAMIndexTest(final String inputFile, final String referenceFile) {
         final File input = new File(TEST_DATA_DIR, inputFile);
-        final SamReader reader = SamReaderFactory.makeDefault().open(input);
+        final File reference = new File(TEST_DATA_DIR, referenceFile);
+        final SamReader reader = SamReaderFactory.makeDefault().referenceSequence(reference).open(input);
```
droazen: This test case should take as input both a .bai-indexed cram and a .crai-indexed cram (to ensure that `hasIndex()` returns true for both). Would also be good (if possible) to test here that the index is actually usable, since the bug you fixed manifested itself as "`hasIndex()` returns true, but queries on the index return no records".

droazen:
Looks good -- one request to address, then this should be ready to merge. Back to @cmnbroad

droazen:
:+1: looks good -- merging!

PR #344, opened 2015-10-08
Yf allow only compatible dictionaries...

yfarjoun:
...in new assertion signature.
- keeps old behavior the same
- includes two tests (positive and negative)

### DIFF
--- a/src/java/htsjdk/samtools/util/SequenceUtil.java
+++ b/src/java/htsjdk/samtools/util/SequenceUtil.java
@@ -103,21 +103,43 @@ public class SequenceUtil {
     }
 
     /**
-     * Throws an exception only if both parameters are not null
+     * default signature that forces the lists to be the same size
      *
      * @param s1 a list of sequence headers
      * @param s2 a second list of sequence headers
      */
     public static void assertSequenceListsEqual(final List<SAMSequenceRecord> s1, final List<SAMSequenceRecord> s2) {
+        assertSequenceListsEqual(s1, s2, false);
+    }
+    /**
+     * Throws an exception only if both (first) parameters are not null
+     * optionally check that one list is a (nonempty) prefix of the other.
+     *
+     * @param s1 a list of sequence headers
+     * @param s2 a second list of sequence headers
+     * @param checkPrefixOnly a flag specifying whether to only look at the first records in the lists. This will then check that the
+     * records of the smaller dictionary are equal to the records of the beginning of the larger dictionary, which can be useful since
+     * sometimes different pipelines choose to use only the first contigs of a standard reference.
+     */
+    public static void assertSequenceListsEqual(final List<SAMSequenceRecord> s1, final List<SAMSequenceRecord> s2, final boolean checkPrefixOnly) {
         if (s1 != null && s2 != null) {
 
-            if (s1.size() != s2.size()) {
-                throw new SequenceListsDifferException(
-                        "Sequence dictionaries are not the same size (" + s1.size() + ", " + s2.size() +
-                                ")");
-            }
+            final int sizeToTest;
 
-            for (int i = 0; i < s1.size(); ++i) {
+            if (checkPrefixOnly) {
+                sizeToTest = Math.min(s1.size(), s2.size());
+                if (sizeToTest == 0) {
+                    throw new SequenceListsDifferException("Neither of the dictionaries can be empty.");
+                }
+            } else {
+                sizeToTest = s1.size();
+                if (s1.size() != s2.size()) {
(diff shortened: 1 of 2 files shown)


src/java/htsjdk/samtools/util/SequenceUtil.java
```diff
@@ -103,21 +103,40 @@ public static double calculateGc(final byte[] bases) {
...
+    }
+        /**
+         * Throws an exception only if both parameters are not null
+         * allows to only check that one list is a prefix of the other.
+         *
+         * @param s1 a list of sequence headers
+         * @param s2 a second list of sequence headers
+         * @param checkPrefixOnly a flag specifying whether to only look at the first records in the lists
+         */
+    public static void assertSequenceListsEqual(final List<SAMSequenceRecord> s1, final List<SAMSequenceRecord> s2, boolean checkPrefixOnly) {
```
eitanbanks: missing 'final' keyword

src/java/htsjdk/samtools/util/SequenceUtil.java
```diff
@@ -103,21 +103,40 @@ public static double calculateGc(final byte[] bases) {
...
+     * default signature that forces the lists to be the same size
      * @param s1 a list of sequence headers
      * @param s2 a second list of sequence headers
      */
     public static void assertSequenceListsEqual(final List<SAMSequenceRecord> s1, final List<SAMSequenceRecord> s2) {
+        assertSequenceListsEqual(s1, s2, false);
+    }
+        /**
+         * Throws an exception only if both parameters are not null
+         * allows to only check that one list is a prefix of the other.
```
eitanbanks: There are 3 params.
Also, please clean up the grammar here.

src/java/htsjdk/samtools/util/SequenceUtil.java
```diff
@@ -176,8 +195,15 @@ public static boolean areSequenceDictionariesEqual(final SAMSequenceDictionary s
      * Throws an exception if both parameters are non-null and unequal.
      */
     public static void assertSequenceDictionariesEqual(final SAMSequenceDictionary s1, final SAMSequenceDictionary s2) {
+        assertSequenceDictionariesEqual(s1, s2, false);
+    }
+
+    /**
+     * Throws an exception if both parameters are non-null and unequal (if checkPrefixOnly, checks prefex of lists only).
+     */
+    public static void assertSequenceDictionariesEqual(final SAMSequenceDictionary s1, final SAMSequenceDictionary s2, boolean checkPrefixOnly) {
```
eitanbanks: 'final'

src/java/htsjdk/samtools/util/SequenceUtil.java
```diff
@@ -103,21 +103,40 @@ public static double calculateGc(final byte[] bases) {
...
-                        "Sequence dictionaries are not the same size (" + s1.size() + ", " + s2.size() +
-                                ")");
-            }
+            final int sizeToTest;
 
-            for (int i = 0; i < s1.size(); ++i) {
+            if (checkPrefixOnly) {
+                sizeToTest = Math.min(s1.size(), s2.size());
+                if (sizeToTest == 0) {
+                    throw new SequenceListsDifferException("At least one of the dictionaries must be not empty.");
```
eitanbanks: I assume that the intended behavior is to allow the 2 dictionaries to be different sizes.  But I have 2 recommendations:
1. Please add a comment somewhere useful explaining that this is the behavior (i.e. that if checkPrefixOnly is true we look only at the first record AND we don't check that the dictionaries contain the same number of entries) and an example of when a user might want that behavior.
2. Is it worth having sizeToTest be an input parameter?  That would allow someone to choose how many of the contigs to check against?  Not critical, just a suggestion.

yfarjoun: 1. OK.
2. no. :-)

akiezun:
@droazen recently did some work on dict compatibility in gatk. We need his input here

yfarjoun:
back to you @eitanbanks

src/java/htsjdk/samtools/util/SequenceUtil.java
eitanbanks: Sorry, I missed one before.  The error message here should be:
"Neither of the dictionaries can be empty."

eitanbanks:
One last comment and then I'm happy.  But no merging until @droazen okays it.

droazen:
@yfarjoun & I discussed this PR in person and I gave my :+1: to the idea, but will take a look at the actual code now.

droazen:
Looks good to me -- merging!

PR #345, opened 2015-10-08
Add OverclippedReadFilter and new tags

mattsooknah:
* Copy OverclippedReadFilter from GATK, so we can use it for detecting cross-species contamination.
* Add the ability to un-map a read but preserve original mapping info in the tags.
* Add two new tags - FT (read filter reason) and OF (original flags), which I will petition to get added to the SAM spec, per request from @yfarjoun

### DIFF
--- a/src/java/htsjdk/samtools/SAMUtils.java
+++ b/src/java/htsjdk/samtools/SAMUtils.java
@@ -546,6 +546,14 @@ public final class SAMUtils {
         }
     }
 
+    /**
+     * Strip mapping information from a SAMRecord.
+     *
+     * WARNING: by clearing the secondary and supplementary flags,
+     * this may have the affect of producing multiple distinct records with the
+     * same read name and flags, which may lead to invalid SAM/BAM output.
+     * Callers of this method should make sure to deal with this issue.
+     */
     public static void makeReadUnmapped(final SAMRecord rec) {
         if (rec.getReadNegativeStrandFlag()) {
             SAMRecordUtil.reverseComplement(rec);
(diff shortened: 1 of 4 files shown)


src/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -558,10 +561,20 @@ public static void makeReadUnmapped(final SAMRecord rec) {
...
+        rec.setSupplementaryAlignmentFlag(false);
         rec.setProperPairFlag(false);
         rec.setReadUnmappedFlag(true);
     }
 
+    /**
+     * Strip mapping information from a SAMRecord, but preserve it in the 'O' tags.
+     */
+    public static void makeReadUnmappedWithOriginalTags(final SAMRecord rec) {
+        rec.setAttribute(SAMTag.OP.name(), rec.getAlignmentStart());
```
yfarjoun: You should only overwrite the 'O' tags if they aren't already set. this will keep their status as "originalX" rather than "previousX"...

src/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -558,10 +561,20 @@ public static void makeReadUnmapped(final SAMRecord rec) {
         rec.setMappingQuality(SAMRecord.NO_MAPPING_QUALITY);
         rec.setInferredInsertSize(0);
         rec.setNotPrimaryAlignmentFlag(false);
+        rec.setSupplementaryAlignmentFlag(false);
```
yfarjoun: This seems wrong. it would lead to out-of-spec bamfiles that have two primary alignments with the same name. the previous line seems similarly wrong (I know you didn't change it). I'll open an issue and see what people think.

src/java/htsjdk/samtools/filter/OverclippedReadFilter.java
```diff
@@ -0,0 +1,53 @@
+package htsjdk.samtools.filter;
```
yfarjoun: license

src/tests/java/htsjdk/samtools/filter/OverclippedReadFilterTest.java
```diff
@@ -0,0 +1,60 @@
+package htsjdk.samtools.filter;
```
yfarjoun: license

yfarjoun:
back to you @mattsooknah

mattsooknah:
@yfarjoun please re-review

yfarjoun:
:+1:

src/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -558,10 +566,31 @@ public static void makeReadUnmapped(final SAMRecord rec) {
...
         rec.setProperPairFlag(false);
         rec.setReadUnmappedFlag(true);
     }
 
+    /**
+     * Strip mapping information from a SAMRecord, but preserve it in the 'O' tags if it isn't already set.
+     */
+    public static void makeReadUnmappedWithOriginalTags(final SAMRecord rec) {
+        if (!hasOriginalMappingInformation(rec)) {
+            rec.setAttribute(SAMTag.OP.name(), rec.getAlignmentStart());
```
yfarjoun: while this is already merged, we need to store the original Seqence index or name....right?

yfarjoun:
so, talking to Bob, we have been working too hard it seems..we might be able to simply unset bit 0x4 and call it a day (without changing anything else.)

Issue #347, opened 2015-10-10
Histogram caches computed mean, but does not recompute after all changes


tfenne:
`Histogram`'s `getMean()` method caches the result in an instance variable.  An attempt is made to nullify the cached value when the histogram is mutated, but the attempt is not completely successful.  Since `Histogram` extends `TreeMap`, one can call methods like `remove()`, `clear()` etc. and after these calls the result of calling `getMean()` is incorrect.  It looks like even the `trimBy..()` methods use `remove()` but don't reset the value of `mean`.

I'm not sure why the mean is singled out for caching, whereas other values like the median are not.  My preference would be to just remove the caching of the mean value altogether.  If others agree, I'll submit a PR.

tfenne:
@nh13 @yfarjoun What do you think?

nh13:
I am trying to figure out why its even there.  Like you said, other values are not cached.  I thought for second that perhaps because there is a copy constructor, but that isn't used anywhere in Picard for example.  On the other hand, ```getStandardDeviation``` needs to use ```getMean``` so we would be computing the mean twice (see ```insertSizeMetricsCollector```).  Seems like a pre-mature optimization, since we don't cache the median and we do have a method ```getMedianAbsoluteDeviation``` that uses the ```getMedian```.  Since it is not clear that caching is needed, I err on the side of having correctness, so I would argue for removing the caching.

yfarjoun:
I looked in our usages of Histogram.mean() and I only see places where the
current implementation may cause problems if someone add a call to mean
earlier in the code. I do not see anywhere that it would actually improve
runtime.

I agree that this should either be done right (by reimplementing all the
TreeMap functions to invalidate the cache, and also all the other member
functions that change the mean value) or as you suggest, remove the
optimization.

Given that I see no need for the optimization, I suppose removing the cache.

Yossi.

davidbenjamin:
Closed by #348.

PR #348, opened 2015-10-13
Removed the broken caching of getMean() and provided a slightly more …

tfenne:
…efficient implementation of getMean() that didn't require traversing all the bins twice.

### DIFF
--- a/src/java/htsjdk/samtools/util/Histogram.java
+++ b/src/java/htsjdk/samtools/util/Histogram.java
@@ -45,7 +45,6 @@ import static java.lang.Math.*;
 public class Histogram<K extends Comparable> extends TreeMap<K, Bin> {
     private String binLabel   = "BIN";
     private String valueLabel = "VALUE";
-    private Double mean;
 
     /** Constructs a new Histogram with default bin and value labels. */
     public Histogram() { }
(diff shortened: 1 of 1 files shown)


tfenne:
Solution for issue #347

src/java/htsjdk/samtools/util/Histogram.java
```diff
@@ -164,12 +161,21 @@ public boolean equals(final Object o) {
...
     }
 
+    /**
+     * Assuming that the key type for the histogram is a Number type, returns the mean of
+     * all the items added to the histogram.
+     */
     public double getMean() {
-        if (mean == null) {
-            mean = getSum() / getCount();
+        double product=0, totalCount=0;
```
yfarjoun: spaces

yfarjoun:
looks  fine to me. Could you add a comment explaining why you aren't using getSum and getCounts? also, the test uses round(), could you change that to use the (newish) double comparator that takes a precision?

tfenne:
@yfarjoun: added the requested comment.  I couldn't find the comparator you mentioned - the only thing I could find was `TestNGUtil.compareDoubleWithAccuracy()` but that's up in picard and not available here.

yfarjoun:
:100:

Issue #349, opened 2015-10-13
CRAIEntry overflow


superbobry:
I have a 25G CRAM file indexed by [`cramtools`](https://github.com/enasequence/cramtools). Reading it with `htsjdk` leads to an exception in `CRAIEntry`:

```java
Exception in thread "main" htsjdk.samtools.cram.CRAIIndex$CRAIIndexException: java.lang.NumberFormatException: For input string: "2147560457"
	at htsjdk.samtools.cram.CRAIEntry.fromCraiLine(CRAIEntry.java:60)
	at htsjdk.samtools.cram.CRAIIndex.readIndex(CRAIIndex.java:43)
	at htsjdk.samtools.cram.CRAIIndex.openCraiFileAsBaiStream(CRAIIndex.java:128)
	at htsjdk.samtools.CRAMFileReader.<init>(CRAMFileReader.java:138)
	at htsjdk.samtools.CRAMFileReader.<init>(CRAMFileReader.java:147)
	at htsjdk.samtools.CRAMFileReader.<init>(CRAMFileReader.java:153)
	at htsjdk.samtools.SamReaderFactory$SamReaderFactoryImpl.open(SamReaderFactory.java:270)
	at htsjdk.samtools.SamReaderFactory$SamReaderFactoryImpl.open(SamReaderFactory.java:133)
```

The exception is caused by the container start byte offset being too large to fit into a signed int. Section "12 Indexing" of the CRAM [spec](http://samtools.github.io/hts-specs/CRAMv3.pdf) doesn't mention any restrictions on this field.

Here's a relevant chunk of the CRAI index:
```
1       102792674       29413   2147560457      1001    261258
1       102821988       22969   2147822745      782     254201
1       102844858       22856   2148077757      743     262815
1       102867614       20780   2148341344      965     251332
1       102888289       20388   2148593670      960     257108
```

PR #350, opened 2015-10-13
Fixed overflow in 'CRAIEntry.fromCraiLine'

superbobry:
See #349 for details.


lbergelson:
@superbobry Thank you for the for catching this error and for contributing a patch.

PR #353, opened 2015-10-19
Make cramFile, inputStream and iterator variables in CRAMFileReader final

pityka:
CRAMFileReader.getFileHeader() forwards to iterator, therefore the same CRAMFileReader instance could return different references from getFileHeader() during the lifetime of the instance.
This will make NPE in SamFileHeaderMerger and MergingSamRecordIterator because they rely on reference equality in IdentityHashMap.

The original version assigned a new CRAMIterator to the `iterator` instance variable in CRAMFileReader each time getIterator() was called and `cramFile` was not null. As a consequence after each call of getIterator() the reference returned by getFileHeader() was different.

I changed the `iterator`, `cramFile` and `inputStream` instance variables to final to ensure that they don't get reassigned. I adapted the constructors to accommodate this change.

### DIFF
--- a/src/java/htsjdk/samtools/CRAMFileReader.java
+++ b/src/java/htsjdk/samtools/CRAMFileReader.java
@@ -44,10 +44,10 @@ import java.util.Iterator;
  */
 @SuppressWarnings("UnusedDeclaration")
 public class CRAMFileReader extends SamReader.ReaderImplementation implements SamReader.Indexing {
-    private File cramFile;
+    private final File cramFile;
     private final ReferenceSource referenceSource;
-    private InputStream inputStream;
-    private CRAMIterator iterator;
+    private final InputStream inputStream;
+    private final CRAMIterator iterator;
     private BAMIndex mIndex;
     private File mIndexFile;
     private boolean mEnableIndexCaching;
@@ -64,7 +64,7 @@ public class CRAMFileReader extends SamReader.ReaderImplementation implements Sa
      * @param cramFile CRAM file to open
      * @param inputStream   CRAM stream to read
      */
-    public CRAMFileReader(final File cramFile, final InputStream inputStream) {
+    public CRAMFileReader(final File cramFile, final InputStream inputStream) throws FileNotFoundException, IOException {
         this(cramFile, inputStream, new ReferenceSource(Defaults.REFERENCE_FASTA));
     }
 
@@ -78,15 +78,8 @@ public class CRAMFileReader extends SamReader.ReaderImplementation implements Sa
      *                        reference sequences
      */
     public CRAMFileReader(final File cramFile, final InputStream inputStream,
-                          final ReferenceSource referenceSource) {
-        if (cramFile == null && inputStream == null)
-            throw new IllegalArgumentException(
-                    "Either file or input stream is required.");
-
-        this.cramFile = cramFile;
-        this.inputStream = inputStream;
-        this.referenceSource = referenceSource;
-        getIterator();
+                          final ReferenceSource referenceSource) throws FileNotFoundException, IOException {        
+        this(cramFile != null ? new FileInputStream(cramFile) : inputStream,cramFile, (SeekableStream)null, referenceSource, (ValidationStringency)null);       
     }
 
     /**
@@ -99,36 +92,43 @@ public class CRAMFileReader extends SamReader.ReaderImplementation implements Sa
      *                        reference sequences
      */
     public CRAMFileReader(final File cramFile, final File indexFile,
-                          final ReferenceSource referenceSource) {
-        if (cramFile == null)
-            throw new IllegalArgumentException("File is required.");
-
-        this.cramFile = cramFile;
-        this.mIndexFile = indexFile;
-        this.referenceSource = referenceSource;
-
-        getIterator();
+                          final ReferenceSource referenceSource) throws FileNotFoundException,IOException {
+        this(new FileInputStream(cramFile),cramFile, new SeekableFileStream(indexFile), referenceSource, null);
     }
 
-    public CRAMFileReader(final File cramFile, final ReferenceSource referenceSource) {
-        if (cramFile == null && inputStream == null)
-            throw new IllegalArgumentException(
-                    "Either file or input stream is required.");
-
-        this.cramFile = cramFile;
-        this.referenceSource = referenceSource;
-
-        getIterator();
+    public CRAMFileReader(final File cramFile, final ReferenceSource referenceSource) throws FileNotFoundException,IOException {
+        this(new FileInputStream(cramFile),cramFile, (SeekableStream)null, referenceSource, null);
     }
 
     public CRAMFileReader(final InputStream inputStream, final SeekableStream indexInputStream,
                           final ReferenceSource referenceSource, final ValidationStringency validationStringency) throws IOException {
-        this.inputStream = inputStream;
-        this.referenceSource = referenceSource;
-        this.validationStringency = validationStringency;
+        this(inputStream,(File)null,indexInputStream,referenceSource,validationStringency);
+    }
 
-        iterator = new CRAMIterator(inputStream, referenceSource);
-        iterator.setValidationStringency(validationStringency);
+    CRAMFileReader(final InputStream inputStream,final File cramFile, final SeekableStream indexInputStream,
+                          final ReferenceSource referenceSource, final ValidationStringency validationStringency) throws IOException {
+        this.cramFile = cramFile;
+
+        if (inputStream != null) {
+        this.inputStream = inputStream;  
+        } else {
+            if (cramFile != null) {
+                this.inputStream = new FileInputStream(cramFile);            
+            } else {
+                throw new RuntimeException("inputStream or cramFile must not be null");
+            }
+        }
+
+        this.referenceSource = referenceSource;
+
+        iterator = new CRAMIterator(this.inputStream, referenceSource);
+
+        if (validationStringency != null) {
+            this.validationStringency = validationStringency;
+            
+        }   
+        iterator.setValidationStringency(validationStringency);    
+             
         if (indexInputStream != null) {
             try {
                 mIndex = new CachingBAMFileIndex(indexInputStream, iterator.getSAMFileHeader().getSequenceDictionary());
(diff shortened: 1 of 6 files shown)


droazen:
@pityka Could you please rebase this branch onto the latest master, which contains changes that make test failures easier to see in travis?

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -99,36 +92,43 @@ public CRAMFileReader(final File cramFile, final InputStream inputStream,
...
-    }
+        if (inputStream != null) {
+        this.inputStream = inputStream;  
+        } else {
+            if (cramFile != null) {
+                this.inputStream = new FileInputStream(cramFile);            
+            } else {
+                throw new RuntimeException("inputStream or cramFile must not be null");
+            }
+        }
```
nh13: un-nest `if/else`.

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -99,36 +92,43 @@ public CRAMFileReader(final File cramFile, final InputStream inputStream,
...
 
-        getIterator();
-    }
+        if (inputStream != null) {
+        this.inputStream = inputStream;  
+        } else {
+            if (cramFile != null) {
+                this.inputStream = new FileInputStream(cramFile);            
+            } else {
+                throw new RuntimeException("inputStream or cramFile must not be null");
```
nh13: Since this a run time message, could this message be improved for the user?

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -99,36 +92,43 @@ public CRAMFileReader(final File cramFile, final InputStream inputStream,
...
 
+    CRAMFileReader(final InputStream inputStream,final File cramFile, final SeekableStream indexInputStream,
+                          final ReferenceSource referenceSource, final ValidationStringency validationStringency) throws IOException {
         this.cramFile = cramFile;
-        this.referenceSource = referenceSource;
 
-        getIterator();
-    }
+        if (inputStream != null) {
+        this.inputStream = inputStream;
```
nh13: indentation

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -99,36 +92,43 @@ public CRAMFileReader(final File cramFile, final InputStream inputStream,
...
         this.referenceSource = referenceSource;
-        this.validationStringency = validationStringency;
 
-        iterator = new CRAMIterator(inputStream, referenceSource);
-        iterator.setValidationStringency(validationStringency);
+        iterator = new CRAMIterator(this.inputStream, referenceSource);
+
+        if (validationStringency != null) {
+            this.validationStringency = validationStringency;
+
```
nh13: extra newline.

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -144,14 +144,13 @@ public CRAMFileReader(final InputStream inputStream, final SeekableStream indexI
     public CRAMFileReader(final InputStream stream,
                           final File indexFile, final ReferenceSource referenceSource,
                           final ValidationStringency validationStringency) throws IOException {
-        this(stream, indexFile == null ? null: new SeekableFileStream(indexFile), referenceSource, validationStringency);
+        this(stream,null, indexFile == null ? null: new SeekableFileStream(indexFile), referenceSource, validationStringency);
```
nh13: space before `null`

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -144,14 +144,13 @@ public CRAMFileReader(final InputStream inputStream, final SeekableStream indexI
...
     }
 
     public CRAMFileReader(final File cramFile,
                           final File indexFile, final ReferenceSource referenceSource,
                           final ValidationStringency validationStringency) throws IOException {
-        this(new FileInputStream(cramFile), indexFile, referenceSource, validationStringency);
-        this.cramFile = cramFile;
+        this(new FileInputStream(cramFile),cramFile, indexFile == null ? null : new SeekableFileStream(indexFile), referenceSource, validationStringency);
```
nh13: space before `cramFile`.

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -376,13 +359,12 @@ public SAMRecordIterator assertSorted(final SortOrder sortOrder) {
             seekableStream.seek(startOfLastLinearBin >>> 16);
             final Container container = ContainerIO.readContainerHeader(newIterator.getCramHeader().getVersion().major, seekableStream);
             seekableStream.seek(seekableStream.position() + container.containerByteSize);
-            iterator = newIterator;
-            iterator.jumpWithinContainerToPos(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX, SAMRecord.NO_ALIGNMENT_START);
+            newIterator.jumpWithinContainerToPos(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX, SAMRecord.NO_ALIGNMENT_START);
+            return newIterator;
```
nh13: Wait, wont this leave this in a bad state, namely `iterator` is no longer set correctly?

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -64,7 +64,7 @@
      * @param cramFile CRAM file to open
      * @param inputStream   CRAM stream to read
      */
-    public CRAMFileReader(final File cramFile, final InputStream inputStream) {
+    public CRAMFileReader(final File cramFile, final InputStream inputStream) throws FileNotFoundException, IOException {
```
akiezun: it's bad style to have 2 exceptions declared and one of them be a subclass of the other: `FileNotFoundException extends IOException`

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -78,15 +78,8 @@ public CRAMFileReader(final File cramFile, final InputStream inputStream) {
...
-                          final ReferenceSource referenceSource) {
-        if (cramFile == null && inputStream == null)
-            throw new IllegalArgumentException(
-                    "Either file or input stream is required.");
-
-        this.cramFile = cramFile;
-        this.inputStream = inputStream;
-        this.referenceSource = referenceSource;
-        getIterator();
+                          final ReferenceSource referenceSource) throws FileNotFoundException, IOException {
```
akiezun: same style issue "it's bad style to have 2 exceptions declared and one of them be a subclass of the other: `FileNotFoundException extends IOException`"

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -99,36 +92,43 @@ public CRAMFileReader(final File cramFile, final InputStream inputStream,
...
      */
     public CRAMFileReader(final File cramFile, final File indexFile,
-                          final ReferenceSource referenceSource) {
-        if (cramFile == null)
-            throw new IllegalArgumentException("File is required.");
-
-        this.cramFile = cramFile;
-        this.mIndexFile = indexFile;
-        this.referenceSource = referenceSource;
+                          final ReferenceSource referenceSource) throws FileNotFoundException,IOException {
```
akiezun: and here: it's bad style to have 2 exceptions declared and one of them be a subclass of the other: FileNotFoundException extends IOException

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -99,36 +92,43 @@ public CRAMFileReader(final File cramFile, final InputStream inputStream,
...
+    }
 
+    CRAMFileReader(final InputStream inputStream,final File cramFile, final SeekableStream indexInputStream,
+                          final ReferenceSource referenceSource, final ValidationStringency validationStringency) throws IOException {
         this.cramFile = cramFile;
-        this.referenceSource = referenceSource;
 
-        getIterator();
-    }
+        if (inputStream != null) {
```
akiezun: this whole nested if would be nicer as 
```
if (inputStream == null && cramFile == null) { throw }
this.inputStream = inputStream == null ? new FileInputStream(cramFile) : inputStream;
```

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -99,36 +92,43 @@ public CRAMFileReader(final File cramFile, final InputStream inputStream,
...
-                          final ReferenceSource referenceSource, final ValidationStringency validationStringency) throws IOException {
-        this.inputStream = inputStream;
         this.referenceSource = referenceSource;
-        this.validationStringency = validationStringency;
 
-        iterator = new CRAMIterator(inputStream, referenceSource);
-        iterator.setValidationStringency(validationStringency);
+        iterator = new CRAMIterator(this.inputStream, referenceSource);
+
+        if (validationStringency != null) {
```
akiezun: what does this null check achieve? If validationStringency is null then this.validationStringency will also remain null so is this a no-op?

cmnbroad:
I like the idea of this change, but I'd suggest that it get rebased before we do much more since the underlying code has change a bit and some of these changes are obsolete. I think it will need to be reviewed again anyway.

droazen:
@pityka Could you please add a test case that fails before your patch (in a separate commit that fails tests), and passes after your patch (in a separate commit that passes tests). Could you also rebase the commit history of this branch so that it only contains your own commits (it looks like you rebased an unrelated commit from master by accident).

cmnbroad:
I think we actually don't want the iterator to be final; in fact in order to fix https://github.com/samtools/htsjdk/issues/563; to make CRAMFileReader more conformant with the other readers; and to honor the contract that only allows a single extant iterator at a time, we should be updating the iterator whenever a new one is created, i.e. like [BAMFileReader does](https://github.com/samtools/htsjdk/blob/master/src/java/htsjdk/samtools/BAMFileReader.java#L388) with mCurrentIterator.

@pityka It would be really helpful to have a test case to illustrate the original problem outlined above with the header if you have one, though, since that sounds like it needs a fix. Thx.

droazen:
Closing since there's been no reply from @pityka -- @pityka please re-open if you still want this merged.

PR #356, opened 2015-10-22
Set TestNG verbosity back to 2, and reduce log file size by grepping out PASSED: lines.

droazen:
This change allows submitters to once again see which tests are failing in
their branches via travis.


droazen:
Discussed in person with @yfarjoun and @lbergelson, who gave their :+1: -- merging.

PR #357, opened 2015-10-22
Make the VCF/BCF Header tests less chatty

droazen:
The VCFHeader tests were dumping extremely large String representations of VCFHeaders
to the TestNG log due to the way they were using DataProviders -- this comprised
the vast majority of htsjdk test output. This refactors these tests to produce
much less output.

With this change, grepping out the "PASSED:" lines should no longer be necessary
(this approach didn't work well anyway, since it didn't correctly propagate the
TestNG exit status to travis).

### DIFF
--- a/src/tests/java/htsjdk/variant/bcf2/BCF2UtilsUnitTest.java
+++ b/src/tests/java/htsjdk/variant/bcf2/BCF2UtilsUnitTest.java
@@ -93,6 +93,21 @@ public final class BCF2UtilsUnitTest extends VariantBaseTest {
         Assert.assertEquals(7,dict_size);
     }
 
+    /**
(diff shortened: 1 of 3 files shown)


src/tests/java/htsjdk/variant/bcf2/BCF2UtilsUnitTest.java
```diff
@@ -93,6 +93,17 @@ public void testCreateDictionary() {
         Assert.assertEquals(7,dict_size);
     }
 
+    private static class HeaderOrderTestCase {
```
lbergelson: Could you add a comment explaining why this class exists.

lbergelson:
Looks good, just add a comment for posterity so they don't just inline that wrapper class.

PR #358, opened 2015-10-26
Throw a more informative exception if reference lookup failed

pityka:
The original version may passed a null pointer to `restoreReadBases`. This throws the same exception as in CRAMIterator.

### DIFF
--- a/src/java/htsjdk/samtools/cram/build/CramNormalizer.java
+++ b/src/java/htsjdk/samtools/cram/build/CramNormalizer.java
@@ -19,6 +19,7 @@ package htsjdk.samtools.cram.build;
 
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SAMRecord;
+import htsjdk.samtools.cram.CRAMException;
 import htsjdk.samtools.cram.encoding.readfeatures.BaseQualityScore;
 import htsjdk.samtools.cram.encoding.readfeatures.Deletion;
 import htsjdk.samtools.cram.encoding.readfeatures.InsertBase;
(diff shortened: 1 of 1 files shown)


yfarjoun:
can you add a test for this simple case?

pityka:
I have a file for which CRAMIterator throws this exception (NPE in master). I tried to isolate the read to make a test case and I noticed that `samtools view` hangs halfway upon reading the file, so that file might be simply truncated or otherwise wrong. 
Unfortunately I don't know how to create a cram file from scratch which would test this.

cmnbroad:
There is a discussion ongoing in #362 about how to tighten up the validation and error checking for the CRAM code.

src/java/htsjdk/samtools/cram/build/CramNormalizer.java
```diff
@@ -123,9 +124,15 @@ public void normalize(final ArrayList<CramCompressionRecord> records,
...
                 record.readBases = SAMRecord.NULL_SEQUENCE;
-            } else
+            } else {
+                if (refBases != null) {
                 record.readBases = restoreReadBases(record, refBases, refOffset_zeroBased,
                         substitutionMatrix);
+                } else {
+                     throw new CRAMException(String.format("Contig %s not found in the reference file.", header.getSequence(record.sequenceId).toString()));
+                }
+            }
```
nh13: why not just `else if { ...} else { ... }` instead of nesting an `if/else` inside an `else`?

nh13: Also, formatting.

akiezun:
agreed, needs a test. looks fine otherwise

vadimzalunin:
Seems ok to me. 
@pityka could you please add a test? A fake record with an empty ref source should do it.

droazen:
Since @pityka has not responded, @vadimzalunin has agreed to incorporate this fix into an upcoming branch, with a test.

PR #359, opened 2015-10-26
Changed to only interrupt the writer thread if the queue size is...

tfenne:
already zero, implying that the writer will be waiting on the queue.  Resolves a nasty threading bug where on close() the writer could recieve an InterruptedException deep inside blocking I/O code and exit.  This would cause an exception to be thrown in AbstractAsyncWriter for "Queue should be empty but is size: #".

### DIFF
--- a/src/java/htsjdk/samtools/util/AbstractAsyncWriter.java
+++ b/src/java/htsjdk/samtools/util/AbstractAsyncWriter.java
@@ -64,7 +64,7 @@ public abstract class AbstractAsyncWriter<T> implements Closeable {
 
         if (!this.isClosed.getAndSet(true)) {
             try {
-            	this.writer.interrupt(); // signal to writer clean up
+                if (this.queue.isEmpty()) this.writer.interrupt(); // signal to writer clean up
             	this.writer.join();
             } catch (final InterruptedException ie) {
             	throw new RuntimeException("Interrupted waiting on writer thread.", ie);


tfenne:
@nh13 mind taking a quick look at this one?  The actual change in tiny.

tfenne:
@d-cameron You might also want to take a look at this change.  I was seeing occasional (rare) instances of the exception being thrown for the queue not being empty in the close() method.  I tracked it down to the addition of the interrupt() call that could be interrupting IO as well as the queue.poll().  I think the change made will still allow for a faster close() but is also safer.

src/java/htsjdk/samtools/util/AbstractAsyncWriter.java
```diff
@@ -64,7 +64,7 @@ public void close() {
 
         if (!this.isClosed.getAndSet(true)) {
             try {
-            	this.writer.interrupt(); // signal to writer clean up
+                if (this.queue.isEmpty()) this.writer.interrupt(); // signal to writer clean up
```
nh13: extra space in front of the ```if```.  My work here is done.

nh13:
Is there any benefit to polling every 100ms to see if the queue is empty to then try to interrupt?  I don't think so, but then again, why where would the original 2s time improvement come from if the queue was empty and we were waiting on the join?

tfenne:
The original speedup came from if you called AAW.close() and the underlying queue was already empty, you could be waiting up to 2s for the `WriterRunnable` to timeout on it's `poll()` call on the queue.  So the interrupt would cause that `poll()` to terminate with an `InterruptedException`, which was fine so long as it the thread was in `poll()`.

If the queue is not yet empty then the `WriterRunnable` will notice that the queue is empty and the closed flag is set, and not drop into that 2s poll.

src/java/htsjdk/samtools/util/AbstractAsyncWriter.java
tfenne: the extra space makes four!  there were only three before!

nh13: +1: more :coffee:

PR #361, opened 2015-10-28
Added an additional getHistogram that takes in an integer index

kbergin:
for pulling out specific histograms from Metrics Files that have more than one histogram in their output like picard's CollectWGSMetrics.


src/java/htsjdk/samtools/metrics/MetricsFile.java
```diff
@@ -84,6 +84,13 @@ public void addAllMetrics(final Iterable<BEAN> beanz) {
         else return null;
     }
 
+    /** Returns the histogram contained in the metrics file if any. Adds capability of retrieving multiple histograms
+     *  from metrics file. */
+    public Histogram<HKEY> getHistogram(final int index) {
```
lbergelson: My personal preference is always to throw an exception if you're out of bounds.

kbergin: Thank you Louis - Do you mean to do just a try return statement and catch and throw an exception instead of checking in the if and returning null in the 'else' ?  

I'm also working on adding tests because Yossi pointed out I need them - there's no tests for either getHistogram function so I'm doing that. So this PR will need to wait a little for that.

.idea/modules/htsjdk.iml
```diff
@@ -39,5 +39,7 @@
         <SOURCES />
       </library>
     </orderEntry>
+    <orderEntry type="library" name="apache-ant-1.8.2-bzip2" level="application" />
+    <orderEntry type="library" name="testng-5.5-jdk15" level="application" />
```
kbergin: Ugh I didn't git add this! I'll need a reminder on how to get this file out of my commits

yfarjoun: git rm --cached --  .idea/modules/htsjdk.iml
<https://github.com/samtools/htsjdk/pull/361#discussion_r43421100>
git commit -m"removing mistakenly added .iml file"

should do it

Y.

.idea/modules/htsjdk.iml
```diff
@@ -1,43 +0,0 @@
...
-          <root url="jar://$MODULE_DIR$/../../lib/commons-logging-1.1.1.jar!/" />
-          <root url="jar://$MODULE_DIR$/../../lib/snappy-java-1.0.3-rc3.jar!/" />
-          <root url="jar://$MODULE_DIR$/../../lib/commons-jexl-2.1.1.jar!/" />
-        </CLASSES>
-        <JAVADOC />
-        <SOURCES />
-      </library>
-    </orderEntry>
-  </component>
-</module>
```
kbergin: @yfarjoun does this seem right? I tried the git commands you suggested but this seems weird.

kbergin:
@lbergelson do you mind checking this code out again? I tried to address your previous review comment and I added tests

lbergelson:
Ah, sure!  Sorry I didn't notice your earlier comment.

.idea/modules/htsjdk.iml
lbergelson: undo this deletion.  

```
git checkout origin/master .idea/modules/htsjdk.iml
git commit .idea/modules/htsjdk.iml -m "resetting htsjdk.iml"
```

src/java/htsjdk/samtools/metrics/MetricsFile.java
```diff
@@ -267,10 +274,9 @@ private void printHistogram(final BufferedWriter out, final FormatUtil formatter
 
             for (final Histogram<HKEY> histo : nonEmptyHistograms) {
                 final Histogram<HKEY>.Bin bin = histo.get(key);
-                final double value = (bin == null ? 0 : bin.getValue());
-
                 out.append(SEPARATOR);
-                out.append(formatter.format(value));
+
+                if(bin != null) out.append(formatter.format(bin.getValue()));
             }
```
kbergin: These changes were added because there was an issue I ran into with the testing format where the 'writeThenReadBack' function was not passing the tests with two differently sized histograms because this was filling in the empty rows with 0s. The original MetricsFile object didn't have these 0s in the histograms but the new file did, making them unequal and causing tests to fail.

src/java/htsjdk/samtools/metrics/MetricsFile.java
```diff
@@ -427,6 +433,7 @@ else if (line.startsWith(MINOR_HEADER_PREFIX)) {
                         final HKEY key = (HKEY) formatter.parseObject(fields[0], keyClass);
 
                         for (int i=1; i<fields.length; ++i) {
+                            if(fields[i].equals("")) continue;
                             final double value = formatter.parseDouble(fields[i]);
```
kbergin: This was added to allow the code to handle multiple differently sized histograms, for example the first one is size 5, second is size 1 and third is size 5. It needs to skip the empty spots to parse correctly.

.idea/modules/htsjdk.iml
kbergin: Then git push again?

src/java/htsjdk/samtools/metrics/MetricsFile.java
```diff
@@ -84,6 +84,13 @@ public void addAllMetrics(final Iterable<BEAN> beanz) {
         else return null;
     }
 
+    /** Returns the histogram contained in the metrics file if any. Adds capability of retrieving multiple histograms
+     *  from metrics file. */
+    public Histogram<HKEY> getHistogram(final int index) {
+        try { return this.histograms.get(index); }
```
lbergelson: I'd format this with a bit more space. 

i.e.
```
try {
  return ...
} catch () {
  throw ...
}
```

kbergin:
No problem @lbergelson ! You'll see that once I started writing tests for the multiple histograms functionality I had to edit some of the Histogram code to read the histograms in correctly.

src/java/htsjdk/samtools/metrics/MetricsFile.java
```diff
@@ -84,6 +84,13 @@ public void addAllMetrics(final Iterable<BEAN> beanz) {
...
+        catch (final ArrayIndexOutOfBoundsException excep) { throw new SAMException("Index is out of bounds: " + index); }
```
lbergelson: standard exception name is usually `e` or `ex`, good to use the standard unless there's a specific reason to change

kbergin: Thank you! didn't realize. Good to know!

lbergelson: Also, in this case the `SAMException` isn't really providing any more information (and it's hiding some...).  I would probably either just not catch this exception and let the `ArrayIndexOutOfBoundsException` propagate up, or if you think a SAMException is more appropriate, I would give it a bit more context in the message.  `"There is no histogram with index " +index+ "."

lbergelson: Might want to include the original exception in the message as well.  I think SAMException has a constructor that takes an exception as a parameter as well as the additional message.

src/java/htsjdk/samtools/metrics/MetricsFile.java
```diff
@@ -267,10 +274,9 @@ private void printHistogram(final BufferedWriter out, final FormatUtil formatter
 
             for (final Histogram<HKEY> histo : nonEmptyHistograms) {
                 final Histogram<HKEY>.Bin bin = histo.get(key);
-                final double value = (bin == null ? 0 : bin.getValue());
-
                 out.append(SEPARATOR);
-                out.append(formatter.format(value));
+
+                if(bin != null) out.append(formatter.format(bin.getValue()));
```
lbergelson: ALWAYS use `{}` with an `if` statement.

src/java/htsjdk/samtools/metrics/MetricsFile.java
kbergin: Ok so just: 

public Histogram<HKEY> getHistogram(final int index) throws ArrayIndexOutOfBoundsException {
        return this.histograms.get(index);
    }

Should I also change the old get histogram function to mimic this functionality?

src/java/htsjdk/samtools/metrics/MetricsFile.java
kbergin: Ok :)

lbergelson: Does this refactoring change the behavior?  Did there used to be 0's output that are now skipped?

src/java/htsjdk/samtools/metrics/MetricsFile.java
```diff
@@ -427,6 +433,7 @@ else if (line.startsWith(MINOR_HEADER_PREFIX)) {
                         final HKEY key = (HKEY) formatter.parseObject(fields[0], keyClass);
 
                         for (int i=1; i<fields.length; ++i) {
+                            if(fields[i].equals("")) continue;
```
lbergelson: `{}` around continue

What's this case for?

src/tests/java/htsjdk/samtools/metrics/MetricsFileTest.java
```diff
@@ -156,6 +158,45 @@ public void testWriteMetricsFile() throws IOException, ClassNotFoundException {
         histo.increment(1000, 10981982);
         file.setHistogram(histo);
 
+        //Get same histogram from file using two different methods
```
lbergelson: These tests should be moved to their own test methods.

i.e. `testGetHistogram()`, `testGetAllHistograms`, `testWritingAndReadingDifferentSizedHistograms`

src/java/htsjdk/samtools/metrics/MetricsFile.java
lbergelson: Oh... right, there's an existing method that returns nulls.. That makes sense why you were doing it that way.  I wasn't thinking about that.

src/java/htsjdk/samtools/metrics/MetricsFile.java
kbergin: You're right, it does change the functionality and now I'm not convinced this is ideal. This is the first time we're running into this because the histogram functionality wasn't tested before. It is possible that things rely on the 0s being output into the files for unequally sized histograms.

With that in mind, maybe we can look at the test format and decide the best way to test this. The current method writes and reads back the files and compares them, which fails because the 0s don't exist in the histogram but exist after the file is written out.

kbergin:
Putting this on hold until we figure out the best way to handle multiple histograms.

src/java/htsjdk/samtools/metrics/MetricsFile.java
nh13: @kbergin have you added tests?

src/java/htsjdk/samtools/metrics/MetricsFile.java
nh13: Space before the parens in `if(`

src/java/htsjdk/samtools/metrics/MetricsFile.java
nh13: 1. `if(` -> `if ()
2. can just check the length of the array.

src/tests/java/htsjdk/samtools/metrics/MetricsFileTest.java
nh13: space after `//` for each.  Makes it a lot easier to read, but I will leave it up to you.

src/tests/java/htsjdk/samtools/metrics/MetricsFileTest.java
```diff
@@ -156,6 +157,45 @@ public void testWriteMetricsFile() throws IOException, ClassNotFoundException {
...
+        Histogram<Integer> histoTest1 = file.getHistogram();
+        Histogram<Integer> histoTest2 = file.getHistogram(0);
```
nh13: finals

kbergin:
@yfarjoun @nh13 I ended up using htsjdk's existing method for getting histograms when there are multiple in a file, instead of what I did here. So I didn't end up finishing these tests.. and I didn't push for a consensus on picard bug 322 which goes along with this. 

I don't think the functionality I added here is necessary, however, I do still think that the testing could be improved for multiple Histograms. 

Maybe, with that in mind, this should be closed without merging until it can be revisited?

lbergelson:
Closed after discussion with @kbergin during #548

PR #363, opened 2015-11-03
Set original RNAME when unmapping reads

mattsooknah:
Add OR (original RNAME) tag and use it when we call ```makeReadUnmappedWithOriginalTags```.


mattsooknah:
@yfarjoun please review

yfarjoun:
:+1:

Issue #365, opened 2015-11-04
SAMFileWriterFactory.makeWriter ignores factory settings and flags when creating CRAM files


cmnbroad:
SAMFileWriterFactory.makeWriter will try to create a CRAM file if the passed in file extension is  ".cram". However, it ignores the "presorted" parameter in that case, which leaves the new CRAMFileWriter with presorted set to true. This can result in downstream exceptions if the input isn't presorted.

If we really require the input to be presorted for CRAM files, we should validate the inputs (the header and the presorted flag) and reject them right in this call.

Also, makeWriter doesn't propagate the filename to the CRAMFileWriter, so when the downstream exceptions occur, the error message contains the  string "null" for the filename.

cmnbroad:
presorted, create index, and md5 settings are all ignored by makeWriter when creating a cram file.

cmnbroad:
https://github.com/samtools/htsjdk/pull/368.

PR #368, opened 2015-11-04
Fixes for CRAM bugs.

cmnbroad:
These are proposed changes for issues discovered when adding CRAM file tests to GATK4. However, we really need to resolve the policy issue about whether/how to require references for CRAM files (partly being discussed in #339) before I can proceed with these, since some of these changes clash with changes made there).

### DIFF
--- a/src/java/htsjdk/samtools/CRAMFileWriter.java
+++ b/src/java/htsjdk/samtools/CRAMFileWriter.java
@@ -32,6 +32,7 @@ import htsjdk.samtools.cram.structure.CramCompressionRecord;
 import htsjdk.samtools.cram.structure.CramHeader;
 import htsjdk.samtools.cram.structure.Slice;
 import htsjdk.samtools.util.Log;
+import htsjdk.samtools.util.RuntimeIOException;
 import htsjdk.samtools.util.StringLineReader;
 
 import java.io.IOException;
@@ -72,22 +73,75 @@ public class CRAMFileWriter extends SAMFileWriterImpl {
     private CRAMIndexer indexer;
     private long offset;
 
-    public CRAMFileWriter(final OutputStream outputStream, final ReferenceSource source, final SAMFileHeader samFileHeader, final String fileName) {
-        this(outputStream, null, source, samFileHeader, fileName);
+    /**
+     * Create a CRAMFileWriter on an output stream. Requires input records to be presorted to match the
+     * sort order defined by the input {@code samFileHeader}.
+     *
+     * @param outputStream where to write the output.
+     * @param source reference source
+     * @param samFileHeader {@link SAMFileHeader} to be used. Sort order is determined by the sortOrder property of this arg.
+     * @param fileName used for display in error messages
+     */
+    public CRAMFileWriter(
+            final OutputStream outputStream,
+            final ReferenceSource source,
+            final SAMFileHeader samFileHeader,
+            final String fileName)
+    {
+        this(outputStream, null, source, samFileHeader, fileName); // defaults to presorted == true
     }
 
-    public CRAMFileWriter(final OutputStream outputStream, final OutputStream indexOS, final ReferenceSource source, final SAMFileHeader samFileHeader, final String fileName) {
+    /**
+     * Create a CRAMFileWriter and index on output streams. Requires input records to be presorted to match the
+     * sort order defined by the input {@code samFileHeader}.
+     *
+     * @param outputStream where to write the output.
+     * @param indexOS where to write the output index. Can be null if no index is required.
+     * @param source reference source
+     * @param samFileHeader {@link SAMFileHeader} to be used. Sort order is determined by the sortOrder property of this arg.
+     * @param fileName used for display in error messages
+     */
+    public CRAMFileWriter(
+            final OutputStream outputStream,
+            final OutputStream indexOS,
+            final ReferenceSource source,
+            final SAMFileHeader samFileHeader,
+            final String fileName)
+    {
+        this(outputStream, indexOS, true, source, samFileHeader, fileName); // defaults to presorted==true
+    }
+
+    /**
+     * Create a CRAMFileWriter and index on output streams.
+     *
+     * @param outputStream where to write the output.
+     * @param indexOS where to write the output index. Can be null if no index is required.
+     * @param presorted if true records written to this writer must already be sorted in the order specified by the header
+     * @param source reference source
+     * @param samFileHeader {@link SAMFileHeader} to be used. Sort order is determined by the sortOrder property of this arg.
+     * @param fileName used for display in error message display
+     */
+    public CRAMFileWriter(final OutputStream outputStream, final OutputStream indexOS, final boolean presorted,
+                          final ReferenceSource source, final SAMFileHeader samFileHeader, final String fileName) {
         this.outputStream = outputStream;
-        this.source = source;
         this.samFileHeader = samFileHeader;
         this.fileName = fileName;
-        setSortOrder(samFileHeader.getSortOrder(), true);
+        initCRAMWriter(indexOS, source, samFileHeader, presorted);
+    }
+
+    private void initCRAMWriter(final OutputStream indexOS, final ReferenceSource source, final SAMFileHeader samFileHeader, final boolean preSorted) {
+        this.source = source;
+        setSortOrder(samFileHeader.getSortOrder(), preSorted);
         setHeader(samFileHeader);
 
-        if (this.source == null) this.source = new ReferenceSource(Defaults.REFERENCE_FASTA);
+        if (this.source == null) {
+            this.source = new ReferenceSource(Defaults.REFERENCE_FASTA);
+        }
 
         containerFactory = new ContainerFactory(samFileHeader, recordsPerSlice);
-        if (indexOS != null) indexer = new CRAMIndexer(indexOS, samFileHeader);
+        if (indexOS != null) {
+            indexer = new CRAMIndexer(indexOS, samFileHeader);
+        }
     }
 
     /**
@@ -99,7 +153,6 @@ public class CRAMFileWriter extends SAMFileWriterImpl {
      */
     protected boolean shouldFlushContainer(final SAMRecord nextRecord) {
         return samRecords.size() >= containerSize || refSeqIndex != REF_SEQ_INDEX_NOT_INITIALIZED && refSeqIndex != nextRecord.getReferenceIndex();
-
     }
 
     private static void updateTracks(final List<SAMRecord> samRecords, final ReferenceTracks tracks) {
@@ -337,10 +390,14 @@ public class CRAMFileWriter extends SAMFileWriterImpl {
      */
     @Override
     protected void writeAlignment(final SAMRecord alignment) {
-        if (shouldFlushContainer(alignment)) try {
-            flushContainer();
-        } catch (final Exception e) {
-            throw new RuntimeException(e);
+        if (shouldFlushContainer(alignment)) {
+            try {
+                flushContainer();
+            } catch (IOException e) {
+                throw new RuntimeIOException(e);
+            } catch (IllegalAccessException e) {
+                throw new RuntimeException(e);
+            }
         }
 
         updateReferenceContext(alignment.getReferenceIndex());
@@ -378,14 +435,18 @@ public class CRAMFileWriter extends SAMFileWriterImpl {
     @Override
     protected void finish() {
         try {
-            if (!samRecords.isEmpty()) flushContainer();
+            if (!samRecords.isEmpty()) {
+                flushContainer();
+            }
             CramIO.issueEOF(cramVersion, outputStream);
             outputStream.flush();
-            if (indexer != null)
+            if (indexer != null) {
                 indexer.finish();
-        } catch (final RuntimeException re) {
-            throw re;
-        } catch (final Exception e) {
+            }
+            outputStream.close();
+        } catch (final IOException e) {
+            throw new RuntimeIOException(e);
+        } catch (final IllegalAccessException e) {
             throw new RuntimeException(e);
         }
     }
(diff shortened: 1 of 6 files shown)


mp15:
Technically CRAM files only really require a reference if they lack @SQ lines in their headers. Otherwise you can just fetch the reference in question from the reference sequence server. The following section from the samtools man page describes the process samtools goes through. It might be beneficial for pipeline authors using both tools in their pipelines if you match this

> The search order to obtain a reference is:
> * Use any local file specified by the command line options (eg -T).
> * Look for MD5 via REF_CACHE environment variable.
> * Look for MD5 in each element of the REF_PATH environment variable.
> * Look for a local file listed in the UR: header tag.

droazen:
@mp15 On-demand reference download is disabled by default in htsjdk, so in practice I think we do need to require a reference for crams when `Defaults.USE_CRAM_REF_DOWNLOAD` is false.

cmnbroad:
Also has a fix for #369.

cmnbroad:
Narrowed scope down to just a fix for #369.

cmnbroad:
Contains fixes for #388, #369, #365. Also has some javadoc improvements.

droazen:
@lbergelson Would you be able to review this one while I'm away?

lbergelson:
I can look at this on monday.

src/java/htsjdk/samtools/CRAMFileWriter.java
```diff
@@ -72,22 +72,69 @@
...
 
+    /**
+     * Create a CRAMFileWriter on an output stream. Requires input records to be presorted to match the
+     * sort order defined by the input header.
+     *
+     * @param outputStream where to write the output.
+     * @param source reference source
+     * @param samFileHeader Header to be used. Sort order is determined by the sortOrder property of this arg.
+     * @param fileName used for display in error messages
+     * @return CRAMFileWriter
```
lbergelson: `@return` shouldn't really be used for constructors

src/java/htsjdk/samtools/CRAMFileWriter.java
```diff
@@ -72,22 +72,69 @@
     private CRAMIndexer indexer;
     private long offset;
 
+    /**
+     * Create a CRAMFileWriter on an output stream. Requires input records to be presorted to match the
+     * sort order defined by the input header.
```
lbergelson: could you refer to `{@code samFileHeader}` to make it clear what it's matched against

src/java/htsjdk/samtools/CRAMFileWriter.java
```diff
@@ -72,22 +72,69 @@
...
+     *
+     * @param outputStream where to write the output.
+     * @param indexOS where to write the output index. Can be null if no index is required.
+     * @param source reference source
+     * @param samFileHeader Header to be used. Sort order is determined by the sortOrder property of this arg.
+     * @param fileName used for display in error messages
+     * @return CRAMFileWriter
+     */
     public CRAMFileWriter(final OutputStream outputStream, final OutputStream indexOS, final ReferenceSource source, final SAMFileHeader samFileHeader, final String fileName) {
```
lbergelson: It looks like this constructor can just delegate to the one below it.

src/java/htsjdk/samtools/CRAMFileWriter.java
```diff
@@ -333,10 +380,12 @@ private static void detach(CramCompressionRecord cramRecord) {
...
     @Override
     protected void writeAlignment(final SAMRecord alignment) {
-        if (shouldFlushContainer(alignment)) try {
-            flushContainer();
-        } catch (final Exception e) {
-            throw new RuntimeException(e);
+        if (shouldFlushContainer(alignment)) {
+            try {
+                flushContainer();
+            } catch (final Exception e) {
```
lbergelson: Could you replace this with more explicit exception rewrapping?

```
       try {
                flushContainer();
            } catch (IOException e) {
                throw new RuntimeIOException(e);
            } catch (IllegalAccessException e) {
                throw new RuntimeException(e);
            }
```

src/java/htsjdk/samtools/CRAMFileWriter.java
```diff
@@ -379,6 +428,7 @@ protected void finish() {
             outputStream.flush();
             if (indexer != null)
                 indexer.finish();
+            outputStream.close();
         } catch (final RuntimeException re) {
```
lbergelson: Could you fix these exception rethrows while you're here?

```
      } catch (final IOException e) {
            throw new RuntimeIOException(e);
        } catch (final IllegalAccessException e) {
           throw new RuntimeException(e);
        }
```

src/java/htsjdk/samtools/SAMFileWriterFactory.java
```diff
@@ -293,16 +291,39 @@ public SAMFileWriter makeSAMOrBAMWriter(final SAMFileHeader header, final boolea
         return makeBAMWriter(header, presorted, outputFile);
     }
 
+    /**
+     *
+     * Create a SAM, BAM or CRAM writer based on examination of the outputFile extension.
+     * For CRAM files, read names and tags are preserved.
```
lbergelson: What does it mean that for CRAM files read names and tags are preserved?  Are they changed in SAM/BAM files?

src/java/htsjdk/samtools/SamFileValidator.java
```diff
@@ -274,11 +274,22 @@ private void validateSamRecordsAndQualityFormat(final Iterable<SAMRecord> samRec
...
                 if (cigarIsValid) {
-                    validateNmTag(record, recordNumber);
+                    try {
+                        validateNmTag(record, recordNumber);
+                    }
+                    catch (SAMException e) {
+                        if (hasValidSortOrder) {
+                            // If a CRAM file has an invalid sort order, the ReferenceFileWalker will throw a
+                            // SAMException due to an out of order request when retrieving reference bases during NM
+                            // tag validation; swallow the exception and carry on validating unless only if the sort
```
lbergelson: typo  "carry on validating **unless only if** the sort order is not valid"

src/java/htsjdk/samtools/SAMFileWriterFactory.java
cmnbroad: Because the CRAM spec allows for a wide variety of lossy and lossless compression strategies, its a bit ambiguous to just say we support CRAM (for example, AFAICT cramtools defaults to not preserving readnames when writing records to a CRAM, and synthesizes readnames on the fly on CRAM decode). This javadoc was an attempt to start making incremental progress on https://github.com/samtools/htsjdk/issues/380. I'm wondering now though if the preservation comment is more confusing than helpful.

src/java/htsjdk/samtools/CRAMFileWriter.java
cmnbroad: Good catch.

src/java/htsjdk/samtools/SAMFileWriterFactory.java
```diff
@@ -311,28 +332,91 @@ public CRAMFileWriter makeCRAMWriter(final SAMFileHeader header, final OutputStr
...
+     *
+     * @param header entire header. Sort order is determined by the sortOrder property of this arg.
+     * @param presorted  if true, SAMRecords must be added to the SAMFileWriter in order that agrees with header.sortOrder.
+     * @param outputFile where to write the output.  Must end with .sam, .bam or .cram.
+     * @param referenceFasta reference sequence file
+     * @return CRAMFileWriter
+     *
+     * Read names and tags are preserved.
+     */
+    private CRAMFileWriter createCRAMWriter(final SAMFileHeader header, final boolean presorted, final File outputFile, final File referenceFasta) {
```
lbergelson: There are 2 overloads of makeCramWriter that call into this method, and a different one that doesn't but duplicates some of the work that this one does.  Could you unify them?  It's seems very likely that they will eventually become confusingly out of sink if they don't call into the same code.

src/java/htsjdk/samtools/SAMFileWriterFactory.java
```diff
@@ -311,28 +332,91 @@ public CRAMFileWriter makeCRAMWriter(final SAMFileHeader header, final OutputStr
...
+        OutputStream cramOS = null;
+        OutputStream indexOS = null ;
 
-            OutputStream indexOS = null ;
-            if (createIndex) {
-                File indexFile = new File(outputFile.getAbsolutePath() + ".bai") ;
-                indexOS = new FileOutputStream(indexFile) ;
+        if (createIndex) {
+            if (!IOUtil.isRegularPath(outputFile)) {
+                System.err.println("Cannot create index for CRAM because output file is not a regular file: " + outputFile.getAbsolutePath());
```
lbergelson: this should be logger warning instead of printed to system.err

src/tests/java/htsjdk/samtools/SAMFileWriterFactoryTest.java
```diff
@@ -145,6 +148,78 @@ private void fillSmallBam(SAMFileWriter writer) {
...
+        final SAMFileWriterFactory factory = createWriterFactory(header);
+
+        final SAMFileWriter samWriter = factory.makeCRAMWriter(header, false, outputFile, new File(TEST_DATA_DIR, "hg19mini.fasta"));
+        fillSmallBam(samWriter);
+
+        samWriter.close();
+        verifyWriterOutput(outputFile);
+    }
+
 }
```
lbergelson: Could you add tests to exercise the other overloads of `makeCRAMWriter()`

src/java/htsjdk/samtools/CRAMFileWriter.java
lbergelson: Could you add tests to exercise all 3 of these constructors.

lbergelson:
@cmnbroad Back to you.  A few minor comments.

src/java/htsjdk/samtools/SAMFileWriterFactory.java
lbergelson: I didn't know this about cram format, and I suspect a lot of people will also not know it.  I think it's more confusing than helpful at the moment.  Possibly something more detailed would be helpful, but I'm having difficulty phrasing it.  

Maybe something like "CRAM compressed files will be written losslessly and maintain the same information as a BAM"

src/java/htsjdk/samtools/SAMFileWriterFactory.java
cmnbroad: It is confusing; the problem is the first (3 arg) makeCRAMWriter method, which doesn't take a presorted flag and takes a stream instead of a file (so it can't and doesn't honor the other factory settings). It creates the CRAMFileWriter directly. If it wasn't already public I would lobby to remove it since I think it has limited use.

The other two overloads call createCRAMWriter who's main job is to propagate the factory settings, but it depends on being passed a file (to create parallel index and md5 files). I'm going to rename createCRAMWriter to reflect that and factor out the 2 duplicated lines of initialization code. Other than that I don't see a refactoring that doesn't add even more confusion, but if you have a suggestion do let me know.

src/java/htsjdk/samtools/SAMFileWriterFactory.java
cmnbroad: That's exactly what I would like to be able say (that by default its completely lossless). But I don't know for sure if its true. Hence #380. For now I'll remove these since I think you're right that they're not helpful without a more comprehensive statement.

src/java/htsjdk/samtools/SAMFileWriterFactory.java
lbergelson: I think factoring out the extra 2 lines of configuration will at least make sure that no one will change one without the other. 

I do like the ability to use a stream instead of a file though.  The fact that we can't do that in many places is a real pain.  I guess you can always directly invoke the CRAMWriter  constructor if you want to write from a stream though and you're planning on ignoring the factory anyway.

src/java/htsjdk/samtools/SAMFileWriterFactory.java
lbergelson: I see.  I agree that it is best for now then.

cmnbroad:
All requested changes are made. Back to @lbergelson.

src/tests/java/htsjdk/samtools/SAMFileWriterFactoryTest.java
```diff
@@ -141,10 +144,123 @@ private void createSmallBamToOutputStream(final OutputStream outputStream,boolea
...
+        factory.setCreateMd5File(true);
+        // index only created if coordinate sorted
+        header.setSortOrder(SAMFileHeader.SortOrder.coordinate);
+        header.addSequence(new SAMSequenceRecord("chr1", 123));
+        header.addReadGroup(new SAMReadGroupRecord("1"));
+        return factory;
+    }
+
+    private void verifyWriterOutput(File outputFile, ReferenceSource refSource, int nRecs, boolean verifySupplementalFiles) {
+        if (true == verifySupplementalFiles) {
```
lbergelson: this should just be `if ( verifySupplementalFiles )`

src/tests/java/htsjdk/samtools/SAMFileWriterFactoryTest.java
```diff
@@ -141,10 +144,123 @@ private void createSmallBamToOutputStream(final OutputStream outputStream,boolea
...
+    @DataProvider(name="bamOrCramWriter")
+    public Object[][] bamOrCramWriter() {
+        return new Object[][] {
+                { BamFileIoUtils.BAM_FILE_EXTENSION, },
+                { CramIO.CRAM_FILE_EXTENSION }
+        };
+    }
+
+    @Test(dataProvider="bamOrCramWriter")
+    public void testMakeWriter(String extension) throws Exception {
```
lbergelson: these 4 tests look like they should share 1 a common method to reduce duplication

lbergelson:
@cmnbroad Two minor very minor comments.  Please rebase and then good to merge.

cmnbroad:
Back to @lbergelson. I don't see a way to factor the test methods any more as you suggested; the main thing that differs between them is the factory method thats used, but all of the duplicated code manipulates local variables that are used multiple times. If you do have a suggestion pls send it back to me, otherwise can you merge as I don't have permissions in this repo.

lbergelson:
Ah, ok.  If there isn't an easy way to refactor them that's fine.  I just glanced at them and assumed there would be, sorry if I sent you on a wild goose chase.

PR #370, opened 2015-11-06
Significant performance improvements in the DuplicateSetIterator and friends...

tfenne:
by both making the sorting/comparing process for duplicate ordering more efficient, and also doing less of it.

The only thing in here that might be contentious is that I added the facility to SAMRecord to store arbitrary transient attributes that are not persisted but are retained either for the lifetime of the record in RAM or until a caller removes them.  This allows one-time computation of expensive derived values (e.g. unclipped mate start/end position) without adding special handling for each and every field within SAMRecord.

Left as two commits to highlight the two different changes made; happy to rebase at the end of the review.

### DIFF
--- a/src/java/htsjdk/samtools/SAMRecord.java
+++ b/src/java/htsjdk/samtools/SAMRecord.java
@@ -32,7 +32,9 @@ import java.io.Serializable;
 import java.lang.reflect.Array;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 
@@ -167,6 +169,9 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
      */
     private transient SAMFileSource mFileSource;
     private SAMFileHeader mHeader = null;
+
+    /** Transient Map of attributes for use by anyone. */
+    private transient Map<Object,Object> transientAttributes;
 
     public SAMRecord(final SAMFileHeader header) {
         mHeader = header;
@@ -1932,7 +1937,39 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
      * shortcut to <pre>SAMFlag.getFlags( this.getFlags() );</pre>
      * @returns a set of SAMFlag associated to this sam record */
     public final Set<SAMFlag> getSAMFlags() {
-        return SAMFlag.getFlags( this.getFlags() );
+        return SAMFlag.getFlags(this.getFlags());
+    }
+
+    /**
+     * Fetches the value of a transient attribute on the SAMRecord, of null if not set.
+     *
+     * The intended use for transient attributes is to store values that are 1-to-1 with the SAMRecord,
+     * may be needed many times and are expensive to compute.  These values can be computed lazily and
+     * then stored as transient attributes to avoid frequent re-computation.
+     */
+    public final Object getTransientAttribute(final Object key) {
+        return (this.transientAttributes == null) ? null : this.transientAttributes.get(key);
+    }
+
+    /**
+     * Sets the value of a transient attribute, and returns the previous value if defined.
+     *
+     * The intended use for transient attributes is to store values that are 1-to-1 with the SAMRecord,
+     * may be needed many times and are expensive to compute.  These values can be computed lazily and
+     * then stored as transient attributes to avoid frequent re-computation.
+     */
+    public final Object setTransientAttribute(final Object key, final Object value) {
+        if (this.transientAttributes == null) this.transientAttributes = new HashMap<Object,Object>();
+        return this.transientAttributes.put(key, value);
+    }
+
+    /**
+     * Removes a transient attribute if it is stored, and returns the stored value. If there is not
+     * a stored value, will return null.
+     */
+    public final Object removeTransientAttribute(final Object key) {
+        if (this.transientAttributes != null) return this.transientAttributes.remove(key);
+        else return null;
     }
 }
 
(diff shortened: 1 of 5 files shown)


tfenne:
@nh13 Would you mind looking at the changes I made?

tfenne:
@yfarjoun @eitanbanks Opinions on the noted change to SAMRecord?  I think the impact is pretty low - just one more reference/pointer in SAMRecord, and it's not initialized to a Map until the first transient attribute is stored, so if it's not used the overhead is low.

lbergelson:
Why add the transient map instead of just using a wrapper class that caches the values you want. That seems like it would be more readable and have essentially identical performance impact?

I'm assuming there is a good reason to do it this way, but I don't see it.

tfenne:
@lbergelson I did initially think about subclassing SAMRecord and using the SAMRecordFactory mechanism to do this but chose not to because a) this seems like something that is generally useful and b) that mechanism doesn't work will with the whole SortingCollection framework that is used in this case and elsewhere.

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1887,5 +1892,16 @@ public String getPairedReadName() {
...
+    /** Fetches the value of a transient attribute on the SAMRecord, of null if not set. */
+    public final Object getTransientAttribute(final Object key) {
+        return (this.transientAttributes == null) ? null : this.transientAttributes.get(key);
+    }
+
+    /** Sets the value of a transient attribute, and returns the previous value if defined. */
+    public final Object setTransientAttribute(final Object key, final Object value) {
+        if (this.transientAttributes == null) this.transientAttributes = new HashMap<Object,Object>();
+        return this.transientAttributes.put(key, value);
+    }
```
nh13: How about at ```public final Object removeTransientAttribute(final Object key)``` method (or returns a ```boolean```)?

src/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
```diff
@@ -39,6 +39,10 @@
  * @author nhomer
  */
 public class SAMRecordDuplicateComparator implements SAMRecordComparator {
+    /** An enum to provide type-safe keys for transient attributes the comparator puts on SAMRecords. */
+    private static enum Attr {
+        LibraryId, ReadCoordinate, MateCoordinate
```
nh13: Perhaps rename ```ReadCoordinate``` to ```UnclippedReadCoordinate```?   The same for ```MateCoordinate```?

nh13: Perhaps these useful attributes should be defined somewhere, like in ```SAMRecordUtils```? Your call as we have not needed it in the past.

src/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
```diff
@@ -198,6 +214,8 @@ private boolean pairedEndAndBothMapped(final SAMRecord record) {
      * properly choose the first end for optical duplicate identification when both ends are mapped to the same position etc. 
      */ 
     public int compare(final SAMRecord samRecord1, final SAMRecord samRecord2) {
+        populateTransientAttributes(samRecord1, samRecord2);
+
```
nh13: extra newline (see line 255).

src/tests/java/htsjdk/samtools/DuplicateSetIteratorTest.java
```diff
@@ -56,7 +57,7 @@ public void testSupplementalReads() {
         }
 
         //we expect 15 duplicate sets one for the initial two reads and one for each of the additional 14 reads.
-        assert (allSets.size() == 15);
-        assert (allSets.get("READ0").size() == 2);
+        Assert.assertEquals(allSets.size(), 15, "Wrong number of duplicate sets.");
+        Assert.assertEquals(allSets.get("READ0").size(), 2, "Should be two reads in the READ0 duplicate set, but there are not.");
     }
```
nh13: Test to put an attribute into a ```SAMRecord```, retrieve it, and remove it?

tfenne: Ok Adam.

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1887,5 +1892,16 @@ public String getPairedReadName() {
...
         return SAMFlag.getFlags( this.getFlags() );
     }
+
+    /** Fetches the value of a transient attribute on the SAMRecord, of null if not set. */
+    public final Object getTransientAttribute(final Object key) {
+        return (this.transientAttributes == null) ? null : this.transientAttributes.get(key);
+    }
+
+    /** Sets the value of a transient attribute, and returns the previous value if defined. */
+    public final Object setTransientAttribute(final Object key, final Object value) {
```
lbergelson: Could you add a comment about how this is intended to be used?  I.e. for caching results of expensive computations.

src/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
tfenne: I really like the approach I've taken where the using class declares a private enum and uses the values as the key.  It essentially makes key collision impossible and hides transient attributes from classes that didn't create them/don't have access to the enum.  So I'd prefer not to centralize.

tfenne:
Addressed comments/requests by @nh13 and @lbergelson .  Once the tests pass will merge unless anyone jumps in with an objection.

src/tests/java/htsjdk/samtools/DuplicateSetIteratorTest.java
akiezun: Watching you

tfenne: image

tfenne: canyoudealwiththat

nh13:
@tfenne :+1:  from me

PR #372, opened 2015-11-06
Changes to provide multiple downsampling strategies.

tfenne:
[deleted]

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -157,7 +157,7 @@
 
         <testng suitename="htsjdk-single-test" classpathref="classpath" outputdir="${test.output}"
                 verbose="${testng.verbosity}">
-            <jvmarg line="-Xmx512M ${debug.jvm.args}"/>
+            <jvmarg line="-Xmx512m ${debug.jvm.args}"/>
             <classpath>
                 <pathelement path="${classes}"/>
                 <pathelement path="${classes.test}"/>
--- a/src/java/htsjdk/samtools/ChainedDownsamplingIterator.java
+++ b/src/java/htsjdk/samtools/ChainedDownsamplingIterator.java
@@ -1,0 +1,90 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015 Tim Fennell
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package htsjdk.samtools;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * A DownsamplingIterator that combines the ConstantMemory and HighAccuracy downsampling techniques to provide an
+ * iterator that has accuracy approaching that of HighAccuracy, but with more limited memory usage.  Instead of
+ * requiring memory proportional to number of read names in the incoming stream of reads, requires memory
+ * approximately proportional to the number of output reads.
+ *
+ * @author Tim Fennell
+ */
+class ChainedDownsamplingIterator extends HighAccuracyDownsamplingIterator {
+    public static final int MIN_ACCURATE_INPUT_READS = 50000;
+
+    /**
+     * Constructs a chained iterator that will read from the provided iterator and attempt to downsampling to the provided proportion.
+     */
+    ChainedDownsamplingIterator(final Iterator<SAMRecord> iterator, final double proportion, final int seed) {
+        super(new ConstantMemoryDownsamplingIterator(iterator, adjustProportion(proportion), seed), proportion, seed);
+
+        // Deal with the fact that the iterator will advance and discard some reads at construction
+        final long discarded = ((ConstantMemoryDownsamplingIterator) getUnderlyingIterator()).getDiscardedCount();
+        recordDiscardRecords(discarded);
+    }
+
+    /**
+     * Calculates the upper bound of 99.9% CI given the proportion, that is used to "buffer" the proportion on
+     * the constant memory downsampler, to make sure it leaves enough reads for us to downsample.
+     *
+     * Uses an assumed number of reads tested as this is often not known until after the fact.
+     */
+    private static double adjustProportion(final double p) {
+        final double ciAdjustment99_9 = 3.3 * Math.sqrt(p/MIN_ACCURATE_INPUT_READS);
+        return Math.min(1, p + ciAdjustment99_9);
+    }
+
+
+    /**
+     * Resets statistics before reading from the underlying iterator.
+     */
+    @Override
+    protected void readFromUnderlyingIterator(final List<SAMRecord> recs, final Set<String> names, final int templatesToRead) {
+        // Reset the stats on the underlying iterator
+        ((ConstantMemoryDownsamplingIterator) getUnderlyingIterator()).resetStatistics();
+
+        // Read from the underlying iterator
+        super.readFromUnderlyingIterator(recs, names, templatesToRead);
+    }
+
+    @Override
+    protected int calculateTemplatesToKeep(final int templatesRead, final double overallProportion) {
+        // Calculate an adjusted proportion to keep, knowing what proportion the underlying iterator discarded
+        final ConstantMemoryDownsamplingIterator iter = (ConstantMemoryDownsamplingIterator) getUnderlyingIterator();
+        final double priorProportion = iter.getAcceptedFraction();
+        final double p = Math.max(0, Math.min(1, overallProportion / priorProportion));
+        final int retval =  super.calculateTemplatesToKeep(templatesRead, p);
+
+        // Record all the discarded records to keep the overall statistics accurate, but do it after
+        // the call to super() so it doesn't affect the proportion calculation.
+        recordDiscardRecords(iter.getDiscardedCount());
+
+        return retval;
+    }
+}
--- a/src/java/htsjdk/samtools/ConstantMemoryDownsamplingIterator.java
+++ b/src/java/htsjdk/samtools/ConstantMemoryDownsamplingIterator.java
@@ -1,0 +1,88 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015 Tim Fennell
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package htsjdk.samtools;
+
+import htsjdk.samtools.util.Murmur3;
+import htsjdk.samtools.util.PeekableIterator;
+
+import java.util.Iterator;
+
+/**
+ * A DownsamplingIterator that runs in constant (and very small) memory. For each read the read name is hashed
+ * using the Murmur3_32 hash algorithm to obtain an integer value that is, enough for our purposes, uniformly
+ * distributed between the min and max int values even for highly similar inputs.  The proportion is used to
+ * calculate a maximum acceptable hash value within the range.  Records whose hash value is below the limit
+ * are emitted, records whose hash value is above the limit are discarded.
+ *
+ * Does not make any attempt to be accurate (have actual proportion == requested proportion) beyond what would
+ * be expected for a random process and so may become quite inaccurate when downsampling to small numbers of
+ * reads.
+ *
+ * @author Tim Fennell
+ */
+class ConstantMemoryDownsamplingIterator extends DownsamplingIterator {
+    private final PeekableIterator<SAMRecord> underlyingIterator;
+    private final int maxHashValue;
+    private final Murmur3 hasher;
+
+
+    /** Constructs a downsampling iterator upon the supplied iterator, using the Random as the source of randomness. */
+    ConstantMemoryDownsamplingIterator(final Iterator<SAMRecord> iterator, final double proportion, final int seed) {
+        super(proportion);
+        this.hasher = new Murmur3(seed);
+        this.underlyingIterator = new PeekableIterator<SAMRecord>(iterator);
+
+        final long range = (long) Integer.MAX_VALUE - (long) Integer.MIN_VALUE;
+        this.maxHashValue = Integer.MIN_VALUE + (int) Math.round(range * proportion);
+
+        advanceToNextAcceptedRead();
+    }
+
+    /** Returns true if there is another record available post-downsampling, false otherwise. */
+    @Override public boolean hasNext() {
+        // The underlying iterator is always left at the next return-able read, so if it has a next read, so do we
+        return this.underlyingIterator.hasNext();
+    }
+
+    /**
+     * Advances the underlying, peekable, iterator until the next records is one that is to be emitted.
+     * @return true if there is at least one emittable record ready for emission, false otherwise
+     */
+    private boolean advanceToNextAcceptedRead() {
+        while (this.underlyingIterator.hasNext() && this.hasher.hashUnencodedChars(this.underlyingIterator.peek().getReadName()) > this.maxHashValue) {
+            this.underlyingIterator.next();
+            recordDiscardedRecord();
+        }
+
+        return this.underlyingIterator.hasNext();
+    }
+
+    /** Returns the next record from the iterator, or throws an exception if there is no next record. */
+    @Override public SAMRecord next() {
+        final SAMRecord rec = this.underlyingIterator.next();
+        recordAcceptedRecord();
+        advanceToNextAcceptedRead();
+        return rec;
+    }
+}
--- a/src/java/htsjdk/samtools/DownsamplingIterator.java
+++ b/src/java/htsjdk/samtools/DownsamplingIterator.java
@@ -25,127 +25,89 @@ package htsjdk.samtools;
 
 import htsjdk.samtools.util.CloseableIterator;
 
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.Map;
-import java.util.NoSuchElementException;
-import java.util.Random;
-
 /**
- * An iterator of SAMRecords that can downsample on the fly.  Allows for inclusion of secondary and/or
- * supplemental records (off by default), though this will cause memory use to increase as the decisions
- * for each read name must be cached permanently.
+ * Abstract base class for all DownsamplingIterators that provides a uniform interface for recording
+ * and reporting statistics bout how many records have been kept and discarded.
+ *
+ * A DownsamplingIterator is an iterator that takes another iterator of SAMRecords and filters out a
+ * subset of those records in a random way, while ensuring that all records for a template (i.e. record name)
+ * are either retained or discarded.  Strictly speaking the proportion parameter applies to templates,
+ * though in most instances it is safe to think about it being applied to records.
  *
  * @author Tim Fennell
  */
-public class DownsamplingIterator implements CloseableIterator<SAMRecord>, Iterable<SAMRecord> {
-    private final Iterator<SAMRecord> underlyingIterator;
-    private final Random random;
-    private final double probabilityOfKeeping;
-    private SAMRecord nextRecord;
-    private long totalReads, keptReads;
-    private final Map<String, Boolean> decisions = new HashMap<String, Boolean>();
-    private boolean allowSecondaryAlignments = false;
-    private boolean allowSupplementalAlignments = false;
-    private boolean includeNoRefReads = true;
+public abstract class DownsamplingIterator implements CloseableIterator<SAMRecord> {
+    private long recordsSeen;
+    private long recordsAccepted;
+    private double targetProportion;
 
-    /** Constructs a downsampling iterator upon the supplied iterator, using the Random as the source of randomness. */
-    public DownsamplingIterator(final Iterator<SAMRecord> iterator, final Random random, final double probabilityOfKeeping) {
-        this.underlyingIterator = iterator;
-        this.random = random;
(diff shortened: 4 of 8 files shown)


tfenne:
@yfarjoun Can I ask you to take a look at this?  @nh13 has already over-the-shoulder reviewed.

tfenne:
A couple of notes:
1) I have a change for Picard coming once this PR is merged that incorporates this change and exposes some other options in DownsampleSam.
2) My goal here is to make the downsampling procedure more accurate, while still requiring only a single pass through the file and not requiring significantly more memory *unless* you want super-high accuracy.

It achieves this by reading in batches of reads (if you want accuracy +/- 1/1000 then in reads in records until it has seen 1000 novel _templates_), and then keeping `PROBABILITY * batch_size` records.  It rounds up or down depending on whether it is currently under or over it's goal.

yfarjoun:
can you explain what you mean by "more accurate downsampling"? In what way
is the current iterator not accurate?

tfenne:
@yfarjoun Sorry - yes should have explained that.  The current version of the iterator simply makes it so that every template has `PROBABILITY` of being retained, but makes no effort beyond that to ensure that on average input * `PROBABILITY` reads are emitted.  When downsampling to very large numbers (several million reads) you get a fairly accurate output.  But if downsampling to smaller numbers the stochasticity of the process means that it's very hard to use the iterator to produce an output with a specific number of reads in it.

I.e. if I have a file with 10m reads and want 250k for some reason, I might actually end up with 250k +/- 1% when using `PROBABILITY=0.025`.  The goal of this change is to make the result have `output_reads = input_reads * (probability +/- accuracy)`.

yfarjoun:
I take issue with your use of "on average", since it is exactly, on average
that input * PROBABILITY records will be emitted...

You seem to want something better than that. You want a smaller
standard-deviation than the expected input*sqrt(P), am I correct?

tfenne:
I'm not really looking for a semantic argument :disappointed: Yes, I'm trying to make it so that for any single invocation of DownsampleSam (which uses the iterator) you get a more predictable and closer to the target number of reads emitted.  Many people (me included) use this with the goal of reducing a BAM with N input reads to a BAM with N*PROBABILITY reads.  This PR is to make it so that the different between what is output and N*PROBABILITY is within known bounds and not stochastic.

yfarjoun:
sorry I can't help it sometime...:-p

src/java/htsjdk/samtools/DownsamplingIterator.java
```diff
@@ -123,6 +152,45 @@ private boolean advance() {
...
+    private boolean bufferNextChunkOfRecords() {
+        final int templatesToRead = (int) Math.ceil(1 / this.targetAccuracy);
+        final Set<String> names = new HashSet<String>();
+        final List<SAMRecord> recs = new ArrayList<SAMRecord>(templatesToRead);
+
+        while (this.underlyingIterator.hasNext() && names.size() < templatesToRead) {
+            final SAMRecord rec = this.underlyingIterator.next();
+            recs.add(rec);
+
+            if (!this.allowSecondaryAlignments && rec.getNotPrimaryAlignmentFlag()) continue;
```
yfarjoun: getNotPrimaryAlignmentFlag=true,  could also mean supplementary, right? 
what if 
```java    
    this.allowSecondaryAlignments==false;
    this.allowSupplementalAlignments==true;
```
and 
```java
    rec.getNotPrimaryAlignmentFlag()==true && rec.getSupplementaryAlignmentFlag()==true
```

I think that the record will not be included, erroneously.

src/java/htsjdk/samtools/DownsamplingIterator.java
```diff
@@ -123,6 +152,45 @@ private boolean advance() {
...
+            if (!this.allowSupplementalAlignments && rec.getSupplementaryAlignmentFlag()) continue;
+            if (this.decisions.containsKey(rec.getReadName())) continue;
+            names.add(rec.getReadName());
+        }
+
+        final int templatesRead   = names.size();
```
yfarjoun: extra space

src/java/htsjdk/samtools/DownsamplingIterator.java
tfenne: No - there's understandably some confusion around this.  `getNotPrimaryAlignmentFlag()` just returns the boolean/flag bit for primary vs. secondary.  There's another method called something like `isSecondaryOrSupplemental()` that does what you're thinking.

src/java/htsjdk/samtools/DownsamplingIterator.java
```diff
@@ -74,12 +85,30 @@ public DownsamplingIterator setIncludeNoRefReads(final boolean includeNoRefReads
...
     }
 
+    /**
+     * Sets the target accuracy of the downsampling iterator.  The value should be thought of as
+     * probability +/- accuracy.  So a value of 0.001 would instruct the downsampling iterator to
+     * attempt to guarantee at accuracy to within 0.1%.  The downsampler will need to buffer reads
+     * for 1/accuracy templates, so setting this to extremely small numbers is not advisable.
+     */
+    public DownsamplingIterator setTargetAccuracy(final double accuracy) {
+        if (accuracy >= 1 || accuracy <= 0) throw new IllegalArgumentException("Illegal value. Must be 0 < accuracy < 1");
```
yfarjoun: to avoid overflow, accuracy must be greater than 1/MAX_INT (this is also true for other reasons, but one could conceive of enough memory to hold 2x10^9 records in memory, and so perhaps it's worth noting, we've been having overflows in other parts of the code due to similar issues.)

src/java/htsjdk/samtools/DownsamplingIterator.java
```diff
@@ -123,6 +152,45 @@ private boolean advance() {
...
+            if (!this.allowSecondaryAlignments && rec.getNotPrimaryAlignmentFlag()) continue;
+            if (!this.allowSupplementalAlignments && rec.getSupplementaryAlignmentFlag()) continue;
+            if (this.decisions.containsKey(rec.getReadName())) continue;
+            names.add(rec.getReadName());
+        }
+
+        final int templatesRead   = names.size();
+        final double rawTemplatesToKeep = templatesRead * this.probabilityOfKeeping;
+        final int templatesToKeep = (keptTemplates / (double) totalTemplates < this.probabilityOfKeeping)
+                ? (int) Math.ceil(rawTemplatesToKeep) : (int) Math.floor(rawTemplatesToKeep);
```
yfarjoun: the alluded to overflow is here

src/java/htsjdk/samtools/DownsamplingIterator.java
```diff
@@ -123,6 +152,45 @@ private boolean advance() {
...
+        final List<String> tmp    = new ArrayList<String>(names);
+        while (names.size() > templatesToKeep) {
+            names.remove(tmp.get(this.random.nextInt(templatesRead)));
```
yfarjoun: This could take a while if you are downsampling a large amount (as you'll be hitting already-removed names). Could you make tmp a linkedlist and also remove the drawn integer position from the list? Otherwise, you could hash the names using murmur or similar and take the first templatesToKeep by hash order.

yfarjoun:
Got Testz?

This is quite complicated and it would be nice to have some sanity checks...just check that the number of records emitted is in the ball-park, I realize it would be difficult to test for correctness, but P=1, P=0, and P=0.5 should return reasonably sized bams, yes?

Also, since I think we spotted a bug in the logic of secondary and supplementary reads, could you add a unit test for that?

yfarjoun:
@tfenne back to you.

src/java/htsjdk/samtools/DownsamplingIterator.java
yfarjoun: I see. the sam-spec calls this flag  "secondary alignment", but the htsjdk
code it's getNotPrimaryAlignmentFlag...perhaps it's time to deprecate and
have a different function?

not in this PR, clearly!



On Fri, Nov 6, 2015 at 5:23 PM, Tim Fennell <notifications@github.com>
wrote:
>
> In src/java/htsjdk/samtools/DownsamplingIterator.java:
>
> > +     * downsampling can be performed and the target accuracy
achieved.  Once reads have been buffered,
> > +     * template names are randomly sampled out for discarding until
the desired number of reads have
> > +     * been discarded.
> > +     *
> > +     * @return True if one or more reads have been buffered, false
otherwise
> > +     */
> > +    private boolean bufferNextChunkOfRecords() {
> > +        final int templatesToRead = (int) Math.ceil(1 /
this.targetAccuracy);
> > +        final Set<String> names = new HashSet<String>();
> > +        final List<SAMRecord> recs = new
ArrayList<SAMRecord>(templatesToRead);
> > +
> > +        while (this.underlyingIterator.hasNext() && names.size() <
templatesToRead) {
> > +            final SAMRecord rec = this.underlyingIterator.next();
> > +            recs.add(rec);
> > +
> > +            if (!this.allowSecondaryAlignments &&
rec.getNotPrimaryAlignmentFlag()) continue;
>
> No - there's understandably some confusion around this.
getNotPrimaryAlignmentFlag() just returns the boolean/flag bit for primary
vs. secondary. There's another method called something like
isSecondaryOrSupplemental() that does what you're thinking.

src/java/htsjdk/samtools/DownsamplingIterator.java
tfenne: I did think about this, and discussed offline with @nh13.  I think a LinkedList would be disastrous for because just getting to the n'th element requires traversing n links in the list.  So while in theory it should be fewer logical operations, the cost of each operation would be many times higher.

The other option I had considered was to just remove elements from the list, and then convert it to a Set at the end.  But I'm not sure that shuffling all that memory is going to be faster either than just putting up with the repeated hits.  I'm happy to change it if you think this would be better.

tfenne:
@yfarjoun Have attempted to address your comments.  I've left them in separate commits that I will squash once we're ready to merge.

The only thing I didn't do (I think) is test your secondary/supplemental case because reading it again, I disagree with you on the correct result.  I think in the case that all of the following are true:

```java    
    this.allowSecondaryAlignments==false;
    this.allowSupplementalAlignments==true;
    rec.getNotPrimaryAlignmentFlag()==true && rec.getSupplementaryAlignmentFlag()==true
```

that the record should be discarded.  There are two hurdles - it makes it past the supplemental hurdle because, although it is supplemental, those are allowed by the second line.  But it fails the secondary/not-primary hurdle because is *not* primary, i.e. secondary, and those are dis-allowed by the first line.

src/java/htsjdk/samtools/DownsamplingIterator.java
yfarjoun: then why not shuffle tmp and then remove the first templatesToKeep?

yfarjoun: rather the first size-templatesToKeep

yfarjoun:
@tfenne yes, my example was incorrect, as it was based on the misinterpretation of getNotPrimaryAlignmentFlag.

tfenne:
@yfarjoun I made the change you suggested, and then decided to try something else.  I'd appreciate it if you and @nh13 could look at my latest commit on this branch/PR.

I've replaced the use of a Map to store retain/discard decisions with a pair of BitSets (one to record decisions, one to record whether or not a decision has been made yet).  And instead of making a decision per template name, it makes a decision per "key", where key is currently implemented as abs(name.hashCode()).  I think the only real question is whether String.hashCode() is distributed uniformly _enough_.  It makes efforts to be uniform, and in practice it seems to work well enogh, but does reduce accuracy a tiny bit.

On the plus side, it now runs in a fixed amount of memory and runs faster due to not querying a HashMap so frequently.

yfarjoun:
I recall we had discussed using this approach a while back, and the
conclusion was that a good hash function to use would be murmur3 (I think).
The problem was finding a java implementation that we could use. I think
that String.hashCode() had too many collisions to be considered a good
hash-code for this purpose.

tfenne:
I did a little testing with a random real BAM from MiSeq data.  The BAM has 17,443,821 actual templates (i.e. unique read names in it).  If I've done my math right, I'd expect by chance to have draw out 17,373,165 unique hashes between 0 and Integer.MAX_VALUE.  With String.hashCode() I get 17,373,668 unique hashes, with Murmur3_32 I get 17,373,368.  Both are higher than the expected, but String.hashCode() is higher than Murmur3_32.  I'm left wondering why we thought String.hashCode would not be random enough - perhaps it is more susceptible to collisions in certain edge cases?

If we do prefer Murmur3, Google Guava has a nice java implementation of Murmur3_32 (and many other hash functions) all under apache 2 licensing which allows for copying and modification of the source.  And conveniently it looks like the clojure guys have already done the work of extracting it to a standalone class here:
    https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/Murmur3.java

yfarjoun:
IIRC, the problem was that if string A and string B collide then A+C and
B+C collide for all strings C.

yfarjoun:
I was basing the analysis on
http://programmers.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed
and it was murmur2 that won (perhaps 3 is new and even better?)

tfenne:
Interesting reading @yfarjoun  - thanks!  So without doing similar testing or figuring out which algorithm String.hashCode() maps to, Murmur would appear to be safest.

That leaves the question of whether this is preferable.  I think in general it is *except* for cases where the input approaches 2e9 templates as that will cause some undesirable behaviour as the amount of reads needed to see N _new_ templates in `bufferNextChunkOfRecords()` will expand further and further, causing ever larger batches of reads to be buffered.  For example, a 60X human whole genome with 2x150 reads might have > 6e8 templates or 0,3x the hash space.

yfarjoun:
Java String hash uses something similar to DJB2 (but with 31 instead of
33). The post has a plot of DJB2a which uses a XOR instead of +. While this
"should" result in even better hash, you can see in the post that DJB2a is
not nearly as good as murmur2/3.

I support the inclusion of murmur3 into htsjdk, though note that the
cloture class isn't standalone, it needs Util.java.

yfarjoun:
Given that the hashcode can be a long not an int, we could also allow more
than 32 bits for the hashcode. From the bamindex one could easily find the
total number of reads, take some multiple of that to inform the hash-space
you need and murmur will provide...right? the memory usage will increase
for large bams, but that was already the case with the map

The only problem is that the cloture implementation provides the 32 bit
version, not the 64 or 128 bit versions that exist in Guava, and if I
understand correctly we do not want to include Guava in HTSJDK..

tfenne:
@yfarjoun @nh13 After spending some more time thinking about this I decided that the correct way forward was to provide multiple ways of downsampling to cater to multiple use cases.  The use cases I am trying to cover are:
1) Downsampling very large input files without needs a ton of memory
2) Downsampling moderately sized input files to small numbers of reads with high accuracy ( @yfarjoun I've converged on defining accuracy as the absolute difference between the requested proportion and the emitted proportion)
3) Downsampling very large input files to relatively small numbers of reads at higher accuracy than #1 provides.

I've rebased and squashed since the latest changes are quite a shift in direction.  I'm much happier with the results though. 

I should probably also note that I have decided that it is *not* the responsibility of downsampling iterators to filter reads in any way other than downsampling (i.e. secondary, supplemental, unmapped) and that can easily be achieved by using a `FilteringIterator` under a `DownsamplingIterator`.

yfarjoun:
Sounds good. I'll take a look. I agree 100% with leaving the filtering to a
subsequent, or previous filteringiterator.

build.xml
```diff
@@ -157,7 +157,7 @@
 
         <testng suitename="htsjdk-single-test" classpathref="classpath" outputdir="${test.output}"
                 verbose="${testng.verbosity}">
-            <jvmarg line="-Xmx512M ${debug.jvm.args}"/>
+            <jvmarg line="-Xmx2g ${debug.jvm.args}"/>
```
nh13: be careful here, as Travis CI only has 3Gb of memory.

src/java/htsjdk/samtools/ConstantMemoryDownsamplingIterator.java
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * A DownsamplingIterator that runs in constant (and very small) memory. For each read the read name is hashed
+ * using the Murmur3_32 hash algorithm to obtain an integer value that is, enough for our purposes, uniformly
+ * distributed between the min and max int values even for highly similar inputs.  The proportion is used to
+ * calculate a maximum acceptable hash value within the range.  Records whose hash value is below the limit
+ * are emitted, records whose hash value is above the limit are discarded.
+ *
+ * Does not make any attempt to be accurate (have actual proportion == requested proportion) beyond what would
+ * be expected for a random process and so may become quite in accurate when downsampling to small numbers of
```
nh13: `in accurate` -> `inaccurate`

src/java/htsjdk/samtools/DownsamplingIteratorFactory.java
```diff
@@ -0,0 +1,117 @@
...
+            "memory that is proportional to the number of template names in the incoming stream of reads, and will thus require " +
+            "large amounts of memory when running on large input files.";
+
+    public static final String CONSTANT_MEMORY_DESCRPTION =
+            "Downsamples a stream or file of SAMRecords using a hash-projection strategy such that it can run in constant memory. " +
+            "The downsampling is stochastic, and therefore the actual retained proportion will vary around the requested proportion. Due " +
+            "to working in fixed memory this strategy is good for large inputs, and due to the stochastic nature the accuracy of this strategy " +
+            "is highest with a high number of output records, and diminishes at low output volumes.";
+
+    public static final String CHAINED_DESCRPTION =
```
nh13: `CHAINED_DESCRPTION` -> `CHAINED_DESCRIPTION`

src/java/htsjdk/samtools/HighAccuracyDownsamplingIterator.java
```diff
@@ -0,0 +1,196 @@
...
+     * @return true if a record is available after advancing, false otherwise
+     */
+    protected boolean advance() {
+        this.nextRecord = null;
+
+        while (this.nextRecord == null && (this.bufferedRecords.hasNext() || bufferNextChunkOfRecords(getTargetProportion(), this.targetAccuracy))) {
+            final SAMRecord rec = this.bufferedRecords.next();
+            final String key = rec.getReadName();
+            final Boolean previous = decisions.get(key);
+            final boolean keeper;
```
nh13: I can't stop thinking about a goalie, how about renaming it to `keepRecord`?

src/java/htsjdk/samtools/HighAccuracyDownsamplingIterator.java
```diff
@@ -0,0 +1,196 @@
...
+                recordDiscardedRecord(rec);
+            }
+        }
+
+        return this.nextRecord != null;
+    }
+
+    /**
+     * Buffers reads until either the end of the file is reached or enough reads have been buffered that
+     * downsampling can be performed and the target accuracy achieved.  Once reads have been buffered,
```
nh13: is achieved? can be achieved?

src/java/htsjdk/samtools/HighAccuracyDownsamplingIterator.java
```diff
@@ -0,0 +1,196 @@
...
+     * Calculates the number of templates to keep in a specific batch of reads having just read templatesRead reads
+     * and wanting to keep proportion of them.  Rounds the final number up or down based on whether, to this point,
+     * the iterator is under or over it's goal proportion.
+     *
+     * Implemented as second method to allow ChainedDownsamplingIterator to tamper with the strategy!
+     */
+    protected int calculateTemplatesToKeep(final int templatesRead, final double proportion) {
+        final double rawTemplatesToKeep = templatesRead * proportion;
+        return (keptTemplates / (double) totalTemplates < proportion)
+                ? (int) Math.ceil(rawTemplatesToKeep) : (int) Math.floor(rawTemplatesToKeep);
```
nh13: cute

nh13:
@tfenne nothing to add here but a few typos.  I am excited to take it for a spin.

build.xml
tfenne: Thanks for catching this!  I had changed it when I was experimenting with using a pair of BitSets that sucked up quite a bit of memory and forgot to change it back.

tfenne:
Thanks @nh13.  Just pushed a commit to address your comments.

@yfarjoun any objects to merging once the tests pass?  I'll get the Picard PR queued up so that it can be reviewed merged quickly.  The changes there are small, but necessary.

yfarjoun:
I still need a bit more time to get through the CHAINED stuff. I read the
other strategies. sorry for the delay.

tfenne:
No problem @yfarjoun.  Appreciate the in depth review!

tfenne:
@yfarjoun any chance you can take a look at the CHAINED stuff today so we can wrap this review?

yfarjoun:
"We're sorry, we're experiencing an unusually large volume of pull
requests. All of our operators are currently busy or on vacation. Your PR
is important to us and will be reviewed as soon as possible."

:-)

I'll try.

nh13:
All pull requests left unattended will be subject to immediate merging.  Please remove all :scissors: and :mans_shoe: from your luggage before proceeding through security.

yfarjoun:
strange test failure.

src/java/htsjdk/samtools/ChainedDownsamplingIterator.java
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @author Tim Fennell
+ */
+class ChainedDownsamplingIterator extends HighAccuracyDownsamplingIterator {
+
+    /**
+     * Constructs a chained iterator that will read from the provided iterator and attempt to downsampling to the provided proportion.
+     */
+    ChainedDownsamplingIterator(final Iterator<SAMRecord> iterator, final double proportion, final int seed) {
+        super(new ConstantMemoryDownsamplingIterator(iterator, Math.min(1, proportion + 0.01), seed), proportion, seed);
```
yfarjoun: presumably, the 1% extra is so that you will "definitely" have the amount you need, right? in that case I'd use 5sd which for the [Bernoulli](https://en.wikipedia.org/wiki/Bernoulli_distribution) is \sqrt(p(1-p)). so I'd recommend using p + 5*sqrt(p*(1-p)) instead of p+0.01. 

Good thing you allow 1 as a valid value for p, :smiley_cat:

src/java/htsjdk/samtools/ChainedDownsamplingIterator.java
```diff
@@ -0,0 +1,79 @@
...
+
+        // Deal with the fact that the iterator will advance and discard some reads at construction
+        final long discarded = ((ConstantMemoryDownsamplingIterator) getUnderlyingIterator()).getDiscardedCount();
+        for (long i=0; i<discarded; ++i) recordDiscardedRecord(null);
```
yfarjoun: sigh-spaces?

src/java/htsjdk/samtools/ChainedDownsamplingIterator.java
tfenne: Thanks - will change.  I was kind of hoping you'd tell me the right math to use here ;)  Incidentally, why 5sd and not 4 or 6?

src/java/htsjdk/samtools/ChainedDownsamplingIterator.java
tfenne: A lifetime's habit is hard to break.  Also, does anyone think that
`for (long i = 0; i < discarded; ++i)`
is actually more readable than
`for (long i=0; i<discarded; ++i)`
because I actually find it harder to read.

src/java/htsjdk/samtools/ChainedDownsamplingIterator.java
tfenne: Wait a minute - that doesn't work :/  E.g. for p=0.8, you get p(1-p) = 0.16.  sqrt(0.16) = 0.4, and 5*0.4 = 2.

tfenne: Thinking more about this, I don't think I can actually estimate a "correct" value to add to P here without knowing in advance the number of reads I will be selecting from.  I.e. I need to know both P and N to calculate standard error or a CI, and I won't know N until I've traversed the data.

src/java/htsjdk/samtools/ChainedDownsamplingIterator.java
nh13: Sadly, I find it harder to read `for (long i=0; i<discarded; ++i)`

src/java/htsjdk/samtools/ChainedDownsamplingIterator.java
yfarjoun: right...we need to take the sqrt of N as well...oops.

tfenne: Ok - how's this for a compromise.  I calculate using a fixed value of N, say 250,000 - and add to the documentation that it is accurate to x% when aiming for >= N reads?

tfenne:
@yfarjoun Did the math with an assumed N=50,000 and added to the documentation.  And added your spaces :sob:   Good to go when the tests pass?

tfenne:
Woo - and no spurious test failure this time :grin:

src/java/htsjdk/samtools/ChainedDownsamplingIterator.java
```diff
@@ -0,0 +1,92 @@
...
+        // Calculate an adjusted proportion to keep, knowing what proportion the underlying iterator discarded
+        final ConstantMemoryDownsamplingIterator iter = (ConstantMemoryDownsamplingIterator) getUnderlyingIterator();
+        final double priorProportion = iter.getAcceptedFraction();
+        final double p = Math.max(0, Math.min(1, overallProportion / priorProportion));
+        final int retval =  super.calculateTemplatesToKeep(templatesRead, p);
+
+        // Record all the discarded records to keep the overall statistics accurate, but do it after
+        // the call to super() so it doesn't affect the proportion calculation.
+        for (long i = 0; i < iter.getDiscardedCount(); ++i) {
+            recordDiscardedRecord(null);
```
yfarjoun: it might make sense to provide a recordDiscarded(int) instead of this silly loop. (and above)

yfarjoun:
small comment and then :+1:

PR #373, opened 2015-11-06
Moving the build to java 8 (and fixing one compile break).

tfenne:


tfenne:
So ... the readme says that as of the end of October 2015 we no longer support anything earlier than Java 8.  This PR is here mostly to start the discussion and see if we're actually ready and willing to push htsjdk up to Java 8.

@nh13, @eitanbanks, comments?

nh13:
:+1: from me.  Pulling @jacarey since it changes the version of the snapshot after a successful master build.

lbergelson:
I support this.  We've been talking about changing the versioning scheme to proper sementic versioning with this change.  So the new version would be 2.0.0. 

We've also been discussing switching from the combination ant/sbt build publishing system to using a single gradle build for both.

yfarjoun:
could you expand on the reasons to move from ant to gradle?

tfenne:
@lbergelson It should be noted that htsjdk is an open source project in the samtools organization, and is not owned by broadinstitute.  Any discussions about things like versioning schemes and build systems should take place somewhere *all* of the committers can be involved, and we should probably have a way for the full group to make decisions also.

lbergelson:
@yfarjoun Ant is ancient.  It's hard to read and write xml scripts.  We already are using a second build system (SBT) for publishing htsjdk since it was easier to implement that way then try to write it in ant.  This causes confusion among users and IDEs.  I recommend moving to gradle for the following reasons:
- we're successfully using gradle in hellbender already
- it's generally cleaner and easier to write code in gradle because it uses groovy, a proper scripting language that runs on the jvm instead of complex xml files
- it has really good intellij support out of the box
- it's trivial to include a gradlew wrapper script which lets people avoid having to manually install any new software
- it doesn't need a scala installation to run like SBT does

lbergelson:
@tfenne That's fair.  Where should these conversations take place?  

Anyway, I propose as part of the move to java 8, we adopt semantic versioning and move the build system to gradle.

Semantic versioning is the global best practice and I don't see a reason not to adopt it.  Moving to a new version of java seems like as good a time as any to bump to 2.0.0 and get started on the right foot.

I have a branch ( lb_gradle_build ) demonstrating a simple gradle build that replicates the essential functionality of the ant build.

tfenne:
@lbergelson I think I'd suggest that each thing be discussed in an issue, unless there's a better option.

I'm not against your suggestion of moving to semantic versioning at the same time.  Is that any more than updating the official release process?

The gradle change is somewhat larger IMHO and I'd suggest that we not couple the two as I suspect we'd all like to start using Java 8 immediately, and I think gradle may take more discussion.

akiezun:
+1 for Java8

vdauwera:
@akiezun @lbergelson We'll need to move GATK 3 up to Java 8 in parallel. There is an open issue to that effect in the repo; I suggest directing resources to doing that in the immediate future to stay in sync with htsjdk.

akiezun:
@vdauwera not as much in parallel as in sequence: htsjdk -> picard -> gatk3

lbergelson:
Lets merge this and update to 8.  Any objections?  @tfenne @yfarjoun @nh13 

It means picard / gatk will have to move up soon.

tfenne:
No objections from me.

lbergelson:
What do we have to do to notify users of a change like this?

vadimzalunin:
@lbergelson cramtools should also be fine

nh13:
@lbergelson probably to the various samtools mailing list at least.

eitanbanks:
@jrobinso just making sure you see this.

jrobinso:
Thanks Eric,  I knew about this but didn't see this git thread.  I will 
have to keep IGV frozen on the current htsjdk until we force igv users 
to java 8.   We haven't made a decision about that yet.

Jim

lbergelson:
@tfenne or @nh13 Could one of you write the announcement to the mailing list?  I don't know what the protocol is.

jmarshall:
Compare https://sourceforge.net/p/samtools/mailman/message/33629377/ .  Only certain anointed people are able to post to samtools-announce; if you want to post this there too, let me know who at Broad should be added to the list.

alecw:
Hi Folks,

Just wondering -- switching to Java 8 is a change that will have significant impact on a lot of users.  Will it be announced anywhere?  Back in the day, major changes like this would be announced on samtools-announce.  Unless I'm missing some other forum, the only way to know that something like this is coming is to watch all the notifications coming out of this repo, which is a huge stream that one quickly starts ignoring.  Could a topic be created in the GATK forum for Picard & HTSJDK announcements, and a message posted when something like this is coming down the pike?

Thanks, Alec

tfenne:
@alecw It should definitely be announced, but it's also been prominently displayed on the project homepage [here](https://github.com/samtools/htsjdk) for some time.

tfenne:
All - I'm going to merge this PR shortly unless someone jumps in and requests otherwise.  I will also send an announcement email to the samtools-* mailing lists letting folks know that this is happening, and edit the project homepage to make it clear the move has happened.

tfenne:
Have sent emails, and just realized in the process that I haven't been subscribed to the samtools-* lists since I left Broad.  Whoops!  I think my messages are stuck in a moderation queue :unamused: .  @jmarshall can you please me out please?

jmarshall:
Email posted on -devel, and now approved on -announce (and the email address you sent it from is now suitably anointed) — though perhaps I should have rejected it for effect/affect abuse :smile:

tfenne:
Thanks John.  Correct usage of effect/affect is one thing I never seem to be able to remember.

akiezun:
link here for posterity and reference: http://sourceforge.net/p/samtools/mailman/message/34644882/

lbergelson:
image

PR #376, opened 2015-11-09
the disk backed queue should set "canAdd" to false whenever we have

nh13:
spilled to disk and have started removing records from RAM, or whenever
we start removing records from disk.  The pathology was that we would
let callers add if there was any room in ram, regardless if we had
spilled records to disk, thus causing records to become out of order.
If the entire queue is empty, we set canAdd to true again.  Added test
to make sure an exception is thrown if we add when we should not.

See: https://github.com/broadinstitute/picard/issues/327

@bradtaylor

### DIFF
--- a/src/java/htsjdk/samtools/util/DiskBackedQueue.java
+++ b/src/java/htsjdk/samtools/util/DiskBackedQueue.java
@@ -25,6 +25,7 @@
 package htsjdk.samtools.util;
 
 import htsjdk.samtools.Defaults;
+import htsjdk.samtools.SAMException;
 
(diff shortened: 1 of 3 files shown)


tfenne:
OMFG the statefulness in these tests is really sucky.  Think I've finally got it this time. :hammer:

tfenne:
@nh13 I reviewed the code and the new test and think this looks good.  I also fixed up the failing test (it was a nasty interdependency between test classes and tmp directory contents, not a problem with the code being tested).  Merging now.

PR #382, opened 2015-11-10
Use buffered streams in binary codec to be much faster on network drives

BeatWolf:
When converting an unsorted SAM to a sorted BAM with index creation on a NAS i noticed that the time it took was longer than expected.
Profiling revealed that almost all the time was spent in the BinaryCodec write methods.

Using buffered streams cuts the time required for that operation in half. 
This is for my use case where the files where stored on a NAS. The speed gain on a local HD or SSD is much lower.
But as i often use code like that where the data is not stored locally, but on a network drive, this will help to speed up some operations.

### DIFF
--- a/src/java/htsjdk/samtools/util/BinaryCodec.java
+++ b/src/java/htsjdk/samtools/util/BinaryCodec.java
@@ -101,10 +101,10 @@ public class BinaryCodec implements Closeable {
         try {
             this.isWriting = writing;
             if (this.isWriting) {
-                this.outputStream = new FileOutputStream(file);
+                this.outputStream = IOUtil.maybeBufferOutputStream(new FileOutputStream(file));
                 this.outputFileName = file.getName();
             } else {
-                this.inputStream = new FileInputStream(file);
+                this.inputStream = IOUtil.maybeBufferInputStream(new FileInputStream(file));
                 this.inputFileName = file.getName();
             }
         } catch (FileNotFoundException e) {


tfenne:
You should probably use IOUtil.maybeBuffer[Output|Input]Stream vs. locally creating a buffered stream with Defaults.BUFFER_SIZE as this will correctly handle the case where buffer size is set to 0 or some invalid number.

tfenne:
Other than that :+1:  I do have some unsettling memory that maybe we left BinaryCodec unbuffered for a reason, but all the tests pass and I can't remember why that would be.

BeatWolf:
Modified the pull request according to your feedback.

BeatWolf:
The travis build failed on this unrelated test:
   [testng] FAILED: testNonExistentFile
   [testng] java.lang.AssertionError: expected:<true> but was:<false>
   [testng] 	at htsjdk.tribble.util.ftp.FTPClientTest.testNonExistentFile(FTPClientTest.java:155)

Looking at the code i would expect that there was a communication error with the ftp.broadinstitute.org server during the test.

Is there a way to relaunch the travis check for this pull request?

tfenne:
Relaunched @BeatWolf

BeatWolf:
Again the tests seem to fail when contacting that FTP server.
I could not find any connection between the failed test and the modified code. I guess the ftp server has some issues. Maybe it will work better tomorrow. All tests pass here locally.

yfarjoun:
restarted.

yfarjoun:
passed.

PR #383, opened 2015-11-10
replacing ant build with gradle build

lbergelson:
Major changes to htsjdk's build:
- updating to java 8
- adding a build.gradle
  - restructuring repo to use gradle default style:
    - src/java -> src/main/java
    - src/tests/java -> src/test/java
    - test-data -> src/test/resources
- removing ant's build.xml
- removing etc/ because it looks like it's for help formatting the test results nicely, and gradle does nice job already
- updating .travis.yml to use gradle
- adding gradlew and gradlew.bat wrapper files, these let users without gradle run ./gradlew instead
- removing intellij and eclipse project files
- deleting lib because we're pulling down our dependencies from maven
- removing intel deflator because it's comments say it's not needed in java 8
  - removing associated c and jni code
- update gitignore with gradle relevant directories

Things not done yet:
- remove sbt
I'd like to remove the sbt code entirely and replace with equivalent gradle code.  This will probably take some coordination with the @jacarey since he handles all the releases and I don't know the process exactly / have signing keys 
- update documentation 
I don't know how to update the github-io site, can that be done from within a pull request?
- building picard on the gradle version
We need to try this to make sure it will work.  Probably take some minor changes to the picard build to make sure jar paths are correct.
- probably other things...

[Review on Reviewable](https://reviewable.io/reviews/samtools/htsjdk/383)


lbergelson:
Quick start:

```
gradle clean
gradle assemble
gradle test
gradle test -Dtest.single=TestClass
gradle javadoc
```

lbergelson:
Easiest way to make a project in intellij from this:
1:
screen shot 2015-11-10 at 1 19 52 pm

2:
screen shot 2015-11-10 at 1 20 05 pm

3:
screen shot 2015-11-10 at 1 20 14 pm

4:
screen shot 2015-11-10 at 1 20 21 pm

lbergelson:
@nh13 @tfenne @droazen @yfarjoun 

I've assigned this to Nils for a first pass, but feedback from everyone is probably important for a major change like this.

nh13:
@lbergelson why not use `gradle idea` task to create the various files for intellij?

nh13:
`src/scripts/release_picard.sh` needs to be updated or removed.  Check with the folks who release Picard and HTSJDK.

lbergelson:
@nh13 I tried using `gradle idea` but the project setup wasn't quite right, intellij load worked better.  I meant to remove the idea plugin but forgot to.

lbergelson:
@jacarey do people use release_picard.sh?

nh13:
A few things before looking at the file changes:
* javadoc final build location has changed: `./javadoc/` -> `./build/docs/javadoc/`
* location of final htsjdk jar has changed: `dist/` -> `build/libs/`
* this will require an update to Picard to clone and build properly from htsjdk.
* missing equivalent of the ant target `write-version-property`, which is used by other folks like Picard, to get the version of htsjdk.
* some tests failed, I think, but it is unclear if they did, with no stack trace.  I suppose the following message means that a test had a `System.err.print` call?
```
Test: Test suite 'Gradle test' produced standard out/err: In SortingCollectionTest.tearDown.  tmpDir: /var/folders/t5/qf7g7gdd2tv7_6n8_918flhr0000gn/T/nhomer/SortingCollectionTest

Gradle suite > Gradle test STANDARD_ERROR
    In SortingCollectionTest.tearDown.  tmpDir: /var/folders/t5/qf7g7gdd2tv7_6n8_918flhr0000gn/T/nhomer/SortingCollectionTest

BUILD SUCCESSFUL
```

gradlew
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env bash
```
nh13: Did you write this yourself, or is does this come from somewhere?  If the latter, could you add a comment with a link.

gradlew.bat
```diff
@@ -0,0 +1,90 @@
+@if "%DEBUG%" == "" @echo off
```
nh13: Same comment as for the Un*x startup script.

src/main/java/htsjdk/samtools/Defaults.java
```diff
@@ -30,9 +30,6 @@
      */
     public static final int NON_ZERO_BUFFER_SIZE;
 
-    /** Should BlockCompressedOutputStream attempt to load libIntelDeflater? */
-    public static final boolean TRY_USE_INTEL_DEFLATER;
```
nh13: Why is this removed?   Do not remove.

src/c/inteldeflater/IntelDeflater.c
```diff
@@ -1,247 +0,0 @@
-/*
```
nh13: Why did this get removed?  The IntelDeflator speeds up deflation and is really quite useful.  Do not remove.

src/java/htsjdk/samtools/util/zip/DeflaterFactory.java
```diff
@@ -1,78 +0,0 @@
-/*
```
nh13: Please restore me.

src/java/htsjdk/samtools/util/zip/IntelDeflater.java
```diff
@@ -1,567 +0,0 @@
-/*
```
nh13: I want to live.

src/java/htsjdk/samtools/util/zip/ZStreamRef.java
```diff
@@ -1,48 +0,0 @@
-/*
```
nh13: Is this no longer needed because of Java 8? Otherwise the IntelDeflator needs this.

nh13:
@lbergelson back to you.

gradlew
lbergelson: these are generated by calling `gradle wrapper`  I'll add a comment

gradlew.bat
lbergelson: is htsjdk supported on windows?  If not I'll jsut remove this wrapper.

lbergelson:
@nh13 

I removed the intel deflator stuff because of this comment in the intel deflator:
```
/**
 * Create zlib-based Deflater if JNI library and other require libraries are available, otherwise create standard
 * JDK Deflater.
 * Java 7 has its own Deflater implementation (libzip.so).  This is almost as fast as a zlib-based Deflater, so in general
 * there isn't a compelling reason to use zlib.  However, Intel has created a hardware-assisted zlib implementation
 * as part of their IPP (Integrated Performance Primitives) package that can run significantly faster on some Intel
 * hardware.  We have seen compression times reduced by 13% to 33% depending on particular hardware, and hope that
 * newer Intel processors will be even better.
 *
 * Note that this class will no longer be necessary once Java 8 is required, because JDK 8 will use zlib instead
 * of libzip implementation.
 */
```

If you believe it's not true I'll replace it.

lbergelson:
That test output your seeing is that test writing some garbage to stderr.  We can either not output test stderr if it's confusing, or fix that test.  Is that the reason you say some tests are failing?

nh13:
@lbergelson can you confirm that JDK 8 uses zlib and not libzip?  My google-fu is not up to par today.  I was confused by the gradle output since it was my first time parsing it, so I don't see any tests failing.  Is there a # of tests count, so we can verify the same number of tests were performed?

lbergelson:
Hmn, the most conclusive thing I see is the (zlib's faq)[http://www.zlib.net/zlib_faq.html#faq34]

>Is there a Java version of zlib?
Probably what you want is to use zlib in Java. zlib is already included as part of the Java SDK in the java.util.zip package. If you really want a version of zlib written in the Java language, look on the zlib home page for links: http://zlib.net/ . 

It's possible that the intel deflator is even better than the java bundled zlib though, and the comment is bogus, in which case we should keep it.  @alecw Do you have any idea about this?  It looks like you added that comment.

lbergelson:
The best thing to do would be to test and see if one is faster than the other.  Do we have any benchmarks for this sort of thing?

lbergelson:
I talked to one of the intell guys in the office and they say they have even better optimized versions of zlib, and zlib specifically for bam encoding that has 3x speedup.  I sounds like those are newer than what we're including here though, so it maybe be worth upgrading the existing .so to a newer one.

nh13:
@lbergelson see: https://software.intel.com/en-us/articles/igzip-a-high-performance-deflate-compressor-with-optimizations-for-genomic-data

lbergelson:
@nh13 Is that what we're including already? Or is that a new thing that we should add?

alecw:
Hi @lbergelson ,

Two things to consider when reading my comment in the code quoted above regarding libIntelDeflater.

* I may have been confused about the situation;
* The situation may have changed.

IIRC, there were 2 reasons that together obviated the need for LibIntelDeflater in order to get the hardware-assisted gzip speed-up:

* Java 8 was going to use libz.so to do deflating;
* Intel distributed a version of libz.so that was compatible with the standard one but did the hardware assist where possible.

If the above is true, then at minimum users would need to be informed of where to get the Intel libz, and how to install it (i.e. put it on LD_LIBRARY_PATH ahead of regular libz).  

I suggest you reach out to Paolo Narvaez at Intel who facilitated this work originally and has been super-helpful in the past.  He should be able to clear up any confusion.

-Alec

lbergelson:
@alecw Thanks for your input.  I figured a comment from several years ago may not be fully reliable, but I figured I'd ask.

nh13:
@lbergelson igzip and IPP are two different things.  Paolo Narvaez and team made the former, why IPP is large and highly-supported project.

tfenne:
@lbergelson What's the latest on this PR?  I'd like to get this and/or the Java 8 PR resolved.  Anything I can do to help?

I just checked out the branch and built and ran the tests.  All seemed pretty straightforward to me.  A few thoughts:

1. I would vote for killing at least `gradlew.bat` and probably also `gradlew`.  It's very easy to install gradle and personally I'd prefer not to have those wrappers cluttering up the root level of the repo.
2. It's a long-standing annoyance, but if it would be possible to supply -Djava.awt.headless=true to the VM running the tests that would be very nice
3. The unit test report is much nicer, both aesthetically and functionally
4. I think the quick start should include information on either setting up the gradle daemon, or making it shut up about the daemon.  I chose the latter:
```bash
touch ~/.gradle/gradle.properties && echo "org.gradle.daemon=false" >> ~/.gradle/gradle.properties
```
5. On the Intel IPP/igzip front, my hazy recollection is in line with what has been said above.  Specifically: a) igzip is a different, custom deflater. b) Java 8 is supposed to use zlib and c) On systems with IPP installed that should be all that is necessary.  However, I don't have access to a system with IPP installed to verify :/  You might try running something *very* simple like SamFormatConverter with JDK8 with the pre-gradle +IntelCompressor version and the post-gradle -IntelCompressor version, but writing at high compression level to /dev/null and see if there is any significant difference in runtime.

lbergelson:
@tfenne I think we should merge your branch as soon as possible.  
1.   I agree with killing gradlew.bat, we don't really support windows anyway.  I'd advise we keep gradlew itself though.  It's a little bit of clutter, but it effectively versions the build system along with project.  It means changes to gradle versions will automatically propagate to travis, IDE's, etc.  It's also pretty useful for things like building on a remote server that doesn't have gradle installed already.  For instance, the broad servers that don't have up to date gradle dotkits.  
2.   Easy to add `-Djava.awt.headless=true`.  What's it for?  I didn't think we had any UI in htsjdk?
3.  Glad you like it.
4.  I find the daemon pretty useful.  It dramatically reduces the annoying lag on gradle startup.  We should definitely include instructions on how to enable / silence it though.

I'd like to run the relative speed tests.  I'm confused about what I need to do to get the intel custom deflator working on my computer. If I build the old version it doesn't seem to use the it.  Even if I manually specify the system property to tell it where it is it still uses the jdk deflator.

```
$ java  -Dsamjdk.intel_deflater_so_path=htsjdk/lib/jni/libIntelDeflater.so -jar dist/picard.jar SamFormatConverter I=testdata/picard/sam/aligned.sam O=out.bam
[Tue Nov 24 11:49:37 EST 2015] picard.sam.SamFormatConverter INPUT=testdata/picard/sam/aligned.sam OUTPUT=out.bam    VERBOSITY=INFO QUIET=false VALIDATION_STRINGENCY=STRICT COMPRESSION_LEVEL=5 MAX_RECORDS_IN_RAM=500000 CREATE_INDEX=false CREATE_MD5_FILE=false GA4GH_CLIENT_SECRETS=client_secrets.json
[Tue Nov 24 11:49:37 EST 2015] Executing as louisb@wm1b0-8ab on Mac OS X 10.9.5 x86_64; Java HotSpot(TM) 64-Bit Server VM 1.8.0_45-b14; Picard version: 1.141(8ece590411350163e7689e9e77aab8efcb622170_1447695087) JdkDeflater
[Tue Nov 24 11:49:37 EST 2015] picard.sam.SamFormatConverter done. Elapsed time: 0.00 minutes.
Runtime.totalMemory()=257425408
```

What do I need to do to use the old intel deflator?

vadimzalunin:
@lbergelson @tfenne I'd prefer to keep gradlew.bat

lbergelson:
@vadimzalunin Do you build htsjdk on windows?  I didn't know it even worked.

tfenne:
@lbergelson Ok - I'll pick up and push the other PR for Java 8, and send out some announcements etc.

Re: headless - I'm not sure what it is, but something in the test suite does something that causes a java icon to jump up in my doc and steal focus from whatever I'm doing.

On the intel deflator, I *think* that only ever worked on linux.  I only ever got it working by putting the libIntelDeflator.so in the same directory alongside the picard.jar or gatk.jar that was being invoked.

lbergelson:
@tfenne Ahah. I've seen that icon.  I wonder what causes it.  

I'll try running on linux I guess.

alecw:
Yes, libIntelDeflator.so has always been Linux-only.  IIRC, there are various ways it can be located, e.g. LD_LIBRARY_PATH, java property, but the standard way is having it in the same directory as the jarfile containing the code that loads it as Tim wrote below.

akiezun:
@nh13 @lbergelson what's the status here? are we moving forward with gradle?

akiezun:
@nh13 @droazen are we moving forward here? if not, what needs to be done?

nh13:
Happy to help review but I have zero experience with Gradle. Please mention me if you need a review.

vadimzalunin:
i'm not a gradle pro but we do use gradle internally, so perhaps I could help as well

lbergelson:
This is stuck because I haven't gotten a chance to update the picard build script to be aware of this change, and I didn't want to merge this until we were sure picard / picard private would be unaffected.

droazen:
@lbergelson Is there a picard ticket to track that part of the gradle migration?

droazen:
We agreed that the order of operations here would be as follows:

1. @lbergelson will rebase & squash this branch (or start a new from scratch). (Make sure not to remove the intel deflator!)
2. @jacarey change `build.xml` to account for the new paths, and confirm that ant still works
3. make sure travis tests both ant and gradle
4. add build instructions to the htsjdk README
5. merge this
6. (in a separate task) migrate picard to gradle
5.

bpow:
If/when PR #533 is accepted, the dependency on org.apache.ant:ant:1.8.2 listed in the build.gradle should be removed.

yfarjoun:
please remember to check the various pathways through the ant:
- documentation building
- single-test (or is it test-single?)
- etc.

lbergelson:
@yfarjoun 
1. Documentation is built.  I'm not sure who consumes it other than attaching a documentation jar to the artifact.  It ends up in a different location than before, so if something is relying on it being in 1 place then we'll have to update that or copy it somewhere.

Single test running is accomplished by running test with the `--tests "*WhateverTestClass"`

Are there other ant tasks that are important?

lbergelson:
@jacarey This is now ready for you to update the ant build.  The gradle build now does to the current best of my knowledge does everything that the ant/sbt builds did before.  (ignoring picard)

I didn't delete the files in `/lib` yet, but I would recommend doing so or we'll get out of sync.  (leave the intelDeflater though).  I provided `./gradlew copyLibsForPicard` which when run will copy the dependencies that gradle downloads into the lib folder.  Might be helpful for the transitionary stage.

Feel free to add additional tasks to the gradle build if you need them (or ask me if you want a hand adding something specific)  

It might be more worthwhile to build picard off the gradle build instead of the ant one as a transitionary step.  I'll leave it up to you though.  It would be good to remove the ant build as soon as possible though I think.

jacarey:
@lbergelson The ant build in picard is ready (https://github.com/broadinstitute/picard/pull/533) The PR will fail until this is merged. Once that happens I'll rerun build and tests.

lbergelson:
@nh13 @droazen @yfarjoun @jacarey @tfenne 

This is now ready in what I believe will be it's final form unless there are objections.  It's 2 commits, 1 is a move of all the files and the second is the actual inclusion of gradle stuff.  That's necessary to make git rename tracking work, otherwise it loses track of some of the test files that had many paths changed.  (it should also make it easier to review if anyone wants to since all the code changes should be in commit 2.

Once this is merged I'll send an email to the user group as well as updating the gh-pages documention.  

There is still a build.xml which prints a large deprecation message with gradle usage instructions.  So people should have some information about what's going on when they update.  It will break people's intellij files, but those are now trivial to create since intellij understand gradle much better than it did ant.

gradlew
lbergelson: Hmn... Just realized this is back to being unaddressed... The funny thing about automated scripts is that they tend to be overwritten automatically...  I'll have to figure out the hook to modify it.

yfarjoun:
I think that the Readme.md should be updated to explain how to setup the intelliJ, rather than only have the explanation in the commit.

akiezun:
@lbergelson the compile and IDE setup works perfectly. The tests fail on the Mac because the test selection is not ported from Ant (see the sections of build.xml about excluded tests groups). Can you fix those?

lbergelson:
@akiezun Ah, thanks for pointing that out.  I did actually port the exclusions, but there was funny handling of inteldeflater test and it wasn't marked as a unix test.  Fixing.

lbergelson:
@yfarjoun Typically all the documentation is on the GH pages site, do you want me to include it in the readme instead?

README.md
```diff
@@ -15,6 +15,61 @@ Please see the [HTSJDK Documentation](http://samtools.github.io/htsjdk) for more
...
 
+#### Building HTSJDK
+
+HTSJDK is now built using [gradle](http://gradle.org/).
+
+A wrapper script (`gradlew`) is included which will download the appropriate version of gradle on the first invocation.
+
+Example gradle usage:
+
+ - compile htsjdk or it's tests
```
droazen: it's -> its

README.md
```diff
@@ -15,6 +15,61 @@ Please see the [HTSJDK Documentation](http://samtools.github.io/htsjdk) for more
...
+
+ - compile htsjdk or it's tests  
+ ```
+ ./gradlew compileJava
+ ./gradlew compileTest
+ ```
+ 
+ - build a jar
+ ```
+ ./gradlew jar
```
droazen: Can we add a default target like we had in ant? Typing "ant" would do a standard build of the library -- could typing "./gradlew" (no args) do a standard build as well (eg., invoke "jar")?

README.md
```diff
@@ -15,6 +15,61 @@ Please see the [HTSJDK Documentation](http://samtools.github.io/htsjdk) for more
...
+ - build a jar that packages all of htsjdk's dependencies in a single jar
+ ```
+ ./gradlew shadowJar
+ ```
+
+ - run tests, a specific test class, or run a test and wait for the debugger to connect
+ ```
+ ./gradlew test
+ ./gradlew test --tests "*AlleleUnitTest"
+ ./gradlew test --tests "*AlleleUnitTest" --debug-jvm
```
droazen: How do you tell the debugger the port to listen on?

README.md
```diff
@@ -15,6 +15,61 @@ Please see the [HTSJDK Documentation](http://samtools.github.io/htsjdk) for more
...
+
+ - build a jar that packages all of htsjdk's dependencies in a single jar
+ ```
+ ./gradlew shadowJar
+ ```
+
+ - run tests, a specific test class, or run a test and wait for the debugger to connect
+ ```
+ ./gradlew test
+ ./gradlew test --tests "*AlleleUnitTest"
```
droazen: Add a non-wildcard example as well

README.md
```diff
@@ -15,6 +15,61 @@ Please see the [HTSJDK Documentation](http://samtools.github.io/htsjdk) for more
...
+ ./gradlew clean
+ ```
+
+ - see an exhaustive list of all available targets
+ ```
+ ./gradlew tasks
+ ```
+
+#### Create an HTSJDK project in Intellij
+To create a project in Intellij IDE for htsjdk do the following:
```
droazen: Intellij -> IntelliJ

README.md
```diff
@@ -15,6 +15,53 @@ Please see the [HTSJDK Documentation](http://samtools.github.io/htsjdk) for more
...
+
+HTSJDK is now built using [gradle](http://gradle.org/).
+
+A wrapper script (`gradlew`) is included which will download the appropriate version of gradle on the first invocation.
+
+Example gradle usage from the htsjdk root directory:
+ - compile and build a jar
+ ```
+ ./gradlew
+ ./gradlew jar
```
droazen: Make it clear that this is an `or` relationship, not an `and` relationship

droazen: And document where the jar ends up

README.md
```diff
@@ -15,6 +15,53 @@ Please see the [HTSJDK Documentation](http://samtools.github.io/htsjdk) for more
...
+
+ ./gradlew test --tests "*AlleleUnitTest" --debug-jvm
+ ```
+
+ - clean the project directory
+ ```
+ ./gradlew clean
+ ```
+
+ - build a jar that includes all of htsjdk's dependencies
```
droazen: jar -> monolithic jar

src/test/java/htsjdk/samtools/DownsamplingIteratorTests.java
```diff
@@ -23,10 +22,14 @@
         put(Strategy.ConstantMemory, 0.01);
     }};
 
+    private static Random getRandom(){
```
droazen: Add a comment explaining why you need to use this particular seed

.gitignore
```diff
@@ -1,16 +1,13 @@
...
-/bin
 /test-output
-target
-.idea/libraries
-.idea/workspace.xml
\ No newline at end of file
+
+#intellij
+.idea/
+src/htsjdk.iml
```
droazen: Ignore *.ipr as well

droazen: And *.iws

droazen: And *.iml

droazen:
:+1: LGTM

akiezun:
wow, a new era begins!

PR #384, opened 2015-11-12
fix for issue 364: cram roundtrip should preserve tlen

vadimzalunin:


droazen:
:+1: looks good -- merging!

PR #385, opened 2015-11-15
javascript filters from SAM and VCF

lindenb:
from picard :  https://github.com/broadinstitute/picard/pull/335

> Could you move the filters into htsjdk?

here are the new sources:
* wrote an abstract javascript filter and two implementations for SAM and VCF
* wrote two Tests for both format as well as some javascript files used by the test.

### DIFF
--- a/src/java/htsjdk/samtools/filter/AbstractJavascriptFilter.java
+++ b/src/java/htsjdk/samtools/filter/AbstractJavascriptFilter.java
@@ -1,0 +1,159 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015 Pierre Lindenbaum @yokofakun Institut du Thorax - Nantes - France
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package htsjdk.samtools.filter;
+
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.Reader;
+import java.io.StringReader;
+
+import javax.script.Bindings;
+import javax.script.Compilable;
+import javax.script.CompiledScript;
+import javax.script.ScriptEngine;
+import javax.script.ScriptEngineManager;
+import javax.script.ScriptException;
+import javax.script.SimpleBindings;
+
+import htsjdk.samtools.util.CloserUtil;
+import htsjdk.samtools.util.RuntimeScriptException;
+
+/**
+ * Javascript filter with HEADER type containing TYPE records. contains two
+ * static method to get a SAM Read filter or a VariantFilter.
+ * 
+ * warning: tools, like galaxy, using this class are not safe because a script
+ * can access the filesystem.
+ * 
+ * @author Pierre Lindenbaum PhD
+ */
+public abstract class AbstractJavascriptFilter<HEADER, TYPE> {
+    public static final String DEFAULT_HEADER_KEY = "header";
+    /** compiled user script */
+    private CompiledScript script = null;
+
+    /** javascript bindings */
+    protected Bindings bindings;
+
+    /**
+     * constructor using a java.io.File script, compiles the script, puts
+     * 'header' in the bindings
+     */
+    protected AbstractJavascriptFilter(final File scriptFile, final HEADER header) throws IOException {
+        this(new FileReader(scriptFile), header);
+    }
+
+    /**
+     * constructor using a java.lang.String script, compiles the script, puts
+     * 'header' in the bindings
+     */
+    protected AbstractJavascriptFilter(final String scriptExpression, final HEADER header) {
+        this(new StringReader(scriptExpression), header);
+    }
+
+    /**
+     * Constructor, compiles script, put header in the bindings
+     * 
+     * @param scriptReader
+     *            reader containing the script. will be closed.
+     * @param header
+     *            the header to be injected in the javascript context
+     */
+    protected AbstractJavascriptFilter(final Reader scriptReader, final HEADER header) {
+        final ScriptEngineManager manager = new ScriptEngineManager();
+        /* get javascript engine */
+        final ScriptEngine engine = manager.getEngineByName("js");
+        if (engine == null) {
+            CloserUtil.close(scriptReader);
+            throw new RuntimeScriptException("The embedded 'javascript' engine is not available in java. "
+                    + "Do you use the SUN/Oracle Java Runtime ?");
+        }
+        if (scriptReader == null) {
+            throw new RuntimeScriptException("missing ScriptReader.");
+        }
+        
+        try {
+            final Compilable compilingEngine = getCompilable(engine);
+            this.script = compilingEngine.compile(scriptReader);
+        } catch (ScriptException err) {
+            throw new RuntimeScriptException("Script error in input", err);
+        } finally {
+            CloserUtil.close(scriptReader);
+        }
+
+        /*
+         * create the javascript bindings and put the file header in that
+         * context
+         */
+        this.bindings = new SimpleBindings();
+        this.bindings.put(DEFAULT_HEADER_KEY, header);
+    }
+
+    /** return a javascript engine as a Compilable */
+    private static Compilable getCompilable(final ScriptEngine engine) {
+        if (!(engine instanceof Compilable)) {
+            throw new IllegalStateException("The current javascript engine (" + engine.getClass()
+                    + ") cannot be cast to Compilable. " + "Do you use the SUN/Oracle Java Runtime ?");
+        }
+        return Compilable.class.cast(engine);
+    }
+
+    /** returns key used for header binding */
+    public String getHeaderKey() {
+        return DEFAULT_HEADER_KEY;
+    }
+
+    /** returns key used for record binding */
+    public abstract String getRecordKey();
+
+    /**
+     * Evaluates this predicate on the given argument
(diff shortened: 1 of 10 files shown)


src/tests/java/htsjdk/samtools/filter/JavascriptSamRecordFilterTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+	@Test(dataProvider = "jsData")
+	public void testJavascriptFilters(final String samFile, final String javascriptFile) {
+		final SamReaderFactory srf = SamReaderFactory.makeDefault();
+		final SamReader samReader = srf.open(new File(testDir, samFile));
+		final JavascriptSamRecordFilter filter = new JavascriptSamRecordFilter(new File(testDir, javascriptFile),
+		        samReader.getFileHeader());
+		final SAMRecordIterator iter = samReader.iterator();
+		while (iter.hasNext()) {
+			filter.filterOut(iter.next());
```
yfarjoun: it would be good to check that you actually get some records out...do you know how many records you expect out of this iterator? might be nice to assert that..

src/tests/java/htsjdk/variant/variantcontext/filter/JavascriptVariantFilterTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+	@Test(dataProvider = "jsData")
+    public void testJavascriptFilters(final String vcfFile, final String javascriptFile)
+    	{
+    	final File vcfInput = new File(testDir, vcfFile);
+    	final File jsInput = new File(testDir, javascriptFile);
+        final VCFFileReader vcfReader = new VCFFileReader(vcfInput, false);
+    	final JavascriptVariantFilter filter = new JavascriptVariantFilter(jsInput, vcfReader.getFileHeader());
+    	final FilteringIterator iter = new FilteringIterator(vcfReader.iterator(), filter);
+    	while(iter.hasNext()) iter.next();
```
yfarjoun: same here, could you verify that the "right" amount of records passes through the filter?

src/java/htsjdk/samtools/filter/AbstractJavascriptFilter.java
```diff
@@ -0,0 +1,168 @@
...
+	 *            javascript source should be either a File or a String
+	 * @param header
+	 *            the header to be injected in the javascript context
+	 */
+	private AbstractJavascriptFilter(final HEADER header, final Object scriptFileOrString) {
+		final ScriptEngineManager manager = new ScriptEngineManager();
+		/* get javascript engine */
+		this.engine = manager.getEngineByName("js");
+		if (this.engine == null) {
+			throw new RuntimeException("The embedded 'javascript' engine is not available in java. "
```
yfarjoun: There were some comments regarding this class in the picard PR. please address them here.

src/java/htsjdk/samtools/filter/JavascriptSamRecordFilter.java
```diff
@@ -0,0 +1,77 @@
...
+	/** 
+	 * constructor using a javascript expression
+	 * @param scriptExpression the javascript expression to be compiled
+	 * @param header the SAMHeader
+	 */
+	public JavascriptSamRecordFilter(final String scriptExpression, final SAMFileHeader header) {
+		super(scriptExpression,header);
+	}
+	
+	/** return true of both records are filteredOut (AND) */
```
yfarjoun: of->if

yfarjoun:
Thanks!

It seems that you didn't address the issues that were noted on this PR when it was in picard. Could you address them please?

lindenb:
@yfarjoun There were so many comments :-) Let me see.


* https://github.com/broadinstitute/picard/pull/335#discussion_r44482967 fix text / specific to picard
* https://github.com/broadinstitute/picard/pull/335#discussion_r44483063 assertion / specific to picard
* https://github.com/broadinstitute/picard/pull/335#discussion_r44483128 author / done
* https://github.com/broadinstitute/picard/pull/335#discussion_r44483192 rename DEFAULT_HEADER_KEY / done
* https://github.com/broadinstitute/picard/pull/335#issuecomment-155603993 moved to VarCtxFilter , move to htsjdk / done 
* https://github.com/broadinstitute/picard/pull/335#discussion_r44483461 fix 'header' / done
* https://github.com/broadinstitute/picard/pull/335#discussion_r44483617 +  https://github.com/broadinstitute/picard/pull/335#discussion_r44483725 + https://github.com/broadinstitute/picard/pull/335#discussion_r44483765 simplify boolean test / done
* https://github.com/broadinstitute/picard/pull/335#discussion_r44680754 + https://github.com/broadinstitute/picard/pull/335#discussion_r44682617 asked for a javscript exception subclass of picardException .. for now I'm throwing a RuntimeException. Should I subclass a specific Throwable of htsjdk ? 
* https://github.com/broadinstitute/picard/pull/335#discussion_r44483872 empty line /done
* https://github.com/broadinstitute/picard/pull/335#discussion_r44483921 complete javadoc: forgot that one... and... done.
* https://github.com/broadinstitute/picard/pull/335#discussion_r44483937 + https://github.com/broadinstitute/picard/pull/335#discussion_r44484480 + https://github.com/broadinstitute/picard/pull/335#discussion_r44484497 VariantContextFilter /done
* https://github.com/broadinstitute/picard/pull/335#discussion_r44483950 fix javadoc/ done as I've subclassed the abstract filter
* https://github.com/broadinstitute/picard/pull/335#discussion_r44484416 fix logic / done in JavascriptSamRecordFilter
* https://github.com/broadinstitute/picard/pull/335#discussion_r44484708 + https://github.com/broadinstitute/picard/pull/335#discussion_r44484866 + https://github.com/broadinstitute/picard/pull/335#discussion_r44741821 + https://github.com/broadinstitute/picard/pull/335#discussion_r44741679 related to picard
* https://github.com/broadinstitute/picard/pull/335#discussion_r44682462 throw illegalArgumentException / done 
* https://github.com/broadinstitute/picard/pull/335#discussion_r44682573 throw RuntimeIOException / done
* https://github.com/broadinstitute/picard/pull/335#discussion_r44684557 remove script engine / forgot that one... and ... done.
* https://github.com/broadinstitute/picard/pull/335#discussion_r44685130 Compilable caster / done
* https://github.com/broadinstitute/picard/pull/335#discussion_r44685520 VariantFilter / uneeded,ignore
* https://github.com/broadinstitute/picard/pull/335#discussion_r44687071 move js compilation in the ctor / done
* https://github.com/broadinstitute/picard/pull/335#discussion_r44687727 specific to picard

I've just added a count for each test

yfarjoun:
Thanks pierre!



On Mon, Nov 16, 2015 at 4:21 PM, Pierre Lindenbaum <notifications@github.com

lindenb:
Ops, I didn't see there was an error in Travis & it took some times before I found the error:

There was a problem in the test file testdata/htsjdk/samtools/samFilter02.js

In java 1.7 + rhino/javascript `String.charAt()` returns an integer while it returns a String in java 1.8  that's why the javascript-test failed. Fixed by using `substring` instead of `charAt` :

```js

/** accept record if second base of DNA is a A */
function accept(r)
	{
        return r.getReadString().length()>2 &&
		r.getReadString().substring(1,2)=="A";
	}
```

lindenb:
@yfarjoun  is it ok to merge ?

yfarjoun:
sorry for the delay,

@lbergelson were all your comments addressed? could you advise regarding the exception?

src/java/htsjdk/samtools/filter/AbstractJavascriptFilter.java
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * Constructor, compiles script, put header in the bindings
+     * 
+     * @param scriptFileOrObject
+     *            javascript source should be either a File or a String
+     * @param header
+     *            the header to be injected in the javascript context
+     */
+    private AbstractJavascriptFilter(final HEADER header, final Object scriptFileOrString) {
```
lbergelson: Could you change this so that it accepts a specific type instead of object (either `File`, `String`, or `Path` which is probably the new standard now that we're  on java 8) and move the logic of converting it into the overloads?

src/java/htsjdk/samtools/filter/AbstractJavascriptFilter.java
```diff
@@ -0,0 +1,173 @@
...
+        final ScriptEngineManager manager = new ScriptEngineManager();
+        /* get javascript engine */
+        final ScriptEngine engine = manager.getEngineByName("js");
+        if (engine == null) {
+            throw new RuntimeException("The embedded 'javascript' engine is not available in java. "
```
lbergelson: Could you add a new exception type `RuntimeScriptException` and use that in all the places you're throwing runtime exception so that someone downstream can catch if it they want?

lbergelson:
@lindenb I made a few comments and then realized that I'd put them on the wrong commit, so I deleted those and rewrote them in the right place.  Sorry for the long delay in review.  This looks great now, thanks for working with us!  Just two minor comments from me now and then I think it's good to merge :)

src/java/htsjdk/samtools/filter/AbstractJavascriptFilter.java
lindenb: this is the private constructor, it's called by the two constructors (String / File) .

I kept both constructors (File and String) this way because the latest JS engine compiler (nashorn in jdk8) can use the java.io.File information to resolve a relative path ( https://wiki.openjdk.java.net/display/Nashorn/Nashorn+extensions )

> Current executing script file, line, directory of the script can be accessed by these global (read-only) properties. This is useful in debug/tracing scripts. __DIR__ can be used like Unix 'dirname' command. Script can load script-file relative path resources using __DIR__.

If we use a java.io.Reader we will lose some information. 

What do you think ?

lbergelson: I appreciate that it's the private constructor, that was definitely the right thing to do.  Even so though, it would be more comfortable if it wasn't accepting all Object.  

It's likely I'm missing something, but it looks like they're both just being converted to a `FileReader` which is the thing that gets passed to the engine.  Does nashorn have special handling for file readers instead of `Readers`?  If it does you should definitely keep passing in whatever gives it the most information.  

Could the constructor simply take a `File`?  It seems like the String case is just being used as a file path anyway.

lindenb: Opsss, my current design is wrong, my original idea was to call one of the two 'compile' function in https://docs.oracle.com/javase/7/docs/api/javax/script/Compilable.html 

```
compile(Reader script)
compile(String script)
```

I'll fix this in my next commit.

> Could the constructor simply take a File? It seems like the String case is just being used as a file path anyway. 

no , a String is an inline script. Very cool for specifying short scripts on the command line.

lbergelson: Ahah.  That makes sense if you want to be able to pass in either a reader or a short script as a string.  I was confused since it wasn't actually doing that.

lindenb:
@lbergelson back to you

* I've added a java.io.Reader constructor, but I've kept the others (String , File).
* The FIle based constructor throws a java.io.IOException
* I've created a RuntimeScriptException.java, thrown in the AbstractJSFilter constructor.

lbergelson:
@lindenb Why have the file based one throw IOException instead of wrapping those in RuntimeIOException?  It's not a problem, I'm just curious why you changed it.

lindenb:
@lbergelson because the constructor with  java.io.File calls

```java
this(new java.io.FileReader(scriptFile),header);
```

wrapping the Ioexception raised by FileReader  into a RuntimeIOException would require a static function isn't it ?

lbergelson:
Ah, right.  Good point.  Java constructor chaining is so limited.

lbergelson:
@lindenb Could you rebase and squash this?  Ready to merge I think.

lindenb:
@lbergelson rebased. Thanks.

PR #386, opened 2015-11-15
A new comparator that produces an ordering of SAMRecords that effecti…

tfenne:
…vely yields SO=unsorted and GO=query using a sort on a hash of queryname.

### DIFF
--- a/src/java/htsjdk/samtools/SAMFileWriterImpl.java
+++ b/src/java/htsjdk/samtools/SAMFileWriterImpl.java
@@ -184,9 +184,6 @@ public abstract class SAMFileWriterImpl implements SAMFileWriter
             alignment.setHeader(header); // re-establish the record header and attempt to resolve reference index values
         }
         if (sortOrder.equals(SAMFileHeader.SortOrder.unsorted)) {
-            if (!header.getGroupOrder().equals(SAMFileHeader.GroupOrder.none)) {
-                throw new UnsupportedOperationException("GroupOrder " + header.getGroupOrder() + " is not supported");
-            }
             writeAlignment(alignment);
         } else if (presorted) {
             assertPresorted(alignment);
--- a/src/java/htsjdk/samtools/SAMRecordQueryHashComparator.java
+++ b/src/java/htsjdk/samtools/SAMRecordQueryHashComparator.java
@@ -1,0 +1,68 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015 Tim Fennell
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package htsjdk.samtools;
+
+import htsjdk.samtools.util.Murmur3;
+
+/**
+ * SAMRecord comparator that provides an ordering based on a hash of the queryname. Has
+ * the useful property that reads with the same name will be grouped together, but that
+ * reads appear in an otherwise random order.  Useful for when the read names in a BAM
+ * are correlated to something else (e.g. position, read group), making a straight
+ * queryname sort undesirable.
+ *
+ * @author Tim Fennell
+ */
+public class SAMRecordQueryHashComparator extends SAMRecordQueryNameComparator {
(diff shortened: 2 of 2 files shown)


tfenne:
@yfarjoun, @nh13 : small PR here, could you take a look please and let me know what you think?  My goal is to provide a way to sort SAMRecords that will give every read for a template together, but otherwise in random order.  Motivated by the fact that the SRA has done some horrible things to 1KG BAMs (read name and mapping position seem highly correlated in at least one example I looked at).

I was tempted to add this as a new SortOrder, so I could leverage it anywhere SAMRecord sorting is provided in Picard, but felt that wasn't ideal since I really want the `@HD` line to correctly report `SO=unsorted GO=query`.

Thoughts on whether this is the right way to do this?  If so, I'll add some basic tests and rebase.

src/java/htsjdk/samtools/SAMRecordQueryHashComparator.java
```diff
@@ -0,0 +1,68 @@
...
+ * are correlated to something else (e.g. position, read group), making a straight
+ * queryname sort undesirable.
+ *
+ * @author Tim Fennell
+ */
+public class SAMRecordQueryHashComparator extends SAMRecordQueryNameComparator {
+    private final Murmur3 hasher = new Murmur3(42);
+
+    /**
+     * Compares two records based on an integer hash of their read name's. If the hash
```
yfarjoun: Their read name's what? ;-)

yfarjoun:
I think it's fine to do in this way, though wonder if an actual sort order would be good. Also, I don't understand why travis is failing...

nh13:
I would err on the side of having a new sort order so, like you said, we could leverage this in `SortSam` and all the `SortingCollection`s.  While we support sort orders found in the spec, does that mean we should only support those in the spec?  I don't think so.

src/java/htsjdk/samtools/SAMFileWriterImpl.java
```diff
@@ -171,9 +171,6 @@ private SAMRecordComparator makeComparator() {
     public void addAlignment(final SAMRecord alignment)
     {
         if (sortOrder.equals(SAMFileHeader.SortOrder.unsorted)) {
-            if (!header.getGroupOrder().equals(SAMFileHeader.GroupOrder.none)) {
-                throw new UnsupportedOperationException("GroupOrder " + header.getGroupOrder() + " is not supported");
-            }
```
nh13: I suppose I would want to understand why this check was here in the first place, given no where else do we consider the group order, and this only is an issue when writing.  I suppose it is because we would pass the alignment to the `alignmentSorter` if we request a sort order (`SO`) but do nothing in the case of group order (`GO`).  In your case, you are adding support `SO=unsorted` and `GO=query`, so shouldn't the alignment be passed to an alignmentSorter based on your new comparator if `SO=unsorted` and `GO=query`?

yfarjoun:
The only problem with defining a new sort order is that it is a very
particular order (in the sense that it is a particular hash with a
particular seed) and that seems like a strange thing to define in the
sam-spec.

nh13:
Well fortunately it is a very well defined ordering, and conforms to the spec (`SO=unsorted GO=query`), so could be our 'reference implementation' of this ordering.

src/java/htsjdk/samtools/SAMFileWriterImpl.java
tfenne: I have no idea why that check was there in the first place, and was rather surprised to find my code (a separate program in Picard) failing when I tried to create a SAMFileWriter with `SO=unsorted` and `GO=query`!  I went and read the spec and found absolutely no support for this assertion, and in fact found the opposite:

    Grouping of alignments, indicating that similar alignment records are grouped together but the
file is not necessarily sorted overall

I think the reality is that `GO` is very rarely used in reality, and has never really been used in HTSJDK or Picard, and that this check was implemented many years ago based on an incorrect understanding.

tfenne:
I don't really like defining a new sort order because 1) I think if someone else did that then HTSJDK validation of a SAM file would fail, and soI don't think we should do it either and 2) the ordering provided by the new comparator is correctly expressed using existing values (`SO=unsorted GO=query`).

The problem I have is that I think the "right" solution is a lot of work.  The current situation is complicated by:

1. The factory methods for SAMFileWriter use the sort order in the provided `SAMFileHeader` to define the ordering of records
1. `SortOrder` has a method to construct a Comparator 
1. The string name of `SortOrder` is placed into the header in SAM/BAM files, and is supposed to be limited to the set of values contained within the spec

Ideally I think things should work as follows:

* HTSJDK defines a new enum `Ordering` (insert better name here) with the following properties:
    * A name (can match SO from spec, but is not limited to that)
    * The `SortOrder` that the ordering produces
    * The `GroupOrder` that the ordering produces
    * A Comparator that defines the Ordering
* SAMFileWriterFactory and friends are modified to take an Ordering instead of using SortOrder from the header
* Command line programs are modified to take an Ordering instead of a SortOrder

The last two could be made backwards compatibly by continuing to allow use of SortOrder and mapping each of the allowed SortOrders to a specific Ordering.

But that seems like a whole lot of work, for not a lot of gain, and a lot of API and CLP churn for end users.  I'm not eager to take on that work personally.

yfarjoun:
but couldn't GO=none be the default (as per the spec) which would allow all
the current behavior to remain functional?

tfenne:
@yfarjoun I'm not sure I understand what you're proposing.

jmarshall:
See also what §2 (Recommended Practices) of the SAM spec has to say about **GO**, and also the motivating conversation in samtools/hts-specs#15.

yfarjoun:
After re-reading I realized that I was suggesting the same as your last
sentence about backwards compatibility, which makes me wonder if there's
really going to be that much "churn", since the usage of a SortOrder
continue to be allowed (with the mapping to GO=none, and the standard
comparator for that SO), but maybe I'm missing something.

Regarding the usage of GO, I wonder if the fact that it is recommended that
both SO and GO shouldn't be specified together implies that at least one of
them isn't defined properly...

tfenne:
Thanks @jmarshall .  I read through the discussion, and now think I should have payed more attention at the time, since I don't really agree with the recommended best practice.  I can see that many/most of the combinations would be invalid (e.g. `SO=coordinate GO=query`), but I think that in some cases positively stating `SO=unsorted GO=query` or `SO=unknown GO=query` can add information and is not incorrect.

tfenne:
@yfarjoun @nh13 Any objections to merging this (after fixing the lone typo)?  I think having the comparator is better than not, and I don't want to wade into the whole SO/GO -> comparator mapping arena right now.

nh13:
@tfenne :+1: rebase and run the tests.

yfarjoun:
:+1 from me.

PR #390, opened 2015-11-19
container parser should return empty record list from EOF container

vadimzalunin:
ContainerParser throws NPE when reading EOF container. The fix is to return an empty list instead.


lbergelson:
@vadimzalunin Could you add a test case for this that passes but didn't pass before this change was made?

lbergelson:
@vadimzalunin Any update on this?

vadimzalunin:
@lbergelson please review: test added and rebased

lbergelson:
Thanks for the update!

PR #395, opened 2015-11-25
A little cleanup to readBean() in MetricsFile and adding the generic …

tfenne:
…type to avoid casting.


tfenne:
A tiny little change for the better.  @nh13 can I get a review and :+1: ?

nh13:
:+1:

PR #397, opened 2015-11-28
Added support for a less-exhaustive (and less time-consuming) validat…

ktibbett:
…ion of the index, which tests only the first and last references, rather than all of them.  (First step towards addressing https://github.com/broadinstitute/picard/issues/346)

### DIFF
--- a/src/java/htsjdk/samtools/BamIndexValidator.java
+++ b/src/java/htsjdk/samtools/BamIndexValidator.java
@@ -25,13 +25,19 @@ package htsjdk.samtools;
 
 import htsjdk.samtools.util.CloseableIterator;
 
+import java.util.Arrays;
+import java.util.List;
+
 /**
- * One crisp, informative sentence or noun phrase that explains
- * the concept modeled by the class.
- * <p/>
+ * Class to validate (at two different levels of thoroughness) the index for a BAM file.
+ *
  * This class is [<em>not</em>] thread safe [because it is immutable].
  */
 public class BamIndexValidator {
+
+    public enum IndexValidationStringency {
+        EXHAUSTIVE, LESS_EXHAUSTIVE, NONE
+    }
 
     public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exception {
         // look at all chunk offsets in a linear index to make sure they are valid
@@ -40,37 +46,37 @@ public class BamIndexValidator {
 
             // content is from an existing bai file
             final CachingBAMFileIndex existingIndex = (CachingBAMFileIndex) reader.indexing().getBrowseableIndex(); // new CachingBAMFileIndex(inputBai, null);
-            final int n_ref = existingIndex.getNumberOfReferences();
+            final int numRefs = existingIndex.getNumberOfReferences();
 
             int chunkCount = 0;
             int indexCount = 0;
-            for (int i = 0; i < n_ref; i++) {
+            for (int i = 0; i < numRefs; i++) {
                 final BAMIndexContent content = existingIndex.getQueryResults(i);
                 for (final Chunk c : content.getAllChunks()) {
                     final CloseableIterator<SAMRecord> iter = ((SamReader.PrimitiveSamReaderToSamReaderAdapter) reader).iterator(new BAMFileSpan(c));
                     chunkCount++;
-                    BAMRecord b = null;
+                    SAMRecord sam = null;
                     try {
-                        b = (BAMRecord) iter.next();
+                        sam = iter.next();
                         iter.close();
                     } catch (final Exception e) {
-                        throw new SAMException("Exception in BamIndexValidator. Last good record " + b + " in chunk " + c + " chunkCount=" + chunkCount, e);
+                        throw new SAMException("Exception in BamIndexValidator. Last good record " + sam + " in chunk " + c + " chunkCount=" + chunkCount, e);
                     }
                 }
                 // also seek to every position in the linear index
                 // final BAMRecordCodec bamRecordCodec = new BAMRecordCodec(reader.getFileHeader());
                 // bamRecordCodec.setInputStream(reader.getInputStream());
 
-                LinearIndex linearIndex = content.getLinearIndex();
-                for (long l : linearIndex.getIndexEntries()) {
+                final LinearIndex linearIndex = content.getLinearIndex();
+                for (final long l : linearIndex.getIndexEntries()) {
                     try {
                         if (l != 0) {
                             final CloseableIterator<SAMRecord> iter = ((SamReader.PrimitiveSamReaderToSamReaderAdapter) reader).iterator(new BAMFileSpan(new Chunk(l, l + 1)));
-                            BAMRecord b = (BAMRecord) iter.next();   // read the first record identified by the linear index
+                            final SAMRecord sam = iter.next();   // read the first record identified by the linear index
                             indexCount++;
                             iter.close();
                         }
-                    } catch (Exception e) {
+                    } catch (final Exception e) {
                         throw new SAMException("Exception in BamIndexValidator. Linear index access failure " + l + " indexCount=" + indexCount, e);
                     }
 
@@ -84,4 +90,62 @@ public class BamIndexValidator {
         return 0;
     }
 
+    /**
+     * A less time-consuming index validation that only looks at the first and last references in the index
+     * and the first and last chunks in each of those
+     *
+     * @param reader
+     * @return # of chunks examined, or 0 if there is no browseable index for the reader
+     */
+    public static int lessExhaustivelyTestIndex(final SamReader reader) {
+        // look at all chunk offsets in a linear index to make sure they are valid
+        if (reader.indexing().hasBrowseableIndex()) {
+
+            // content is from an existing bai file
+            final CachingBAMFileIndex existingIndex = (CachingBAMFileIndex) reader.indexing().getBrowseableIndex();
+            final int numRefs = existingIndex.getNumberOfReferences();
+
+            int chunkCount = 0;
+            int indexCount = 0;
+            for (int i = 0; i < numRefs; i++) {
+
+                final BAMIndexContent content = existingIndex.getQueryResults(i);
+
+                final List<Chunk> chunks = content.getAllChunks();
+                final int numChunks = chunks.size();
+                // We are looking only at the first and last chunks
+                for (final int chunkNo : Arrays.asList(0, numChunks - 1)) {
+                    chunkCount++;
+
+                    final Chunk c = chunks.get(chunkNo);
+                    final CloseableIterator<SAMRecord> iter = ((SamReader.PrimitiveSamReaderToSamReaderAdapter) reader).iterator(new BAMFileSpan(c));
+                    try {
+                        final SAMRecord sam = iter.next();
+                        iter.close();
+                    } catch (final Exception e) {
+                        throw new SAMException("Exception querying chunk " + chunkNo + " from reference index " + i, e);
+                    }
+                }
+
+                // also seek to first and last position in the linear index
+                final long linearIndexEntries[] = content.getLinearIndex().getIndexEntries();
+                for (final int binNo : Arrays.asList(0, linearIndexEntries.length - 1)) {
+                    indexCount++;
+                    final long l = linearIndexEntries[binNo];
+                    try {
+                        if (l != 0) {
+                            final CloseableIterator<SAMRecord> iter = ((SamReader.PrimitiveSamReaderToSamReaderAdapter) reader).iterator(new BAMFileSpan(new Chunk(l, l + 1)));
+                            final SAMRecord sam = iter.next();   // read the first record identified by the linear index
+                            iter.close();
+                        }
+                    } catch (final Exception e) {
+                        throw new SAMException("Exception in BamIndexValidator. Linear index access failure " + l + " indexCount=" + indexCount, e);
+                    }
+                }
+            }
+            return chunkCount;
+        }
+        // else it's not a bam file with a browseable index
+        return 0;
+    }
 }
--- a/src/java/htsjdk/samtools/SamFileValidator.java
+++ b/src/java/htsjdk/samtools/SamFileValidator.java
@@ -25,6 +25,7 @@
 package htsjdk.samtools;
 
 import htsjdk.samtools.SAMValidationError.Type;
+import htsjdk.samtools.BamIndexValidator.IndexValidationStringency;
 import htsjdk.samtools.metrics.MetricBase;
 import htsjdk.samtools.metrics.MetricsFile;
 import htsjdk.samtools.reference.ReferenceSequence;
@@ -90,7 +91,7 @@ public class SamFileValidator {
     private Set<Type> errorsToIgnore = EnumSet.noneOf(Type.class);
     private boolean ignoreWarnings = false;
     private boolean bisulfiteSequenced = false;
-    private boolean validateIndex = false;
+    private IndexValidationStringency indexValidationStringency = IndexValidationStringency.NONE;
     private boolean sequenceDictionaryEmptyAndNoWarningEmitted = false;
     private final int maxTempFiles;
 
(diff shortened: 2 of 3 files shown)


src/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -84,4 +84,72 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
         return 0;
     }
 
+    /**
+     * A less time-consuming index validation that only looks at the first and last references in the index
+     *
+     * @param reader
+     * @return # of chunks examined, or 9 if there is no browseable index for the reader
```
eitanbanks: 9 or 0?  :)

src/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -84,4 +84,72 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
...
+
+
+            int chunkCount = 0;
+            int indexCount = 0;
+            for (int i = 0; i < n_ref; i++) {
+
+                // Whether we are validating this reference
+                final boolean validateThisReference = (i == 0 || i == (n_ref-1));
+
+                final BAMIndexContent content = existingIndex.getQueryResults(i);
```
eitanbanks: Why do we need this call if it's not the first or last reference contig?

src/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -84,4 +84,72 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
...
+                   chunkCount += content.getAllChunks().size();
+                }
+
+                // also seek to every position in the linear index
+                final LinearIndex linearIndex = content.getLinearIndex();
+                if (validateThisReference) {
+                    for (final long l : linearIndex.getIndexEntries()) {
+                        try {
+                            if (l != 0) {
+                                final CloseableIterator<SAMRecord> iter = ((SamReader.PrimitiveSamReaderToSamReaderAdapter) reader).iterator(new BAMFileSpan(new Chunk(l, l + 1)));
```
eitanbanks: What happens if all of the reads are on the same (first) contig?  Won't this be really slow?

tfenne: I think that in theory this method will get slower as the index fills up - i.e. as each base on the (first) contig goes from zero coverage to coverage >= 1, as more index bins will be populated.  Beyond that it shouldn't change significantly as you go from 1x on up, as this is querying and reading a single SAMRecord per index bin, and the number of bins is fixed by the size of the contig.

eitanbanks: @tfenne @ktibbett Since the first contig is generally the biggest (esp. in Kathleen's use case), doesn't this still mean that this method won't be super fast?  Human chr1 is 10% of the genome, so at best we're reducing the runtime by 1 order of magnitude - but previous runtimes on whole genomes were so bad that I don't think that's good enough.

src/java/htsjdk/samtools/BamIndexValidator.java
ktibbett: Needed in order to have the chunkCount incremented properly (else branch), as it's reported if an exception is thrown.  @tfenne @eitanbanks , is having this chunk count really useful?  Can we skip it in the light-validation?

src/java/htsjdk/samtools/BamIndexValidator.java
eitanbanks: @tfenne do you have any insight into why we need the linear index check here?  Can't we just remove this entire block of code?  At the very least, just check the first record for the contig and that's it?

eitanbanks: Okay, sounds good.  I'm fine with the first and last bin for each contig in both the rtree and linear indices.  That'll be fast.
So @ktibbett you'll need to change the code a bit.

ktibbett:
@tfenne @eitanbanks , revised per your suggestions -- thanks!  Would you take another look?

src/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -84,4 +85,70 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
...
+    public static int lessExhaustivelyTestIndex(final SamReader reader) {
+        // look at all chunk offsets in a linear index to make sure they are valid
+
+        if (reader.indexing().hasBrowseableIndex()) {
+
+            // content is from an existing bai file
+            final CachingBAMFileIndex existingIndex = (CachingBAMFileIndex) reader.indexing().getBrowseableIndex();
+            final int n_ref = existingIndex.getNumberOfReferences();
+
+
```
eitanbanks: I feel like Nils now, but there's an extra newline here.  And on lines 126 and 145.  :(

src/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -84,4 +85,70 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
...
+                final BAMIndexContent content = existingIndex.getQueryResults(i);
+
+                final List<Chunk> chunks = content.getAllChunks();
+                for (int j = 0; j < chunks.size(); j++) {
+
+                    chunkCount++;
+
+                    //validate the first and last chunk in each contig
+                    if (j == 0 || j==(chunks.size()-1)) {
```
eitanbanks: My understanding is that there could in theory be 1 chunk for every position in the contig.  For large contigs that means many millions of calls to chunks.size() here and on line 111, whereas that could be avoided by making it a final variable above.

src/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -84,4 +85,70 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
...
+                        final Chunk c = chunks.get(j);
+                        final CloseableIterator<SAMRecord> iter = ((SamReader.PrimitiveSamReaderToSamReaderAdapter) reader).iterator(new BAMFileSpan(c));
+                        BAMRecord b = null;
```
eitanbanks: We can put this in the try block.  It's only out here because it's used in the catch block, but its usage there doesn't make any sense (we're not keeping track of the last good record; b is always null).

eitanbanks:
Minor comments.  @tfenne wanna take a quick look too?

ktibbett:
@tfenne any further comments on this?

src/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -24,6 +24,7 @@
 package htsjdk.samtools;
 
 import htsjdk.samtools.util.CloseableIterator;
+import java.util.List;
 
 /**
  * One crisp, informative sentence or noun phrase that explains
```
tfenne: Would you mind fixing the javadoc while you're in here please?

src/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -84,4 +85,70 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
...
+     * @return # of chunks examined, or 0 if there is no browseable index for the reader
+     */
+    public static int lessExhaustivelyTestIndex(final SamReader reader) {
+        // look at all chunk offsets in a linear index to make sure they are valid
+
+        if (reader.indexing().hasBrowseableIndex()) {
+
+            // content is from an existing bai file
+            final CachingBAMFileIndex existingIndex = (CachingBAMFileIndex) reader.indexing().getBrowseableIndex();
+            final int n_ref = existingIndex.getNumberOfReferences();
```
tfenne: `n_ref`?  Shouldn't that be `nRef` or `numRefs`?

src/java/htsjdk/samtools/BamIndexValidator.java
tfenne: Does this even need to iterate any more?  I think you could either:
1. Check that `chunks.size()` > 0, and then just grab `chunks.get(0)` and `chunks.get(chunks.size()-1)` and test those?
2. Modify your for loop to be `for (int j = 0; j < chunks.size(); j+=chunks.size()-1)` or `for (int j : new int[]{0, chunks.size()-1})`

src/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -84,4 +85,70 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
...
+
+                    chunkCount++;
+
+                    //validate the first and last chunk in each contig
+                    if (j == 0 || j==(chunks.size()-1)) {
+                        final Chunk c = chunks.get(j);
+                        final CloseableIterator<SAMRecord> iter = ((SamReader.PrimitiveSamReaderToSamReaderAdapter) reader).iterator(new BAMFileSpan(c));
+                        BAMRecord b = null;
+                        try {
+                            b = (BAMRecord) iter.next();
```
tfenne: I there any reason to downcast this to `BAMRecord`?  I'm not sure how the CRAM indexing works, but assuming CRAM has a browseable index, this would presumably blow up on a CRAM file.

src/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -84,4 +85,70 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
...
+                            iter.close();
+                        } catch (final Exception e) {
+                            throw new SAMException("Exception in BamIndexValidator. Last good record " + b + " in chunk " + c + " chunkCount=" + chunkCount, e);
```
tfenne: As Eric points out above, the message here should probably change, to something like:
`"Exception querying chunk " + j + " from reference index " +i`

src/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -84,4 +85,70 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
...
+                        }
+
+                    }
+                }
+
+                // also seek to first and last position in the linear index
+                final long linearIndexEntries[] = content.getLinearIndex().getIndexEntries();
+                for (int k = 0; k < linearIndexEntries.length; k++) {
```
tfenne: Same comment about not really needing to loop here to check just the first and last entries.

src/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -84,4 +85,70 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
...
+                    indexCount++;
+                    if (k == 0 || k == (linearIndexEntries.length-1)) {
+                        final long l = linearIndexEntries[k];
+                        try {
+                            if (l != 0) {
+                                final CloseableIterator<SAMRecord> iter = ((SamReader.PrimitiveSamReaderToSamReaderAdapter) reader).iterator(new BAMFileSpan(new Chunk(l, l + 1)));
+                                final BAMRecord b = (BAMRecord) iter.next();   // read the first record identified by the linear index
```
tfenne: And again about the unnecessary cast.

src/java/htsjdk/samtools/SamFileValidator.java
```diff
@@ -90,7 +90,7 @@
     private Set<Type> errorsToIgnore = EnumSet.noneOf(Type.class);
     private boolean ignoreWarnings = false;
     private boolean bisulfiteSequenced = false;
-    private boolean validateIndex = false;
+    private ValidationStringency indexValidationStringency = ValidationStringency.SILENT;
```
tfenne: Two things:
1. I don't love the use of `ValidationStringency` here for controlling how much index validation is done, since the values tend to imply how errors are _reported_, not how much validation is applied.  I won't object to leaving it this way, but it might be better to either have a boolean that controls whether validation is exhaustive or less exhaustive
2. The default should not be SILENT, since this changes default behaviour very significantly from exhaustive validation to no validation.  I'd be ok with changing the default to less exhaustive, but not omitting validation entirely.

tfenne:
Back to you @ktibbett, sorry for the delay in looking at this one.

src/java/htsjdk/samtools/SamFileValidator.java
ktibbett: I'm going to change the use of ValidationStringency here, but please note that changing it to "SILENT" was just implementing the "validateIndex = false;" in the line above.  I think the default behavior would have been the same.

ktibbett: There's a setValidateIndex method that turns it on.

ktibbett:
@tfenne @eitanbanks : updated in response to code review comments.  Care to take another look?

src/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -84,4 +88,64 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
...
+    /**
+     * A less time-consuming index validation that only looks at the first and last references in the index
+     * and the first and last chunks in each of those
+     *
+     * @param reader
+     * @return # of chunks examined, or 0 if there is no browseable index for the reader
+     */
+    public static int lessExhaustivelyTestIndex(final SamReader reader) {
+        // look at all chunk offsets in a linear index to make sure they are valid
+
```
eitanbanks: Brace yourself, I'm about to do my best impression of Nils:
extra newline here

src/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -84,4 +88,64 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
...
+                            iter.close();
+                        }
+                    } catch (final Exception e) {
+                        throw new SAMException("Exception in BamIndexValidator. Linear index access failure " + l + " indexCount=" + indexCount, e);
+                    }
+                }
+            }
+            return chunkCount;
+        }
+        // else  not a bam file with a browseable index
```
eitanbanks: and an extra space here

src/tests/java/htsjdk/samtools/ValidateSamFileTest.java
```diff
@@ -411,7 +411,7 @@ public void testIndexFileValidation() throws Exception {
         final File outFile = File.createTempFile("validation", ".txt");
         outFile.deleteOnExit();
         final PrintWriter out = new PrintWriter(outFile);
-        new SamFileValidator(out, 8000).setValidateIndex(true).validateSamFileSummary(samReader, reference);
+        new SamFileValidator(out, 8000).setIndexValidationStringency(BamIndexValidator.IndexValidationStringency.EXHAUSTIVE).validateSamFileSummary(samReader, reference);
         final LineNumberReader reader = new LineNumberReader(new FileReader(outFile));
```
eitanbanks: Could you please also add a test here (or even add it to this existing test) to test the lessExhaustivelyTestIndex() method?

eitanbanks:
Looks good to me other than a few minor quibbles.

src/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -84,4 +88,64 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
...
+            int chunkCount = 0;
+            int indexCount = 0;
+            for (int i = 0; i < numRefs; i++) {
+
+                final BAMIndexContent content = existingIndex.getQueryResults(i);
+
+                final List<Chunk> chunks = content.getAllChunks();
+                final int numChunks = chunks.size();
+                // We are looking only at the first and last chunks
+                for (int j = 0; j < numChunks; j+=(numChunks-1)) {
```
tfenne: Up to you, but I think this would be clearer as:
`for (final chunkNo : Arrays.asList(0, numChunks-1)) {`

src/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -84,4 +88,64 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
...
+                        final SAMRecord sam = iter.next();
+                        iter.close();
+                    } catch (final Exception e) {
+                        throw new SAMException("Exception querying chunk " + j + " from reference index " + i, e);
+                    }
+                }
+
+                // also seek to first and last position in the linear index
+                final long linearIndexEntries[] = content.getLinearIndex().getIndexEntries();
+                for (int k = 0; k < linearIndexEntries.length; k+=(linearIndexEntries.length-1)) {
```
tfenne: Same suggestion here to substitute:
`for (final int binNo : Arrays.asList(0, linearIndexEntries.length - 1))`

tfenne:
Minor comments from me which you should feel free to either ignore, or incorporate then merge.

ktibbett:
Thanks guys.  Changes made.

PR #398, opened 2015-11-30
function to parse an interval in a SAMSequenceDictionary + tests

lindenb:
Added a small function to parse an interval + test in  SAMSequenceDictionary .

```java
 public Interval parseInterval(final String region) 
```

The function checks that the contig exist in the dict, and clip the positions if pos<1 or pos>contig.length()

Various syntaxes are supported:

     *     CHROM -> CHROM:1-seq.length()
     *     CHROM:POS ->  CHROM:POS-POS
     *     CHROM:POS+NUM -> CHROM:{POS-NUM}-{POS+NUM}
     *     CHROM:BEGIN- ->  CHROM:BEGIN-seq.length()
     *     CHROM:BEGIN-END ->  CHROM:BEGIN-END


src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -215,9 +217,128 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
-
+    
+    /** 
+     * called by `parseInterval`. Parse a coordinate string, remove the comma look at the suffix (bp,kb,mp,gb) to scale,
+     * verify that the result is a positive integer
+     * @author Pierre Lindenbaum 
+     * @param numStr the coordinate
+     * @return the coordinate as integer
+     */
+    private static int parseCoordinate(final String numStr)
```
yfarjoun: is SamDictionary the right place for this? isn't this a way to define an Interval? seems like Interval is a more natural place for it. What do you think?

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -215,9 +217,128 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
+        {
+        long factor = 1L;
+        String num = numStr.replace(",","").trim().toLowerCase();
+        if(num.endsWith("bp"))
+            {
+            factor = 1L;
+            num = num.substring(0,num.length()-2).trim();
```
yfarjoun: refactor out this line please, and fix the alignment:

```code
if␣(num.endsWith("bp"))␣{
            factor = 1L;
            num = num.substring(0,␣num.length()␣-␣2).trim();
} else { 
``` 

etc.

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -215,9 +217,128 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
+            }
+        else if(num.endsWith("mb"))
+            {
+            factor = 1000000L;//1E6
+            num = num.substring(0,num.length()-2).trim();
+            }
+        else if(num.endsWith("gb"))
+            {
+            factor = 1000000000L;//1E9
```
yfarjoun: use underscores for easy reading:
```code
factor = 1_000_000_000L; / 1E9
```

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -215,9 +217,128 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
+            num = num.substring(0,num.length()-2).trim();
+            }
+
```
yfarjoun: should you throw an exception if there substring doesn't match any of the things you attempted?

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -215,9 +217,128 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
+        else if(num.endsWith("gb"))
+            {
+            factor = 1000000000L;//1E9
+            num = num.substring(0,num.length()-2).trim();
+            }
+        
+        try {
+            long pos = Long.parseLong(num);
+            pos *= factor;
+            if(pos < 1L || pos > Integer.MAX_VALUE)  throw new IllegalArgumentException("Cannot convert "+numStr+" to a positive integer");
```
yfarjoun: double space and missing spaces around `+` signs

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -215,9 +217,128 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
+            return (int)pos;
+        } catch (NumberFormatException e) {
+            throw new IllegalArgumentException("Cannot convert "+numStr+" to a coordinate");
```
yfarjoun: spaces

src/java/htsjdk/samtools/SAMSequenceDictionary.java
yfarjoun: Nevermind. You are using the dictionary itself for the contig order. Ignore.

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -215,9 +217,128 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
+            }
+        }
+    
+    /** 
+     * parses the interval String and returns a htsjdk.samtools.util.Interval
+     * The contig/chromosome must exist in the dictionary. out-of-bound index <code>!(1<= index <= sequence.length())</code> will be clipped
```
yfarjoun: above zero throws an exception, doesn't clip. please correct comment.

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -215,9 +217,128 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
+            return (int)pos;
+        } catch (NumberFormatException e) {
+            throw new IllegalArgumentException("Cannot convert "+numStr+" to a coordinate");
+            }
+        }
+    
+    /** 
+     * parses the interval String and returns a htsjdk.samtools.util.Interval
+     * The contig/chromosome must exist in the dictionary. out-of-bound index <code>!(1<= index <= sequence.length())</code> will be clipped 
+     * coordinate suffixes like 'bp,Mb,Gb' will be interpreted.
```
yfarjoun: ...as 1,1000,and 1000000 bases

src/tests/java/htsjdk/samtools/SAMSequenceDictionaryTest.java
```diff
@@ -39,6 +41,55 @@
...
 
 public class SAMSequenceDictionaryTest {
+    
+    @Test void testIntervalParser() {
+        final int contigLength = 100;
+        final SAMSequenceRecord ssr1 = new SAMSequenceRecord("1", contigLength);
+        final SAMSequenceDictionary dict = new SAMSequenceDictionary(
+                Arrays.asList(ssr1));
+        Assert.assertEquals(dict.size(), 1);
+        Assert.assertEquals(
```
yfarjoun: please add test for "1:0"

yfarjoun: please use a @DataProvider

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -215,9 +217,128 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
+     * Comma in the coordinate will be removed.
+     * left/right white spaces will be trimmed.
+     * Position=0 will be set to 1.
+     * A Position larger than the contig size will be set to the contig size
+     * 
+     * Valid syntax -> convertion :
```
yfarjoun: conversion

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -215,9 +217,128 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
+            final SAMSequenceRecord record = getSequence(region);
+            if(record == null) throw new IllegalArgumentException("Unknown chromosome \""+region+"\".");
+            // return whole chromosome
+            return new Interval(
+                    record.getSequenceName(),
+                    1,
+                    record.getSequenceLength()
+                    );
+        }
+        if(colon == 0)  throw new IllegalArgumentException("Empty chromosome in "+region);
```
yfarjoun: lots of spaces missing from here and down. esp. around `+` signs.

yfarjoun: and after commas

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -215,9 +217,128 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
+            }
+        }
+    
+    /** 
+     * parses the interval String and returns a htsjdk.samtools.util.Interval
+     * The contig/chromosome must exist in the dictionary. out-of-bound index <code>!(1<= index <= sequence.length())</code> will be clipped 
+     * coordinate suffixes like 'bp,Mb,Gb' will be interpreted.
+     * Comma in the coordinate will be removed.
+     * left/right white spaces will be trimmed.
+     * Position=0 will be set to 1.
```
yfarjoun: comment disagrees with code.

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -215,9 +217,128 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
+            final int pos = parseCoordinate(region.substring(colon+1));
+            if( pos < 1 || pos > record.getSequenceLength()) throw new IllegalArgumentException("In "+region +", position="+pos+" out of bounds 1<=pos<="+record.getSequenceLength());
+            return new Interval( record.getSequenceName(),pos,pos);
+        }
+        //CHROM:POS+NUM
+        if(hyphen == -1 && plus > colon) {
+            final int pos = parseCoordinate(region.substring(colon+1,plus));
+            final int extend = parseCoordinate(region.substring(plus+1));
+            if( extend < 0)  throw new IllegalArgumentException("Extend cannot be negative in "+region);
+            if( pos - extend  > record.getSequenceLength() || pos+extend  <1 ) throw new IllegalArgumentException("In "+region +", position="+pos+"+/-"+extend+" out of bounds 1<=pos<="+record.getSequenceLength());
```
yfarjoun: code disagrees with java doc. are you clipping or throwing an exception?

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -215,9 +217,128 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
+        if(hyphen == -1 && plus == -1) {
+            final int pos = parseCoordinate(region.substring(colon+1));
+            if( pos < 1 || pos > record.getSequenceLength()) throw new IllegalArgumentException("In "+region +", position="+pos+" out of bounds 1<=pos<="+record.getSequenceLength());
+            return new Interval( record.getSequenceName(),pos,pos);
+        }
+        //CHROM:POS+NUM
+        if(hyphen == -1 && plus > colon) {
+            final int pos = parseCoordinate(region.substring(colon+1,plus));
+            final int extend = parseCoordinate(region.substring(plus+1));
+            if( extend < 0)  throw new IllegalArgumentException("Extend cannot be negative in "+region);
```
yfarjoun: parseCoordinate would never return a negative number...

src/tests/java/htsjdk/samtools/SAMSequenceDictionaryTest.java
```diff
@@ -39,6 +41,55 @@
...
+                dict.parseInterval("1:10-20"),
+                new Interval("1", 10, 20)
+               );
+        Assert.assertEquals(
+                dict.parseInterval("1:10-"+(contigLength+100)),
+                new Interval("1", 10, contigLength)
+               );
+    }
+
```
yfarjoun: could you please add some test that expectedly throw exceptions due to formatting of input string?

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -215,9 +217,128 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
+       //CHROM:START-
+       if( region.substring(hyphen+1).trim().isEmpty()) {
+           return new Interval( record.getSequenceName(),
+                   Math.max(1,chromStart),
+                   record.getSequenceLength()
+                   );
+           }
+       //CHROM:START-END
+       final int chromEnd = parseCoordinate(region.substring(hyphen+1));
+       if(chromEnd < chromStart)  throw new IllegalArgumentException("In "+region +", start>end");
```
yfarjoun: Interval actually supports the end being 1 less than the start to indicate an empty interval between the two bases:

1:10-9 can point to the place of an insertion between the 9th and 10th bases. Would you like to allow for that in the parsing?

src/tests/java/htsjdk/samtools/SAMSequenceDictionaryTest.java
yfarjoun: please add tests that have spaces in them (not only in the begining and end)

src/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -215,9 +217,128 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
+     * @param region a 1-based coordinate region.  
+     * @return the parsed htsjdk.samtools.util.Interval
+     */
+    public Interval parseInterval(final String region) {
+        if( region == null) throw new IllegalArgumentException("parseInterval: region is null");
+        if(region.isEmpty())  throw new IllegalArgumentException("parseInterval: region is empty");
+        final int colon = region.indexOf(':');
+        // no colon, chromosome only
+        if(colon == -1) {
+            final SAMSequenceRecord record = getSequence(region);
```
yfarjoun: There's a lot of parsing going on here. Would it be possible to use a regex to capture the parsing? 
something like this:
```code
^(.*?)((:)(.*?)(([-+])?(.*?))?)?$
```

Then the first group would be the CHROM
if there is a second group, the fourth would be POS1 the sixth (if there) would be the + or -  and the seventh (if there) would the POS2, then you can separate the parsing from the logic of constructing the interval, which currently is combined in a complicated way.

I'm not saying it has to be my regex...that is just an example...but I think it would be good to separate the parsing of the string and the logic of creating the Interval.

src/tests/java/htsjdk/samtools/SAMSequenceDictionaryTest.java
```diff
@@ -39,6 +41,55 @@
...
+                dict.parseInterval("1:10"),
+                new Interval("1", 10, 10)
+               );
+        Assert.assertEquals(
+                dict.parseInterval("1:10bp"),
+                new Interval("1", 10, 10)
+               );
+        Assert.assertEquals(
+                dict.parseInterval("1:10-10kb"),
+                new Interval("1", 10, contigLength)
```
yfarjoun: mb is not tested 
kb isn't tested to return the correct interval (only clipped)

yfarjoun:
review done.

src/java/htsjdk/samtools/SAMSequenceDictionary.java
lbergelson: I'd definitely vote against having a regex like that.  Separating the parsing from the interpretation is a good idea though.

lindenb:
@yfarjoun  One month later I'm back to this pull request.

   * As you suggested, I've moved everything to InrervalUtil (rather than Interval or SamSequenceDictionary )
   * I've added positive and negative tests, using DataProvider 
   * updated the javadoc, messages, etc...
  
thanks.

droazen:
@yfarjoun Is this one ok to merge? Has @lindenb has addressed your original comments to your satisfaction?

Issue #394, opened 2015-11-25
CRAMFileReader.query does not propagate validationStringency to created iterators


akiezun:
`CRAMFileReader.query` creates a new `CRAMIterator` but does not propagate validationStringency so it's always STRICT and blows up even if the CRAMFileReader's stringency is SILENT

akiezun:
@vadimzalunin can you have a look? I'm not a gatekeeper so i won't assign it to you but please grab it if is it a bug

droazen:
Assigning to @vadimzalunin -- please look at this when you have a chance, as it's a blocker for some downstream development.

cmnbroad:
Closed via https://github.com/samtools/htsjdk/pull/399.

Issue #378, opened 2015-11-09
process for evolving api


lbergelson:
I think we should formalize a process for evolving the api and removing deprecated functions.  We have some things that have been deprecated for years and never removed.  

I'd recommend something like "functions that are deprecated are subject to removal after n months."  Where N is something like 3-6.  We'd commit to documenting what was removed and how to replace it if relevant in those releases.

yfarjoun:
I agree with the sprit, but I think that 3-6 months is too fast. We already
have users (eg igv) who inform us of the trouble they go through when they
rev htsjdk, removing deprecated functions after 3 months would make it much
harder for users like that to use the library. I think that something on
the order of 1 year, assuming there was a discussion of the (specific)
deprecation as an issue, and an agreement among the users would be more
appropriate...but that's just my 2ct.

Y

tfenne:
I tend to agree with @yfarjoun - deprecated things should have a fairly long twilight period before being yanked.  While high volume users of the library and likely to update frequently, I suspect there is a fairly long tail of users who don't update all that often and would get bitten by that kind of change.  I would think something in the 9-12 month range would be more realistic.

lbergelson:
I think it would be a good idea to pair a deprecation policy with a 'experimental' or 'beta' tag policy, so that new functionality wouldn't be locked in immediately.  

Basically: new functionality that isn't quite solid gets tagged with `@beta` or `@experimental`.

We reserve the rights to change these methods at will.  That way we can experiment with new ideas without being immediately committed to supporting them forever.  At some point, experimental methods lose their experimental status.  Then we've committed to not changing them until they've been deprecated.

Three months is probably fast.  A year seems slow to me.  Whatever the number is isn't that important, it's just important that we choose a number and can then act to remove things at some point.

vadimzalunin:
I would advise to separate API and implementation even further. APIs are generally much more stable and that's ok. At the moment it looks like we tend to classify all public classes/methods as API but they are not API imo. Ideally an API would have a version, which is different from htsjdk version.

cmnbroad:
[...]

PR #399, opened 2015-12-01
proppagate ValidationStringency from CRAMFileReader to CRAMIterator

vadimzalunin:
https://github.com/samtools/htsjdk/issues/394

### DIFF
--- a/src/java/htsjdk/samtools/CRAMFileReader.java
+++ b/src/java/htsjdk/samtools/CRAMFileReader.java
@@ -127,8 +127,7 @@ public class CRAMFileReader extends SamReader.ReaderImplementation implements Sa
         this.referenceSource = referenceSource;
         this.validationStringency = validationStringency;
 
-        iterator = new CRAMIterator(inputStream, referenceSource);
-        iterator.setValidationStringency(validationStringency);
+        iterator = new CRAMIterator(inputStream, referenceSource, validationStringency);
         if (indexInputStream != null) {
             try {
                 mIndex = new CachingBAMFileIndex(indexInputStream, iterator.getSAMFileHeader().getSequenceDictionary());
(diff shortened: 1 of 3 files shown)


vadimzalunin:
@droazen @akiezun please review

lbergelson:
@vadimzalunin Looks good.  Could you add a test that proves that it's fixed for the case that is described in #394?  I.e. an example query call that fails with strict validation but passes with silent validation.

cmnbroad:
We already have a setter/getter for validation stringency on the iterator, so I'm wondering if we can't just use those to fix this bug. It looks like this change breaks backward compatibility for CRAMIterators - I'm just not sure how those kinds of changes are handled with htsjdk. Can we make changes like that to public APIs ?

lbergelson:
@cmnbroad It's a good point.  My thinking was the following:

1. The coming release is going to break backwards compatibility in a number of small ways already.
2.  The cram support is still very new and I was thinking it's essentially still and alpha interface. 
3.  Respecting validation stringency is pretty important, and it's easy to forget to set it if it's not done in the constructor.  

If people think it's dangerous to change these constructors I'm fine with that too though.

lbergelson:
We need to define a policy for experimental API's so that we can feel free to revise things while they're still under heavy development.  I started a thread about it here, but we didn't reach any conclusion...  #378

cmnbroad:
Yeah we definitely need something like that. It sounds like we should move ahead with this for now though.

vadimzalunin:
@cmnbroad  I was thinking about using the setter but then realized that all calls to CRAMIterator constructors are followed by setValidation method. 
As for changes to public API, I agree that CRAM should be allowed to change for now.

lbergelson:
we merged this in so that it's fixed for the release we're doing now.  We should still add that test.

tomwhite:
I agree with @cmnbroad's point about not breaking backward compatibility. This has caused a problem already, see https://github.com/broadinstitute/gatk/issues/1247.

Issue #362, opened 2015-10-30
Require all CRAM files to have a reference


cmnbroad:
Currently its possible to create a CRAMFileReader that doesn't have an associated reference, which can result in several downstream problems (NPE, failure to locate contigs, etc). I have some changes to lock this down tighter but its unclear how strict we should be.

@vadimzalunin - is there *any* valid use case/operations for reading CRAM files without the accompanying reference, or are there cases where we need to allow this ?

vadimzalunin:
I think this has been fixed in https://github.com/samtools/htsjdk/pull/339 but perhaps needs testing.

CRAM supports unmapped readds as well as embedded reference, and in both cases ref file is not required. 
Also there is a feature of downloading a sequence on demand from ENA ref service.

cmnbroad:
@vadimzalunin - we still have several cases where we have bad failure modes if a reference isn't provided (including NPEs). Do you have any suggestions for how we can validate the referenceSource at creation time (ie., in CRAMFileReader, Writer and Iterator) that isn't overly constraining? Verifying that the referenceSource is backed by a file is too stringent (and breaks some tests that use fake or in memory sources), but the common case of using the SAMFileWriter and Reader factories on CRAM files can lead to problems if the reference isn't provided or is bad.

Currently, CRAMFileReader and Writer do no checking, and CRAMIterator checks for a null referenceSource, but thats ineffective since the default referenceSource is almost always used, but it points to the null DEFAULT_FASTA. So you get NPEs further on.

Some examples and stacks:

Create a CRAMFIleWriter with no reference:
			at htsjdk.samtools.cram.build.Sam2CramRecordFactory.addSubstitutionsAndMaskedBases(Sam2CramRecordFactory.java:295)
			at htsjdk.samtools.cram.build.Sam2CramRecordFactory.createVariations(Sam2CramRecordFactory.java:239)
			at htsjdk.samtools.cram.build.Sam2CramRecordFactory.checkedCreateVariations(Sam2CramRecordFactory.java:184)
			at htsjdk.samtools.cram.build.Sam2CramRecordFactory.createCramRecord(Sam2CramRecordFactory.java:139)
			at htsjdk.samtools.CRAMFileWriter.flushContainer(CRAMFileWriter.java:201)
			at htsjdk.samtools.CRAMFileWriter.writeAlignment(CRAMFileWriter.java:353)


Create a CRAMFIleWriter with a non-null but  not valid reference:

[...]

PR #400, opened 2015-12-02
Require a reference for CRAM files, add javadoc and some tests.

cmnbroad:
Per the discussion in https://github.com/samtools/htsjdk/issues/362, require a valid reference source for CRAMReader, CRAMWriter, CRAMIterator and CRAMNormalizer. The SAMReader factories now use a method to create a default reference source (which throws if the reference_fasta property is not set and the download services isn't enabled) if one is not supplied directly, and the CRAM* classes throw if passed a null reference source.

### DIFF
--- a/src/java/htsjdk/samtools/CRAMFileReader.java
+++ b/src/java/htsjdk/samtools/CRAMFileReader.java
@@ -58,32 +58,40 @@ public class CRAMFileReader extends SamReader.ReaderImplementation implements Sa
     private ValidationStringency validationStringency;
 
     /**
-     * Open CRAM data for reading using either the file or the input stream
-     * supplied in the arguments. The
-     * {@link htsjdk.samtools.Defaults#REFERENCE_FASTA default} reference fasta
-     * file will be used.
+     * Create a CRAMFileReader from either a file or input stream using the reference source returned by
+     * {@link ReferenceSource#getDefaultCRAMReferenceSource() getDefaultCRAMReferenceSource}.
+     *
      *
      * @param cramFile CRAM file to open
-     * @param inputStream   CRAM stream to read
+     * @param inputStream CRAM stream to read
+     *
+     * @throws IllegalArgumentException if the {@code cramFile} and the {@code inputStream} are both null
+     * @throws IllegalStateException if a {@link ReferenceSource#getDefaultCRAMReferenceSource() default}
+     * reference source cannot be acquired
      */
     public CRAMFileReader(final File cramFile, final InputStream inputStream) {
-        this(cramFile, inputStream, new ReferenceSource(Defaults.REFERENCE_FASTA));
+        this(cramFile, inputStream, ReferenceSource.getDefaultCRAMReferenceSource());
     }
 
     /**
-     * Open CRAM data for reading using either the file or the input stream
-     * supplied in the arguments.
+     * Create a CRAMFileReader from either a file or input stream using the supplied reference source.
      *
-     * @param cramFile            CRAM file to read
-     * @param inputStream              index file to be used for random access
+     * @param cramFile        CRAM file to read
+     * @param inputStream     CRAM stream to read
      * @param referenceSource a {@link htsjdk.samtools.cram.ref.ReferenceSource source} of
-     *                        reference sequences
+     *                        reference sequences. May not be null.
+     *
+     * @throws IllegalArgumentException if the {@code cramFile} and the {@code inputStream} are both null
+     * or if the {@code ReferenceSource} is null
      */
     public CRAMFileReader(final File cramFile, final InputStream inputStream,
                           final ReferenceSource referenceSource) {
-        if (cramFile == null && inputStream == null)
-            throw new IllegalArgumentException(
-                    "Either file or input stream is required.");
+        if (cramFile == null && inputStream == null) {
+            throw new IllegalArgumentException("Either file or input stream is required.");
+        }
+        if (referenceSource == null) {
+            throw new IllegalArgumentException("A reference is required for CRAM readers");
+        }
 
         this.cramFile = cramFile;
         this.inputStream = inputStream;
@@ -92,18 +100,22 @@ public class CRAMFileReader extends SamReader.ReaderImplementation implements Sa
     }
 
     /**
-     * Open CRAM file for reading. If index file is supplied than random access
-     * will be available.
+     * Create a CRAMFileReader from a file and optional index file using the supplied reference source. If index file
+     * is supplied then random access will be available.
      *
-     * @param cramFile        CRAM file to read
-     * @param indexFile       index file to be used for random access
+     * @param cramFile        CRAM file to read. May not be null.
+     * @param indexFile       index file to be used for random access. May be null.
      * @param referenceSource a {@link htsjdk.samtools.cram.ref.ReferenceSource source} of
-     *                        reference sequences
+     *                        reference sequences. May not be null.
+     * @throws IllegalArgumentException if the {@code cramFile} or the {@code ReferenceSource} is null
      */
     public CRAMFileReader(final File cramFile, final File indexFile,
                           final ReferenceSource referenceSource) {
         if (cramFile == null)
             throw new IllegalArgumentException("File is required.");
+        if (referenceSource == null) {
+            throw new IllegalArgumentException("A reference is required for CRAM readers");
+        }
 
         this.cramFile = cramFile;
         this.mIndexFile = indexFile;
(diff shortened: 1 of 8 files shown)


src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -56,32 +56,39 @@
...
 
     /**
-     * Open CRAM data for reading using either the file or the input stream
-     * supplied in the arguments.
+     * Create a CRAMFileReader from either a file or input stream using the supplied reference source.
      *
-     * @param cramFile            CRAM file to read
-     * @param inputStream              index file to be used for random access
+     * @param cramFile        CRAM file to read
+     * @param inputStream     index file to be used for random access
```
droazen: `inputStream` is for the index file? The docs for the two-arg constructor `CRAMFileReader(final File cramFile, final InputStream inputStream)` that calls into this three-arg constructor imply that `inputStream` is connected to the cram file itself, not the index. Something is fishy here...

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -56,32 +56,39 @@
...
-     * @param inputStream   CRAM stream to read
+     * @param inputStream CRAM stream to read
+     *
+     * @throws IllegalArgumentException if the {@code cramFile} and the {@code inputStream} are both null
+     * @throws IllegalStateException if a {@link ReferenceSource#getDefaultCRAMReferenceSource() default}
+     * reference source cannot be acquired
      */
     public CRAMFileReader(final File cramFile, final InputStream inputStream) {
-        this(cramFile, inputStream, new ReferenceSource(Defaults.REFERENCE_FASTA));
+        this(cramFile, inputStream, ReferenceSource.getDefaultCRAMReferenceSource());
```
droazen: The docs for the constructor you're calling here suggest that `inputStream` is assumed to be connected to the index, not the cram itself. Is this actually the case, or do the docs just need to be updated?

droazen: It's also silly and confusing to have a constructor that takes both a `File` and `InputStream` for the same file -- seems like it should be split into two 1-arg constructors (unless `inputStream` is really meant to be for the index).

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -90,18 +97,22 @@ public CRAMFileReader(final File cramFile, final InputStream inputStream,
...
-     * Open CRAM file for reading. If index file is supplied than random access
-     * will be available.
+     * Create a CRAMFileReader from a file and and optional index file using the supplied reference source. If index file
+     * is supplied than random access will be available.
      *
-     * @param cramFile        CRAM file to read
+     * @param cramFile        CRAM file to read. May not be null.
      * @param indexFile       index file to be used for random access
      * @param referenceSource a {@link htsjdk.samtools.cram.ref.ReferenceSource source} of
      *                        reference sequences
```
droazen: "May not be null"

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -90,18 +97,22 @@ public CRAMFileReader(final File cramFile, final InputStream inputStream,
...
 
     /**
-     * Open CRAM file for reading. If index file is supplied than random access
-     * will be available.
+     * Create a CRAMFileReader from a file and and optional index file using the supplied reference source. If index file
+     * is supplied than random access will be available.
      *
-     * @param cramFile        CRAM file to read
+     * @param cramFile        CRAM file to read. May not be null.
      * @param indexFile       index file to be used for random access
```
droazen: "May be null"

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -121,8 +142,26 @@ public CRAMFileReader(final File cramFile, final ReferenceSource referenceSource
         getIterator();
     }
 
+    /**
+     * Create a CRAMFileReader from an input stream and optional index stream using the supplied reference source.
+     *
+     * @param inputStream      index file to be used for random access. Can not be null.
```
droazen: This can't be right -- `inputStream` here is for the cram file itself, not the index, right?

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -121,8 +142,26 @@ public CRAMFileReader(final File cramFile, final ReferenceSource referenceSource
...
+     * @param indexInputStream index file to be used for random access
```
droazen: "May be null"

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -141,12 +180,34 @@ public CRAMFileReader(final InputStream inputStream, final SeekableStream indexI
         }
     }
 
+    /**
+     * Create a CRAMFileReader from an input stream and optional index file using the supplied reference source.
+     *
+     * @param stream            index file to be used for random access. Can not be null.
```
droazen: Here too -- `stream` is for the cram itself, not the index, right?

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -141,12 +180,34 @@ public CRAMFileReader(final InputStream inputStream, final SeekableStream indexI
...
+     * @param indexFile         index file to be used for random access
```
droazen: "May be null"

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -141,12 +180,34 @@ public CRAMFileReader(final InputStream inputStream, final SeekableStream indexI
...
     public CRAMFileReader(final InputStream stream,
                           final File indexFile, final ReferenceSource referenceSource,
                           final ValidationStringency validationStringency) throws IOException {
         this(stream, indexFile == null ? null: new SeekableFileStream(indexFile), referenceSource, validationStringency);
     }
 
+    /**
+     * Create a CRAMFileReader from a CRAM file and optional index file using the supplied reference source.
+     *
+     * @param cramFile        index file to be used for random access. Can not be null.
```
droazen: And here -- `cramFile` is not for the index.

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -141,12 +180,34 @@ public CRAMFileReader(final InputStream inputStream, final SeekableStream indexI
...
+     * @param indexFile       index file to be used for random access
```
droazen: "May be null"

src/java/htsjdk/samtools/CRAMFileWriter.java
```diff
@@ -134,10 +149,6 @@ private void initCRAMWriter(final OutputStream indexOS, final ReferenceSource so
         setSortOrder(samFileHeader.getSortOrder(), preSorted);
         setHeader(samFileHeader);
 
-        if (this.source == null) {
-            this.source = new ReferenceSource(Defaults.REFERENCE_FASTA);
-        }
```
droazen: If `Defaults.REFERENCE_FASTA` is set, is there any remaining mechanism for propagating it to `CRAMFileWriter`?

src/java/htsjdk/samtools/cram/ref/ReferenceSource.java
```diff
@@ -63,6 +63,40 @@ public ReferenceSource(final ReferenceSequenceFile rsFile) {
         this.rsFile = rsFile;
     }
 
+    /**
+     * Construct a default ReferenceSource.
+     *
+     * @return ReferenceSource if one can be acquired
```
droazen: Mention whether this is guaranteed to be non-null

src/java/htsjdk/samtools/cram/ref/ReferenceSource.java
```diff
@@ -63,6 +63,40 @@ public ReferenceSource(final ReferenceSequenceFile rsFile) {
...
+                throw new IllegalArgumentException(
+                        "The file specified by the reference property does not exist: " + Defaults.REFERENCE_FASTA.getName());
+            }
+        }
+        else if (Defaults.USE_CRAM_REF_DOWNLOAD) {
+            return new ReferenceSource();
+        }
+        else {
+            throw new IllegalStateException(
+                    "A valid CRAM reference was not supplied and one cannot be acquired via property settings");
```
droazen: Mention what these property settings are in the error message.

src/java/htsjdk/samtools/cram/ref/ReferenceSource.java
```diff
@@ -63,6 +63,40 @@ public ReferenceSource(final ReferenceSequenceFile rsFile) {
         this.rsFile = rsFile;
     }
 
+    /**
+     * Construct a default ReferenceSource.
```
droazen: Mention that this is specifically for CRAM, and should only be called when an explicitly-supplied `ReferenceSource` is not available.

droazen:
Review complete -- back to @cmnbroad. Major confusion in the `CRAMFileReader ` docs (and possibly the constructor code itself) as to which streams are connected to the CRAM itself and which are connected to the index. Also some question as to whether you should be calling `ReferenceSource.getDefaultCRAMReferenceSource()` in `CRAMFileWriter`

src/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: The docs for the called constructor (and several of the other inputStreams below) are wrong. I'll fix them as part of this PR.

And I totally agree this constructor is silly and confusing, not only because of the stream/file thing, but also because its the only one of the (7!) CRAMFileReader constructors that doesn't have a reference argument, so its only useful if you're using a default reference source. But its already public. Its only use in htsjdk is from SamFileReader, which is deprecated - should we deprecate it ?

src/java/htsjdk/samtools/CRAMFileWriter.java
cmnbroad: If I understand the question, yes. The consumer should create a reference source by whatever mechanism it wants, perhaps by calling getDefaultReferenceSource, and pass it in to one of these constructors. The whole idea of this PR was to have code at this level (CRAMFileReader and CRAMFileWriter) always require that a reference be provided, and give the higher levels (i.e. SAMFileWriterFactory) a unified way to create a default to provide to the lower levels via getDefaultReferenceSource. Not sure if that answers the question.

Issue #316, opened 2015-08-31
Usage of SAMRecord.validateCigar


dariober:
Hello,
I'm surely missing something trivial here... I can't make `validateCigar` to actually capture errors in a `SAMRecord`. In the example below I create a `SAMRecord` with nonsensical cigar. Still, `validateCigar` returns null instead of a list of errors. This is on htsjdk-1.130 and Java 1.6.

Thanks!
Dario

```
package validateSAMRecord;

import htsjdk.samtools.SAMRecord;
import htsjdk.samtools.SAMValidationError;
import htsjdk.samtools.SamReader;
import htsjdk.samtools.SamReaderFactory;
import htsjdk.samtools.ValidationStringency;

import java.io.File;
import java.io.IOException;
import java.util.ArrayList;

public class ValidateSAMRecord {

	public static void main(String[] args) throws IOException {
		
		// Read dummy header
		SamReaderFactory sf = SamReaderFactory.
				makeDefault().
				validationStringency(ValidationStringency.STRICT);
		SamReader sam= sf.open(new File("/tmp/header.sam"));
		
		// Create nonsense record
		SAMRecord rec= new SAMRecord(sam.getFileHeader());
		rec.setCigarString("nonsense");

		// Validate record
		ArrayList<SAMValidationError> err= 
				(ArrayList<SAMValidationError>) rec.validateCigar(-1);
		
		System.out.println(err); // Returns null!!
		
		sam.close();
	}
}
```

yfarjoun:
@dariober, sorry it took so long to get to your issue. I added a test in PR #405. As you can see there, the problem with your code is that the default validationStringency for a SamRecord is SILENT. So, before you validate you have to setValidationStringency to STRICT and then the validation will indeed fail.

PR #405, opened 2015-12-06
investigate issue 316 in htsjdk. 

yfarjoun:
Added a test to explore issue #316

### DIFF
--- a/src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
+++ b/src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
@@ -31,6 +31,7 @@ import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
 import java.io.*;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
@@ -456,7 +457,7 @@ public class SAMRecordUnitTest {
             Assert.fail("Unexpected exception", e);
         }
 
-        record.setAttribute(tag, (long)Integer.MIN_VALUE-1L);
+        record.setAttribute(tag, (long) Integer.MIN_VALUE - 1L);
     }
 
     @Test(expectedExceptions = SAMException.class)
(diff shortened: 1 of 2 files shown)


yfarjoun:
@droazen care to look at this tiny PR. it's just another test, no new code.

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -777,6 +778,23 @@ public void testNullHeaderSerialization(final File inputFile) throws Exception {
...
+    @Test(expectedExceptions = IllegalArgumentException.class)
+    public void testValidateNonsenseCigar(){
+        // Create nonsense record
+        SAMRecord rec = createTestRecordHelper();
+        rec.setCigarString("nonsense");
+
+        //The default validationStringency of a sam record is SILENT.
+        rec.setValidationStringency(ValidationStringency.STRICT);
+        // Validate record
+        List<SAMValidationError> err = rec.validateCigar(-1);
```
lbergelson: This test is kind of deceptive.  The test is throwing on this line in `rec.validateCigar(-1)`  So it never actually reaches those asserts below it.  

It looks like this test is proving that `validateCigar()` has a bug.  The contract says that it will `@return List of errors, or null if no errors.`, but instead of doing so it throws an `IllegalArgumentException` if there is a completely invalid cigar.

lbergelson:
@yfarjoun I think you've exposed a bug in `validateCigar()`.  Your thoughts?

yfarjoun:
right! thanks for pointing this out. I've fixed the code. care to look again?

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
yfarjoun: point taken. expected exception removed.

lbergelson:
:+1: Looks good to me.  Test driven development indeed!   Rebase and merge when tests pass.

yfarjoun:
waiting for tests. will merge when passes.

yfarjoun:
thanks.

PR #406, opened 2015-12-08
SAM/BAM writers should always call setHeader on alignment records.

cmnbroad:
Includes some DRY-refactoring to support cleaning up mate/reference index state management and javadoc.

### DIFF
--- a/src/java/htsjdk/samtools/SAMRecord.java
+++ b/src/java/htsjdk/samtools/SAMRecord.java
@@ -100,9 +100,10 @@ import java.util.Set;
  * A record with null a header may be validated by the isValid method, but the reference and mate reference indices,
  * read group, sequence dictionary, and alignment start will not be fully validated unless a header is present.
  * <p>
- * Also, SAMTextWriter, BAMFileWriter, and CRAMFileWriter all require records to have a valid header in order to be
- * written. Any record that does not have a header at the time it is added to the writer will be updated to use the
- * header associated with the writer.
+ * Also, SAMTextWriter, BAMFileWriter, and CRAMFileWriter all require the reference and mate reference names to be valid
+ * in order to be written. At the time a record is added to a writer it will be updated to use the header associated
+ * with the writer and the reference and mate reference names must be valid for that header. If the names cannot be
+ * resolved using the writer's header, an exception will be thrown.
  * <p>
  * @author alecw@broadinstitute.org
  * @author mishali.naik@intel.com
@@ -343,26 +344,24 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
      * the reference index to NO_ALIGNMENT_REFERENCE_INDEX.
      *
      * @param referenceName - must not be null
+     * @throws IllegalArgumentException if {@code referenceName} is null
      */
     public void setReferenceName(final String referenceName) {
         if (null == referenceName) {
             throw new IllegalArgumentException(
                     "Reference name must not be null. Use SAMRecord.NO_ALIGNMENT_REFERENCE_NAME to reset the reference name.");
         }
+        if (null != mHeader) {
+            mReferenceIndex = resolveIndexFromName(referenceName, mHeader, false);
+            // String.intern() is surprisingly expensive, so avoid it by calling resolveNameFromIndex
+            // and using the interned value in the sequence dictionary if possible
+            mReferenceName = null == mReferenceIndex ?
+                                    referenceName.intern() :
+                                    resolveNameFromIndex(mReferenceIndex, mHeader);
+        }
         else if (NO_ALIGNMENT_REFERENCE_NAME.equals(referenceName)) {
             mReferenceName = NO_ALIGNMENT_REFERENCE_NAME;
             mReferenceIndex = NO_ALIGNMENT_REFERENCE_INDEX;
-        }
-        else if (null != mHeader) {
-            // String.intern() is surprisingly expensive, so avoid it by looking up in sequence dictionary if possible
-            final int referenceIndex = mHeader.getSequenceIndex(referenceName);
-            if (-1 != referenceIndex) {
-                setReferenceIndex(referenceIndex); // sets reference name and index
-            }
-            else {
-                mReferenceName = referenceName.intern();
-                mReferenceIndex = null;
-            }
         }
         else {
             mReferenceName = referenceName.intern();
@@ -383,20 +382,14 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
      * @return Index in the sequence dictionary of the reference sequence. If the read has no reference sequence, or if
      * the reference name is not found in the sequence index, NO_ALIGNMENT_REFERENCE_INDEX (-1) is returned.
      *
-     * @throws IllegalStateException if the reference index cannot be resolved because the SAMFileHeader for the
-     * record is null.
+     * @throws IllegalStateException if the reference index must be resolved but cannot be because the SAMFileHeader
+     * for the record is null.
      */
     public Integer getReferenceIndex() {
-        if (null == mReferenceIndex) {
-            // try to resolve the reference index
-            if (NO_ALIGNMENT_REFERENCE_NAME.equals(mReferenceName)) {
+        if (null == mReferenceIndex) { // try to resolve the reference index
+            mReferenceIndex = resolveIndexFromName(mReferenceName, mHeader, false);
+            if (null == mReferenceIndex) {
                 mReferenceIndex = NO_ALIGNMENT_REFERENCE_INDEX;
-            }
-            else if (null != mHeader) {
-                mReferenceIndex = mHeader.getSequenceIndex(mReferenceName);
-            }
-            else {
-                throw new IllegalStateException("A non-null SAMFileHeader is required to resolve the reference index");
             }
         }
         return mReferenceIndex;
@@ -411,27 +404,16 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
      * @param referenceIndex Must either equal NO_ALIGNMENT_REFERENCE_INDEX (-1) indicating no reference, or the
      *                       record must have a SAMFileHeader and the index must exist in the associated sequence
      *                       dictionary.
-     * @throws IllegalStateException if the SAMFileHeader is null for this record or the reference index is not
-     *         found in the sequence dictionary for this record.
+     * @throws IllegalStateException if {@code referenceIndex} is not equal to NO_ALIGNMENT_REFERENCE_INDEX and the
+     * SAMFileHeader is null for this record
+     * @throws IllegalArgumentException if {@code referenceIndex} is not found in the sequence dictionary in the header
+     * for this record.
      */
     public void setReferenceIndex(final int referenceIndex) {
-        if (referenceIndex == NO_ALIGNMENT_REFERENCE_INDEX) {
-            mReferenceIndex = NO_ALIGNMENT_REFERENCE_INDEX;
-            mReferenceName = NO_ALIGNMENT_REFERENCE_NAME;
-        }
-        else if (null == mHeader) {
-            throw new IllegalStateException("A non-null SAMFileHeader is required to resolve the reference index");
-        }
-        else {
-            SAMSequenceRecord samSequence = mHeader.getSequence(referenceIndex);
-            if (null != samSequence) {
-                mReferenceIndex = referenceIndex;
-                mReferenceName = samSequence.getSequenceName();
-            }
-            else {
-                throw new IllegalArgumentException("Reference index " + referenceIndex + " not found in sequence dictionary.");
-            }
-        }
+        // resolveNameFromIndex throws if the index can't be resolved
+        setReferenceName(resolveNameFromIndex(referenceIndex, mHeader));
+        // setReferenceName does this as a side effect, but set the value here to be explicit
+        mReferenceIndex = referenceIndex;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -514,27 +478,79 @@ else if (null != mHeader) {
...
+     *                {@code referenceName} is not NO_ALIGNMENT_REFERENCE_NAME.
+     * @param strict if true, throws if {@code referenceName}  does not appear in the header's sequence dictionary
+     * @returns the reference index corresponding to the {@code referenceName}, or null if strict is false and {@code referenceName}
+     * does not appear in the header's sequence dictionary.
+     * @throws IllegalStateException if {@code referenceName} is not equal to NO_ALIGNMENT_REFERENCE_NAME and the header is null
+     * @throws IllegalArgumentException if strict is true and the name does not appear in header's sequence dictionary.
+     *
+     * Does not mutate the SAMRecord.
+     */
+    protected static Integer resolveIndexFromName(final String referenceName, final SAMFileHeader header, final boolean strict) {
```
droazen: Please add direct unit tests for both `resolveIndexFromName()` and `resolveNameFromIndex()` that exercise all the possible cases.

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -514,27 +478,79 @@ else if (null != mHeader) {
...
+        String referenceName = NO_ALIGNMENT_REFERENCE_NAME;
+        if (NO_ALIGNMENT_REFERENCE_INDEX != referenceIndex) {
+            if (null == header) {
+                throw new IllegalStateException("A non-null SAMFileHeader is required to resolve the reference index or name");
             }
-            else {
-                throw new IllegalArgumentException("Reference index " + mateReferenceIndex + " not found in sequence dictionary.");
+            SAMSequenceRecord samSeq = header.getSequence(referenceIndex);
+            if (null == samSeq) {
+                throw new IllegalArgumentException("Reference name for '" + referenceIndex + "' not found in sequence dictionary.");
```
droazen: So there is no valid use case for a "strict" mode here as well to control whether this exception is thrown, correct?

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -342,26 +343,19 @@ private boolean hasMateReferenceName() {
...
-            if (-1 != referenceIndex) {
-                setReferenceIndex(referenceIndex); // sets reference name and index
-            }
-            else {
-                mReferenceName = referenceName.intern();
-                mReferenceIndex = null;
-            }
+            mReferenceName = null == mReferenceIndex ?
+                                    referenceName.intern() :
+                                    resolveNameFromIndex(mReferenceIndex, mHeader);
```
droazen: Why do we make this additional call to `resolveNameFromIndex`() here? Won't this always just give us back the `referenceName` we already had, provided the earlier call to `resolveIndexFromName()` returned non-null?

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -342,26 +343,19 @@ private boolean hasMateReferenceName() {
...
+     * @throws IllegalArgumentException if {@code referenceName} is null
      */
     public void setReferenceName(final String referenceName) {
         if (null == referenceName) {
             throw new IllegalArgumentException(
                     "Reference name must not be null. Use SAMRecord.NO_ALIGNMENT_REFERENCE_NAME to reset the reference name.");
         }
-        else if (NO_ALIGNMENT_REFERENCE_NAME.equals(referenceName)) {
-            mReferenceName = NO_ALIGNMENT_REFERENCE_NAME;
-            mReferenceIndex = NO_ALIGNMENT_REFERENCE_INDEX;
```
droazen: Previously if `NO_ALIGNMENT_REFERENCE_NAME` was passed in here, `mReferenceIndex` would get set to `NO_ALIGNMENT_REFERENCE_INDEX` even if the header was null. Now if the header is null, `mReferenceIndex` will get set to null in this case. Is there a reason for this behavioral change? Ideally we shouldn't change any existing behavior when refactoring to use the new `resolve*` methods.

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -382,20 +376,14 @@ else if (null != mHeader) {
...
-            }
-            else if (null != mHeader) {
-                mReferenceIndex = mHeader.getSequenceIndex(mReferenceName);
-            }
-            else {
-                throw new IllegalStateException("A non-null SAMFileHeader is required to resolve the reference index");
+        if (null == mReferenceIndex) { // try to resolve the reference index
+            mReferenceIndex = resolveIndexFromName(mReferenceName, mHeader, false);
+            if (null == mReferenceIndex) {
+                return NO_ALIGNMENT_REFERENCE_INDEX;
```
droazen: Previously this method returned null in this case -- why the change?

droazen: This new code leaves `mReferenceIndex` set to null in the case where the lookup failed, whereas the previous version would leave it set to `NO_ALIGNMENT_REFERENCE_INDEX`. Is this what we want?

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -410,27 +398,15 @@ else if (null != mHeader) {
...
-                mReferenceIndex = referenceIndex;
-                mReferenceName = samSequence.getSequenceName();
-            }
-            else {
-                throw new IllegalArgumentException("Reference index " + referenceIndex + " not found in sequence dictionary.");
-            }
-        }
+        // resolveNameFromIndex throws if the index can't be resolved
+        setReferenceName(resolveNameFromIndex(referenceIndex, mHeader));
+        mReferenceIndex = referenceIndex;
```
droazen: Won't `setReferenceName()` set the reference index for us (since we know the header lookup succeeded if the method call even happens)?

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -447,24 +423,18 @@ public String getMateReferenceName() {
...
+     * @throws IllegalArgumentException if {@code mateReferenceName} is null
      */
     public void setMateReferenceName(final String mateReferenceName) {
         if (null == mateReferenceName) {
-            throw new IllegalArgumentException("Mate reference name must not be null");
+            throw new IllegalArgumentException("Mate reference name must not be null. Use SAMRecord.NO_ALIGNMENT_REFERENCE_NAME to reset the mate reference name.");
         }
-        else if (NO_ALIGNMENT_REFERENCE_NAME.equals(mateReferenceName)) {
-            mMateReferenceName = NO_ALIGNMENT_REFERENCE_NAME;
-            mMateReferenceIndex = NO_ALIGNMENT_REFERENCE_INDEX;
```
droazen: Same comment as for `setReferenceName()` above -- the behavior has changed so that if the header is null and `NO_ALIGNMENT_REFERENCE_NAME` is passed in here, `mMateReferenceIndex` ends up as null rather than `NO_ALIGNMENT_REFERENCE_INDEX`. Is this desired?

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -447,24 +423,18 @@ public String getMateReferenceName() {
...
-            else {
-                mMateReferenceName = mateReferenceName.intern();
-                mMateReferenceIndex = null;
-            }
+        if (null != mHeader) {
+            mMateReferenceIndex = resolveIndexFromName(mateReferenceName, mHeader, false);
+            // String.intern() is surprisingly expensive, so try to avoid it by looking up in sequence dictionary if possible
+            mMateReferenceName = null == mMateReferenceIndex ?
+                                   mateReferenceName.intern() :
+                                   resolveNameFromIndex(mMateReferenceIndex, mHeader);
```
droazen: Same comment as for `setReferenceName()` -- why do we make this additional call to resolveNameFromIndex() here?

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -486,20 +456,14 @@ else if (null != mHeader) {
...
-            }
-            else if (null != mHeader) {
-                mMateReferenceIndex = mHeader.getSequenceIndex(mMateReferenceName);
-            }
-            else {
-                throw new IllegalStateException("A non-null SAMFileHeader is required to resolve the mate reference index");
+        if (null == mMateReferenceIndex) { // try to resolve the mate reference index
+            mMateReferenceIndex = resolveIndexFromName(mMateReferenceName, mHeader, false);
+            if (null == mMateReferenceIndex) {
+                return NO_ALIGNMENT_REFERENCE_INDEX;
```
droazen: Same comment as for `getReferenceIndex()` -- This new code leaves mMateReferenceIndex set to null in the case where the lookup failed, whereas the previous version would leave it set to `NO_ALIGNMENT_REFERENCE_INDEX`. Do we care?

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -514,27 +478,79 @@ else if (null != mHeader) {
...
     public void setMateReferenceIndex(final int mateReferenceIndex) {
-        if (mateReferenceIndex == NO_ALIGNMENT_REFERENCE_INDEX) {
-            mMateReferenceIndex = NO_ALIGNMENT_REFERENCE_INDEX;
-            mMateReferenceName = NO_ALIGNMENT_REFERENCE_NAME;
-        }
-        else if (null == mHeader) {
-            throw new IllegalStateException("A non-null SAMFileHeader is required to resolve the mate reference index");
+        // resolveNameFromIndex throws if the index can't be resolved
+        setMateReferenceName(resolveNameFromIndex(mateReferenceIndex, mHeader));
+        mMateReferenceIndex = mateReferenceIndex;
```
droazen: Same comment as for `setReferenceIndex()` -- Won't setMateReferenceName() set the reference index for us (since we know the header lookup succeeded if resolveNameFromIndex() did not throw)?

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -847,4 +847,51 @@ private void testNullHeadGetCigarBAM() {
 
         testNullHeaderCigar(bamRec);
     }
+
+    @Test
+    public void testSetHeaderStrictValid() {
+        SAMRecord sam = createTestRecordHelper();
+        final SAMFileHeader samHeader = sam.getHeader();
```
droazen: Add a second `testSetHeaderStrictValid` test case that uses `setHeaderStrict()` to set the header on a read to a header that is compatible, but different from the original header.

droazen: Also test setting a valid header on a headerless record.

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -847,4 +847,51 @@ private void testNullHeadGetCigarBAM() {
...
+        Assert.assertEquals(sam.getReferenceIndex(), originalRefIndex);
+    }
+
+    @Test(expectedExceptions=IllegalArgumentException.class)
+    public void testSetHeaderStrictInvalidReference() {
+        SAMRecord sam = createTestRecordHelper();
+        final SAMFileHeader samHeader = sam.getHeader();
+
+        sam.setReferenceName("unresolveable");
+        int originalRefIndex = sam.getReferenceIndex();
```
droazen: This is odd -- didn't `setReferenceName()` clear the reference index? (and so "originalRefIndex" is not actually the "original" reference index)

droazen:
First-pass review complete -- back to @cmnbroad

src/java/htsjdk/samtools/SAMRecord.java
cmnbroad: Its just to maintain the interning pattern that was already in place before, and retrieve the already interned object. Not sure how much its buying us. I must make that comment more explicit because even with it I had to think about it a bit to recall why I did that.

src/java/htsjdk/samtools/SAMRecord.java
cmnbroad: Yes, I think I was just trying to be explicit about it since its the main point of the call.

src/java/htsjdk/samtools/SAMRecord.java
cmnbroad: Good point.

src/java/htsjdk/samtools/SAMRecord.java
cmnbroad: Also a good point.

src/java/htsjdk/samtools/SAMRecord.java
cmnbroad: No. Needs fixing.

src/java/htsjdk/samtools/SAMRecord.java
cmnbroad: Interning.

src/java/htsjdk/samtools/SAMRecord.java
cmnbroad: Also needs fixing.

src/java/htsjdk/samtools/SAMRecord.java
cmnbroad: Just being explicit.

src/java/htsjdk/samtools/SAMRecord.java
cmnbroad: Done.

src/java/htsjdk/samtools/SAMRecord.java
cmnbroad: Right - I originally had a strict mode for this as well but the policy was and is that you can't set an unresolveable index on a record, only an unresolveable name.

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
cmnbroad: Both done.

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
cmnbroad: Right, done.

cmnbroad:
Back to @droazen.

Also, a while back we agreed that you should be able to use setHeaderStrict to set a null header and just mirror the behavior of setHeader for that case. So I made that change in this round, which affected one test.

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -847,4 +847,138 @@ private void testNullHeadGetCigarBAM() {
...
+    public void testResolveIndexResolveable() {
+        final SAMRecord sam = createTestRecordHelper();
+        final SAMFileHeader samHeader = sam.getHeader();
+        final String contigName = sam.getContig();
+        Assert.assertEquals(SAMRecord.resolveIndexFromName(contigName, samHeader, true), new Integer(samHeader.getSequenceIndex(contigName)));
+    }
+
+    @Test(expectedExceptions=IllegalStateException.class)
+    public void testResolveIndexUnresolveableStrict() {
+        Assert.assertEquals(SAMRecord.resolveIndexFromName("unresolveable", null, false), new Integer(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX));
```
droazen: Test case is mis-named (strict mode is off here). This is testing whether you throw when the header is null and strict mode is off. You also need a separate test with an unresolvable contig, a non-null header, and strict mode on that expects an exception.

droazen:
:+1: Branch looks good now. One final TODO in the tests, then please merge this yourself. Back to @cmnbroad

PR #407, opened 2015-12-11
Added checks that can very early report that path is not an SRA archive and not load SRA engine and its libraries

a-nikitiuk:
We implemented several checks to see if the path looks like an SRA archive and if not fail immediately and try other readers.

This PR should address issues for users working with non-SRA data sources and limited internet connections. (for instance, that one: https://github.com/broadinstitute/picard/issues/354)

The file recognition in `SamReaderFactory` will work as follows:

1. Checks if the file is a BAM file and then loads it;
2. Checks if the file is a CRAM file and then loads it;
3. Checks if the file is:
  * An existing file, then checks its header to match SRA signature. If matched - load SRA engine and pass the file to it
  * Non-existing file path, then checks if it matches to SRA accession pattern (`^[SED]RR[0-9]{6,9}$`). If matched - load SRA engine and make it check if that was a real SRA accession
4. Tries to read the file as a text SAM file

You can also pass `InputResource` of type `InputResource.Type.SRA_ACCESSION`
 to `SamReaderFactory`, then we will load SRA engine and try to read that `InputResource` as SRA (because you specified SRA type explicitly).

### DIFF
--- a/src/java/htsjdk/samtools/sra/SRAAccession.java
+++ b/src/java/htsjdk/samtools/sra/SRAAccession.java
@@ -29,11 +29,19 @@ package htsjdk.samtools.sra;
 import htsjdk.samtools.util.Log;
 import gov.nih.nlm.ncbi.ngs.NGS;
 
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
 import java.io.Serializable;
+import java.util.Arrays;
 
 /**
- * Describes a single SRA accession
+ * Describes a single SRA accession for SRA read collection
  * Also provides app string functionality and allows to check if working SRA is supported on the running platform
+ *
+ * Important: due to checks performed in SRAAccession.isValid(), we won't recognise any accessions other
+ * than ones that follow the pattern "^[SED]RR[0-9]{6,9}$", e.g. SRR000123
  */
 public class SRAAccession implements Serializable {
     private static final Log log = Log.getInstance(SRAAccession.class);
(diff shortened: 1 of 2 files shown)


src/java/htsjdk/samtools/sra/SRAAccession.java
```diff
@@ -75,11 +83,34 @@ public static boolean isSupported() {
...
-            return false;
+        boolean looksLikeSRA = false;
+        File f = new File(acc);
+        if (f.isFile()) {
+            try {
+                byte[] buffer = new byte[8];
+                byte[] signature1 = "NCBI.sra".getBytes();
+                byte[] signature2 = "NCBInenc".getBytes();
+
+                InputStream is = new FileInputStream(f);
```
lbergelson: We're java 8 now!  So this can be made into a try-with resources and the close can be done implicitly.

src/java/htsjdk/samtools/sra/SRAAccession.java
```diff
@@ -29,11 +29,19 @@
...
 import java.io.Serializable;
+import java.util.Arrays;
 
 /**
- * Describes a single SRA accession
+ * Describes a single SRA accession for SRA read collection
  * Also provides app string functionality and allows to check if working SRA is supported on the running platform
+ *
+ * Important: due to checks performed in SRAAccession.isValid(), we won't recognise any accessions other
+ * than ones that follow the pattern "^[SED]RR[0-9]{6,9}$"
```
lbergelson: Please add an example accession ID that matches this pattern to the comment.

src/java/htsjdk/samtools/sra/SRAAccession.java
```diff
@@ -29,11 +29,19 @@
...
  */
 public class SRAAccession implements Serializable {
```
lbergelson: Since this method now has somewhat non-trivial logic in it, could you add a unit test for it.

lbergelson:
@a-nikitiuk Thanks for this fix.  A have a few minor comments.

src/java/htsjdk/samtools/sra/SRAAccession.java
```diff
@@ -75,11 +83,34 @@ public static boolean isSupported() {
...
+            }
+        } else if (f.exists()) {
+            // anything else local other than a file is not an SRA archive
+            looksLikeSRA = false;
+        } else {
+            looksLikeSRA = acc.toUpperCase().matches ( "^[SED]RR[0-9]{6,9}$" );
         }
 
-        return NGS.isValid(acc);
+        if (!looksLikeSRA)
```
lbergelson: always `{}` even for 1 liners

tfenne: Unless on the same line ala `if (!looksLikeSRA) return false;`

src/java/htsjdk/samtools/sra/SRAAccession.java
a-nikitiuk: DONE

src/java/htsjdk/samtools/sra/SRAAccession.java
a-nikitiuk: DONE

src/java/htsjdk/samtools/sra/SRAAccession.java
a-nikitiuk: DONE: added tests

src/java/htsjdk/samtools/sra/SRAAccession.java
a-nikitiuk: Changed to `if (!looksLikeSRA) return false;`

a-nikitiuk:
@lbergelson I updated PR to have all your comments fixed. Please check

lbergelson:
Thank you. I will merge as soon as the tests run. Travis is having technical difficulties today it seems.

Issue #387, opened 2015-11-16
Request: SamLocusIterator including indel and pair information


magicDGS:
SamLocusIterator is a great class to provide "pileup" information, but it lacks insertion and deletions that could be useful for develop tools for indel calling at the same time as SNP calling. On the other hand, pair-end information in the iterator could be very useful to use linkage in the calling procedure.

If the problem for indel calling is solve somehow with a different class or the pair-end information could be retrieved in a different way, I apologize for opening the issue.

Thank you very much.

yfarjoun:
pair-end information can be found by processing the data with either
MergeBamAlignment or FixMateInformation (Picard tools). This will place the
mate CIGAR string in each read in a MC: tag. I think this could provide
much of the information you require.

Regarding indels I need to look further into the issue, perhaps someone
else can respond.

magicDGS:
I was thinking more on developing new tools (independent of Picard), but it could be a requirement to pre-process the BAM file to add the MC tag. Thank you very much for the suggestion.

tfenne:
@yfarjoun I think including, in LocusInfo, the sets of reads that have insertions and/or deletions at that position would be relatively straightforward.  I think it would involve:
1. Either modifying `LocusInfo` to have separate collections of records for 'covered with base', 'deleted in record', and 'insertion in record' _or_ modifying `RecordAndOffset` to have a discriminator type.
2. Deciding at which locus insertions should be presented (base before or base after)
3. Modifying `SamLocusIterator.accumulateSamRecord()` to also process insertions/deletions and not just alignment blocks.

magicDGS:
@tfenne That idea sounds great and it could help a lot. With respect to the second point, the base before is more reasonable for my point of view, as in the output from samtools mpileup. Thank you very much.

yfarjoun:
This is a great request, but I'm not sure at all who will get around to
actually doing it..is there a downstream tool that would greatly benefit
from having this option?

magicDGS:
@yfarjoun I'm developing now a tool that needs that information, so I could try to implement in the way suggested by @tfenne: adding to `LocusInfo` two lists (deletedInRecord and insertionInRecord) and methods to retrieve them and a modified accumulator (accumulateSamRecordWithIndels) for processing indels. Some minor changes will be include a boolean field for the indel option and setter/getter methods for it in a similar way as the filters or quality threshold.

[...]

PR #408, opened 2015-12-15
Added indel support to SamLocusIterator

magicDGS:
Added indel tracking to `SamLocusIterator` (as discussed in issue #387) and some tests. The default behaviour is still do not keep reads spanning indels.

### DIFF
--- a/src/main/java/htsjdk/samtools/util/SamLocusIterator.java
+++ b/src/main/java/htsjdk/samtools/util/SamLocusIterator.java
@@ -24,6 +24,9 @@
 package htsjdk.samtools.util;
 
 import htsjdk.samtools.AlignmentBlock;
+import htsjdk.samtools.Cigar;
+import htsjdk.samtools.CigarElement;
+import htsjdk.samtools.CigarOperator;
 import htsjdk.samtools.SAMException;
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SAMRecord;
@@ -75,12 +78,16 @@ public class SamLocusIterator implements Iterable<SamLocusIterator.LocusInfo>, C
 
     /**
      * The unit of iteration.  Holds information about the locus (the SAMSequenceRecord and 1-based position
-     * on the reference), plus List of ReadAndOffset objects, one for each read that overlaps the locus
+     * on the reference), plus List of ReadAndOffset objects, one for each read that overlaps the locus;
+     * two more List_s_ of ReadAndOffset objects include reads that overlap the locus with insertions and deletions
+     * respectively
      */
     public static final class LocusInfo implements Locus {
         private final SAMSequenceRecord referenceSequence;
         private final int position;
         private final List<RecordAndOffset> recordAndOffsets = new ArrayList<RecordAndOffset>(100);
+        private List<RecordAndOffset> deletedInRecord = null;
+        private List<RecordAndOffset> insertedInRecord = null;
 
         LocusInfo(final SAMSequenceRecord referenceSequence, final int position) {
             this.referenceSequence = referenceSequence;
@@ -92,14 +99,49 @@ public class SamLocusIterator implements Iterable<SamLocusIterator.LocusInfo>, C
             recordAndOffsets.add(new RecordAndOffset(read, position));
         }
 
+        /** Accumulate info for one read with a deletion */
+        public void addDeleted(final SAMRecord read, int previousPosition) {
+            if(deletedInRecord == null) {
+                deletedInRecord = new ArrayList<>();
+            }
+            deletedInRecord.add(new RecordAndOffset(read, previousPosition));
+        }
+
+        /**
+         * Accumulate info for one read with an insertion.
+         * For this locus, the reads in the insertion are included also in recordAndOffsets
+         */
+        public void addInserted(final SAMRecord read, int firstPosition) {
+            if(insertedInRecord == null) {
+                insertedInRecord = new ArrayList<>();
+            }
+            insertedInRecord.add(new RecordAndOffset(read, firstPosition));
+        }
+
         public int getSequenceIndex() { return referenceSequence.getSequenceIndex(); }
 
         /** @return 1-based reference position */
         public int getPosition() { return position; }
         public List<RecordAndOffset> getRecordAndPositions() { return Collections.unmodifiableList(recordAndOffsets); }
+        public List<RecordAndOffset> getDeletedInRecord() {
+            return (deletedInRecord == null) ? Collections.emptyList() : Collections.unmodifiableList(deletedInRecord);
+        }
+        public List<RecordAndOffset> getInsertedInRecord() {
+            return (insertedInRecord == null) ? Collections.emptyList() : Collections.unmodifiableList(insertedInRecord);
+        }
         public String getSequenceName() { return referenceSequence.getSequenceName(); }
         @Override public String toString() { return referenceSequence.getSequenceName() + ":" + position; }
         public int getSequenceLength() {return referenceSequence.getSequenceLength();}
+
+        /**
+         * @return <code>true</code> if all the RecordAndOffset lists are empty;
+         *         <code>false</code> if at least one have records
+         */
+        public boolean isEmpty() {
+            return recordAndOffsets.isEmpty() &&
+                    (deletedInRecord == null || deletedInRecord.isEmpty()) &&
+                    (insertedInRecord == null || insertedInRecord.isEmpty());
+        }
     }
 
 
@@ -154,6 +196,12 @@ public class SamLocusIterator implements Iterable<SamLocusIterator.LocusInfo>, C
 
     // Set to true when we have enforced the accumulation limit for the first time
     private boolean enforcedAccumulationLimit = false;
+
+    /**
+     * If true, include indels in the LocusInfo
+     */
+    private boolean includeIndels = false;
+
 
     // When there is a target mask, these members remember the last locus for which a LocusInfo has been
     // returned, so that any uncovered locus in the target mask can be covered by a 0-coverage LocusInfo
@@ -300,8 +348,14 @@ public class SamLocusIterator implements Iterable<SamLocusIterator.LocusInfo>, C
                 continue;
             }
 
-            final Locus alignmentStart = new LocusImpl(rec.getReferenceIndex(), rec.getAlignmentStart());
-
+            int start = rec.getAlignmentStart();
+            // only if we are including indels and the record does not start in the first base of the reference
+            // the stop locus to populate the queue is not the same if the record starts with an insertion
+            if(includeIndels && start != 1 && startWithInsertion(rec.getCigar())) {
+                // the start to populate is one less
+                start--;
+            }
+            final Locus alignmentStart = new LocusImpl(rec.getReferenceIndex(), start);
             // emit everything that is before the start of the current read, because we know no more
             // coverage will be accumulated for those loci.
             while (!accumulator.isEmpty() && locusComparator.compare(accumulator.get(0), alignmentStart) < 0) {
@@ -316,17 +370,22 @@ public class SamLocusIterator implements Iterable<SamLocusIterator.LocusInfo>, C
             }
 
             // at this point, either the accumulator list is empty or the head should
-            // be the same position as the first base of the read
+            // be the same position as the first base of the read (or insertion if first)
             if (!accumulator.isEmpty()) {
                 if (accumulator.get(0).getSequenceIndex() != rec.getReferenceIndex() ||
-                        accumulator.get(0).position != rec.getAlignmentStart()) {
+                        accumulator.get(0).position != start) {
                     throw new IllegalStateException("accumulator should be empty or aligned with current SAMRecord");
                 }
             }
 
             // Store the loci for the read in the accumulator
-            if (!surpassedAccumulationThreshold()) accumulateSamRecord(rec);
-
+            if (!surpassedAccumulationThreshold()) {
+                accumulateSamRecord(rec);
+                // Store the indels if requested
+                if(includeIndels) {
+                    accumulateIndels(rec);
+                }
+            }
             samIterator.next();
         }
 
@@ -369,19 +428,29 @@ public class SamLocusIterator implements Iterable<SamLocusIterator.LocusInfo>, C
     }
 
     /**
+     * Check if cigar start with an insertion, ignoring softclip
+     * @param cigar the cigar element
+     * @return <code>true</code> if it start with an insertion; <code>false</code> otherwise
+     */
+    private static boolean startWithInsertion(final Cigar cigar) {
+        for(final CigarElement element: cigar.getCigarElements()) {
+            switch(element.getOperator()) {
+                case I: return true;
+                case S: continue;
+                default: break;
+            }
+        }
+        return false;
+    }
+
+    /**
      * Capture the loci covered by the given SAMRecord in the LocusInfos in the accumulator,
      * creating new LocusInfos as needed.
      */
     private void accumulateSamRecord(final SAMRecord rec) {
-        final SAMSequenceRecord ref = getReferenceSequence(rec.getReferenceIndex());
-        final int alignmentStart  = rec.getAlignmentStart();
-        final int alignmentEnd    = rec.getAlignmentEnd();
-        final int alignmentLength = alignmentEnd - alignmentStart;
 
-        // Ensure there are LocusInfos up to and including this position
-        for (int i=accumulator.size(); i<=alignmentLength; ++i) {
-            accumulator.add(new LocusInfo(ref, alignmentStart + i));
-        }
+        // get the accumulator offset
+        int accOffset = getAccumulatorOffset(rec);
 
         final int minQuality = getQualityScoreCutoff();
         final boolean dontCheckQualities = minQuality == 0;
@@ -397,13 +466,48 @@ public class SamLocusIterator implements Iterable<SamLocusIterator.LocusInfo>, C
                 // 0-based offset into the read of the current base
                 final int readOffset = readStart + i - 1;
 
-                // 0-based offset from the aligned position of the first base in the read to the aligned position of the current base.
-                final int refOffset =  refStart + i - alignmentStart;
-
                 // if the quality score cutoff is met, accumulate the base info
                 if (dontCheckQualities || baseQualities[readOffset] >= minQuality) {
+                    // 0-based offset from the aligned position of the first base in the read to the aligned position of the current base.
+                    final int refOffset = refStart + i - accOffset;
                     accumulator.get(refOffset).add(rec, readOffset);
                 }
+            }
+        }
+    }
+
+    /**
+     * Requires that the accumulator for the record is previously fill with
+     * {@link #accumulateSamRecord(htsjdk.samtools.SAMRecord)}.
+     * Include in the LocusInfo the indels; the quality threshold does not affect insertions/deletions
+     */
+    private void accumulateIndels(final SAMRecord rec) {
+        // get the cigar elements
+        final List<CigarElement> cigar = rec.getCigar().getCigarElements();
+        // 0-based offset into the read of the current base
+        int readBase = 0;
+        // 0-based offset for the reference of the current base
+        // the accumulator could have the previous position because an indel is accumulating
+        int refBase = rec.getAlignmentStart() - getAccumulatorOffset(rec);
+        // iterate over the cigar element
+        for (int elementIndex = 0; elementIndex < cigar.size(); elementIndex++) {
+            final CigarElement e = cigar.get(elementIndex);
+            final CigarOperator operator = e.getOperator();
+            if (operator.equals(CigarOperator.I)) {
+                System.err.println("");
+                // insertions are included in the previous base
+                accumulator.get(refBase - 1).addInserted(rec, readBase);
+                readBase += e.getLength();
+            } else if (operator.equals(CigarOperator.D)) {
+                // accumulate for each position that spans the deletion
+                for (int i = 0; i < e.getLength(); i++) {
+                    // the offset is the one for the previous base
+                    accumulator.get(refBase + i).addDeleted(rec, readBase - 1);
+                }
+                refBase += e.getLength();
+            } else {
+                if (operator.consumesReadBases()) readBase += e.getLength();
+                if (operator.consumesReferenceBases()) refBase += e.getLength();
             }
         }
     }
@@ -456,8 +560,8 @@ public class SamLocusIterator implements Iterable<SamLocusIterator.LocusInfo>, C
      */
     private void populateCompleteQueue(final Locus stopBeforeLocus) {
         // Because of gapped alignments, it is possible to create LocusInfo's with no reads associated with them.
-        // Skip over these.
-        while (!accumulator.isEmpty() && accumulator.get(0).getRecordAndPositions().isEmpty() &&
+        // Skip over these if not including indels
+        while (!accumulator.isEmpty() && accumulator.get(0).isEmpty() &&
                locusComparator.compare(accumulator.get(0), stopBeforeLocus) < 0) {
             accumulator.remove(0);
         }
@@ -492,6 +596,29 @@ public class SamLocusIterator implements Iterable<SamLocusIterator.LocusInfo>, C
 
         lastReferenceSequence = sequenceIndex;
         lastPosition = locusInfo.getPosition();
+    }
+
+    /**
+     * Ensure that the queue is populated and get the accumulator offset for the current record
+     */
+    private int getAccumulatorOffset(SAMRecord rec) {
+        final SAMSequenceRecord ref = getReferenceSequence(rec.getReferenceIndex());
+        final int alignmentStart  = rec.getAlignmentStart();
+        final int alignmentEnd    = rec.getAlignmentEnd();
+        final int alignmentLength = alignmentEnd - alignmentStart;
+        // get the offset for an insertion if we are tracking them
+        final int insOffset =
+                (includeIndels && startWithInsertion(rec.getCigar())) ? 1 : 0;
+        // if there is an insertion in the first base and it is not tracked in the accumulator, add it
+        if (insOffset == 1 && accumulator.isEmpty()) {
+            accumulator.add(new LocusInfo(ref, alignmentStart - 1));
+        }
+
+        // Ensure there are LocusInfos up to and including this position
+        for (int i=accumulator.size(); i<=alignmentLength+insOffset; ++i) {
+            accumulator.add(new LocusInfo(ref, alignmentStart + i - insOffset));
+        }
+        return alignmentStart - insOffset;
     }
 
     private SAMSequenceRecord getReferenceSequence(final int referenceSequenceIndex) {
@@ -548,5 +675,14 @@ public class SamLocusIterator implements Iterable<SamLocusIterator.LocusInfo>, C
      * cap is applied (the first maxReadsToAccumulatePerLocus reads are kept and all subsequent ones are dropped).
      */
     public void setMaxReadsToAccumulatePerLocus(final int maxReadsToAccumulatePerLocus) { this.maxReadsToAccumulatePerLocus = maxReadsToAccumulatePerLocus; }
+
+    public boolean isIncludeIndels() {
+        return includeIndels;
+    }
+
+    public void setIncludeIndels(final boolean includeIndels) {
+        this.includeIndels = includeIndels;
+    }
+
 }
 
--- a/src/main/java/htsjdk/samtools/util/SamLocusIterator.java
+++ b/src/main/java/htsjdk/samtools/util/SamLocusIterator.java
@@ -23,26 +23,10 @@
  */
 package htsjdk.samtools.util;
 
-import htsjdk.samtools.AlignmentBlock;
-import htsjdk.samtools.Cigar;
-import htsjdk.samtools.CigarElement;
-import htsjdk.samtools.CigarOperator;
-import htsjdk.samtools.SAMException;
-import htsjdk.samtools.SAMFileHeader;
-import htsjdk.samtools.SAMRecord;
-import htsjdk.samtools.SAMSequenceRecord;
-import htsjdk.samtools.SamReader;
-import htsjdk.samtools.filter.AggregateFilter;
-import htsjdk.samtools.filter.DuplicateReadFilter;
-import htsjdk.samtools.filter.FilteringSamIterator;
-import htsjdk.samtools.filter.SamRecordFilter;
-import htsjdk.samtools.filter.SecondaryOrSupplementaryFilter;
+import htsjdk.samtools.*;
+import htsjdk.samtools.filter.*;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
+import java.util.*;
 
 /**
  * Iterator that traverses a SAM File, accumulating information on a per-locus basis.
@@ -101,7 +85,7 @@ public class SamLocusIterator implements Iterable<SamLocusIterator.LocusInfo>, C
 
         /** Accumulate info for one read with a deletion */
         public void addDeleted(final SAMRecord read, int previousPosition) {
-            if(deletedInRecord == null) {
+            if (deletedInRecord == null) {
                 deletedInRecord = new ArrayList<>();
             }
             deletedInRecord.add(new RecordAndOffset(read, previousPosition));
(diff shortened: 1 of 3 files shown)


tfenne:
Hi @magicDGS, thanks for doing this!  I'll try and take a look at this in the next day or so.  In the mean time two quick things:
1. It looks like the test that is failing is a new test that you added - do you know what's going on?
2. Your edits introduce a lot of hard-tabs.  Some time before merging those'll need to get turned into soft-tabs with tab size=4.

magicDGS:
Sorry, I forgot to add some changes in my local repository before PR. It should be fixed in commit b48aa0aba328f3bf6bb2741904c607a011c55292

magicDGS:
Fixed hard-tabs in commit 1572366678cd51a5ea64eb8c31b888be95331a26

src/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -75,12 +77,15 @@ public RecordAndOffset(final SAMRecord record, final int offset) {
 
     /**
      * The unit of iteration.  Holds information about the locus (the SAMSequenceRecord and 1-based position
-     * on the reference), plus List of ReadAndOffset objects, one for each read that overlaps the locus
+     * on the reference), plus List of ReadAndOffset objects, one for each read that overlaps the locus;
+     * two more List of ReadAndOffset objects includes reads that overlaps the locus with insertions/deletions
```
tfenne: this doesn't read quite right to me.  How about:
two more List_s_ of ReadAndOffset objects include reads that overlap the locus with insertions and deletions respectively.

src/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -75,12 +77,15 @@ public RecordAndOffset(final SAMRecord record, final int offset) {
...
      */
     public static final class LocusInfo implements Locus {
         private final SAMSequenceRecord referenceSequence;
         private final int position;
         private final List<RecordAndOffset> recordAndOffsets = new ArrayList<RecordAndOffset>(100);
+        private final List<RecordAndOffset> deletedInRecord = new ArrayList<RecordAndOffset>();
```
tfenne: This class is already quite slow due to all the object creation involved.  What do you think about creating these two ArrayList objects lazily at the first call to `addDeleted` and `addInsertion` since they may frequently be empty?

src/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -75,12 +77,15 @@ public RecordAndOffset(final SAMRecord record, final int offset) {
...
+        private final List<RecordAndOffset> insertionInRecord = new ArrayList<RecordAndOffset>();
```
tfenne: The two new names, and their methods, should have matching names.  Either `deletedInRecord` and `insertedInRecord` or `deletionInRecord` and `insertionInRecord`.  Personally I prefer the first pair.

src/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -353,9 +395,17 @@ private void accumulateSamRecord(final SAMRecord rec) {
         final int alignmentEnd    = rec.getAlignmentEnd();
         final int alignmentLength = alignmentEnd - alignmentStart;
 
+        // get the offset for an insertion if we are tracking them
```
tfenne: Now that it's a little more complicated, could you please extract lines ~398-408 into a separate method called something like `ensureAccumulatorIsPopulated()`?

src/tests/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -64,6 +64,34 @@ public void testBasicIterator() {
...
+            Assert.assertEquals(li.getDeletedInRecord().size(), 0);
+            Assert.assertEquals(li.getInsertionInRecord().size(), 0);
+        }
+
+    }
+
+    @Test
+    public void testBasicIteratorWithIndels() {
+
+        final String sqHeader = "@HD\tSO:coordinate\tVN:1.0\n@SQ\tSN:chrM\tAS:HG18\tLN:100000\n";
```
tfenne: I feel bad making this comment, because this is totally not your fault.  You've made changes in a class that uses an outdated way of building SAMRecords for testing, and followed the pattern you found.  But I'd really prefer not to add lots more tests that do things this way.  Would you mind taking a look at `SAMRecordSetBuilder` and seeing if you could convert the tests to use that instead?

src/tests/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -64,6 +64,34 @@ public void testBasicIterator() {
...
             Assert.assertEquals(2, li.getRecordAndPositions().size());
+            // make sure that we are not accumulating indels
+            Assert.assertEquals(li.getDeletedInRecord().size(), 0);
+            Assert.assertEquals(li.getInsertionInRecord().size(), 0);
+        }
+
+    }
+
+    @Test
+    public void testBasicIteratorWithIndels() {
```
tfenne: `testIteratorWithIndelsOnButNoIndels`?

src/tests/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -93,6 +121,43 @@ public void testEmitUncoveredLoci() {
...
+            // make sure that we are not accumulating indels
+            Assert.assertEquals(li.getDeletedInRecord().size(), 0);
+            Assert.assertEquals(li.getInsertionInRecord().size(), 0);
+        }
+        Assert.assertEquals(pos, 100001);
+
+    }
+
+    @Test
+    public void testEmitUncoveredLociWithIndels() {
```
tfenne: Instead of copy/pasting the whole method with an additional call to `setIncludeIndels(true)` you could just wrap the body of the original method with `for (final boolean incIndels : new boolean[][ {true, false}) {`

src/tests/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -119,6 +184,37 @@ public void testQualityFilter() {
...
             Assert.assertEquals(pos++, li.getPosition());
+            // make sure that we are not accumulating indels
+            Assert.assertEquals(li.getDeletedInRecord().size(), 0);
+            Assert.assertEquals(li.getInsertionInRecord().size(), 0);
+        }
+
+    }
+
+    @Test
+    public void testQualityFilterWithIndels() {
```
tfenne: Same thing here about not duplicating the test code.

tfenne:
Hi @magicDGS there's one case that I think doesn't work if I've understood the code correctly.  Either way it would be great to add a test for it to ensure/show that it works.  My concern is that I think if you encounter a stream of record with the same alignment start position, and the _first_ read starts with an `M` operator and some subsequent read starts with an `I` operator, things will go badly.  E.g.

```
R1	65	chr4	5000	60	101M	...
R2	65	chr4	5000	60	6I95M	...
R3	65	chr4	5000	60	101M	...
R4	65	chr4	5000	60	101M	...
```

I think in this case that when `R1` is encountered that the accumulator will be flushed up to position `5000`, and then all the bases from `R1` accumulated.  And then when `R2` is encountered, it will expect that `accumulator[0]` is in fact for position 4999, and the exception on line 349 will be thrown.

Since you can't rely on any specific ordering of reads with the same alignment start position, you may need to change things further such that you pull off all the reads with the same alignment start position, order them by whether or not they start with an indel, and then accumulate from all of them.

src/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -384,6 +434,56 @@ private void accumulateSamRecord(final SAMRecord rec) {
...
+        final List<CigarElement> cigar = rec.getCigar().getCigarElements();
+        // 0-based offset into the read of the current base
+        int readBase = 0;
+        // 0-based offset for the reference of the current base
+        // the accumulator could have the previous position because an indel is accumulating
+        int refBase = 0;
+        // iterate over the cigar element
+        for (int elementIndex = 0; elementIndex < cigar.size(); elementIndex++) {
+            final CigarElement e = cigar.get(elementIndex);
+            switch (e.getOperator()) {
```
tfenne: This switch statements would, I think, be clearer written as follows:
```java
if (operator == I) // handle insertion
else if (operator == D) // handle deletion
else {
    if (operator.consumesReadBases()) readBase += e.getLength();
    if (operator.consumesReferenceBases()) refBase += e.getLength();
}
```

tfenne:
@magicDGS I've made a first pass through the code, but will likely have more comments once you've addressed the first batch.

@alecw Would you mind taking a look at this PR, perhaps after @magicDGS has made the first round of revisions?  As the original author of `SamLocusIterator` I think it would be good for you to take a look at this.

src/java/htsjdk/samtools/util/SamLocusIterator.java
magicDGS: Done in ac5c8b5360ae8fca32325ae304578bd0929736dc

src/java/htsjdk/samtools/util/SamLocusIterator.java
magicDGS: Done in ac5c8b5360ae8fca32325ae304578bd0929736dc

src/java/htsjdk/samtools/util/SamLocusIterator.java
magicDGS: Done in ac5c8b5360ae8fca32325ae304578bd0929736dc

src/java/htsjdk/samtools/util/SamLocusIterator.java
magicDGS: I extracted the computation of the offset for the queue in  the method `getAccumulatorOffset` in commit 7c9be2b1d221233cadb5180b50984e6760888b87

src/java/htsjdk/samtools/util/SamLocusIterator.java
magicDGS: Done in 5d23bdd2f9fca3027257e060a3dd8741678eaf2c

src/tests/java/htsjdk/samtools/util/SamLocusIteratorTest.java
magicDGS: I've just followed the pattern found to keep the class without many changes, but I could have a look to `SAMRecordSetBuilder` and use it instead. In addition, I also found that some of the `Asser.assertEquals()` from the original class are in the wrong order (expected/actual instead of actual/expected). Should I change that too?

src/tests/java/htsjdk/samtools/util/SamLocusIteratorTest.java
magicDGS: Sure, it will also reduce the number of methods (I will do it also with `testBasicIterator` to avoid the suggested `testIteratorWithIndelsOnButNoIndels`) in the previous comment.

src/tests/java/htsjdk/samtools/util/SamLocusIteratorTest.java
tfenne: If you're willing that would be great.

magicDGS:
Hi @tfenne, it is true that the iterator is not working on the test case that you mention and I will modify the `testStartWithInsertion` method to develop what you suggest. Thank you very much for your revision, and I will let you know when I fix that issue.

src/tests/java/htsjdk/samtools/util/SamLocusIteratorTest.java
magicDGS: Done in dbf3c75b170adda515385f52e6302701fd5cde38

src/tests/java/htsjdk/samtools/util/SamLocusIteratorTest.java
magicDGS: Added to `testBasicIterator` in dbf3c75b170adda515385f52e6302701fd5cde38

src/tests/java/htsjdk/samtools/util/SamLocusIteratorTest.java
magicDGS: Done in dbf3c75b170adda515385f52e6302701fd5cde38

src/tests/java/htsjdk/samtools/util/SamLocusIteratorTest.java
magicDGS: Done in dbf3c75b170adda515385f52e6302701fd5cde38

magicDGS:
Hello @tfenne, for a reason that I do not fully understand, `testEmitUncoveredLoci` is broken when using the `SAMRecordSetBuilder` (commit 9d14bb9fa74a6fbd375515cd9ced0a61c442d4c2, see the file [here](https://github.com/samtools/htsjdk/blob/9d14bb9fa74a6fbd375515cd9ced0a61c442d4c2/src/tests/java/htsjdk/samtools/util/SamLocusIteratorTest.java)), but the implementation is the same as in the commit where the test pass (5d23bdd2f9fca3027257e060a3dd8741678eaf2c, see the file [here](https://github.com/magicDGS/htsjdk/blob/5d23bdd2f9fca3027257e060a3dd8741678eaf2c/src/tests/java/htsjdk/samtools/util/SamLocusIteratorTest.java)), although separating the `setIncludeIndels`.

Do you have any clue about what is happening?

alecw:
Truth be told, I'm not the original author.  We inherited this code from @kcibul.  Anyway, this code is pretty complicated (apart from the changes) and I don't think I have the time to get my head into this enough to comment intelligently. 

One suggestion I would make, however, is that it would be good for the javadoc to explain how insertions and deletions will be handled differently according to the value of includeIndels. As it stands now I don't know what the existing behavior is and what the new behavior will be.

-Alec

magicDGS:
Although I'm still working on this, I found that GATK4 already implemented a `LocusIteratorByState` (https://github.com/broadinstitute/gatk/pull/1442) that solve the initial issue.

I do not know if it is worth to implement this option in htsjdk or if someone is interested in that behavior use directly the GATK4 framework. What do you think, @tfenne?

droazen:
@tfenne Are you still interested in getting this into htsjdk? It sounds like @magicDGS has found a solution in GATK4 for the original need that prompted this PR. If you are interested, could you comment before the next htsjdk review party on 5/3? Otherwise, we might close this due to lack of interest.

magicDGS:
I found the solution in GATK4, but I think that it is still important in htsjdk. But it will probably be better for the new API if it is implemented (#520), using `AlignmentContext` as in GATK.

src/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -287,8 +333,14 @@ public LocusInfo next() {
                 continue;
             }
 
-            final Locus alignmentStart = new LocusImpl(rec.getReferenceIndex(), rec.getAlignmentStart());
-
+            int start = rec.getAlignmentStart();
+            // only if we are including indels and the record does not start in the first base of the reference
+            // the stop locus to populate the queue is not the same if the record starts with an insertion
+            if(includeIndels && start != 1 && rec.getCigar().getCigarElement(0).getOperator().equals(CigarOperator.I)) {
```
yfarjoun: what if there's a clip and then an insertion, e.g 5S6I4M . In this case, you'll not notice that it is in-fact an insertion that comes before the alignment start. (needs a test)

src/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -431,8 +515,9 @@ else if (nextbit >= stopBeforeLocus.getPosition()) {
      */
     private void populateCompleteQueue(final Locus stopBeforeLocus) {
         // Because of gapped alignments, it is possible to create LocusInfo's with no reads associated with them.
-        // Skip over these.
-        while (!accumulator.isEmpty() && accumulator.get(0).getRecordAndPositions().isEmpty() &&
+        // Skip over these if not including indels
+        while (!accumulator.isEmpty() && accumulator.get(0).isEmpty() &&
+                // accumulator.get(0).getRecordAndPositions().isEmpty() && !(includeIndels && !accumulator.get(0).getDeletedInRecord().isEmpty()) &&
```
yfarjoun: why is this code commented? should it be removed?

src/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -469,6 +554,29 @@ private void populateCompleteQueue(final Locus stopBeforeLocus) {
...
+    /**
+     * Ensure that the queue is populated and get the accumulator offset for the current record
+     */
+    private int getAccumulatorOffset(SAMRecord rec) {
+        final SAMSequenceRecord ref = getReferenceSequence(rec.getReferenceIndex());
+        final int alignmentStart  = rec.getAlignmentStart();
+        final int alignmentEnd    = rec.getAlignmentEnd();
+        final int alignmentLength = alignmentEnd - alignmentStart;
+        // get the offset for an insertion if we are tracking them
+        final int insOffset =
```
yfarjoun: same issue as before regarding softclips.

src/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -469,6 +554,29 @@ private void populateCompleteQueue(final Locus stopBeforeLocus) {
...
+                (includeIndels && rec.getCigar().getCigarElement(0).getOperator().equals(CigarOperator.I)) ? 1 : 0;
+        // if there is an insertion in the first base and it is not tracked in the accumulator, add it
+        if(insOffset == 1 && accumulator.isEmpty()) {
```
yfarjoun:
```code
if (   
```
(needs a space)

src/tests/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -161,10 +196,91 @@ public void testSimpleGappedAlignment() {
...
+            Assert.assertEquals(li.getInsertedInRecord().size(), 0);
+            ++i;
+        }
+    }
+
+    /**
+     * Try all CIGAR operands (except H and P) and confirm that loci produced by SamLocusIterator are as expected with indels
+     */
+    @Test
+    public void testSimpleGappedAlignmentWithIndels() {
```
yfarjoun: would be better to include this in a loop over `includeIndels` as you did in other tests.

src/java/htsjdk/samtools/util/SamLocusIterator.java
magicDGS: Added test. Bug fix soon.

src/java/htsjdk/samtools/util/SamLocusIterator.java
magicDGS: Removed.

src/tests/java/htsjdk/samtools/util/SamLocusIteratorTest.java
magicDGS: Done.

src/java/htsjdk/samtools/util/SamLocusIterator.java
magicDGS: Done.

src/java/htsjdk/samtools/util/SamLocusIterator.java
magicDGS: Done.

magicDGS:
I think that I fixed the bugs. Back to @yfarjoun.

magicDGS:
@akiezun, could you have a look to the failed test when both `testSimpleGappedAlignment()` were merged together? I don't understand why it is failing (it could be possible that it was the reason why I have them separate).

yfarjoun:
I think that this is a victim of the switch to gradle. Can you rebase this (without squashing) on top of master?

magicDGS:
It is already rebased, @yfarjoun, and I force the push with the squashed commits (so I can't retrieve the previous test). I can implement separate tests again from scratch to check if it is still not working.

yfarjoun:
In case you do not know: if you really want to recover your unsquashed
commits, you can look in `git reflog`..

yfarjoun:
@lbergelson @droazen I think that this is due to the move to gradle...could you take a look?

magicDGS:
Thanks @yfarjoun, I didn't know and it will be useful. I think that there is a bug and not a gradle problem, but let's see what they think about.

magicDGS:
@lbergelson @droazen, it is not a problem of gradle. There is a bug in the code when accumulating deletions in a complicated case. I'll work on it.

yfarjoun:
Thanks Daniel.

magicDGS:
@yfarjoun, I reorganized commits and squashed them for an easier review. I solved the bug in the code when tracking indels, and added some simple tests for them. Could you have a look?

src/main/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -369,19 +428,29 @@ private boolean surpassedAccumulationThreshold() {
...
     /**
+     * Check if cigar start with an insertion, ignoring softclip
+     * @param cigar the cigar element
+     * @return <code>true</code> if it start with an insertion; <code>false</code> otherwise
+     */
+    private static boolean startWithInsertion(final Cigar cigar) {
+        for(final CigarElement element: cigar.getCigarElements()) {
+            switch(element.getOperator()) {
+                case I: return true;
+                case S: continue;
```
yfarjoun: this should be any operator that doesn't consume refbases (except "I"). I think it would be better to implement this as an if statement. Perhaps something like: 
```
<SNIP>
if( element.getOperator()==CigarOperator.I ) return true;
if ( ! element.getOperator().consumesReferenceBases() )  continue;
break;
</SNIP>
```

src/main/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -369,19 +428,29 @@ private boolean surpassedAccumulationThreshold() {
     }
 
     /**
+     * Check if cigar start with an insertion, ignoring softclip
```
yfarjoun: ...ignoring other operators that do not consume references bases

src/main/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -369,19 +428,29 @@ private boolean surpassedAccumulationThreshold() {
...
+     * @param cigar the cigar element
+     * @return <code>true</code> if it start with an insertion; <code>false</code> otherwise
```
yfarjoun: ...if the first operator to consume reference bases or be an insertion, is an insertion...

src/main/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -369,19 +428,29 @@ private boolean surpassedAccumulationThreshold() {
     }
 
     /**
+     * Check if cigar start with an insertion, ignoring softclip
+     * @param cigar the cigar element
```
yfarjoun: remove the word element.

src/main/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -300,8 +348,14 @@ public LocusInfo next() {
                 continue;
             }
 
-            final Locus alignmentStart = new LocusImpl(rec.getReferenceIndex(), rec.getAlignmentStart());
-
+            int start = rec.getAlignmentStart();
+            // only if we are including indels and the record does not start in the first base of the reference
+            // the stop locus to populate the queue is not the same if the record starts with an insertion
+            if(includeIndels && start != 1 && startWithInsertion(rec.getCigar())) {
```
yfarjoun:
```
if (
```
(space)

src/main/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -494,6 +598,29 @@ private void populateCompleteQueue(final Locus stopBeforeLocus) {
...
+        // get the offset for an insertion if we are tracking them
+        final int insOffset =
+                (includeIndels && startWithInsertion(rec.getCigar())) ? 1 : 0;
+        // if there is an insertion in the first base and it is not tracked in the accumulator, add it
+        if (insOffset == 1 && accumulator.isEmpty()) {
+            accumulator.add(new LocusInfo(ref, alignmentStart - 1));
+        }
+
+        // Ensure there are LocusInfos up to and including this position
+        for (int i=accumulator.size(); i<=alignmentLength+insOffset; ++i) {
```
yfarjoun: spaces around operators

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -23,128 +23,327 @@
...
-        final String seq1 = "ACCTACGTTCAATATTACAGGCGAACATACTTACTA";
-        final String qual1 = "++++++++++++++++++++++++++++++++++++"; // phred 10
-        final String s1 = "3851612\t16\tchrM\t165\t255\t36M\t*\t0\t0\t" + seq1 + "\t" + qual1 + "\n";
-        final String exampleSam = sqHeader + s1 + s1;
+        final SAMRecordSetBuilder builder = getRecordBuilder();
+        // add records up to coverage for the test in that position
+        final int startPosition = 165;
+        for(int i = 0; i < coverage; i++) {
+            // add a negative-strand fragment mapped on chrM with base quality of 10
+            builder.addFrag("record"+i, 0, startPosition, true, false, "36M", null, 10);
```
yfarjoun: spaces

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -23,128 +23,327 @@
...
+     */
     @Test
     public void testQualityFilter() {
+        final SAMRecordSetBuilder builder = getRecordBuilder();
+        // add records up to coverage for the test in that position
+        final int startPosition = 165;
+        for(int i = 0; i < coverage; i++) {
+            final String qualityString;
+            // half of the reads have a different quality
+            if(i % 2 == 0) {
```
yfarjoun:
``` 
if (
```

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -23,128 +23,327 @@
...
+                qualityString = null;
+            } else {
+                qualityString = "+*+*+*+*+*+*+*+*+*+*+*+*+*+*+*+*+*+*";
+            }
+            // add a negative-strand fragment mapped on chrM with base quality of 10
+            builder.addFrag("record"+i, 0, startPosition, true, false, "36M", qualityString, 10);
```
yfarjoun: spaces

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -23,128 +23,327 @@
...
+     * Test a simple deletion, with both including or not indels
+     */
+    @Test
+    public void testSimpleDeletion() {
+        final SAMRecordSetBuilder builder = getRecordBuilder();
+        // add records up to coverage for the test in that position
+        final int startPosition = 165;
+        for(int i = 0; i < coverage; i++) {
+            // add a negative-strand fragment mapped on chrM with base quality of 10
+            builder.addFrag("record"+i, 0, startPosition, true, false, "18M10D18M", null, 10);
```
yfarjoun: spaces

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -23,128 +23,327 @@
...
+            // make sure we accumulated depth for each position
+            int pos = startPosition;
+            for (final SamLocusIterator.LocusInfo li : sli) {
+                boolean isDeletedPosition = (pos >= deletionStart && pos <= deletionEnd);
+                if(!incIndels && isDeletedPosition) {
+                    pos = deletionEnd + 1;
+                    isDeletedPosition = false;
+                }
+                Assert.assertEquals(li.getPosition(), pos++);
+                if(isDeletedPosition) {
```
yfarjoun:
```
if (
```

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -23,128 +23,327 @@
...
+            sli.setIncludeIndels(incIndels);
+            // make sure we accumulated depth for each position
+            int pos = startPosition;
+            for (final SamLocusIterator.LocusInfo li : sli) {
+                Assert.assertEquals(li.getPosition(), pos++);
+                // make sure we are accumulating normal coverage
+                Assert.assertEquals(li.getRecordAndPositions().size(), coverage);
+                // make sure that we are not accumulating deletions
+                Assert.assertEquals(li.getDeletedInRecord().size(), 0);
+                if(incIndels && li.getPosition() == insStart) {
```
yfarjoun:
```
if (
```

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -23,128 +23,327 @@
...
 
+	/**
+     * Test a simple insertion, with both including or not indels
+     */
+    @Test
+    public void testSimpleInsertion() {
+        final SAMRecordSetBuilder builder = getRecordBuilder();
+        // add records up to coverage for the test in that position
+        final int startPosition = 165;
+        for(int i = 0; i < coverage; i++) {
```
yfarjoun:
```
for (
```

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -23,128 +23,327 @@
...
+        final int deletionEnd = 192;
+        // test both for include indels and do not include indels
+        for (final boolean incIndels : new boolean[]{false, true}){
+            final SamLocusIterator sli = createSamLocusIterator(builder);
+            sli.setIncludeIndels(incIndels);
+            // make sure we accumulated depth for each position
+            int pos = startPosition;
+            for (final SamLocusIterator.LocusInfo li : sli) {
+                boolean isDeletedPosition = (pos >= deletionStart && pos <= deletionEnd);
+                if(!incIndels && isDeletedPosition) {
```
yfarjoun:
```
if (
```

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -23,128 +23,327 @@
...
+        // add records up to coverage for the test in that position
+        final int startPosition = 165;
+        for(int i = 0; i < coverage; i++) {
+            // add a negative-strand fragment mapped on chrM with base quality of 10
+            builder.addFrag("record"+i, 0, startPosition, true, false, "18M10D18M", null, 10);
+        }
+        final int deletionStart = 183;
+        final int deletionEnd = 192;
+        // test both for include indels and do not include indels
+        for (final boolean incIndels : new boolean[]{false, true}){
```
yfarjoun:
```
[] {false, true}) {
```

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -23,128 +23,327 @@
...
-        sli.setQualityScoreCutoff(10);
+	/**
+     * Test a simple deletion, with both including or not indels
+     */
+    @Test
+    public void testSimpleDeletion() {
+        final SAMRecordSetBuilder builder = getRecordBuilder();
+        // add records up to coverage for the test in that position
+        final int startPosition = 165;
+        for(int i = 0; i < coverage; i++) {
```
yfarjoun: space

yfarjoun:
Thanks for your work on this PR @magicDGS.

I've commented about your bug fix, and a few of the places where the spaces in your code are missing (sorry for the nit-picking, we are trying to have a consistent code style). There are more that I missed, perhaps you can set your IDE to fix these for you?

I think that there are some CIGAR cases that are missing. 

e.g.

```
2M4N3I4M
2M4N4D5M (is this legal?)
```

magicDGS:
Back to you for (final?) review, @yfarjoun:

- [x] Coding style fixed with IDE
- [x] `startWithInsertion()` implemented as you suggested
- [x] New tests for examples of CIGAR in previous comment

yfarjoun:
Thanks @magicDGS I'm happy with this. 👍 

@tfenne you took a look at this first, and said you might have more comments. Are you happy enough with my review, or would you like another glance?

yfarjoun:
Thanks @magicDGS !!

PR #411, opened 2015-12-17
Switched around the logic if the if-else check.

eitanbanks:
In theory it should never be a problem (the BAM would have to be all sorts of f###ed up),
but in the original version the if statement could error out if the read is not paired.

### DIFF
--- a/src/java/htsjdk/samtools/SAMUtils.java
+++ b/src/java/htsjdk/samtools/SAMUtils.java
@@ -945,14 +945,14 @@ public final class SAMUtils {
             } else {
                 if (getMateCigarString(rec) != null) {
                     ret = new ArrayList<SAMValidationError>();
-                    if (rec.getMateUnmappedFlag()) {
+                    if (!rec.getReadPairedFlag()) {
+                        // If the read is not paired, and the Mate Cigar String (MC Attribute) exists, that is a validation error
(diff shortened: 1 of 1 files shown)


src/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -945,14 +945,14 @@ public static int getMateUnclippedEnd(final SAMRecord rec) {
             } else {
                 if (getMateCigarString(rec) != null) {
                     ret = new ArrayList<SAMValidationError>();
-                    if (rec.getMateUnmappedFlag()) {
+                    if (!rec.getReadPairedFlag()) {
```
yfarjoun: can you explain what you worried about here? I think that this is equivalent due to the if statement above.

eitanbanks: We don't know why the if statement above was false; it's either because the read is not paired or the mate is unmapped.
If the statement failed because the read is not paired, then when it gets to this if statement here (in the original version) it will call rec.getMateUnmappedFlag() - but that will throw an exception because the code there has this check:
        requireReadPaired();
So, we need to switch the order here and have the if make sure that the read is paired.
Make sense?

yfarjoun:
got it. that's sneaky. :+1:

PR #412, opened 2015-12-18
Genotype now treats empty filter strings identically to null filter strings

lbergelson:
There was a potentially invalid case where a genotype could be constructed with an empty filter string.
This would result in `isFiltered()` returning `true` and `getFilters()` being empty.  There are explicit checks in gatk for `genotype.isFiltered() && genotype.getFilters().isEmpty()` so there must be cases where this has happened.  

I don't believe there is any case where this would be meaningful state to represent.


tfenne:
It makes me sad that `VariantContext` represents filters as a `Set<String>`, while `Genotype` represents them as a plain `String` (presumably lifted straight from the VCF).  But this seems like a reasonable fix given the situation.

PR #413, opened 2015-12-18
Open up DiskBasedBAMFileIndex as a public class.

fnothaft:
I was trying to make raw access to BAIs in a downstream project (see https://github.com/HadoopGenomics/Hadoop-BAM/issues/34) and found that the [AbstractBAMFileIndex](https://github.com/samtools/htsjdk/blob/master/src/java/htsjdk/samtools/AbstractBAMFileIndex.java) has no public concrete implementations. As such, I'd like to open this class up as public.


lbergelson:
This seems reasonable.  Sorry for the delay in looking at it.

droazen:
:+1:

PR #416, opened 2015-12-27
Yf add istransition

yfarjoun:
- added isTransition to VariantContext, to tell if a VariantContext is a transition or a transversion. Works on bi-allelic SNPs.
- Added a java8 verion of CollectionUtils.partition which is much easier to use than the previous version (@Deprecated)

### DIFF
--- a/src/java/htsjdk/samtools/util/CollectionUtil.java
+++ b/src/java/htsjdk/samtools/util/CollectionUtil.java
@@ -34,6 +34,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Function;
 
 /**
  * Small utility methods for dealing with collection classes.
@@ -42,34 +43,29 @@ import java.util.Set;
 public class CollectionUtil {
 
     /** Simple case-insensitive lexical comparator of objects using their {@link Object#toString()} value. */
-    final public static Comparator<Object> OBJECT_TOSTRING_COMPARATOR = new Comparator<Object>() {
-        @Override
-        public int compare(final Object o1, final Object o2) {
-            return o1.toString().compareToIgnoreCase(o2.toString());
-        }
-    };
+    final public static Comparator<Object> OBJECT_TOSTRING_COMPARATOR = (o1, o2) -> o1.toString().compareToIgnoreCase(o2.toString());
 
-    public static <T> List<T> makeList (final T... list) {
-        final List<T> result = new ArrayList<T>();
+    public static <T> List<T> makeList(final T... list) {
+        final List<T> result = new ArrayList<>();
         Collections.addAll(result, list);
 
         return result;
     }
-    
-    public static <T> Set<T> makeSet (final T... list) {
-        final Set<T> result = new HashSet<T>();
+
+    public static <T> Set<T> makeSet(final T... list) {
+        final Set<T> result = new HashSet<>();
         Collections.addAll(result, list);
         return result;
     }
-    
-    public static <T> Collection<T> makeCollection (final Iterator<T> i) {
-        final List<T> list = new LinkedList<T>();
+
+    public static <T> Collection<T> makeCollection(final Iterator<T> i) {
+        final List<T> list = new LinkedList<>();
         while (i.hasNext()) {
             list.add(i.next());
         }
         return list;
     }
-    
+
     /** Construct a string by toString()ing each item in the collection with inBetween between each item. */
     public static String join(final Collection<?> items, final String inBetween) {
         final StringBuilder builder = new StringBuilder();
@@ -86,7 +82,7 @@ public class CollectionUtil {
             throw new IllegalArgumentException(String.format("Expected a single element in %s, but found %s.", items, items.size()));
         return items.iterator().next();
     }
-    
+
     /** Simple multi-map for convenience of storing collections in map values. */
     public static class MultiMap<K, V> extends HashMap<K, Collection<V>> {
         public void append(final K k, final V v) {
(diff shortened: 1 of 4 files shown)


yfarjoun:
@tfenne could you take a look here too? This is needed for https://github.com/broadinstitute/picard/issues/368

src/java/htsjdk/samtools/util/CollectionUtil.java
```diff
@@ -101,24 +97,38 @@ public void appendAll(final K k, final Collection<? extends V> v) {
...
+
     public static abstract class Partitioner<V, K> {
         public abstract K getPartition(final V v);
     }
+
+    /** 
+     * Partitions a collection into groups based on a characteristics of that group.  Partitions are embodied in a map, whose keys are the
+     * value of that characteristic, and the values are the partition of elements whose characteristic evaluate to that key.
+     */
+    public static <K, V> Map<K, Collection<V>> partition(final Collection<V> collection, final Function<? super V, ? extends K> keyer) {
```
tfenne: Two thoughts on this:
1) Is this even necessary any more with the ability to do things like:
```collection.stream().collect(Collectors.groupBy(x -> ....)))```
2) Why deprecate the old version and create this new version when you could just make Partitioner a functional interface and be done?

File: src/java/htsjdk/variant/variantcontext/VariantContext.java [hunk 2]
```diff
@@ -1687,30 +1689,17 @@ public boolean hasSymbolicAlleles() {
...
-            final int ac = getCalledChrCount(a);
-            if ( ac >= maxAC1 ) {
-                maxAC1 = ac;
-                best = a;
-            }
-
-        }
-        return best;
+        return getAlternateAlleles().stream()
+                .max((a1, a2) -> Integer.compare(getCalledChrCount(a1), getCalledChrCount(a2)))
```
tfenne: While this is neater, it's going to be a lot slower on big VCFs with many samples and multiple alleles, since `getCalledChrCount(allele)` traverses all the genotypes each time it's called.

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1730,10 +1719,9 @@ public int getAlleleIndex(final Allele allele) {
      * @return a list of indices for each allele, in order
      */
     public List<Integer> getAlleleIndices(final Collection<Allele> alleles) {
-        final List<Integer> indices = new LinkedList<Integer>();
-        for ( final Allele allele : alleles )
-            indices.add(getAlleleIndex(allele));
-        return indices;
+        return alleles.stream()
+                .map(this::getAlleleIndex)
+                .collect(Collectors.toCollection(LinkedList::new));
```
tfenne: I wonder why this uses a LinkedList?  It's almost always better to use an `ArrayList` of appropriate size.

src/java/htsjdk/variant/variantcontext/VariantContextUtils.java
```diff
@@ -338,6 +338,55 @@ public static boolean match(VariantContext vc, Genotype g, JexlVCMatchExp exp) {
...
+     * T->A
+     * T->G
+     *
+     * @param vc a biallelic polymorphic SNP
+     * @return true if a transition and false if transversion
+     * @throws IllegalArgumentException if vc is monomorphic, not a SNP or not bi-allelic.
+
+          */
+
+    static public boolean isTransition(final VariantContext vc) throws IllegalAccessException {
```
tfenne: I think you mean `IllegalArgumentException`

tfenne:
A few minor comments, otherwise :+1:

src/java/htsjdk/variant/variantcontext/VariantContext.java
yfarjoun: boo. OK. I'll revert this.

src/java/htsjdk/samtools/util/CollectionUtil.java
yfarjoun: 1) good point. I guess I could change the usages of partition to use groupBy instead.
2) I did this initially, but then I felt bad for breaking some possible use of a downstream tool that uses htsjdk...so I did it this way to avoid breaking the API...

src/java/htsjdk/variant/variantcontext/VariantContext.java
yfarjoun: sure. I didn't want to change the result without understanding why it was a linked-list to begin with...

src/java/htsjdk/variant/variantcontext/VariantContextUtils.java
yfarjoun: oops. strange I didn't get a warning about an undeclared Exception. thanks.

src/java/htsjdk/variant/variantcontext/VariantContext.java
yfarjoun: actually, I have a workaround that still uses streams, but maps first to a pair...see what you think....

src/java/htsjdk/variant/variantcontext/VariantContextUtils.java
eitanbanks: FYI this is already implemented in GATKVariantContextUtils.java (in the GATK obviously).  We only moved half of the utility code over to htsjdk and kept the other half in the GATK.  You might want to look at the implementation in the GATK and move that over here instead.

yfarjoun:
@tfenne I made some changes in response to your comments. can you please take another look?

yfarjoun:
argh. finally passing tests. @tfenne, what do you think of the solution using AbstractMap.SimpleImmutableEntry?

tfenne:
Looks good to me @yfarjoun.  Btw. did you consider using `htsjdk.samtools.util.Tuple`?

yfarjoun:
I'll use Tuple...thanks!

PR #417, opened 2015-12-29
added utilities to Cigar*

lindenb:
I've added a few functions to Cigar* . I'm working with split reads and I often need to know if the cigar is clipped, what is the nature if the 5' or 3' operator ,etc ...

in **CigarOperator**:
* isClipping
* isPadding
* isIndel
* isIndelOrSkippedRegion
* isAlign

in **Cigar**
* cigarElements is final and so it is never nulll
* now implements Iterable<CIgarElement>
* added static builder from a List of CigarOperator (it cannot be a constructor because a similar constructor exists)
* added 'containsOperator', isLeftClipped, isRightClipped, isClipped
* prevent IDE warnings in 'switch/case' by adding 'default'
* get first/last cigarElement

in **CigarElement**:
* added toString

in **CigarTest**:

added one test for the above functions.

### DIFF
--- a/src/java/htsjdk/samtools/Cigar.java
+++ b/src/java/htsjdk/samtools/Cigar.java
@@ -26,6 +26,7 @@ package htsjdk.samtools;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 
 /**
@@ -37,7 +38,7 @@ import java.util.List;
  *
  * c.f. http://samtools.sourceforge.net/SAM1.pdf for complete CIGAR specification.
  */
-public class Cigar implements Serializable {
+public class Cigar implements Serializable, Iterable<CigarElement> {
     public static final long serialVersionUID = 1L;
 
     private final List<CigarElement> cigarElements = new ArrayList<CigarElement>();
@@ -82,6 +83,8 @@ public class Cigar implements Serializable {
                 case EQ:
                 case X:
                     length += element.getLength();
+                    break;
+                default: break;
             }
         }
         return length;
@@ -101,6 +104,8 @@ public class Cigar implements Serializable {
                 case X:
                 case P:
                     length += element.getLength();
+                    break;
+                default: break;
             }
         }
         return length;
@@ -236,15 +241,15 @@ public class Cigar implements Serializable {
     }
 
     private static boolean isInDelOperator(final CigarOperator op) {
-        return op == CigarOperator.I || op == CigarOperator.D;
+        return op !=null && op.isIndel();
     }
 
     private static boolean isClippingOperator(final CigarOperator op) {
-        return op == CigarOperator.S || op == CigarOperator.H;
+        return op !=null && op.isClipping();
     }
 
     private static boolean isPaddingOperator(final CigarOperator op) {
-        return op == CigarOperator.P;
+        return op !=null && op.isPadding();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


src/java/htsjdk/samtools/Cigar.java
```diff
@@ -254,15 +259,69 @@ public boolean equals(final Object o) {
 
         final Cigar cigar = (Cigar) o;
 
-        if (cigarElements != null ? !cigarElements.equals(cigar.cigarElements) : cigar.cigarElements != null)
-            return false;
+        return cigarElements.equals(cigar.cigarElements);
```
yfarjoun: why did you remove the null protections?

lindenb: cigarElements is never null:

```java
private final List<CigarElement> cigarElements = new ArrayList<CigarElement>();
```

src/java/htsjdk/samtools/Cigar.java
```diff
@@ -254,15 +259,69 @@ public boolean equals(final Object o) {
...
+    }
+    
+    /** build a new cigar from a list of cigar operators */
+    public static Cigar fromCigarOperators(final List<CigarOperator> cigarOperators) {
```
yfarjoun: what is the use-case here?, could you explain in the javadoc?

src/java/htsjdk/samtools/Cigar.java
```diff
@@ -254,15 +259,69 @@ public boolean equals(final Object o) {
...
+    
+    /** shortcut to <code>getCigarElements().iterator()</code> */
+    @Override
+    public Iterator<CigarElement> iterator() {
+        return this.getCigarElements().iterator();
+    }
+    
+    /** returns true if the cigar string contains the given operator */
+    public boolean containsOperator(final CigarOperator operator) {
+        for(final CigarElement cigarElement : this.cigarElements) {
```
yfarjoun: care to java8-ify this function?

cigarElements.stream().anyMatch(el->el.getOperator()==op);

src/java/htsjdk/samtools/Cigar.java
yfarjoun: I see. cigarElements should never be null. OK.

src/java/htsjdk/samtools/CigarOperator.java
```diff
@@ -179,6 +179,31 @@ public static byte enumToCharacter(final CigarOperator e) {
...
+        return this == I || this == D;
+    }
+
+    /** Returns true if the operator is a Skipped Region Insertion or Deletion operator */
+    public boolean isIndelOrSkippedRegion() {
+        return this == N || isIndel();
+    }
+
+    /** Returns true if the operator is a M, a X or a EQ */
+    public boolean isAlign() {
```
yfarjoun: should be a noun, like the others, not a verb. isAlignment ?

src/java/htsjdk/samtools/Cigar.java
```diff
@@ -254,15 +259,69 @@ public boolean equals(final Object o) {
...
-        if (cigarElements != null ? !cigarElements.equals(cigar.cigarElements) : cigar.cigarElements != null)
-            return false;
+        return cigarElements.equals(cigar.cigarElements);
+    }
+    
+    /** build a new cigar from a list of cigar operators */
+    public static Cigar fromCigarOperators(final List<CigarOperator> cigarOperators) {
+        if( cigarOperators == null ) throw new IllegalArgumentException("cigarOperators is null");
+        final List<CigarElement> cigarElementList = new ArrayList<>();
+        int i=0;
```
yfarjoun: spaces in this function

src/java/htsjdk/samtools/Cigar.java
```diff
@@ -254,15 +259,69 @@ public boolean equals(final Object o) {
...
+    }
+    
+    /** returns the first cigar element */
+    public CigarElement getFirstCigarElement() {
+        return isEmpty() ? null : this.cigarElements.get(0); 
+    }
+    
+    /** returns the last cigar element */
+    public CigarElement getLastCigarElement() {
+        return isEmpty() ? null : this.cigarElements.get(this.numCigarElements() -1);
```
yfarjoun: -<space>1

yfarjoun:
minor comments. looks good!

back to you @lindenb

lindenb:
@yfarjoun  I've updated the code + squashed.

by the way : in https://github.com/samtools/htsjdk/blob/master/src/java/htsjdk/samtools/Cigar.java#L192 the old code only looks at 'I' and 'D' but not 'N' as it is declared.

```java
                // There should be an M or P operator between any pair of IDN operators
                if (isInDelOperator(op)) {
(...)

  private static boolean isInDelOperator(final CigarOperator op) {
        return op == CigarOperator.I || op == CigarOperator.D;
    }
```

I kept the old code but I wonder if you shouldn't consider 'op == CigarOperator.N' too ?

yfarjoun:
If the current code doesn't do the right thing (please add a test!) then,
by all means, I'd love if you could add a fix!

can go in a separate PR though.

this one is fine! thanks! :+1:

lindenb:
@yfarjoun ok, I'll do this in another PR.

PR #425, opened 2016-01-06
bug fix https://github.com/samtools/htsjdk/pull/417#issuecomment-1687…

lindenb:
as discussed in https://github.com/samtools/htsjdk/pull/417#issuecomment-168789234

changed 'isInDelOperator'

```
        return op !=null && op.isIndel();
```

to

```
        return op !=null && op.isIndelOrSkippedRegion();
```

because the comment in the code is 

> There should be an M or P operator between any pair of IDN operators

### DIFF
--- a/src/java/htsjdk/samtools/Cigar.java
+++ b/src/java/htsjdk/samtools/Cigar.java
@@ -241,7 +241,7 @@ public class Cigar implements Serializable, Iterable<CigarElement> {
     }
 
     private static boolean isInDelOperator(final CigarOperator op) {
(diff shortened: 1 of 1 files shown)


lindenb:
ok travis failed, because 

```
   [testng] FAILED: testPositive("10M1I5N1I12M")
   [testng] java.lang.AssertionError: expected:<true> but was:<false>
   [testng] 	at htsjdk.samtools.CigarTest.testPositive(CigarTest.java:58)
````

so the original code was valid. My bad. 
Closing this PR.

PR #427, opened 2016-01-06
Factor out CRAM container writing code from CRAMFileWriter.

cmnbroad:
This PR factors out the code (from CRAMFileWriter) that bundles SAMRecords into CRAM containers , and repackages it in a new class CRAMContainerStreamWriter, which exposes an API for finer grained control over the writing of the CRAM file header, SAMFileHeader, and EOF container blocks than is possible with CRAMFileWriter. There is no net  new functionality; CRAMFileWriter delegates to the new class.

### DIFF
--- a/src/java/htsjdk/samtools/CRAMContainerStreamWriter.java
+++ b/src/java/htsjdk/samtools/CRAMContainerStreamWriter.java
@@ -1,0 +1,459 @@
+package htsjdk.samtools;
+
+import htsjdk.samtools.cram.build.ContainerFactory;
+import htsjdk.samtools.cram.build.Cram2SamRecordFactory;
+import htsjdk.samtools.cram.build.CramIO;
+import htsjdk.samtools.cram.build.CramNormalizer;
+import htsjdk.samtools.cram.build.Sam2CramRecordFactory;
+import htsjdk.samtools.cram.common.CramVersions;
+import htsjdk.samtools.cram.common.Version;
+import htsjdk.samtools.cram.lossy.PreservationPolicy;
+import htsjdk.samtools.cram.lossy.QualityScorePreservation;
+import htsjdk.samtools.cram.ref.ReferenceSource;
+import htsjdk.samtools.cram.ref.ReferenceTracks;
+import htsjdk.samtools.cram.structure.Container;
+import htsjdk.samtools.cram.structure.ContainerIO;
+import htsjdk.samtools.cram.structure.CramCompressionRecord;
+import htsjdk.samtools.cram.structure.Slice;
+import htsjdk.samtools.util.Log;
+import htsjdk.samtools.util.RuntimeIOException;
+import htsjdk.samtools.util.StringLineReader;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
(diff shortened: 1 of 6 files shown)


vadimzalunin:
I like the idea of splitting CRAMFileWriter as it is responsible for too many things. Just to understand your approach better, do you intend CRAMContainerStreamWriter to be used outside of CRAMFileWriter?

droazen:
@vadimzalunin Can you review this one when you have time? Thanks!

cmnbroad:
@vadimzalunin Yes, we'd like to extend [Hadoop-BAM](https://github.com/HadoopGenomics/Hadoop-BAM) with support for writing CRAM containers in parallel on multiple Spark endpoints, and then aggregate them together into a single file/stream at the end. These changes enable that.

I've done some testing with using these changes with modified versions of Hadoop-BAM and GATK, but I'd be interested in your input on what kinds of problems we might run into using this approach. Specifically, I'm wondering whether we need to do any fixup at aggregation time, i.e., for container or slice global record counts, crc sums, etc.

src/java/htsjdk/samtools/cram/build/CramIO.java
```diff
@@ -101,6 +101,25 @@ public static long issueEOF(final Version version, final OutputStream outputStre
...
+     * @param cramID
+     * @return the offset in the stream after writing the headers
+     */
+
+    public static long writeHeader(final Version cramVersion, final OutputStream outStream, final SAMFileHeader samFileHeader, String cramID) {
+        final CramHeader cramHeader = new CramHeader(cramVersion, cramID, samFileHeader);
+        try {
+            return CramIO.writeCramHeader(cramHeader, outStream);
+        } catch (final IOException e) {
+            throw new RuntimeException(e);
```
vadimzalunin: perhaps RuntimeIOException?

src/java/htsjdk/samtools/CRAMContainerStreamWriter.java
```diff
@@ -0,0 +1,460 @@
...
+     * Write a CRAM file header and SAM header to the stream.
+
+     * @param textHeader
+     */
+    public void writeHeader(final String textHeader) {
+        // TODO: header must be written exactly once per writer life cycle.
+        final SAMFileHeader header = new SAMTextHeaderCodec().decode(new StringLineReader(textHeader), (cramID != null ? cramID : null));
+        offset = CramIO.writeHeader(cramVersion, outputStream, header, cramID);
+    }
+
```
vadimzalunin: I would also add:
`public void writeHeader(final SAMFileHeader samFileHeader) `

vadimzalunin:
@cmnbroad looks good to me, a couple of minor notes.

src/java/htsjdk/samtools/CRAMContainerStreamWriter.java
cmnbroad: Good idea. Once I do that I no longer need the text version here, so I'll just replace it and let CRAMFileWriter do the conversion since it has to implement the text based once.

src/java/htsjdk/samtools/cram/build/CramIO.java
cmnbroad: Done.

cmnbroad:
@vadimzalunin I made the changes you mentioned and added one more test. If it looks good can you either merge or reassign back to me - I have other changes waiting on these.

PR #428, opened 2016-01-07
XML serialization (marshalling ) of SAM with JAXB

lindenb:
## About JAXB
Adding a few annotations for the XML binding API (JAXB) doesn't change the behavior of htsjdk and add it allows to serialize/marshall a SAM file to XML without writing a specific tool, to use the objects as a source of XSLT transformation (http://blog.bdoughan.com/2012/11/using-jaxb-with-xslt-to-produce-html.html ) . Some 3rd party implementation of JAXB allows to write as JSON (http://stackoverflow.com/questions/19158056).

with the current changes , we can quickly build a (Jersey) REST server writing BAM as XML, JSON, text.... https://jersey.java.net/documentation/1.19/xml.html

## Code change

* I've added some `@XMLElement`, `@XMLAttribute` ... to the fields I want to serialize. I tried to use the names defined by the GA4GH. 
* `@XMLTransient` to fields/methods that should not be serialized to XML (e.g: SamRecord.getValidationStringency() )
* some **empty** **private** constructors (required by JAXB) e.g: 
```java
    /* used for xml serialization */
    @SuppressWarnings("unused")
    private SAMRecord() {
    }
```
* A `SAMXmlWriter implements SAMFileWriter` write a XML file from a SAM.
* A test SAMXmlWriterTest

### Example
here is a file produced by the test.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<sam-file>
  <sam-header group-order="none" sort-order="unsorted" version="1.0">
    <comments/>
    <program-records/>
    <read-groups>
      <read-group>
        <sample>Hi,Mom!</sample>
      </read-group>
    </read-groups>
    <dictionary>
      <Reference index="0" length="101">chr1</Reference>
      <Reference index="1" length="101">chr2</Reference>
      <Reference index="2" length="101">chr3</Reference>
    </dictionary>
  </sam-header>
  <reads>
    <readAlignment>
      <position>1</position>
      <attributes>
        <attribute>RG:Z:0</attribute>
      </attributes>
      <alignedQuality>)'.*.+2,))</alignedQuality>
      <cigar>
        <cigarUnit operationLength="10" operation="M"/>
      </cigar>
      <flags>73</flags>
      <fragmentLength>0</fragmentLength>
      <mappingQuality>255</mappingQuality>
      <mate-position>0</mate-position>
      <mate-fragmentName>*</mate-fragmentName>
      <name>A</name>
      <alignedSequence>CAACAGAAGC</alignedSequence>
      <fragmentName>chr2</fragmentName>
    </readAlignment>
(...)
  </reads>
</sam-file>
```

## Misc

* I've not considered XML->object (unmarshalling) because there are too many non-empty constructors.
* if the PR is accepted, I'll do the same for VCF

### DIFF
--- a/src/java/htsjdk/samtools/Cigar.java
+++ b/src/java/htsjdk/samtools/Cigar.java
@@ -29,6 +29,11 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlElementWrapper;
+import javax.xml.bind.annotation.XmlRootElement;
+import javax.xml.bind.annotation.XmlTransient;
+
 /**
  * A list of CigarElements, which describes how a read aligns with the reference.
  * E.g. the Cigar string 10M1D25M means
@@ -38,6 +43,7 @@ import java.util.List;
  *
  * c.f. http://samtools.sourceforge.net/SAM1.pdf for complete CIGAR specification.
  */
+@XmlRootElement(name="cigar")
 public class Cigar implements Serializable, Iterable<CigarElement> {
     public static final long serialVersionUID = 1L;
 
@@ -50,6 +56,7 @@ public class Cigar implements Serializable, Iterable<CigarElement> {
         this.cigarElements.addAll(cigarElements);
     }
 
+    @XmlElement(name="element")
     public List<CigarElement> getCigarElements() {
         return Collections.unmodifiableList(cigarElements);
     }
@@ -66,6 +73,7 @@ public class Cigar implements Serializable, Iterable<CigarElement> {
         return cigarElements.size();
     }
 
+    @XmlTransient
     public boolean isEmpty() {
         return cigarElements.isEmpty();
     }
@@ -73,6 +81,7 @@ public class Cigar implements Serializable, Iterable<CigarElement> {
     /**
      * @return The number of reference bases that the read covers, excluding padding.
      */
+    @XmlTransient
     public int getReferenceLength() {
         int length = 0;
         for (final CigarElement element : cigarElements) {
@@ -114,6 +123,7 @@ public class Cigar implements Serializable, Iterable<CigarElement> {
     /**
      * @return The number of read bases that the read covers.
      */
+    @XmlTransient
     public int getReadLength() {
         return getReadLength(cigarElements);
     }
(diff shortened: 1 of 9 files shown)


tfenne:
My initial reaction to this is, why?  It's already possible to serve up slices of indexed BAMs over HTTP, and serializing to XML bloats the data several hundred percent from it's binary format, not to mention the overhead of formatting and parsing everything from markup.

lindenb:
@tfenne My aim here is not to convert & produce large XML+sam files. I agree you: this would be stupid :-) 

However, imagine a **web-based genome browser** using a XML/JSON REST server as a source of data : there is no need to write/include a javascript parser for BAM/SAM on the client side. Standard javascript API (json/xml) will do the job.

Web services (**SOAP**): we can use wsgen ( https://docs.oracle.com/javase/6/docs/technotes/tools/share/wsgen.html ) to produce a WSDL+XSD (web service directory + xml schema) service. A java client doesn't need htsjdk, it only needs to run **wsimport** to call some remote methods and parse the basic structure of a read.

```java
@WebMethod(operationName="getReadsInRegion")
public List<SAMRecord> getReadsInRegion(
   @WebParam(name="contig")String contig,
   @WebParam(name="start")int start,
   @WebParam(name="end")int end
   ) throws Exception
```

Associating a **XSLT** stylesheet at the top of the XML could also be used to transform this XML to another format (SVG view of the region, statistics of the region,... etc... ).

For another project I had to write a plain XML serializer for sam (using a bunch of startElement(...), endElement(...) but I'm reinventing the wheel instead of taking advantage of a few java annotations.

I do understand it sounds like an non  necessary feature,  and I won't fight if the administrators doesn't think it's useful :-)

vadimzalunin:
If it's cheap and easy to maintain (which it seems) I'd vote yes.

lindenb:
@tfenne i you feel more comfortable with this: I can remove the new class SAmXMLwriter and move the serialization test into another existing class = no new classes. But I think it would be nice to keep those stdandard annotations.

lbergelson:
I'm not a fan of having these additional tags annotating all the core classes.  I

t seems kind of ugly and confusing (it's not immediately apparent to me which methods are `@XmlTransient`, which are `@XmlValue`, `@XmlElement` etc...)

I understand your pain of having to write a custom serializer when it could be auto-generated though.  I see how this could be very convenient for the application you're talking about.  Is there a way that these could be exposed through an adapter or subclass instead?  That way they could be kept out of the core classes but still conveniently available?

lindenb:
> It seems kind of ugly

and I agree with you.

> and confusing

... if you don't know JAXB ... :-) on my side Java 8 Stream/lambda still confuse me :-)

> . Is there a way that these could be exposed through an adapter or subclass instead?

if I write an adaptater class or an **XmlAdapter** ( http://blog.bdoughan.com/2010/07/xmladapter-jaxbs-secret-weapon.html ), I think we will lose all the interest of writing those annotations in the base classes of htsjdk and I can write this in my private code.

FYI some precisions about the JAXB annotations :-)

  * `@XmlTransient` : -> don't serialize to XML / ignore ( used for example in getProperPairFlag() because an exception would be raised by htsjdk for the unpaired reads )
  * `@XMLValue` -> the class will be serialized as a literal element . Use this field as the text Node `<samtag att1="123">TEXTNODE</samtag>`
  * `@XMLElement`: -> use this method as the name of the XML tag : `<cigar>100M</cigar>`
  * `@XMLRootElement` -> name of the top root xml element when an instance is the saved to a XML element : `<?xml version="1.0"?><samRecord>(....)`

 
Again, if you don't feel comfortable with this PR,  feel free to close it

lindenb:
ok, I'll close that one and handle the code on my side :-)

lbergelson:
@lindenb  We had some debate going on about this locally as you could probably infer from the very slow responses.  I think it's probably best to handle on your side as you've decided.  Thank you for the submission though.

PR #431, opened 2016-01-12
adding a method to convert a FASTQ header (read name) to a read name

nh13:
string usable for output in a SAM or BAM file.

[Review on Reviewable](https://reviewable.io/reviews/samtools/htsjdk/431)

### DIFF
--- a/src/java/htsjdk/samtools/util/SequenceUtil.java
+++ b/src/java/htsjdk/samtools/util/SequenceUtil.java
@@ -1089,4 +1089,27 @@ public class SequenceUtil {
 
         return sofar;
     }
+
+    /**
+     * Returns a read name from a FASTQ header string suitable for use in a SAM/BAM file.  Any letters after the first space are ignored.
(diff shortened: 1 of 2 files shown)


nh13:
@yfarjoun, @tfenne, @droazen  could you take a look?  I found this method useful in a few different tools outside of htsjdk, so I decided to add it in.

tfenne:
Looking at the tests I was surprised by this one:
`{"Name/1", true, false, "Name/1"}`

Wouldn't you want to always remove any trailing `/1` or `/2` from paired reads?  Otherwise the template names won't match in the resulting SAM file?  

I'd also be tempted to change it to take a single parameter `stripMateNumber` instead of the two booleans, and let the user decide how to set it.

tfenne:
@nh13 I think this PR is waiting on you to reply to my comment/question, and then rebase.

akiezun:
@tfenne should we reassign to @nh13 then (if it's waiting on @nh13 )?

nh13:
@tfenne rebased and simplified based on your suggestions.  Thanks for the review.

Issue #404, opened 2015-12-04
CRAMIterator fails on queryname sorted input with unmapped reads


cmnbroad:
CRAMWriter can generate a CRAM file with queryname sorted reads, but when reading the file get an exception when iterating on the first record that is unmappped : 

htsjdk.samtools.util.RuntimeEOFException: java.lang.RuntimeException: Slice mapped outside of the reference.

	at htsjdk.samtools.CRAMIterator.hasNext(CRAMIterator.java:264)
	at htsjdk.samtools.SamReader$AssertingIterator.hasNext(SamReader.java:568)
	at org.broadinstitute.hellbender.tools.picard.sam.SortSam.doWork(SortSam.java:50)

cmnbroad:
@vadimzalunin Can you take a look ?

cmnbroad:
The stack for the original exception is:

Caused by: java.lang.RuntimeException: Slice mapped outside of the reference.
	at htsjdk.samtools.cram.structure.Slice.alignmentBordersSanityCheck(Slice.java:74)
	at htsjdk.samtools.cram.structure.Slice.validateRefMD5(Slice.java:84)
	at htsjdk.samtools.CRAMIterator.nextContainer(CRAMIterator.java:179)
	at htsjdk.samtools.CRAMIterator.hasNext(CRAMIterator.java:257)

vadimzalunin:
@cmnbroad 
I can't reproduce this, htsjdk seems to read the file just fine: 
https://github.com/broadinstitute/gatk/blob/1ecb871ce375799f9245bf932ed5da40f045d7ec/src%2Ftest%2Fresources%2Forg%2Fbroadinstitute%2Fhellbender%2Ftools%2Fcount_reads.cram

Also tried enforcing unmapped flags on some random reads - no exception. 

Am I missing some test condition?

cmnbroad:
Interesting. I should be able to get you a reproducible test case.

cmnbroad:
I pushed a [commit](https://github.com/cmnbroad/htsjdk/commit/62c37af07b1d1322533197e81e0ad67df2a191a1) to a [branch](https://github.com/cmnbroad/htsjdk/commits/cn_cram_sort) on my [fork](https://github.com/cmnbroad/htsjdk) with a test case that repros this. The test code is in SAMFileReaderTest.java.

[...]

PR #355, opened 2015-10-22
improving CRAM compression

vadimzalunin:
the aim is to shift from custom encodings onto external like GZIP and RANS. This allows for better and faster compression in most cases.

PR #432, opened 2016-01-13
Cram multiref roundtrip fixes

vadimzalunin:
two goals:
1) properly support multiref https://github.com/samtools/htsjdk/issues/404
2) compression improvements (re-make of https://github.com/samtools/htsjdk/pull/355)


src/java/htsjdk/samtools/CRAMFileWriter.java
```diff
@@ -46,8 +46,9 @@
 
 @SuppressWarnings("UnusedDeclaration")
 public class CRAMFileWriter extends SAMFileWriterImpl {
-    private static final int REF_SEQ_INDEX_NOT_INITIALIZED = -2;
+    private static final int REF_SEQ_INDEX_NOT_INITIALIZED = -3;
```
cmnbroad: I'm assuming that this change was necessary to avoid collision with the new Slice values; it looks like the refSeqIndex variable in this class now takes on values drawn from 3 different namespaces (SAMRecord, Slice and CRAMFIleWriter) that all have to be mutually exclusive. I think it would be much cleaner and safer to define the sentinel values together someplace in an enum and use a variable of that type here that is separate from the int-valued reference index.

src/java/htsjdk/samtools/CRAMFileWriter.java
```diff
@@ -46,8 +46,9 @@
...
     static int DEFAULT_RECORDS_PER_SLICE = 10000;
+    static int SWITCH_TO_MULTIREF_IF_MORE_THAN = 1000;
```
cmnbroad: This could use a better name, maybe something like MAX_SINGLE_REF_RECORDS.

src/java/htsjdk/samtools/CRAMIndexer.java
```diff
@@ -101,16 +105,72 @@ public CRAMIndexer(final OutputStream output, final SAMFileHeader fileHeader) {
...
+     * reads mapped to a single reference.
      * If this alignment starts a new reference, write out the old reference.
-     * Requires a non-null value for rec.getFileSource().
      *
-     * @param slice The CRAM slice
+     * @param slice The CRAM slice, single ref or unmapped only.
+     * @throws htsjdk.samtools.SAMException if slice refers to multiple reference sequences.
      */
-    public void processAlignment(final Slice slice) {
+    public void processSingleReferenceSlice(final Slice slice) {
```
cmnbroad: Much better name!

src/java/htsjdk/samtools/CRAMIndexer.java
```diff
@@ -101,16 +105,72 @@ public CRAMIndexer(final OutputStream output, final SAMFileHeader fileHeader) {
...
+        System.out.println(slice.toString());
```
cmnbroad: This should be removed.

src/java/htsjdk/samtools/CRAMIndexer.java
```diff
@@ -350,28 +412,30 @@ void startNewReference() {
...
 
         int totalRecords = 0;
         Container container = null;
         do {
-            if (++totalRecords % 10 == 0)
-                if (null != log) log.info(totalRecords + " slices processed ...");
+            if (++totalRecords % 10 == 0) {
+                if (null != log) {
+                    log.info(totalRecords + " slices processed ...");
+                }
```
cmnbroad: Use a [SAMProgressLogger](https://github.com/samtools/htsjdk/blob/e64903b5e977a6ee88748ee71581db341758d457/src/java/htsjdk/samtools/util/ProgressLogger.java#L20) for this.

src/java/htsjdk/samtools/CRAMIterator.java
```diff
@@ -165,6 +165,7 @@ private void nextContainer() throws IOException, IllegalArgumentException,
 
         if (container.sequenceId == SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX) {
             refs = new byte[]{};
+            prevSeqId = -1;
         } else if (container.sequenceId == -2) {
             refs = null;
             prevSeqId = -2;
```
cmnbroad: We should replace magic numbers with constants. In this case its hard to tell what the value -2 means, or what namespace the sequenceID is drawn from.

src/java/htsjdk/samtools/cram/build/CompressionHeaderFactory.java
```diff
@@ -60,745 +47,543 @@
...
+    private final Map<Integer, EncodingDetails> bestEncodings = new HashMap<>();
+    private final ByteArrayOutputStream baosForTagValues;
 
-        { // compression bit flags encoding:
-            getOptimalIntegerEncoding(header, EncodingKey.CF_CompressionBitFlags, 0, records);
-        }
-
-        { // ref id:
+    public CompressionHeaderFactory() {
+        baosForTagValues = new ByteArrayOutputStream(1024 * 1024);
```
cmnbroad: Magic number -> symbolic constant.

src/tests/java/htsjdk/samtools/CramFileWriterTest.java
```diff
@@ -231,4 +231,24 @@ public void test_roundtrip_tlen_preserved() throws IOException {
...
+            }
+        }
+
+        try (final SamReader outReader = SamReaderFactory.makeDefault().referenceSequence(reference).open(outputFile)) {
+            for (final SAMRecord rec : outReader) {
+            }
+        }
+
+    }
+
```
cmnbroad: This looks like the test case for writing and then reading back in a query-sorted cram. Now that it doesn't throw it really should have an assert to verify the result is correct. Also we please rename the test file to something relevant to this context, maybe "querysorted.cram".

cmnbroad: Also, please align the names of the .fasta and .fai files to match whatever name you use for the .cram.

src/java/htsjdk/samtools/cram/encoding/huffman/HuffmanParamsCalculator.java
```diff
@@ -0,0 +1,135 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+
+/**
+ * A utility class to calculate Huffman encoding parameters based on the values to be encoded.
+ */
+public class HuffmanParamsCalculator {
```
cmnbroad: Does this class need to be public, or can it have reduced scope ?

src/java/htsjdk/samtools/cram/encoding/reader/RefSeqIdReader.java
```diff
@@ -0,0 +1,237 @@
...
+import java.util.Map;
+
+/**
+ * A reader that only keeps track of alignment spans. The intended use is for
+ * CRAI index.
+ * 
+ * @author vadim
+ *
+ */
+public class RefSeqIdReader extends AbstractReader {
```
cmnbroad: Does this class need to be public ?

src/java/htsjdk/samtools/cram/encoding/reader/RefSeqIdReader.java
```diff
@@ -0,0 +1,237 @@
...
+
+			final Integer tagIdList = tagIdListCodec.readData();
+			final byte[][] ids = tagIdDictionary[tagIdList];
+			if (ids.length > 0) {
+				final int tagCount = ids.length;
+				cramRecord.tags = new ReadTag[tagCount];
+				for (int i = 0; i < ids.length; i++) {
+					final int id = ReadTag.name3BytesToInt(ids[i]);
+					final DataReader<byte[]> dataReader = tagValueCodecs.get(id);
+					final ReadTag tag = new ReadTag(id, dataReader.readData(), ValidationStringency.SILENT);
```
cmnbroad: Why is the validation stringency hardcoded to SILENT ?

src/java/htsjdk/samtools/CRAMFileWriter.java
vadimzalunin: @cmnbroad Ideally I'd like SAMRecord.MULTIPLE_REFERENCES_INDEX=-2 and then Slice and others could rely on SAMRecord's ref id space. Do you think this is possible?

src/java/htsjdk/samtools/cram/encoding/huffman/HuffmanParamsCalculator.java
vadimzalunin: @cmnbroad not sure if the class is needed at all. This is intended for advanced users who choose to override default compression but it seems marginally useful. I'll remove the class and the tests.

src/java/htsjdk/samtools/cram/encoding/reader/RefSeqIdReader.java
vadimzalunin: @cmnbroad Good question. All readers reside in htsjdk.samtools.cram.encoding.reader package now and a few more are to be transferred from cramtools. I could limit their visibility via some public factory but is it worth it?

src/java/htsjdk/samtools/CRAMFileWriter.java
cmnbroad: Having looked at the code and spec a bit more closely, and it looks like some of the constants used in Slice are really just alternate names for, and must have the same value as, names already defined in SAMRecord i.e., Slice.UNMAPPED_OR_NO_REFERENCE is really just an alias for SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX and Slice.NO_ALIGNMENT_START is really just an alias for SAMRecord.NO_ALIGNMENT_START. Is that correct ? It it is, can we remove these symbols entirely and just use the SAMRecord constants, or at the very least initialize them in Slice to use the values in SAMRecord to make the relationship explicit?

Maybe others have a different opinion., but for Slice.MULTI_REFERENCE, it would seem strange to me to expose a value through the public interface of SAMRecord that is never used by SAMREcord and is only used by the CRAM code. Since the specific value -2 is prescribed by the CRAM spec, my take would be that it should live in the CRAM package somewhere.

src/java/htsjdk/samtools/CRAMFileWriter.java
cmnbroad: Its kind of hard to tell what the significance of this number is. It looks like it somehow affects alignment span in downstream code, but it would be helpful to have a comment somewhere describing why we would want to force a switch to multi-ref in the case where there are > 1000 records, even if they have the same same reference index (if I have that right).

src/java/htsjdk/samtools/CRAMFileWriter.java
```diff
@@ -198,11 +222,21 @@ protected void flushContainer() throws IllegalArgumentException, IllegalAccessEx
...
-            refs = source.getReferenceBases(sequence, true);
-            refSeqName = sequence.getSequenceName();
+        switch (refSeqIndex) {
+            case Slice.MULTI_REFERENCE:
+                if (preservation != null && preservation.areReferenceTracksRequired()) {
+                    throw new SAMException("Cannot apply reference-based lossy compression on non-coordinate sorted reads.");
+                }
+                refs = new byte[0];
+                break;
+            case Slice.UNMAPPED_OR_NO_REFERENCE:
```
cmnbroad: Per above, can we replace this with SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX ?

src/java/htsjdk/samtools/CRAMIndexer.java
```diff
@@ -101,16 +105,72 @@ public CRAMIndexer(final OutputStream output, final SAMFileHeader fileHeader) {
...
+                } else {
+                    processSingleReferenceSlice(slice);
+                }
+            }
+
+        } catch (final IOException e) {
+            throw new RuntimeIOException("Failed to read cram container", e);
+        } catch (final IllegalAccessException e) {
+            throw new SAMException("Failed to read cram container", e);
+        }
```
cmnbroad: It looks like these exceptions (i.e. the IllegalAccessException) originate deep down at the encoding level, and are propagated up through several layers as checked exceptions, but they aren't really handled anywhere. I think it would be a lot more useful to catch these where they originate (DataReader/Factory ??) and translate them into unchecked exceptions with a more meaningful message saying what went wrong. Then all the other layers wouldn't have to declare and catch them, since they can't do anything with them anyway.

src/java/htsjdk/samtools/cram/build/CompressionHeaderFactory.java
```diff
@@ -60,745 +47,543 @@
...
+    /**
+     * Build an array of substitution frequencies for the given CRAM records.
+     *
+     * @param records
+     *            CRAM records to scan
+     * @return a 2D array of frequencies, see
+     *         {@link htsjdk.samtools.cram.structure.SubstitutionMatrix}
+     */
+    static long[][] buildFrequencies(final List<CramCompressionRecord> records) {
+        final long[][] frequencies = new long[200][200];
```
cmnbroad: This array is indexed with byte values - what is significant about 200 ??

src/java/htsjdk/samtools/CRAMFileWriter.java
vadimzalunin: this is to prevent many tiny containers, like a threshold when to switch to multiref rather than start another container. I'm changing the constant name and the code to reflect this better.

src/java/htsjdk/samtools/cram/build/CompressionHeaderFactory.java
```diff
@@ -60,745 +47,543 @@
...
-        { // next fragment alignment start
-            header.encodingMap.put(EncodingKey.NP_NextFragmentAlignmentStart, ExternalIntegerEncoding.toParam(mateInfoID));
+    /**
+     * Find a byte value never mentioned in the array
+     * @param array bytes
+     * @return byte value or -1 if the array contains all possible byte values.
+     */
+    static int getUnusedByte(final byte[] array) {
+        final byte[] usage = new byte[256];
+        for (final byte b : array) {
```
cmnbroad: The size of the usage array should really be based on the length if the input array. If 256 is really required, we should probably check to ensure the input array is that size.

src/java/htsjdk/samtools/cram/build/ContainerFactory.java
```diff
@@ -192,8 +179,15 @@ private static Slice buildSlice(final List<CramCompressionRecord> records,
         int prevAlStart = slice.alignmentStart;
         for (final CramCompressionRecord record : records) {
             record.alignmentDelta = record.alignmentStart - prevAlStart;
+            if (record.alignmentDelta <0) {
+                System.out.println("hgotcha");
```
cmnbroad: Not sure if this needs to be handled here but we should get rid of the println.

src/java/htsjdk/samtools/cram/build/ContainerFactory.java
```diff
@@ -192,8 +179,15 @@ private static Slice buildSlice(final List<CramCompressionRecord> records,
...
+            }
             prevAlStart = record.alignmentStart;
-            writer.write(record);
+            try {
+                writer.write(record);
+            } catch (ArrayIndexOutOfBoundsException e) {
+                e.printStackTrace();
+            }
```
cmnbroad: We should remove this, or at the very least we shouldn't catch without re-throwing.

src/java/htsjdk/samtools/cram/build/ContainerParser.java
```diff
@@ -51,14 +54,16 @@ public ContainerParser(final SAMFileHeader samFileHeader) {
...
+        if (records == null) {
             records = new ArrayList<CramCompressionRecord>(container.nofRecords);
+        }
 
-        for (final Slice slice : container.slices)
+        for (final Slice slice : container.slices) {
             records.addAll(getRecords(slice, container.header, validationStringency));
+        }
 
         final long time2 = System.nanoTime();
```
cmnbroad: Can we remove this instrumentation (maybe in another PR) ? There is quite a bit of it in the cram code.

src/java/htsjdk/samtools/cram/encoding/reader/DataReaderFactory.java
```diff
@@ -39,6 +40,7 @@
 
 @SuppressWarnings("unchecked")
```
cmnbroad: I dont think we need this ?

src/java/htsjdk/samtools/cram/build/CompressionHeaderFactory.java
```diff
@@ -60,745 +47,543 @@
...
+     *            the matrix to be updated
+     */
+    static void updateSubstitutionCodes(final List<CramCompressionRecord> records,
+                                                final SubstitutionMatrix substitutionMatrix) {
+        for (final CramCompressionRecord record : records) {
+            if (record.readFeatures != null) {
+                for (final ReadFeature recordFeature : record.readFeatures) {
+                    if (recordFeature.getOperator() == Substitution.operator) {
+                        final Substitution substitution = ((Substitution) recordFeature);
+                        if (substitution.getCode() == -1) {
```
cmnbroad: Should use a symbolic constant here.

src/java/htsjdk/samtools/cram/build/ContainerParser.java
```diff
@@ -144,14 +204,15 @@ public ContainerParser(final SAMFileHeader samFileHeader) {
             if (!nanosecondsMap.containsKey(key)) {
                 nanosecondsMap.put(key, 0L);
                 value = 0;
-            } else
+            } else {
                 value = nanosecondsMap.get(key);
+            }
             nanosecondsMap.put(key, value + statMap.get(key).nanos);
         }
```
cmnbroad: Is this code only used for perf analysis/instrumentation ? Can we remove it - it looks like some of the supporting code is exposed through the public API (i.e, Container.parseTime, buildHeaderTime, etc) ?

src/java/htsjdk/samtools/cram/encoding/reader/RefSeqIdReader.java
cmnbroad: Either that or change the package structure, but its probably not worth complicating the code by doing that now.

src/java/htsjdk/samtools/cram/encoding/reader/RefSeqIdReader.java
```diff
@@ -0,0 +1,237 @@
...
+	public Map<Integer, AlignmentSpan> getReferenceSpans() {
+		return spans;
+	}
+
+	public void read() {
+		cramRecord.sequenceId = globalReferenceSequenceId;
+		try {
+			cramRecord.flags = bitFlagsCodec.readData();
+			cramRecord.compressionFlags = compressionBitFlagsCodec.readData();
+			if (refId == -2)
```
cmnbroad: -2 -> constant.

src/java/htsjdk/samtools/cram/encoding/reader/RefSeqIdReader.java
```diff
@@ -0,0 +1,237 @@
...
+					}
+				}
+			}
+
+			recordCounter++;
+
+		} catch (final Exception e) {
+			System.err.printf("Failed at record %d. \n", recordCounter);
+			throw new RuntimeException(e);
+		}
```
cmnbroad: I'm not sure why we need such wide exception handling here (if there is a specific case that can happen, maybe a narrower exception would be good) but we should at least change the printf.

src/java/htsjdk/samtools/cram/structure/Slice.java
```diff
@@ -66,6 +68,7 @@
     public SAMBinaryTagAndValue sliceTags;
 
     private void alignmentBordersSanityCheck(final byte[] ref) {
+        if (sequenceId == UNMAPPED_OR_NO_REFERENCE) return ;
         if (alignmentStart > 0 && sequenceId >= 0 && ref == null) throw new NullPointerException("Mapped slice reference is null.");
```
cmnbroad: This seems like it should be either IllegalArgumentException or IllegalStateException.

src/tests/java/htsjdk/samtools/CRAMIndexerTest.java
```diff
@@ -0,0 +1,70 @@
...
+        record.alignmentStart =start;
+        record.readBases = record.qualityScores = bases;
+        record.readName = Integer.toString(recordIndex);
+        record.readLength = readLength;
+        record.readFeatures = Collections.emptyList();
+
+        return record;
+    }
+    @Test
+    public void test_processMultiContainer() throws IOException, IllegalAccessException {
```
cmnbroad: There is really only one test case in this file I think (a single multi-ref container). I think we should add index tests for multiple containers and single reference slices as well, unless they exist somewhere else.

cmnbroad: Also, since this is a new file, can we rename it to something that reflects that its BAI index oriented, i.e., CRAMBAIIndexerTest or something like that.

cmnbroad:
Back to @vadimzalunin.

src/java/htsjdk/samtools/cram/structure/AlignmentSpan.java
```diff
@@ -0,0 +1,92 @@
+package htsjdk.samtools.cram.structure;
+
+import htsjdk.samtools.SAMRecord;
+
+/**
+ * A span of reads on a single reference.
+ */
+public class AlignmentSpan {
```
vadimzalunin: the class is used in more than one package, for example:
htsjdk.samtools.cram.encoding.reader.RefSeqIdReader
htsjdk.samtools.CRAMIndexer

Not sure how to refactor this without 'public' modifier. I guess htsjdk.samtools.util.Interval could be used instead but it lacks record count.

vadimzalunin:
@cmnbroad mostly fixed, not sure what to do about public classes like AlignmentSpan and RefSeqIdReader - they are accessed in more than one package. Separate PR?

Also some unrelated improvements like removing performance stats I would like to leave for another PR - this one is already getting out of hands.

cmnbroad:
@vadimzalunin Yes - lets not address the class visibility issues in this PR.  I do want to make another pass through the recent changes - I'll be able to do that in the next day or so.

src/java/htsjdk/samtools/CRAMIndexer.java
```diff
@@ -187,7 +262,7 @@ private int computeIndexingBin(final Slice slice) {
         /**
          * Record any index information for a given BAM record
          *
-         * @param slice The BAM record. Requires rec.getFileSource() is non-null.
+         * @param slice CRAM slice, single ref or unmapped only.
          */
         public void processAlignment(final Slice slice) {
```
cmnbroad: The javadoc param has been changed here, but its still confusing that this is called processAlignment when it takes a slice.

src/java/htsjdk/samtools/cram/encoding/reader/RefSeqIdReader.java
```diff
@@ -0,0 +1,240 @@
...
+
+					if (cramRecord.isForcePreserveQualityScores()) {
+						cramRecord.qualityScores = qualityScoresCodec.readDataArray(cramRecord.readLength);
+					}
+				}
+			}
+
+			recordCounter++;
+
+		} catch (final Exception e) {
```
cmnbroad: I think this can be reduced to IOException.

cmnbroad:
@vadimzalunin A couple of more minor comments. Also, I'm assuming there are no backward compatibility issues with these compression changes, but can you just confirm that ? I'll make tickets for the remaining issues I think we should do in other PRs (i.e., removing the instrumentation). Can you squash these down to one commit, and rebase?

vadimzalunin:
@cmnbroad minor comments fixed, rebased/squashed.
This PR should be fully compatible with prev versions.

PR #434, opened 2016-01-13
skip failing SRA test native code is unavailable

lbergelson:
adding code to skip an SRA test if the SRA native code is unavailable.


droazen:
:+1:, but please:

- Look for any other SRA tests that don't get skipped automatically when the library is not present, and add the same line there

- Create a ticket to bump the ancient version of TestNG used by htsjdk, and use `SkipException` in these tests once that's done.

PR #438, opened 2016-01-17
modify SeekableFileStream

zong0jie:
I tried jsr203-hadoop (@damiencarol ) and found it is a good solution for hdfs read and write. @tomwhite do good works on "IndexedFastaSequenceFile" and I think maybe I can help to do some thing else.
So I modified SeekableFileStream so that it can also use jsr203-hadoop.
Could any one let me know whether the work is useful, or maybe you have any other plan.

### DIFF
--- a/src/java/htsjdk/samtools/seekablestream/SeekableFileStream.java
+++ b/src/java/htsjdk/samtools/seekablestream/SeekableFileStream.java
@@ -17,10 +17,15 @@
  */
 package htsjdk.samtools.seekablestream;
 
+import htsjdk.samtools.util.IOUtil;
+
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
-import java.io.RandomAccessFile;
+import java.nio.ByteBuffer;
+import java.nio.channels.SeekableByteChannel;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


zong0jie:
I find here may have a problem. Since we use nio Path instead of File for hdfs, that means all the functional method of File, such as File#list(), File#length() ... will be invalide, because hdfs file cannot be recognized by File. What I think is to rewrite all these method in IOUTil.
I don't know whether someone may have good idea.

zong0jie:
maybe it isn't in your plan. close it.

droazen:
@zong0jie On the contrary, expanding use of `java.nio.file.Path` in htsjdk to enable hdfs support via jsr203-hadoop is definitely useful, and something that we are interested in doing. If you'd like us to consider this branch, feel free to re-open and I'll ask @tomwhite to have a look.

zong0jie:
@dorazen Thank you for your response and I'm apologize for my impatience.

droazen:
@tomwhite Could you look at this when you have a chance?

tomwhite:
Moving to support `java.nio.file.Path` in htsjdk is definitely the right direction in my opinion.

Rather than having special semantics for `File` (i.e. an `hdfs://` name), I think it would be better to have a `SeekablePathStream`, that is like `SeekableFileStream`, but for `Path`, and then overload methods and constructors in htsjdk as needed so they take `Path` objects. What do you think @zong0jie?

zong0jie:
@droazen @tomwhite I have another opinion, you can deside which is better.
 I agree with  not using special semantics for File (i.e. an hdfs:// name). 

However, sicne ```SeekablePathStream``` can do everything of ```SeekableFileStream```, I prefer overload constructor method like ```SeekableFileStream(Path path)``` and may deprecate ```SeekableFileStream(File file)``` just like @tomwhite did in ```IndexedFastaSequenceFile```. 

In this way the test unit can work with out change and ```SeekableStreamFactory``` also needs very little changes. Of cause the method I add in ```IOUtil``` should be deleted.

tomwhite:
@zong0jie thanks for your comment, sorry I missed it until today. I wrote an implementation of `SeekablePathStream` here https://github.com/samtools/htsjdk/compare/master...tomwhite:seekable_path_stream. Does this look OK?

I don't think we should deprecate the File-based methods since they are still valid, and widely used. (Note that I didn't deprecate anything in #308.)

@droazen I didn't add an overloaded `open` method to `SamReaderFactory` that takes a `Path` since it would be an incompatible change, and it's not needed since you can use `SamInputResource.of(path).index(indexPath)` instead.

droazen:
@tomwhite How would adding an overload for `SamReaderFactory.open()` that takes a `Path` break backwards compatibility? The existing overloads wouldn't change? I still think that's a better solution, as we want `Path` to be a first-class citizen in the factory interface, just like it is now in `ReferenceSequenceFileFactory` thanks to your earlier work. It would be inconsistent for `ReferenceSequenceFileFactory` to take `Path` objects directly but for `SamReaderFactory` to take them only indirectly.

droazen:
(and if you're concerned that adding a new abstract method to `SamReaderFactory` that takes a `Path` might break implementations other than `SamReaderFactoryImpl`, why not make it non-abstract, have it do the `SamInputResource.of(path).index(indexPath)` dancing internally, and call the existing `open(SamInputResource)`?)

zong0jie:
@tomwhite seems you provide a better solution again. If you make the change more early I may not make this PR :). @droazen maybe I should close it this time.

tomwhite:
@droazen Good point about making it non-abstract. Thanks for the comment @zong0jie. I've opened #458.

zong0jie:
ok, close it.

PR #439, opened 2016-01-18
Implementation of a phred/bwa-style quality trimming utility method (…

tfenne:
…and tests).

### DIFF
--- a/src/java/htsjdk/samtools/util/TrimmingUtil.java
+++ b/src/java/htsjdk/samtools/util/TrimmingUtil.java
@@ -1,0 +1,67 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Tim Fennell
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


tfenne:
@yfarjoun @jacarey I'd like to submit a PR to Picard based on this change, that adds optional quality trimming to `SamToFastq`.  Would you mind taking a quick look at this?  It's a re-coding into legible java of the same algorithm used in `bwa aln`.

src/java/htsjdk/samtools/util/TrimmingUtil.java
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * If the entire read is of low quality this function may return 0!  It is left to the caller
+     * to decide whether or not to trim reads down to 0-bases, or to enforce some minimum length.
+     *
+     * @param quals a byte[] of quality scores in phred scaling (i.e. integer values between 0 and ~60)
+     * @param trimQual the lowest quality that is considered "good". In the simplest case
+     *                 where a read is composed exclusively of "good" qualities followed by
+     *                 "bad" qualities, this is the lowest quality value left after trimming.
+     * @return The zero-based index of the first base within the quality string that should be trimmed.
+     *         When no trimming is required, quals.length (i.e. an index noe greater than the last valid
```
jacarey: Typo: noe -> no

src/java/htsjdk/samtools/util/TrimmingUtil.java
```diff
@@ -0,0 +1,67 @@
...
+     *         index) will be returned.
+     */
+    public static int findQualityTrimPoint(final byte[] quals, final int trimQual) {
+        final int length = quals.length;
+        int score = 0, maxScore = 0, trimPoint = length;
+        if (trimQual < 1 || length == 0) return 0;
+
+        for (int i=length-1; i >= 0; --i) {
```
jacarey: Formatting.

jacarey:
:+1:

The description doesn't quite capture the case in test case "testLowQualityWithOccasionalHighQuality". Reading it I had assumed it would return the index of the left most score less than the trimming quality. 

Likely my confusion was in not really understanding "first index within the byte array that should be clipped" without actually looking at the algorithm.

I'm not sure there is a succinct way of describing that and maybe "phred-style quality trimming" means more to most people.

tfenne:
Thanks @jacarey - will fix those couple of things and merge.
I agree that the algorithm is a little hard to describe - it's clearly looking to identify a region of "mostly low quality" at the end of the read, and does so in a way that can handle the occasionally high quality base mixed in with a lot of low quality.  The algorithm was apparently pioneered in phred for sanger base calling/scoring where presumably even messy peaks will align once in a while.

PR #440, opened 2016-01-18
Allow the duplicate set comparator to be set.

nh13:
@tfenne @yfarjoun simple change to allow a user-provided comparator, for example, if we wanted to extend `SAMRecordDuplicateComparator` with our own implementation.

### DIFF
--- a/src/java/htsjdk/samtools/DuplicateSetIterator.java
+++ b/src/java/htsjdk/samtools/DuplicateSetIterator.java
@@ -51,12 +51,21 @@ public class DuplicateSetIterator implements CloseableIterator<DuplicateSet> {
         this(iterator, header, false);
     }
 
+    public DuplicateSetIterator(final CloseableIterator<SAMRecord> iterator,
+                                final SAMFileHeader header,
+                                final boolean preSorted) {
+        this(iterator, header, preSorted, new SAMRecordDuplicateComparator(Collections.singletonList(header)));
+    }
+
     /**
      * Allows the user of this iterator to skip the sorting of the input if the input is already sorted.  If the records are said to be
      * sorted but not actually sorted in the correct order, an exception during iteration will be thrown.
      */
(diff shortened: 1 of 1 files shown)


eitanbanks:
@nh13 wouldn't you want to add this to the constructor?  There's already a DuplicateSet being generated there with the default comparator.

nh13:
@eitanbanks I think the proliferation of alternate constructors would start to make this difficult to understand.  If you would prefer, I can move it to more of a builder pattern?   I can have `setComparator` and `setScoringStrategy` return a `DuplicateSetIterator`?   I could go further and replace the constructors to make it truly a builder pattern but that would make it not backwards compatible.  Your call for all of this, I just need to set the comparator for my own purposes.

eitanbanks:
Sorry, I probably wasn't clear.  I just meant that if you don't do something then your results might not be accurate.  The constructor already has this in it:
this.duplicateSet = new DuplicateSet(this.comparator);
So if you wait until after the constructor to set your comparator then you will have missed at least one creation of a duplicate said with the correct comparator.

nh13:
@eitanbanks I suppose then `setScoringStrategy` is also being set after the first duplicate set is populated.  I think that affects Picard's `SimpleMarkDuplicatesWithMateCigar`.  What do you think about making `setScoringStrategy` to deprecated, since now folks can set the scoring strategy on the passed in comparator?

src/java/htsjdk/samtools/DuplicateSetIterator.java
```diff
@@ -85,6 +94,8 @@ public DuplicateSetIterator(final CloseableIterator<SAMRecord> iterator, final S
 
     }
 
+    @Deprecated
+    /** Do not use as the first duplicate set will not be compared with this scoring strategy. */
```
eitanbanks: "Instead, set the scoring strategy on the comparator"

eitanbanks:
Yeah, this looks better

nh13:
@eitanbanks thanks for the review, I will merge once the tests pass.

PR #444, opened 2016-01-20
fixes the case where mates are not clipped if they start at the same

nh13:
position


nh13:
@tfenne little bug fix, added a test to make sure the behavior is as we want.

tfenne:
Looks good to me.  Merge at will once the tests pass.

PR #448, opened 2016-01-22
Escape quotes in VCF header values

cwhelan:
This is a small change that protects against writing unescaped quotes in VCF header lines, which can cause files written with htsjdk to fail validation.

For some attribute values in VCF header lines, htsjdk currently wraps the string value with double quotes. However, if the value passed in itself contains double quotes, this results in unquoted or unbalanced quotes in the resulting VCF header line. An example tool that can cause this issue is GATK VariantFiltration, which creates VCF header lines (of the form "GATKCommandLine=") with broken quotes if run with a JEXL expression that contains double quotes.

The VCF spec states that double quotes can be escaped with a backlash. This change automatically does this when necessary. 

Included a couple of unit tests of the functionality as well as an end-to-end test that proves you can write out and read a file with escaped quotes, and then re-write and read it without runaway escaping.

### DIFF
--- a/src/java/htsjdk/variant/vcf/VCFHeaderLine.java
+++ b/src/java/htsjdk/variant/vcf/VCFHeaderLine.java
@@ -158,9 +158,17 @@ public class VCFHeaderLine implements Comparable, Serializable {
             builder.append("=");
             builder.append(entry.getValue().toString().contains(",") ||
                            entry.getValue().toString().contains(" ") ||
-                           entry.getKey().equals("Description") ? "\""+ entry.getValue() + "\"" : entry.getValue());
+                           entry.getKey().equals("Description") ? "\""+ escapeQuotes(entry.getValue().toString()) + "\"" : entry.getValue());
         }
         builder.append(">");
         return builder.toString();
+    }
+
+    private static String escapeQuotes(final String value) {
+        // java escaping in a string literal makes this harder to read than it should be
+        // without string literal escaping and quoting the regex would be: replaceAll( ([^\])" , $1\" )
+        // ie replace: something that's not a backslash ([^\]) followed by a double quote
+        // with: the thing that wasn't a backslash ($1), followed by a backslash, followed by a double quote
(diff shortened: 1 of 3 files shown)


src/tests/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
```diff
@@ -61,6 +69,19 @@ private VCFHeader createHeader(String headerStr) {
         return header;
     }
 
+    @BeforeClass
+    private void createTemporaryDirectory() {
```
lbergelson: we need to move GATK BaseTest utils down so people don't have to do this anymore.  @droazen any objections?

droazen: Fine as long as in the GATK we still have a method in `BaseTest` that calls into the underlying utility, wherever it may live.

src/java/htsjdk/variant/vcf/VCFHeaderLine.java
```diff
@@ -158,9 +158,13 @@ public static String toStringEncoding(Map<String, ? extends Object> keyValues) {
...
                            entry.getValue().toString().contains(" ") ||
-                           entry.getKey().equals("Description") ? "\""+ entry.getValue() + "\"" : entry.getValue());
+                           entry.getKey().equals("Description") ? "\""+ escapeQuotes(entry.getValue().toString()) + "\"" : entry.getValue());
         }
         builder.append(">");
         return builder.toString();
     }
+
+    private static String escapeQuotes(final String value) {
+        return value.replaceAll("([^\\\\])\"", "$1\\\\\"");
```
lbergelson: Could you add a comment here showing an example of what this replace does?  It's hard for me to read this regex.

cwhelan: Yes, Java regexes are horrible to read. Added explanatory comment.

lbergelson:
:+1:

PR #450, opened 2016-01-23
private members for variant ctx

lindenb:
I've moved a few fields in variant context from protected to private.
Those fields were used in VariantContextBuilder but a 'builder' should use the getter/setter of a variantcontext if we are using a derived class of VariantContext.

Added a few 'final' too.

### DIFF
--- a/src/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
+++ b/src/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
@@ -94,7 +94,7 @@ public class VariantContextBuilder {
      * Create an empty VariantContextBuilder where all values adopt their default values, but the bare min.
      * of info (source, chr, start, stop, and alleles) have been provided to start.
      */
-    public VariantContextBuilder(String source, String contig, long start, long stop, Collection<Allele> alleles) {
+    public VariantContextBuilder(final String source, final String contig, final long start, final long stop, final Collection<Allele> alleles) {
         this.source = source;
         this.contig = contig;
         this.start = start;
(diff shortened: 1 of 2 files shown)


src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1654,7 +1652,7 @@ private final Genotype fullyDecodeGenotypes(final Genotype g, final VCFHeader he
     //
     // ---------------------------------------------------------------------------------------------------------
     @Deprecated
-    public String getChr() {
+    public final String getChr() {
```
yfarjoun: why make this method final?

yfarjoun:
one comment. otherwise :+1: from me.

lindenb:
@yfarjoun final because getChrom is deprecated and calls getContig . You don't want people extending  VariantContext to do things with 'getChrom' , so it  should be final :-)

yfarjoun:
:+1:

lbergelson:
I definitely approve of encapsulating these fields.  This is technically a breaking change though, since someone could be accessing those directly.  I'd wager that very few people are extending `VariantContext` so it's unlikely to cause problems.  If we merge this we should probably bump our version to 3.0.0.  A test of our semantic versioning resolve...

lindenb:
@lbergelson I've moved VariantContext.java back to its original state and I kept the getter in VariantContextBuilder, so it shouldn't affect any sub-class of VariantContext.java

lbergelson:
:+1: Although I do still agree with making those variables private at some point.  There's no reason they should have been exposed in the first place.

PR #451, opened 2016-01-23
VCFEncoder : vcf line capacity

lindenb:
in VCFEncoder Added a new private variable 

```
	private int vcfLineLengthCapacity = 100;
```

used to set the stringBuilder capacity. It could improve things for large VCF lines (Exac.... )

```
final StringBuilder stringBuilder = new StringBuilder(this.vcfLineLengthCapacity);
(...)
this.vcfLineLengthCapacity = Math.max(this.vcfLineLengthCapacity, stringBuilder.length());
```

### DIFF
--- a/src/java/htsjdk/variant/vcf/VCFEncoder.java
+++ b/src/java/htsjdk/variant/vcf/VCFEncoder.java
@@ -37,6 +37,9 @@ public class VCFEncoder {
 
 	private boolean outputTrailingFormatFields = false;
 
+	/** length of longest VCF line seen so far, used in StringBuilder(capacity) */
+	private int vcfLineLengthCapacity = 100;
+
 	/**
 	 * Prepare a VCFEncoder that will encode records appropriate to the given VCF header, optionally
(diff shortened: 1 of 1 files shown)


lbergelson:
@lindenb Does this improve anything empirically?  It seems like it could improve things by avoiding StringBuilder expansions, but it seems just as likely that it will slow things down if line length is highly uneven and the each StringBuilder ends up being very large.

I'm slightly skeptical that it will make any difference either way that will show up on a profiler.  Happy to proven wrong though! (It seems like adding a default size might make sense even if adding a variable one doesn't)

Definitely +1 for cleaning up those little issues in the rest of the code though.

lindenb:
made some tests and... It improve the speed ... but not as much as I expected... :-/
Closing.
Thank you for the review anyway :-)

lbergelson:
If it does make some improvement we can merge it.  Small gains are still good, I don't want to block small improvements for not being large enough.  I just wanted to hear to hear that it is actually a measurable improvement.

PR #452, opened 2016-01-24
extended Tuple to a ComparableTuple so it can be used in a Histogram

yfarjoun:
- add equals, toString, and hashCode to Tuple
- extend Tuple to ComparableTuple for when the Types are themselves Comparable

This makes it possible to use ComparableTuple in the Histogram class.

### DIFF
--- a/src/java/htsjdk/samtools/util/ComparableTuple.java
+++ b/src/java/htsjdk/samtools/util/ComparableTuple.java
@@ -1,0 +1,28 @@
+package htsjdk.samtools.util;
+
+/**
+ * A simple extension of the Tuple class that, for comparable Types, allows comparing Tuples of non-null elements.
+ * <p>
+ * The comparison will compare the first arguments and if equal (compareTo returns 0) compare the second arguments.
+ *
+ * @author farjoun
+ */
+public class ComparableTuple<A extends Comparable<A>, B extends Comparable<B>> extends Tuple<A, B> implements Comparable<ComparableTuple<A, B>> {
+
+    public ComparableTuple(final A a, final B b) {
+        super(a, b);
+
+        if (a == null || b == null) {
+            throw new IllegalArgumentException("ComparableTuple's behavior is undefined when containing a null.");
+        }
+    }
+
+    @Override
+    public int compareTo(final ComparableTuple<A, B> o) {
+        int retval = a.compareTo(o.a);
+        if (retval == 0) {
+            retval = b.compareTo(o.b);
+        }
+        return retval;
+    }
+}
--- a/src/java/htsjdk/samtools/util/Tuple.java
+++ b/src/java/htsjdk/samtools/util/Tuple.java
@@ -13,4 +13,30 @@ public class Tuple<A, B> {
         this.a = a;
         this.b = b;
     }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        final Tuple<?, ?> tuple = (Tuple<?, ?>) o;
+
+        if (a != null ? !a.equals(tuple.a) : tuple.a != null) return false;
+        return !(b != null ? !b.equals(tuple.b) : tuple.b != null);
+
+    }
+
+    @Override
+    public int hashCode() {
+        int result = a != null ? a.hashCode() : 0;
+        result = 31 * result + (b != null ? b.hashCode() : 0);
+
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return "[" + a + ", " + b + "]";
+    }
+
 }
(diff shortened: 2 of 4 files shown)


src/java/htsjdk/samtools/util/ComparableTuple.java
```diff
@@ -0,0 +1,20 @@
+package htsjdk.samtools.util;
+
+/**
+ * Created by farjoun on 1/17/16.
+ */
+public class ComparableTuple<A extends Comparable<A>, B extends Comparable<B>> extends Tuple<A, B> implements Comparable<ComparableTuple<A, B>> {
```
lbergelson: Could you add a class comment explaining that these will be sorted by A first and then B?

src/java/htsjdk/samtools/util/Tuple.java
```diff
@@ -13,4 +13,26 @@ public Tuple(final A a, final B b) {
         this.a = a;
         this.b = b;
     }
+
+    @Override
+    public int hashCode() {
```
lbergelson: These will all break if A or B is null.  Recommend using the autogenerated intellij equals + hashcode unless you have a good reason not to

src/java/htsjdk/samtools/util/ComparableTuple.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public ComparableTuple(final A a, final B b) {
+        super(a, b);
```
lbergelson: This should probably reject nulls unless you want to have the `compareTo` deal with them.

lbergelson:
Tests?  At least show they sort properly.

src/java/htsjdk/samtools/util/Tuple.java
yfarjoun: cool. I didn't know that existed!

yfarjoun:
thanks @lbergelson. back to you.

src/java/htsjdk/samtools/util/Tuple.java
```diff
@@ -13,4 +13,28 @@ public Tuple(final A a, final B b) {
...
+        int result = a != null ? a.hashCode() : 0;
+        result = 31 * result + (b != null ? b.hashCode() : 0);
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return "[" + a.toString() + ", " + b.toString() + "]";
```
lbergelson: So this will explode if there is a null still...  This is actually a weird case, because 
```
 return "[" + a+ ", " + b + "]";
```

will be just fine.  I never thought about it before, but there must be special syntactic sugar that checks nulls in string concatenations.

lbergelson:
One more comment about the toString method.  It's a nasty surprise when toString crashes... 

Feel free to merge once that's updated.  Thanks for adding some tests.

src/tests/java/htsjdk/samtools/util/TupleTest.java
```diff
@@ -0,0 +1,62 @@
...
+    @Test
+    public void testHashCode() throws Exception {
+        Assert.assertEquals(new Tuple<>(1, "hi").hashCode(), new Tuple<>(1, "hi").hashCode());
+
+        Assert.assertEquals(new Tuple<>(1, null).hashCode(), new Tuple<>(1, null).hashCode());
+        Assert.assertEquals(new Tuple<>(null, "hi").hashCode(), new Tuple<>(null, "hi").hashCode());
+        Assert.assertEquals(new Tuple<>(null, null).hashCode(), new Tuple<>(null, null).hashCode());
+
+        //even though these are of different types, the value is null and so I have to make these equal...
+        Assert.assertEquals(new Tuple<Integer, Integer>(1, null).hashCode(), new Tuple<Integer, String>(1, null).hashCode());
```
yfarjoun: is this strange? two different class instances getting the same hashCode? I couldn't find a way around it...if you think it should not be this way, and know how to fix, please let me know.

yfarjoun:
@lbergelson I know you said I could merge, but I have a question for you (in the code). responded to other issues.

src/tests/java/htsjdk/samtools/util/TupleTest.java
lbergelson: Doesn't seem strange to me.  It's ok for different things to share the same hashCode.  The contract for hashCode is that two objects that are equal must have the same hashCode, but there's no guarantee that having the same hashcode means object are equal.  

This case is weird because those objects ARE equal, which is probably not what you'd expect, but I think is an unavoidable consequence of type erasure.

yfarjoun: ok. rebased. will merge when tests pass.

lbergelson: :+1:

PR #454, opened 2016-01-27
Updated testng from an ancient version (5.5) to the latest (6.9.10).

tfenne:
And the new version produces reports that are actually legible!


tfenne:
@lbergelson or @droazen any objections?  This should/will be superceded when @lbergelson's gradle build happens, but in the mean time it'd be nice to be using a more modern testng.  My tests all work locally (tested the Picard build also).  If the test here pass, any reason not to upgrade?

droazen:
Objections? No way, this is great! Finally we can use `SkipException` in the SRA tests  :+1:

lbergelson:
:+1:

PR #456, opened 2016-01-31
Protect against massive pileups by allowing for a cap on the accumulation

eitanbanks:
I have not touched the default behavior and have tried to be as minimally invasive
as possible.  But we are seeing OOM problems for tools that use this class and it
is forcing us to up RAM to unacceptably high numbers.

### DIFF
--- a/src/java/htsjdk/samtools/util/SamLocusIterator.java
+++ b/src/java/htsjdk/samtools/util/SamLocusIterator.java
@@ -142,6 +142,19 @@ public class SamLocusIterator implements Iterable<SamLocusIterator.LocusInfo>, C
      */
     private boolean emitUncoveredLoci = true;
 
+    /**
+     * If set, this will cap the number of reads we accumulate for any given position.
+     * Note that if we hit the maximum threshold at the first position in the accumulation queue,
+     * then we throw further reads overlapping that position completely away (including for subsequent positions).
+     * This is a useful feature if one wants to minimize the memory footprint in files with a few massively large pileups,
+     * but it must be pointed out that it could cause major bias because of the non-random nature with which the cap is
+     * applied (the first maxReadsToAccumulatePerLocus reads are kept and all subsequent ones are dropped).
+     */
+    private int maxReadsToAccumulatePerLocus = Integer.MAX_VALUE;
+
+    // Set to true when we have enforced the accumulation limit for the first time
+    private boolean enforcedAccumulationLimit = false;
+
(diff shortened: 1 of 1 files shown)


eitanbanks:
@yfarjoun @ktibbett this addresses the high memory consumption (and subsequent OOMs) of the metrics collectors in Picard (along with a PR that will come there once this gets merged).

src/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -142,6 +142,14 @@ public void add(final SAMRecord read, final int position) {
      */
     private boolean emitUncoveredLoci = true;
 
+    /**
+     * If set, this will cap the number of reads we accumulate for any given position.
+     * Note that if we hit the maximum threshold at the first position in the accumulation queue,
+     * then we throw further reads overlapping that position completely away (including for subsequent positions).
+     * This is a useful feature if one wants to minimize the memory footprint in files with a few massively large pileups.
+     */
+    private long maxReadsToAccumulatePerLocus = -1L;
```
tfenne: Two quick things:
1. Since this is compared to `list.size()` which is int, why not just make it an `int`?
2. Defaulting to `Integer.MAX_VALUE` would give the same default behavior and simplify the check below

tfenne:
@eitanbanks My main concern with this is the non-random nature with which the cap is supplied.  If it were set such that it removed some appreciable fraction of reads it could cause fairly major bias.  I get that implementing it to randomly pick between reads that are already accumulated and reads that are to be accumulated would be difficult though.

src/java/htsjdk/samtools/util/SamLocusIterator.java
eitanbanks: Sure, good point.

eitanbanks:
@tfenne In the use cases I'm seeing, I'm totally okay with that bias.  These are pileups of the same exact read sequence that are millions deep that don't affect the bottom line.  May be some sort of bacterial sequence.
Would it be sufficient then to warn of the potential for this bias in the java docs for the variable and again on the setter?  I can add that now.

tfenne:
Sure - I think a large warning, and perhaps a `log.warn()` the first time any given `SamLocusIterator` enforces the limit would be sufficient.

eitanbanks:
Okay @tfenne I've made those changes and rebased.  Do you want/need to look again?
(Thanks for the early AM review!)

PR #458, opened 2016-02-03
Add SeekablePathStream for java.nio.file.Path.

tomwhite:


zong0jie:
Good works. With this PR I think It can manipulate hdfs-bam-files to some extent. Next time maybe I can just ask here may need a function instead of make a PR, and you will give good solutions :)

tomwhite:
Thanks @zong0jie. Please continue to submit PRs :)

The test failure seems to be an unrelated transient failure (all tests pass locally for me), but I don't seem to have permission to rerun it.

droazen:
@tomwhite I restarted the travis build just now.

tomwhite:
New test failure in RelativeIso8601DateTest is again unrelated.

tomwhite:
@droazen can you review this please when you have time? The test failure is unrelated.

droazen:
@cmnbroad can you take a look at this when you have time?

src/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -56,6 +57,13 @@
     
     abstract public SamReader open(final File file);
 
+    public SamReader open(final Path path) {
```
cmnbroad: We should probably add a test for this, since I don't think its covered elsewhere.

src/java/htsjdk/samtools/SamInputResource.java
```diff
@@ -192,6 +209,62 @@ public File asFile() {
     }
 
     @Override
+    public Path asPath() {
+        return null;
```
cmnbroad: Should return a Path for the file here.

src/java/htsjdk/samtools/SamInputResource.java
```diff
@@ -234,6 +307,11 @@ public File asFile() {
...
+    }
```
cmnbroad: We should probably try to return a Path here as well ?

src/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
```diff
@@ -0,0 +1,123 @@
...
+
+    public static synchronized void closeAllInstances() {
+        Collection<SeekablePathStream> clonedInstances = new HashSet<SeekablePathStream>();
+        clonedInstances.addAll(allInstances);
+        for (SeekablePathStream sfs : clonedInstances) {
+            try {
+                sfs.close();
+            } catch (IOException e) {
+                //TODO
+                //log.error("Error closing SeekableFileStream", e);
```
cmnbroad: Seems like we should re-throw, or at the very least log.

cmnbroad:
Sorry about the delay in getting to this @tomwhite. Looks pretty good; a few minor comments, mostly about filling out the SamInputResource conversion methods.

tomwhite:
Thanks for the review @cmnbroad! I've addressed all your review comments now.

akiezun:
@droazen ping

tomwhite:
@droazen could you take a look at this when you have time.

src/java/htsjdk/samtools/SamFiles.java
```diff
@@ -19,7 +21,19 @@
...
+    }
+
+    /**
+     * Finds the index file associated with the provided SAM file.  The index file must exist and be reachable to be found.
+     *
+     * If the file is a symlink and the index cannot be found, try to unsymlink the file and look for the bai in the actual file path.
+     *
+     * @return The index for the provided SAM, or null if one was not found.
+     */
+    public static Path findIndex(final Path samFile) {
```
yfarjoun: variable name samPath?

tomwhite:
Addressed @yfarjoun's feedback about naming (thanks!), and rebased. This should be ready for merge now.

droazen:
@tomwhite Ahem, well actually this branch has not been formally reviewed yet :) @yfarjoun's comment was just a one-off made while we were looking at the branch during our regular htsjdk review meeting this week. I'll do a full review now.

src/main/java/htsjdk/samtools/SamFiles.java
```diff
@@ -19,61 +21,73 @@
...
+    }
+
+    /**
+     * Finds the index file associated with the provided SAM file.  The index file must exist and be reachable to be found.
+     *
+     * If the file is a symlink and the index cannot be found, try to unsymlink the file and look for the bai in the actual file path.
+     *
+     * @return The index for the provided SAM, or null if one was not found.
+     */
+    public static Path findIndex(final Path samPath) {
```
droazen: `findIndex(Path)` needs direct unit testing -- one test case with a regular bai index file represented as a `Path`, and another with a symlink to a bai, so that we can confirm the existing functionality has not been broken by switching to `Path` here.

src/main/java/htsjdk/samtools/SamFiles.java
```diff
@@ -19,61 +21,73 @@
...
             return null;
         }
     }
 
-    private static File lookForIndex(final File samFile) {// If input is foo.bam, look for foo.bai
-        File indexFile;
-        final String fileName = samFile.getName();
+    private static Path lookForIndex(final Path samPath) {// If input is foo.bam, look for foo.bai
+        Path indexPath;
+        final String fileName = samPath.getFileName().toString();
```
droazen: Will this work if `samPath` is not a normal file (eg., is an HDFS URI or GCS URI?)

src/main/java/htsjdk/samtools/SamFiles.java
```diff
@@ -19,61 +21,73 @@
...
         if (fileName.endsWith(BamFileIoUtils.BAM_FILE_EXTENSION)) {
             final String bai = fileName.substring(0, fileName.length() - BamFileIoUtils.BAM_FILE_EXTENSION.length()) + BAMIndex.BAMIndexSuffix;
-            indexFile = new File(samFile.getParent(), bai);
-            if (indexFile.isFile()) {
-                return indexFile;
+            indexPath = samPath.resolveSibling(bai);
+            if (Files.isRegularFile(indexPath)) {
```
droazen: Same question here -- what happens if `indexPath` is a non-file URI?

src/main/java/htsjdk/samtools/SamFiles.java
```diff
@@ -19,61 +21,73 @@
...
         }
 
         // If foo.bai doesn't exist look for foo.bam.bai
-        indexFile = new File(samFile.getParent(), samFile.getName() + BAMIndex.BAMIndexSuffix);
-        if (indexFile.isFile()) {
-            return indexFile;
+        indexPath = samPath.resolveSibling(fileName + BAMIndex.BAMIndexSuffix);
+        if (Files.isRegularFile(indexPath)) {
+            return indexPath;
         }
```
droazen: Add comment that this code is here and not in the `if (fileName.endsWith(BamFileIoUtils.BAM_FILE_EXTENSION)) {` block because we want to fall back to looking for bai's on cram files as well (at least, that is my interpretation of why this is here instead of above)

src/main/java/htsjdk/samtools/SamFiles.java
```diff
@@ -19,61 +21,73 @@
...
      * @return an index file or null if no index is found.
      */
-    private static File unsymlinkAndLookForIndex(File samFile) {
+    private static Path unsymlinkAndLookForIndex(Path samPath) {
         try {
-            final File canonicalSamFile = samFile.getCanonicalFile();
-            final File canonicalIndexFile = lookForIndex(canonicalSamFile);
-            if ( canonicalIndexFile != null) {
-                System.err.println("The index file " + canonicalIndexFile.getPath()
+            final Path canonicalSamPath = samPath.toRealPath();
```
droazen: Add comment that `toRealPath()` resolves any symbolic links

src/main/java/htsjdk/samtools/SamInputResource.java
```diff
@@ -147,6 +164,9 @@ public String toString() {
             case FILE:
                 childToString = asFile().toString();
                 break;
+            case PATH:
+                childToString = asPath().toString();
```
droazen: Guard against null here (since `asPath()` can return null)

src/main/java/htsjdk/samtools/SamInputResource.java
```diff
@@ -192,6 +212,62 @@ public File asFile() {
     }
 
     @Override
+    public Path asPath() {
+        return fileResource.toPath();
+    }
+
+    @Override
+    public URL asUrl() {
+        return null;
```
droazen: It's strange that `FileInputResource`s can be converted to `Path` but not `URL`, since `Path` itself is convertible to `URL` -- do you want to address this now?

src/main/java/htsjdk/samtools/SamInputResource.java
```diff
@@ -192,6 +212,62 @@ public File asFile() {
...
+
+
+    PathInputResource(final Path pathResource) {
+        super(Type.PATH);
+        this.pathResource = pathResource;
+    }
+
+    @Override
+    public File asFile() {
+        return null;
```
droazen: Should you attempt `Path.toFile()` here?

src/main/java/htsjdk/samtools/SamInputResource.java
```diff
@@ -192,6 +212,62 @@ public File asFile() {
...
+    }
+
+    @Override
+    public Path asPath() {
+        return pathResource;
+    }
+
+    @Override
     public URL asUrl() {
         return null;
```
droazen: Call `Path.toUri().toURL()` here?

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -57,6 +58,13 @@
     
     abstract public SamReader open(final File file);
 
+    public SamReader open(final Path path) {
```
droazen: Should live in `SamReaderFactoryImpl` along with the other `open()` implementations

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -57,6 +58,13 @@
...
+        final SamInputResource r = SamInputResource.of(path);
+        final Path indexMaybe = SamFiles.findIndex(path);
+        if (indexMaybe != null) r.index(indexMaybe);
+        return open(r);
+    }
```
droazen: What if the `Path` is a cram, and the client wants to call `referenceSequence()` on the factory? Do we have a plan to make that work?

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Override
+    public long skip(long n) throws IOException {
+        long initPos = position();
+        sbc.position(initPos + n);
+        return position() - initPos;
+    }
+    
+    public int read(final byte[] buffer, final int offset, final int length) throws IOException {
```
droazen: `@Override` here and for all other implementations of methods from `SeekableStream` in this class

src/main/java/htsjdk/samtools/SamFiles.java
```diff
@@ -19,61 +21,73 @@
...
+            final Path canonicalIndexPath = lookForIndex(canonicalSamPath);
+            if ( canonicalIndexPath != null) {
+                System.err.println("The index file " + canonicalIndexPath.toAbsolutePath()
```
droazen: Gross! Could you change this to print the warning to the logger instead?

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
```diff
@@ -0,0 +1,108 @@
...
+     */
+    static Collection<SeekablePathStream> allInstances =
+            Collections.synchronizedCollection(new HashSet<SeekablePathStream>());
+
+    private final Path path;
+    private final SeekableByteChannel sbc;
+    private final long length;
+    private final ByteBuffer oneByteBuf = ByteBuffer.allocate(1);
+
+    public SeekablePathStream(final Path path) throws IOException {
```
droazen: Ensure `path` is non-null

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
```diff
@@ -0,0 +1,108 @@
...
+        this.path = path;
+        this.sbc = Files.newByteChannel(path);
+        this.length = sbc.size();
+        allInstances.add(this);
+    }
+
+    public long length() {
+        return length;
```
droazen: Could the length change between construction time and a call to `length()`?

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public boolean eof() throws IOException {
+        return length() == position();
+    }
+
+    public void seek(final long position) throws IOException {
```
droazen: What do you expect seeking performance to be like if the `Path` is something like an HDFS file?

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public long position() throws IOException {
+        return sbc.position();
+    }
+
+    @Override
+    public long skip(long n) throws IOException {
+        long initPos = position();
+        sbc.position(initPos + n);
```
droazen: What is the behavior here if this brings us past `eof()`?

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
```diff
@@ -0,0 +1,108 @@
...
+public class SeekablePathStream extends SeekableStream {
+
+    private final static Log LOG = Log.getInstance(SeekablePathStream.class);
+
+    /**
+     * Collection of all open instances.  SeekablePathStream objects are usually open and kept open for the
+     * duration of a session.  This collection supports a method to close them all.
+     */
+    static Collection<SeekablePathStream> allInstances =
+            Collections.synchronizedCollection(new HashSet<SeekablePathStream>());
```
droazen: Can this be private?

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
```diff
@@ -0,0 +1,108 @@
...
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.SeekableByteChannel;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+
+public class SeekablePathStream extends SeekableStream {
```
droazen: Add direct unit tests for this class that fully exercise `read()` (both versions), `skip()`, `seek()`, `position()`, and `length()` (can use a `Path` backed by a regular file for the tests).

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -214,6 +217,24 @@ public void exhaustInputResourcePermutation(final SamInputResource resource) thr
...
+        final SAMFileHeader fileHeader = reader.getFileHeader();
+        reader.hasIndex();
+        reader.indexing().hasBrowseableIndex();
+        reader.close();
+
+        /* Ensure all tests have read the same records in the same order or, if this is the first test, set it as the template. */
+        observedHeaders.add(fileHeader);
+        observedRecordOrdering.add(slurped);
+        Assert.assertEquals(observedHeaders.size(), 1, "read different headers than other testcases");
+        Assert.assertEquals(observedRecordOrdering.size(), 1, "read different records than other testcases");
```
droazen: This test is not meaningful if it happens to run *before* `exhaustInputResourcePermutation()` has populated  `observedHeaders` and `observedRecordOrdering` instead of after it. This is a problem since TestNG by default randomizes test ordering.

droazen: Recommend that instead of using `observedHeaders` and `observedRecordOrdering` directly in this test case, you build the expected values separately within this method using `SamReaderFactory.open(file)`, and then assert that the results you get from `SamReaderFactory.open(path)` match the ones you got from `open(file)`. That way you don't share data with `exhaustInputResourcePermutation()` and don't have to worry about test ordering issues.

droazen:
Review complete -- back to @tomwhite for changes.

src/main/java/htsjdk/samtools/SamFiles.java
tomwhite: Yes, it will. `samPath.getFileName()` returns the last component of the Path, which is then converted to a string.

src/main/java/htsjdk/samtools/SamFiles.java
tomwhite: I just tried it with an HDFS Path (in the jsr-203 project) and it works as expected.

src/main/java/htsjdk/samtools/SamFiles.java
tomwhite: The first block looks for foo.bai, whereas this one looks for foo.bam.bai. I think the comments cover this, but let me know if there's an improvement you'd like to see.

src/main/java/htsjdk/samtools/SamInputResource.java
tomwhite: This is an existing issue with all the other resource types, so I've left it to be consistent. This should really be fixed separately.

src/main/java/htsjdk/samtools/SamInputResource.java
tomwhite: OK.

src/main/java/htsjdk/samtools/SamInputResource.java
tomwhite: Yes, done.

src/main/java/htsjdk/samtools/SamReaderFactory.java
tomwhite: The reason I didn't add an abstract method here was to maintain backwards compatibility. Is there a better way of adding the overloaded method?

src/main/java/htsjdk/samtools/SamReaderFactory.java
tomwhite: That should work fine as the implementation of `open` will use the reference sequence set on the factory (if any).

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
tomwhite: Yes, done.

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
tomwhite: I didn't do this because the other `Seekable*Stream` implementations don't.

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
tomwhite: It could. The reason I didn't do this before was because `SeekableByteChannel.size()` can throw an `IOException`, but `SeekableStream.length` cannot. I've caught the exception, logged it, and returned 0 now (which is consistent with `java.io.File`).

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
tomwhite: Seeks should be avoided if possible, since it's better to stream through a file in HDFS. Having said that, in HDFS seeks are implemented pretty efficiently - as a regular file seek if within the same HDFS block, or as a transparent seek to another datanode (if the seek takes you to another block).

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
tomwhite: It will succeed, but then any attempt to read will get an EOF. This is the same as `SeekableFileStream`, and is specified here: https://docs.oracle.com/javase/8/docs/api/java/nio/channels/SeekableByteChannel.html#position-long-

tomwhite:
Thanks for the review @droazen. I've addressed all the points you made - a few (which I've addressed in a comment) are longstanding issues that would be best addressed separately. Back to you.

src/test/java/htsjdk/samtools/SamFilesTest.java
```diff
@@ -58,6 +60,15 @@ public void testFindIndexForSuffixes(final String dataFileSuffix, final String i
 
         Assert.assertNotNull(foundIndexFile);
         Assert.assertTrue(foundIndexFile.getName().endsWith(expectIndexSuffix));
+
+        final Path foundIndexPath = SamFiles.findIndex(dataFile.toPath());
+        if (expectIndexSuffix == null) {
+            Assert.assertNull(foundIndexPath);
+            return;
```
droazen: Refactor out these `return` statements from this test case, replacing with `else` blocks -- otherwise when `expectIndexSuffix == null` you'll only test the `File` case and not the `Path` case

src/main/java/htsjdk/samtools/SamFiles.java
```diff
@@ -19,61 +24,73 @@
...
+    private static Path unsymlinkAndLookForIndex(Path samPath) {
         try {
-            final File canonicalSamFile = samFile.getCanonicalFile();
-            final File canonicalIndexFile = lookForIndex(canonicalSamFile);
-            if ( canonicalIndexFile != null) {
-                System.err.println("The index file " + canonicalIndexFile.getPath()
+            final Path canonicalSamPath = samPath.toRealPath(); // resolve symbolic links
+            final Path canonicalIndexPath = lookForIndex(canonicalSamPath);
+            if ( canonicalIndexPath != null) {
+                LOG.error("The index file " + canonicalIndexPath.toAbsolutePath()
```
droazen: I think this should be a `LOG.warn()`, not `LOG.error()`

src/main/java/htsjdk/samtools/SamFiles.java
droazen: Can you add a comment here documenting that this will work correctly with URIs?

src/main/java/htsjdk/samtools/SamFiles.java
droazen: The improvement I'd like to see is to mention that this fallback code at the bottom looks for bai indices on cram files as well (which is allowed). Eg., `look for foo.bam.bai or foo.cram.bai`

src/main/java/htsjdk/samtools/SamReaderFactory.java
droazen: Ok, fair enough

src/main/java/htsjdk/samtools/SamReaderFactory.java
droazen: What I mean, though, is that `referenceSequence()` takes a `File` -- there is no overload that takes a `Path`. If you're not going to address this here, can you open a ticket for it?

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
```diff
@@ -0,0 +1,118 @@
...
+import htsjdk.samtools.util.Log;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.SeekableByteChannel;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+
```
droazen: Add class-level javadoc for this new class

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
droazen: Ok

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
```diff
@@ -0,0 +1,118 @@
...
+public class SeekablePathStream extends SeekableStream {
+
+    private final static Log LOG = Log.getInstance(SeekablePathStream.class);
+
+    /**
+     * Collection of all open instances.  SeekablePathStream objects are usually open and kept open for the
+     * duration of a session.  This collection supports a method to close them all.
+     */
+    private static Collection<SeekablePathStream> allInstances =
+            Collections.synchronizedCollection(new HashSet<SeekablePathStream>());
```
droazen: Can this be final?

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -214,6 +217,24 @@ public void exhaustInputResourcePermutation(final SamInputResource resource) thr
         Assert.assertEquals(observedRecordOrdering.size(), 1, "read different records than other testcases");
     }
 
+    @Test
+    public void openPath() throws IOException {
+        final Path path = localBam.toPath();
+        final SamReader reader = SamReaderFactory.makeDefault().open(path);
```
droazen: Close both readers in this test case when done using try-with-resources

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -214,6 +217,24 @@ public void exhaustInputResourcePermutation(final SamInputResource resource) thr
...
+        LOG.info(String.format("Reading from %s ...", path));
+        final List<SAMRecord> records = Iterables.slurp(reader);
+        final SAMFileHeader fileHeader = reader.getFileHeader();
+        reader.hasIndex();;
```
droazen: Do you want to assert that `hasIndex()` is true? Also, remove stray semicolon.

droazen:
Second-pass review complete, back to @tomwhite. Please squash/rebase after addressing remaining comments so we can merge.

src/main/java/htsjdk/samtools/SamReaderFactory.java
tomwhite: You can construct a `ReferenceSource` with a `Path` and pass that to `referenceSequence()` - so there is a way of doing it a bit less directly.

tomwhite:
Thanks for the second review @droazen. I've addressed all the remaining comments.

droazen:
:+1: merging!

tomwhite:
Thanks!

Issue #461, opened 2016-02-04
Fix inexplicable SRA test failures


lbergelson:
We've seen a case of  `Invalid SRA accession was passed to SRA reader` in many tests. Before running the tests we checked that `SRAAccession.isSupported()`  The accession number usually passes, so I think it must be failing `NGS.isValid(acc)` (or `SRAAccession.isSupported()` has changed between invocations... is that possible?)

Following the trail leads down to an undocumented native method call  `private native static boolean IsValid ( String spec );` so it's hard to tell why things are failing.

These errors were transient and we've been unable to reproduce them, but they've caused a major headache for production so it's necessary that we prevent this from happening again.  See #459 for the temporary solution.

lbergelson:
@a-nikitiuk @kwrodarmer  Here's a fun one for you.  Sorry I don't have more to go on.  There's a stack trace, but at the moment it's lost due to atlassian failures.  I'll update when that comes back online.

kwrodarmer:
It would be helpful enough to at least give us a reproducible command line, or accession in question. These things have a lot to do with behaviors. We need to know, for example, if an accession leads to something already downloaded and in a user's cache area, or whether it is only accessible with dbGaP authorization, or whether the original submitter has for whatever reason withdrawn its publication.

lbergelson:
Here's the testng output from one of them.  22 SRA tests failed with similar stack traces.  We're unable to reproduce the failure and subsequent attempts in the same environment passed.  

It's a bit worrisome that something like the contents of the users cache could affect test results.  Is there a way to isolate the tests from the rest of the system?
 
```
[...]

Issue #459, opened 2016-02-04
Disable failing SRA tests until they can be made more robust


lbergelson:
We're seeing intermittent SRA test failures despite #434.  This is causing headaches for production.  

We're seeing `IllegalArgumentException` "Invalid SRA accession was passed to SRA reader" in many tests.  Before running the tests we checked that `SRAAccession.isSupported()` The accession number usually passes, so I think it must be failing `NGS.isValid(acc)` (or `SRAAccession.isSupported() has changed between invocations... is that possible?`) 

Following the trail leads down to an undocumented native method call to  ` private native static boolean IsValid ( String spec );`  so it's hard to tell why things are failing.

These errors were transient and we've been unable to reproduce them.

We should place all the SRA tests into their own test group, and not run that as part of the standard tests until the tests are more robust and #442 is fixed.

cmnbroad:
Can this be closed (fixed in https://github.com/samtools/htsjdk/pull/638 ?)

magicDGS:
I think that this is already fix, no?

lbergelson:
Marking this as done.

Issue #442, opened 2016-01-20
SRA test creating a huge local file/cache


cmnbroad:
Running code pulled from master (up to commit 96a55713e673daaab8fc2938615eb47725eb4c6b), [this SRA test](https://github.com/samtools/htsjdk/blob/00337afdb712380722ff484e689f78fa79badfdd/src/tests/java/htsjdk/samtools/sra/SRATest.java#L104) tests fill up the hard drive on my MacBook running Yosemite. The same code/build/test (ish) running on my Ubuntu VM on a Windows host leaves only a 16M file. In both case, the tests all pass. I'm fairly sure its not downloading that much data since I'm behind a relatively slow T1 line and the test runs much too quickly. If I delete the ncbi folder and rerun the test the same thing reappears:

~$ du -h ncbi
193G	ncbi/public/sra
193G	ncbi/public
193G	ncbi

~$ ls -l ncbi/public/sra
total 403865600
-rw-r--r--  1 cmn  staff  206779187200 Jan 20 14:47 SRR822962.sra.cache

cmnbroad:
BTW, its the [first test](https://github.com/samtools/htsjdk/blob/00337afdb712380722ff484e689f78fa79badfdd/src/tests/java/htsjdk/samtools/sra/SRATest.java#L93) in the data provider that causes the problem. If I run just the second test (or even all the tests in the SRA package except that first one), the cache is there but much smaller.

droazen:
@a-nikitiuk Could you have a look at this?

a-nikitiuk:
Thanks for the information provided. We are working on a fix

kwrodarmer:
The actual fix for this will be to supply a sparse-file implementation for HFS, which resides in the C library that will either need to be automatically downloaded or require a manual download step on the part of the user. We'll let you know when that's available.

We can quickly disable the first test, if that helps in the meantime.

lbergelson:
In general tests should not leave any files lying around.  Please have tests delete any files they create.  A lot of people would be unpleasantly surprised to have ~/ncbi show up in their home directory without explanation, even if it only takes up 16mb. 

Is there a way for users to specify where the SRA cache goes?

kwrodarmer:
I've been considering how to address this one... It seems to me that we might want to be more clear about requirements, especially those that are in conflict with one another.

[...]

PR #465, opened 2016-02-08
moving SRA tests to their own test group

### LINKED ISSUE #436: SRA tests should be skipped with a SkipException if they can't be run
Currently SRA tests pass if the SRA native code isn't available.  This is very deceptive.  They should be skipped instead. (or possibly run in their own test group)

requires #435

lbergelson:
removing SRA tests from `ant test` and moving them to their own `ant sra-test` target

   this should isolate SRA tests so that they can no longer fail during a picard release
   
once #461 and  #442 are fixed they should be moved back into the standard `ant test`

- added a new ant target sra-test which exclusively runs the sra tests
- made all SRA test classes descend from a new `AbstractSRATest` base class
  - this ensures they are all set in the appropriate test group
  - each SRA test method will run `assertSRAIsSupported()` and be skipped if it is not
    - removed the newly redundant checks in every SRA test
 - pulled `checkAlignedUnalignedCountsByIterator` up into the base class to remove duplicated code

- updated .travis.yml to run sra-test after the rest of the tests

fixes #436 and #459

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -119,31 +119,45 @@
     </target>
 
     <!-- TEST -->
+    <macrodef name="run-test">
+        <attribute name="excludedTestGroups" default=""/>
+        <attribute name="includedTestGroups" default=""/>
+        <sequential>
+                <taskdef resource="testngtasks" classpathref="classpath"/>
+                <testng suitename="htsjdk-tests" classpathref="classpath" outputdir="${test.output}"
+                        failureproperty="tests.failed" excludedgroups="@{excludedTestGroups}" groups="@{includedTestGroups}"
+                        workingDir="${basedir}"
+                        verbose="${testng.verbosity}">
+                    <classpath>
+                        <pathelement path="${classes}"/>
+                        <pathelement path="${classes.test}"/>
+                        <pathelement path="${scripts}"/>
+                    </classpath>
+                    <classfileset dir="${classes.test}">
+                        <include name="**/Test*.class"/>
+                        <include name="**/*Test.class"/>
+                    </classfileset>
+                    <jvmarg value="-Xmx2G"/>
+                </testng>
+                <junitreport todir="${dist}/test" >
+                    <fileset dir="${test.output}">
+                        <include name="*.xml"/>
+                    </fileset>
+                    <report format="noframes" todir="${dist}/test" styledir="etc/test"/>
+                </junitreport>
+                <copy file="etc/test/testng.css" todir="${dist}/test" overwrite="true"/>
+                <fail if="tests.failed" message="There were failed unit tests"/>
+        </sequential>
+    </macrodef>
+
     <target name="test" depends="compile, set_excluded_test_groups" description="Run unit tests">
-        <taskdef resource="testngtasks" classpathref="classpath"/>
-        <testng suitename="htsjdk-tests" classpathref="classpath" outputdir="${test.output}"
-                failureproperty="tests.failed" excludedgroups="${excludedTestGroups}" workingDir="${basedir}"
-                verbose="${testng.verbosity}">
-            <classpath>
-                <pathelement path="${classes}"/>
-                <pathelement path="${classes.test}"/>
-                <pathelement path="${scripts}"/>
-            </classpath>
-            <classfileset dir="${classes.test}">
-                <include name="**/Test*.class"/>
-                <include name="**/*Test.class"/>
-            </classfileset>
-            <jvmarg value="-Xmx2G"/>
-        </testng>
-        <junitreport todir="${dist}/test" >
-            <fileset dir="${test.output}">
-                <include name="*.xml"/>
-            </fileset>
-            <report format="noframes" todir="${dist}/test" styledir="etc/test"/>
-        </junitreport>
-        <copy file="etc/test/testng.css" todir="${dist}/test" overwrite="true"/>
-        <fail if="tests.failed" message="There were failed unit tests"/>
+            <run-test excludedTestGroups="${excludedTestGroups}, sra"/>
     </target>
+
+    <target name="sra-test" depends="compile, set_excluded_test_groups" description="Run SRA unit tests">
+        <run-test includedTestGroups="sra" excludedTestGroups="${excludedTestGroups}"/>
+    </target>
+
 
     <target name="single-test"
             depends="compile, compile-tests"
--- a/src/tests/java/htsjdk/samtools/sra/AbstractSRATest.java
+++ b/src/tests/java/htsjdk/samtools/sra/AbstractSRATest.java
@@ -1,0 +1,57 @@
+package htsjdk.samtools.sra;
+
+import htsjdk.samtools.SAMRecord;
+import htsjdk.samtools.SAMRecordIterator;
+import org.testng.Assert;
+import org.testng.SkipException;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import java.util.NoSuchElementException;
+
+@Test(groups = "sra")
+public abstract class AbstractSRATest {
+
+    @BeforeMethod
+    public final void assertSRAIsSupported(){
+        if(!SRAAccession.isSupported()){
+            throw new SkipException("Skipping SRA Test because SRA native code is unavailable.");
+        }
+    }
+
+    /**
+     * Exhaust the iterator and check that it produce the expected number of mapped and unmapped reads.
+     * Also checks that the hasNext() agrees with the actual results of next() for the given iterator.
+     * @param expectedNumMapped expected number of mapped reads, specify -1 to skip this check
+     * @param expectedNumUnmapped expected number of unmapped reads, specify -1 to skip this check
+     */
+    static void assertCorrectCountsOfMappedAndUnmappedRecords(SAMRecordIterator samRecordIterator,
+                                                                        int expectedNumMapped, int expectedNumUnmapped) {
+        int numMapped = 0, numUnmapped = 0;
+        while (true) {
+            boolean hasRecord = samRecordIterator.hasNext();
+            SAMRecord record;
+            try {
+                record = samRecordIterator.next();
+                Assert.assertNotNull(record);
+                Assert.assertTrue(hasRecord); // exception is not thrown if we came to this point
+            } catch (final NoSuchElementException e) {
+                Assert.assertFalse(hasRecord);
+                break;
+            }
+
+            if (record.getReadUnmappedFlag()) {
+                numUnmapped++;
+            } else {
+                numMapped++;
+            }
+        }
+
+        if (expectedNumMapped != -1) {
+            Assert.assertEquals(numMapped, expectedNumMapped);
+        }
+        if (expectedNumUnmapped != -1) {
+            Assert.assertEquals(numUnmapped, expectedNumUnmapped);
+        }
+    }
+}
--- a/src/tests/java/htsjdk/samtools/sra/SRAIndexTest.java
+++ b/src/tests/java/htsjdk/samtools/sra/SRAIndexTest.java
@@ -35,11 +35,9 @@ import org.testng.Assert;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Iterator;
-import java.util.List;
 import java.util.Set;
 
 /**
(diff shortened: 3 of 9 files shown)


lbergelson:
@droazen I know you're swamped, but I was hoping you could suggest how to avoid the big copy/paste in the build.xml.  I don't know how you get ant tasks to inherit from one another.

build.xml
```diff
@@ -50,10 +50,10 @@
         <os family="unix"/>
     </condition>
     <target name="set_excluded_test_groups_unix" if="isUnix">
-        <property name="excludedTestGroups" value="slow, broken"/>
+        <property name="excludedTestGroups" value="slow, broken, SRA"/>
     </target>
     <target name="set_excluded_test_groups_non_unix" unless="isUnix">
-        <property name="excludedTestGroups" value="slow, broken, unix"/>
+        <property name="excludedTestGroups" value="slow, broken, unix, SRA"/>
```
nh13: the other groups are lowercase, would it make sense to do the same here?

src/tests/java/htsjdk/samtools/sra/AbstractSRATest.java
```diff
@@ -0,0 +1,56 @@
...
+        if(!SRAAccession.isSupported()){
+            throw new SkipException("Skipping SRA Test because SRA native code is unavailable.");
+        }
+    }
+
+    protected static void checkAlignedUnalignedCountsByIterator(SAMRecordIterator samRecordIterator,
+                                                              int numberAlignments, int numberUnalignments) {
+        int countAlignments = 0, countUnalignments = 0;
+        while (
+                true) {
```
nh13: looks like there's an errant newline

src/tests/java/htsjdk/samtools/sra/AbstractSRATest.java
```diff
@@ -0,0 +1,56 @@
...
+    @BeforeMethod
+    public final void assertSRAIsSupported(){
+        if(!SRAAccession.isSupported()){
+            throw new SkipException("Skipping SRA Test because SRA native code is unavailable.");
+        }
+    }
+
+    protected static void checkAlignedUnalignedCountsByIterator(SAMRecordIterator samRecordIterator,
+                                                              int numberAlignments, int numberUnalignments) {
+        int countAlignments = 0, countUnalignments = 0;
```
nh13: while we're here, can we change these?
```
numberAlignments   -> expectedNumMapped
numberUnalignments -> expectedNumUnmapped
countAlignments    -> numMapped
countUnalignments  -> numUnmapped
```

src/tests/java/htsjdk/samtools/sra/AbstractSRATest.java
```diff
@@ -0,0 +1,56 @@
...
+        while (
+                true) {
+            boolean hasRecord = samRecordIterator.hasNext();
+            SAMRecord record = null;
+            try {
+                record = samRecordIterator.next();
+                Assert.assertTrue(hasRecord); // exception is not thrown if we came to this point
+            } catch (NoSuchElementException e) {
```
nh13: can be `final`

src/tests/java/htsjdk/samtools/sra/AbstractSRATest.java
```diff
@@ -0,0 +1,56 @@
...
+                Assert.assertFalse(hasRecord);
+            }
+
+            Assert.assertEquals(hasRecord, record != null);
```
nh13: Is this needed given we check `hasRecord` on line 31 and 33?

src/tests/java/htsjdk/samtools/sra/AbstractSRATest.java
```diff
@@ -0,0 +1,56 @@
...
+public abstract class AbstractSRATest {
+
+    @BeforeMethod
+    public final void assertSRAIsSupported(){
+        if(!SRAAccession.isSupported()){
+            throw new SkipException("Skipping SRA Test because SRA native code is unavailable.");
+        }
+    }
+
+    protected static void checkAlignedUnalignedCountsByIterator(SAMRecordIterator samRecordIterator,
```
nh13: Could you rename this while we are making changes?  Perhaps `checkMappedStatusCounts`?

src/tests/java/htsjdk/samtools/sra/AbstractSRATest.java
```diff
@@ -0,0 +1,56 @@
...
+            }
+
+            if (record.getReadUnmappedFlag()) {
+                countUnalignments++;
+            } else {
+                countAlignments++;
+            }
+        }
+
+        if (numberAlignments != -1) {
```
nh13: Could you add doc stating that if this is `-1` that the values wont be checked, only `hasRecord`?

src/tests/java/htsjdk/samtools/sra/AbstractSRATest.java
nh13: If you change them here, can you change them in the tests too?

src/tests/java/htsjdk/samtools/sra/AbstractSRATest.java
```diff
@@ -0,0 +1,56 @@
...
+            Assert.assertEquals(numberAlignments, countAlignments);
+        }
+        if (numberUnalignments != -1) {
```
nh13: Same comment.

src/tests/java/htsjdk/samtools/sra/SRAIndexTest.java
```diff
@@ -64,15 +62,13 @@ public void testLevelSize() {
 
     @Test
     public void testLevelForBin() {
-        if (!SRAAccession.isSupported()) return;
-
         SRAIndex index = getIndex(DEFAULT_ACCESSION);
```
nh13: So many missing `final`s; my IntelliJ would not be pleased.

src/tests/java/htsjdk/samtools/sra/SRAQueryTest.java
```diff
@@ -45,8 +39,6 @@ public void testUnmappedCounts(String acc, int numberUnalignments) {
 
     @Test(dataProvider = "testReferenceAlignedCounts")
     public void testReferenceAlignedCounts(String acc, String reference, int refernceStart, int numberAlignments) {
```
nh13: type in `refernceStart`; also, most if not all parameters in the test methods can be final.  Since we are touching so much of the code, would it make sense to add in these finishing touches?

src/tests/java/htsjdk/samtools/sra/SRAQueryTest.java
```diff
@@ -68,8 +60,6 @@ public void testReferenceAlignedCounts(String acc, String reference, int refernc
 
     @Test(dataProvider = "testQueryCounts")
     public void testQueryCounts(String acc, String reference, int refernceStart, int referenceEnd, boolean contained, int numberAlignments, int numberUnalignment) {
```
nh13: typo here too.

droazen:
@lbergelson Discussed in person ways to avoid the duplication in `build.xml` -- ping me once you've made that change, and I'll review.

build.xml
lbergelson: lower cased

src/tests/java/htsjdk/samtools/sra/AbstractSRATest.java
lbergelson: done

src/tests/java/htsjdk/samtools/sra/AbstractSRATest.java
lbergelson: done

src/tests/java/htsjdk/samtools/sra/AbstractSRATest.java
lbergelson: removed, seems redundant

lbergelson: Adding an explicit non-null check for record as well.

src/tests/java/htsjdk/samtools/sra/SRAQueryTest.java
lbergelson: fixed

src/tests/java/htsjdk/samtools/sra/SRAQueryTest.java
lbergelson: fixed

src/tests/java/htsjdk/samtools/sra/SRAIndexTest.java
lbergelson: I think I got them all.  For some reason my intellij isn't reporting them in the test files and I can't figure out why not...

src/tests/java/htsjdk/samtools/sra/AbstractSRATest.java
lbergelson: fixed

src/tests/java/htsjdk/samtools/sra/AbstractSRATest.java
lbergelson: fixed

src/tests/java/htsjdk/samtools/sra/AbstractSRATest.java
lbergelson: going with `assertCorrectCountsOfMappedAndUnmappedRecords` verbose but hopefully clear

lbergelson:
@nh13 Thanks for the review.  I believe I've addressed your concerns now.

@droazen could you give my ant file changes a glance to be sure they're sane?

build.xml
```diff
@@ -38,6 +38,7 @@
     <property name="javac.target" value="1.8"/>
     <property name="javac.debug" value="true"/>
 
+    <property name="includedTestGroups" value=""/>
```
droazen: Can you make this a param in the `antcall` for the `test` target? It is confusing to have an empty value defined up here like this.

build.xml
```diff
@@ -145,6 +147,20 @@
         <fail if="tests.failed" message="There were failed unit tests"/>
     </target>
 
+    <target name="test" depends="compile, set_excluded_test_groups" description="Run unit tests">
+        <antcall target="run-test">
```
droazen: Pass in `includedTestGroups` as a param here as well instead of relying on a property.

build.xml
```diff
@@ -145,6 +147,20 @@
...
+            <param name="excludedTestGroups" value="${excludedTestGroups}, sra"/>
```
droazen: What if `excludedTestGroups` is empty? Then you'd get `value=", sra"`. Will this be ok?

droazen: Never mind, I see that `excludedTestGroups` is guaranteed to be set to a non-empty value elsewhere in the build.

build.xml
```diff
@@ -145,6 +147,20 @@
...
+        </antcall>
+
+    </target>
+    <target name="sra-test" depends="compile, set_excluded_test_groups" description="Run SRA unit tests">
+        <antcall target="run-test">
+            <param name="includedTestGroups" value="sra"/>
+            <param name="excludedTestGroups" value="${excludedTestGroups}"/>
```
droazen: The params in an `antcall` will fail to override properties that are explicitly set on the command line, as I understand things. Are these include/exclude properties ever explicitly set on the command line (using `-D`)?

build.xml
```diff
@@ -119,10 +120,11 @@
     </target>
 
     <!-- TEST -->
-    <target name="test" depends="compile, set_excluded_test_groups" description="Run unit tests">
+    <target name="run-test" depends="compile, set_excluded_test_groups" description="Run unit tests">
```
droazen: I think you might want to turn `run-test` into a macro using `macrodef` (there are examples of this elsewhere in this build file). Macros can be parameterized a bit more cleanly (without using `antcall`).

droazen:
@lbergelson I took a look, and my main suggestion is to turn `run-test` into a macro. Macros can be parameterized using attributes (no `antcall` needed), and you don't have to worry about potential collision issues with properties set explicitly on the command line, etc.

tfenne:
@lbergelson: I'm not trying to be a pain, and I wish I'd thought to say this sooner but would it not be less work to wrap up your conversion to gradle (and test the Picard build), than to invest all this time with the ant build?

lbergelson:
@tfenne That's a good point.... this would be trivial in gradle.

lbergelson:
I sort of expected it to be trivial in ant too... but I've already had to learn more about ant than I wanted to know.

tfenne:
Yeah - Ant's 100% declarative model makes some things way way more difficult than they need to be :(

droazen:
I think this qualifies as an urgent change that should not wait for the gradle migration (it's causing major headaches for people right now)

Issue #437, opened 2016-01-14
GenotypeLikelihoods should handle anyploid, not just diploid 


ronlevine:
`GenotypeLikelihoodsAllelePair GenotypeLikelihoods.getAllelePair(final int PLindex)` returns the cached alleles corresponding to the PLindex. This cache is diploid because the method that initializes it, `GenotypeLikelihoodsAllelePair[] GenotypeLikelihoods.calculatePLcache(final int altAlleles)`,  uses a diploid number of likelihoods ( `numLikelihoods(1 + altAlleles, 2)`) and adds an allele pair each PLindex, instead of the ploidy number of allele indices. If the methods are generalized, they should return
`ArrayList<Integer> ` instead of `GenotypeLikelihoodsAllelePair`.

yfarjoun:
hmmm. I think that the optimization of using a GenotypeLikelihoodsAllelePair overweighs the generalizability of multipliody. 

is there a usecase for this?

yfarjoun:
is it needed for select variants? perhaps a non-diploid method could be **added** rather than changing the signature of this function.

ronlevine:
Here's the usecase: https://github.com/broadinstitute/gatk/issues/1382
The plan is to keep the diploid optimization.

ronlevine:
@yfarjoun Could you review the accompanying pull request? It should be ready momentarily.

yfarjoun:
It will not be this week, as I"m out conferencing in agbt. but if it can
wait for next week then sure.

Y.

ronlevine:
@yfarjoun Thanks. I'll look at your comments next week while I'm in Miami.

ronlevine:
I think this should be closed.

PR #466, opened 2016-02-09
Handle aneuploid likelihoods

ronlevine:
Implements https://github.com/samtools/htsjdk/issues/437.

Previously, only diploid allele pair indices could be accessed by PL index:
`GenotypeLikelihoodsAllelePair getAllelePair(final int PLindex)`

This pull request adds the ability to get the allele indices for any ploidy:
`List<Integer> GenotypeLikelihoods.getAlleles(final int PLindex, final int ploidy)`

### DIFF
--- a/src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
+++ b/src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
@@ -29,14 +29,17 @@ import htsjdk.tribble.TribbleException;
 import htsjdk.variant.utils.GeneralUtils;
 import htsjdk.variant.vcf.VCFConstants;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.EnumMap;
 import java.util.List;
+import java.util.Map;
+import java.util.HashMap;
 
 public class GenotypeLikelihoods {
     private final static int NUM_LIKELIHOODS_CACHE_N_ALLELES = 5;
     private final static int NUM_LIKELIHOODS_CACHE_PLOIDY = 10;
-    // caching numAlleles up to 5 and ploidy up to 10
+    // caches likelihoods up to 5 alleles and up to 10 ploidy
     private final static int[][] numLikelihoodCache = new int[NUM_LIKELIHOODS_CACHE_N_ALLELES][NUM_LIKELIHOODS_CACHE_PLOIDY];
 
     public final static int MAX_PL = Integer.MAX_VALUE;
@@ -48,7 +51,6 @@ public class GenotypeLikelihoods {
     //
     private double[] log10Likelihoods = null;
     private String likelihoodsAsString_PLs = null;
-
 
     /**
      * initialize num likelihoods cache
@@ -63,14 +65,21 @@ public class GenotypeLikelihoods {
     }
 
     /**
-     * The maximum number of alleles that we can represent as genotype likelihoods
+     * The maximum number of diploid alternate alleles that we can represent as genotype likelihoods
      */
-    public final static int MAX_ALT_ALLELES_THAT_CAN_BE_GENOTYPED = 50;
+    public final static int MAX_DIPLOID_ALT_ALLELES_THAT_CAN_BE_GENOTYPED = 50;
 
-    /*
-    * a cache of the PL index to the 2 alleles it represents over all possible numbers of alternate alleles
-    */
-    private final static GenotypeLikelihoodsAllelePair[] PLIndexToAlleleIndex = calculatePLcache(MAX_ALT_ALLELES_THAT_CAN_BE_GENOTYPED);
+    /**
+     * A cache of the PL index to the 2 alleles it represents over all possible numbers of alternate alleles
+     */
+    private final static GenotypeLikelihoodsAllelePair[] diploidPLIndexToAlleleIndex = calculateDiploidPLcache(MAX_DIPLOID_ALT_ALLELES_THAT_CAN_BE_GENOTYPED);
+
+    /**
+     * A cache of PL index to a list of alleles for any ploidy.
+     * For example, for a ploidy of 3, the allele lists for each PL index is:
+     * {0,0,0}, {0,0,1}, {0,1,1}, {1,1,1}, {0,0,2}, {0,1,2}, {1,1,2}, {0,2,2}, {1,2,2}, {2,2,2}
+     */
+    protected final static Map<Integer, List<List<Integer>>> anyploidPloidyToPLIndexToAlleleIndices = new HashMap<Integer, List<List<Integer>>>();
 
     public final static GenotypeLikelihoods fromPLField(String PLs) {
         return new GenotypeLikelihoods(PLs);
@@ -101,11 +110,11 @@ public class GenotypeLikelihoods {
     }
 
     /**
-     * Returns the genotypes likelihoods in negative log10 vector format.  pr{AA} = x, this
+     * The genotypes likelihoods in -10log10(x) vector format.  pr{AA} = x, this
      * vector returns math.log10(x) for each of the genotypes.  Can return null if the
      * genotype likelihoods are "missing".
      *
-     * @return
+     * @return genotypes likelihoods in negative log10 vector format
      */
     public double[] getAsVector() {
         // assumes one of the likelihoods vector or the string isn't null
@@ -318,7 +327,13 @@ public class GenotypeLikelihoods {
         }
     }
 
-    private static GenotypeLikelihoodsAllelePair[] calculatePLcache(final int altAlleles) {
+    /**
+     * Calculate the cache of diploid alleles for each PL index
+     *
+     * @param altAlleles   number of alternate alleles
+     * @return cache of diploid alleles for each PL index
+     */
+    private static GenotypeLikelihoodsAllelePair[] calculateDiploidPLcache(final int altAlleles) {
         final int numLikelihoods = numLikelihoods(1 + altAlleles, 2);
         final GenotypeLikelihoodsAllelePair[] cache = new GenotypeLikelihoodsAllelePair[numLikelihoods];
 
@@ -338,6 +353,60 @@ public class GenotypeLikelihoods {
         return cache;
     }
 
+
+    /**
+     * Calculate the alleles for each PL index for a ploidy.
+     * Creates the ordering for all possible combinations of ploidy alleles. Computed recursively and the
+     * result is stored in a cache.
+     *
+     * The implementation is described in The Variant Call Format Specification VCF 4.3, Section 1.6.2 Genotype fields
+     * The likelihoods are ordered for ploidy P and N alternate alleles as follows:
+     * for aP = 0...N
+     *  for aP-1 = 0...aP
+     *      ...
+     *      for a1 = 0..…a2
+     *          a1,a2..aP
+     *
+     * This is implemented recursively:
+     *
+     * PLIndexToAlleleIndices(N, P, suffix=empty):
+     *      for a in 0...N
+     *          if (P == 1) accum += (a + suffix)  // have all the alleles for a PL index
+     *          if (P > 1) PLIndexToAlleleIndices(a, P-1, a + suffix )
+     *
+     * @param altAlleles     Number of alternate alleles
+     * @param ploidy         Number of chromosomes in set
+     * @param anyploidPLIndexToAlleleIndices PL index to the alleles of general ploidy over all possible alternate alleles
+     * @param genotype       An entry of ploidy alleles
+     */
+    private static void calculatePLIndexToAlleleIndices(final int altAlleles, final int ploidy, final List<List<Integer>> anyploidPLIndexToAlleleIndices,
+                                                   final List<Integer> genotype) {
+        for (int a=0; a <= altAlleles; a++) {
+            final List<Integer> gt = new ArrayList<Integer>(Arrays.asList(a));
+            gt.addAll(genotype);
+            if ( ploidy == 1 ) {// have all ploidy alleles for a PL index
+                anyploidPLIndexToAlleleIndices.add(gt);
+            } else if ( ploidy > 1 ) {
+                calculatePLIndexToAlleleIndices(a, ploidy - 1, anyploidPLIndexToAlleleIndices, gt);
+            }
+        }
+    }
+
+    /**
+     * Calculate the cache of allele indices for each PL index for a ploidy.
+     * Calculation in @see #calculatePLIndexToAlleleIndices
+     *
+     * @param altAlleles Number of alternate alleles
+     * @param ploidy    Number of chromosomes in set
+     * @return PL index to the alleles of general ploidy over all possible alternate alleles
+     * @return The alleles for each PL index for a ploidy
+     */
+    protected static List<List<Integer>> calculateAnyploidPLcache(final int altAlleles, final int ploidy) {
+        List<List<Integer>> anyploidPLIndexToAlleleIndices = new ArrayList<List<Integer>>();
+        calculatePLIndexToAlleleIndices(altAlleles, ploidy, anyploidPLIndexToAlleleIndices, new ArrayList<Integer>());
+        return anyploidPLIndexToAlleleIndices;
+    }
+
     // -------------------------------------------------------------------------------------
     //
     // num likelihoods given number of alleles and ploidy
@@ -347,9 +416,9 @@ public class GenotypeLikelihoods {
     /**
      * Actually does the computation in @see #numLikelihoods
      *
-     * @param numAlleles
-     * @param ploidy
-     * @return
+     * @param numAlleles    number of alleles
+     * @param ploidy        number of chromosomes
+     * @return  number of likelihoods
      */
     private static final int calcNumLikelihoods(final int numAlleles, final int ploidy) {
         if (numAlleles == 1)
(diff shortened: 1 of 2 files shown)


ronlevine:
@yfarjoun Please review.

ronlevine:
@yfarjoun Please review if you can.

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -101,11 +108,11 @@ private GenotypeLikelihoods(double[] asVector) {
     }
 
     /**
-     * Returns the genotypes likelihoods in negative log10 vector format.  pr{AA} = x, this
+     * The genotypes likelihoods in negative log10 vector format.  pr{AA} = x, this
      * vector returns math.log10(x) for each of the genotypes.  Can return null if the
```
yfarjoun: seems like it returns -10*log10(x), right?

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -318,6 +325,12 @@ public GenotypeLikelihoodsAllelePair(final int alleleIndex1, final int alleleInd
         }
     }
 
+    /**
+     * Calculate the cache of diploid alleles for each PL index
+     *
+     * @param altAlleles   number of alternate alleles
+     * @return cache of diploid alleles for each PL index
+     */
     private static GenotypeLikelihoodsAllelePair[] calculatePLcache(final int altAlleles) {
```
yfarjoun: might as well rename this function (its private anyhow) to calculateDiploidPLcache

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -338,6 +351,29 @@ public GenotypeLikelihoodsAllelePair(final int alleleIndex1, final int alleleInd
...
+    /**
+     * Calculate the cache of anyploid alleles for each PL index
+     * Creates the ordering for all possible combinations of ploidy alleles. Computed recursively and the
+     * result is stored in a cache.
+     *
+     * @param numAlleles    Number of alleles (including ref)
+     * @param ploidy        Number of chromosomes in set
+     * @param genotype      An entry of ploidy alleles
+     */
+    protected static void calculatePLcache(final int numAlleles, final int ploidy, final List<Integer> genotype){
```
yfarjoun: and rename this one calculateAnyPloidyPLcache (BTW, the thing you are refering to is spelled 'aneuploid' or 'aneuploidy' I'm specifically using the spelling above to indicate 'any ploidy'...i.e. not only diploid.

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -338,6 +351,29 @@ public GenotypeLikelihoodsAllelePair(final int alleleIndex1, final int alleleInd
...
+        if (genotype.size() == ploidy) {
+            plIndexToAlleleIndices.add(genotype);
+        }
```
yfarjoun: please use } else { on the same line.

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -338,6 +351,29 @@ public GenotypeLikelihoodsAllelePair(final int alleleIndex1, final int alleleInd
...
+        else {
+            for (int a=0; a < numAlleles; a++) {
+                List gt = new ArrayList<Integer>(Arrays.asList(a));
```
yfarjoun: why is gt not a List<Integer>?

ronlevine: Isn't it a List?

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -338,6 +351,29 @@ public GenotypeLikelihoodsAllelePair(final int alleleIndex1, final int alleleInd
         return cache;
     }
 
+
+    /**
+     * Calculate the cache of anyploid alleles for each PL index
+     * Creates the ordering for all possible combinations of ploidy alleles. Computed recursively and the
+     * result is stored in a cache.
```
yfarjoun: could you include the recursive formula you are using and a reference for where it comes from?

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
yfarjoun: sorry, markdown ate my comment:
why is gt not a `List<Integer>`?

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -338,6 +351,29 @@ public GenotypeLikelihoodsAllelePair(final int alleleIndex1, final int alleleInd
...
+     * @param genotype      An entry of ploidy alleles
+     */
+    protected static void calculatePLcache(final int numAlleles, final int ploidy, final List<Integer> genotype){
+        if (genotype.size() == ploidy) {
+            plIndexToAlleleIndices.add(genotype);
+        }
+        else {
+            for (int a=0; a < numAlleles; a++) {
+                List gt = new ArrayList<Integer>(Arrays.asList(a));
+                gt.addAll(genotype);
```
yfarjoun: looks like `gt` is a simple copy of `genotype`, was that intended?

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
ronlevine: Because my backspace key ate my `<Integer>`. ;-)

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -72,6 +72,13 @@
     */
     private final static GenotypeLikelihoodsAllelePair[] PLIndexToAlleleIndex = calculatePLcache(MAX_ALT_ALLELES_THAT_CAN_BE_GENOTYPED);
```
yfarjoun: change the name of the static int to something that references 'diploid'

yfarjoun: nm. I guess it's the same limit on non-diploid.

yfarjoun: also, it would be good to add "diploid" to the variable itself.

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
yfarjoun: I'm slowly getting the combinatorial construction you are doing here....bear with me...

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -338,6 +351,29 @@ public GenotypeLikelihoodsAllelePair(final int alleleIndex1, final int alleleInd
...
+     * Creates the ordering for all possible combinations of ploidy alleles. Computed recursively and the
+     * result is stored in a cache.
+     *
+     * @param numAlleles    Number of alleles (including ref)
+     * @param ploidy        Number of chromosomes in set
+     * @param genotype      An entry of ploidy alleles
+     */
+    protected static void calculatePLcache(final int numAlleles, final int ploidy, final List<Integer> genotype){
+        if (genotype.size() == ploidy) {
+            plIndexToAlleleIndices.add(genotype);
```
yfarjoun: perhaps add a comment here 
`// and we are done with the calculation`
so help the poor reader...

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
ronlevine: This will help: refer to "Algorithm for enumerating the genotypes given ploidy and alleles" in  http://genome.sph.umich.edu/wiki/Relationship_between_Ploidy,_Alleles_and_Genotypes

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
yfarjoun: protect against someone putting genotype with .size()>ploidy, as that will not terminate.

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
yfarjoun: Use Collections.SingletonList(a) instead of asList?

ronlevine: Yes. Safer to make it immutable and preface it with a "final".

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -72,6 +72,13 @@
     */
     private final static GenotypeLikelihoodsAllelePair[] PLIndexToAlleleIndex = calculatePLcache(MAX_ALT_ALLELES_THAT_CAN_BE_GENOTYPED);
 
+
+    /*
+    * a cache of the PL index to the anyploid alleles it represents over all possible numbers of alternate alleles
+    */
+    protected final static List<List<Integer>> plIndexToAlleleIndices = new ArrayList<List<Integer>>();
```
yfarjoun: This should be a `Map<Integer, List<Integer>>`, not a `List<List<Integer>>`, no? That way it is clear that you put something in the right place. regardless of order of insertion. Also it is much clearer then that the map is ploidy->plIndexToAlleleIndeciesList(of that ploidy)

yfarjoun: scratch that. I got it wrong. it's not at all what I though...and I don't like it! The problem is that one cannot deal with two different ploidies at the same time since this is a singleton object I think that this should be encapsulated in a map (`Map<Integer,List<List<Integer>>>`) or an inner class so that one could have multiple instances of NonDiplodGenotypeLikelihoods with different ploidies...somewhat like LogMath in MathUtil.

src/tests/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
```diff
@@ -182,22 +187,122 @@ public void testCalculatePLindex(){
...
+                                Arrays.asList(1, 2, 2),
+                                Arrays.asList(2, 2, 2)
+                        )
+                }
+        };
+    }
+
+    @Test(dataProvider = "testCalculatePLcacheData")
+    public void testCalculatePLcache(final int numAlleles, final int ploidy,  final List<List<Integer>> expected){
+        GenotypeLikelihoods.plIndexToAlleleIndices.clear();
```
yfarjoun: Test should be using `initializePLIndexToAlleleIndices`, not replecating the code.

src/tests/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
```diff
@@ -182,22 +187,122 @@ public void testCalculatePLindex(){
...
+        for ( int i=0; i < GenotypeLikelihoods.plIndexToAlleleIndices.size() ; i++){
+            Assert.assertTrue(GenotypeLikelihoods.plIndexToAlleleIndices.get(i).equals(expected.get(i)));
+        }
+    }
+
+
+    @DataProvider
+    public Object[][] testGetAllelesData() {
+        return new Object[][]{
+                {0, 2, 3, Arrays.asList(0,0,0)},
```
yfarjoun: spaces after commas in lists

src/tests/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
```diff
@@ -182,22 +187,122 @@ public void testCalculatePLindex(){
...
+                {7, 1, 3},
+                {8, 2, 3},
+                {9, 3, 3}
+        };
     }
-        
-    private void allelePairTest(int PLindex, int allele1, int allele2) {
+
+    @Test(dataProvider = "testGetAllelePairData")
+    public void testGetAllelePair(final int PLindex, final int allele1, final int allele2){
```
yfarjoun: space between `)` and `{`

src/tests/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
```diff
@@ -182,22 +187,122 @@ public void testCalculatePLindex(){
...
+                                Arrays.asList(0, 2, 2),
+                                Arrays.asList(1, 2, 2),
+                                Arrays.asList(2, 2, 2)
+                        )
+                }
+        };
+    }
+
+    @Test(dataProvider = "testCalculatePLcacheData")
+    public void testCalculatePLcache(final int numAlleles, final int ploidy,  final List<List<Integer>> expected){
```
yfarjoun: space here too.

src/tests/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
```diff
@@ -182,22 +187,122 @@ public void testCalculatePLindex(){
...
+        GenotypeLikelihoods.plIndexToAlleleIndices.clear();
+        GenotypeLikelihoods.calculatePLcache(numAlleles, ploidy, new ArrayList<Integer>());
+        for ( int i=0; i < GenotypeLikelihoods.plIndexToAlleleIndices.size() ; i++){
```
yfarjoun: space

src/tests/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
```diff
@@ -182,22 +187,122 @@ public void testCalculatePLindex(){
...
+                {6, 2, 3, Arrays.asList(1,1,2)},
+                {7, 2, 3, Arrays.asList(0,2,2)},
+                {8, 2, 3, Arrays.asList(1,2,2)},
+                {9, 2, 3, Arrays.asList(2,2,2)},
+                {1, 2, 1, Arrays.asList(1)}
+        };
+    }
+
+    @Test(dataProvider = "testGetAllelesData")
+    public void testGetAlleles(final int PLindex, final int numAltAlleles, final int ploidy, final List<Integer> expected ){
```
yfarjoun: space

src/tests/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
```diff
@@ -182,22 +187,122 @@ public void testCalculatePLindex(){
...
+    @DataProvider
+    public Object[][] testGetAllelesIndexOutOfBoundsData() {
+        return new Object[][]{
+                {-1, 2, 3},
+                {10, 2, 3}
+        };
+    }
+
+    @Test(dataProvider = "testGetAllelesIndexOutOfBoundsData", expectedExceptions = IndexOutOfBoundsException.class)
+    public void testGetAllelesOutOfBounds(final int PLindex, final int numAltAlleles, final int ploidy){
```
yfarjoun: space

src/tests/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
```diff
@@ -182,22 +187,122 @@ public void testCalculatePLindex(){
...
+        GenotypeLikelihoods.initializePLIndexToAlleleIndices(numAltAlleles, ploidy);
+        final List<Integer> alleles = GenotypeLikelihoods.getAlleles(PLindex, ploidy);
+    }
+
+    @Test(expectedExceptions = IllegalStateException.class)
+    public void testGetAllelesUnitialized(){
```
yfarjoun: space

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
yfarjoun: space

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -422,6 +459,43 @@ public static GenotypeLikelihoodsAllelePair getAllelePair(final int PLindex) {
         return PLIndexToAlleleIndex[PLindex];
     }
 
+    /**
+     * initialize cache of allele ploidy indices
+     *
+     * @param altAlleles number of alternate alleles
+     * @param ploidy    number of chromosomes
+     */
+    public static synchronized void initializePLIndexToAlleleIndices(final int altAlleles, final int ploidy){
```
yfarjoun: space

src/tests/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
```diff
@@ -29,12 +29,17 @@
 // the imports for unit testing.
 
 
+import com.sun.glass.ui.SystemClipboard;
```
yfarjoun: SystemClipboard?

src/tests/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
```diff
@@ -29,12 +29,17 @@
...
+import com.sun.tools.javac.api.ClientCodeWrapper;
```
yfarjoun: I think you have a few extra import statements

yfarjoun:
back to you @ronlevine. please do not rebase your next commit.

src/tests/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
ronlevine: Y. InelliJ black magiic

ronlevine:
Isn't aneuploidy "the presence of an abnormal number of chromosomes in a cell"? I was using anyploidy to be equivalent to general ploidy.

yfarjoun:
yes. in that case: CamelCase please.

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
ronlevine: How about keeping the data structure as is but lazily initializing it with the double checked lock pattern in `initializeAneuploidPLIndexToAlleleIndices(final int altAlleles, final int ploidy))`? `aneuploidPLIndexToAlleleIndices` will only exist for one ploidy and the operations will be thread safe.

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
ronlevine: Actually, gt is mutable since the next statement adds `genotype` so will make it `final List<Integer>`.

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
yfarjoun: So you'll also store a static `ploidy` variable that will tell you for what ploidy the last is initialized and if it doesn't match, you'll reinitialize it? is that the idea? I'm not sure what the advantage is. Using a map will be thread safe will be just as "lazy" and allow for the odd case where two different ploidies are requested to remain as efficient. happy to be convinced otherwise...just interested in hearing the reasoning.

ronlevine: Yes, the ploidy would have to be saved. My logic is to only deal with one ploidy. With the `map`, the application developer could initialize for multiple ploidys, which would waste memory.

yfarjoun: well, if the developer needs more than one ploidy, what else should they
do? if they do it without needing, that's their fault. I would put a
comment in the initialization function that makes it clear that one should
only initialize ploidies that one is interested in...

ronlevine: :+1: I'll go with that.

ronlevine:
@yfarjoun Please review. I think I addressed all of your comments. Note that I changed the allele ordering (calculateAneuploidPLcache()) to be consistent with the VCF 4.3 spec. Please comment if the accompanying doc is not clear or sufficient.

ronlevine:
@yfarjoun Comments addressed. Please review.

ronlevine:
@yfarjoun Please review. Responded to comments and made code/comments more legible.

ronlevine:
@yfarjoun Please take another look.

src/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -338,6 +353,59 @@ public GenotypeLikelihoodsAllelePair(final int alleleIndex1, final int alleleInd
...
+     * @param anyploidPLIndexToAlleleIndices PL index to the alleles of general ploidy over all possible alternate alleles
+     * @param genotype       An entry of ploidy alleles
+     */
+    private static void calculatePLIndexToAlleleIndices(final int altAlleles, final int ploidy, final List<List<Integer>> anyploidPLIndexToAlleleIndices,
+                                                   final List<Integer> genotype) {
+        for (int a=0; a <= altAlleles; a++) {
+            final List<Integer> gt = new ArrayList<Integer>(Arrays.asList(a));
+            gt.addAll(genotype);
+            if ( ploidy == 1 ) // have all ploidy alleles for a PL index
+                anyploidPLIndexToAlleleIndices.add(gt);
```
yfarjoun: sorry, missed this before: multiple lines, please use braces:
```
if () {
...
} else {
..
}

yfarjoun:
thanks @ronlevine for this. minor comment and then :+1:  for rebase/squash/merge.

PR #477, opened 2016-02-16
squid:S2162 - equals methods should be symmetric and work for subclasses

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule squid:S2162 - "equals" methods should be symmetric and work for subclasses

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:S2162

M-Ezzat

### DIFF
--- a/src/java/htsjdk/samtools/Bin.java
+++ b/src/java/htsjdk/samtools/Bin.java
@@ -77,8 +77,7 @@ public class Bin implements Comparable<Bin> {
     @Override
     public boolean equals(final Object other) {
         if(other == null) return false;
-        if(!(other instanceof Bin)) return false;
-
+        if(this.getClass() != other.getClass()) return false;
         final Bin otherBin = (Bin)other;
         return this.referenceSequence == otherBin.referenceSequence && this.binNumber == otherBin.binNumber;
     }
--- a/src/java/htsjdk/samtools/SAMSequenceRecord.java
+++ b/src/java/htsjdk/samtools/SAMSequenceRecord.java
@@ -176,7 +176,8 @@ public class SAMSequenceRecord extends AbstractSAMHeaderRecord implements Clonea
     @Override
     public boolean equals(final Object o) {
         if (this == o) return true;
-        if (!(o instanceof SAMSequenceRecord)) return false;
+        if (o == null) return false;
+        if (this.getClass() != o.getClass()) return false;
 
         final SAMSequenceRecord that = (SAMSequenceRecord) o;
 
--- a/src/java/htsjdk/samtools/util/Interval.java
+++ b/src/java/htsjdk/samtools/util/Interval.java
@@ -159,12 +159,12 @@ public class Interval implements Comparable<Interval>, Cloneable, Locatable {
 
     /** Equals method that agrees with {@link #compareTo(Interval)}. */
     public boolean equals(final Object other) {
-        if (!(other instanceof Interval)) return false;
-        else if (this == other) return true;
-        else {
-            Interval that = (Interval)other;
-            return (this.compareTo(that) == 0);
-        }
+        if (this == other) return true;
+        if (other == null) return false;
+        if (this.getClass() != other.getClass()) return false;
+
+        Interval that = (Interval)other;
+        return (this.compareTo(that) == 0);
     }
 
     @Override
--- a/src/java/htsjdk/samtools/Cigar.java
+++ b/src/java/htsjdk/samtools/Cigar.java
@@ -255,7 +255,8 @@ public class Cigar implements Serializable, Iterable<CigarElement> {
     @Override
     public boolean equals(final Object o) {
         if (this == o) return true;
-        if (!(o instanceof Cigar)) return false;
+        if (o == null) return false;
+        if (this.getClass() != o.getClass()) return false;
 
         final Cigar cigar = (Cigar) o;
 
--- a/src/java/htsjdk/samtools/CigarElement.java
+++ b/src/java/htsjdk/samtools/CigarElement.java
@@ -51,7 +51,8 @@ public class CigarElement implements Serializable {
     @Override
     public boolean equals(final Object o) {
         if (this == o) return true;
-        if (!(o instanceof CigarElement)) return false;
+        if (o == null) return false;
+        if (this.getClass() != o.getClass()) return false;
 
         final CigarElement that = (CigarElement) o;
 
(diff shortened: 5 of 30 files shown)


src/java/htsjdk/samtools/Bin.java
```diff
@@ -76,8 +76,12 @@ public int getBinNumber() {
      */
     @Override
     public boolean equals(final Object other) {
-        if(other == null) return false;
-        if(!(other instanceof Bin)) return false;
+        if(other == null) {
```
tfenne: Can you please stick to single-line style `if`s for guard statements?  I.e.
```java
if (other == null) return false;`
```
is preferred over
```java
if (other == null) {
    return false;
}
```
especially when used to return early at the top of methods.

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1782,8 +1782,15 @@ protected void eagerDecode() {
...
     public boolean equals(final Object o) {
-        if (this == o) return true;
-        if (!(o instanceof SAMRecord)) return false;
+        if (this == o) {
+            return true;
+        }
+        if(o == null) {
+            return false;
+        }
+        if (this.getClass() != o.getClass()) {
```
tfenne: This change needs some real discussion.  @nh13 @yfarjoun @droazen @vadimzalunin - what are your thoughts on this?  With this change a `SAMRecord`, `BAMRecord` and `CRAMRecord` that are all logically equal, will no longer be equal.  I think this is a bad idea - it's more important that `SAMRecord` implement logical equality than to slavishly follow java convention that `equals` require type equality also.

src/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -175,8 +175,15 @@ private URI makeURI(final String s) throws URISyntaxException {
...
     public boolean equals(final Object o) {
-        if (this == o) return true;
-        if (!(o instanceof SAMSequenceRecord)) return false;
+        if (this == o) {
+            return true;
+        }
+        if (o == null) {
+            return false;
+        }
+        if (this.getClass() != o.getClass()) {
```
tfenne: Similarly here - I don't think people tend to subclass SAMSequenceRecord, but I think this class should continue to implement logical equality and not require type equality.

src/java/htsjdk/samtools/util/Interval.java
```diff
@@ -159,12 +159,17 @@ public int compareTo(final Interval that) {
...
-        else {
-            Interval that = (Interval)other;
-            return (this.compareTo(that) == 0);
+        if (this == other) {
+            return true;
         }
+        if (other == null) {
+            return false;
+        }
+        if (this.getClass() != other.getClass()) {
```
tfenne: This is another one that might cause problems.  @lbergelson you've done a lot of refactoring to pull interval/region-like methods out into a single class; do you see changing `Interval.equals()` to not accept subclasses?

tfenne:
@m-ezzat Thanks for doing this; this is a change that will need much more careful review from the team.

@vadimzalunin Lots of Cram classes have their equality changed to not accept sub-classes in this PR.  Can you take a look and decide if that's correct in all cases?

@droazen Similarly, can you or someone more familiar review the various `variant` classes?

Does anyone know enough to review the changes to tribble classes?

yfarjoun:
there seems to be a failure due to SamRecord.equals change throwing the test off....

cmnbroad:
For some reason, SAMRecordSetBuilder.getSAMReader actually serializes the records out to a .sam file through a BAMWriter (!), so the records that come back when reading that file are actually BAM records. Its not clear to me to what extent the current implementation which allows for BAM/SAM record equality is intentional.

tfenne:
@cmnbroad It's entirely intentional.  The sub-classing of `SAMRecord` is generally used to add optimizations and perhaps extended behaviour.  If a pair of SAMRecords are identical other than one coming from a `.sam` file another from a `.bam` or `.cram` file, I would expect `equals()` to return true.

src/java/htsjdk/samtools/SAMRecord.java
lbergelson: Changing this a very scary idea.  I would vote against it.

droazen:
@tfenne @yfarjoun @cmnbroad @lbergelson @m-ezzat I'd advise against making this change, and recommend that this PR be closed. This is a somewhat contentious issue in the Java community, but for the purposes of htsjdk I have to agree with Josh Bloch (from http://www.artima.com/intv/bloch17.html):

> The reason that I favor the instanceof approach is that when you use the getClass approach, you have the restriction that objects are only equal to other objects of the same class, the same run time type. If you extend a class and add a couple of innocuous methods to it, then check to see whether some object of the subclass is equal to an object of the super class, even if the objects are equal in all important aspects, you will get the surprising answer that they aren't equal. In fact, this violates a strict interpretation of the Liskov substitution principle, and can lead to very surprising behavior. In Java, it's particularly important because most of the collections (HashTable, etc.) are based on the equals method. If you put a member of the super class in a hash table as the key and then look it up using a subclass instance, you won't find it, because they are not equal.

tfenne:
I tend to agree @droazen.  I'll give a day or two for folks to offer other viewpoints, and close it if not persuaded.

src/java/htsjdk/samtools/util/Interval.java
lbergelson: @tfenne We primarily `Locatable` and gatk`s `SimpleInterval` so this wouldn't really effect our work.  It seems like a bad idea though.  Anyone inheriting from `Interval` is expecting this equality to work in one way and changing it now is going to cause major headaches.

lbergelson:
I also agree with @droazen.  There are cases where equality should be restricted to instances of the identical class, but classes like `SAMRecord` have been built around the assumption that subclasses will be comparable to each other.  This does come  with added responsibility to make sure that subclasses don't implement equals or do so only in a way that is compatible with the original implementor.  

If we have classes that are violating their logical equality we should definitely fix that (I've found a few that had equals / hashcode issues in the past).  I don't think we can or should enforce identical classes as part of that though.

droazen:
One compromise solution that I've often seen recommended is to continue using `instanceof` where desired, but mark `equals()` methods that use it as `final`. This enforces the contract that the supertype has exclusive responsibility for defining equality between its subtypes, and preserves symmetry/commutativity of equality operations.

There would be a few places in htsjdk where we might not be able to do this -- eg., `SRALazyRecord` has an override of `SAMRecord.equals()` that just loads some lazy data before returning `super.equals(other)` directly -- but we could probably make most of these `equals()` methods final.

droazen:
Closing this for the reasons given above

PR #478, opened 2016-02-16
write CRAM3 by default

vadimzalunin:
Mostly works out of the box.

### DIFF
--- a/src/main/java/htsjdk/samtools/cram/common/CramVersions.java
+++ b/src/main/java/htsjdk/samtools/cram/common/CramVersions.java
@@ -7,5 +7,5 @@ public class CramVersions {
     /**
      * The default CRAM version when creating a new CRAM output file or stream.
      */
-    public static final Version DEFAULT_CRAM_VERSION = CRAM_v2_1;
+    public static final Version DEFAULT_CRAM_VERSION = CRAM_v3;
 }
--- a/src/main/java/htsjdk/samtools/cram/structure/ContainerHeaderIO.java
+++ b/src/main/java/htsjdk/samtools/cram/structure/ContainerHeaderIO.java
@@ -64,6 +64,14 @@ class ContainerHeaderIO {
         return true;
     }
 
+    /**
+     * Write CRAM {@link Container} out into the given {@link OutputStream}.
+     * @param major CRAM major version
+     * @param container container to be written
+     * @param outputStream the output stream to write the container to
+     * @return number of bytes written out to the output stream
+     * @throws IOException as per java IO contract
+     */
     public int writeContainerHeader(final int major, final Container container, final OutputStream outputStream)
             throws IOException {
         final CRC32OutputStream crc32OutputStream = new CRC32OutputStream(outputStream);
@@ -80,7 +88,7 @@ class ContainerHeaderIO {
 
         if (major >= 3) {
             outputStream.write(crc32OutputStream.getCrc32_LittleEndian());
-            length += 4 * 8;
+            length += 4 ;
         }
 
         return length;
--- a/src/main/java/htsjdk/samtools/seekablestream/SeekableMemoryStream.java
+++ b/src/main/java/htsjdk/samtools/seekablestream/SeekableMemoryStream.java
@@ -50,7 +50,7 @@ public class SeekableMemoryStream extends SeekableStream {
     @Override
     public int read() throws IOException {
         if (buf.position() < buf.limit()) {
-            return buf.get();
+            return buf.get() & 0xFF;
         } else {
             return -1;
         }
--- a/src/test/java/htsjdk/samtools/cram/VersionTest.java
+++ b/src/test/java/htsjdk/samtools/cram/VersionTest.java
@@ -1,0 +1,106 @@
+package htsjdk.samtools.cram;
+
+import htsjdk.samtools.CRAMFileWriter;
+import htsjdk.samtools.SAMFileHeader;
+import htsjdk.samtools.SAMRecord;
+import htsjdk.samtools.cram.build.CramIO;
+import htsjdk.samtools.cram.common.CramVersions;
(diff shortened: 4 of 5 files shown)


droazen:
@cmnbroad please review

src/java/htsjdk/samtools/cram/common/CramVersions.java
```diff
@@ -7,5 +7,5 @@
     /**
      * The default CRAM version when creating a new CRAM output file or stream.
      */
-    public static final Version DEFAULT_CRAM_VERSION = CRAM_v2_1;
+    public static final Version DEFAULT_CRAM_VERSION = CRAM_v3;
```
cmnbroad: Please add a test that verifies that the files we're creating now are actually v3 files, and that we can read them back in once we create one.

cmnbroad:
@vadimzalunin It looks like we already have quite a few tests that read existing v3 files, but I'm not sure how much coverage there is for writing them. As far as I can tell the version differences are the crc, the  version stamp in the file header, the eof container, and the unknown base handling. Is there anything else different between v3 and v2.1 that we should have a validation test for ?

vadimzalunin:
@cmnbroad added version3 tests
Also fixed a bug in SeekableMemoryStream, my fault!

src/tests/java/htsjdk/samtools/cram/VersionTest.java
```diff
@@ -0,0 +1,111 @@
...
+        w.addAlignment(record);
+        w.close();
+
+        byte[] cramBytes = baos.toByteArray();
+        // expecting 2529 bytes but the size may vary in the future:
+        Assert.assertTrue(cramBytes.length > 100);
+
+        // ensure version bytes:
+        Assert.assertEquals(cramBytes[4], version.major);
+        Assert.assertEquals(cramBytes[5], version.minor);
```
cmnbroad: I would remove lines 55-60. The tests in the code that follow below seem like the right way to test this, and the 100 byte threshold doesn't seem to have any meaning.

cmnbroad:
@vadimzalunin Thanks for adding those tests. One more minor comment and then :+1: from me.

vadimzalunin:
@cmnbroad fixed but Travis failed due to some internet issue.

cmnbroad:
@vadimzalunin Looks like the travis issue was intermittent - I just restarted the tests and it seems fine. But I just noticed the issues in https://github.com/samtools/htsjdk/issues/506, and I think we should hold off on merging this one in we fix those and then we can validate those for > v3.0 files.

src/tests/java/htsjdk/samtools/seekablestream/SeekableMemoryStreamTest.java
```diff
@@ -0,0 +1,93 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009 The Broad Institute
```
yfarjoun: 2016

src/tests/java/htsjdk/samtools/cram/VersionTest.java
```diff
@@ -0,0 +1,105 @@
...
+        long containerStart = cramSeekableStream.position();
+        Container container = ContainerIO.readContainer(version, cramSeekableStream);
+        Assert.assertNotNull(container);
+
+        // ensure EOF follows:
+        Container eof = ContainerIO.readContainer(version, cramSeekableStream);
+        Assert.assertNotNull(eof);
+        Assert.assertTrue(eof.isEOF());
+
+        // TODO: ensure there is nothing left in the stream.
```
yfarjoun: TODO?

src/java/htsjdk/samtools/seekablestream/SeekableMemoryStream.java
```diff
@@ -50,7 +50,7 @@ public void seek(final long position) throws IOException {
     @Override
     public int read() throws IOException {
         if (buf.position() < buf.limit()) {
-            return buf.get();
+            return buf.get() & 0xFF;
```
akiezun: can you comment what this change is for?

src/java/htsjdk/samtools/cram/structure/ContainerHeaderIO.java
```diff
@@ -80,7 +88,7 @@ public int writeContainerHeader(final int major, final Container container, fina
 
         if (major >= 3) {
             outputStream.write(crc32OutputStream.getCrc32_LittleEndian());
-            length += 4 * 8;
+            length += 4 ;
```
akiezun: can you comment what this change is for?

cmnbroad:
@vadimzalunin Were you able to get to the bottom of why this change is necessary ? We'd like to get this change in but would like to understand why we need this change. Thanks.

src/java/htsjdk/samtools/cram/structure/ContainerHeaderIO.java
cmnbroad: Looking at the htslib code, I'm assuming this code was originally ported incorrectly (see https://github.com/samtools/htslib/blob/develop/cram/cram_io.c#L3159), but it never mattered  until we made the change in this PR to switch to writing v3 files because this code path was never executed.

vadimzalunin:
@cmnbroad yes, are correct: writeContainerHeader methods must return the number of bytes written in contrast to for example ITF8.writeUnsignedITF8 which returns number of bits. 

The bug originated in cramtools.

yfarjoun:
needs a rebase

vadimzalunin:
@yfarjoun rebased but SRA tests are failing, is it intermittent?

cmnbroad:
@vadimzalunin It looks like the sra tests are failing elsewhere as well. I'd wait a bit and then restart.

cmnbroad:
The SRA tests are working now so I restarted these.

Issue #471, opened 2016-02-11
Investigate potential Fasta reader performance regression


lbergelson:
There was a report about terrible performance reading zipped fasta files in GATK.  Since no changes were made to FASTA reading in GATK itself it's possible that this is due to changes in htsjdk.  

Those with access can view it here: https://github.com/broadinstitute/dsde-docs/issues/570

The relevant report is:

```
I noticed a performance bottleneck with newer GATK versions (I'm on the latest version right now v3.5-0-g36282e4, Compiled 2015/11/25 04:03:56) where (simple) analysis types such as SelectVariants or VariantsToTable would run for days when they would previously complete in less than an hour. I have to add that I am using bgzip-compressed and tabix-indexed VCF files throughout the analysis and that I see the performance problem on our cluster running several processes in parallel. Most Java processes would show a rather low percentage of CPU consumption, while our cluster file system maxes out at around 100,000 read requests per second. Using 'strace' of an active Java thread, I noticed that millions of "pread" system calls of 183 bytes against the reference sequence *.fasta file seem to be issued.

  [ ~]$ strace -p 28395
  Process 28395 attached - interrupt to quit
  pread(11, "TGGTAAAATGTGGTTGGATGAAGCGTACGCTT"..., 183, 46038724) = 183
  pread(11, "TAGACCATTCTATCAAAATGCTCTTTCTACAG"..., 183, 46038907) = 183
  ...
The question is, what am I doing wrong? Alternatively, are you aware of any changes to how the reference sequence FASTA file is read in recent GATK versions especially in the context of bgzip-compressed and tabix-indexed VCF files. Are you aware that so many read requests are issued? I tried to go back to version v3.3-0-g37228af, Compiled 2014/10/24 01:07:22 and although the "pread" requests seem to include larger regions, the run time of a single process seems similar. I would have to go back further, but you may be quicker in doing so.

Suggested by our system administrator, I managed to overcome the performance bottleneck by copying the reference sequence *.fasta, *.fasta.fai and *.dict files into the Linux shared memory tmpfs file system (/dev/shm/). As an example, VariantsToTable run times have fallen from 5 hours to just 31 seconds (13h vs 74s per 1M sites), SelectVariants from 5 hours to 18 minutes (26m vs 87s per 1M sites). In case I'm not the only one seeing this, would it make sense to read (and keep in memory) larger chunks of the reference sequence?
```

[...]

PR #479, opened 2016-02-16
Fix perf bug in IndexedFastaSequenceFile.

cmnbroad:
Fix for perf regression described in https://github.com/samtools/htsjdk/issues/471,  which was introduced in https://github.com/samtools/htsjdk/pull/252. Affects the size of the read buffer used when reading fasta files. I don't see any obvious way to write a test to verify this programmatically.


droazen:
:+1: from me, assuming that you've manually tested the performance effects of this change, and confirmed that performance of queries for long sequences is greatly improved, and performance of queries for short sequences is not adversely affected.

Back to @cmnbroad

cmnbroad:
I've done a bit more crude measurement of how this affects performance, and I only see on average about 20-40% time improvement for a 1MB reference sequence, which is considerably less than the huge performance degradation originally reported. However, the change greatly reduces the number of read calls needed for the 1MB sequence from about 400 to 8 (for small sequences there is little or no change) which sounds in line with what the OP reported.

droazen:
:+1: merging!

PR #480, opened 2016-02-17
squid:S2272 - Iterator.next() methods should throw NoSuchElementExcep…

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule squid:S2272 - "Iterator.next()" methods should throw "NoSuchElementException"

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:S2272

M-Ezzat


src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -343,6 +344,7 @@ public boolean hasNext() {
 
         @Override
         public SAMRecord next() {
+            if(!hasNext()) { throw new NoSuchElementException(); }
```
yfarjoun: unneeded here...

yfarjoun:
This needs a thorough inspection I think as some of the iterators are setup to return null when empty, and I do not know if the callers are using that instead of try/catch anywhere....

lbergelson:
@yfarjoun If that's true then the iterators are just implemented wrong.

yfarjoun:
I am not arguing with that, but a cursory glance seemed to indicate that
some iterators might be doing it anyway...

lbergelson:
@yfarjoun You're definitely right.  It's just an unpleasant thing to find out.

src/java/htsjdk/samtools/SamPairUtil.java
```diff
@@ -466,6 +467,7 @@ private void advance() {
         }
 
         public SAMRecord next() {
+            if(!hasNext()) { throw new NoSuchElementException(); }
```
yfarjoun: hasNext() here seems to throw an exception if the iterator is closed. Is this a reasonable thing to do? I would imagine that a better action would be to return false...not your fault, but an indication of the state of affairs. @alecw I think you added these assertions, do you remember why?

src/java/htsjdk/samtools/cram/build/CramContainerIterator.java
```diff
@@ -51,6 +52,7 @@ public boolean hasNext() {
 
     @Override
     public Container next() {
+        if(!hasNext()) { throw new NoSuchElementException(); }
```
yfarjoun: Here's another example of how iterators were broken, before this change, calling next() twice in a row would return null in the second time...

yfarjoun:
@m-ezzat you seemed to have miss some private classes. for example 
private class Boundary inside CramSpanContainerIterator, which does the heavy lifting for it.

yfarjoun:
also the anonymous class that extends IterableOnceIterator<String> inside IOUtil

yfarjoun:
nm, you did that one.

src/java/htsjdk/samtools/util/PeekableIterator.java
```diff
@@ -50,6 +51,7 @@ public boolean hasNext() {
 
     /** Returns the next object and advances the iterator. */
     public Object next() {
+        if(!hasNext()) { throw new NoSuchElementException(); }
         Object retval = this.nextObject;
         advance();
         return retval;
```
yfarjoun: peek() should probably also have this test.

yfarjoun:
MergingSamRecordIterator::next() also lacks hasNext()

src/java/htsjdk/samtools/SamPairUtil.java
alecw: Hi Yossi,

I’m not sure what assertion I added.  Can you clarify?

Having next() throw NoSuchElementException is merely conforming to the java.util.Iterator.next() contract.  Are you saying that PeekableIterator.hasNext() throws an exception if the iterator is closed?  I don’t see where that happens.  This is all I see for PeekableIterator.hasNext():


    public boolean hasNext() {
        return this.nextObject != null;
    }

I don’t see how this would throw an exception.  

Anyway, the CloseableIterator interface doesn’t specify what should happen when you call next() or hasNext() on a closed iterator.  IMHO, I think throwing an exception when a call is made on a closed iterator (other than closing again) seems like a desirable behavior.

-Alec

yfarjoun: Sorry @alecw I wasn't clear enough. I meant the assertOpen() in htsjdk.samtools.SAMRecord#hasNext() and htsjdk.samtools.SAMRecord#next()

shouldn't hasNext() simply return false if the iterator is closed, not throw an exception? And the exception being thrown in this case should be a NoSuchElementException rather than a AssertionError, no?  The only reason I mentioned you here was that you seemed to have added the AbstractBamIterator about two years ago, and together with that change, added the calls to assertOpen.

yfarjoun:
PeekIterator seems to rely on the fact that objects returned by the iterator will never be null. @lbergelson @tfenne @droazen  is that a reasonable assumption to make?

yfarjoun:
@lbergelson, I think I've been confused about the null-returning iterators. I've done an in-depth review and I do not see anything suspicious (except for the things I've stated specifically in the comments here)

src/java/htsjdk/samtools/SamPairUtil.java
alecw: Hi Yossi,

OK, I assume you're referring to AbstractBamIterator.assertOpen().  Let me
preface the following by saying that reasonable people can disagree on this
question.

The behavior of calling next() on an exhausted iterator is clear from the
Iterator interface -- it should throw NoSuchElementException.  The question
is: is a closed iterator an exhausted iterator?  I would argue that it is
not.  I would further argue that it is a bug to call next() or hastNext()
on a closed iterator.  An analogy would be calling a read() method on a
Reader or InputStream that has been closed.  I wouldn't expect that I would
get a result that indicated EOF; rather I would expect to get an exception
indicating that I tried to read from a closed InputStream.

When we started writing Picard, part of our design philosophy was to be
strict and throw an exception at anything that smelled funny.  Not everyone
agrees with this philosophy, and over the years we have received numerous
complaints that samtools accepts something that Picard does not, and many
times our response was that we were comfortable with the strict behavior.

Just my $0.02.

-Alec

yfarjoun: Thanks. Alec. As you know I'm not familiar with the history as much as you
are, which is why I asked. I agree that a closed iterator should not allow
any further reading, but it seems that it will even throw an exception on
hasNext(), which is what confused me.

droazen:
Closing -- this is problematic given that some `hasNext()` methods in htsjdk have side effects.

PR #482, opened 2016-02-18
Asynchronous read performance enhancements #78

d-cameron:
Moved BAM decompression to a dedicated thread if asyncio is enabled.
Added AsyncBufferedIterator class to allow for improved read performance
Fixed resource leaks from unclosed BlockCompressedInputStream objects

### DIFF
--- a/.classpath
+++ b/.classpath
@@ -3,8 +3,12 @@
 	<classpathentry kind="src" path="src/java"/>
 	<classpathentry kind="src" path="src/tests/java"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
-	<classpathentry kind="lib" path="lib/testng/testng-5.5-jdk15.jar"/>
 	<classpathentry kind="lib" path="lib/commons-jexl-2.1.1.jar"/>
 	<classpathentry kind="lib" path="lib/snappy-java-1.0.3-rc3.jar"/>
+	<classpathentry kind="lib" path="lib/testng/testng-6.9.10.jar"/>
+	<classpathentry kind="lib" path="lib/testng/jcommander-1.48.jar"/>
+	<classpathentry kind="lib" path="lib/ngs-java-1.2.2.jar"/>
+	<classpathentry kind="lib" path="lib/commons-compress-1.4.1.jar"/>
+	<classpathentry kind="lib" path="lib/apache-ant-1.8.2-bzip2.jar"/>
 	<classpathentry kind="output" path="bin"/>
 </classpath>
--- a/src/java/htsjdk/samtools/BAMFileReader.java
+++ b/src/java/htsjdk/samtools/BAMFileReader.java
@@ -52,7 +52,7 @@ class BAMFileReader extends SamReader.ReaderImplementation {
     private BinaryCodec mStream = null;
 
     // Underlying compressed data stream.
-    private final BlockCompressedInputStream mCompressedInputStream;
+    private BlockCompressedInputStream mCompressedInputStream;
     private SAMFileHeader mFileHeader = null;
 
     // One of these is populated if the file is seekable and an index exists
@@ -262,12 +262,20 @@ class BAMFileReader extends SamReader.ReaderImplementation {
     public void setEagerDecode(final boolean desired) { this.eagerDecode = desired; }
     
     public void close() {
+    	if (mCompressedInputStream != null) {
+    		try {
+				mCompressedInputStream.close();
+			} catch (IOException e) {
+				throw new SAMException("Exception closing compressed input stream.", e);
+			}
+    	}
         if (mStream != null) {
             mStream.close();
         }
         if (mIndex != null) {
             mIndex.close();
         }
+        mCompressedInputStream = null;
         mStream = null;
         mFileHeader = null;
         mIndex = null;
--- a/src/java/htsjdk/samtools/BAMFileReader.java
+++ b/src/java/htsjdk/samtools/BAMFileReader.java
@@ -262,13 +262,13 @@ class BAMFileReader extends SamReader.ReaderImplementation {
     public void setEagerDecode(final boolean desired) { this.eagerDecode = desired; }
     
     public void close() {
-    	if (mCompressedInputStream != null) {
-    		try {
-				mCompressedInputStream.close();
-			} catch (IOException e) {
-				throw new SAMException("Exception closing compressed input stream.", e);
-			}
-    	}
+        if (mCompressedInputStream != null) {
+            try {
+                mCompressedInputStream.close();
+            } catch (IOException e) {
+                throw new SAMException("Exception closing compressed input stream.", e);
+            }
+        }
         if (mStream != null) {
             mStream.close();
         }
--- a/src/java/htsjdk/samtools/SAMFileReader.java
+++ b/src/java/htsjdk/samtools/SAMFileReader.java
@@ -709,7 +709,7 @@ public class SAMFileReader implements SamReader, SamReader.Indexing {
         readBytes(stream, buffer, 0, buffSize);
         stream.reset();
         final byte[] magicBuf = new byte[4];
-        BlockCompressedInputStream bcis = new BlockCompressedInputStream(new ByteArrayInputStream(buffer), false);
+        final BlockCompressedInputStream bcis = new BlockCompressedInputStream(new ByteArrayInputStream(buffer), false);
         final int magicLength = readBytes(bcis, magicBuf, 0, 4);
         bcis.close();
         return magicLength == BAMFileConstants.BAM_MAGIC.length && Arrays.equals(BAMFileConstants.BAM_MAGIC, magicBuf);
--- a/src/java/htsjdk/samtools/SAMFileReader.java
+++ b/src/java/htsjdk/samtools/SAMFileReader.java
@@ -709,7 +709,9 @@ public class SAMFileReader implements SamReader, SamReader.Indexing {
         readBytes(stream, buffer, 0, buffSize);
         stream.reset();
         final byte[] magicBuf = new byte[4];
-        final int magicLength = readBytes(new BlockCompressedInputStream(new ByteArrayInputStream(buffer)), magicBuf, 0, 4);
+        BlockCompressedInputStream bcis = new BlockCompressedInputStream(new ByteArrayInputStream(buffer), false);
+        final int magicLength = readBytes(bcis, magicBuf, 0, 4);
+        bcis.close();
         return magicLength == BAMFileConstants.BAM_MAGIC.length && Arrays.equals(BAMFileConstants.BAM_MAGIC, magicBuf);
     }
 
--- a/src/java/htsjdk/samtools/SamStreams.java
+++ b/src/java/htsjdk/samtools/SamStreams.java
@@ -55,7 +55,7 @@ public class SamStreams {
         readBytes(stream, buffer, 0, buffSize);
         stream.reset();
         final byte[] magicBuf = new byte[4];
-        BlockCompressedInputStream bcis = new BlockCompressedInputStream(new ByteArrayInputStream(buffer), false);
+        final BlockCompressedInputStream bcis = new BlockCompressedInputStream(new ByteArrayInputStream(buffer), false);
         final int magicLength = readBytes(bcis, magicBuf, 0, 4);
         bcis.close();
         return magicLength == BAMFileConstants.BAM_MAGIC.length && Arrays.equals(BAMFileConstants.BAM_MAGIC, magicBuf);
--- a/src/java/htsjdk/samtools/SamStreams.java
+++ b/src/java/htsjdk/samtools/SamStreams.java
@@ -55,7 +55,9 @@ public class SamStreams {
         readBytes(stream, buffer, 0, buffSize);
         stream.reset();
         final byte[] magicBuf = new byte[4];
-        final int magicLength = readBytes(new BlockCompressedInputStream(new ByteArrayInputStream(buffer)), magicBuf, 0, 4);
+        BlockCompressedInputStream bcis = new BlockCompressedInputStream(new ByteArrayInputStream(buffer), false);
+        final int magicLength = readBytes(bcis, magicBuf, 0, 4);
+        bcis.close();
         return magicLength == BAMFileConstants.BAM_MAGIC.length && Arrays.equals(BAMFileConstants.BAM_MAGIC, magicBuf);
     }
 
--- a/src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
+++ b/src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
@@ -1,0 +1,218 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Daniel Cameron
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package htsjdk.samtools.util;
+
+import htsjdk.samtools.util.CloseableIterator;
+import htsjdk.samtools.util.CloserUtil;
+import htsjdk.samtools.util.Log;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Iterator that uses a background thread to perform read-ahead to improve
+ * throughput at the expense of increased latency.
+ * 
+ * @author Daniel Cameron
+ *
+ */
+public class AsyncBufferedIterator<T> implements CloseableIterator<T> {
+	private static final Log log = Log.getInstance(AsyncBufferedIterator.class);
+	/**
+	 * Count of threads created
+	 */
+	private static AtomicInteger threadsCreated = new AtomicInteger(0);
+	private final Thread reader;
+	private final ReaderRunnable readerRunnable;
+	private final AtomicReference<Throwable> ex = new AtomicReference<Throwable>(null);
+	private final Iterator<T> underlying;
+	private final BlockingQueue<List<T>> buffer;
+	private final int batchSize;
+	private boolean closeCalled = false;
+	/**
+	 * Current batch of records returned from the background thread. An empty
+	 * list indicates the background thread has encountered the end of the
+	 * underlying stream.
+	 */
+	private List<T> currentList = null;
+	private Iterator<T> currentBuffer = null;
+
+	/**
+	 * Creates a new iterator that traverses the given iterator on a background
+	 * thread
+	 * 
+	 * @param iterator
+	 *            iterator to traverse
+	 * @param bufferCount
+	 *            number of read-ahead buffers
+	 * @param batchSize
+	 *            ize of each read-ahead buffer. A larger batch size will
+	 *            increase both throughput and latency.
+	 */
+	public AsyncBufferedIterator(Iterator<T> iterator, int bufferCount, int batchSize) {
+		this(iterator, null, bufferCount, batchSize);
+	}
+
+	/**
+	 * Creates a new iterator that traverses the given iterator on a background
+	 * thread
+	 * 
+	 * @param iterator
+	 *            iterator to traverse
+	 * @param threadName
+	 *            background thread name
+	 * @param bufferCount
+	 *            number of read-ahead buffers
+	 * @param batchSize
+	 *            size of each read-ahead buffer. A larger batch size will
+	 *            increase both throughput and latency.
+	 */
+	public AsyncBufferedIterator(Iterator<T> iterator, String threadName, int bufferCount, int batchSize) {
+		if (iterator == null)
+			throw new IllegalArgumentException();
+		if (bufferCount <= 0 || batchSize <= 0)
+			throw new IllegalArgumentException("Buffer size must be at least 1.");
+		this.underlying = iterator;
+		this.buffer = new ArrayBlockingQueue<List<T>>(bufferCount);
+		this.batchSize = batchSize;
+		this.readerRunnable = new ReaderRunnable();
+		this.reader = new Thread(readerRunnable,
+				threadName == null ? getThreadNamePrefix() + threadsCreated.incrementAndGet() : threadName);
+		this.reader.setDaemon(true);
+		log.debug("Starting thread " + this.reader.getName());
+		this.reader.start();
+	}
+
+	protected String getThreadNamePrefix() {
+		return "AsyncBufferedIterator";
+	}
+
+	@Override
+	public void close() {
+		closeCalled = true;
+		try {
+			reader.interrupt();
+			buffer.clear(); // flush buffer so EOS indicator can be written if
+							// writer is blocking
+			reader.join();
+		} catch (InterruptedException ie) {
+			throw new RuntimeException("Interrupted waiting for background thread to complete", ie);
+		} finally {
+			CloserUtil.close(underlying);
+		}
+	}
+
+	@Override
+	public boolean hasNext() {
+		throwOnCallingThread();
+		if (closeCalled)
+			return false;
+		if (currentList == null || (currentList.size() > 0 && !currentBuffer.hasNext())) {
+			try {
+				currentList = buffer.take();
+				currentBuffer = currentList.iterator();
+			} catch (InterruptedException e) {
+				throw new RuntimeException("Error reading from background thread", e);
+			}
+			// rethrow any exceptions raised on the background thread while we
+			// were blocking on the next record
+			throwOnCallingThread();
+		}
+		return currentBuffer.hasNext();
+	}
+
+	@Override
+	public T next() {
+		if (hasNext())
+			return (T) currentBuffer.next();
+		throw new NoSuchElementException("next");
+	}
+
+	/**
+	 * Rethrows any exception encountered on the background
+	 * thread to the caller 
+	 */
+	private final void throwOnCallingThread() {
+		final Throwable t = this.ex.get();
+		if (t != null) {
+			if (t instanceof Error) {
+				throw (Error) t;
+			} else if (t instanceof RuntimeException) {
+				throw (RuntimeException) t;
+			} else {
+				throw new RuntimeException(t);
+			}
+		}
+	}
+
+	/**
+	 * Reads the given iterator, passing back records to the calling thread in batches
+	 */
+	private class ReaderRunnable implements Runnable {
+		public void run() {
+			boolean eosWritten = false;
+			try {
+				while (underlying.hasNext()) {
+					List<T> readAhead = new ArrayList<T>(batchSize + 1);
+					for (int i = 0; i < batchSize && underlying.hasNext(); i++) {
+						readAhead.add(underlying.next());
+					}
+					buffer.put(readAhead);
+					if (readAhead.size() == 0) {
+						eosWritten = true;
+					}
+				}
+			} catch (InterruptedException ie) {
+				// log.debug("Thread interrupt received - closing on background thread.");
+			} catch (Throwable t) {
+				ex.set(t);
+				throw new RuntimeException(t);
+			} finally {
+				Thread.interrupted();
+				try {
+					if (!eosWritten) {
+						buffer.put(new ArrayList<T>());
+					}
+				} catch (InterruptedException e2) {
+					log.warn("Thread interrupt received whilst writing end of stream indicator");
+				}
+			}
+		}
+	}
+
+	@Override
+	public void remove() {
+		throw new UnsupportedOperationException();
+	}
+
+	protected String getBackgroundThreadName() {
+		return this.reader.getName();
+	}
+}
--- a/src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
+++ b/src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
@@ -44,175 +44,166 @@ import java.util.concurrent.atomic.AtomicReference;
  *
  */
 public class AsyncBufferedIterator<T> implements CloseableIterator<T> {
-	private static final Log log = Log.getInstance(AsyncBufferedIterator.class);
-	/**
-	 * Count of threads created
-	 */
-	private static AtomicInteger threadsCreated = new AtomicInteger(0);
-	private final Thread reader;
-	private final ReaderRunnable readerRunnable;
-	private final AtomicReference<Throwable> ex = new AtomicReference<Throwable>(null);
-	private final Iterator<T> underlying;
-	private final BlockingQueue<List<T>> buffer;
-	private final int batchSize;
-	private boolean closeCalled = false;
-	/**
-	 * Current batch of records returned from the background thread. An empty
-	 * list indicates the background thread has encountered the end of the
-	 * underlying stream.
-	 */
-	private List<T> currentList = null;
-	private Iterator<T> currentBuffer = null;
+    private static final Log log = Log.getInstance(AsyncBufferedIterator.class);
+    /** Count of threads created */
+    private static AtomicInteger threadsCreated = new AtomicInteger(0);
+    private final Thread reader;
+    private final ReaderRunnable readerRunnable;
+    private final AtomicReference<Throwable> ex = new AtomicReference<Throwable>(null);
+    private final Iterator<T> underlying;
+    private final BlockingQueue<List<T>> buffer;
+    private final int bufferSize;
+    private boolean closeCalled = false;
+    /**
+     * Current buffer of records returned from the background thread. An empty
+     * list indicates the background thread has encountered the end of the
+     * underlying stream.
+     */
+    private List<T> currentList = null;
+    private Iterator<T> currentBuffer = null;
 
-	/**
-	 * Creates a new iterator that traverses the given iterator on a background
-	 * thread
-	 * 
-	 * @param iterator
-	 *            iterator to traverse
-	 * @param bufferCount
-	 *            number of read-ahead buffers
-	 * @param batchSize
-	 *            ize of each read-ahead buffer. A larger batch size will
-	 *            increase both throughput and latency.
-	 */
-	public AsyncBufferedIterator(Iterator<T> iterator, int bufferCount, int batchSize) {
-		this(iterator, null, bufferCount, batchSize);
-	}
+    /**
+     * Creates a new iterator that traverses the given iterator on a background
+     * thread
+     * 
+     * @param iterator iterator to traverse
+     * @param bufferCount number of read-ahead buffers
+     * @param bufferSize size of each read-ahead buffer. A larger size will increase both throughput and latency.
+     */
+    public AsyncBufferedIterator(final Iterator<T> iterator, final int bufferCount, final int bufferSize) {
+        this(iterator, null, bufferCount, bufferSize);
+    }
 
-	/**
-	 * Creates a new iterator that traverses the given iterator on a background
-	 * thread
-	 * 
-	 * @param iterator
-	 *            iterator to traverse
-	 * @param threadName
-	 *            background thread name
-	 * @param bufferCount
-	 *            number of read-ahead buffers
-	 * @param batchSize
-	 *            size of each read-ahead buffer. A larger batch size will
-	 *            increase both throughput and latency.
-	 */
-	public AsyncBufferedIterator(Iterator<T> iterator, String threadName, int bufferCount, int batchSize) {
-		if (iterator == null)
-			throw new IllegalArgumentException();
-		if (bufferCount <= 0 || batchSize <= 0)
-			throw new IllegalArgumentException("Buffer size must be at least 1.");
-		this.underlying = iterator;
-		this.buffer = new ArrayBlockingQueue<List<T>>(bufferCount);
-		this.batchSize = batchSize;
-		this.readerRunnable = new ReaderRunnable();
-		this.reader = new Thread(readerRunnable,
-				threadName == null ? getThreadNamePrefix() + threadsCreated.incrementAndGet() : threadName);
-		this.reader.setDaemon(true);
-		log.debug("Starting thread " + this.reader.getName());
-		this.reader.start();
-	}
+    /**
+     * Creates a new iterator that traverses the given iterator on a background
+     * thread
+     * 
+     * @param iterator iterator to traverse
+     * @param threadName background thread name
+     * @param bufferCount number of read-ahead buffers
+     * @param bufferSize size of each read-ahead buffer. A larger size will increase both throughput and latency.
+     */
+    public AsyncBufferedIterator(final Iterator<T> iterator, final String threadName, final int bufferCount, final int bufferSize) {
+        if (iterator == null) throw new IllegalArgumentException();
+        if (bufferCount <= 0) throw new IllegalArgumentException("Must use at least 1 buffer.");
+        if (bufferSize <= 0) throw new IllegalArgumentException("Buffer size must be at least 1 record.");
+        this.underlying = iterator;
+        this.buffer = new ArrayBlockingQueue<List<T>>(bufferCount);
+        this.bufferSize = bufferSize;
+        this.readerRunnable = new ReaderRunnable();
+        this.reader = new Thread(readerRunnable, threadName == null ? getThreadNamePrefix() + threadsCreated.incrementAndGet() : threadName);
+        this.reader.setDaemon(true);
+        log.debug("Starting thread " + this.reader.getName());
+        this.reader.start();
+    }
 
-	protected String getThreadNamePrefix() {
-		return "AsyncBufferedIterator";
-	}
+    protected String getThreadNamePrefix() {
+        return "AsyncBufferedIterator";
+    }
 
-	@Override
-	public void close() {
-		closeCalled = true;
-		try {
-			reader.interrupt();
-			buffer.clear(); // flush buffer so EOS indicator can be written if
-							// writer is blocking
-			reader.join();
-		} catch (InterruptedException ie) {
-			throw new RuntimeException("Interrupted waiting for background thread to complete", ie);
-		} finally {
-			CloserUtil.close(underlying);
-		}
-	}
+    @Override
+    public void close() {
+        closeCalled = true;
+        try {
+            reader.interrupt();
+            // flush buffer so EOS indicator can be written if writer is blocking
+            buffer.clear();
+            reader.join();
+        } catch (InterruptedException ie) {
+            throw new RuntimeException("Interrupted waiting for background thread to complete", ie);
+        } finally {
+            CloserUtil.close(underlying);
+        }
+    }
 
-	@Override
-	public boolean hasNext() {
-		throwOnCallingThread();
-		if (closeCalled)
-			return false;
-		if (currentList == null || (currentList.size() > 0 && !currentBuffer.hasNext())) {
-			try {
-				currentList = buffer.take();
-				currentBuffer = currentList.iterator();
-			} catch (InterruptedException e) {
-				throw new RuntimeException("Error reading from background thread", e);
-			}
-			// rethrow any exceptions raised on the background thread while we
-			// were blocking on the next record
-			throwOnCallingThread();
-		}
-		return currentBuffer.hasNext();
-	}
+    @Override
+    public boolean hasNext() {
+        throwOnCallingThread();
+        if (closeCalled) return false;
+        if (currentList != null && currentList.isEmpty()) {
+            // encountered end of stream
+            return false;
+        }
+        if (currentList == null || !currentBuffer.hasNext()) {
+            try {
+                currentList = buffer.take();
+                currentBuffer = currentList.iterator();
+            } catch (InterruptedException e) {
+                throw new RuntimeException("Error reading from background thread", e);
+            }
+            // rethrow any exceptions raised on the background thread while we
+            // were blocking on the next record
+            throwOnCallingThread();
+        }
+        return currentBuffer.hasNext();
+    }
 
-	@Override
-	public T next() {
-		if (hasNext())
-			return (T) currentBuffer.next();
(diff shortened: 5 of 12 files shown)


src/java/htsjdk/samtools/BAMFileReader.java
```diff
@@ -262,12 +262,20 @@ public BAMIndex getIndex() {
     public void setEagerDecode(final boolean desired) { this.eagerDecode = desired; }
     
     public void close() {
+    	if (mCompressedInputStream != null) {
+    		try {
+				mCompressedInputStream.close();
+			} catch (IOException e) {
+				throw new SAMException("Exception closing compressed input stream.", e);
+			}
+    	}
```
nh13: probably a comment for all your code, but would it be possible to convert every tab to four spaces to match the style of the rest of the code?

src/java/htsjdk/samtools/SAMFileReader.java
```diff
@@ -709,7 +709,9 @@ private boolean isBAMFile(final InputStream stream)
         readBytes(stream, buffer, 0, buffSize);
         stream.reset();
         final byte[] magicBuf = new byte[4];
-        final int magicLength = readBytes(new BlockCompressedInputStream(new ByteArrayInputStream(buffer)), magicBuf, 0, 4);
+        BlockCompressedInputStream bcis = new BlockCompressedInputStream(new ByteArrayInputStream(buffer), false);
```
nh13: `bcis` can be final

src/java/htsjdk/samtools/SamStreams.java
```diff
@@ -55,7 +55,9 @@ public static boolean isBAMFile(final InputStream stream)
         readBytes(stream, buffer, 0, buffSize);
         stream.reset();
         final byte[] magicBuf = new byte[4];
-        final int magicLength = readBytes(new BlockCompressedInputStream(new ByteArrayInputStream(buffer)), magicBuf, 0, 4);
+        BlockCompressedInputStream bcis = new BlockCompressedInputStream(new ByteArrayInputStream(buffer), false);
```
nh13: `final` here too.

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,218 @@
...
+ * throughput at the expense of increased latency.
+ * 
+ * @author Daniel Cameron
+ *
+ */
+public class AsyncBufferedIterator<T> implements CloseableIterator<T> {
+	private static final Log log = Log.getInstance(AsyncBufferedIterator.class);
+	/**
+	 * Count of threads created
+	 */
```
nh13: `/** Count of threads created */`

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,218 @@
...
+	 *            number of read-ahead buffers
+	 * @param batchSize
+	 *            size of each read-ahead buffer. A larger batch size will
+	 *            increase both throughput and latency.
+	 */
+	public AsyncBufferedIterator(Iterator<T> iterator, String threadName, int bufferCount, int batchSize) {
+		if (iterator == null)
+			throw new IllegalArgumentException();
+		if (bufferCount <= 0 || batchSize <= 0)
+			throw new IllegalArgumentException("Buffer size must be at least 1.");
```
nh13: buffer size *and* batch size

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,218 @@
...
+	 *            background thread name
+	 * @param bufferCount
+	 *            number of read-ahead buffers
+	 * @param batchSize
+	 *            size of each read-ahead buffer. A larger batch size will
+	 *            increase both throughput and latency.
+	 */
+	public AsyncBufferedIterator(Iterator<T> iterator, String threadName, int bufferCount, int batchSize) {
+		if (iterator == null)
+			throw new IllegalArgumentException();
```
nh13: could these go on the same line?

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,218 @@
...
+	 *            iterator to traverse
+	 * @param threadName
+	 *            background thread name
+	 * @param bufferCount
+	 *            number of read-ahead buffers
+	 * @param batchSize
+	 *            size of each read-ahead buffer. A larger batch size will
+	 *            increase both throughput and latency.
+	 */
+	public AsyncBufferedIterator(Iterator<T> iterator, String threadName, int bufferCount, int batchSize) {
```
nh13: parameters could be `final`

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,218 @@
...
+		return "AsyncBufferedIterator";
+	}
+
+	@Override
+	public void close() {
+		closeCalled = true;
+		try {
+			reader.interrupt();
+			buffer.clear(); // flush buffer so EOS indicator can be written if
+							// writer is blocking
```
nh13: put on previous line

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,218 @@
...
+		} finally {
+			CloserUtil.close(underlying);
+		}
+	}
+
+	@Override
+	public boolean hasNext() {
+		throwOnCallingThread();
+		if (closeCalled)
+			return false;
```
nh13: put on previous line

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,218 @@
...
+
+	/**
+	 * Reads the given iterator, passing back records to the calling thread in batches
+	 */
+	private class ReaderRunnable implements Runnable {
+		public void run() {
+			boolean eosWritten = false;
+			try {
+				while (underlying.hasNext()) {
+					List<T> readAhead = new ArrayList<T>(batchSize + 1);
```
nh13: my last `final` comment, but loading this into an IDE like `IntelliJ` should give you markers where you can add `final` and such.

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,218 @@
...
+					for (int i = 0; i < batchSize && underlying.hasNext(); i++) {
+						readAhead.add(underlying.next());
+					}
+					buffer.put(readAhead);
+					if (readAhead.size() == 0) {
```
nh13: use `isEmpty` instead.

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -507,6 +541,119 @@ private static boolean preambleEqual(final byte[] preamble, final byte[] buf, fi
...
+    }
+    
+	/**
+	 * Performs block decompression
+	 */
+	private class DecompressionRunnable implements Runnable {
+		/**
+		 * Number of block to eagerly decompress
+		 */
+		public static final int READ_AHEAD_DEPTH = 1;
```
nh13: How can `READ_AHEAD_DEPTH` changed elsewhere?  Should it be added to `Defaults`?

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
d-cameron: I've gone with bufferCount and bufferSize to hopefully make the meanings of these fields more obvious

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
d-cameron: I've updated this to be driven from Defaults.NON_ZERO_BUFFER_SIZE as it already drives a range of buffers from this single value.

d-cameron:
When tracking down unclosed BlockCompressedInputStreams (which now leak an thread a aysnc mode), I noticed that there are quite a few locations is htsjdk that leak file pointers. Quite a bit of the header parsing and file type determination code leak file handles because they don't close their file streams afterwards (AbstractVCFCodec.canDecodeFile is a typical example; it also explains why IGV doesn't unload tracks properly and I have to restart it whenever I modify a file). Is it worth raising a generic bug unclosed file handles in general?

lbergelson:
@d-cameron Good idea to file that bug.  It's something that should be easier to tackle systematically now that we're on java  8 with try-with-resources.

akiezun:
@nh13 @droazen can someone on our side confirm the performance improvements, ie the speedup numbers?

nh13:
@akiezun unfortunately I am going to be unavailable for the rest of the week to look at this.  If this can wait until next week I can try looking at it again.  Could someone un-assign me?

d-cameron:
Given the rather embarrassing performance regression (#479) from my last performance optimisation patch, my preference would be for a second party to verify that:
* async sequential access is indeed considerably faster
* async additional CPU overhead is minimal
* the use of seek() with asyncio enabled returns the correct records even when the background decompression thread is not idle
 * ie thread synchronization and buffer flushing has been implemented correctly
* the throughput penalty of async random access is minimal
 * eager decompression of unneeded blocks will increase CPU usage, but should have minimal effect on throughput. It's possible that on low-core machines throughput may well suffer and reducing the priority of the background decompression thread through Thread.setPriority() would reduce the number of wasted CPU cycles.
* the removal of BlockCompressedInputStream buffering (in both async and sync mode) when parsing headers does not cause a significant performance penalty even for very large headers on unbuffered input.
 * ie, the change to the unbuffered BlockCompressedInputStream constructor in SAMFileReader.java, SamStreams.java, TabixFeatureReader.java, AbstractVCFCodec.java
  * An extra thread for every file type check or header parse seemed excessive to I went with the least impact (from an API perspective) of changing to constructor used in these locations.

nh13:
@akiezun since this is likely a big win for work that you are doing, could you spare some time for someone on your team to help out?

akiezun:
sure, happy to. @d-cameron how do you recommend I test this? As in, what exact commandline do I need to run to measure performance?

d-cameron:
I used htsjdk.samtools.ValidateSamFileTest as a test harness to iterate over a 300Mb BAM file a few times and used the test duration as my performance benchmark.
You'll need to run with both -Dsamjdk.use_async_io=true and  -Dsamjdk.use_async_io=false to compare.
For the random access validation, I'd set up a test case that seeks() back and forward between reads at least a Mb away from each other a few thousand times, checking that the reads are the expected ones (this test would cover both performance and correctness).

If you repeat but pin the test hardness process a single core you should get a good estimate of the additional CPU overhead.

akiezun:
@d-cameron is it intentional that this PR is from master of your fork? It makes me uneasy to sit on a moving target like master

d-cameron:
There's nothing particularly meaningful in it, nor is it necessarily any less stable than a named branch. I created the fork for the purpose of contributing this patch so I just inherited master from samtools/htsjdk.

akiezun:
My testing results confirm that he proposed change does improve wall clock time by ~2x and the CPU overhead is minimal (ie it just takes 2 cores and fills them to 100% each)

Using a BAM file 1.6GB, only reading and eager decoding.

**Testing reading only with eager decoding
use_async_io=false  71, 35, 35, 35, 35 seconds (median 35)
use_async_io=true   22, 18, 19, 19, 25 seconds  (median 19)**

To measure CPU overhead I pinned it to 1 core (using taskset -c 0). Overhead is minimal.
use_async_io=false 35 seconds
use_async_io=true 36 seconds 

To compare this to current htsjdk master - on current master.

Testing reading only with eager decoding:
use_async_io=false 34 seconds
use_async_io=true  35 seconds

Testing reading only with eager decoding, pinned to 1 CPU
use_async_io=false 35 seconds
use_async_io=true 35 seconds

akiezun:
un-assigning myself -  someone should check the other conditions and carefully examine the code. @droazen @nh13 @tfenne who should it be? I think we should prioritize this PR given the 2x reading speedup it offers (in wallclock time).

akiezun:
@tomwhite Can you help review this code? @tfenne @nh13 @droazen FYI, please chime in if you have comments.

src/java/htsjdk/samtools/SAMFileReader.java
```diff
@@ -709,7 +709,9 @@ private boolean isBAMFile(final InputStream stream)
         readBytes(stream, buffer, 0, buffSize);
         stream.reset();
         final byte[] magicBuf = new byte[4];
-        final int magicLength = readBytes(new BlockCompressedInputStream(new ByteArrayInputStream(buffer)), magicBuf, 0, 4);
+        final BlockCompressedInputStream bcis = new BlockCompressedInputStream(new ByteArrayInputStream(buffer), false);
```
akiezun: use try-with-resources to guarantee that the stream is always closed
```
try(final BlockCompressedInputStream bcis = new BlockCompressedInputStream(new ByteArrayInputStream(buffer), false)){
   final int magicLength = readBytes(bcis, magicBuf, 0, 4);
   return magicLength == BAMFileConstants.BAM_MAGIC.length && Arrays.equals(BAMFileConstants.BAM_MAGIC, magicBuf);
}
```

src/java/htsjdk/samtools/SamStreams.java
```diff
@@ -55,7 +55,9 @@ public static boolean isBAMFile(final InputStream stream)
         readBytes(stream, buffer, 0, buffSize);
         stream.reset();
         final byte[] magicBuf = new byte[4];
-        final int magicLength = readBytes(new BlockCompressedInputStream(new ByteArrayInputStream(buffer)), magicBuf, 0, 4);
+        final BlockCompressedInputStream bcis = new BlockCompressedInputStream(new ByteArrayInputStream(buffer), false);
```
akiezun: same here - use try-with-resources

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package htsjdk.samtools.util;
+
+import htsjdk.samtools.util.CloseableIterator;
```
akiezun: these 3 imports are unused

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+/**
+ * Iterator that uses a background thread to perform read-ahead to improve
+ * throughput at the expense of increased latency.
+ * 
+ * @author Daniel Cameron
+ *
+ */
+public class AsyncBufferedIterator<T> implements CloseableIterator<T> {
+    private static final Log log = Log.getInstance(AsyncBufferedIterator.class);
+    /** Count of threads created */
```
akiezun: this comment is not redundant with the names of the field and can be removed

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+    private static AtomicInteger threadsCreated = new AtomicInteger(0);
```
akiezun: make this field final

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+    private final Thread reader;
+    private final ReaderRunnable readerRunnable;
+    private final AtomicReference<Throwable> ex = new AtomicReference<Throwable>(null);
```
akiezun: `new AtomicReference<>`

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Iterator that uses a background thread to perform read-ahead to improve
+ * throughput at the expense of increased latency.
+ * 
+ * @author Daniel Cameron
+ *
+ */
+public class AsyncBufferedIterator<T> implements CloseableIterator<T> {
```
akiezun: IMPORTANT: This code has been copied with modification from a GPL library (as I understand you wrote it): license https://github.com/PapenfussLab/gridss/blob/master/COPYING
code: https://github.com/PapenfussLab/gridss/blob/master/src/main/java/au/edu/wehi/idsv/util/AsyncBufferedIterator.java

I don't know if this is problem but please make it very clear in the class comment that this is the case while we review the implications of this.

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+    private static final Log log = Log.getInstance(AsyncBufferedIterator.class);
+    /** Count of threads created */
+    private static AtomicInteger threadsCreated = new AtomicInteger(0);
+    private final Thread reader;
```
akiezun: I'm not a fan of a variable called `reader` and being of type `Thread` (readers are a specific abstraction in Java and this variable is not one of them) - how about `readerThread`?

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+    private final ReaderRunnable readerRunnable;
```
akiezun: This class does not seem to need a name - I suggest `private final Runnable readerRunnable` and making the class anonymous

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+    private final AtomicReference<Throwable> ex = new AtomicReference<Throwable>(null);
+    private final Iterator<T> underlying;
```
akiezun: `underlying` is a strange name, seems to miss a noun, how about `underlyingIterator`?

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
d-cameron: Not a problem in this case. As I am indeed the author and copyright holder, I am free to relicence the code under a non-GPL licence, which I have explicitly done under an MIT licence in the file header of the contributed file.

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+     * Creates a new iterator that traverses the given iterator on a background
+     * thread
+     * 
+     * @param iterator iterator to traverse
+     * @param threadName background thread name
+     * @param bufferCount number of read-ahead buffers
+     * @param bufferSize size of each read-ahead buffer. A larger size will increase both throughput and latency.
+     */
+    public AsyncBufferedIterator(final Iterator<T> iterator, final String threadName, final int bufferCount, final int bufferSize) {
+        if (iterator == null) throw new IllegalArgumentException();
```
akiezun: add a description

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+        if (bufferCount <= 0) throw new IllegalArgumentException("Must use at least 1 buffer.");
+        if (bufferSize <= 0) throw new IllegalArgumentException("Buffer size must be at least 1 record.");
+        this.underlying = iterator;
+        this.buffer = new ArrayBlockingQueue<List<T>>(bufferCount);
```
akiezun: use the diamond types here and elsewhere, ie `new ArrayBlockingQueue<>(bufferCount)`

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+    public AsyncBufferedIterator(final Iterator<T> iterator, final int bufferCount, final int bufferSize) {
+        this(iterator, null, bufferCount, bufferSize);
+    }
+
+    /**
+     * Creates a new iterator that traverses the given iterator on a background
+     * thread
+     * 
+     * @param iterator iterator to traverse
+     * @param threadName background thread name
```
akiezun: mention that null is ok as a value for `threadName`, eg "background thread name or null if an auto-generated unique name is to be used" (or something like that)

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+        this.bufferSize = bufferSize;
+        this.readerRunnable = new ReaderRunnable();
+        this.reader = new Thread(readerRunnable, threadName == null ? getThreadNamePrefix() + threadsCreated.incrementAndGet() : threadName);
+        this.reader.setDaemon(true);
+        log.debug("Starting thread " + this.reader.getName());
+        this.reader.start();
+    }
+
+    protected String getThreadNamePrefix() {
+        return "AsyncBufferedIterator";
```
akiezun: how about `AsyncBufferedIterator.class.getSimpleName()` instead (for the ease of refactoring etc)?

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    @Override
+    public void close() {
```
akiezun: given that this class is not `final`, leaving this method as non-final is problematic i think. How about making the class `final`? I'm a big advocate for[ "design and document for inheritance or else prohibit it"](http://users.csc.calpoly.edu/~jdalbey/305/Assign/Bloch17.pdf)

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+            // were blocking on the next record
+            throwOnCallingThread();
+        }
+        return currentBuffer.hasNext();
+    }
+
+    @Override
+    public T next() {
+        if (hasNext()) {
+            return (T) currentBuffer.next();
```
akiezun: the cast to T is not needed here

d-cameron:
AsyncBufferedIterator coding style was designed to follow the style of the existing htsjdk.samtools.util.AbstractAsyncWriter code as closely as possible. Should I make the suggested changes that cause a deviation between the styles of the two files?

akiezun:
I see, I did not realize that. I would not have a problem with these two classes not following exactly the same style (and they already don't, eg threadsCreated is a `static volatile int` in `AbstractAsyncWriter` and `isClosed` is an `AtomicBoolean` etc)

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
akiezun: can you rename `ex` to something more meaningful?

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+    private final BlockingQueue<List<T>> buffer;
+    private final int bufferSize;
+    private boolean closeCalled = false;
+    /**
+     * Current buffer of records returned from the background thread. An empty
+     * list indicates the background thread has encountered the end of the
+     * underlying stream.
+     */
+    private List<T> currentList = null;
+    private Iterator<T> currentBuffer = null;
```
akiezun: do we need both the list and the list's iterator as fields here? If yes, add a comment as to why - the list seems to be used in only 1 method and may be eliminated I think

d-cameron:
In the threadsCreated case, I though it was better not to duplicate the race condition present in AbstractAsyncWriter, even if it is inconsequential.

You are correct that the closed signaling logic does indeed differ slightly, and reviewing the coding, it looks like the AbstractAsyncWriter approach is correct. If the AsyncBufferedIterator reader thread won't abort until it encounters an operation that throws InterruptedException which, over a computationally expensive non-blocking iterator with a large buffer size, could be a non-trivial amount of time. The AbstractAsyncWriter approach reduces this window to a single next() call which, whilst technically is still unbounded since the iterator is arbitrary, is better than having to fill the entire buffer.

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+            } catch (InterruptedException ie) {
+                // log.debug("Thread interrupt received - closing on background thread.");
+            } catch (Throwable t) {
+                ex.set(t);
+                throw new RuntimeException(t);
+            } finally {
+                Thread.interrupted();
+                try {
+                    if (!eosWritten) {
+                        buffer.put(new ArrayList<T>());
```
akiezun: use `Collections.emptyList()` and add a comment that this empty list indicates the end of the underlying stream

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+                        eosWritten = true;
+                    }
+                }
+            } catch (InterruptedException ie) {
+                // log.debug("Thread interrupt received - closing on background thread.");
+            } catch (Throwable t) {
+                ex.set(t);
+                throw new RuntimeException(t);
+            } finally {
+                Thread.interrupted();
```
akiezun: what does this call `Thread.interrupted()` do? Remove or comment why it's necessary

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -192,45 +203,45 @@ public int read(final byte[] buffer)
...
-        	available = available();    
-        	if(available == 0){
-        		// EOF
-        		done = true;
-        	}
-        }
-    	return buf.toString();
+            int linetmpPos = mCurrentOffset;
+            int bCnt = 0;
+            while((available-- > 0)){
```
akiezun: this line is so horrible `(available-- > 0)` - can you use this as a opportunity to make it cleaner?

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -507,6 +542,117 @@ private static boolean preambleEqual(final byte[] preamble, final byte[] buf, fi
...
+                    mOperations.take();
+                    BlockingQueue<DecompressedBlock> result = mResult;
+                    if (result != null) {
+                        DecompressedBlock decompressed = nextBlock();
+                        result.put(decompressed);
+                    }
+                }
+            } catch (InterruptedException e) {
+                // Reset interrupt status
+                Thread.interrupted();
```
akiezun: my understanding of the doc is that this call does not reset the interrupt status `Tests whether this thread has been interrupted.  The _interrupted status_ of the thread is unaffected by this method.`

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -614,14 +615,24 @@ private static void parseSingleAltAllele(List<Allele> alleles, String alt, int l
...
 
     public final static boolean canDecodeFile(final String potentialInput, final String MAGIC_HEADER_LINE) {
+    	FileInputStream fis = null;
+    	GZIPInputStream giz = null;
+    	BlockCompressedInputStream bcis = null;
         try {
-            return isVCFStream(new FileInputStream(potentialInput), MAGIC_HEADER_LINE) ||
-                    isVCFStream(new GZIPInputStream(new FileInputStream(potentialInput)), MAGIC_HEADER_LINE) ||
-                    isVCFStream(new BlockCompressedInputStream(new FileInputStream(potentialInput)), MAGIC_HEADER_LINE);
+        	fis = new FileInputStream(potentialInput);
```
akiezun: all three streams could be put in a single try-with-resources

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -507,6 +542,117 @@ private static boolean preambleEqual(final byte[] preamble, final byte[] buf, fi
...
+            mDecompressionThread.interrupt();
+        }
+    }
+    
+    /**
+     * Performs block decompression
+     */
+    private class DecompressionRunnable implements Runnable {
+        /**
+         * Number of block to eagerly decompress
```
akiezun: dangling comment

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -507,6 +542,117 @@ private static boolean preambleEqual(final byte[] preamble, final byte[] buf, fi
...
+         */
+        /**
+         * Each entry in this queue indicates a block to be decompressed. 
+         */
+        private BlockingQueue<Object> mOperations = new ArrayBlockingQueue<Object>(READ_AHEAD_BUFFERS);
```
akiezun: add a comment that the actual values in this queue are not important

akiezun: I'm not a fan of classes meddling in private members of other classes (even their own inner classes). Can you either encapsulate those or make them explicitly non-private (to lower the reader's surprise)?

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
d-cameron: I've previous raised #494 to address this in a more general context as there are a number of places that I don't touch that have the same problems and to do it properly requires a code review of all opened streams. Should I be performing these sort of bug fixes which are not directly related to the performance enhancements in this PR?

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -507,6 +542,117 @@ private static boolean preambleEqual(final byte[] preamble, final byte[] buf, fi
...
+        /**
+         * Results of block decompression.
+         */
+        private volatile BlockingQueue<DecompressedBlock> mResult = null;
+        private AtomicBoolean mTerminate = new AtomicBoolean(false);
+        public void decompressBlocks() {
+            try {
+                while (true) {
+                    mOperations.take();
+                    BlockingQueue<DecompressedBlock> result = mResult;
```
akiezun: can you just use the field here rather than copy the reference into a local var first? If not, can you add a comment why this is necessary?

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
akiezun: I don't see why not make them now that you're in the codebase. You've already fixed a whole bunch of resource leaks in this branch, which is great.

akiezun:
Done with part 1 of the review. Will resume tomorrow and try to wrap my head around the multi-threading issues. Hopefully @tomwhite and/or @tfenne can help too, this is tricky stuff.

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
akiezun: The `AsyncBufferedIterator` class seems unused outside of tests. Can we remove it?

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -507,6 +542,117 @@ private static boolean preambleEqual(final byte[] preamble, final byte[] buf, fi
...
+        /**
+         * Each entry in this queue indicates a block to be decompressed. 
+         */
+        private BlockingQueue<Object> mOperations = new ArrayBlockingQueue<Object>(READ_AHEAD_BUFFERS);
+        /**
+         * Results of block decompression.
+         */
+        private volatile BlockingQueue<DecompressedBlock> mResult = null;
+        private AtomicBoolean mTerminate = new AtomicBoolean(false);
+        public void decompressBlocks() {
```
akiezun: decompressBlocks should be private

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -507,6 +542,117 @@ private static boolean preambleEqual(final byte[] preamble, final byte[] buf, fi
...
+                        DecompressedBlock decompressed = nextBlock();
+                        result.put(decompressed);
+                    }
+                }
+            } catch (InterruptedException e) {
+                // Reset interrupt status
+                Thread.interrupted();
+            }
+        }
+        public void run() {
```
akiezun: add "@Override"

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
akiezun: nevermind - i was reading the wrong thing.

akiezun:
Looks fine to me so far. @d-cameron can you remove the `AsyncBufferedIterator` class and the tests for it? It's unused thus need not be reviewed or included in this PR. 
@tomwhite has been looking into blockgzip compression recently and I deref to him for opinion on the correctness of this change.

d-cameron:
I could, but the reason I included it is that I found, for me, it provides an additional speedup in BAM reading by separating the processing from the parsing of the SAM/BAM/CRAM file. When the per-record processing time is the same as the parsing time, program running time can be halved. For example, wrapping picard.analysis.SinglePassSamProgram.java:129 in an AsyncBufferedIterator improves the speed of many Picard tools with only a one-line change required. To do this, AsyncBufferedIterator need to exist in htsjdk.

The reason it is unused is because I wanted to minimise the impact of the PR. The design options are:

1. Current design: explicit wrapping by the user
  - Pros:
    - Greatest control over threading
    - Can use on any iterator to separate the iterator traversal from the processing
  - Cons:
    - Unused by default
2. Remove
  - Pros:
    - Simple PR, easier to review
    - Smaller codebase
  - Cons:
    - Loss of up to 100% additional speed improvement for some Picard tools
3. Add to key htsjdk iterators
  - Pros:
    - Automatically used
  - Cons:
    - Only works for specific iterators
    - Unclear exactly where it should be included

Take SamRecord as an example. Wrapping iterator() improves throughput when reading the whole file, but wrapping query() would be cause an excessive thread creation when used for random access.

In my usage, I found that were to async an interator depended on the workload. For example, in my usage, I read and write from multiple intermediate files per chromosome. In that scenario, I found that it a single async iterator wrapping the merged iterator, was sufficient. If I was only counting record, an async iterator per input file would have been more appropriate since the workload was much lower.

An alternate to the thread per iterator design is to back with a shared ThreadPool and use the ThreadPool to perform the I/O. This design would be more appropriate for inclusion in htsjdk iterators as each iterator would be more light-weight. Such as design would be best applied to all async IO classes and would be a relatively elegant way to limit the thread creation

akiezun:
I see, thanks @d-cameron for the explanation - very helpful. I'll put together some code to test how fast this works.

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -507,6 +542,117 @@ private static boolean preambleEqual(final byte[] preamble, final byte[] buf, fi
...
+         * Number of block to eagerly decompress
+         */
+        /**
+         * Each entry in this queue indicates a block to be decompressed. 
+         */
+        private BlockingQueue<Object> mOperations = new ArrayBlockingQueue<Object>(READ_AHEAD_BUFFERS);
+        /**
+         * Results of block decompression.
+         */
+        private volatile BlockingQueue<DecompressedBlock> mResult = null;
```
tomwhite: Rather than updating `mResult` on seek, it might be safer and cleaner to create a new `Runnable` (this would not mean a new thread if you used a fixed size thread pool created from `Executors`). Then the `BlockingQueue` could be `final`.

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
tomwhite: Agreed, this tripped me up too.

.classpath
```diff
@@ -3,8 +3,12 @@
...
 	<classpathentry kind="src" path="src/tests/java"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
-	<classpathentry kind="lib" path="lib/testng/testng-5.5-jdk15.jar"/>
 	<classpathentry kind="lib" path="lib/commons-jexl-2.1.1.jar"/>
 	<classpathentry kind="lib" path="lib/snappy-java-1.0.3-rc3.jar"/>
+	<classpathentry kind="lib" path="lib/testng/testng-6.9.10.jar"/>
+	<classpathentry kind="lib" path="lib/testng/jcommander-1.48.jar"/>
+	<classpathentry kind="lib" path="lib/ngs-java-1.2.2.jar"/>
+	<classpathentry kind="lib" path="lib/commons-compress-1.4.1.jar"/>
+	<classpathentry kind="lib" path="lib/apache-ant-1.8.2-bzip2.jar"/>
```
tomwhite: Are all these library changes needed?

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -507,6 +542,117 @@ private static boolean preambleEqual(final byte[] preamble, final byte[] buf, fi
...
+            mException = exception;
+        }
+    }
+    
+    private void startDecompressionThread() {
+        if (mDecompressionThread == null) {
+            mDecompressionRunnable = new DecompressionRunnable();
+            mDecompressionThread = new Thread(mDecompressionRunnable, "DecompressBGZF");
+            mDecompressionThread.setDaemon(true);
+            mDecompressionThread.start();
```
tomwhite: I would recommend using the Java `Executors` library for thread management.

tomwhite:
I added a couple of comments. I agree with the previous comment about using a shared thread pool - since using Java executors will simplify some of the code here.

src/tests/java/htsjdk/samtools/util/AsyncBufferedIteratorTest.java
```diff
@@ -0,0 +1,115 @@
...
+        @Override
+        public Integer next() {
+            return results[offset++];
+        }
+        public int consumed() {
+            return offset;
+        }
+    }
+    @Test
+    public void testWrapUnderlying() {
```
tomwhite: Test the `hasNext()` method too.

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,209 @@
...
+ */
+public class AsyncBufferedIterator<T> implements CloseableIterator<T> {
+    private static final Log log = Log.getInstance(AsyncBufferedIterator.class);
+    /** Count of threads created */
+    private static AtomicInteger threadsCreated = new AtomicInteger(0);
+    private final Thread reader;
+    private final ReaderRunnable readerRunnable;
+    private final AtomicReference<Throwable> ex = new AtomicReference<Throwable>(null);
+    private final Iterator<T> underlying;
+    private final BlockingQueue<List<T>> buffer;
```
tomwhite: Why is this buffering a list, rather than just the elements directly, i.e. why not `BlockingQueue<T>`?

d-cameron: The intent is to reduce the thread synchronisation overhead by transferring records in bulk. In my use, I just use two largish buffers as I do purely sequential SAMRecord access.

.classpath
d-cameron: They are needed to get htsjdk working in eclipse.

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
d-cameron: That results in a race condition. If a seek() is called when the old thread is still running and the thread stop times out, the orphaned thread could still be running. Creating a new BlockQueue ensures that even if the pre-seek() thread writes a record after the seek() has been called, this record be written to an unused BlockingQueue (this is why mResult is copied to a local variable).

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
d-cameron: A slow/misbehaving background processing thread may still be performing decompression when a seek() is called. This approach prevents the writing of an old pre-seek() result to mResult (since a new mResult is updated in the seek() method).

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
d-cameron: We need to differentiate between an empty list (end of stream), and a list iterator with no more records (need to get next list of records).

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
d-cameron: Also, BlockingQueue does not allow null.

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
d-cameron: Should  I expand the PR to include a thread pool for async writing as well?

d-cameron:
Is this back with me to update the PR with new changes? It seems like I should split this into multiple PRs: a straight-forward one with the multithreaded decompression that everyone seems relatively happy with, and an async iterator one that refactors async writing to use a thread pool as well. The former should be a straight-forward PR since the leaks are already merged as #526.

akiezun:
Yes splitting would be great. It's back with you. Thanks

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -88,12 +95,14 @@ public BlockCompressedInputStream(final File file)
         throws IOException {
         mFile = new SeekableFileStream(file);
         mStream = null;
+        mAllowBuffering = true;
 
     }
 
     public BlockCompressedInputStream(final URL url) {
```
akiezun: add a comment to indicate that see can also be used with this ctor

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -104,6 +113,7 @@ public BlockCompressedInputStream(final URL url) {
     public BlockCompressedInputStream(final SeekableStream strm) {
```
akiezun: and here, add a comment to indicate that see can also be used with this ctor

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -354,53 +366,76 @@ private static boolean isValidBlockHeader(final byte[] buffer) {
...
-        }
-        int count = readBytes(mFileBuffer, 0, BlockCompressedStreamConstants.BLOCK_HEADER_LENGTH);
-        if (count == 0) {
-            // Handle case where there is no empty gzip block at end.
-            mCurrentOffset = 0;
-            mBlockAddress += mLastBlockLength;
-            mCurrentBlock = new byte[0];
-            return;
+            throws IOException {
+        if (mAllowBuffering && Defaults.USE_ASYNC_IO) {
```
akiezun: switch to using `Defaults.USE_ASYNC_IO_FOR_SAMTOOLS`

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -354,53 +366,76 @@ private static boolean isValidBlockHeader(final byte[] buffer) {
...
+                throw new RuntimeException("Interrupted waiting for decompression thread", e);
+            }
+        } else {
+            mCurrentBlock = nextBlock();
         }
-        if (count != BlockCompressedStreamConstants.BLOCK_HEADER_LENGTH) {
-            throw new IOException("Premature end of file");
+        mCurrentOffset = 0;
+        if (mCurrentBlock.mException != null) {
+            if (mCurrentBlock.mException instanceof IOException) {
```
akiezun: extract to a meanningful method like `checkAndRethrow` or something like that

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -354,53 +366,76 @@ private static boolean isValidBlockHeader(final byte[] buffer) {
                 buffer[12] == BlockCompressedStreamConstants.BGZF_ID1 &&
                 buffer[13] == BlockCompressedStreamConstants.BGZF_ID2);
     }
-
+    
     private void readBlock()
```
akiezun: Add a comment what `readBlock` does (now that there is a `readBlock()` and `nextBlock()` we need better comments

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -354,53 +366,76 @@ private static boolean isValidBlockHeader(final byte[] buffer) {
...
+            if (blockLength < BlockCompressedStreamConstants.BLOCK_HEADER_LENGTH || blockLength > mFileBuffer.length) {
+                return new DecompressedBlock(blockAddress, blockLength, new IOException("Unexpected compressed block length: " + blockLength));
+            }
+            final int remaining = blockLength - BlockCompressedStreamConstants.BLOCK_HEADER_LENGTH;
+            final int dataByteCount = readBytes(mFileBuffer, BlockCompressedStreamConstants.BLOCK_HEADER_LENGTH, remaining);
+            mStreamOffset += dataByteCount;
+            if (dataByteCount != remaining) {
+                return new DecompressedBlock(blockAddress, blockLength, new FileTruncatedException("Premature end of file"));
+            }
+            final byte[] decompressed = inflateBlock(mFileBuffer, blockLength, !Defaults.USE_ASYNC_IO && mCurrentBlock != null ? mCurrentBlock.mBlock : null);
```
akiezun: here too, use `Defaults.USE_ASYNC_IO_FOR_SAMTOOLS`. Also, this conditional exception is kind of hidden at the end of the line - extract a new temp var, something like `bufferForReuse` maybe

akiezun: also, add a comment re: why the buffer cannot be reused in the multithreading context

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -354,53 +366,76 @@ private static boolean isValidBlockHeader(final byte[] buffer) {
...
         final int uncompressedLength = unpackInt32(compressedBlock, compressedLength-4);
-        byte[] buffer = mCurrentBlock;
-        mCurrentBlock = null;
-        if (buffer == null || buffer.length != uncompressedLength) {
-            try {
-                buffer = new byte[uncompressedLength];
-            } catch (final NegativeArraySizeException e) {
-                throw new RuntimeIOException("BGZF file has invalid uncompressedLength: " + uncompressedLength, e);
-            }
+        if (uncompressedLength < 0) {
```
akiezun: thanks for refactoring this code - catching `NegativeArraySizeException` was really nasty especially given that it's so easy to check before

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
akiezun: makes total sense. Can you add this as a comment? This is important to explain.

src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
akiezun: Rather than expanding, I'd rather shrink the PR to split it to changes in BlockCompressedInputStream.java+related and separate PR for AsyncBufferedIterator. Having small PRs would greatly simplify our life and enable faster merging.

Using a thread pool for async writing would be fabulous (see also #559 ) but definitely a separate PR.

akiezun:
@d-cameron Thanks again for this contribution - I've re-read the code and understand it much better now. Do you think you could split it?

d-cameron:
Async decompression now in #576 rebaselined to latest htsjdk master with feedback incorporated.

akiezun:
@d-cameron can you create a new PR for the new AsyncBufferedIterator+tests? Then we can close this PR in favor of #576 and the new one

d-cameron:
Now split into two independent PRs.

PR #473, opened 2016-02-12
adding test code coverage collection and report using jacoco

akiezun:
this adds a target to collect coverage and make a report. To run it, use `ant test-coverage-report` (report is generate in the `report/` folder)

This PR is an analogue to https://github.com/broadinstitute/picard/pull/404 in picard.

The current line coverage is 64% and some packages are a bit underrepresented, particularly `htsjdk.samtools.sra`
image

PR #484, opened 2016-02-19
Revert "adding test code coverage collection and report using jacoco"

akiezun:
Reverts samtools/htsjdk#473

Master is failing after the merge of #473. Let's revert and see if reverting helps.


droazen:
:+1: merging

PR #488, opened 2016-02-19
ValidateSamFile doesn't catch duplicated unmapped reads depending on their order

meganshand:
I found a case where ValidateSamFile should catch an error, but it currently passes with no error. As you can see in the test below, if there are two identical unmapped reads in the order of first read in pair, second read in pair, first read in pair, second read in pair, then the file passes ValidateSamFile with no errors. However, if the order is changed to first read in pair, first read in pair, second read in pair, second read in pair, then ValidateSamFile correctly marks it with the Mates Are Same End error. 

It seems that the order of the reads shouldn't matter and ValidateSamFile should return the same error for both of these cases, but if each read needs to be checked against all of the others regardless of order then that would be an expensive validation. I'm not sure what the best solution for this would be.


yfarjoun:
Thanks for this @meganshand,
Could you add the successful test as well?

meganshand:
@yfarjoun Sure, the test is now added.

yfarjoun:
looks good. thanks Megan. :+1:  please rebase-squash and merge.

yfarjoun:
Could you also file an issue in Picard for this. It can have a very similar description and should probably point to this PR. Also please note that this isn't due to the reads being unmapped. if we really try,we could map a read as primary to two different places and then it will still validate, incorrectly.

meganshand:
@yfarjoun I don't have write access to this repo. Could you merge please?

PR #489, opened 2016-02-19
Update VCF header line parser to handle escaping quotes and backslashes

cwhelan:
Since VCF 4.1, the specification has stated that within the 'Description' values in INFO header lines, double quote and backslash characters can be escaped with a backslash character. The htsjdk parser does not properly handle escape sequences in quoted header line values, and currently produces strange results when lines with internal quotes are parsed and re-written. This PR updates the VCF header line parser to handle escape sequences in quoted header line values.

https://github.com/samtools/htsjdk/pull/448 added backslash-escaping of quotes to values when they are written out, but did not modify the parser.

There is currently some ambiguity in the specification about where escaping should be allowed. Escaping is explicitly mentioned only in reference to the Description value of INFO lines. It seems to me that escaping should be allowed in other Description values. In particular, FILTER lines and GATK command line header lines are the most likely header lines to contain quotes in the current use cases I've seen. For example a filter that was applied to a site info field of type 'String' will often leave a FILTER header line with an internal double quote. 

Other tools like bcftools fail to validate htsjdk-written VCF files if, for example, they contain a FILTER header line, or 'GATK.CommandLine' header line, that has an unescaped internal quote.


src/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
```diff
@@ -83,13 +83,29 @@
...
+                } else if (c != '\\') {
+                    builder.append(c);
+                } else {
+                    escape = true;
+                }
             } else {
+                escape = false;
                 switch (c) {
                     case ('<') : if (index == 0) break; // if we see a open bracket at the beginning, ignore it
                     case ('>') : if (index == valueLine.length()-1) ret.put(key,builder.toString().trim()); break; // if we see a close bracket, and we're at the end, add an entry to our list
```
droazen: Is it possible to finish parsing the line and still be in the `inQuote == true` state?

src/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
```diff
@@ -83,13 +83,29 @@
...
+                if (escape) {
+                    builder.append(c);
+                    escape = false;
+                } else {
+                    inQuote = !inQuote;
+                }
             } else if ( inQuote ) {
-                builder.append(c);
+                if (escape) {
+                    builder.append(c);
```
droazen: Can you confirm that the way this treats `\n` and such is consistent with the spec?

src/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
droazen: Actually the spec says that the only supported escape sequences are `\\` and `\"`, but your code would strip the backslash in other cases as well.

droazen: Also, is it appropriate to do this for non-INFO lines?

src/tests/java/htsjdk/variant/vcf/VCFHeaderLineTranslatorUnitTest.java
```diff
@@ -0,0 +1,35 @@
...
+        Assert.assertEquals(values3.size(), 2);
+        Assert.assertEquals(values3.get("ID"), "ANNOTATION");
+        Assert.assertEquals(values3.get("Description"), "ANNOTATION \\= \"NA\" || ANNOTATION <= 0.01");
+
+        final Map<String,String> values4 = VCFHeaderLineTranslator.parseLine(VCFHeaderVersion.VCF4_2, "<ID=ANNOTATION,Description=\"ANNOTATION \\\\= \\\"NA\\\" || ANNOTATION <= 0.01\", Description2=\"foo\\\"bar\">", null);
+        Assert.assertEquals(values4.size(), 3);
+        Assert.assertEquals(values4.get("ID"), "ANNOTATION");
+        Assert.assertEquals(values4.get("Description"), "ANNOTATION \\= \"NA\" || ANNOTATION <= 0.01");
+        Assert.assertEquals(values4.get("Description2"), "foo\"bar");
+
```
droazen: Add test case for unclosed quotes that expects an exception.

droazen: Add test case(s) for `\` before chars other than `\` or `"`, after patching the code to no longer escape other chars.

src/tests/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
```diff
@@ -333,6 +333,14 @@ public void testVCFHeaderQuoteEscaping() throws Exception {
         final VCFSimpleHeaderLine addedHeaderLine = new VCFSimpleHeaderLine("GATKCommandLine.Test", attributes);
         originalHeader.addMetaDataLine(addedHeaderLine);
 
+        final VCFFilterHeaderLine originalCopyAnnotationLine1 = originalHeader.getFilterHeaderLine("ANNOTATION");
+        Assert.assertNotNull(originalCopyAnnotationLine1);
+        Assert.assertEquals(originalCopyAnnotationLine1.getGenericFieldValue("Description"), "ANNOTATION != \"NA\" || ANNOTATION <= 0.01", originalCopyAnnotationLine1.toString());
+
+        final VCFFilterHeaderLine originalCopyAnnotationLine2 = originalHeader.getFilterHeaderLine("ANNOTATION2");
+        Assert.assertNotNull(originalCopyAnnotationLine2);
+        Assert.assertEquals(originalCopyAnnotationLine2.getGenericFieldValue("Description"), "ANNOTATION with quote \" that is unmatched");
```
droazen: This test case will probably fail if you accept the changes above to handle unclosed quotes.

testdata/htsjdk/variant/HiSeq.10000.vcf
```diff
@@ -10,6 +10,8 @@
 ##FILTER=<ID=Indel,Description="Overlaps a user-input mask">
 ##FILTER=<ID=LowQual,Description="Low quality">
 ##FILTER=<ID=LowQual,Description="QUAL  50.0">
+##FILTER=<ID=ANNOTATION,Description="ANNOTATION != \"NA\" || ANNOTATION <= 0.01">
+##FILTER=<ID=ANNOTATION2,Description="ANNOTATION with quote \" that is unmatched">
```
droazen: Is this file used by tests for the legacy VCF3 codec? (since I think you only changed the VCF4 parser)

droazen:
First-pass review complete, back to @cwhelan

src/tests/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
cwhelan: Actually, it doesn't since it's escaped in the input file that was parsed (HiSeq.10000.vcf). I've updated the new file to be more clear, and pulled these tests out into their own test file, along with a couple of new ones.

src/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
cwhelan: The VCF spec only mentions escaping in the context of the Description tags for INFO lines. However, it seems like the rules should also apply to other quoted text fields in the header, and these are causing problems in the wild. For example, GATK's VariantFiltration will add header lines with unescaped quotes to VCF headers. Issues samtools/hts-specs#89 and samtools/hts-specs#125 both address this lack of clarity in the spec. Is it OK to add this escaping capability in the meantime?

cwhelan: Added an exception if the line ends with a quote open. Also made sure that the parser leaves the backslash if the character following is not a quote or a backslash.

src/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
cwhelan: I don't see anything in the spec that says what to do with a backslash preceeding a character other than double quote or backslash, so I've made the parser code not do anything special in that case.

src/tests/java/htsjdk/variant/vcf/VCFHeaderLineTranslatorUnitTest.java
cwhelan: done

cwhelan:
Thanks for the review. I've addressed the code-related comments. Please see my comment responses on how this relates to the spec. Again, the spec only talks about doing this escaping within "Description" values of INFO header lines, but that leaves other types of header lines broken. Hopefully issues samtools/hts-specs#89 and samtools/hts-specs#125 will resolve this in the next version of the spec. 

Back to you @droazen

src/test/java/htsjdk/variant/vcf/VCFHeaderLineTranslatorUnitTest.java
```diff
@@ -0,0 +1,64 @@
...
+        Assert.assertEquals(values5.size(), 2);
+        Assert.assertEquals(values5.get("ID"), "ANNOTATION");
+        Assert.assertEquals(values5.get("Description"), "ANNOTATION \\n with a newline in it");
+
+        // test with an unclosed quote
+        try {
+            final Map<String, String> values6 = VCFHeaderLineTranslator.parseLine(VCFHeaderVersion.VCF4_2, "<ID=ANNOTATION,Description=\"ANNOTATION \\n with a newline in it>", null);
+            Assert.fail("Should have thrown a TribbleException for having an unclosed quote in the description line");
+        } catch (TribbleException.InvalidHeader e) {
+            Assert.assertEquals("Your input file has a malformed header: Unclosed quote in header line value <ID=ANNOTATION,Description=\"ANNOTATION \\n with a newline in it>", e.getMessage());
```
droazen: Don't check the contents of the exception message -- this makes the test too brittle. It's enough to prove that `TribbleException.InvalidHeader` was thrown.

src/test/java/htsjdk/variant/vcf/VCFHeaderLineTranslatorUnitTest.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        // test with an escaped quote at the end
+        try {
+            final Map<String, String> values7 = VCFHeaderLineTranslator.parseLine(VCFHeaderVersion.VCF4_2, "<ID=ANNOTATION,Description=\"ANNOTATION \\n with a newline in it\\\">", null);
+            Assert.fail("Should have thrown a TribbleException for having an unclosed quote in the description line");
+        } catch (TribbleException.InvalidHeader e) {
+            Assert.assertEquals("Your input file has a malformed header: Unclosed quote in header line value <ID=ANNOTATION,Description=\"ANNOTATION \\n with a newline in it\\\">", e.getMessage());
```
droazen: Same here -- don't check the contents of the exception message in the test, just make sure `TribbleException.InvalidHeader` is thrown.

droazen:
:+1: Looks good now @cwhelan. Two remaining comments, then feel free to squash/merge.

cwhelan:
Updated tests to not make assertions about the content of InvalidHeader TribbleExceptions messages. Squashed and rebased.

Back to you @droazen

droazen:
:+1: merging!

PR #491, opened 2016-02-22
Small change to return false when testing to see if a file is an SRA …

tfenne:
…accession, if the SRA toolkit is unavailable.

This just caused me an hour or so of tail chasing.  I upgraded to the latest htsjdk build for a project, and then started getting exceptions from tools I had written that are processing SAM and BAM via `/dev/stdin` and `/dev/stdout`.  Previously it worked fine to not have the sra java library present at runtime _if_ you didn't want SRA support.  This patch makes it that way again by treating exceptions in testing a file for SRA-ness, as if the file is not an SRA accession.

### DIFF
--- a/src/java/htsjdk/samtools/SamReaderFactory.java
+++ b/src/java/htsjdk/samtools/SamReaderFactory.java
@@ -1,12 +1,5 @@
 package htsjdk.samtools;
 
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
-import java.util.Collections;
-import java.util.EnumSet;
-import java.util.zip.GZIPInputStream;
-
 import htsjdk.samtools.cram.ref.ReferenceSource;
 import htsjdk.samtools.seekablestream.SeekableStream;
 import htsjdk.samtools.sra.SRAAccession;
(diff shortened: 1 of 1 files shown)


droazen:
@a-nikitiuk or @kwrodarmer can you review?

tfenne:
@droazen FWIW the change is isolated to `SamReaderFactory`

droazen:
(This looks fine to me @tfenne -- if you don't hear from @a-nikitiuk or @kwrodarmer within the next day or two feel free to merge)

src/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -302,6 +302,15 @@ public SamReader open(final SamInputResource resource) {
             }
         }
 
+        /** Attempts to detect whether the file is an SRA accessioned file. If SRA support is not available, returns false. */
+        private boolean isSra(final File sourceFile) {
+            try {
+                return SRAAccession.isValid(sourceFile.getPath());
+            } catch (Exception e) {
```
nh13: `final`

a-nikitiuk:
I can add that having such wrapper function in this single place of code brings some inconsistency, since this code will be return false when there is a problem with checking validity of provided input file/accession, while calling SRAAccession.isValid directly will still throw an exception.

It can also hide some bugs (or make them very hard to diagnose), since even NullPointerException inherits from Exception class and we will still catch it and return false instead of seeing the actual problem.

tfenne:
@kwrodarmer & @a-nikitiuk : I don't disagree with your assessment, but for various projects it is necessary for me to exclude `ngs-java.jar`.  Having lost quite a bit of time dealing with the unintended consequences of the sra code in htsjdk, and knowing you're working on changing things, I'm not inclined to spend more time on a better solution.  I'm happy to have this particular change reverted in your upcoming PR so long as the result is that `htsjdk` continues to work for all SAM/BAM reading use cases even when `ngs-java` isn't present at runtime.

PR #493, opened 2016-02-22
make field 'long processed' volatile

jabbarish:
Make field 'long processed' volatile to provide atomic read in getCount()

According to Java Language Specification, "Programmers are encouraged to declare shared 64-bit values as volatile or synchronize their programs correctly to avoid possible complications."
https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.7

It is enough to make the field 'processed' volatile in AbstractProgressLogger to provide thread safety.

### DIFF
--- a/src/java/htsjdk/samtools/util/AbstractProgressLogger.java
+++ b/src/java/htsjdk/samtools/util/AbstractProgressLogger.java
@@ -12,13 +12,13 @@ import java.text.NumberFormat;
  * Concrete subclasses must provide the logger
  */
 abstract public class AbstractProgressLogger implements ProgressLoggerInterface {
-    protected final int n;
-    protected final String verb;
-    protected final String noun;
+    private final int n;
+    private final String verb;
+    private final String noun;
     private final long startTime = System.currentTimeMillis();
     private final NumberFormat fmt = new DecimalFormat("#,###");
     private final NumberFormat timeFmt = new DecimalFormat("00");
-    private volatile long processed = 0;
+    private long processed = 0;
     // Set to -1 until the first record is added
     private long lastStartTime = -1;
 
@@ -93,7 +93,7 @@ abstract public class AbstractProgressLogger implements ProgressLoggerInterface 
     }
 
     /** Returns the count of records processed. */
-    public long getCount() { return this.processed; }
+    public synchronized long getCount() { return this.processed; }
 
     /** Returns the number of seconds since progress tracking began. */
     public long getElapsedSeconds() { return (System.currentTimeMillis() - this.startTime) / 1000; }
--- a/src/java/htsjdk/samtools/util/AbstractProgressLogger.java
+++ b/src/java/htsjdk/samtools/util/AbstractProgressLogger.java
@@ -18,7 +18,7 @@ abstract public class AbstractProgressLogger implements ProgressLoggerInterface 
     private final long startTime = System.currentTimeMillis();
     private final NumberFormat fmt = new DecimalFormat("#,###");
     private final NumberFormat timeFmt = new DecimalFormat("00");
-    private long processed = 0;
+    private volatile long processed = 0;
     // Set to -1 until the first record is added
     private long lastStartTime = -1;
 


tfenne:
I don't think this is a good idea.  It looks like @lbergelson started to make `AbstractProgressLogger` synchronized, but the protection is not 100% since several instance variables are protected, not private.  @lbergelson could those be made private?

lbergelson:
So the code in `AbstractProgressLogger` is essentially just lifted out of the original `ProgressLogger`.  I didn't intend to make it any more or less synchronized than it was.  

Those variables are final constants.  I was under the impression that that wouldn't introduce any synchronization problems since they're essentially read only once constructed.  They could be made private if it's problematic.  It would be a very minor incompatibility since I suspect GATK is the only client out side of htsjdk for `AbstractProgressLogger`  and I don't think it even accesses those.

@tfenne It looks to me like this could be made safe either by making `processed` volatile or by making `getCount` synchronized.  I'm not sure I follow why you think it's a bad idea.  Multithreaded programming is admittedly not my greatest strength though, so it's likely I'm missing something though.

tfenne:
It's been a while since I've looked into this kind of thing in great depth, but my understanding is that the JVM can do a lot to eliminate synchronization overhead when things are uncontended _or_ accessible only from a single thread, but that volatile will still require cache flushes on every write (of which there are typically millions in progress logger).  That said, I'm not 100% sure of this.

My vote would be to make those variables private and the synchronize getCount() as we can be certain that'll have minimal impact, whereas we can't be certain of that with changing to `volatile` .

ronlevine:
@tfenne I think the volatile keyword should be fine. For a multiprocessor machine, the this will keep the local caches synchronized. For a nice explanation of volatile, look at http://tutorials.jenkov.com/java-concurrency/volatile.html.

lbergelson:
@tfenne Ron pointed out to me that while synchronizing `getCounts()` will prevent half written variables from being read, it may not provide a consistent view of `processed` across threads unless we explicitly make `processed` volatile as well.  

That article makes that point that if multiple threads are reading and writing the same variable you need to have those operations synchronized and make the underlying variables volatile.

ronlevine:
Synchronize will lock the `getCounts()` until the thread is done with it. Otherwise, a [race condition](http://tutorials.jenkov.com/java-concurrency/race-conditions-and-critical-sections.html) will occur and the threads will see different values for that variable.

jabbarish:
As far as I know, in this conrete case with counter 'processed' it is enough to make it volatile to provide correct work of getCount(). There is no need to sync it, race condition is more about concurrent writes when speaking about counters.  But to provide counter visibility and consistency for 64-bit values one either need to use synchronized or volatile (or Atomic).  Race condition also is the problem when reading complex object's state (i.e. reading multiple fields at once). Again, it's not the case with getCount() method - we only read one field's value.

On the other hand, I understand Tim's concerns about the performance.  May be, we could use another solution apperead in Java 8 -- LongAdder?  See this article: http://blog.takipi.com/java-8-longadders-the-fastest-way-to-add-numbers-concurrently/

jabbarish:
@tfenne :  and one more article about volatile's performance: https://brooker.co.za/blog/2012/09/10/volatile.html

tfenne:
@lbergelson @ronlevine - I'm not sure what you mean when you say `synchronized` may "may not provide a consistent view of processed across threads".  Synchronized does exactly this - the contract of synchronized is that any state modified when synchronized on an object/monitor will be visible to another other thread that synchronizes on the same object/monitor when it acquires the lock.  So you're right that it wouldn't guarantee that non-syncrhonized access would yield the correct value, but that's why you make the values private and ensure all access is synchronized.   I.e. the simplest path is making `getCount()` synchronized.

The upside, as I see it, is that with synchronized if all access is from a single thread, the JVM does not need to flush and re-acquire state from main memory each time, whereas with volatile it does.  And in this case volatile alone would be insufficient since `record()` accesses `this.processed` multiple times, and has other non-thread-safe variables in use, notably `this.fmt`.

ronlevine:
The other option is to make `processed` [atomic](https://docs.oracle.com/javase/tutorial/essential/concurrency/atomicvars.html).

lbergelson:
I'm just confused I guess.  My initial thought was that making the getter synchronized would be sufficient to make this safe.  @ronlevine made it sound like we'd need to make it both synchronized and volatile or we risked issues with threads seeing out of date information.  

@tfenne is supporting the synchronized is safe side of things.  

After reflection, it seems to me that synchronized wouldn't be a useful construct if it didn't guarantee a consistent view of data across multiple threads.  The documentation seems to support that:
> [... when a synchronized method exits, it automatically establishes a happens-before relationship with any subsequent invocation of a synchronized method for the same object. This guarantees that changes to the state of the object are visible to all threads.](https://docs.oracle.com/javase/tutorial/essential/concurrency/syncmeth.html)

I don't have any intuition about the relative performance of synchronized vs volatile.  My assumption would be that they would be similar since they have to perform similar operations.  I'm going to go with Tim's thought on the matter that synchronization potentially enables better optimization though.

Unless someone objects I think we should do the following:

@jabbarish Could you make `getCount()` synchronized, remove the volatile from `processed` and make the protected variables private instead?

ronlevine:
@lbergelson I still think it should also be volatile. Otherwise, the local processor caches could have different values. As @tfenne  pointed out, there will be a performance hit.

tfenne:
@ronlevine I don't understand the case for making it volatile if all access to it is synchronized.  I would politely suggest that if you or @jabbarish really want to have it be `volatile` that you a) synchronize access to `getCount()` and then submit a test case that fails with an inconsistent view unless the variable is also declared volatile.

lbergelson:
@ronlevine I read the oracle doc page that says that synchronized guarantees changes to that changes to the state of the object are visible to all threads as covering that case.  I assume it essentially treats everything within the synchronized block as being volatile.

jabbarish:
I drafted small testing class for different approaches.  Please check it and compare results.
Note that when we remove synchronized, we need to make **lastTime** field AtomicLong to make it threadsafe.  Maybe this will help us to choose correct solution.

package counters;

import java.time.Duration;
import java.time.Instant;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.TimeUnit;
import java.util.concurrent.atomic.AtomicLong;
import java.util.concurrent.atomic.LongAdder;

public class TestCounters {

	public interface Logger {
		void record();

		long get();
	}

	static class Runner implements Runnable {

		private static final int ITERATIONS = 10_000_000;
		private Logger logger;

		public void setLogger(Logger logger) {
			this.logger = logger;
		}

		@Override
		public void run() {
			Instant start = Instant.now();
			for (int i = 0; i < ITERATIONS; i++) {
				logger.record();
			}
			Instant stop = Instant.now();

			long elapsed = Duration.between(start, stop).toMillis();

			System.out.println(elapsed);

		}

	}

	static class SynchronizedLogger implements Logger {

		private long lastTime = System.currentTimeMillis();
		private long processed;
		private final int n;

		public SynchronizedLogger(int n) {
			this.n = n;
		}

		@Override
		synchronized public void record() {
			if (++processed % n == 0) {
				lastTime = System.currentTimeMillis();
			}
		}

		@Override
		public long get() {
			return processed;
		}

	}

	static class SynchronizedVolatileLogger implements Logger {

		private long lastTime = System.currentTimeMillis();
		volatile private long processed;
		private final int n;

		public SynchronizedVolatileLogger(int n) {
			this.n = n;
		}

		@Override
		synchronized public void record() {
			if (++processed % n == 0) {
				lastTime = System.currentTimeMillis();
			}
		}

		@Override
		public long get() {
			return processed;
		}

	}

	static class LongAdderLogger implements Logger {
		
		private AtomicLong lastTime = new AtomicLong(System.currentTimeMillis());
		LongAdder processed = new LongAdder();
		private final int n;
		
		public LongAdderLogger(int n) {
			this.n = n;
		}
		
		@Override
		public void record() {
			processed.increment();
			if ( processed.intValue() % n == 0) {
				lastTime.get();
				lastTime.set(System.currentTimeMillis());
			}
		}
		
		@Override
		public long get() {
			return processed.intValue();
		}
		
	}
	
	static class AtomicLogger implements Logger {
		
		private AtomicLong lastTime = new AtomicLong(System.currentTimeMillis());
		private AtomicLong processed = new AtomicLong();
		private final int n;
		
		public AtomicLogger(int n) {
			this.n = n;
		}
		
		@Override
		public void record() {
			if ( processed.incrementAndGet() % n == 0) {
				lastTime.get();
				lastTime.set(System.currentTimeMillis());
			}
		}
		
		@Override
		public long get() {
			return processed.intValue();
		}
		
	}
	
	public static void main(String[] args) throws InterruptedException {
		Runner runner = new Runner();

		for (int i = 0; i < 3; i++) {

			test(new LongAdderLogger(100_000), runner);
			test(new AtomicLogger(100_000), runner);
			test(new SynchronizedVolatileLogger(100_000), runner);
			test(new SynchronizedLogger(100_000), runner);
			
			System.out.println("---");
			
		}

	}
	
	static void test(Logger logger, Runner runner) throws InterruptedException {
		System.out.println(logger.getClass().getSimpleName() + ": ");
		ExecutorService service = Executors.newFixedThreadPool(4);
		runner.setLogger(logger);
		service.submit(runner);
		service.submit(runner);
		service.shutdown();
		service.awaitTermination(1, TimeUnit.HOURS);
		System.out.println(logger.get());
	}

}

jabbarish:
@lbergelson :  no, if value read is not synchronized as well (as change).
This code stucks because of changes visibility problem (get is not synchronized, as in out class):

package counters;

public class Visibility {
	
	static class Counter implements Runnable {
		
		private long count = 0;

		@Override
		public void run() {
			while (true) {
				inc();
			}
			
		}

		synchronized private void inc() {
			count++;
		}
		
		public long get() {
			return count;
		}
		
	}
	
	public static void main(String[] args) {
		Counter counter = new Counter();
		new Thread(counter).start();
		
		while (counter.get() < 1000000) {
		}
		
		System.out.println(counter.get());
	}

}

vadimzalunin:
wow!
i always thought volatile only guards against partial write, so for example a long value may be written as two ops: lower int and higher int. In this case the value is not guaranteed to be consistent (it may be corrupt) unless it is defined as volatile. 
and synchronization is for states of multiple variables. If this issue is about a single variable I'd say volatile will do.

jabbarish:
Another great article regarding volatile by Brian Goetz: http://www.ibm.com/developerworks/java/library/j-jtp06197/index.html
Scroll down there directly to **Pattern #5: The cheap read-write lock trick**.
This is exactly what we have.

Let me resume the info:  
1. volatile is the safe, but not the fastest solution in our concrete case. 
2. it is enought to make synchronized getCount() 
3. the fastest solution in multi-threaded environment is LongAdder for 'processed' and volatile for 'lastTime'
4. difference in time consumed between various versions will be within tens of seconds total on a big bam file

jabbarish:
@lbergelson I did what you asked to do: "make getCount() synchronized, remove the volatile from processed and make the protected variables private instead".  Please check it.

lbergelson:
@jabbarish :+1: Thanks for fixing this problem! I'll merge when tests pass.

tfenne:
Ditto - thanks @jabbarish !

PR #495, opened 2016-02-23
Modified VariantContextWriterBuilder to automatically set output type…

tfenne:
… to VCF_STREAM when the output file exists but is not a symlink, regular file or directory.

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
@@ -157,7 +157,7 @@ public class VariantContextWriterBuilder {
     public VariantContextWriterBuilder setOutputFile(final File outFile) {
         this.outFile = outFile;
         this.outStream = null;
-        determineOutputTypeFromFilename();
+        determineOutputTypeFromFilename(outFile);
         return this;
     }
 
@@ -171,7 +171,7 @@ public class VariantContextWriterBuilder {
     public VariantContextWriterBuilder setOutputFile(final String outFile) {
         this.outFile = new File(outFile);
         this.outStream = null;
-        determineOutputTypeFromFilename();
+        determineOutputTypeFromFilename(this.outFile);
         return this;
     }
 
@@ -465,16 +465,25 @@ public class VariantContextWriterBuilder {
         return writer;
      }
 
-    private void determineOutputTypeFromFilename() {
-        if (isBCF(this.outFile)) {
+    private void determineOutputTypeFromFilename(final File f) {
+        if (isBCF(f)) {
             this.outType = OutputType.BCF;
-        } else if (isCompressedVCF(this.outFile)) {
+        } else if (isCompressedVCF(f)) {
             this.outType = OutputType.BLOCK_COMPRESSED_VCF;
-        } else if (isVCF(this.outFile)) {
+        } else if (isVCF(f)) {
             this.outType = OutputType.VCF;
         }
         else {
-            this.outType = OutputType.UNSPECIFIED;
+            // See if we have a special file (device, named pipe, etc.)
+            final File canonical = new File(IOUtil.getFullCanonicalPath(this.outFile));
+            if (!canonical.equals(f)) {
+                determineOutputTypeFromFilename(canonical);
+            }
+            else if (f.exists() && !f.isFile() && !f.isDirectory()) {
+                this.outType = OutputType.VCF_STREAM;
+            } else {
+                this.outType = OutputType.UNSPECIFIED;
+            }
         }
     }
 
--- a/src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
@@ -157,7 +157,7 @@ public class VariantContextWriterBuilder {
     public VariantContextWriterBuilder setOutputFile(final File outFile) {
         this.outFile = outFile;
         this.outStream = null;
-        determineOutputTypeFromFilename(outFile);
+        this.outType = determineOutputTypeFromFile(outFile);
         return this;
     }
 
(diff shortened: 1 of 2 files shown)


tfenne:
@droazen Can you take a look at this one please?  My goal here is just to be able to stream VCF out of any Picard tool or similar that uses `VariantContextWriterBuilder` to create it's output.  This change should do the right thing both with redirects:
`java -jar picard.jar SortVcf I=foo.vcf O=/dev/stdout > /tmp/foo.vcf.gz`
and with pipes
`java -jar picard.jar SortVcf I=foo.vcf O=/dev/stdout | wc -l`

In the former case the conversion to canonical path will (at least on mac) result in finding `/tmp/foo.vcf.gz` and inferring the filetype from that(!).  In the latter, it will default to `VCF_STREAM`.

yfarjoun:
That's pretty cool!

src/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
```diff
@@ -445,16 +445,25 @@ else if (STREAM_TYPES.contains(this.outType))
...
             this.outType = OutputType.BLOCK_COMPRESSED_VCF;
-        } else if (isVCF(this.outFile)) {
+        } else if (isVCF(f)) {
             this.outType = OutputType.VCF;
         }
         else {
-            this.outType = OutputType.UNSPECIFIED;
+            // See if we have a special file (device, named pipe, etc.)
+            final File canonical = new File(IOUtil.getFullCanonicalPath(this.outFile));
+            if (!canonical.equals(f)) {
```
droazen: Can you document in *precisely* which cases this would be true (`canonical` not equal to `f`)? I understand that resolving a symlink could cause this, obviously, but could you list/explain any other cases you had in mind?

Also, am I wrong in thinking that a symlink to a `vcf` (for example) would typically be caught by the `isVCF()` check above, assuming the symlink itself ends in `.vcf`?

src/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
```diff
@@ -445,16 +445,25 @@ else if (STREAM_TYPES.contains(this.outType))
         return writer;
      }
 
-    private void determineOutputTypeFromFilename() {
-        if (isBCF(this.outFile)) {
+    private void determineOutputTypeFromFilename(final File f) {
```
droazen: Since this method now has some subtlety to it, I think it needs a good unit test. Recommend that you make this method `protected static` with an `OutputType` return value so that it can be independently tested, and move the assignment to `this.outType` into the constructors for the builder. Then add a quick unit test covering all the cases we care about (file, symlink, stream, etc.).

droazen:
Looks good to me @tfenne -- I've requested some extra explanatory docs + a unit test, and then we can merge this.

yfarjoun:
Hey @tfenne we'd really love to have this merged, but are concerned about the lack of a regression test...I think that your use-case in the commit message is a fine example, could you wrap that in a unit test?

yfarjoun:
@tfenne having happy thoughts about this PR in the review party. Would love to see this merged!

tfenne:
@yfarjoun Now with added tests.  Want to take a final peek?

src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
```diff
@@ -169,10 +169,7 @@ public VariantContextWriterBuilder setOutputFile(final File outFile) {
      * @return this <code>VariantContextWriterBuilder</code>
      */
     public VariantContextWriterBuilder setOutputFile(final String outFile) {
-        this.outFile = new File(outFile);
-        this.outStream = null;
-        determineOutputTypeFromFilename(this.outFile);
-        return this;
+        return setOutputFile(new File(outFile));
```
yfarjoun: I like this reuse of code...

src/test/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilderUnitTest.java
```diff
@@ -123,6 +122,26 @@ public void testSetOutputFile() throws IOException {
     }
 
     @Test
+    public void testDetermineOutputType() {
+        Assert.assertEquals(OutputType.VCF, VariantContextWriterBuilder.determineOutputTypeFromFile(this.vcf));
+        Assert.assertEquals(OutputType.BCF, VariantContextWriterBuilder.determineOutputTypeFromFile(this.bcf));
+        Assert.assertEquals(OutputType.VCF_STREAM, VariantContextWriterBuilder.determineOutputTypeFromFile(new File("/dev/stdout")));
```
yfarjoun: in the discussion you had mentioned `/tmp/foo.vcf.gz`. does that work? can you test it?

src/test/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilderUnitTest.java
```diff
@@ -123,6 +122,26 @@ public void testSetOutputFile() throws IOException {
...
+
+        // Test symlinking
+        try {
+            final Path link = Files.createTempFile("foo.", ".tmp");
```
yfarjoun: what happens if the symlink is purposefully confusing? I'm not sure I know what I want the answer to be, but I'd like the behaviour to be clear....so if `link.vcf` points to a `file.bcf` what happens?

yfarjoun:
back to you @tfenne

src/test/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilderUnitTest.java
yfarjoun: OK. Thanks.

src/test/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilderUnitTest.java
tfenne: I've added a test for `.vcf.gz`.  What I'd mentioned though was using redirects.  This was previously unsupported, and I'm not even really sure how to test it as I'd have to fork another process and use the shell to redirect it's output stream in order to test it.  I don't think that's worth the effort.

yfarjoun: specifically, is it worth testing the various BLOCK_COMPRESSED_EXTENSIONS for vcf as is done above?

yfarjoun:
👍

PR #501, opened 2016-02-24
squid:SwitchLastCaseIsDefaultCheck - switch statements should end wit…

m-ezzat:
This pull request is focused on resolving occurrences of Sonar squid:SwitchLastCaseIsDefaultCheck - "switch" statements should end with a "default" clause

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:SwitchLastCaseIsDefaultCheck

M-Ezzat

### DIFF
--- a/src/java/htsjdk/samtools/BAMFileReader.java
+++ b/src/java/htsjdk/samtools/BAMFileReader.java
@@ -927,6 +927,8 @@ class BAMFileReader extends SamReader.ReaderImplementation {
                     // Either found a good record, or else keep scanning SAMRecords
                     case OVERLAPPING: return
                             (contained ? FilteringIteratorState.CONTINUE_ITERATION : FilteringIteratorState.MATCHES_FILTER);
+                    default:
+                        break;
                 }
             }
             // Went past the last interval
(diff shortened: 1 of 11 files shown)


lbergelson:
Restarted tests.  This seems pointless to me.  Just adding ` default: break;` to these may be worse than leaving them default less.  It will hide warnings that might be generated if there were missing cases for a number of these.  Some of these would probably be improved by throwing exceptions if an unexpected case is encountered.  It's too bad java doesn't have compilation failures for switches with missing enum values and no default...

akiezun:
I think this is fine. Having an explicit default is cleaner. Let's merge this unless someone objects, wdyt?

yfarjoun:
I second @lbergelson's objection. if not all cases are dealt with, an IDE
could generate a warning (if someone adds an ENUM value in the future),
with a "default" case it will not.

:-1: from me.

akiezun:
ok, in that case, unless we find a champion for this PR, let's close.

lbergelson:
Closing this. Thanks @m-ezzat, but we're going to pass on this.

PR #505, opened 2016-02-29
Split the two FilteringIterators into two differently named Iterators

yfarjoun:
### Description
There are two very different FilteringIterators in htsjdk, SAM and VariantContext flavored ones. This is confusing and can lead to very verbose code when one wants to use both in the same class. 

This PR renames these two iterators FilteringSamIterator and FilteringVariantContextIterator while keeping the "old" FilteringIterators as @Deprecated but accessible subclasses. This makes it possible to remove these classes later, once folks have shifted over to using the explicitly named iterators.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/java/htsjdk/samtools/filter/FilteringIterator.java
+++ b/src/java/htsjdk/samtools/filter/FilteringIterator.java
@@ -1,38 +1,8 @@
-/*
- * The MIT License
- *
- * Copyright (c) 2009 The Broad Institute
- *
- * Permission is hereby granted, free of charge, to any person obtaining a copy
- * of this software and associated documentation files (the "Software"), to deal
- * in the Software without restriction, including without limitation the rights
- * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
- * copies of the Software, and to permit persons to whom the Software is
- * furnished to do so, subject to the following conditions:
- *
- * The above copyright notice and this permission notice shall be included in
- * all copies or substantial portions of the Software.
- *
- * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
- * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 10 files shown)


src/java/htsjdk/samtools/filter/FilteringIterator.java
```diff
@@ -1,154 +1,29 @@
...
-            } else if (!filter.filterOut(record)) {
-                return record;
-            }
-        }
-
-        return null;
-    }
-}
\ No newline at end of file
+package htsjdk.samtools.filter;
```
nh13: license

lbergelson:
If we're changing these should we extract out a functional interface to use here as well.  It would be nice if you could instantiate a filter as a lambda.

tfenne:
I should start with saying that I'm not against this PR.  But then I'm going to go on and suggest a bigger change that I think would be nicer and in line with what I think @lbergelson is suggesting.

It would be nice to consolidate things further and simply have:
```java
class FilteringIterator<T> extends Iterator { ... }
@FunctionalInterface interface Filter<T> { public boolean filterOut(T thing); }
```

The concrete filters and filtering iterators could be subclasses of these things, and have a completely uniform interface, etc.

lbergelson:
I retract my comment!   I failed to read correctly.

yfarjoun:
so @tfenne the only problem with this is that the sam filters need to know how to filter both individual sam records and pairs....that seems to be a bit of a sticking point...but if we can resolve that I'm in favor!

tfenne:
```java
class SamRecordFilter implements Filter<SAMRecord>, Filter<Tuple<SAMRecord,SAMRecord>>
``` 
?

yfarjoun:
Sure, that would work It should probably implement Predicate<> to keep with
standard Java notation, no?

Do you think it "needs" to be included in this PR, or just a nice to have
at some point?

tfenne:
Totally a nice to have.

yfarjoun:
Anyone feel inclined to give me a thumbs up here?

lbergelson:
:+1:

Issue #433, opened 2016-01-13
Implement support for creating CRAI files


cmnbroad:
Ideally, both from an existing CRAM and on-the-fly during CRAM writing. @vadimzalunin it looks like this is implemented in cramtools - do you know if any reason that implementation couldn't be moved to htsjdk ? I'm happy to do the work since we want it for htsjdk.

vadimzalunin:
I think it was agreed before that we focus on BAI rather than CRAI in htsjdk to avoid confusion. But I'm :+1 on full CRAI support.

GWarsow:
> Ideally, both from an existing CRAM file and on-the-fly during CRAM writing. @vadimzalunin it looks like this is implemented in cramtools

Creation of cram index files during CRAM writing is not yet supported in cramtools (neither bai nor crai). I commented on this just to avoid that others may get confused when they read above cited statement.

Best
Gregor

cmnbroad:
Update: there is now support for creating .crai (https://github.com/samtools/htsjdk/pull/538), but its not currently turned on by default due to https://github.com/samtools/htsjdk/issues/531.

tfenne:
Closing -- CRAI creation is implemented.

- From an existing CRAM: `CRAMCRAIIndexer.writeIndex(SeekableStream, OutputStream)`
- On the fly while writing: `CRAMCRAIIndexer` implements `CRAMIndexer` and can be handed to `CRAMContainerStreamWriter`/`CRAMFileWriter`

Verified by writing an indexed CRAM and querying it back through the generated index.

One related caveat, tracked separately in #1672: when you ask `SAMFileWriterFactory` for an index with `setCreateIndex(true)`, the *default* index emitted for a CRAM is still a BAI rather than a CRAI. The CRAI capability this issue asked for exists; which index the factory produces by default is the open question in #1672.

PR #510, opened 2016-03-02
CRAM index code/test refactoring.

cmnbroad:
Proposed refactoring of some existing CRAM indexing code and tests, both as prep for https://github.com/samtools/htsjdk/issues/433 and to better align the CRAM test classes with the code being tested; proposed removal of some unused code; some minor test bug fixes. No net new functionality.

Summary of changes:

- CRAMIndexer - Renamed to CRAMBAIIndexer to better reflect it's intent; all of the code is for generating BAI indices from a CRAM file (and some of the test code for this class is already in a file
called CRAMBAIIndexerTest).

- CRAIIndex - Removed two public methods (findLastAlignedEntry and getLeftMost) that were only referenced by their test code in CRAIIndexTest. (Are these useful to keep as public methods - htsjdk never uses .crai indices directly other than to convert them to .bai ?)

- SamIndexes - Propose to delete. The methods in this class are all for directly opening or discerning between .bai and .crai files/streams, but are not used by htsjdk to do this and are not used anywhere in htsjdk other than SamIndexesTest. Is it worth maintaining these as public methods ?

- SamIndexesTest - Propose to delete. Contains tests for the unused the methods in SamIndexes, with the exception of two tests that were using intermediate methods in SamIndexes to test CRAIIndex methods; those two tests have been moved to CRAIIndexTest and changed to call CRAIndex directly.

- CRAMComplianceTest - Removed some unnecessary/unused code (all tests remain) and renamed -> CRAMRoundTripTest, since all of the code is for testing round tripping between BAM and CRAM v2.1/v3.0.

- CramFileWriterTest - Renamed ->CRAMFileWriterTest to match the case-sense of the other files and the class it contains. Moved one general round-trip test into CRAMRoundTripTests.

- CRAMFileWriterWithIndexTest - Propose to delete. There are two tests in here; one was moved to CRAMFileIndexTest. The other one (testUnnecessaryIO) seems overly complicated and unnatural. I'm not sure what its testing for; is there a useful test here ?

### DIFF
--- a/src/java/htsjdk/samtools/cram/CRAIEntry.java
+++ b/src/java/htsjdk/samtools/cram/CRAIEntry.java
@@ -2,7 +2,10 @@ package htsjdk.samtools.cram;
 
 import htsjdk.samtools.cram.structure.Container;
 import htsjdk.samtools.cram.structure.Slice;
+import htsjdk.samtools.util.RuntimeIOException;
 
+import java.io.IOException;
+import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
@@ -20,35 +23,23 @@ public class CRAIEntry implements Comparable<CRAIEntry>, Cloneable {
     public int sliceSize;
     public int sliceIndex;
 
+    private static int CRAI_INDEX_COLUMNS = 6;
+    private static String entryFormat = "%d\t%d\t%d\t%d\t%d\t%d";
+
     public CRAIEntry() {
     }
 
-    public static List<CRAIEntry> fromContainer(final Container container) {
-        final List<CRAIEntry> entries = new ArrayList<CRAIEntry>(container.slices.length);
-        for (int i = 0; i < container.slices.length; i++) {
-            final Slice s = container.slices[i];
-            final CRAIEntry e = new CRAIEntry();
-            e.sequenceId = s.sequenceId;
-            e.alignmentStart = s.alignmentStart;
-            e.alignmentSpan = s.alignmentSpan;
-            e.containerStartOffset = s.containerOffset;
-            e.sliceOffset = container.landmarks[i];
-            e.sliceSize = s.size;
-
-            e.sliceIndex = i;
-            entries.add(e);
-        }
-        return entries;
-    }
-
-    public static CRAIEntry fromCraiLine(final String line) {
-        return new CRAIEntry(line);
-    }
-
+    /**
+     * Create a CRAI Entry from a serialized CRAI index line.
+     *
+     * @param line string formatted as a CRAI index entry
+     * @throws CRAIIndex.CRAIIndexException
+     */
     public CRAIEntry(final String line) throws CRAIIndex.CRAIIndexException {
         final String[] chunks = line.split("\t");
-        if (chunks.length != 6) {
-            throw new CRAIIndex.CRAIIndexException("Expecting 6 columns but got " + chunks.length);
+        if (chunks.length != CRAI_INDEX_COLUMNS) {
+            throw new CRAIIndex.CRAIIndexException(
+                    "Malformed CRAI index entry: expecting " + CRAI_INDEX_COLUMNS + " columns but got " + chunks.length);
         }
 
         try {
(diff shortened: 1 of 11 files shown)


cmnbroad:
@vadimzalunin Can you take a look at this and let me know if there is anything here you think we should keep ? I don't want to throw out anything useful, just clean up before I integrate the .crai writing code.

vadimzalunin:
@cmnbroad sounds good to me. 

CRAMComplianceTest - the goal was to test how htsjdk and htslib comply with each other. The CRAM files in the test have been generated with htslib. I don't mind renaming the class though.

CRAMFileWriterWithIndexTest.testUnnecessaryIO() - the goal was to ensure that only relevant regions of a CRAM file are read during search queries. For example the code might read too many bytes and then filter irrelevant records out. Normally this could be detected by bad performance on large files but this is difficult to achieve in tests. Moving/renaming the method is fine with me but I would prefer to keep the test unless there is a better way to test for unnecessary IO.

droazen:
Vote to keep (and start actually using) `SamIndexes`

cmnbroad:
Ok, I'll restore the I'm going to restore the testUnnecessaryIO test and revert the deletion of SamIndexes.

cmnbroad:
I've folded in the various comments I've received; the updated summary of changes is:

- CRAMIndexer - Renamed to CRAMBAIIndexer to reflect its content (all of the code is for generating BAI indices), to align it with the existing CRAMBAIIndexerTest, and to distinguish it from the soon-to-come CRAMCRAIIndexer
- CRAIIndex/CRAIndexTest - Removed two public methods (findLastAlignedEntry and getLeftMost) that were only referenced by their test code in CRAIIndexTest
- CRAMComplianceTest - Removed some unnecessary/unused code (all tests remain)
- CramFileWriterTest - Renamed ->CRAMFileWriterTest to match the case-sense of the other files and the class it contains.
- CRAIEntry - minor code cleanup

src/java/htsjdk/samtools/cram/CRAIEntry.java
```diff
@@ -63,13 +54,51 @@ public CRAIEntry(final String line) throws CRAIIndex.CRAIIndexException {
         }
     }
 
-    @Override
-    public String toString() {
-        return String.format("%d\t%d\t%d\t%d\t%d\t%d", sequenceId, alignmentStart, alignmentSpan,
+    /**
+     * Serializ the entry to a CRAI index stream.
```
droazen: Serializ -> Serialize

src/tests/java/htsjdk/samtools/CRAMComplianceTest.java
```diff
@@ -147,16 +136,15 @@ private void assertSameRecords(int majorVersion, SAMRecord record1, SAMRecord re
...
-                Assert.assertEquals(record2.getReadBases(), record1.getReadBases());
-        }
+        /**
+         * Known issue: CRAM v2.1 doesn't handle reads with missing bases correctly. This causes '*' bases to arise when reading CRAM.
+         * Skipping the base comparison asserts.
+         */
+        if (record1.getReadBases() == SAMRecord.NULL_SEQUENCE && majorVersion < CramVersions.CRAM_v3.major)
+            ;
+        else
+            Assert.assertEquals(record2.getReadBases(), record1.getReadBases());
```
droazen: Refactor to remove empty if clause, and add curly braces

src/tests/java/htsjdk/samtools/cram/CRAIIndexTest.java
```diff
@@ -77,57 +93,81 @@ private boolean allFoundEntriesIntersectQueryInFind(final List<CRAIEntry> index,
         return foundCount > 0;
     }
 
+    @Test(expectedExceptions = NullPointerException.class)
+    public void testCraiRequiresDictionary() throws IOException {
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        final GZIPOutputStream gos = new GZIPOutputStream(baos);
+        gos.close();
+        final BufferedInputStream bis = new BufferedInputStream(new ByteArrayInputStream(baos.toByteArray()));
+        CRAIIndex.openCraiFileAsBaiStream(bis, null);
```
droazen: Close `bis`

src/tests/java/htsjdk/samtools/cram/CRAIIndexTest.java
```diff
@@ -77,57 +93,81 @@ private boolean allFoundEntriesIntersectQueryInFind(final List<CRAIEntry> index,
...
+        final SeekableStream baiStream = CRAIIndex.openCraiFileAsBaiStream(new SeekableBufferedStream(new SeekableFileStream(file)), dictionary);
+        Assert.assertNotNull(baiStream);
+
+        final DiskBasedBAMFileIndex bamIndex = new DiskBasedBAMFileIndex(baiStream, dictionary);
+        final BAMFileSpan span = bamIndex.getSpanOverlapping(entry.sequenceId, entry.alignmentStart, entry.alignmentStart);
+        Assert.assertNotNull(span);
+        final long[] coordinateArray = span.toCoordinateArray();
+        Assert.assertEquals(coordinateArray.length, 2);
+        Assert.assertEquals(coordinateArray[0] >> 16, entry.containerStartOffset);
+        Assert.assertEquals(coordinateArray[1] & 0xFFFF, 1);
```
droazen: Looks like these two test cases share much of the same code -- can you extract a shared helper method?

droazen:
Review complete -- back to @cmnbroad. Merge after addressing comments.

PR #511, opened 2016-03-03
Fix setHeaderStrict(null) bug.

cmnbroad:
setHeaderStrict(null) isn't resetting the mHeader field, only the ref indices.


droazen:
:+1: Looks good -- merging!

PR #515, opened 2016-03-07
Fast BAM compression with igzip integrated as part of IntelDeflater

mishalinaik:
### Description
igzip provides fast BAM compression at level 1. This PR attempts to integrate igzip fast BAM compression as part of IntelDeflater to enable low latency BAM compression in htsjdk. PR can be summarized as follows:

1)	Modified IntelDeflater.c to add support for igzip API
2)     Added headers required by.igzip under src/c/inteldeflater
2)	Updated src/c/scripts/build_intel_deflater.sh to build libIntelDeflater.so with igzip.

### Checklist

- Code compiles correctly


akiezun:
This is an important one to review. @nh13 I believe you have worked with igzip, can you comment how your experience?

@mishalinaik can you clarify whether this code can be made part of this open source project (htsjdk uses an MIT license, I think you'd need to release your code under the same, though I Am Not A Lawyer).

I can take on the task of benchmarking the performance of this proposed change and I'll report here. Any objections to considering this contribution? (i'll assume no unless someone does object)

src/c/inteldeflater/IntelDeflater.c
```diff
@@ -57,12 +126,27 @@ Java_htsjdk_samtools_util_zip_IntelDeflater_initIDs(JNIEnv *env, jclass cls)
...
 }
 
 JNIEXPORT jlong JNICALL
-Java_htsjdk_samtools_util_zip_IntelDeflater_init(JNIEnv *env, jclass cls, jint level,
+Java_net_sf_samtools_util_zip_IntelDeflater_init(JNIEnv *env, jclass cls, jint level,
                                  jint strategy, jboolean nowrap)
 {
+
+  if (level == FAST_COMPRESSION && is_sse42_supported()) { //Use igzip
+    printf("Using igzip\n");
```
nh13: remove

src/c/inteldeflater/IntelDeflater.c
```diff
@@ -90,10 +174,11 @@ Java_htsjdk_samtools_util_zip_IntelDeflater_init(JNIEnv *env, jclass cls, jint l
             return jlong_zero;
         }
     }
+  }
 }
 
 JNIEXPORT void JNICALL
-Java_htsjdk_samtools_util_zip_IntelDeflater_setDictionary(JNIEnv *env, jclass cls, jlong addr,
+Java_net_sf_samtools_util_zip_IntelDeflater_setDictionary(JNIEnv *env, jclass cls, jlong addr,
```
nh13: Why are these all renamed to `net_sf_samtools` from `htsjdk_samtools`?  The package name in the java code is `htsjdk.samtools`.

src/c/inteldeflater/IntelDeflater.c
```diff
@@ -57,12 +126,27 @@ Java_htsjdk_samtools_util_zip_IntelDeflater_initIDs(JNIEnv *env, jclass cls)
...
+    useIGZIP = IGZIP_TRUE;
+    LZ_Stream2 *strm = calloc(1, sizeof(LZ_Stream2));
+    if (strm == 0) {
+      JNU_ThrowOutOfMemoryError(env, 0);
+      return jlong_zero;
+    } else {
+      init_stream(strm); //CHECK RETURN VALUE
+      return ptr_to_jlong(strm);
+    }
+
```
nh13: extra line

src/c/inteldeflater/IntelDeflater.c
```diff
@@ -116,20 +201,65 @@ Java_htsjdk_samtools_util_zip_IntelDeflater_setDictionary(JNIEnv *env, jclass cl
 }
 
 JNIEXPORT jint JNICALL
-Java_htsjdk_samtools_util_zip_IntelDeflater_deflateBytes(JNIEnv *env, jobject this, jlong addr,
+Java_net_sf_samtools_util_zip_IntelDeflater_deflateBytes(JNIEnv *env, jobject this, jlong addr,
                                          jarray b, jint off, jint len, jint flush)
 {
-    z_stream *strm = jlong_to_ptr(addr);
+  jarray this_buf = (*env)->GetObjectField(env, this, bufID);
```
nh13: indentation is two spaces here; I think it should be four throughout.

src/c/inteldeflater/IntelDeflater.c
```diff
@@ -116,20 +201,65 @@ Java_htsjdk_samtools_util_zip_IntelDeflater_setDictionary(JNIEnv *env, jclass cl
...
+    strm = jlong_to_ptr(addr);
+  //igzip only supports one compression level so setParamsID should not be set when using igzip 
+  //igzip does not support flush
+  if (useIGZIP && (((*env)->GetBooleanField(env, this, setParamsID) && strm->total_in != 0)  || flush == 1)) {
+    JNU_ThrowInternalError(env, "igzip doesn't support this");
+  } else if (useIGZIP) {
+    in_buf = (*env)->GetPrimitiveArrayCritical(env, this_buf, 0);
+    if (in_buf == NULL) {
+      // Throw OOME only when length is not zero
+      if (this_len != 0)
```
nh13: braces around `if`s make it much clearer, or put it on the same line throughout please

src/c/inteldeflater/IntelDeflater.c
```diff
@@ -116,20 +201,65 @@ Java_htsjdk_samtools_util_zip_IntelDeflater_setDictionary(JNIEnv *env, jclass cl
...
+    if (out_buf == NULL) {
+      (*env)->ReleasePrimitiveArrayCritical(env, this_buf, in_buf, 0);
+      if (len != 0)
+	JNU_ThrowOutOfMemoryError(env, 0);
+      return 0;
+    }
+    strm->next_in = (Bytef *) (in_buf + this_off);
+    strm->next_out = (Bytef *) (out_buf + off);
+    strm->avail_in = this_len;
+    strm->avail_out = len;
```
nh13: variable naming could be clearer.  `this_off` vs `off` and `this_len` vs `len`.

src/c/inteldeflater/IntelDeflater.c
```diff
@@ -116,20 +201,65 @@ Java_htsjdk_samtools_util_zip_IntelDeflater_setDictionary(JNIEnv *env, jclass cl
...
+    assert(strm->avail_in != 0);
+    assert(strm->avail_out != 0);
+    jboolean finish = (*env)->GetBooleanField(env, this, finishID);
+    if (finish) 
+      strm->end_of_stream = 1;
+    else
+      strm->end_of_stream = 0;
+
+    fast_lz(strm);
```
nh13: just curious, if `lz_stream.avail_in > 0` here does that mean that there was not enough room in the block for the compressed bytes?  If so, perhaps it should default to using `zlib`?  See a similar thought here: https://github.com/nh13/pbgzip/blob/master/consumer.c#L297

src/c/inteldeflater/IntelDeflater.c
```diff
@@ -116,20 +201,65 @@ Java_htsjdk_samtools_util_zip_IntelDeflater_setDictionary(JNIEnv *env, jclass cl
...
-    jint this_len = (*env)->GetIntField(env, this, lenID);
-    jbyte *in_buf;
-    jbyte *out_buf;
+    (*env)->ReleasePrimitiveArrayCritical(env, b, out_buf, 0);
+    (*env)->ReleasePrimitiveArrayCritical(env, this_buf, in_buf, 0);
+    if (finish)
+      (*env)->SetBooleanField(env, this, finishedID, JNI_TRUE);
+    this_off += this_len - strm->avail_in;
+    (*env)->SetIntField(env, this, offID, this_off);
+    (*env)->SetIntField(env, this, lenID, strm->avail_in);
```
nh13: should you check that `lz_stream.total_out` plus the header and footer lengths are not greater than the buffer/blocks size.  See a similar thought here: https://github.com/nh13/pbgzip/blob/master/consumer.c#L302

src/c/inteldeflater/types.h
```diff
@@ -0,0 +1,46 @@
...
+otherwise. Any license under such intellectual property rights must be
+express and approved by Intel in writing.
+ 
+Unless otherwise agreed by Intel in writing, you may not remove or alter this
+notice or any other notice embedded in Materials by Intel or Intel's suppliers
+or licensors in any way.
+*/
+
+/* NOTICE:
+ * Usage of this header file in new projects is deprecated.
```
nh13: is this still needed?

nh13:
@akiezun I found `igzip` very useful, enough so that I incorporated into my parallel-block gzip tool [`pbgzip`](https://github.com/nh13/pbgzip).  I tried my best to highlight some "gotchas" that I found when using it.  Also, I think since there a so many changes to the C code, it would be a great time to clean it up (function naming, consistent if/else statements, local variable naming, etc.), but I will leave it up to you to decide if that is necessary.  Finally, please make sure that when you run a tool that uses this branch, like Picard, it uses the intel deflator properly (Picard up-front states if it is using the intel deflator or jdk native deflator).

tfenne:
I guess I'm a little curious about this. We had some discussion recently about how things work in a Java 8+ world, and whether the custom code around the IntelDeflator was still necessary now that we think the JDK delegates internally to the platform's zlib library.  Our hope had been to remove the custom code around IntelDeflator, and rely on the fact that when it was present on a system, it overrode the standard zlib, and would be picked up by default with a Java 8+ VM.

So, what I'm wondering is:

1. Can we verify if that assumption is true?
2. If it is, would `igzip` also be picked up automatically if present on the library path, or does this require custom hooks?
3. If 1 is true, but igzip would continue to require custom hooks, is it worth the burden of maintaining custom JNI bindings?

I don't mean this to be a :-1: on this PR - I don't have any experience with `igzip` or anything against it.  I'd just like to understand the implications for the IntelDeflator/JNI code in htsjdk so we can make an informed decision.

src/c/inteldeflater/IntelDeflater.c
mishalinaik: Nils, 
Yes, it needs to be htsjdk_samtools, overwrote with a stale version before the commit hence the mismatch. Fixed in the new commit.
- Mishali

mishalinaik:
@akiezun @nh13 

Incorporated the changes requested in the latest commit.

The igzip code has been released under MIT license (https://software.intel.com/en-us/articles/igzip-a-high-performance-deflate-compressor-with-optimizations-for-genomic-data)

nh13:
@mishalinaik I had a few questions in the outdated diff, but didn't see any answers.
1. question about `lz_stream.avail_in > 0`.
2. question about `lz_stream.total_out`.
Thanks for getting this together.

mishalinaik:
@nh13 
1) I handle the case lz_stream.avail_in > 0 with this line 'this_off += this_len - strm->avail_in'. this_off and strm->avail_in which are in turn returned by setting the field in JNIEnv *env. So you don't need to fall back on zlib. 
2) The Deflater itself doesn't have visibility into the header and footer its works at the granularity of a stream buffer; this applies to any Deflater in general.

nh13:
@mishalinaik thank-you, very informative!

akiezun:
@mishalinaik when I run this I see a warning I do not see when running on master
```
Java HotSpot(TM) 64-Bit Server VM warning: You have loaded library /dsde/working/akiezun/tmp/testing/samtools/htsjdk/lib/jni/libIntelDeflater.so which might have disabled stack guard. The VM will try to fix the stack guard now.
It's highly recommended that you fix the library with 'execstack -c <libfile>', or link it with '-z noexecstack'.
```

Can you fix this?

akiezun:
I can confirm that this change speeds up compression at level 1 (but not at level 5)

I'm running this on a file ~800MB 

(compression level 5):
```
new IntelDeflater   73, 66, 64, 61, 65 seconds. Median 65
old IntelDeflater   62, 62, 70, 64, 62 seconds. Median 62 
JdkDeflater         79, 79, 81, 80, 80 seconds. Median 80
```

(compression level 1)
```
new IntelDeflater   43, 35, 31, 37, 44 seconds Median 37 
old IntelDeflater   54, 50, 49, 42, 48 seconds Median 49
JdkDeflater         49, 49, 51, 51, 58 seconds Median 51  
```

These numbers are certainly impressive.

eitanbanks:
But compression level 1 is not the recommended compression level, right?  Don't we want to optimize for the standard use case?

akiezun:
for any intermediate files it does not matter which level you have, does it? For final files yes but for intermediates we're optimizing for speed. Using igzip gives you speed for this case.

nh13:
Can you benchmark it against Snappy, since that's what is used in [SortingCollection](https://github.com/samtools/htsjdk/blob/master/src/java/htsjdk/samtools/util/SortingCollection.java)?

akiezun:
See also this article http://www.htslib.org/benchmarks/zlib.html

tfenne:
@eitanbanks : What @akiezun said.  For intermediate files that you can't stream between steps you're often trying to find the balance point between burning CPU to compress the temporary files and the amount of time it takes to read/write the files to storage.  If the storage is network based (e.g. NFS, S3, etc.) then going from level=0 to level=1 can make the overall time quite a bit shorter (and put less load on a shared resource in the case of NFS), whereas going up from level=1 yields fairly marginal reductions in size that are useful only for long-term storage of files, but at significant CPU cost.

mishalinaik:
@akiezun: Fixed the warning you are seeing. igzip has been optimized for compression level 1, new IntelDeflater only uses igzip at compression level 1 and continue.

akiezun:
I like this contribution and would like to merge - I like the speedup! My only reservation is that we can't make the so file ourselves (the build script does not work) and that it does not work on non linux, eg mac. A potential way out (maybe longer term) would be to have the whole deflater (java, jni and all) be moved to a separate project at Intel which htsjdk would then include like any other library. 

What do others think? Any objections to merging as is (as a drop-in replacement for the current deflater)?

pgrosu:
Maybe this is a naive question, but if I use [pbzip2](http://compression.ca/pbzip2/) - which is a parallel implementation of bzip2 - I seem to be faster than all of the benchmarks using a randomly generated 7 GB file:

```Bash
$ for i in {1..7}; do $(openssl rand -base64 $(( 2**30 * 3/4 )) >> 7gb-file.txt) ; done
$
$ du -h 7gb-file.txt
7.2G    7gb-file.txt
$
$ time( pbzip2 7gb-file.txt )

real    1m25.203s
user    21m50.263s
sys     0m47.491s

$
$ du -h 7gb-file.txt.bz2
5.4G    7gb-file.txt.bz2
$
```
Am I misinterpreting something?

Thanks,
~p

droazen:
Recommend adding a disclaimer at the top of `build_intel_deflater.sh` stating that the script can't actually be used to build the library without proprietary software, and also add a README to the directory with the C source files stating the same thing.

akiezun:
@mishalinaik can you add the information that @droazen is requesting and squash your commits? Then I'll merge the PR

mishalinaik:
@akiezun I have incorporated changes suggested by @droazen and squashed the commits.

akiezun:
Thank you

PR #516, opened 2016-03-08
Add optional logging in duplicate set iterator.

nh13:
Sorting the inputs may take quite some time, and so it is useful for a
user of the iterator to have logging and progress information to monitor
its progress.

@tfenne @yfarjoun

### DIFF
--- a/src/java/htsjdk/samtools/DuplicateSetIterator.java
+++ b/src/java/htsjdk/samtools/DuplicateSetIterator.java
@@ -24,6 +24,8 @@
 package htsjdk.samtools;
 
 import htsjdk.samtools.util.CloseableIterator;
+import htsjdk.samtools.util.Log;
+import htsjdk.samtools.util.ProgressLogger;
 import htsjdk.samtools.util.SortingCollection;
 
 import java.io.File;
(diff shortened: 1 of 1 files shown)


yfarjoun:
Personally, I'm happy with an info-level logging of the sorting. We've also been annoyed by the radio silence of the sort (not only in the duplicate-set-iterator) and so I think that you can spare the api changes and simply get the log object and log the progress at INFO-level. @tfenne, what do you think?

tfenne:
I'm happy with either approach, and just to muddy the waters will suggest a third option ;)  How about:

1. `DuplicateSetIterator` has a private instance `ProgressLogger` with a default value that will do logging every 1m reads
2. No changing the constructors
3. A `withProgress(ProgressLogger logger)` method or similar that allows callers to set a custom progress logger, or null if they want no logging, and returns `this` for easy chaining?

But really, I'm fine with any of the approaches.

yfarjoun:
I'm happy as long as there's no change in api, and we get logging on by
default. :-)

nh13:
@yfarjoun Pass a log, and you get logging. Could you review?

eitanbanks:
@yfarjoun is on vacation and he's not supposed to be doing work this week!
If you'll accept my review, I do have 1 request:
the comments in the code for the main constructor don't say anything about the logging only applying to the upfront sorting.  As a user, I would have expected the logging to apply across all the work in this class.  So could you please add to the comments for the constructor that the logging only applies to the sorting?

nh13:
@eitanbanks added a comment as requested, merging.

nh13:
@eitanbanks silly SRA errors: `Invalid SRA accession was passed to SRA reader`.

nh13:
Blocked by https://github.com/samtools/htsjdk/issues/545

PR #519, opened 2016-03-09
Added an example PrintReads

akiezun:
### Description

I made a simple PrintReads program that illustrates how to use the readers and writers and also is very useful for benchmarking changes such as #515 or #78 

It's useful to have something like this in htsjdk directly rather than going through users of the library eg picard or gatk to test/benchmark reader/writer improvements.

### Checklist

- [x] Code compiles correctly

### DIFF
--- a/src/java/htsjdk/samtools/example/PrintReadsExample.java
+++ b/src/java/htsjdk/samtools/example/PrintReadsExample.java
@@ -1,0 +1,106 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 The Broad Institute
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ */
+package htsjdk.samtools.example;
+
+import htsjdk.samtools.*;
+import htsjdk.samtools.util.Log;
+import htsjdk.samtools.util.ProgressLogger;
+import htsjdk.samtools.util.zip.DeflaterFactory;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.InetAddress;
+import java.util.Arrays;
+
+
+/**
+ * This is a example program showing how to use SAM readers and (optionally) writers.
+ * It's also useful for measuring time.
+ * An example invocation is:
+ * java -cp dist/htsjdk-2.1.1.jar htsjdk.samtools.example.PrintReadsExample in.bam false a.bam
+ * <p>
+ * or (to test the IntelDeflator)
+ * java -Dsamjdk.intel_deflater_so_path=$PWD/lib/jni/libIntelDeflater.so -cp dist/htsjdk-2.1.1.jar htsjdk.samtools.example.PrintReadsExample in.bam false a.bam
+ * <p>
+ * Arguments:
+ * - the first argument is the input file (SAM or BAM)
+ * - the second argument is a boolean (true or false) that indicates whether reads are to be eagerly decoded (useful for benchmarking)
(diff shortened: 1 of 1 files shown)


src/java/htsjdk/samtools/PrintReads.java
```diff
@@ -0,0 +1,64 @@
...
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ */
+package htsjdk.samtools;
```
yfarjoun: please move this to an examples package or something to make clear that this isn't part of the distribution

src/java/htsjdk/samtools/PrintReads.java
```diff
@@ -0,0 +1,64 @@
...
+            System.exit(1);
+        }
+        final File inputFile = new File(args[0]);
+        final boolean decodeEverything = Boolean.parseBoolean(args[1]); //useful to test (realistic) scenarios in which every record is always fully decoded.
+        final File outputFile = new File(args[2]);
+
+        try( SamReader reader = SamReaderFactory.makeDefault().validationStringency(ValidationStringency.SILENT).open(inputFile)) {
+            final SAMFileHeader header = reader.getFileHeader();
+            try (SAMFileWriter writer = new SAMFileWriterFactory().makeBAMWriter(header, true, outputFile)) {
+                final ProgressLogger pl= new ProgressLogger(log, 100000);
```
yfarjoun: should probably log less (1e6? 1e7?) given how little work you do per read.

yfarjoun: Does the progress logger emit time upon completion? if not, might as well do that from the program rather than have to wrap this with an external "time" command.

src/java/htsjdk/samtools/PrintReads.java
```diff
@@ -0,0 +1,64 @@
...
+import htsjdk.samtools.util.Log;
+import htsjdk.samtools.util.ProgressLogger;
+
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * This is a example program showing how to use SAM readers and writers.
+ * It's also useful for measuring time.
+ */
```
yfarjoun: To be a useful example, please explain here what the arguments will be and how to use, with an example invocation.

src/java/htsjdk/samtools/PrintReads.java
```diff
@@ -0,0 +1,64 @@
...
+public final class PrintReads {
```
yfarjoun: Please change name to PrintReadsExample or something similar to further explain that this is not even pretending to emulate GATK PrintReads, or something like that.

src/java/htsjdk/samtools/PrintReads.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2010 The Broad Institute
```
yfarjoun: 2016

src/java/htsjdk/samtools/PrintReads.java
akiezun: done, renamed to PrintReadsExample. I can't move it to the example package though, here's why (comment from code):
```
//This method SAMRecord::eagerDecode() is protected and there's no way to set
//eager decoding on the reader via a factory so this example must live in the samtools.htsjdk package
//(or extend SAMRecord but that would be bizarre)
```

src/java/htsjdk/samtools/PrintReads.java
akiezun: done. changed logging to 1e6 records (~every 10 seconds or so) and added time stamp at the end

src/java/htsjdk/samtools/PrintReads.java
akiezun: done

src/java/htsjdk/samtools/PrintReads.java
akiezun: done

src/java/htsjdk/samtools/PrintReads.java
akiezun: done

akiezun:
@yfarjoun I addressed your comments. Can you rereview?

src/java/htsjdk/samtools/PrintReadsExample.java
```diff
@@ -0,0 +1,96 @@
...
+			   " " + System.getProperty("java.runtime.version") +
+			   " " + (DeflaterFactory.usingIntelDeflater()? "IntelDeflater": "JdkDeflater"));
+
+        try( SamReader reader = SamReaderFactory.makeDefault().validationStringency(ValidationStringency.SILENT).open(inputFile)) {
+            final SAMFileHeader header = reader.getFileHeader();
+            try (SAMFileWriter writer = outputFile != null ? new SAMFileWriterFactory().makeBAMWriter(header, true, outputFile) : null) {
+                final ProgressLogger pl = new ProgressLogger(log, 1000000);
+                for (final SAMRecord record : reader) {
+                    if (eagerDecode) {
+                        //This method is protected and there's no way to set
```
yfarjoun: SamReaderFactory.makeDefault()
                .enable(SamReaderFactory.Option.EAGERLY_DECODE)

yfarjoun:
back to you @akiezun. please use .enable and move to a different package.

akiezun:
@yfarjoun addressed your comments

yfarjoun:
thanks for humoring me. :+1:

PR #522, opened 2016-03-11
Add a method to expose whether the stream is positioned at the end of a BGZF block.

tomwhite:
### Description

Add a method to BlockCompressedInputStream to expose whether the stream is positioned at the end of a BGZF block. This is needed by Hadoop-BAM (and GATK4 indirectly via its dependency on Hadoop-BAM) to add support for reading block compressed VCF files in parallel. See https://github.com/HadoopGenomics/Hadoop-BAM/issues/68 and https://github.com/HadoopGenomics/Hadoop-BAM/pull/70. /cc @akiezun 

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
+++ b/src/java/htsjdk/samtools/util/BlockCompressedInputStream.java
@@ -133,6 +133,14 @@ public class BlockCompressedInputStream extends InputStream implements LocationA
     }
 
     /**
+     * @return <code>true</code> if the stream is at the end of a BGZF block,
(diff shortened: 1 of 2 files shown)


akiezun:
:+1: from me but I'd :+1: it even more if there was a test (this class has actually no tests so adding even 1 would be huge)

tomwhite:
Thanks @akiezun. I've added a test for this now.

akiezun:
LGTM - if there are no objections, I'll merge in 1 day

PR #524, opened 2016-03-17
High Priority Build Fix: Fixed igzip build issue which resulted in 'invalid block type' error during Inflate

mishalinaik:
### Description

Fixed igzip build issue. igzip should be built per the src/c/inteldeflater/README. Also added a test  src/test/java/htsjdk/samtools/util/zip/IntelDeflaterTest.java. 

### Checklist

Build details added to src/c/inteldeflater/README


yfarjoun:
Tests please!!!

Can you add a simple test that take a sam, and round-trips it? Also in the
test please assert that the intelDeflator has been loaded.

Thanks.

mishalinaik:
@yfarjoun @akiezun @lbergelson @ktibbett @jacarey  As part of this PR I have added the test along with the fix

yfarjoun:
Should there be an ant test added to run this (and then run it in travis?) how will this get tested in every build?

src/tests/java/htsjdk/samtools/util/zip/IntelDeflaterTest.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    private static void printConfigurationInfo() throws IOException {
+        log.info("Executing as " +
+                System.getProperty("user.name") + '@' + InetAddress.getLocalHost().getHostName() +
+                " on " + System.getProperty("os.name") + ' ' + System.getProperty("os.version") +
+                ' ' + System.getProperty("os.arch") + "; " + System.getProperty("java.vm.name") +
+                ' ' + System.getProperty("java.runtime.version") +
+                ' ' + (DeflaterFactory.usingIntelDeflater() ? "IntelDeflater" : "JdkDeflater"));
+        log.info("CREATE_INDEX:" + Defaults.CREATE_INDEX +
```
yfarjoun: this is misleading. should probably use the values from the writer. I just tried is with compression level 1 and it said 5...

yfarjoun:
I tested this in one of our unix machines and it worked. 

My question to @jacarey  and @ktibbett is how should this test be hooked up in the travis build? Should travis issue the command and test the return value? Should it be run at various compression levels and with/without eager decoding? should the inputs be varied? Should modifying the travis build be the responsibility of the submitter or will someone else hook it up?

I tried putting a version of this as a @Test but ant runs without the `-Dsamjdk.intel_deflater_so_path=$PWD/lib/jni/libIntelDeflater.so ` argument, so it fails. Also, one could surmise that some users might not want to use this deflator, so it shouldn't be in the regular test suite.

src/tests/java/htsjdk/samtools/util/zip/IntelDeflaterTest.java
mishalinaik: @yfarjoun Done.

yfarjoun:
this is now irrelevant as #536  includes this and more.

PR #526, opened 2016-03-18
close opened streams. Thanks to @d-cameron

akiezun:
### Description

Closed a bunch of streams that were left open. Thanks to @d-cameron - these changes come from https://github.com/samtools/htsjdk/pull/482

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/java/htsjdk/samtools/BAMFileReader.java
+++ b/src/java/htsjdk/samtools/BAMFileReader.java
@@ -52,7 +52,7 @@ class BAMFileReader extends SamReader.ReaderImplementation {
     private BinaryCodec mStream = null;
 
     // Underlying compressed data stream.
-    private BlockCompressedInputStream mCompressedInputStream;
+    private final BlockCompressedInputStream mCompressedInputStream;
     private SAMFileHeader mFileHeader = null;
 
     // One of these is populated if the file is seekable and an index exists
@@ -224,9 +224,6 @@ class BAMFileReader extends SamReader.ReaderImplementation {
     }
 
     @Override void enableCrcChecking(final boolean enabled) {
-        if (mCompressedInputStream == null) {
-            throw new IllegalStateException("File reader is closed");
-        }
         this.mCompressedInputStream.setCheckCrcs(enabled);
     }
 
@@ -263,7 +260,8 @@ class BAMFileReader extends SamReader.ReaderImplementation {
     }
 
     public void setEagerDecode(final boolean desired) { this.eagerDecode = desired; }
-    
+
+    @Override
     public void close() {
         if (mCompressedInputStream != null) {
             try {
@@ -278,7 +276,6 @@ class BAMFileReader extends SamReader.ReaderImplementation {
         if (mIndex != null) {
             mIndex.close();
         }
-        mCompressedInputStream = null;
         mStream = null;
         mFileHeader = null;
         mIndex = null;
@@ -311,9 +308,6 @@ class BAMFileReader extends SamReader.ReaderImplementation {
         if (mStream == null) {
             throw new IllegalStateException("File reader is closed");
         }
-        if (mCompressedInputStream == null) {
-            throw new IllegalStateException("File reader is closed");
-        }
         if (mCurrentIterator != null) {
             throw new IllegalStateException("Iteration in progress");
         }
@@ -331,9 +325,6 @@ class BAMFileReader extends SamReader.ReaderImplementation {
     @Override
     public CloseableIterator<SAMRecord> getIterator(final SAMFileSpan chunks) {
         if (mStream == null) {
-            throw new IllegalStateException("File reader is closed");
-        }
-        if (mCompressedInputStream == null) {
             throw new IllegalStateException("File reader is closed");
         }
         if (mCurrentIterator != null) {
@@ -475,9 +466,6 @@ class BAMFileReader extends SamReader.ReaderImplementation {
      */
     public CloseableIterator<SAMRecord> queryUnmapped() {
         if (mStream == null) {
-            throw new IllegalStateException("File reader is closed");
-        }
-        if (mCompressedInputStream == null) {
             throw new IllegalStateException("File reader is closed");
         }
         if (mCurrentIterator != null) {
--- a/src/java/htsjdk/samtools/BAMFileReader.java
+++ b/src/java/htsjdk/samtools/BAMFileReader.java
@@ -52,7 +52,7 @@ class BAMFileReader extends SamReader.ReaderImplementation {
     private BinaryCodec mStream = null;
 
     // Underlying compressed data stream.
-    private final BlockCompressedInputStream mCompressedInputStream;
+    private BlockCompressedInputStream mCompressedInputStream;
     private SAMFileHeader mFileHeader = null;
 
     // One of these is populated if the file is seekable and an index exists
@@ -224,6 +224,9 @@ class BAMFileReader extends SamReader.ReaderImplementation {
     }
 
     @Override void enableCrcChecking(final boolean enabled) {
+        if (mCompressedInputStream == null) {
+            throw new IllegalStateException("File reader is closed");
+        }
         this.mCompressedInputStream.setCheckCrcs(enabled);
     }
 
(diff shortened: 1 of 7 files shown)


src/java/htsjdk/samtools/SAMFileReader.java
```diff
@@ -708,9 +708,11 @@ private boolean isBAMFile(final InputStream stream)
         final byte[] buffer = new byte[buffSize];
         readBytes(stream, buffer, 0, buffSize);
         stream.reset();
-        final byte[] magicBuf = new byte[4];
-        final int magicLength = readBytes(new BlockCompressedInputStream(new ByteArrayInputStream(buffer)), magicBuf, 0, 4);
-        return magicLength == BAMFileConstants.BAM_MAGIC.length && Arrays.equals(BAMFileConstants.BAM_MAGIC, magicBuf);
+        try(final BlockCompressedInputStream bcis = new BlockCompressedInputStream(new ByteArrayInputStream(buffer))){
```
tfenne: Your PR has made some code duplication painfully obvious ;)  Could this entire method just delegate to `SamStreams.isBAMfile()`?

src/java/htsjdk/tribble/index/IndexFactory.java
```diff
@@ -166,10 +166,11 @@ public static IndexType getIndexType(final BufferedInputStream is) {
      */
     public static Index loadIndex(final String indexFile) {
         final Index idx = null;
+        InputStream inputStream = null;
         BufferedInputStream bufferedInputStream = null;
         final LittleEndianInputStream dis = null;
         try {
-            InputStream  inputStream = ParsingUtils.openInputStream(indexFile);
+            inputStream = ParsingUtils.openInputStream(indexFile);
```
tfenne: There's still a leak here because the original code is rather odd.  It's opening an input stream here, and then possibly re-assigning `inputStream` immediately in the `if` below (I have no idea why).  Perhaps move line 173 into an `else` block after line 179?

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -614,10 +614,13 @@ private static void parseSingleAltAllele(List<Allele> alleles, String alt, int l
     }
 
     public final static boolean canDecodeFile(final String potentialInput, final String MAGIC_HEADER_LINE) {
-        try {
-            return isVCFStream(new FileInputStream(potentialInput), MAGIC_HEADER_LINE) ||
-                    isVCFStream(new GZIPInputStream(new FileInputStream(potentialInput)), MAGIC_HEADER_LINE) ||
-                    isVCFStream(new BlockCompressedInputStream(new FileInputStream(potentialInput)), MAGIC_HEADER_LINE);
+        try(
+            final FileInputStream fis = new FileInputStream(potentialInput);
```
tfenne: Not your doing, and doesn't _need_ to be fixed, but this method is icky.  Why open up three streams (which will open three filehandles and do buffering) when you may only need the first one or two in many cases??

src/java/htsjdk/samtools/BAMFileReader.java
```diff
@@ -52,7 +52,7 @@
     private BinaryCodec mStream = null;
 
     // Underlying compressed data stream.
-    private final BlockCompressedInputStream mCompressedInputStream;
+    private BlockCompressedInputStream mCompressedInputStream;
```
tfenne: I don't love making this non-final and doing null checking everywhere.  It's probably reasonable to think that upon `BAMFileReader.close()` the reader itself is released and ready for GC soon thereafter, such that there's not much to gain from releasing the `BlockCompressedInputStream` a little faster.

src/java/htsjdk/samtools/SAMFileReader.java
akiezun: yes, done. and method inlined

src/java/htsjdk/tribble/index/IndexFactory.java
akiezun: that wont work because line 175 uses the `inputStream` created on 173

akiezun: actually, reading this again, whatever happens, `inputStream` always ends up wrapped inside of `bufferedInputStream` and that gets a `close()` call, which delegates to the wrapped `inputStream` (and that would close any other stream wrapped inside of `inputStream` - both `BlockCompressedInputStream` and `GZIPInputStream` do that). So the call to `inputStream.close()` is not really needed at all. Am I reading this right?

Also, the `Index` object that gets returned has a stream inside it and that stream is always closed at the beginning. Is that strange?

src/java/htsjdk/variant/vcf/AbstractVCFCodec.java
akiezun: i can't imagine `canDecodeFile` being a bottleneck and rewriting this may make it less readable. Let's leave as is for now, wdyt?

src/java/htsjdk/samtools/BAMFileReader.java
akiezun: fair enough. done

akiezun:
@tfenne i addressed the issues you mentioned. Do you mind taking another look?

akiezun:
@lbergelson can you review?

src/java/htsjdk/samtools/BAMFileReader.java
```diff
@@ -260,8 +260,16 @@ public BAMIndex getIndex() {
...
     public void setEagerDecode(final boolean desired) { this.eagerDecode = desired; }
-    
+
+    @Override
     public void close() {
+        if (mCompressedInputStream != null) {
+            try {
+                mCompressedInputStream.close();
+            } catch (IOException e) {
+                throw new SAMException("Exception closing compressed input stream.", e);
```
lbergelson: Maybe use the somewhat more specific `RuntimeIOException`

src/java/htsjdk/samtools/SAMFileReader.java
```diff
@@ -694,25 +694,6 @@ private void init(final InputStream stream, File file, final File indexFile, fin
         }
     }
 
-    /**
```
lbergelson: While we're killing this, could you also kill `isGzippedSAMFile` and delegate to the SamStreams variant?

src/java/htsjdk/tribble/index/IndexFactory.java
```diff
@@ -165,16 +165,19 @@ public static IndexType getIndexType(final BufferedInputStream is) {
      * @param indexFile from which to load the index
      */
     public static Index loadIndex(final String indexFile) {
```
lbergelson: This still seems ugly.  I propose the following instead:

```java
   /**
     * Load in index from the specified file.   The type of index (LinearIndex or IntervalTreeIndex) is determined
     * at run time by reading the type flag in the file.
     *
     * @param indexFile from which to load the index
     */
    public static Index loadIndex(final String indexFile) {
        try (InputStream inputStream = openIndexFile(indexFile);
             // Must be buffered, because getIndexType uses mark and reset
             BufferedInputStream bufferedInputStream = new BufferedInputStream(inputStream, Defaults.NON_ZERO_BUFFER_SIZE))
        {
            final Class<Index> indexClass = IndexType.getIndexType(bufferedInputStream).getIndexType();

            final Constructor<Index> ctor = indexClass.getConstructor(InputStream.class);

            return ctor.newInstance(bufferedInputStream);
        } catch (final IOException ex) {
            throw new TribbleException.UnableToReadIndexFile("Unable to read index file", indexFile, ex);
        } catch (final Exception ex) {
            throw new RuntimeException(ex);
        }
    }

    private static InputStream openIndexFile(final String indexFile) throws IOException {
        final InputStream inputStreamInitial = ParsingUtils.openInputStream(indexFile);
        if (indexFile.endsWith(".gz")) {
            return new GZIPInputStream(inputStreamInitial);
        }
        else if (indexFile.endsWith(TabixUtils.STANDARD_INDEX_EXTENSION)) {
            return new BlockCompressedInputStream(inputStreamInitial);
        }
        else {
            return inputStreamInitial;
        }
    }
```

also needs `Class getIndexType()` ->` Class<Index> getIndexType()`

lbergelson:
@akiezun back to you.

src/java/htsjdk/samtools/BAMFileReader.java
akiezun: done

src/java/htsjdk/samtools/SAMFileReader.java
akiezun: done + added tests for all 2: isBAMFile isCRAMFile and isGzippedSAMFile

src/java/htsjdk/tribble/index/IndexFactory.java
akiezun: done

akiezun:
back to @lbergelson

src/java/htsjdk/samtools/SAMFileReader.java
lbergelson: excellent, thanks for adding tests

lbergelson:
:+1:

PR #527, opened 2016-03-19
print variants example

akiezun:
### Description

Similar to `PrintReadsExample`, this is a `PrintVariantsExample`. The purpose it to demonstrate how to use readers and writers but also to enable easy performance testing etc.

Run it like this
```
java -cp dist/htsjdk-2.1.1.jar htsjdk.variant.example.PrintVariantsExample file.vcf
```

optionally add output file 
```
java -cp dist/htsjdk-2.1.1.jar htsjdk.variant.example.PrintVariantsExample file.vcf out.vcf
```

### DIFF
--- a/src/java/htsjdk/variant/example/PrintVariantsExample.java
+++ b/src/java/htsjdk/variant/example/PrintVariantsExample.java
@@ -1,0 +1,105 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 The Broad Institute
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ */
+package htsjdk.variant.example;
+
+import htsjdk.samtools.Defaults;
+import htsjdk.samtools.util.Log;
+import htsjdk.samtools.util.ProgressLogger;
+import htsjdk.samtools.util.zip.DeflaterFactory;
+import htsjdk.tribble.AbstractFeatureReader;
+import htsjdk.tribble.readers.LineIterator;
+import htsjdk.variant.variantcontext.VariantContext;
+import htsjdk.variant.variantcontext.writer.Options;
+import htsjdk.variant.variantcontext.writer.VariantContextWriter;
+import htsjdk.variant.variantcontext.writer.VariantContextWriterBuilder;
+import htsjdk.variant.vcf.VCFCodec;
+import htsjdk.variant.vcf.VCFHeader;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.InetAddress;
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+/**
+ * This is a example program showing how to use Feature readers and (optionally) writers.
+ * It's also useful for measuring time.
+ * An example invocation is:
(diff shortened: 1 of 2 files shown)


src/java/htsjdk/variant/example/PrintVariantsExample.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private static void printConfigurationInfo() throws IOException {
+        log.info("Executing as " +
+                System.getProperty("user.name") + '@' + InetAddress.getLocalHost().getHostName() +
+                " on " + System.getProperty("os.name") + ' ' + System.getProperty("os.version") +
+                ' ' + System.getProperty("os.arch") + "; " + System.getProperty("java.vm.name") +
+                ' ' + System.getProperty("java.runtime.version") +
+                ' ' + (DeflaterFactory.usingIntelDeflater() ? "IntelDeflater" : "JdkDeflater"));
+        log.info("CREATE_INDEX:" + Defaults.CREATE_INDEX +
```
yfarjoun: is Options.INDEX_ON_THE_FLY the same as Defaults.CREATE_INDEX?

akiezun: i don't know. You specify defaults on the commandline and Options are programmatic

akiezun: answer  - it's not the same `Defaults.CREATE_INDEX` is only for BAMs as per doc (and code concurs)
```
 /** Should BAM index files be created when writing out coordinate sorted BAM files?  Default = false. */
```

akiezun:
@yfarjoun ok to merge?

akiezun:
@yfarjoun i switched to using `Defaults.allDefaults()` to print all options.

akiezun:
@lbergelson can you review in YF's absence?

src/java/htsjdk/variant/example/PrintVariantsExample.java
```diff
@@ -0,0 +1,107 @@
...
+
+    private static void printConfigurationInfo() throws IOException {
+        log.info("Executing as " +
+                System.getProperty("user.name") + '@' + InetAddress.getLocalHost().getHostName() +
+                " on " + System.getProperty("os.name") + ' ' + System.getProperty("os.version") +
+                ' ' + System.getProperty("os.arch") + "; " + System.getProperty("java.vm.name") +
+                ' ' + System.getProperty("java.runtime.version") +
+                ' ' + (DeflaterFactory.usingIntelDeflater() ? "IntelDeflater" : "JdkDeflater"));
+
+        final List<String> list = Defaults.allDefaults().entrySet().stream().map(e -> e.getKey() + ':' + e.getValue()).collect(Collectors.toList());
```
lbergelson: There's a joining collector if you want to save a line `.collect(Collectors.joining(" "))`.

src/java/htsjdk/variant/example/PrintVariantsExample.java
```diff
@@ -0,0 +1,107 @@
...
+                }
+                pl.record(vc.getContig(), vc.getStart());
+            }
+        }
+
+        final long end = System.currentTimeMillis();
+        log.info(String.format("Done. Elapsed time %.3f seconds", (end - start) / 1000.0));
+    }
+
+    private static void printConfigurationInfo() throws IOException {
```
lbergelson: Would this method be useful to extract somewhere?  I can see many programs wanting to log all the relevant htsjdk settings.  Especially the defaults and the deflator factory.

src/java/htsjdk/variant/example/PrintVariantsExample.java
```diff
@@ -0,0 +1,107 @@
...
+ * This is a example program showing how to use Feature readers and (optionally) writers.
+ * It's also useful for measuring time.
+ * An example invocation is:
+ * java -cp dist/htsjdk-2.1.1.jar htsjdk.variant.example.PrintVariantsExample in.vcf out.vcf
+ * <p>
+ * Arguments:
+ * - the first argument is the input file (VCF)
+ * - the second argument is optional and is the name of the output file (nothing gets written if this argument is missing)
+ */
+public final class PrintVariantsExample {
```
lbergelson: This should probably have a test case of some kind, (although the test code is likely to reproduce the tool in complexity...) It's extremely frustrating to have example programs that don't function correctly, and it's hard to be sure we won't break it in the future without a test.

lbergelson:
@akiezun It looks good to me.  I would like to have a test, just to verify that it doesn't get completely broken and out of date.  We might want to lift the settings logging to somewhere in defaults or utils as well.  Merge when you're happy with it.

src/java/htsjdk/variant/example/PrintVariantsExample.java
akiezun: cool, done

src/java/htsjdk/variant/example/PrintVariantsExample.java
akiezun: if you have a suggestion of where to put it, i'm open to it.

akiezun:
@lbergelson ok to merge?

src/tests/java/htsjdk/variant/PrintVariantsExampleTest.java
```diff
@@ -0,0 +1,65 @@
+/*
+* Copyright (c) 2012 The Broad Institute
```
lbergelson: This should probably be 2016

src/java/htsjdk/variant/example/PrintVariantsExample.java
lbergelson: Not sure where exactly it could go, maybe in defaults.  Leave it how it is for now. We can figure it out later if we need it.

lbergelson:
:+1: Fix the license if that's important.  Merge at will.

PR #528, opened 2016-03-19
remove asynchronous reading from Tribble LineReaderUtil

akiezun:
### Description

This PR removes asynchronous reading from Tribble. The reason is that performance was very bad. I got hit by it here: https://github.com/broadinstitute/gatk/issues/1597 and it can easily be reproduced. 

To see the problem, run the PrintVariantsExample (from #527 ) 

like this (sync IO):
```
java -Dsamjdk.use_async_io=false -cp dist/htsjdk-2.1.1.jar htsjdk.variant.example.PrintVariantsExample dbsnp_138.b37.excluding_sites_after_129.chr1_4.vcf
```

and then like this (async IO)
```
java -Dsamjdk.use_async_io=true -cp dist/htsjdk-2.1.1.jar htsjdk.variant.example.PrintVariantsExample dbsnp_138.b37.excluding_sites_after_129.chr1_4.vcf
```

(pick a file for which this takes at least a few seconds). On my Mac, the results are dramatic - asych is 25% *slower* (async IO is supposed to be faster!)

This PR removes async from tribble readers. With this change, performance of the two commandlines above is the same (reviewer should verify that)

### DIFF
--- a/src/java/htsjdk/tribble/TabixFeatureReader.java
+++ b/src/java/htsjdk/tribble/TabixFeatureReader.java
@@ -128,7 +128,7 @@ public class TabixFeatureReader<T extends Feature, SOURCE> extends AbstractFeatu
     public CloseableTribbleIterator<T> iterator() throws IOException {
         final InputStream is = new BlockCompressedInputStream(ParsingUtils.openInputStream(path));
         final PositionalBufferedStream stream = new PositionalBufferedStream(is);
-        final LineReader reader = LineReaderUtil.fromBufferedStream(stream, LineReaderUtil.LineReaderOption.SYNCHRONOUS);
+        final LineReader reader = LineReaderUtil.fromBufferedStream(stream);
         return new FeatureIterator<T>(reader, 0, Integer.MAX_VALUE);
     }
 
--- a/src/java/htsjdk/tribble/readers/LineReaderUtil.java
+++ b/src/java/htsjdk/tribble/readers/LineReaderUtil.java
@@ -1,9 +1,7 @@
 package htsjdk.tribble.readers;
 
-import htsjdk.samtools.Defaults;
 import htsjdk.samtools.util.CloserUtil;
 import htsjdk.samtools.util.RuntimeIOException;
-import htsjdk.tribble.TribbleException;
 
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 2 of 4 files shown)


yfarjoun:
Without understanding why this is the case, I think that this requires substantial testing on a wide range of data-type, size, access patterns, operating systems, and file-systems (perhaps more dimensions).

I'm not sure how much of that 5-dimensional array we need to look at to feel comfortable, but in my opinion, the test you describe is not enough.

akiezun:
Doing a full 5D analysis may be a bit much (particularly if we then apply the same bar to all PRs). But here are some stats to prove the point that **asych reading of tribble feature files** (and the PR only extends to this 1 case) is terribly inefficient. In fact, most users of `LineReaderUtil` already recognize this and only use SYNCHRONOUS as the argument (I did a search for users of that class)

image

OK, here are some stats showing that sync IO loses every time even when multiple CPUs are available. On 1 CPU it's a disaster:

1) on Unix, data on NFS, using mutliple CPUs. **Async loses 48sec vs 39sec**

the file was 2.3GB size. I ran each at least 2 times and the timings were very similar (down to 0.5 second)

Sync
```
time java -Dsamjdk.use_async_io=false -cp dist/htsjdk-2.1.1.jar htsjdk.variant.example.PrintVariantsExample /dsde/working/akiezun/data/dbsnp_138.b37.excluding_sites_after_129.vcf

real	0m39.275s
user	0m48.004s
sys	0m1.992s
```

Async
```
time java -Dsamjdk.use_async_io=true -cp dist/htsjdk-2.1.1.jar htsjdk.variant.example.PrintVariantsExample /dsde/working/akiezun/data/dbsnp_138.b37.excluding_sites_after_129.vcf


real	0m47.992s
user	1m19.060s
sys	0m14.745s
```

2) on Unix, data on NFS, using 1 CPU (pinned JVM to 1 core using taskset). ***Async loses 86 seconds to 39 !***

Sync
```
taskset -c 0 time java -Dsamjdk.use_async_io=false -cp dist/htsjdk-2.1.1.jar htsjdk.variant.example.PrintVariantsExample /dsde/working/akiezun/data/dbsnp_138.b37.excluding_sites_after_129.vcf

Elapsed time 39.242 seconds
38.01 user 
1.35 system
```

async
```
taskset -c 0 time java -Dsamjdk.use_async_io=true -cp dist/htsjdk-2.1.1.jar htsjdk.variant.example.PrintVariantsExample /dsde/working/akiezun/data/dbsnp_138.b37.excluding_sites_after_129.vcf

Elapsed time 85.934 seconds
55.68 user
30.30system
```

3) Using linux and local disk - same results as NFS. 
4) Using linux and a VCF file with a Tribble index - same results as the others.

droazen:
Discussed in person with @akiezun -- I suggested that instead of deleting this functionality from tribble, we add a separate `Defaults.USE_ASYNC_IO_FOR_TRIBBLE` property that controls async I/O for the `tribble` package only (with a warning in the docs that it may not improve performance in many cases), and clarify that existing `Defaults.USE_ASYNC_IO` applies only to the `samtools` package. This seems defensible given that the documentation for the `USE_ASYNC_IO` option reads `Should asynchronous I/O be used when writing out SAM and BAM files`.

Thoughts on this?

droazen:
Back to @akiezun for a response. If you're convinced that deletion is the better option, and that `AsynchronousLineReader` should never be used by anyone, I'll go along with that.

akiezun:
The commit comment mentions long lines so I'll try on wide VCFs (so far I've only tried on narrow VCFs, ie few samples and lots of lines). Maybe @mccowan still remembers the scenarios where async read worked well? Mike, can you help out?

pfxuan:
it seems [AsynchronousLineReader](https://github.com/samtools/htsjdk/blob/master/src/java/htsjdk/tribble/readers/AsynchronousLineReader.java) has never been updated since it's implementated. Is it possible to refactor this class to improve the async-read performance?

akiezun:
It may be easier to delete and write from scratch. It's a different project
though

pfxuan:
@akiezun, I agree! Java NIO has a built-in async I/O, like [AsynchronousFileChannel](https://docs.oracle.com/javase/7/docs/api/java/nio/channels/AsynchronousFileChannel.html), which could be used to reimplment this class. It looks like htsjdk doesn't use any Java NIO on it readers.

akiezun:
@pfxuan yes, all this was implemenred pre-java7

pfxuan:
@akiezun, I see! I guess this is where the issue comes from.

akiezun:
OK, so I tested it on a big wide VCF and do see a little big of benefit of asynchronous reading of such VCFs:

Data was a 4.6GB file with 5780 lines and 40537 samples, with a tribble index, sitting on NFS. I ran it on `on Linux 2.6.32-573.3.1.el6.x86_64 amd64; Java HotSpot(TM) 64-Bit Server VM 1.8.0-b132`

I ran 5 times with synch and async IO:
async IO = 15.885, 10.425, 16.311, 10.534, 10.439 -> mean 12.72, stdev 3
sync IO = 15.283, 15.617, 14.989, 14.894, 14.867 -> mean 15.13, stdev 0.31

Given that, let's not delete this code right away. I just need a way to disable it for my usecase. @yfarjoun wdyt about @droazen proposal?

yfarjoun:
hmmm. I like the fine-grained controls..but I do not like the naming
convention. I think that if there is a Defaults.USE_ASYNC_IO_FOR_TRIBBLE
then there should be a Defaults.USE_ASYNC_IO_FOR_SAMTOOLS otherwise it will
be confusing in the future...

can we have Defaults.USE_ASYNC_IO mean: "for everything" and add two new
Defaults?
Defaults.USE_ASYNC_IO_FOR_SAMTOOLS and
Defaults.USE_ASYNC_IO_FOR_TRIBBLE

droazen:
@yfarjoun :+1: from me, provided that we alter the docs for the existing `USE_ASYNC_IO` option to no longer say `"when writing out SAM and BAM files"`

yfarjoun:
which is how it was interpreted anyhow...

akiezun:
how do you want conflicting settings handled?
eg user specifies `true` for `USE_ASYNC_IO_FOR_SAMTOOLS ` but `false` for `USE_ASYNC_IO`

akiezun:
If `USE_ASYNC_IO` is specified, then it's easy - it overrides everything else. 
But ft it is not specified, and someone says `USE_ASYNC_IO_FOR_SAMTOOLS=true` and `USE_ASYNC_IO_FOR_TRIBBLE=false` then what? 

For now, I'll go over all internal users of `USE_ASYNC_IO` and change them to use `USE_ASYNC_IO_FOR_SAMTOOLS` or `USE_ASYNC_IO_FOR_TRIBBLE` but external clients will break and so I have to make the field deprecated.

droazen:
@akiezun If `USE_ASYNC_IO` is specified, set both `USE_ASYNC_IO_FOR_SAMTOOLS` and `USE_ASYNC_IO_FOR_TRIBBLE` to true unconditionally in `Defaults` (even if they're also specified). The relevant code in the tribble/samtools packages should check `USE_ASYNC_IO_FOR_SAMTOOLS` or `USE_ASYNC_IO_FOR_TRIBBLE` as appropriate, rather than looking at `USE_ASYNC_IO`. I don't think this should cause any breakage for external clients?

akiezun:
wait, what if `USE_ASYNC_IO` is specified and set to `false`? Also, external clients are going to be broken because async IO may be used when though it is not specified on commandline (which is different than before the change) in the case when `USE_ASYNC_IO` is *not* specified.

akiezun:
closing this in favor of split async samtools and tribble options #529

PR #532, opened 2016-03-21
Clean up Genotype JEXL Filtering

bbimber:
I am hoping someone more familiar with JEXL in htsjdk could do a quick review of what i'm proposing to see if this looks reasonable.  As noted in your code, genotype-level JEXL filtering is supported, but not implemented very cleanly.  I tried to improve that and also make it more consistent w/ the context available for VariantContext.

Here's a related GATK forum thread:

http://gatkforums.broadinstitute.org/gatk/discussion/comment/28654#Comment_28654

### DIFF
--- a/src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
+++ b/src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
@@ -6,7 +6,12 @@ import java.util.HashMap;
 import java.util.Map;
 
 /**
- * Created by bimber on 3/21/2016.
+ *
+ * @author bbimber
+ *
+ * implements the JEXL context for Genotype; this saves us from
+ * having to generate a JEXL context lookup map everytime we want to evaluate an expression.
+ *
  */
 public class GenotypeJEXLContext extends VariantJEXLContext {
     private Genotype g;
--- a/src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
+++ b/src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
@@ -1,0 +1,53 @@
+package htsjdk.variant.variantcontext;
+
+import htsjdk.variant.vcf.VCFConstants;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Created by bimber on 3/21/2016.
+ */
+public class GenotypeJEXLContext extends VariantJEXLContext {
+    private Genotype g;
+
+    private interface AttributeGetter {
+        public Object get(Genotype g);
+    }
+
+    private static Map<String, AttributeGetter> attrs = new HashMap<String, AttributeGetter>();
+
+    static {
+        attrs.put("g", new AttributeGetter() { public Object get(Genotype g) { return g; }});
+        attrs.put(VCFConstants.GENOTYPE_KEY, new AttributeGetter() { public Object get(Genotype g) { return g.getGenotypeString(); }});
+
+        attrs.put("isHom", new AttributeGetter() { public Object get(Genotype g) { return g.isHom() ? "1" : "0"; }});
+        attrs.put("isHomRef", new AttributeGetter() { public Object get(Genotype g) { return g.isHomRef() ? "1" : "0"; }});
+        attrs.put("isHet", new AttributeGetter() { public Object get(Genotype g) { return g.isHet() ? "1" : "0"; }});
+        attrs.put("isHomVar", new AttributeGetter() { public Object get(Genotype g) { return g.isHomVar() ? "1" : "0"; }});
+        attrs.put("isCalled", new AttributeGetter() { public Object get(Genotype g) { return g.isCalled() ? "1" : "0"; }});
+        attrs.put("isNoCall", new AttributeGetter() { public Object get(Genotype g) { return g.isNoCall() ? "1" : "0"; }});
+        attrs.put("isMixed", new AttributeGetter() { public Object get(Genotype g) { return g.isMixed() ? "1" : "0"; }});
+        attrs.put("isAvailable", new AttributeGetter() { public Object get(Genotype g) { return g.isAvailable() ? "1" : "0"; }});
+        attrs.put("isPassFT", new AttributeGetter() { public Object get(Genotype g) { return g.isFiltered() ? "0" : "1"; }});
+        attrs.put(VCFConstants.GENOTYPE_FILTER_KEY, new AttributeGetter() { public Object get(Genotype g) { return g.isFiltered()?  g.getFilters() : "PASS"; }});
+        attrs.put(VCFConstants.GENOTYPE_QUALITY_KEY, new AttributeGetter() { public Object get(Genotype g) { return g.getGQ(); }});
+    }
+
+    public GenotypeJEXLContext(VariantContext vc, Genotype g) {
+        super(vc);
+        this.g = g;
+    }
+
+    public Object get(String name) {
+        //should matching genotype attributes always supersede vc?
+        if ( attrs.containsKey(name) ) { // dynamic resolution of name -> value via map
+            return attrs.get(name).get(g);
+        } else if ( g.hasAnyAttribute(name) ) {
+            return g.getAnyAttribute(name);
+        } else if ( g.getFilters().contains(name) ) {
+            return "1";
+        } else
+            return super.get(name);
+    }
+}
--- a/src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
+++ b/src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
@@ -20,23 +20,23 @@ public class GenotypeJEXLContext extends VariantJEXLContext {
         public Object get(Genotype g);
     }
 
-    private static Map<String, AttributeGetter> attrs = new HashMap<String, AttributeGetter>();
+    private static Map<String, AttributeGetter> attributes = new HashMap<String, AttributeGetter>();
 
     static {
-        attrs.put("g", new AttributeGetter() { public Object get(Genotype g) { return g; }});
-        attrs.put(VCFConstants.GENOTYPE_KEY, new AttributeGetter() { public Object get(Genotype g) { return g.getGenotypeString(); }});
+        attributes.put("g", (Genotype g) -> g);
+        attributes.put(VCFConstants.GENOTYPE_KEY, Genotype::getGenotypeString);
 
-        attrs.put("isHom", new AttributeGetter() { public Object get(Genotype g) { return g.isHom() ? "1" : "0"; }});
-        attrs.put("isHomRef", new AttributeGetter() { public Object get(Genotype g) { return g.isHomRef() ? "1" : "0"; }});
-        attrs.put("isHet", new AttributeGetter() { public Object get(Genotype g) { return g.isHet() ? "1" : "0"; }});
-        attrs.put("isHomVar", new AttributeGetter() { public Object get(Genotype g) { return g.isHomVar() ? "1" : "0"; }});
-        attrs.put("isCalled", new AttributeGetter() { public Object get(Genotype g) { return g.isCalled() ? "1" : "0"; }});
-        attrs.put("isNoCall", new AttributeGetter() { public Object get(Genotype g) { return g.isNoCall() ? "1" : "0"; }});
-        attrs.put("isMixed", new AttributeGetter() { public Object get(Genotype g) { return g.isMixed() ? "1" : "0"; }});
-        attrs.put("isAvailable", new AttributeGetter() { public Object get(Genotype g) { return g.isAvailable() ? "1" : "0"; }});
-        attrs.put("isPassFT", new AttributeGetter() { public Object get(Genotype g) { return g.isFiltered() ? "0" : "1"; }});
-        attrs.put(VCFConstants.GENOTYPE_FILTER_KEY, new AttributeGetter() { public Object get(Genotype g) { return g.isFiltered()?  g.getFilters() : "PASS"; }});
-        attrs.put(VCFConstants.GENOTYPE_QUALITY_KEY, new AttributeGetter() { public Object get(Genotype g) { return g.getGQ(); }});
+        attributes.put("isHom", (Genotype g) -> g.isHom() ? "1" : "0");
+        attributes.put("isHomRef", (Genotype g) -> g.isHomRef() ? "1" : "0");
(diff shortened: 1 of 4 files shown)


vdauwera:
@yfarjoun Think you could have a look at this? Seems like it would improve JEXL capabilities in GATK.

bbimber:
Hello,

The GATK folks suggested I include an example of how this alters filtering.  Here are some key points:

1) This is general code cleanup.  The existing code has a lot of comments about how this codepath needed improvement.

2) When filtering at the site level, the VariantContext object was in scope.  However, when filtering at the Genotype level, neither vc nor genotype are in scope.  That's a big limitation.  This patch enabled genotype filters like:

-G_filter "isHet && g.hasAd() && g.getAD().1 < 5"

3) It fixes what I think was a bug in the original implementation, which omits AD + PL from the JEXL map. This is because they were not getting included in Genotype.getExtendedAttributes().

I'm happy to incorporate suggestions on alternate approaches.

lindenb:
@bbimber for you information: there is now a javascript-based engine in picard filtervcf FilterVcf . It's, IMHO, much  more powerful than JEXL . See 

https://github.com/samtools/htsjdk/blob/master/testdata/htsjdk/variant/variantFilter02.js

for an example.

vdauwera:
I agree that the javascript filters are very cool but we still need this PR which cleans up existing code and extends functionality, so that we can still offer straightforward one-line querying via JEXL.

bbimber:
hi guys - sorry to bug on this, but is there anything else you need from me?  is there a process through which hts-jdk submissions are reviewed?  as noted by @vdauwera, even if there are newer filter options, JEXL is still used in GATK.  thanks.

vdauwera:
@bbimber I don't think we need anything directly from you right now; I need to find someone to do a technical review of your PR. Since it mostly benefits GATK it will have to be someone from our team, but my first choice of reviewer is on vacation. I'll try to catch someone's attention. Sorry for the delay!

src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
```diff
@@ -0,0 +1,58 @@
...
+        public Object get(Genotype g);
+    }
+
+    private static Map<String, AttributeGetter> attrs = new HashMap<String, AttributeGetter>();
+
+    static {
+        attrs.put("g", new AttributeGetter() { public Object get(Genotype g) { return g; }});
+        attrs.put(VCFConstants.GENOTYPE_KEY, new AttributeGetter() { public Object get(Genotype g) { return g.getGenotypeString(); }});
+
+        attrs.put("isHom", new AttributeGetter() { public Object get(Genotype g) { return g.isHom() ? "1" : "0"; }});
```
yfarjoun: you can use java8 magic to make this a lot cleaner (intelliJ gives nice recommendations, and while you are at it, could you do the same for the VariantJEXLContext?)

bbimber: do you mean lambdas?  can i introduce those if you're using java6?

src/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -55,68 +55,17 @@ private void initialize(Collection<JexlVCMatchExp> jexlCollection) {
...
+            } else {
+                jContext = new GenotypeJEXLContext(vc, g);
             }
-
-            // create the internal context that we can evaluate expressions against
-
-            jContext = new MapContext(infoMap);
+        }
+        else {
+            jContext = new MapContext(new HashMap<String, Object>());
```
yfarjoun: Use Collections.emptyMap

src/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -55,68 +55,17 @@ private void initialize(Collection<JexlVCMatchExp> jexlCollection) {
...
-                infoMap.put(VCFConstants.GENOTYPE_QUALITY_KEY, g.getGQ());
-                if ( g.hasDP() )
-                    infoMap.put(VCFConstants.DEPTH_KEY, g.getDP());
-                for ( Entry<String, Object> e : g.getExtendedAttributes().entrySet() ) {
-                    if ( e.getValue() != null && !e.getValue().equals(VCFConstants.MISSING_VALUE_v4) )
-                        infoMap.put(e.getKey(), e.getValue());
-                }
+        // QUESTION: the original code could have created a new VariantJEXLContext with a null vc.
+        // i dont think this is right, though i have no idea if vc could actually be null
+        if ( vc != null ) {
```
yfarjoun: I'd prefer if you kept the nesting lower by inverting the if statement and returning Collections.emptyMap() if vc==null.

src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
```diff
@@ -0,0 +1,58 @@
...
+        attrs.put("isHet", new AttributeGetter() { public Object get(Genotype g) { return g.isHet() ? "1" : "0"; }});
+        attrs.put("isHomVar", new AttributeGetter() { public Object get(Genotype g) { return g.isHomVar() ? "1" : "0"; }});
+        attrs.put("isCalled", new AttributeGetter() { public Object get(Genotype g) { return g.isCalled() ? "1" : "0"; }});
+        attrs.put("isNoCall", new AttributeGetter() { public Object get(Genotype g) { return g.isNoCall() ? "1" : "0"; }});
+        attrs.put("isMixed", new AttributeGetter() { public Object get(Genotype g) { return g.isMixed() ? "1" : "0"; }});
+        attrs.put("isAvailable", new AttributeGetter() { public Object get(Genotype g) { return g.isAvailable() ? "1" : "0"; }});
+        attrs.put("isPassFT", new AttributeGetter() { public Object get(Genotype g) { return g.isFiltered() ? "0" : "1"; }});
+        attrs.put(VCFConstants.GENOTYPE_FILTER_KEY, new AttributeGetter() { public Object get(Genotype g) { return g.isFiltered()?  g.getFilters() : "PASS"; }});
+        attrs.put(VCFConstants.GENOTYPE_QUALITY_KEY, new AttributeGetter() { public Object get(Genotype g) { return g.getGQ(); }});
```
yfarjoun: what about DP? and what about the extended attributes?

bbimber: these would be captured by GenotypeJEXLContext.get(), correct?  DP could be added to the static map, though I'm not sure how one supports extended (variable) attributes there.  

I was assuming the primary point of the map was to support non-standard keys (i.e. isHom, isHomRef).  Otherwise, we defer back to Genotype.hasAnyAttribute().  This is comparable to what exists in VariantJEXLContext.get().

src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
```diff
@@ -0,0 +1,58 @@
...
+        this.g = g;
+    }
+
+    public Object get(String name) {
+        //should matching genotype attributes always supersede vc?
+        if ( attrs.containsKey(name) ) { // dynamic resolution of name -> value via map
+            return attrs.get(name).get(g);
+        } else if ( g.hasAnyAttribute(name) ) {
+            return g.getAnyAttribute(name);
+        } else if ( g.getFilters().contains(name) ) {
```
yfarjoun: This is a surprising behavior...is it possible to add a function that takes a filtername and responds with 1 or 0 depending on whether the filter is applied to the genotype?

src/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -55,68 +55,17 @@ private void initialize(Collection<JexlVCMatchExp> jexlCollection) {
...
-                infoMap.put(VCFConstants.GENOTYPE_FILTER_KEY, g.isFiltered()?  g.getFilters() : "PASS");
-
-                infoMap.put(VCFConstants.GENOTYPE_QUALITY_KEY, g.getGQ());
-                if ( g.hasDP() )
-                    infoMap.put(VCFConstants.DEPTH_KEY, g.getDP());
-                for ( Entry<String, Object> e : g.getExtendedAttributes().entrySet() ) {
-                    if ( e.getValue() != null && !e.getValue().equals(VCFConstants.MISSING_VALUE_v4) )
-                        infoMap.put(e.getKey(), e.getValue());
-                }
+        // QUESTION: the original code could have created a new VariantJEXLContext with a null vc.
```
yfarjoun: this is fine. remove comment.

yfarjoun:
back to you @bbimber

src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
yfarjoun: htsjdk is on java 8.

src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
yfarjoun: I just mentioned it since it was in the original code and so I feel that it would change the API.

src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
bbimber: I tend to agree, but it's exactly what VariantJEXLContext does.  With all my changes, I was trying to keep GenotypeJEXLContext behavior consistent with existing behavior in VariantJEXLContext.  Thoughts on consistency?

src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ */
+public class GenotypeJEXLContext extends VariantJEXLContext {
+    private Genotype g;
+
+    private interface AttributeGetter {
+        public Object get(Genotype g);
+    }
+
+    private static Map<String, AttributeGetter> attrs = new HashMap<String, AttributeGetter>();
```
yfarjoun: no reason to skimp on verbosity here, use attributes. Could you also change the name of the attributes in VariantJEXLContext from "x" to "attributes"?

src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
yfarjoun: hmmm OK. keep as before.

src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
bbimber: I believe the original code in JEXLMap actually had bugs here.  I think you're referring to lines 109-113 in the previous version:

```
                if ( g.hasDP() )
                    infoMap.put(VCFConstants.DEPTH_KEY, g.getDP());
                for ( Entry<String, Object> e : g.getExtendedAttributes().entrySet() ) {
                    if ( e.getValue() != null && !e.getValue().equals(VCFConstants.MISSING_VALUE_v4) )
                        infoMap.put(e.getKey(), e.getValue());

```
I'm not seeing how we put anything in the static map that is dependent on the Genotype itself, which I why I'm having GenotypeJEXLContext supply these in get().

Also, the original code special-cased some genotype fields and then deferred to Genotype.getExtendedAttributes() for the rest.  The problem is that FastGenotype and JEXLMap were not consistent about which fields they special case.  This mostly means that there was no way to ask for AD.  

The implementation I'm proposing will change behavior, but that's on purpose.  Any value in the Genotype should be available to JEXL.  I assume that was the original intent anyway.

bbimber:
I believe the last commit addresses all the comments

src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
```diff
@@ -20,23 +20,23 @@
...
-        attrs.put("isHomVar", new AttributeGetter() { public Object get(Genotype g) { return g.isHomVar() ? "1" : "0"; }});
-        attrs.put("isCalled", new AttributeGetter() { public Object get(Genotype g) { return g.isCalled() ? "1" : "0"; }});
-        attrs.put("isNoCall", new AttributeGetter() { public Object get(Genotype g) { return g.isNoCall() ? "1" : "0"; }});
-        attrs.put("isMixed", new AttributeGetter() { public Object get(Genotype g) { return g.isMixed() ? "1" : "0"; }});
-        attrs.put("isAvailable", new AttributeGetter() { public Object get(Genotype g) { return g.isAvailable() ? "1" : "0"; }});
-        attrs.put("isPassFT", new AttributeGetter() { public Object get(Genotype g) { return g.isFiltered() ? "0" : "1"; }});
-        attrs.put(VCFConstants.GENOTYPE_FILTER_KEY, new AttributeGetter() { public Object get(Genotype g) { return g.isFiltered()?  g.getFilters() : "PASS"; }});
-        attrs.put(VCFConstants.GENOTYPE_QUALITY_KEY, new AttributeGetter() { public Object get(Genotype g) { return g.getGQ(); }});
+        attributes.put("g", (Genotype g) -> g);
+        attributes.put(VCFConstants.GENOTYPE_KEY, Genotype::getGenotypeString);
```
yfarjoun: is the `(** Genotype** g) -> g` needed? will not g->g also work?

Also, g is now an overloaded variable. Could you rename the member one to be `genotype`? 

(both comments have clear parallels in the VC part.)

yfarjoun:
Thanks @bbimber I think this is great. It would be even better if you could add some tests that showcase usages beyond the very simple tests I see in VariantJEXLContextUnitTest....otherwise I made some minor stylistic comments. 

Thanks again.

src/java/htsjdk/variant/variantcontext/GenotypeJEXLContext.java
bbimber: FWIW, the checked in project IPR file has the languageLevel as 1.6.  If you're officially on 1.8, it might be nice to update that file (not in this PR).

yfarjoun: you are correct, though the README.md states that we are on JAVA8...

bbimber:
now with unit test

bbimber:
Hey - just wanted to check back on this.  So far as i can see the previous comments have been addressed, and I expanded the unit test to exercise genotype filtering.

yfarjoun:
thanks!

Issue #360, opened 2015-10-27
In maven central pom file lists compile time dependency on org.testng:testng:6.8.8


priimak:
Looking at this file https://repo1.maven.org/maven2/com/github/samtools/htsjdk/1.140/htsjdk-1.140.pom you can see compile scope dependency on org.testng:testng:6.8.8

```xml
<dependency>
<groupId>org.testng</groupId>
<artifactId>testng</artifactId>
<version>6.8.8</version>
</dependency>
```

but as far as I can see there is no compile time dependency on testng. If so, could that be resolved in a next version of htsjdk?

magicDGS:
This is fixed now, no?

lbergelson:
This was fixed in the move to gradle.

Issue #377, opened 2015-11-09
switch hdsjdk build system to gradle


lbergelson:
I recommend that we switch htsjdk to use gradle as our build system.

Ant is old and not very beautiful.  I know from personal experience and anecdotal experience from those working around me that the ant build isn't particularly well supported by intellij.

The GATK4 team has now gained experience working with gradle and it's been a very pleasant tool to use.  Code for common things is very short and readable, and more complex things are reasonable because it's written in groovy instead of XML.

The following is a simple build.gradle script that lets you pull down dependencies, build htsjdk, and run tests.  A slightly more fully featured version can be seen in my branch https://github.com/samtools/htsjdk/tree/lb_gradle_build.  (It has code to reduce test verbosity on travis, which is something that we've had problems controlling using ant, as well as hooks for coveralls.)

```
buildscript {
    repositories {
        mavenCentral()
    }
}

apply plugin: 'java'
apply plugin: 'idea'

repositories {
    mavenCentral()
}

dependencies {
    compile files("${System.properties['java.home']}/../lib/tools.jar")

    compile "org.apache.commons:commons-jexl:2.1.1"
    compile "commons-logging:commons-logging:1.1.1"
    compile "org.xerial.snappy:snappy-java:1.0.3-rc3"
    compile "org.apache.commons:commons-compress:1.4.1"
    compile "org.tukaani:xz:1.5"
    compile "org.apache.ant:ant:1.8.2"

    testCompile "org.testng:testng:6.8.8"
}

sourceCompatibility = 1.8
targetCompatibility = 1.8

version="2.0.0"

logger.info("build for version:" + version)
group = 'com.github.samtools'

jar {
    manifest {
        attributes 'Implementation-Title': 'HTSJDK',
                'Implementation-Vendor' : 'Broad Institute',
                'Implementation-Version': version
    }
}


test {
    useTestNG{}

    // set heap size for the test JVM(s)
    minHeapSize = "1G"
    maxHeapSize = "2G"
}

task wrapper(type: Wrapper) {
    gradleVersion = '2.8'
}

// This is a hack to disable the java 8 default javadoc lint until we fix the html formatting
if (JavaVersion.current().isJava8Compatible()) {
    tasks.withType(Javadoc) {
        options.addStringOption('Xdoclint:none', '-quiet')
    }
}

```

I think gradle's builds are much more readable and understandable than ant.  Intellij also has excellent gradle support which is helpful.  

[...]

PR #533, opened 2016-03-22
Bzip2 fixes in cram.io.ExternalCompression

bpow:
### Description

A couple of fixes to BZip2 support in h.s.cram.io.ExternalCompression:

1. The existing version of `bzip2` uses a BZip2CompressorInputStream. This is the Apache commons compress class for *uncompressing* from bzip2. This should be BZip2CompressorOutputStream. There wasn't a test for this before, so I added one

2. `unbzip2` was using CBZip2InputStream (from ant) when there is a perfectly good BZip2CompressorInputStream in commons compress which is already a dependency. The classpath dependency on ant had been removed in dd90c3e6 but was put back in d92460dd. The logic for its removal is similar to before, but now with an additional reason: the build.sbt that produces the maven artifact uses the whole ant jar file instead of the pared-down apache-ant-bzip2.jar, so that pulls in a few megs of dependency for downstream users when commons already provides bzip2 compress/decompress support

3. Not directly related, but testng is no longer a "compile" configuration dependency (no classes in the main code depend on it, to what I can tell... I think there used to be some), and can be marked as a "test" configuration dependency so downstream code won't automatically pull it in with maven. It would be helpful if someone could run sbt to make sure this works.

### Checklist

- [x] Code compiles correctly
- [x] New tests covering changes and new functionality _(and to demonstrate old class failing)_
- [x] All tests passing
- [?] Extended the README / documentation, if necessary

_not sure if any additional documentation is necessary..._

The pull request is split into three commits to show the logic:

1. The tests that show existing code failing
2. The fixes
3. Removal of dependencies.

I would be happy to squash, of course.


cmnbroad:
@bpow - thanks for finding (and fixing!) this issue. The current cram implementation doesn't appear to ever use bzip for compression, so this appears to have gone unnoticed. Its part of the spec so it should work in the case that we need to decompress a cram file that does use it.

The code itself looks good to me (@vadimzalunin can you have a look).

I'm not sure about the sbt changes though, or how to verify those. @lbergelson ??.

cmnbroad:
If we take this it will also fix https://github.com/samtools/htsjdk/issues/360.

build.sbt
```diff
@@ -20,9 +20,7 @@ libraryDependencies += "org.apache.commons" % "commons-compress" % "1.4.1"
 
 libraryDependencies += "org.tukaani" % "xz" % "1.5"
  
-libraryDependencies += "org.apache.ant" % "ant" % "1.8.2"
-
-libraryDependencies += "org.testng" % "testng" % "6.8.8"
+libraryDependencies += "org.testng" % "testng" % "6.8.8" % "test"
```
lbergelson: @cmnbroad  These changes look fine to me from an htsjdk perspective.  testng is only used in tests.  We could probably just remove it from the sbt completely since we don't actually use the sbt for anything other than publishing to maven.  

I can't see any usage of the ant dependency either.  It's possible it's needed for the ant build, but that would be unaffected by removing it.  Likewise, picard probably requires these be included by htsjdk, but since it doesn't access them through maven it won't matter if we remove them from the sbt.

bpow:
Rebased to current master/HEAD

cmnbroad:
@vadimzalunin - Can you want to take a look at these changes before we merge them since they mostly affect CRAM ?

cmnbroad:
@bpow Thanks for your patience. Can you squash this down to 2 commits (one with the compression code changes and a separate one with the sbt changes).

bpow:
Rebased and squashed.

Note that the second commit also removes the binary jar `lib/apache-ant-1.8.2-bzip2.jar` (it can be easy to overlook file deletions in git logs). It is possible that some downstream users could be relying on this jar being coalesced into the htsjdk jar, but if so they should be encouraged to switch to commons-compress or to depend on ant explicitly.

vadimzalunin:
looks good to me

cmnbroad:
@bpow I can't test these sbt changes, and I don't want to merge this without testing them. So we either need to take those out of the PR (which is fine since I think we're headed towards [gradle soon](https://github.com/samtools/htsjdk/issues/377) or we need to hold off until we get the local publish working.

bpow:
Rebased to move deletion of `lib/apache-ant-1.8.2-bzip2.jar` to the first commit and remove the second commit for now.

Should I make a separate issue/pull request for the sbt edits, knowing that it would be on the back burner until someone can test it or until unnecessary because of a switch to gradle? Is the sbt build used for maven publishing? If so, then not removing the dependencies I had listed before bloats the maven artifact.

lbergelson:
@cmnbroad publishLocal should be fixed now. #573   @bpow I'd recommend removing the ant requirement from the sbt if you're deleting it here.

bpow:
I'm not sure of the best way to proceed. Should I put the sbt changes back in this pull request or open a separate pull request for the sbt changes?

I'm not able to fully test the sbt changes either-- `sbt test` on my machine gives several test failures for java.lang.OutOfMemoryError. I notice that the `build.xml` increases memory to 2G for tests.

lbergelson:
@bpow Sorry for my slow response. I think you should add the sbt changes back in.  They're only for publishing to maven, and if the library is not need with this change than it should be removed in all places.  If downstream people depend on have it here than they'll have to fix it themselves.

bpow:
The removal of unneeded dependencies is back in, and everything is squashed down to one commit.

bpow:
@cmnbroad : I updated this for the gradle build process now that #383 has been committed.

Note that in addition to fixing the bzip2 compression forcram files, removing the compile dependency on ant results in a shadowJar that is 36% smaller.

bpow:
I'm trying not to be a pest... Is there any additional information I can supply regarding this PR, or has it just fallen off people's radar?

It seems to me like a relatively simple fix for a clear error in the code (complete with a unit test that demonstrates the problem and passes when the problem is fixed). I've been responsive to all suggestions made so far, and have kept the PR up-to-date with other changes in the codebase (including the switch to gradle).

People that may have input:

- @cmnbroad, to whom the PR is assigned
- @vadimzalunin, since it affects cram
- @lbergelson, since it affects the build dependencies in the build.gradle file
- @nh13, since his commit (d92460dd) was the one that added the ant dependency that had been removed in dd90c3e

It is rebased again to master.

cmnbroad:
@bpow We haven't forgotten about this, we've just been conservative regarding the dependency change (the code change itself I think is uncontroversial). I had hoped @lbergelson would review the last post-gradle PR, but he is away on vacation now, and I have successfully built with this. I'll merge this week if no one objects.

cmnbroad:
@bpow thanks for your help with this.

jmarshall:
Post 909381a31321deb1e411a34fca3d1f75e4fc2410, _src/tests/java/htsjdk/samtools/cram/io/ExternalCompressionTest.java_ and _testdata/htsjdk/samtools/io/bzip2-test.bz2_ are all on their lonesome and presumably in the wrong place.

PR #539, opened 2016-03-23
Modifications to IntelDeflater.c to avoid issues when multiple threads try to access

mishalinaik:
### Description

Added struct Stream in IntelDeflater.c which enables useIGZIP check to avoid potential issues when multiple threads try to access. 

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

… issues with multiple threads access


yfarjoun:
thank-you  @mishalinaik for the super-fast turnaround. I'll take it out for a spin on different machines because I'm paranoid and then I'll merge it in.

Issue #520, opened 2016-03-10
Appetite for a non-backwards compatible upgrade to core APIs?


tfenne:
Calling @akiezun @lbergelson @vadimzalunin @nh13 @yfarjoun @droazen (and anyone else who's interested)!

I'd like to start some discussion to see if there would be appetite in collaborating on a newer, non-backwards compatible API for some or all of the major APIs in htsjdk?  By _major_ I mean SAM/BAM/CRAM/etc. and also VCF/BCF, with a focus on the former first.

To be specific, what I'm proposing is that we embark on a path where we:

1. Leave the old APIs in place in the `samtools` and `variant` packages, for a long time, perhaps ever
2. Create new parallel packages (e.g. `samtools2` or `reads` etc.)
3. Spend a good amount of time coming up with clean designs for all the major parts (readers, writers, factories, records etc.) before we start implementing anything beyond interfaces/abstract classes.
4. Implement a battery of tests against the new APIs 
5. Implement all the new APIs by copy/pasting and refactoring code from the old implementations and/or writing new code, but _not_ linking/using classes from the old packages

Point 5 isn't meant to be absolute - i.e. it might make sense to move a bunch of utility classes into shared packages and use them directly, but rather to indicate that I mean this would not be writing wrappers around existing classes (like `SamReader` did).

My motivations for this are many-fold:

1. Many of the existing APIs are clunky.  This ranges from unwieldy method names (e.g. `getReadUnmappedFlag()` instead of `isUnmapped()`) to more cross-cutting design problems (e.g. `SAMRecord` being a class not an interface).
2. I would like to have alternative implementations of various classes like `SAMRecord` that optimize for, e.g., low-memory consumption vs. low-latency access.
3. I'd like to have uniform access to Files/Streams/URLs regardless of whether the contents is SAM, BAM, CRAM or SRA (currently support for anything other than files is rather type-dependent).
4. I'd like `htsjdk` APIs to work better with other JVM languages, particularly scala.  You can use it now (and I do), but I think there are likely ways we could make the APIs easier to consume from scala or elsewhere without making the Java usage worse.

I'm willing to dedicate time and effort to this _if_:

1. Others also see value in a much cleaner but not backwards compatible API
2. Others would engage in both discussion, and porting of some of the existing code
3. The new API would likely, over time, be adopted by GATK & Picard

Would anyone else be interested enough to participate and help?

lindenb:
Interfaces , Factories, etc.. : +1 
count me in, please

[...]

PR #542, opened 2016-03-24
improve variantcontext.* javadoc

lindenb:
### Description

* I've added a few sentences in the javadoc to highlight what is happening under the hood of some functions
* I've added a few 'final' keyworkds
* added the method 'hasFilter' in VariantContext and CommonInfo
* added some tests for CommonINfo
* fix getAttributeAsList where the array is a int[] or a double[]

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/java/htsjdk/variant/variantcontext/CommonInfo.java
+++ b/src/java/htsjdk/variant/variantcontext/CommonInfo.java
@@ -101,12 +101,12 @@ public final class CommonInfo implements Serializable {
         return filters == null ? NO_FILTERS : Collections.unmodifiableSet(filters);
     }
 
-    /** @return true if no filter has been defined  <code>(getFiltersMaybeNull()==null)</code> */
+    /** @return true if filters were applied and the VariantContext did not pass filters */
     public boolean filtersWereApplied() {
         return filters != null;
     }
 
-    /** @return true if any filter been defined  <code>(!filters.isEmpty())</code> */
+    /** @return true if any filter been defined. This method won't work for <code>PASS</code>, since it's not included in the set of filters */
     public boolean isFiltered() {
         return filters == null ? false : !filters.isEmpty();
     }
@@ -173,7 +173,7 @@ public final class CommonInfo implements Serializable {
      */
     public double getPhredScaledQual() { return (getLog10PError() * -10) + 0.0; }
 
-    /** set the Phred scaled quality score */
+    /** set the quality score */
     public void setLog10PError(final double log10PError) {
         if ( log10PError > 0 && log10PError != NO_LOG10_PERROR)
             throw new IllegalArgumentException("BUG: log10PError cannot be > 0 : " + this.log10PError);
@@ -203,18 +203,18 @@ public final class CommonInfo implements Serializable {
     }
 
     // todo -- define common attributes as enum
-    /** clear all attributes, and set 'map' as attributes */
+    /** clear all attributes, and set 'map' as attributes. Subsequent changes to <code>map</code>  won't be reflected in 'this' record */
     public void setAttributes(final Map<String, ?> map) {
         clearAttributes();
         putAttributes(map);
     }
 
-    /** shortcut of <code>putAttribute(key, value, false)</code> */
+    /** put a new attribute, raises an exception if the attribute exists */
     public void putAttribute(final String key, final Object value) {
         putAttribute(key, value, false);
     }
 
-    /** insert a new attribute, raises an exception if allowOverwrites==true and the attribute exists */
+    /** insert a new attribute, raises an exception if allowOverwrites==false and the attribute exists */
     public void putAttribute(final String key, final Object value, final boolean allowOverwrites) {
         if ( ! allowOverwrites && hasAttribute(key) )
             throw new IllegalStateException("Attempting to overwrite key->value binding: key = " + key + " this = " + this);
@@ -232,7 +232,7 @@ public final class CommonInfo implements Serializable {
         attributes.remove(key);
     }
 
-    /** insert the attibutes as map using <code>putAttribute(key,value,false)</code> */
+    /** insert the attibutes in the map using <code>putAttribute(key,value,false)</code> */
     public void putAttributes(final Map<String, ?> map) {
         if ( map != null ) {
             // for efficiency, we can skip the validation if the map is empty
@@ -248,7 +248,7 @@ public final class CommonInfo implements Serializable {
         }
     }
 
-    /** @return true if the key is present */
+    /** @return true if the specified attribute is present */
     public boolean hasAttribute(final String key) {
         return attributes.containsKey(key);
     }
@@ -305,13 +305,13 @@ public final class CommonInfo implements Serializable {
         final Object x = getAttribute(key);
         if ( x == null ) return defaultValue;
         if ( x instanceof String ) return (String)x;
-        return String.valueOf(x); // throws an exception if this isn't a string
+        return String.valueOf(x);
     }
 
     /**
      * return an attribute as an integer.
      * if given key is not found the defaultValue is returned.
-     * if the value is a String, the value of <code>Integer.parseInt((String)x)</code> is returned.
+     * if the value is a String, the value of <code>Integer.valueOf((String)x)</code> is returned.
      * If the value is not a String or an Integer, an exception is thrown
      * 
      * @param key the attribute key
@@ -329,8 +329,8 @@ public final class CommonInfo implements Serializable {
     /**
      * return an attribute as a double.
      * if given key is not found the defaultValue is returned.
-     * if the value is as an Integer, this value is returned.
-     * if the value is a String, the value of <code>Double.parseDouble((String)x)</code> is returned.
+     * if the value is as an Integer or a Douvle, the value is returned as a double.
+     * if the value is a String, the value of <code>Double.valueOf((String)x)</code> is returned.
      * If the value is not a Double, a String or an Integer, an exception is thrown
      * 
      * @param key the attribute key
@@ -348,7 +348,7 @@ public final class CommonInfo implements Serializable {
     /**
      * return an attribute as a boolean.
      * if given key is not found the defaultValue is returned.
-     * if the value is a String, the value of <code>Boolean.parseBoolean((String)x)</code> is returned.
+     * if the value is a String, the value of <code>Boolean.valueOf((String)x)</code> is returned.
      * If the value is not a Boolean or a String an exception is thrown
      * 
      * @param key the attribute key
--- a/src/java/htsjdk/variant/variantcontext/CommonInfo.java
+++ b/src/java/htsjdk/variant/variantcontext/CommonInfo.java
@@ -132,7 +132,7 @@ public final class CommonInfo implements Serializable {
             filters = new HashSet<String>();
 
         if ( filter == null ) throw new IllegalArgumentException("BUG: Attempting to add null filter " + this);
-        if ( hasFilter(filter) ) throw new IllegalArgumentException("BUG: Attempting to add duplicate filter " + filter + " at " + this);
+        if ( getFilters().contains(filter) ) throw new IllegalArgumentException("BUG: Attempting to add duplicate filter " + filter + " at " + this);
         filters.add(filter);
     }
 
@@ -285,21 +285,10 @@ public final class CommonInfo implements Serializable {
         as a Collections.singletonList if there is only one value */
     @SuppressWarnings("unchecked")
     public List<Object> getAttributeAsList(final String key) {
-        final Object o = getAttribute(key);
+        Object o = getAttribute(key);
         if ( o == null ) return Collections.emptyList();
-        else if ( o instanceof List ) return (List<Object>)o;
-        else if ( o instanceof int[]) {
-            final int array[]=(int[])o;
-            final List<Object> list = new ArrayList<>( array.length );
-            for( int i : array) list.add(i);
-            return list;
-        } else if ( o instanceof double[]) {
-            final double array[]=(double[])o;
-            final List<Object> list = new ArrayList<>( array.length );
-            for( double i : array) list.add(i);
-            return list;
-        }
-        else if ( o.getClass().isArray() ) return Arrays.asList((Object[])o);
+        if ( o instanceof List ) return (List<Object>)o;
+        if ( o.getClass().isArray() ) return Arrays.asList((Object[])o);
         return Collections.singletonList(o);
     }
 
@@ -331,16 +320,10 @@ public final class CommonInfo implements Serializable {
      * @return the attribute as an integer
      */
     public int getAttributeAsInt(final String key, final int defaultValue) {
-        final Object x = getAttribute(key);
+        Object x = getAttribute(key);
         if ( x == null || x == VCFConstants.MISSING_VALUE_v4 ) return defaultValue;
         if ( x instanceof Integer ) return (Integer)x;
-        if ( !( x instanceof String ) ) throw new IllegalArgumentException(
-                "Cannot get attribute(key=\"" + key + ") as an integer because it's not a Number or a String. Class = " + x.getClass());
-        try {
-            return Integer.parseInt((String)x);
-        } catch (final NumberFormatException err) {
-            throw new IllegalArgumentException( "Cannot convert attribute(key=\"" + key + ") = "+ x + " as an integer.", err);
-        }
+        return Integer.valueOf((String)x); // throws an exception if this isn't a string
     }
     
     /**
@@ -355,17 +338,11 @@ public final class CommonInfo implements Serializable {
      * @return the attribute as a double
      */
     public double getAttributeAsDouble(final String key, final double defaultValue) {
-        final Object x = getAttribute(key);
+        Object x = getAttribute(key);
         if ( x == null ) return defaultValue;
         if ( x instanceof Double ) return (Double)x;
         if ( x instanceof Integer ) return (Integer)x;
-        if ( !( x instanceof String ) ) throw new IllegalArgumentException(
-                "Cannot get attribute(key=\"" + key + ") as a double because it's not a Double, a Integer or a String. Class = " + x.getClass());
-        try {
-            return Double.parseDouble((String)x);
-        } catch (final NumberFormatException err) {
-            throw new IllegalArgumentException( "Cannot convert attribute(key=\"" + key + ") = "+ x + " as a double.", err);
-        }
+        return Double.valueOf((String)x); // throws an exception if this isn't a string
     }
 
     /**
@@ -379,27 +356,10 @@ public final class CommonInfo implements Serializable {
      * @return the attribute as a boolean
      */
     public boolean getAttributeAsBoolean(final String key, final boolean defaultValue) {
-        final Object x = getAttribute(key);
+        Object x = getAttribute(key);
         if ( x == null ) return defaultValue;
         if ( x instanceof Boolean ) return (Boolean)x;
-        if ( !( x instanceof String ) ) throw new IllegalArgumentException(
-                "Cannot get attribute(key=\"" + key + ") as a boolean because it's not a Boolean or a String. Class = " + x.getClass());
-        return Boolean.parseBoolean((String)x);
-    }
-
-    @Override
-    public String toString() {
-        return new StringBuilder().
-            append("CommonInfo [log10PError=").
-            append(log10PError).
-            append(", name=").
-            append(name).
-            append(", filters=").
-            append(filters).
-            append(", attributes=").
-            append(attributes).
-            append("]").
-            toString();
+        return Boolean.valueOf((String)x); // throws an exception if this isn't a string
     }
 
 //    public String getAttributeAsString(String key)      { return (String.valueOf(getExtendedAttribute(key))); } // **NOTE**: will turn a null Object into the String "null"
--- a/src/java/htsjdk/variant/variantcontext/CommonInfo.java
+++ b/src/java/htsjdk/variant/variantcontext/CommonInfo.java
@@ -106,14 +106,14 @@ public final class CommonInfo implements Serializable {
         return filters != null;
     }
 
-    /** @return true if any filter been defined. This method won't work for <code>PASS</code>, since it's not included in the set of filters */
+    /** @return true if any filter been defined*/
     public boolean isFiltered() {
         return filters == null ? false : !filters.isEmpty();
     }
 
     /**
      * @param filter the filter ID 
-     * @return true if the filters contains 'filter' */
+     * @return true if the filters contains 'filter'. This method won't work for <code>PASS</code>, since it's not included in the set of filters */
     public boolean hasFilter(final String filter) {
         return filters == null ? false : filters.contains(filter);
     }
--- a/src/java/htsjdk/variant/variantcontext/CommonInfo.java
+++ b/src/java/htsjdk/variant/variantcontext/CommonInfo.java
@@ -29,6 +29,7 @@ package htsjdk.variant.variantcontext;
 import htsjdk.variant.vcf.VCFConstants;
 
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
@@ -57,7 +58,7 @@ public final class CommonInfo implements Serializable {
     private Set<String> filters = null;
     private Map<String, Object> attributes = NO_ATTRIBUTES;
 
-    public CommonInfo(String name, double log10PError, Set<String> filters, Map<String, Object> attributes) {
+    public CommonInfo(final String name, final double log10PError, final Set<String> filters, final Map<String, Object> attributes) {
         this.name = name;
         setLog10PError(log10PError);
         this.filters = filters;
@@ -78,7 +79,7 @@ public final class CommonInfo implements Serializable {
      *
      * @param name    the name associated with this information
      */
-    public void setName(String name) {
+    public void setName(final String name) {
         if ( name == null ) throw new IllegalArgumentException("Name cannot be null " + this);
         this.name = name;
     }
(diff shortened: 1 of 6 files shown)


droazen:
@ldgauthier Would you be able to review this one?

ldgauthier:
Can do.  I should be able to get around to it by the end of next week.

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -151,7 +173,8 @@ public boolean hasLog10PError() {
      */
     public double getPhredScaledQual() { return (getLog10PError() * -10) + 0.0; }
 
-    public void setLog10PError(double log10PError) {
+    /** set the Phred scaled quality score */
```
ldgauthier: Is this value Phred-scaled at this point?  It will become the QUAL (in the writer?), but I thought it's just stored as log10prob.

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -90,38 +91,58 @@ public void setName(String name) {
...
     public Set<String> getFiltersMaybeNull() {
         return filters;
     }
 
+    /** @return an unmodifiable Set of filters. Can be empty by never null */
     public Set<String> getFilters() {
         return filters == null ? NO_FILTERS : Collections.unmodifiableSet(filters);
     }
 
+    /** @return true if no filter has been defined  <code>(getFiltersMaybeNull()==null)</code> */
```
ldgauthier: It would be nice to note somewhere that in the parsing in the VCF codec if a site is unfiltered (represented by VCFConstants.UNFILTERED=".") then the filters list is set to null and that if the VC passes filters ("PASS"), its common info gets an empty Set<String>.

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -721,33 +721,50 @@ public String getID() {
     // get routines to access context info fields
     //
     // ---------------------------------------------------------------------------------------------------------
+    /** shortcut to {@link CommonInfo#getName()} */
```
ldgauthier: I would rather have the javadoc in VariantContext.java instead of CommonInfo.  Or both.

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -247,41 +284,122 @@ public Object getAttribute(String key, Object defaultValue) {
...
         as a Collections.singletonList if there is only one value */
     @SuppressWarnings("unchecked")
-    public List<Object> getAttributeAsList(String key) {
-        Object o = getAttribute(key);
+    public List<Object> getAttributeAsList(final String key) {
+        final Object o = getAttribute(key);
         if ( o == null ) return Collections.emptyList();
-        if ( o instanceof List ) return (List<Object>)o;
-        if ( o.getClass().isArray() ) return Arrays.asList((Object[])o);
+        else if ( o instanceof List ) return (List<Object>)o;
```
ldgauthier: I appreciate that this is a problem, but I hate to add more instanceof calls.  I guess that's an implicit vote for the proposal by @tfenne in #520

lindenb:
@ldgauthier back to you and thank you for the review.  I've copied the javadoc from CommonInfo to the shotcuts in VariantContext.
Regarding  your other comments, there is not much I can do  isnt't it ? :-) ( setLog10PError , 'instanceof' calls...)

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -721,45 +721,122 @@ public String getID() {
...
-    /** shortcut to {@link CommonInfo#hasLog10PError()} */
+    
+    /** shortcut to {@link CommonInfo#hasLog10PError()}
+     * @return true if log10-based error estimate has been set 
+     */
     public boolean hasLog10PError()             { return commonInfo.hasLog10PError(); }
-    /** shortcut to {@link CommonInfo#getLog10PError()} */
+    
+    /** shortcut to {@link CommonInfo#getLog10PError()}
+     * @return the -1 * log10-based error estimate
```
ldgauthier: Where does the -1 come from?

lindenb:
@ldgauthier 
> Where does the -1 come from?

it's here: https://github.com/samtools/htsjdk/blob/master/src/java/htsjdk/variant/variantcontext/CommonInfo.java#L139

ldgauthier:
I don't see where in the code the sign is flipped, except in getPhredScaledQual, which is a -10. Log10PError should be negative and that method just returns it. @droazen is it possible you made a mistake in that original comment?

ldgauthier:
I'm still hesitant to add more instanceof calls, but I guess it's just a question of doing them here or in whatever package is working with the VCF.  This way seems safer. @droazen Do you have any objections? Ultimately, it would be nice to be able to use the type declared in the header for each info field attribute, but that's a pretty significant change.

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -90,38 +91,58 @@ public void setName(String name) {
     //
     // ---------------------------------------------------------------------------------------------------------
 
+    /** @return a <b>modifiable</b> Set of filters. Can be  null. All changes in this set will be reflected in the CommonInfo */
```
nh13: extra space `be__null`

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -90,38 +91,58 @@ public void setName(String name) {
...
     public Set<String> getFiltersMaybeNull() {
         return filters;
     }
 
+    /** @return an unmodifiable Set of filters. Can be empty by never null */
```
nh13: typo (`by`)

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -247,41 +284,122 @@ public Object getAttribute(String key, Object defaultValue) {
...
+     * if given key is not found the defaultValue is returned.
+     * if the value is as an Integer, this value is returned.
+     * if the value is a String, the value of <code>Double.parseDouble((String)x)</code> is returned.
+     * If the value is not a Double, a String or an Integer, an exception is thrown
+     * 
+     * @param key the attribute key
+     * @param defaultValue the default value
+     * @return the attribute as a double
+     */
+    public double getAttributeAsDouble(final String key, final double defaultValue) {
```
nh13: I am so confused why in `VariantContext.java` we have these `getAttributeAs*` methods but in `Genotype.java` the similar methods are deprecated and implemented differently.  Likely this is for a different PR, but perhaps with some consensus a new PR could emerge?

SHuang-Broad:
To add a somewhat relevant comment, if one looks at the first constructor in the `Allele` classes--`protected Allele(final byte[] bases, final boolean isRef)`--it says "null alleles are no longer allowed".

But the documentation for the public method `public static Allele create(final byte[] bases, final boolean isRef)` says "If bases == '-', a Null allele is created", which seems like an outdated doc, because the code in the `create` method doesn't actually check for the case of dash valued bases.

src/java/htsjdk/variant/variantcontext/CommonInfo.java
lbergelson: I recently ran into that and was also confused.

yfarjoun:
@lindenb This PR claims to be a documentation PR, but has many code changes, not all of which are trivial. Could you please split it into two separate PRs one with the doc changes and another one with the code changes together with appropriate tests for the new functionality?

src/java/htsjdk/variant/variantcontext/CommonInfo.java
lindenb: @lbergelson what do you mean ? do you mean you don't understand the original javadoc or you don't understand my changes ?

lbergelson: Oh, I was agreeing with @nh13  that we had weird api differences between `VariantContext` and `Genotype` that should probably be looked at in a different PR.  Not a comment on the doc at all.

lindenb:
@yfarjoun I've removed the code changes. I've only kept the 'final' keywords. Shall I remove them too ?

yfarjoun:
I think it's fine to leave the 'final's. thanks.

On Wed, Jun 29, 2016 at 5:57 AM, Pierre Lindenbaum <notifications@github.com

src/java/htsjdk/variant/variantcontext/CommonInfo.java
droazen: This comment is backwards -- the method returns true if `filters != null`, not if "no filter has been defined"

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -90,27 +91,43 @@ public void setName(String name) {
...
     public Set<String> getFilters() {
         return filters == null ? NO_FILTERS : Collections.unmodifiableSet(filters);
     }
 
+    /** @return true if no filter has been defined  <code>(getFiltersMaybeNull()==null)</code> */
     public boolean filtersWereApplied() {
         return filters != null;
     }
 
+    /** @return true if any filter been defined  <code>(!filters.isEmpty())</code> */
```
droazen: Better: "true if filters were applied and the VC did not pass filters"

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -90,27 +91,43 @@ public void setName(String name) {
...
     public boolean isFiltered() {
         return filters == null ? false : !filters.isEmpty();
     }
 
+    /**
+     * @param filter the filter ID 
+     * @return true if the filters contains 'filter' */
```
droazen: Document that this method won't work for `PASS`, since it's not included in the set of filters

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -151,7 +173,8 @@ public boolean hasLog10PError() {
      */
     public double getPhredScaledQual() { return (getLog10PError() * -10) + 0.0; }
 
-    public void setLog10PError(double log10PError) {
+    /** set the Phred scaled quality score */
+    public void setLog10PError(final double log10PError) {
```
droazen: The comment for this method is incorrect -- `log10PError` is not Phred-scaled, as @ldgauthier suspected. You can see this by looking at the implementation of the `getPhredScaledQual()` method above.

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -166,29 +189,33 @@ public void setLog10PError(double log10PError) {
...
      */
     public Map<String, Object> getAttributes() {
         return Collections.unmodifiableMap(attributes);
     }
 
     // todo -- define common attributes as enum
-
-    public void setAttributes(Map<String, ?> map) {
+    /** clear all attributes, and set 'map' as attributes */
+    public void setAttributes(final Map<String, ?> map) {
```
droazen: Document whether subsequent changes to the `map` passed in will be reflected in this record (they won't).

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -166,29 +189,33 @@ public void setLog10PError(double log10PError) {
...
         clearAttributes();
         putAttributes(map);
     }
 
-    public void putAttribute(String key, Object value) {
+    /** shortcut of <code>putAttribute(key, value, false)</code> */
```
droazen: Add proper javadoc for this method (and for the other methods where you've written "shortcut of x")

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -166,29 +189,33 @@ public void setLog10PError(double log10PError) {
...
+    public void putAttribute(final String key, final Object value) {
         putAttribute(key, value, false);
     }
 
-    public void putAttribute(String key, Object value, boolean allowOverwrites) {
+    /** insert a new attribute, raises an exception if allowOverwrites==true and the attribute exists */
```
droazen: This isn't right -- the exception is raised if allowOverwrites is **false** and the attribute exists.

droazen: Also, use a proper javadoc-style `@throws` to document exceptions

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -198,31 +225,35 @@ public void putAttribute(String key, Object value, boolean allowOverwrites) {
...
-    public void removeAttribute(String key) {
+    /** removes the attribute identified by key */
+    public void removeAttribute(final String key) {
         if ( attributes == NO_ATTRIBUTES ) // immutable -> mutable
             attributes = new HashMap<String, Object>();
         attributes.remove(key);
     }
 
-    public void putAttributes(Map<String, ?> map) {
+    /** insert the attibutes as map using <code>putAttribute(key,value,false)</code> */
```
droazen: "as map" -> "in the map"

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -198,31 +225,35 @@ public void putAttribute(String key, Object value, boolean allowOverwrites) {
...
-                for ( Map.Entry<String, ?> elt : map.entrySet() ) {
+                for ( final Map.Entry<String, ?> elt : map.entrySet() ) {
                     putAttribute(elt.getKey(), elt.getValue(), false);
                 }
             }
         }
     }
 
-    public boolean hasAttribute(String key) {
+    /** @return true if the key is present */
```
droazen: key -> attribute

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -247,37 +284,78 @@ public Object getAttribute(String key, Object defaultValue) {
...
+     * 
+     * @param key the attribute key
+     * @param defaultValue  the default value
+     * @return the attribute as a String
+     */
+    public String getAttributeAsString(final String key, final String defaultValue) {
+        final Object x = getAttribute(key);
         if ( x == null ) return defaultValue;
         if ( x instanceof String ) return (String)x;
         return String.valueOf(x); // throws an exception if this isn't a string
```
droazen: This comment "throws an exception if this isn't a string" isn't right -- `String.valueOf(x)` calls `x.toString()`, regardless of whether x is already a String or not.

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -247,37 +284,78 @@ public Object getAttribute(String key, Object defaultValue) {
...
     }
 
-    public int getAttributeAsInt(String key, int defaultValue) {
+    /**
+     * return an attribute as an integer.
+     * if given key is not found the defaultValue is returned.
+     * if the value is a String, the value of <code>Integer.parseInt((String)x)</code> is returned.
```
droazen: `parseInt` -> `valueOf`

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -247,37 +284,78 @@ public Object getAttribute(String key, Object defaultValue) {
...
         return Integer.valueOf((String)x); // throws an exception if this isn't a string
     }
-
-    public double getAttributeAsDouble(String key, double defaultValue) {
+    
+    /**
+     * return an attribute as a double.
+     * if given key is not found the defaultValue is returned.
+     * if the value is as an Integer, this value is returned.
+     * if the value is a String, the value of <code>Double.parseDouble((String)x)</code> is returned.
```
droazen: `parseDouble` -> `valueOf`

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -247,37 +284,78 @@ public Object getAttribute(String key, Object defaultValue) {
...
         if ( x instanceof Integer ) return (Integer)x;
         return Integer.valueOf((String)x); // throws an exception if this isn't a string
     }
-
-    public double getAttributeAsDouble(String key, double defaultValue) {
+    
+    /**
+     * return an attribute as a double.
+     * if given key is not found the defaultValue is returned.
+     * if the value is as an Integer, this value is returned.
```
droazen: "if the value is an Integer or Double, it is returned as a Double"

src/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -247,37 +284,78 @@ public Object getAttribute(String key, Object defaultValue) {
...
         if ( x instanceof Double ) return (Double)x;
         if ( x instanceof Integer ) return (Integer)x;
         return Double.valueOf((String)x); // throws an exception if this isn't a string
     }
 
-    public boolean getAttributeAsBoolean(String key, boolean defaultValue) {
+    /**
+     * return an attribute as a boolean.
+     * if given key is not found the defaultValue is returned.
+     * if the value is a String, the value of <code>Boolean.parseBoolean((String)x)</code> is returned.
```
droazen: `parseBoolean` -> `valueOf`

src/java/htsjdk/variant/variantcontext/Genotype.java
```diff
@@ -181,6 +181,10 @@ public boolean hasDP() {
     // ---------------------------------------------------------------------------------------------------------
 
     /**
+     * if {@link #getAlleles()} is empty, the type is {@link GenotypeType#UNAVAILABLE}.
+     * if all the  alleles are NoCall,  the type is {@link GenotypeType#NO_CALL}.
+     * if some, but not all, alleles are NoCall,  the type is {@link GenotypeType#MIXED}.
+     *
```
droazen: Your doc for this method is missing any description of what happens when **none** of the alleles are no call (ie., you get `GenotypeType.HET`, `GenotypeType.HOM_REF`, or `GenotypeType.HOM_VAR` as appropriate).

src/java/htsjdk/variant/variantcontext/Genotype.java
```diff
@@ -255,16 +259,19 @@ else if ( !allele.equals(observedAllele) )
     public boolean isNoCall() { return getType() == GenotypeType.NO_CALL; }
 
     /**
+     * shortcut of ( {@link #getType()} != NO_CALL AND {@link #getType()} != UNAVAILABLE )
```
droazen: `NO_CALL` -> `{@link GenotypeType#NO_CALL}`
`UNAVAILABLE` -> `{@link GenotypeType#UNAVAILABLE}`

src/java/htsjdk/variant/variantcontext/Genotype.java
```diff
@@ -255,16 +259,19 @@ else if ( !allele.equals(observedAllele) )
...
      * @return true if this genotype is comprised of any alleles that are not no-calls (even if some are).
      */
     public boolean isCalled() { return getType() != GenotypeType.NO_CALL && getType() != GenotypeType.UNAVAILABLE; }
 
     /**
+     * shortcut of ( {@link #getType()} == MIXED )
```
droazen: `MIXED` -> `{@link GenotypeType#MIXED}`

src/java/htsjdk/variant/variantcontext/Genotype.java
```diff
@@ -255,16 +259,19 @@ else if ( !allele.equals(observedAllele) )
...
      * @return true if this genotype is comprised of both calls and no-calls.
      */
     public boolean isMixed() { return getType() == GenotypeType.MIXED; }
 
     /**
+     * shortcut of ( {@link #getType()} != UNAVAILABLE )
```
droazen: `UNAVAILABLE` -> `{@link GenotypeType#UNAVAILABLE}`

src/java/htsjdk/variant/variantcontext/GenotypesContext.java
```diff
@@ -660,6 +660,7 @@ public boolean containsSample(final String sample) {
         return sampleNameToOffset.containsKey(sample);
     }
 
+    /** return true if all {@link GenotypesContext#getSampleNames() contains all sample */
```
droazen: Wording is awkward here -- "return true if the list of samples in this context as returned by `{@link #getSampleNames()}` contains all samples in the Collection passed in"

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -721,33 +721,125 @@ public String getID() {
...
+     */
     public String getSource()                   { return commonInfo.getName(); }
+    
+    /** shortcut to {@link CommonInfo#getFiltersMaybeNull()} 
+     *  @return a <b>modifiable</b> Set of filters. Can be  null. All changes in this set will be reflected in the CommonInfo 
+     */
     public Set<String> getFiltersMaybeNull()    { return commonInfo.getFiltersMaybeNull(); }
+    
+    /** shortcut to {@link CommonInfo#getFilters()} 
+     * @return an unmodifiable Set of filters. Can be empty by never null
```
droazen: by -> but

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -721,33 +721,125 @@ public String getID() {
...
+     */
     public Set<String> getFilters()             { return commonInfo.getFilters(); }
+    
+    /** shortcut to {@link CommonInfo#isFiltered()}
+     * @return true if any filter been defined  <code>(!filters.isEmpty())
```
droazen: "true if filters were applied and the VC did not pass filters"

Fix broken `<code>` tag

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -721,33 +721,125 @@ public String getID() {
...
+     */
     public boolean isFiltered()                 { return commonInfo.isFiltered(); }
+    
+    /** shortcut to {@link CommonInfo#isNotFiltered()} */
     public boolean isNotFiltered()              { return commonInfo.isNotFiltered(); }
+    
+    /** shortcut to {@link CommonInfo#filtersWereApplied()}
+     *  @return true if no filter has been defined  <code>(getFiltersMaybeNull()==null)</code>
```
droazen: This is backwards -- this returns true if `filters != null`

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -721,33 +721,125 @@ public String getID() {
...
+     * @return an unmodifiable Set of filters. Can be empty by never null
+     */
     public Set<String> getFilters()             { return commonInfo.getFilters(); }
+    
+    /** shortcut to {@link CommonInfo#isFiltered()}
+     * @return true if any filter been defined  <code>(!filters.isEmpty())
+     */
     public boolean isFiltered()                 { return commonInfo.isFiltered(); }
+    
+    /** shortcut to {@link CommonInfo#isNotFiltered()} */
```
droazen: "equivalent to ! isFiltered()"

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -721,33 +721,125 @@ public String getID() {
...
+     */
     public boolean filtersWereApplied()         { return commonInfo.filtersWereApplied(); }
+    
+    /** shortcut to {@link CommonInfo#hasLog10PError()}
+     * @return true if log10-based error estimate has been set 
+     */
     public boolean hasLog10PError()             { return commonInfo.hasLog10PError(); }
+    
+    /** shortcut to {@link CommonInfo#getLog10PError()}
+     * @return the -1 * log10-based error estimate
```
droazen: This isn't right -- it's just log10(perror), not -1 * log10(perror). You can see this is true because in `commonInfo.getPhredScaledQual()` we multiply `getLog10PError()` by -10.

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -721,33 +721,125 @@ public String getID() {
...
     public double getPhredScaledQual()          { return commonInfo.getPhredScaledQual(); }
-
+    
+    /** shortcut to {@link CommonInfo#getAttributes()}
+     * @return an unmodifiable Map of attributes
+     */
     public Map<String, Object>  getAttributes() { return commonInfo.getAttributes(); }
+    
+    /** shortcut to {@link CommonInfo#hasAttribute(String)}
+     * @return true if the key is present
```
droazen: key -> specified attribute

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -721,33 +721,125 @@ public String getID() {
...
+    /** shortcut to {@link CommonInfo#getAttribute(String, Object)}
+     * @return the attribute value for the given key (or defaultValue if not set)
+     */
     public Object getAttribute(String key, Object defaultValue) {
         return commonInfo.getAttribute(key, defaultValue);
     }
-
+    
+    /** shortcut to {@link CommonInfo#getAttributeAsString(String, String)}
+     * return an attribute as a String.
```
droazen: You may want to use the javadoc `@see` tag in these methods to reference the appropriate docs in `CommonInfo`

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -721,33 +721,125 @@ public String getID() {
...
+     * if given key is not found the defaultValue is returned.
+     * if the value is not a String <code> String.valueOf(x)</code> is returned.
+     * @return the attribute as a String
+     */
     public String getAttributeAsString(String key, String defaultValue)   { return commonInfo.getAttributeAsString(key, defaultValue); }
+    
+    /** shortcut to {@link CommonInfo#getAttributeAsInt(String, int)}
+     * if given key is not found the defaultValue is returned.
```
droazen: All the comments I made in `CommonInfo` for these `getAttributeAs*` methods should be applied here as welll.

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -759,17 +851,18 @@ public CommonInfo getCommonInfo() {
...
-        Allele ref = REF;
+        final Allele ref = REF;
         if ( ref == null )
             throw new IllegalStateException("BUG: no reference allele found at " + this);
         return ref;
     }
 
 
     /**
+     * shortcut to <code>{@link #getNAlleles()} == 2</code<
```
droazen: Fix broken `</code<` tag

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -940,7 +1035,7 @@ public boolean hasGenotypes() {
         return ! genotypes.isEmpty();
     }
 
-    public boolean hasGenotypes(Collection<String> sampleNames) {
+    public boolean hasGenotypes(final Collection<String> sampleNames) {
```
droazen: No docs for this one?

src/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -997,6 +1092,9 @@ public GenotypesContext getGenotypes(Set<String> sampleNames) {
         return getGenotypes().getSampleNames();
     }
 
+    /**
+     * @return the list of all sample names in this context, ordered
```
droazen: ordered -> ordered by name

droazen:
Review complete, back to @lindenb for changes.

yfarjoun:
Friendly poke... @lindenb we would be very happy to merge once you respond to the comments.

lindenb:
Thanks & back to you, all.
I've move the comments in VariantContext to  `@see CommonInfo#...`

Thanks
P.

lindenb:
ok, I'm going to close that one for now, this is currently diverging too much from the master.
But I thank the reviewers for their time.

Issue #512, opened 2016-03-03
BAMIndexer should throw if file is not coordinate sorted


lbergelson:
Since creating a `.bai` index requires that a bam be coordinate sorted, it seems like `BAMIndexer` should check that the file it is indexing is in the correct order.  Since it has access to the `SAMHeader` it can trivially check this for essentially no cost.

Issue #402, opened 2015-12-03
SAMRecord equals should not use reference equality for attributes


cmnbroad:
SAMRecord.equals should use value equality for attributes now that we have deepCopy. As it is now, rec.deepCopy().equals(rec) always fails if the record has attributes.

cmnbroad:
This is actually overstated; its only array valued attributes where the reference equality fails to do the right thing.

cmnbroad:
@vadimzalunin In the course of trying to fix this bug, I'm trying to tighten up the contract for SAMBinaryTagAndValue and SAMBinaryTagAndUnsignedArrayValue (by, for instance, moving the isAllowedAttributeValue test code from SAMRecord into these classes, since restricting the types to the known list is required in order to get the copy/hash/equality semantics of those classes consistent).

But we have tests that fail when I do this, and the test comments indicate that we're depending on being able to [circumvent such guards](https://github.com/samtools/htsjdk/blob/master/src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java#L376). It looks like they're CRAM related - is there something about CRAM that assumes that we can populate a SAMBinaryTagAndUnsignedArray object with a value that is neither [unsigned nor an array](https://github.com/samtools/htsjdk/blob/master/src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java#L379) ? These tests are exploiting an inconsistency in the attribute APIs that I'd like to close off, but I can't tell if its just a test convenience or if CRAM is really dependent on it (in which case I'd like to find another way).

vadimzalunin:
This is for test convenience. I think I wanted to test that an out of range value is prohibited explicitly by the SAMRecord checks and everything else does not impose such restrictions. Perhaps this is not required any more.

cmnbroad:
Fixed via https://github.com/samtools/htsjdk/pull/543. As part of this change the checks for allowed attribute types and values have been moved inside of the SAMBinaryTagAndValue classes.

PR #543, opened 2016-03-30
Bug fixes for SAMRecord, SAMBinaryTagAndValue, and BAMIndexer.

cmnbroad:
Fixes https://github.com/samtools/htsjdk/issues/512, https://github.com/samtools/htsjdk/issues/402, and a couple of gratuitous javadoc fixes.

### DIFF
--- a/src/java/htsjdk/samtools/BAMIndexer.java
+++ b/src/java/htsjdk/samtools/BAMIndexer.java
@@ -27,6 +27,7 @@ import htsjdk.samtools.util.Log;
 
 import java.io.File;
 import java.io.OutputStream;
+import java.util.function.Function;
 
 /**
  * Class for both constructing BAM index content and writing it out.
@@ -49,15 +50,14 @@ public class BAMIndexer {
     // content is built up from the input bam file using this
     private final BAMIndexBuilder indexBuilder;
 
+    private static final Log log = Log.getInstance(BAMIndexer.class);
+
     /**
      * @param output     binary BAM Index (.bai) file
      * @param fileHeader header for the corresponding bam file
      */
     public BAMIndexer(final File output, final SAMFileHeader fileHeader) {
-
-        numReferences = fileHeader.getSequenceDictionary().size();
-        indexBuilder = new BAMIndexBuilder(fileHeader.getSequenceDictionary());
-        outputWriter = new BinaryBAMIndexWriter(numReferences, output);
+        this(fileHeader, numRefs -> new BinaryBAMIndexWriter(numRefs, output));
     }
 
     /**
@@ -67,10 +67,28 @@ public class BAMIndexer {
      * @param fileHeader header for the corresponding bam file.
      */
     public BAMIndexer(final OutputStream output, final SAMFileHeader fileHeader) {
+        this(fileHeader, numRefs -> new BinaryBAMIndexWriter(numRefs, output));
+    }
 
+    /*
+     * Prepare to index a BAM.
+     *
+     * @param fileHeader header for the corresponding bam file.
+     * @param  createWrite a lambda that, given an Integer numReferences value, will create a BinaryBAMIndexWriter
+     *                     with that value and an appropriate output.
+      */
+    private BAMIndexer(final SAMFileHeader fileHeader, Function<Integer, BinaryBAMIndexWriter> createWriter) {
+        if (fileHeader.getSortOrder() != SAMFileHeader.SortOrder.coordinate) {
+            if (fileHeader.getSortOrder() == SAMFileHeader.SortOrder.unsorted) {
+                log.warn("For indexing, the BAM file is required to be coordinate sorted. Attempting to index \"unsorted\" BAM file.");
+            }
+            else {
+                throw new SAMException("Indexing requires a coordinate-sorted input BAM.");
+            }
+        }
         numReferences = fileHeader.getSequenceDictionary().size();
         indexBuilder = new BAMIndexBuilder(fileHeader.getSequenceDictionary());
-        outputWriter = new BinaryBAMIndexWriter(numReferences, output);
+        outputWriter = createWriter.apply(numReferences);
     }
 
     /**
--- a/src/java/htsjdk/samtools/SAMBinaryTagAndValue.java
+++ b/src/java/htsjdk/samtools/SAMBinaryTagAndValue.java
@@ -24,6 +24,7 @@
 package htsjdk.samtools;
 
 import java.io.Serializable;
+import java.util.Arrays;
 
 /**
  * Holds a SAMRecord attribute and the tagname (in binary form) for that attribute.
@@ -31,7 +32,12 @@ import java.io.Serializable;
  * See SAMTagUtil to convert the tag to String form.
  *
  * Values associated with attribute tags must be of a type that implements {@link Serializable} or else
- * serialization will fail.
+ * serialization will fail. Accepted types are String, scalar types Short, Integer, Character, Float,
+ * and Long (see below); array types byte[], short[], int[] and float[]. Cannot be null.
+ *
+ * Long valued attributes are constrained to the range [Integer.MIN_VALUE, BinaryCodec.MAX_UINT],
+ * which includes the entire range of signed ints [Integer.MIN_VALUE, Integer.MAX_VALUE] and
+ * the entire range of unsigned ints that can be stored per the BAM spec [0, (Integer.MAX_VALUE * 2) + 1].
  *
  * @author alecw@broadinstitute.org
  */
@@ -44,15 +50,44 @@ public class SAMBinaryTagAndValue implements Serializable {
 
     /**
      * @param tag tagname (in binary form) for this attribute
-     * @param value value for this attribute (must be of a type that implements {@link Serializable} or else serialization will fail)
-     *              Cannot be null.
+     * @param value value for this attribute (must be of a type that implements {@link Serializable}
+     *              or else serialization will fail). Cannot be null.
      */
     public SAMBinaryTagAndValue(final short tag, final Object value) {
         if (null == value) {
             throw new IllegalArgumentException("SAMBinaryTagAndValue value may not be null");
         }
+        if (!isAllowedAttributeValue(value)) {
+            throw new IllegalArgumentException("Attribute type " + value.getClass() + " not supported. Tag: " +
+                    SAMTagUtil.getSingleton().makeStringTag(tag));
+        }
         this.tag = tag;
         this.value = value;
+    }
+
+    // Inspect the proposed value to determine if it is an allowed value type,
+    // and if the value is in range.
+    protected static boolean isAllowedAttributeValue(final Object value) {
+            if (value instanceof Byte ||
+                value instanceof Short ||
+                value instanceof Integer ||
+                value instanceof String ||
+                value instanceof Character ||
+                value instanceof Float ||
+                value instanceof byte[] ||
+                value instanceof short[] ||
+                value instanceof int[] ||
+                value instanceof float[]) {
+            return true;
+        }
+
+        // A special case for Longs: we require Long values to fit into either a uint32_t or an int32_t,
+        // as that is what the BAM spec allows.
+        if (value instanceof Long) {
+            return SAMUtils.isValidUnsignedIntegerAttribute((Long) value)
+                    || ((Long) value >= Integer.MIN_VALUE && (Long) value <= Integer.MAX_VALUE);
+        }
+        return false;
     }
 
     @Override public boolean equals(final Object o) {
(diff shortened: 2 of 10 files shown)


src/java/htsjdk/samtools/BAMIndexer.java
```diff
@@ -55,6 +55,10 @@
      */
     public BAMIndexer(final File output, final SAMFileHeader fileHeader) {
 
+        if (fileHeader.getSortOrder() != SAMFileHeader.SortOrder.coordinate) {
+            throw new SAMException("BAM file be coordinate-sorted for indexing.");
```
lbergelson: Lost a word here,  BAM file **must** be.

Could you also include what the sort order is in the error message?

src/java/htsjdk/samtools/BAMIndexer.java
```diff
@@ -55,6 +55,10 @@
      */
     public BAMIndexer(final File output, final SAMFileHeader fileHeader) {
 
+        if (fileHeader.getSortOrder() != SAMFileHeader.SortOrder.coordinate) {
```
lbergelson: It's possible we should allow "unsorted" to pass with a warning.  There may be bams that don't have the sort order set that we want to index anyway assuming they're coordinate sorted.

src/java/htsjdk/samtools/BAMIndexer.java
```diff
@@ -68,6 +72,10 @@ public BAMIndexer(final File output, final SAMFileHeader fileHeader) {
      */
     public BAMIndexer(final OutputStream output, final SAMFileHeader fileHeader) {
 
+        if (fileHeader.getSortOrder() != SAMFileHeader.SortOrder.coordinate) {
+            throw new SAMException("BAM file be coordinate-sorted for indexing.");
```
lbergelson: Since these constructors are getting some logic in them I think it would be good to refactor them to call out to a private constructor that takes a BinaryBAMIndexWriter.

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2144,20 +2144,15 @@ public Object clone() throws CloneNotSupportedException {
...
      *  - The file source field, which will always always be set to null in the copy
      *
      *  Note that some fields, i.e. the cigar elements, alignment blocks, and
      *  indexing bin, are not explicitly populated in the copy since they are lazily
-     *  generated on demand.
-     *
-     *  Also note that this fails:
+     *  generated on demand, which can cause this to fail:
      *
      *     original.deepCopy().equals(original)
```
lbergelson: Is there a ticket to fix the remainder of the differences causing original.deepCopy().equals(original) to not be true?  It feels like a bug.

src/java/htsjdk/samtools/SAMBinaryTagAndValue.java
```diff
@@ -73,6 +74,34 @@ private boolean typeSafeEquals(final SAMBinaryTagAndValue that) {
...
+        else if (this.value instanceof int[]) {
+            return that.value instanceof int[] ?
+                    Arrays.equals((int[])this.value, (int[])that.value) : false;
+        }
+        else if (this.value instanceof float[]) {
+            return that.value instanceof float[] ?
+                    Arrays.equals((float[])this.value, (float[])that.value) : false;
+        }
+        else {
+            return this.value.equals(that.value);
```
lbergelson: should this include a `this.value instanceOf Object[]` case?   What happens with an `Integer[]`?.  Based on the comment in `cloneValue()` it seems like Integer[] is possible, but `Array.equals` is an identity comparison I believe, not an element by element equality.

lbergelson: Could you include a note explaining why double, char, and long aren't included?

src/java/htsjdk/samtools/SAMBinaryTagAndValue.java
```diff
@@ -73,6 +74,34 @@ private boolean typeSafeEquals(final SAMBinaryTagAndValue that) {
         }
     }
 
+    private boolean valueEquals(SAMBinaryTagAndValue that) {
```
lbergelson: I think this update forces and update to `hashCode` as well.  The current hashCode implementation is based on the array identity, not the array values.

src/java/htsjdk/samtools/BAMIndexer.java
lbergelson: @vdauwera in your experience with weird bams, have you seen many that don't specify their sort order but are actually coordinate sorted?  Should we worry about that case failing to index?  Rewriting the bam header is expensive I believe, so we may want to allow it.

lbergelson:
@cmnbroad back to you.  Looks good, but I think there may be a few issues with the equality changes.

src/java/htsjdk/samtools/SAMBinaryTagAndValue.java
cmnbroad: Those cases are all prohibited (for *SAMRecords* anyway) by the [isAllowedAttributeValue](https://github.com/samtools/htsjdk/blob/master/src/java/htsjdk/samtools/SAMRecord.java#L1397) guard. That should really be in SAMBinaryTagAndValue since its a public class. I wonder if we should move it in there (which could break backward compatibility for some uses), since I don't see a way to keep the copy/equals semantics that we want for SAMRecord attributes if SAMBinaryTagAndValue can take any object as an attribute.

cmnbroad:
I deleted one of Louis's comments by accident, wherein he correctly pointed out that I need to update the hash implementation of SAMBinaryTagAndValue as part of this PR.

src/java/htsjdk/samtools/SAMBinaryTagAndValue.java
```diff
@@ -73,6 +74,34 @@ private boolean typeSafeEquals(final SAMBinaryTagAndValue that) {
...
+                    Arrays.equals((float[])this.value, (float[])that.value) : false;
+        }
+        else {
+            return this.value.equals(that.value);
+        }
+    }
+
     @Override
     public int hashCode() {
         int result = (int) tag;
```
cmnbroad: Per Louis' comment, this needs to be updated to reflect the new equals implementation.

cmnbroad:
This is on hold until I hear back from @vadimzalunin about whether the SAMRecord tests like [test_getUnsignedIntegerAttribute_valid](https://github.com/samtools/htsjdk/blob/master/src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java#L316) and friends that depend on SAMBinaryTagAndUnsignedArrayValue objects with values that [fall outside](https://github.com/samtools/htsjdk/blob/master/src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java#L376) the range enforced by SAMRecord are just for test convenience, or whether CRAM really depends on the such attribute values.

vadimzalunin:
@cmnbroad sorry, I thought I replied already... The test was to ensure that the restriction is applied in the right place. I mean out of range value could blow up in somewhere else (CRAM) but it didn't.

cmnbroad:
@vadimzalunin You had replied (I wrote that comment before you did)!  But I'm still a little unclear. Moving the place where the restriction is applied I exactly what I want to do. I want to move it from SAMRecord right into the SAMBnaryTagAndValue classes. [This comment](https://github.com/samtools/htsjdk/blob/9cb4d73bb3a20f15a66388b104902c1b5111def8/src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java#L353) seem like this test is intended to prevent such a change. So I just want to double check that there is no reason that we're depending on such behavior.

vadimzalunin:
@cmnbroad that was probably over-engineering on my part. The idea was: if an out of range value breaks something outside of the validation method then it's probably a bug. I believe it is safe to delete the test method.

cmnbroad:
Second commit with all comments addressed. This should also fix the last of the known issues with samRec.equals(samRec.deepCopy()) weirdness, and makes hashCode/equals/deepCopy/clone consistent and correct.

src/java/htsjdk/samtools/BAMIndexer.java
```diff
@@ -71,14 +67,28 @@ public BAMIndexer(final File output, final SAMFileHeader fileHeader) {
...
+    }
 
+    /*
+     * Prepare to index a BAM.
+     *
+     * @param fileHeader header for the corresponding bam file.
+     * @param  createWrite a lambda that, given an Integer numReferences value, will create a BinaryBAMIndexWriter
+     *                     with that value and an appropriate output.
+      */
+    private BAMIndexer(final SAMFileHeader fileHeader, Function<Integer, BinaryBAMIndexWriter> createWriter) {
```
lbergelson: I might have made this just take a `BAMIndexWriter` directly and have the callers generate it on their end.  That seems simpler, but maybe I'm missing a reason they can't do that?

cmnbroad: I did try that; you can't create the BAMIndexWriter until you have the number of references, but you can't get that until the common code runs. So using the lambda allows deferral of the BAMIndexWriter creation with the appropriate output type until the end of the execution of  the common code, when we have the refs. I think its either that or duplicate code. Let me know if you see another way.

src/java/htsjdk/samtools/SAMBinaryTagAndValue.java
```diff
@@ -45,17 +50,46 @@
...
+                    SAMTagUtil.getSingleton().makeStringTag(tag));
+        }
         this.tag = tag;
         this.value = value;
     }
 
+    // Inspect the proposed value to determine if it is an allowed value type,
+    // and if the value is in range.
+    protected static boolean isAllowedAttributeValue(final Object value) {
+        if (value instanceof String ||
```
lbergelson: I noticed the order is different here than it was before.  Do we have any idea about the relative distribution of sam attribute types?  I assume the most common should be on top to short circuit them.  I don't know if the original order was carefully designed or just thrown together, but unless we know that this ordering is better I might keep the origianl.

src/java/htsjdk/samtools/BAMIndexer.java
lbergelson: Oh right, I wasn't paying enough attention.  Thanks for explaining.

src/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1394,42 +1394,22 @@ protected void setAttribute(final short tag, final Object value) {
      * @param value the value to be checked
      * @return true if the value is valid and false otherwise
      */
+    @Deprecated
```
lbergelson: could you add an `* @deprecated ` block to the javadoc explaining that it's deprecated because it moved into `SAMBinaryTagAndValue.isAllowedAttributeValue`  It's easy to figure out it out, but  it's even easier to have it tell you that it's safe to use the new method.

src/java/htsjdk/samtools/SAMBinaryTagAndValue.java
cmnbroad: God point. I organized it for my own sanity to make sure I covered everything in each place where I enumerated them, but there is really no need to change the order. I'll change that back.

src/java/htsjdk/samtools/SAMRecord.java
cmnbroad: I actually made it protected in SAMBinaryTagAndValue on the theory that it doesn't need to be public since its enforced by SAMBinaryTagAndValue now. But let me know if you think it needs to be public there too.

src/tests/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -800,8 +769,8 @@ private SAMRecord testNullHeaderDeepCopy() {
         final SAMRecord deepCopy = sam.deepCopy();
 
         // force the indexing bins to be computed in order to satisfy equality test
-        sam.setIndexingBin(sam.computeIndexingBin());
-        deepCopy.setIndexingBin(deepCopy.computeIndexingBin());
+//        sam.setIndexingBin(sam.computeIndexingBin());
+//        deepCopy.setIndexingBin(deepCopy.computeIndexingBin());
```
lbergelson: Should these lines be deleted?

cmnbroad: Yes.

src/java/htsjdk/samtools/SAMRecord.java
lbergelson: I don't think there's any compelling reason to make it public.  I don't see any known subclasses of samtools using this directly.  And I think if they did they would just choose to blow up if the tag isn't valid, so that will happen automatically now.  If they really need to test it and do something else they can always catch the failure.

lbergelson:
@cmnbroad looks good to me.  Merge when you're happy with it.

lbergelson:
could you make sure the final commit message includes information about change in where the guards live,

Issue #544, opened 2016-03-30
Change AD Number in VCF Header


davidbenjamin:
See issue #1286 in broadinstitute/gsa-unstable.  VCF headers have

`##FORMAT=<ID=AD,Number=.,Type=Integer,Description="Allelic depths for the ref and alt alleles in the order listed">`

when they should have

`##FORMAT=<ID=AD,Number=R,Type=Integer,Description="Allelic depths for the ref and alt alleles in the order listed">`.

The former denotes an unlimited values for the AD field, the latter (correctly) denotes that there are as many AD values as there are alleles.

The necessary change is at line 186 of `htsjdk.variant.vcf.VCFStandardHeaderLines.java`, where `VCFHeaderLineCount.UNBOUNDED` should become `VCFHeaderLineCount.R`.

davidbenjamin:
Made the change; the build is failing with `Invalid SRA accession was passed to SRA reader`  @lbergelson or @droazen any clue what that means?

nh13:
@davidbenjamin I am seeing the same thing in my PR.  Logged an issue: https://github.com/samtools/htsjdk/issues/545

Sigh.

davidbenjamin:
Thanks for letting me know @nh13!

magicDGS:
This could be close, couldn't it?

davidbenjamin:
Yes, closed by PR #546.

PR #546, opened 2016-04-01
Fixed AD number if VCF header

davidbenjamin:
See description is issue #544.  Briefly, VCF header used to say AD had unlimited number of values, this fixes it to correctly say it has as many values as there are alleles.


davidbenjamin:
@ldgauthier Can I give this to you in your capacity as GATK bug sensei?

ldgauthier:
Sure -- I'll give it a look today.

Issue #551, opened 2016-04-04
CRAM multi-interval queries return multiple instances of the same record


cmnbroad:
If multiple intervals that do not overlap are specified, a single record will be returned multiple times, once for each interval that it matches.

droazen:
@vadimzalunin This is a pretty serious problem with the existing htsjdk CRAM implementation -- could we get your thoughts on how to fix this?

nh13:
Does that record overlap more than one interval?

droazen:
@nh13 Yes, but with bam multi-interval queries you only get each record that overlaps the array of `QueryIntervals` once (not once per individual interval it overlaps), which is the correct behavior according to the `SamReader` API:

     * Iterate over records that overlap any of the given intervals.  This may be more efficient than querying
     * each interval separately, because multiple reads of the same SAMRecords is avoided.
    public SAMRecordIterator queryOverlapping(final QueryInterval[] intervals);

cmnbroad:
The current implementation creates a new query iterator for each interval, which is why the records get returned once for each interval they overlap. This is also inefficient as the entire stream is re-parsed from the beginning for each interval.

There is a proposed fix in https://github.com/samtools/htsjdk/pull/552, which creates a single CRAMIterator backed by the minimal file spans required to satisfy the intervals, then filters to find the matching records from that. @vadimzalunin let us know your thoughts on this.

vadimzalunin:
@cmnbroad  I like the proposed solution very much given that iterator implementation is shared between BAM and CRAM.

yfarjoun:
hmmm. the description of #552 states that it doesn't fix #551...

cmnbroad:
@yfarjoun PR #552 originally didn't include a fix for this, but it was added and is included, and is listed further down in the comments.

yfarjoun:
OK. it would be good to change the PR comments to make it easier to see.

Issue #550, opened 2016-04-04
CRAM index queries don't handle unmapped/placed records correctly


cmnbroad:
Records that are positioned but unmapped can be returned in both contained and overlap queries even if the records are positioned outside of the requested interval.

Issue #481, opened 2016-02-18
CRAM files do not support zero for start/end coordinates when doing interval queries


bhandsaker:
I believe there is a bug in CRAM support where it is not handling zero correctly for the start/end coordinates for query(), queryContained() and queryOverlapping().  The API specifies that zero should treated as either beginning of sequence (for start coordinate) or end of sequence (for end coordinate). In particular, using query(chr,0,0) should return the whole chromosome, but currently returns no records.

droazen:
@vadimzalunin, could you take this one?

cmnbroad:
@vadimzalunin I'm taking this one since I'm fixing some other bugs with indexing. Let me know if you've started on this already.

cmnbroad:
Closed via #552.

PR #552, opened 2016-04-04
Fixes and tests for CRAM index queries.

cmnbroad:
Fixes https://github.com/samtools/htsjdk/issues/550 and https://github.com/samtools/htsjdk/issues/481. Does not fix https://github.com/samtools/htsjdk/issues/551, which is going to require a different MultiIntervaliterator implementation and will be in a separate PR.

The new tests cases were validated using the corresponding bam/bai; I've left that code commented out in the test file but will remove it once this code has been reviewed.


droazen:
@vadimzalunin Can you review? We were somewhat surprised to learn that cram queries are broken in so many cases..

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -567,47 +571,65 @@ public void close() {
...
+                    refIndex != interval.referenceIndex) {
+                return false;
+            }
+            else {
+                int alignStart = samRecord.getAlignmentStart();
+                int alignEnd = samRecord.getAlignmentEnd();
+                if (alignEnd == 0) {
+                    // handle placed but "unmapped" since they can pass the overlap test
+                    return (alignStart >= interval.start) &&
+                            (interval.end <= 0 || alignStart <= interval.end);
```
droazen: Handle case where `alignStart < interval.start && alignEnd == 0` in a manner consistent with the BAM-querying code.

droazen: (`BAMQueryMultipleIntervalsIteratorFilter.compareIntervalToRecord()` suggests that unmapped reads with a start position are assigned an end position equal to start for BAM queries)

cmnbroad:
I've added a second commit to this PR with fixes for the comments here, as well as a proposed fix for https://github.com/samtools/htsjdk/issues/551. The second commit reduces the two CRAMFileReader interval iterator implementations (IntervalIterator and MultipleIntervalIterator) into a single implementation that uses (a temporarily cloned copy of the) interval matching code used by the BAM iterators iterators to do interval matching. If we get agreement that this is the right strategy, I'll refactor the cloned code so it can be shared between the implementations.

vadimzalunin:
@cmnbroad the approach looks good. A shared iterator impl would be ideal.

cmnbroad:
This now includes fixes for #481, #550, #551, and #562; uses a multiple interval filter iterator implementation that is shared with BAMFileReader; and adds a bunch of new CRAM tests.

cmnbroad:
@vadimzalunin Please review. thx.

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -364,61 +362,14 @@ public SAMRecordIterator assertSorted(final SortOrder sortOrder) {
...
-            } catch (IllegalAccessException e) {
-                throw new SAMException(e);
-            }
-        }
-        throw new SAMException("Failed to query alignment start: " + sequence + " at " + start);
+        return new CRAMIntervalIterator(new QueryInterval[]{new QueryInterval(referenceIndex, start, -1)}, true);
     }
 
     CloseableIterator<SAMRecord> query(final int referenceIndex,
                                        final int start, final int end, final boolean overlap) throws IOException {
```
cmnbroad: Note to self: Delete this method. I meant to remove it as its no longer used internally and should just be deleted (its not part of the public query interface, and is confusing since it reverses the polarity of overlap).

src/tests/java/htsjdk/samtools/CRAMIndexQueryTest.java
```diff
@@ -0,0 +1,464 @@
...
+public class CRAMIndexQueryTest {
+
+    private static final File TEST_DATA_DIR = new File("testdata/htsjdk/samtools/cram");
+
+    private static final String cramQueryWithBAI = "cramQueryWithBAI.cram";
+    private static final String cramQueryWithCRAI = "cramQueryWithCRAI.cram";
+    private static final String cramQueryReference = "human_g1k_v37.20.21.10M-10M200k.fasta";
+
+    //TODO: this is the bam /bai equivalent of the cram test file, used to validate the query results
+    //private static final String bamQueryWithBAI = "samtools.bam";
```
cmnbroad: Per my comment in the PR, I'll remove all the BAM comparison test cases in this file on my last squash after review.

src/tests/java/htsjdk/samtools/CRAMIndexQueryTest.java
```diff
@@ -0,0 +1,464 @@
...
+            {cramQueryWithCRAI, cramQueryReference,
+                    new QueryInterval[]{new QueryInterval(0, 100010, 100010), new QueryInterval(0, 100011, 100011)},
+                    new String[]{"a", "b", "c", "d", "e"}},
+            {cramQueryWithBAI, cramQueryReference,
+                    new QueryInterval[]{new QueryInterval(0, 100010, 100010), new QueryInterval(0, 100011, 100011)},
+                    new String[]{"a", "b", "c", "d", "e"}},
+        };
+    }
+
+    // tests are disabled until https://github.com/samtools/htsjdk/issues/551 is fixed
```
cmnbroad: This comment is now irrelevant and should be removed.

droazen:
@vadimzalunin Do you have time to do a final review on this? This is blocking us from safely using CRAM as input to the GATK, since our queries on CRAMs are currently not returning the correct results. Thanks!

src/java/htsjdk/samtools/BAMQueryMultipleIntervalsIteratorFilter.java
```diff
@@ -0,0 +1,59 @@
+package htsjdk.samtools;
+
+import htsjdk.samtools.util.CoordMath;
+
+/**
+ * Filters out records that do not match any of the given intervals and query type.
+ */
+public class BAMQueryMultipleIntervalsIteratorFilter implements BAMIteratorFilter {
```
droazen: Since you've made this formerly private class public, can you add direct unit tests for it?

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -491,68 +445,47 @@ void enableFileSource(final SamReader reader, final boolean enabled) {
...
-        private SAMRecord next;
-        private boolean noMore = false;
-        private final boolean overlap;
-
-        public IntervalIterator(final CloseableIterator<SAMRecord> delegate, final QueryInterval interval) {
-            this(delegate, interval, true);
-        }
+        // convert queries -> merged BAMFileSpan -> coordinate array
+        private long[] coordinatesFromQueryIntervals(BAMIndex index, QueryInterval[] queries) {
+            getIndex();
```
droazen: Is this necessary to load the index?

src/java/htsjdk/samtools/CRAMFileReader.java
droazen: :+1: on deletion if possible

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -434,7 +385,10 @@ public SAMRecordIterator assertSorted(final SortOrder sortOrder) {
             final Container container = ContainerIO.readContainerHeader(newIterator.getCramHeader().getVersion().major, seekableStream);
             seekableStream.seek(seekableStream.position() + container.containerByteSize);
             iterator = newIterator;
-            iterator.jumpWithinContainerToPos(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX, SAMRecord.NO_ALIGNMENT_START);
+            boolean atAlignments;
+            do {
+                atAlignments = iterator.jumpToAlignment(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX, SAMRecord.NO_ALIGNMENT_START);
```
droazen: Method name here is a little strange, as `jumpToAlignment` implies that it will immediately position the iterator at the requested position, but instead it appears to be iterating repeatedly in order to reach the requested position. Recommend renaming to make this more clear.

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -434,7 +385,10 @@ public SAMRecordIterator assertSorted(final SortOrder sortOrder) {
...
+            } while (!atAlignments && iterator.hasNext());
```
droazen: What happens if the iterator runs out of containers before finding the requested position? Make sure this case is tested (eg., cram with no unmapped reads)

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -491,68 +445,47 @@ void enableFileSource(final SamReader reader, final boolean enabled) {
...
-                    try {
-                        iterator = query(query.referenceIndex, query.start, query.end, overlap);
-                    } catch (final IOException e) {
-                        throw new RuntimeException(e);
-                    }
-                } while (!iterator.hasNext() && queries.hasNext());
+        public CRAMIntervalIterator(final QueryInterval[] queries, final boolean contained) {
+            super(queries, contained);
+            if (queries.length== 0) {
+                throw new IllegalArgumentException("Empty interval list passed to interval iterator");
```
droazen: What does the `SamReader` contract say about the case where one of the `query` methods is called with an empty interval list? Make sure you're being consistent with the guarantees of the interface here.

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -491,68 +445,47 @@ void enableFileSource(final SamReader reader, final boolean enabled) {
...
-        }
-
-        @Override
-        public void remove() {
-            iterator.remove();
+            long[] coordinates = coordinatesFromQueryIntervals(getIndex(), queries);
+            if (coordinates.length != 0) {
+                hasMore = true;
+                try {
+                    unfilteredIterator = new CRAMIterator(
```
droazen: Add a comment explaining why this iterator is "unfiltered" even though we passed in the coordinates from query intervals

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -491,68 +445,47 @@ void enableFileSource(final SamReader reader, final boolean enabled) {
...
-        public SAMRecord next() {
-            return iterator.next();
-        }
-
-        @Override
-        public void remove() {
-            iterator.remove();
+            long[] coordinates = coordinatesFromQueryIntervals(getIndex(), queries);
+            if (coordinates.length != 0) {
+                hasMore = true;
```
droazen: Could it ever be the case that you could get one or more containers back from the query on the `coordinates`, but the containers don't actually contain any records of interest?

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -562,64 +495,46 @@ public SAMRecordIterator assertSorted(final SortOrder sortOrder) {
...
         @Override
         public void close() {
-            delegate.close();
+            if (unfilteredIterator != null) {
+                unfilteredIterator.close();
+            }
         }
 
         @Override
         public boolean hasNext() {
```
droazen: Not a fan of `hasNext()` methods that have side effects like this -- ideally the next record would be preloaded in the constructor and then updated in `next()`. If this is too hard to change, though, we can keep this as-is.

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -562,64 +495,46 @@ public SAMRecordIterator assertSorted(final SortOrder sortOrder) {
...
-            if (!refMatch) return false;
-
-            final int start = record.getAlignmentStart();
-            final int end = record.getAlignmentEnd();
-            if (overlap) {
-                return CoordMath.overlaps(start, end, interval.start, interval.end < 0 ? Integer.MAX_VALUE : interval.end);
-            } else {
-                // contained:
-                return CoordMath.encloses(interval.start, interval.end < 0 ? Integer.MAX_VALUE : interval.end, start, end);
+            while (nextRec == null && hasMore && unfilteredIterator.hasNext()) {
```
droazen: Why do you need both `hasMore` and `nextRec`? Can you just use `nextRec == null` to signal the end of the iteration (this is what most iterators do)?

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -562,64 +495,46 @@ public SAMRecordIterator assertSorted(final SortOrder sortOrder) {
...
-
+            return nextRec != null;
         }
 
         @Override
         public SAMRecord next() {
-            final SAMRecord result = next;
-            next = null;
+            final SAMRecord result = nextRec;
+            nextRec = null;
```
droazen: As discussed above, ideally `next()` should advance, and `nextRec == null` would be the termination condition for the iteration.

src/java/htsjdk/samtools/CRAMIterator.java
```diff
@@ -227,7 +227,7 @@ void nextContainer() throws IOException, IllegalArgumentException,
      * @param refIndex reference sequence index
      * @param pos      alignment start to skip to
      */
-    public boolean jumpWithinContainerToPos(final int refIndex, final int pos) {
+    public boolean jumpToAlignment(final int refIndex, final int pos) {
```
droazen: Should be renamed to reflect the fact that it iterates rather than jumps directly (as mentioned above).

src/tests/java/htsjdk/samtools/CRAMFileWriterWithIndexTest.java
```diff
@@ -124,8 +130,9 @@ public void testUnnecessaryIO() throws IOException {
 
         CRAMFileReader reader = new CRAMFileReader(tabuIS, new ByteArraySeekableStream(indexBytes), source, ValidationStringency.SILENT);
         try {
-            reader.queryAlignmentStart(header.getSequence(refID).getSequenceName(), 1);
+            CloseableIterator<SAMRecord> it = reader.queryAlignmentStart(header.getSequence(refID).getSequenceName(), 1);
             // attempt to read 1st container must fail:
+            it.hasNext();
```
droazen: Might need to change to `it.next()` if you take my advice above.

src/tests/java/htsjdk/samtools/CRAMIndexQueryTest.java
```diff
@@ -0,0 +1,464 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package htsjdk.samtools;
```
droazen: Should this be in a cram package?

src/tests/java/htsjdk/samtools/CRAMIndexQueryTest.java
```diff
@@ -0,0 +1,464 @@
...
+ * Test index query calls against a *SamReader* on a CRAM File, which will use
+ * whatever index format (.bai or .crai converted to .bai) is available for the
+ * target file.
+ */
+public class CRAMIndexQueryTest {
+
+    private static final File TEST_DATA_DIR = new File("testdata/htsjdk/samtools/cram");
+
+    private static final String cramQueryWithBAI = "cramQueryWithBAI.cram";
+    private static final String cramQueryWithCRAI = "cramQueryWithCRAI.cram";
```
droazen: Do these have more than 1 container's worth of reads? Do we have test coverage for the multi-container case?

src/tests/java/htsjdk/samtools/CRAMIndexQueryTest.java
```diff
@@ -0,0 +1,464 @@
...
+
+            // interval with 0 end
+            //{bamQueryWithBAI, null, new QueryInterval(0, 100015, 0), new String[]{"a", "b", "c", "d", "e", "f"}},
+            {cramQueryWithCRAI, cramQueryReference, new QueryInterval(0, 100015, 0), new String[]{"a", "b", "c", "d", "e", "f"}},
+            {cramQueryWithBAI, cramQueryReference, new QueryInterval(0, 100015, 0), new String[]{"a", "b", "c", "d", "e", "f"}},
+
+            // interval with 0 start and 0 end
+            //{bamQueryWithBAI, null, new QueryInterval(0, 0, 0), new String[]{"a", "b", "c", "d", "e", "f",  "f"}},
+            {cramQueryWithCRAI, cramQueryReference, new QueryInterval(0, 0, 0), new String[]{"a", "b", "c", "d", "e", "f",  "f"}},
+            {cramQueryWithBAI, cramQueryReference, new QueryInterval(0, 0, 0), new String[]{"a", "b", "c", "d", "e", "f",  "f"}},
```
droazen: Two reads with name "f" are returned here? Is this intended?

src/tests/java/htsjdk/samtools/CRAMIndexQueryTest.java
```diff
@@ -0,0 +1,464 @@
...
+        return new Object[][]{
+            //{bamQueryWithBAI, null,
+            //        new QueryInterval[]{new QueryInterval(0, 100010, 100010), new QueryInterval(0, 100011, 100011)},
+            //        new String[]{"a", "b", "c", "d", "e"}},
+            {cramQueryWithCRAI, cramQueryReference,
+                    new QueryInterval[]{new QueryInterval(0, 100010, 100010), new QueryInterval(0, 100011, 100011)},
+                    new String[]{"a", "b", "c", "d", "e"}},
+            {cramQueryWithBAI, cramQueryReference,
+                    new QueryInterval[]{new QueryInterval(0, 100010, 100010), new QueryInterval(0, 100011, 100011)},
+                    new String[]{"a", "b", "c", "d", "e"}},
```
droazen: Recommend having more than two test cases here

droazen: Can also try porting the test cases from `ReadsDataSourceUnitTest` here, which use the `reads_data_source_test1.bam` file mentioned above.

droazen:
First-pass review complete, back to @cmnbroad

src/tests/java/htsjdk/samtools/CRAMIndexQueryTest.java
cmnbroad: These files have a small number of reads and so are single container. There are additional query tests in CRAMFileBAIIndexTest that convert a BAM to a CRAM and force it to have multiple containers by using a small DEFAULT_RECORDS_PER_SLICE value, and then do some queries against it. There are also some unmapped tests below that use multiple containers to verify a fix to a bug that only occurred with multiple containers.

src/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: Added a test with no unmapped reads to the unmapped tests.

src/java/htsjdk/samtools/BAMQueryMultipleIntervalsIteratorFilter.java
cmnbroad: Done.

src/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: Done.

src/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: Done.

src/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: I don't see this stated explicitly anywhere, but the BAMReader silently returns an empty iterator, so I matched this behavior and added a test.

src/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: Done.

src/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: I don't actually know if thats possible (I would not expect it to happen - if it did I think it would be a bug in the BAMFileSpan code). However the iterator filter should filter those records out in either case for.

src/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: Fixed.

src/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: Fixed.

src/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: Fixed.

src/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: Fixed.

src/java/htsjdk/samtools/CRAMIterator.java
cmnbroad: Fixed.

src/tests/java/htsjdk/samtools/CRAMIndexQueryTest.java
cmnbroad: Maybe ideally, but this is where all of the existing CRAM (and BAM) index tests are. If you think we should move them we should probably move them all.

src/tests/java/htsjdk/samtools/CRAMIndexQueryTest.java
cmnbroad: These have all been removed.

src/tests/java/htsjdk/samtools/CRAMIndexQueryTest.java
cmnbroad: Added quite a few more tests for both overlapped and contained.

src/tests/java/htsjdk/samtools/CRAMIndexQueryTest.java
cmnbroad: Done.

src/tests/java/htsjdk/samtools/CRAMFileWriterWithIndexTest.java
cmnbroad: We don't actually need this line at all anymore since it the advance now happens when the iterator is initialized.

cmnbroad:
Didn't squash or rebase yet but otherwise I think this covers it.

droazen:
@cmnbroad could you rebase to resolve the conflicts?

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -491,68 +440,44 @@ void enableFileSource(final SamReader reader, final boolean enabled) {
...
+                    unfilteredIterator = new CRAMIterator(
+                            getSeekableStreamOrFailWithRTE(),
+                            referenceSource,
+                            coordinates,
+                            validationStringency
+                    );
+                } catch (final IOException e) {
+                    throw new RuntimeEOFException(e);
+                }
+                next(); // advance to the first record that matches the filter criteria
```
droazen: Have you confirmed that this iterator works with an empty cram file?

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -562,64 +487,40 @@ public SAMRecordIterator assertSorted(final SortOrder sortOrder) {
...
-            if (record.getReferenceIndex() > interval.referenceIndex) return true;
-            if (record.getReferenceIndex() != interval.referenceIndex) return false;
-
-            return interval.end != -1 && record.getAlignmentStart() > interval.end;
-
+            return nextRec != null;
         }
 
         @Override
         public SAMRecord next() {
```
droazen: Throw `NoSuchElementException` if `next()` is called when `hasNext()` is false

src/tests/java/htsjdk/samtools/CRAMIndexQueryTest.java
```diff
@@ -0,0 +1,554 @@
...
+ */
+public class CRAMIndexQueryTest {
+
+    private static final File TEST_DATA_DIR = new File("testdata/htsjdk/samtools/cram");
+
+    private static final String cramQueryWithBAI = "cramQueryWithBAI.cram";
+    private static final String cramQueryWithCRAI = "cramQueryWithCRAI.cram";
+    private static final String cramQueryReference = "human_g1k_v37.20.21.10M-10M200k.fasta";
+
+    private static final String cramQueryReadsWithBAI = "reads_data_source_test1.cram";
```
droazen: Should probably rename `reads_data_source_test1.cram`, since there is no `ReadsDataSource` class here. Perhaps `cram_query_test.cram`

src/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -562,64 +487,40 @@ public SAMRecordIterator assertSorted(final SortOrder sortOrder) {
...
+                        throw new SAMException("Unexpected return from compareToFilter");
+                }
+            }
             return result;
         }
 
         @Override
         public void remove() {
-            throw new RuntimeException("Not available.");
+            throw new RuntimeException("Method \"remove\" not implemented for CRAMFileReader.");
```
droazen: `for CRAMIntervalIterator`

droazen:
Review complete, back to @cmnbroad. Merge after addressing comments and rebasing.

src/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: Added a couple of tests to validate this.

PR #553, opened 2016-04-04
packaging the intel deflator into the published jar

lbergelson:
including the intel deflator in the published jar, it appears in the jar as `lib/jni/libIntelDeflator.so`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -26,6 +26,10 @@ libraryDependencies += "org.testng" % "testng" % "6.8.8"
 
 unmanagedBase := baseDirectory.value
 
+mappings in (Compile, packageBin) ++= Seq(
+  (baseDirectory.value / "lib/jni/libIntelDeflater.so") -> "lib/jni/libIntelDeflator.so"
+)
+
 javaSource in Compile := baseDirectory.value / "src/java"
 
 javaSource in Test := baseDirectory.value / "src/tests"


build.sbt
```diff
@@ -26,6 +26,10 @@ libraryDependencies += "org.testng" % "testng" % "6.8.8"
 
 unmanagedBase := baseDirectory.value
 
+mappings in (Compile, packageBin) ++= Seq(
+  (baseDirectory.value / "lib/jni/libIntelDeflater.so") -> "lib/jni/libIntelDeflator.so"
+)
+
```
droazen: What if we're building with `ant`?

lbergelson:
@droazen We can put the lib anywhere in the jar, not sure where we should put it...

droazen:
@lbergelson With this change, will the intel deflator be used by htsjdk classes by default when using the published jar?

lbergelson:
No.  This doesn't include code to fix the loading for dependencies since testing that is much more involved.

build.sbt
lbergelson: Everything is unchanged if we're building with ant.  It will continue to do whatever it already does.

droazen:
@lbergelson Have you built the sbt jar and confirmed that the lib is in the proper place inside it? If you have, merge when ready

lbergelson:
@droazen I have built the jar and it gets packaged in lib/jni. The question is, should we move that somewhere else?  Is that a good place for it?

droazen:
@lbergelson Seems fine to me!

PR #554, opened 2016-04-05
Switched to relative lib path for deflator so Picard public release s…

jacarey:


lbergelson:
@jacarey Was this necessary because of the sbt changes I made?  I didn't think that would affect picard at all since it shouldn't change the layout of the checkout out htsjdk repo..

jacarey:
@lbergelson The issue is that the picard release script calls htsjdk tests for its own directory which gets set as user.dir and so the intel tests attempt to find the deflator in picard/lib/jni instead of picard/htsjdk/lib/jni

lbergelson:
@jacarey Ah, so sounds unrelated to what I was doing.  I didn't break everything! Woo.

jacarey:
@lbergelson Nope in fact it only broke the release script and the change was simple enough to fix.

PR #557, opened 2016-04-06
100% test coverage for FastqRecord

akiezun:
### Description

Adding tests for `FastqRecord`. It now has 100% line coverage

### Checklist

- [X] Code compiles correctly
- [X] All tests passing

### DIFF
--- a/src/test/java/htsjdk/samtools/fastq/FastqRecordTest.java
+++ b/src/test/java/htsjdk/samtools/fastq/FastqRecordTest.java
@@ -1,0 +1,204 @@
+package htsjdk.samtools.fastq;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public final class FastqRecordTest {
+
+    @Test
+    public void testBasic() {
+        final String seqHeaderPrefix = "FAKE0003 Original version has Solexa scores from 62 to -5 inclusive (in that order)";
+        final String seqLine = "ACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGT";
+        final String qualHeaderPrefix = "";
+        final String qualLine = ";<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~";
+        final FastqRecord fastqRecord = new FastqRecord(seqHeaderPrefix, seqLine, qualHeaderPrefix, qualLine);
+
+        Assert.assertNull(fastqRecord.getBaseQualityHeader());
+
+        Assert.assertEquals(fastqRecord.getReadHeader(), seqHeaderPrefix);
+        Assert.assertEquals(fastqRecord.getBaseQualityString(), qualLine);
+        Assert.assertEquals(fastqRecord.getReadString(), seqLine);
+        Assert.assertNotNull(fastqRecord.toString());//just check not nullness
+        Assert.assertNotEquals(fastqRecord, null);
+        Assert.assertFalse(fastqRecord.equals(null));
+        Assert.assertNotEquals(null, fastqRecord);
+        Assert.assertEquals(fastqRecord, fastqRecord);
+        Assert.assertNotEquals(fastqRecord, "fred");
+        Assert.assertNotEquals("fred", fastqRecord);
+        Assert.assertEquals(fastqRecord.length(), seqLine.length());
+        Assert.assertEquals(fastqRecord.getBaseQualityString().length(), fastqRecord.getReadString().length());
+        Assert.assertEquals(fastqRecord.getReadString().length(), fastqRecord.length());
+    }
+
+    @Test
+    public void testBasicEmptyHeaderPrefix() {
+        final String seqHeaderPrefix = "";
+        final String seqLine = "ACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGT";
+        final String qualHeaderPrefix = "";
+        final String qualLine = ";<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~";
+        final FastqRecord fastqRecord = new FastqRecord(seqHeaderPrefix, seqLine, qualHeaderPrefix, qualLine);
+        Assert.assertNull(fastqRecord.getReadHeader());
+        Assert.assertNull(fastqRecord.getBaseQualityHeader());
+    }
+
+    @Test
+    public void testCopy() {
+        final String seqHeaderPrefix = "FAKE0003 Original version has Solexa scores from 62 to -5 inclusive (in that order)";
+        final String seqLine = "ACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGTACGT";
+        final String qualHeaderPrefix = "";
+        final String qualLine = ";<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~";
+        final FastqRecord fastqRecord = new FastqRecord(seqHeaderPrefix, seqLine, qualHeaderPrefix, qualLine);
+        final FastqRecord fastqRecordCopy = new FastqRecord(fastqRecord);
+
+        Assert.assertEquals(fastqRecord, fastqRecordCopy);
+        Assert.assertNotSame(fastqRecord, fastqRecordCopy);
+        Assert.assertSame(fastqRecord.getReadString(), fastqRecordCopy.getReadString());
+        Assert.assertSame(fastqRecord.getBaseQualityString(), fastqRecordCopy.getBaseQualityString());
+        Assert.assertSame(fastqRecord.getBaseQualityHeader(), fastqRecordCopy.getBaseQualityHeader());
+    }
+
+    @Test
+    public void testNullSeq() {
+        final String seqHeaderPrefix = "header";
+        final String seqLine = null;
+        final String qualHeaderPrefix = "";
+        final String qualLine = ";<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~";
+        new FastqRecord(seqHeaderPrefix, seqLine, qualHeaderPrefix, qualLine);
+        //Note: this does not blow up now but it will once we enforce non null seqLine
+    }
+
+    @Test
+    public void testEqualsWithNullHeader() {
+        final String seqLine = "GATTACA";
+        final String qualHeaderPrefix = "";
+        final String qualLine = "ABCDEFG";
+        final FastqRecord fastqRecord1 = new FastqRecord("", seqLine, qualHeaderPrefix, qualLine);
+        final FastqRecord fastqRecord2 = new FastqRecord("header", seqLine, qualHeaderPrefix, qualLine);
+        Assert.assertNotEquals(fastqRecord1, fastqRecord2);
+        Assert.assertNotEquals(fastqRecord2, fastqRecord1);
+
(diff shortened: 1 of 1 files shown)


src/tests/java/htsjdk/samtools/fastq/FastqRecordTest.java
```diff
@@ -0,0 +1,136 @@
+package htsjdk.samtools.fastq;
```
nh13: license

src/tests/java/htsjdk/samtools/fastq/FastqRecordTest.java
```diff
@@ -0,0 +1,136 @@
...
+        Assert.assertSame(fastqRecord.getBaseQualityString(), fastqRecordCopy.getBaseQualityString());
+        Assert.assertSame(fastqRecord.getBaseQualityHeader(), fastqRecordCopy.getBaseQualityHeader());
+    }
+
+    @Test
+    public void testEqualsWithNullSeq() {
+        final String seqHeaderPrefix = "header";
+        final String seqLine = "GATTACA";
+        final String qualHeaderPrefix = "";
+        final String qualLine = ";<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~";
```
nh13: I don't think it is valid for the quality line to have a different number of characters as the sequence line.

src/tests/java/htsjdk/samtools/fastq/FastqRecordTest.java
```diff
@@ -0,0 +1,136 @@
...
+        Assert.assertNotSame(fastqRecord, fastqRecordCopy);
+        Assert.assertSame(fastqRecord.getReadString(), fastqRecordCopy.getReadString());
+        Assert.assertSame(fastqRecord.getBaseQualityString(), fastqRecordCopy.getBaseQualityString());
+        Assert.assertSame(fastqRecord.getBaseQualityHeader(), fastqRecordCopy.getBaseQualityHeader());
+    }
+
+    @Test
+    public void testCopyWithNullSeq() {
+        final String seqHeaderPrefix = "header";
+        final String seqLine = null;
```
nh13: I think if `seqLine` is `null`, then `qualLine` should be required to be `null`.

src/tests/java/htsjdk/samtools/fastq/FastqRecordTest.java
```diff
@@ -0,0 +1,136 @@
...
+        Assert.assertEquals(fastqRecord.getReadHeader(), seqHeaderPrefix);
+        Assert.assertEquals(fastqRecord.getBaseQualityString(), qualLine);
+        Assert.assertEquals(fastqRecord.getReadString(), seqLine);
+        Assert.assertNotNull(fastqRecord.toString());//just check not nullness
+        Assert.assertNotEquals(fastqRecord, null);
+        Assert.assertNotEquals(null, fastqRecord);
+        Assert.assertEquals(fastqRecord, fastqRecord);
+        Assert.assertNotEquals(fastqRecord, "fred");
+        Assert.assertNotEquals("fred", fastqRecord);
+        Assert.assertEquals(fastqRecord.length(), seqLine.length());
```
nh13: assert quality line and sequence line are the same length?

src/tests/java/htsjdk/samtools/fastq/FastqRecordTest.java
```diff
@@ -0,0 +1,136 @@
...
+        Assert.assertNotEquals(fastqRecord2.hashCode(), fastqRecord1.hashCode());
+        Assert.assertEquals(fastqRecord1.hashCode(), fastqRecord1.hashCode());
+        Assert.assertEquals(fastqRecord2.hashCode(), fastqRecord2.hashCode());
+    }
+
+    @Test
+    public void testEqualsWithNullHeader() {
+        final String seqLine = "GATTACA";
+        final String qualHeaderPrefix = "";
+        final String qualLine = ";<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~";
```
nh13: same comment about lengths

src/tests/java/htsjdk/samtools/fastq/FastqRecordTest.java
```diff
@@ -0,0 +1,136 @@
...
+        Assert.assertNotEquals(fastqRecord1.hashCode(), fastqRecord2.hashCode());
+        Assert.assertNotEquals(fastqRecord2.hashCode(), fastqRecord1.hashCode());
+        Assert.assertEquals(fastqRecord1.hashCode(), fastqRecord1.hashCode());
+        Assert.assertEquals(fastqRecord2.hashCode(), fastqRecord2.hashCode());
+    }
+    @Test
+    public void testEqualsWithNullBaseQualityHeader() {
+        final String seqHeaderPrefix = "header";
+        final String seqLine = "GATTACA";
+        final String qualLine = ";<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~";
```
nh13: same comment about lengths and `null`

src/tests/java/htsjdk/samtools/fastq/FastqRecordTest.java
```diff
@@ -0,0 +1,136 @@
...
+        Assert.assertNotEquals(fastqRecord2.hashCode(), fastqRecord1.hashCode());
+        Assert.assertEquals(fastqRecord1.hashCode(), fastqRecord1.hashCode());
+        Assert.assertEquals(fastqRecord2.hashCode(), fastqRecord2.hashCode());
+    }
+
+    @Test
+    public void testEqualsWithNullQualLine() {
+        final String seqHeaderPrefix = "header";
+        final String seqLine = "GATTACA";
+        final String qualLine = ";<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~";
```
nh13: same comment about lengths

src/tests/java/htsjdk/samtools/fastq/FastqRecordTest.java
akiezun: the ctor does not blow up. Should it? if yes, that would change the current semantics. The same question for other comments: should we change or keep the sematics of `FastqRecord`?

nh13: Various sites (ex. wikipedia, MAQ's page, and even here [1]) all say that the seq and qual fields should be the same length, so I would prefer a change in semantics.  `FastqReader` also enforces this prior to creating a `FastqRecord`.

[1] http://www.ncbi.nlm.nih.gov/pmc/articles/PMC2847217/

src/tests/java/htsjdk/samtools/fastq/FastqRecordTest.java
yfarjoun: Really? Hardly any of the test files have licenses...is this a thing we want to start now? (I'm not sure why they don't have a license, but I've noticed it...)

nh13: Crap, totally my fault for not remembering that this was a test file, just a habit.

akiezun: ok, no license then

src/tests/java/htsjdk/samtools/fastq/FastqRecordTest.java
akiezun: the spec also indicated that both seq and quals are non-null and not empty. Is that a correct understanding?

src/tests/java/htsjdk/samtools/fastq/FastqRecordTest.java
akiezun: i think neither can be null or empty according to the spec

src/tests/java/htsjdk/samtools/fastq/FastqRecordTest.java
akiezun: done

src/tests/java/htsjdk/samtools/fastq/FastqRecordTest.java
nh13: I think so, having empty sequences makes no sense, and `nulls` seem wrong too.  I think we get to decide here.

src/tests/java/htsjdk/samtools/fastq/FastqRecordTest.java
nh13: If we are going to follow the spec, then yes.

src/java/htsjdk/samtools/fastq/FastqRecord.java
```diff
@@ -24,18 +24,23 @@
...
+import java.util.Objects;
 
 /**
  * Represents a fastq record, fairly literally, i.e. without any conversion.
  */
 public class FastqRecord implements Serializable {
     private static final long serialVersionUID = 1L;
     private final String seqHeaderPrefix;
-    private final String seqLine;
+    private final String seqLine;            //not null, not empty
```
akiezun: according to the FASTQ spec that @nh13 pointed out, these are not null and not empty. So I enforce it here and use that fact to simplify a bunch of code in this class. Tests get updated accordingly

akiezun:
@nh13 can you re-review?

src/java/htsjdk/samtools/fastq/FastqRecord.java
tfenne: Just commenting to note that, while I think this is good and the right thing to do, this is a breaking change that will effect many places in the ecosystem, and so should be clearly noted in release notes and a corresponding bump in version number made.

akiezun:
1 sec

src/java/htsjdk/samtools/fastq/FastqRecord.java
akiezun: agreed, this is a breaking API change. Where do we make notes of such things?

tfenne: I'm not sure...  Perhaps we should have an open ticket per release, or use the release prep features to keep a running list of things like this for the next release?  @jacarey what do you think?

akiezun:
weird, there seems to be some transient travis problem - some FTP tests failing (?) what is that?

lindenb:
```
private final String seqLine;            //not null, not empty
```

but some cleaning tools like cutadapt could remove the whole sequence. What happends when reading such fastq files with FastqReader ?

lbergelson:
@akiezun There are a bunch of tests that hit a live ftp server.  They fail sometimes.  It's annoying.  I've created #560 to replace them with a mocked ftp server at some point.

nh13:
@lindenb my personal preference would be for the authors of `cutadapt` to update the tools to output valid FASTQs instead of zero length reads.  Myself and others have certainly run into issues with the output of `cutadapt` when fed into aligners, especially RNA-seq aligners, that can't handle zero-length reads (and they should have to).

@akiezun I am assigning this back to you.  Looking more closely, `FastqReader` enforces this spec, but `FastqWriter` doesn't.  I think in Picard, checking that `SamToFastq` and `IlluminaBasecallsToFastq` do the right thing, in particular when trimming (adapter, quality, or other) may cause a zero-length read.

akiezun:
@nh13  can you clarify what you want to see? Should I change this code or client code or something else? I'm not sure.

akiezun:
I guess @nh13 may be too busy. Anyone else have an opinion? I'm fine with just dropping this PR it's too much hassle

nh13:
I would like any FASTQ code to follow the spec rather than allow things like empty reads. Others should weigh in too as the code has been non-conforming for a while so an argument for not changing or for pragmatism could be made.

yfarjoun:
@akiezun could you split this into two PR's:

1. increase test coverage
2. change behavior regarding empty reads

??

akiezun:
@yfarjoun removed breaking changes from this PR. can you review?

akiezun:
@yfarjoun good to merge?

yfarjoun:
👍

Issue #564, opened 2016-04-10
race condition in AbstractAsyncWriter causing "java.lang.RuntimeException: Queue should be empty but is size:"


akiezun:
```
java.lang.RuntimeException: Queue should be empty but is size: 276
        at htsjdk.samtools.util.AbstractAsyncWriter.close(AbstractAsyncWriter.java:75)
        at org.broadinstitute.hellbender.utils.read.ArtificialBAMBuilder.makeBAMFile(ArtificialBAMBuilder.java:143)
        at org.broadinstitute.hellbender.utils.read.ArtificialBAMBuilder.makeTemporaryBAMFile(ArtificialBAMBuilder.java:130)
        at org.broadinstitute.hellbender.utils.read.ArtificialBAMBuilderUnitTest.testBamProvider(ArtificialBAMBuilderUnitTest.java:69)
```

- [x] Make sure your issue is not already in the [htsjdk issue tracker](https://github.com/samtools/htsjdk/issues?q=)

### Subject of the issue
I think there's a race condition in AbstractAsyncWriter that is causing https://github.com/broadinstitute/gatk/issues/1699

The race is exposed by a context switch in the middle of evaluating the `||` condition of the writer thread

The scenario is like this. 
1) let's say at a specific moment the queue is empty (but the writer is not closed)

2) writer thread starts executing this the loop condition

` while (!queue.isEmpty() || !isClosed.get()) {`

it evaluates the **first** condition and queue is empty so it moves on to evaluating the second condition.
But before it gets to it, ... <SWITCH>

3) main thread calls `write` a few times - the queue becomes non empty

4) main thread calls `close()`

main runs this line and thus flips isClosed to true `!this.isClosed.getAndSet(true)`

then main checks that `this.queue.isEmpty()` which it is not so it moves on 
then main blocks on writer.join() and waits

5) writer thread wakes up and evaluates the second condition of the while loop: `isClosed.get()` and that returns true now so the loop is not entered - because the first part of the condition is still remembered as `false`)

Writer thread ends the `run()` method and dies

6) main thread wakes up, and blows up on
```
if (!this.queue.isEmpty()) { throw new RuntimeException("Queue should be empty but is size: " + this.queue.size()); }
``

**Comment by akiezun** [comment]
i think reversing the order of operations in the while loop will fix it, ie
```
while (!isClosed.get() || !queue.isEmpty()) 
```
[...]

PR #565, opened 2016-04-11
fix for race condition in AbstractAsyncWriter

akiezun:
### Description

Fixes https://github.com/samtools/htsjdk/issues/564 by reverting the order of conditions in the while loop and making checking them effectively atomic (ie nothing can change the values of once the `while` condition starts being evaluated

### DIFF
--- a/src/java/htsjdk/samtools/util/AbstractAsyncWriter.java
+++ b/src/java/htsjdk/samtools/util/AbstractAsyncWriter.java
@@ -11,6 +11,9 @@ import java.util.concurrent.atomic.AtomicReference;
  * Abstract class that is designed to be extended and specialized to provide an asynchronous
  * wrapper around any kind of Writer class that takes an object and writes it out somehow.
  *
+ * NOTE: Objects of subclasses of this class are not intended to be shared between threads.
+ * In particular there must be only one thread that calls {@link #write} and {@link #close}.
+ *
  * @author Tim Fennell
  */
 public abstract class AbstractAsyncWriter<T> implements Closeable {
@@ -56,7 +59,7 @@ public abstract class AbstractAsyncWriter<T> implements Closeable {
     }
 
     /**
-     * Attempts to finishing draining the queue and then calls synchronoslyClose() to allow implementation
+     * Attempts to finish draining the queue and then calls synchronouslyClose() to allow implementation
      * to do any one time clean up.
      */
     public void close() {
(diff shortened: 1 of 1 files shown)


droazen:
Can you run the test suite a sufficient number of times to convince yourself that the race is gone, and post the results here?

akiezun:
@droazen @d-cameron @tfenne can you have a look at this and confirm (or debunk) that a) there was a race condition, b) this fixes it and c) suggest how to test it?

droazen:
@akiezun I've added my comments to https://github.com/samtools/htsjdk/issues/564

akiezun:
@droazen the htsjdk test suite is insufficient here and it never did fail with this error

droazen:
@akiezun I meant run the GATK4 test suite repeatedly (with this patch)

droazen:
It would also be good if we could craft a direct htsjdk test that fails (even intermittently) with the same error.

akiezun:
can you make me a snaphot build of htsjdk?

droazen:
@akiezun You can temporarily modify GATK4's `build.gradle` in your branch to point to a custom local htsjdk build.

tfenne:
@akiezun I followed the discussion in the issue, and I think this will do the trick.  The other thing you could do to further protect things would be to remove the assertion in the `close()` method and replace it with code to drain any remaining items, once the writer thread is dead, from the queue and write them using `writeSynchronously()`.

akiezun:
@tfenne thanks. I added queue draining as a second commit. can you look at it?

tfenne:
Looks good to me @akiezun .  :+1:

akiezun:
thanks @tfenne. Unless someone objects, I will squash and merge by EOB tomorrow (~24 hours). I have not been able to write a test that reproduces this (either in htjdk or gatk). But I have successfully run GATK4 test suite with this update.

d-cameron:
There's still a data loss race condition but it's between write() and close(). A thread already in write() can add elements to the queue once close() has set isClosed and this additional element may or may not get written by the safety flush that was added.

This could be fixed by putting write() and close() in synchronized blocks, or just left as-is and documentation added to say that the class does not expose a thread-safe API and if you call write() and close() at the same time, the write may or may not get written.

tfenne:
You're right @d-cameron, and I would go with the second suggestion.  The goal of this class and it's subclasses has never been to create writers that can be invoked from multiple threads - just like the non async `FastqWriter` and `SAMFileWriter`, the expectation is that the writer is owned and used by a single thread.  No harm in documenting that more clearly.

akiezun:
Agreed, I'll add documentation that this class (write and close methods) is
to be called from a single thread.

On Tuesday, April 12, 2016, Tim Fennell <notifications@github.com> wrote:

> You're right @d-cameron <https://github.com/d-cameron>, and I would go
> with the second suggestion. The goal of this class and it's subclasses has
> never been to create writers that can be invoked from multiple threads -
> just like the non async FastqWriter and SAMFileWriter, the expectation is
> that the writer is owned and used by a single thread. No harm in
> documenting that more clearly.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/samtools/htsjdk/pull/565#issuecomment-208814268>
>


-- 
Sent from Gmail Mobile

PR #569, opened 2016-04-13
fixing a spelling mistake in the packaged intel deflater path

lbergelson:
Fixing a typo the build.sbt that causes the libIntelDeflater.so to be named incorrectly.


akiezun:
👍

PR #571, opened 2016-04-15
Query methods for FeatureReader

magicDGS:
### Description

Added `query(Locatable loc)` and `query(List<Locatable> locs)` methods for `FeatureReader` and `VCFFileReader`. In addition some javadoc improvements and refactoring.

### Checklist

- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/java/htsjdk/tribble/AbstractFeatureReader.java
+++ b/src/java/htsjdk/tribble/AbstractFeatureReader.java
@@ -18,6 +18,9 @@
 
 package htsjdk.tribble;
 
+import htsjdk.samtools.util.IntervalList;
+import htsjdk.samtools.util.IntervalUtil;
+import htsjdk.samtools.util.Locatable;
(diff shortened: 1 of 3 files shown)


droazen:
@magicDGS This is a major breaking change to a core interface relied on by multiple different downstream projects (not just the GATK -- IGV and other projects also depend on this tribble interface). Since the change appears motivated by convenience rather than enabling clients to do things that weren't previously possible, I'm inclined to reject this for now, at least until we have a better procedure for making breaking interface changes in htsjdk (https://github.com/samtools/htsjdk/issues/378).

droazen:
I'll add, however: if you could re-implement this in a way that would not force any client code to change (eg., using Java 8 default methods), we might be able to accept this.

magicDGS:
I think that I can wait till #378, as I understand the breaking is not good.

PR #577, opened 2016-04-18
AsynchronousLineReader should terminate worker thread

akiezun:
### Description

AsynchronousLineReader did not terminate the worker thread properly which resulted in errors like https://github.com/broadinstitute/gatk/issues/1638 (where the worker kept trying to read from a stream after it was supposed to be terminated).

Note: the AsynchronousLineReader is actually still very slow, slower than syncIO (which is a shame). This PR does not address this - only the thread and resource leakage.

Also added:
- closing the reader in `TribbleIndexedFeatureReader.readHeader` - it was left open which is always a problem but particularly when you have multiple threads
- closing resources in AsynchronousLineReaderTest.java and PrintVariantsExample.java

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/java/htsjdk/tribble/TribbleIndexedFeatureReader.java
+++ b/src/java/htsjdk/tribble/TribbleIndexedFeatureReader.java
@@ -213,6 +213,7 @@ public class TribbleIndexedFeatureReader<T extends Feature, SOURCE> extends Abst
     private void readHeader() throws IOException {
         InputStream is = null;
         PositionalBufferedStream pbs = null;
+        SOURCE source = null;
         try {
             is = ParsingUtils.openInputStream(path);
             if (path.endsWith("gz")) {
@@ -220,11 +221,12 @@ public class TribbleIndexedFeatureReader<T extends Feature, SOURCE> extends Abst
                 is = new GZIPInputStream(new BufferedInputStream(is));
             }
             pbs = new PositionalBufferedStream(is);
-            final SOURCE source = codec.makeSourceFromStream(pbs);
+            source = codec.makeSourceFromStream(pbs);
             header = codec.readHeader(source);
         } catch (Exception e) {
             throw new TribbleException.MalformedFeatureFile("Unable to parse header with error: " + e.getMessage(), path, e);
         } finally {
+            if (source != null) codec.close(source);
             if (pbs != null) pbs.close();
             else if (is != null) is.close();
         }
--- a/src/java/htsjdk/tribble/readers/AsynchronousLineReader.java
+++ b/src/java/htsjdk/tribble/readers/AsynchronousLineReader.java
@@ -1,12 +1,15 @@
 package htsjdk.tribble.readers;
 
 import htsjdk.samtools.util.CloserUtil;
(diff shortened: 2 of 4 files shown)


akiezun:
@d-cameron can you help review? thanks

d-cameron:
Looking at just the stack traces from linked issue, it appears that the problem is not a race condition, but the fact that the worker thread is autoclosing the underlying stream at EOF. Both the @lbergelson error message and the @akiezun one are caused by a closed stream.

Not having seen the test cases themselves, my guess is the error is reproduceable with a test case that calls TribbleIndexedFeatureReader.query() to within 100 lines of the end of a file, waits (so the async thread hits eof then closes the stream), then calling query() again to throw the closed stream error that they are encountering.

d-cameron:
More concerning to me is the apparent API design flaw that allows foreground threads to change the stream position while the background thread is still reading.

TribbleIndexedFeatureReader.query() leaves it up the caller to dispose of the iterator and doesn't hold a reference to ensure that close() is indeed called when a new query() is made. This seems rather dangerous to me. A good example is FeatureReaderTest.testBedQuery() which leaks a thread per testQuery() call in async mode.

Given the comments about it giving no performance advantage and the significant refactoring required to fix these flaws, I'd be inclined to remove the class entirely.

akiezun:
Funny you should say that, removing it was my original thought too, see https://github.com/samtools/htsjdk/pull/528

Who's in favor of killing this flawed class: @droazen @tfenne wdyt?

akiezun:
@d-cameron do you mean something like this?  This test blows up very reliably:
```
    @Test
    public void testEOFShouldNotCloseStream() throws Exception {
        final File filePath = new File(TestUtils.DATA_DIR + "trio.vcf");//short file
        final InputStreamReader is = new InputStreamReader(new FileInputStream(filePath));
        new AsynchronousLineReader(is);
        Thread.sleep(2000L);
        is.read(); //this should not blow up
    }
```

blows up like this
```
java.io.IOException: Stream closed

	at sun.nio.cs.StreamDecoder.ensureOpen(StreamDecoder.java:46)
	at sun.nio.cs.StreamDecoder.read(StreamDecoder.java:148)
	at sun.nio.cs.StreamDecoder.read0(StreamDecoder.java:127)
	at sun.nio.cs.StreamDecoder.read(StreamDecoder.java:112)
	at java.io.InputStreamReader.read(InputStreamReader.java:168)
	at htsjdk.tribble.readers.AsynchronousLineReaderTest.testEOFShouldNotCloseStream(AsynchronousLineReaderTest.java:41)
```

d-cameron:
That's exactly the sort of test I was thinking of.

akiezun:
Thanks @d-cameron. I'm closing this PR in favor of https://github.com/samtools/htsjdk/issues/579

droazen:
@akiezun I would be in favor of removing this class -- it appears to be far too dangerous for anyone to use.

akiezun:
@droazen can you add your comment to https://github.com/samtools/htsjdk/issues/579?

Issue #580, opened 2016-04-20
Not immutable Genotypes from GenotypeBuilder regarding extended attributes 


magicDGS:
### Subject of the issue
When I try to re-use a `GenotypeBuilder` with an already initialized extended attribute map, the previously generated `Genotype`s contain the new attribute too. That means that a `Genotype` could be modified if you have access to the `GenotypeBuilder` that make it.

This is happening because the `extendedAttributes` field is passed by reference to the `FastGenotype` constructor. Changing the `FastGenotype` constructor will solve the problem.

### Your environment
* htsjdk 2.1.1
* Java(TM) SE Runtime Environment (build 1.8.0_60-b27)

### Steps to reproduce
```java
public class GenotypeBuilderFailTest {
	@Test
	public void test() {
		final GenotypeBuilder gb = new GenotypeBuilder("test");
		final int[] first = new int[]{1, 2};
		final int[] second = new int[]{3, 4};
		final Genotype firstG = gb.attribute("first", first).make();
		final Genotype secondG = gb.attribute("second", second).make();
		// both genotypes have the first field
		Assert.assertEquals(first, firstG.getExtendedAttribute("first"));
		Assert.assertEquals(first, secondG.getExtendedAttribute("first"));
		// only the second genotype should have the second field, but next line fails
		Assert.assertEquals(null, firstG.getExtendedAttribute("second"));
		Assert.assertEquals(second, secondG.getExtendedAttribute("second"));
	}
}
```

### Expected behaviour
A `Genotype` created by a `GenotypeBuilder` can't be modify if the builder is reused.

### Actual behaviour
A `Genotype` created by a `GenotypeBuilder` will contain extended attributes from later calls of `builder.attribute(key, value)`, unless the builder is reset.

lindenb:
related to https://github.com/samtools/htsjdk/issues/541

magicDGS:
Closed by #581.

PR #581, opened 2016-04-20
GenotypeBuilder.makeWithShallowCopy() for safely re-use it

magicDGS:
### Description

Fixing `GenotypeBuilder`, that generates mutable `Genotypes` (regarding extended attributes). Found in #580.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/FastGenotype.java
+++ b/src/main/java/htsjdk/variant/variantcontext/FastGenotype.java
@@ -29,7 +29,10 @@ import java.util.List;
 import java.util.Map;
 
 /**
- * This class encompasses all the basic information about a genotype.  It is immutable.
+ * This class encompasses all the basic information about a genotype.
+ *
+ * For the sake of performance, it does not make a copy of the Collections/arrays it's constructed from, and so
+ * subsequent changes to those Collections/arrays will be reflected in the FastGenotype object
  *
  * A genotype has several key fields
  *
--- a/src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
+++ b/src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
@@ -28,6 +28,7 @@ package htsjdk.variant.variantcontext;
 import htsjdk.tribble.util.ParsingUtils;
 import htsjdk.variant.vcf.VCFConstants;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
@@ -48,6 +49,11 @@ import java.util.Map;
  * this builder.  Can be called multiple times to create independent copies,
  * or with intervening sets to conveniently make similar Genotypes with
  * slight modifications.
+ *
+ * Re-using the same GenotypeBuilder to build multiple Genotype objects via calls
+ * to make() is dangerous, since reference types in the builder (eg., Collections/arrays)
+ * don't get copied when making each Genotype. To safely re-use the same builder object
+ * multiple times, use makeWithShallowCopy() instead of make().
  *
  * @author Mark DePristo
  * @since 06/12
@@ -185,11 +191,32 @@ public final class GenotypeBuilder {
      * created, althrough the contents of array values like PL should never be modified
      * inline as they are not copied for efficiency reasons.
      *
+     * Note: if attributes are added via this builder after a call to make(), the new Genotype will
+     * be modified. Use {@link #makeWithShallowCopy} to safely re-use the same builder object
+     * multiple times.
+     *
      * @return a newly minted Genotype object with values provided from this builder
      */
     public Genotype make() {
-        final Map<String, Object> ea = extendedAttributes == null ? NO_ATTRIBUTES : extendedAttributes;
+        final Map<String, Object> ea = (extendedAttributes == null) ? NO_ATTRIBUTES : extendedAttributes;
         return new FastGenotype(sampleName, alleles, isPhased, GQ, DP, AD, PL, filters, ea);
+    }
+
+    /**
+     * Create a new Genotype object using the values set in this builder, and perform a
+     * shallow copy of reference types to allow safer re-use of this builder
+     *
+     * After creation the values in this builder can be modified and more Genotypes
+     * created.
+     *
+     * @return a newly minted Genotype object with values provided from this builder
+     */
+    public Genotype makeWithShallowCopy() {
+        final Map<String, Object> ea = (extendedAttributes == null) ? NO_ATTRIBUTES : new HashMap<>(extendedAttributes);
+        final List<Allele> al = new ArrayList<>(alleles);
+        final int[] copyAD = (AD == null) ? null : Arrays.copyOf(AD, AD.length);
+        final int[] copyPL = (PL == null) ? null : Arrays.copyOf(PL, PL.length);
+        return new FastGenotype(sampleName, al, isPhased, GQ, DP, copyAD, copyPL, filters, ea);
     }
 
     /**
@@ -303,9 +330,9 @@ public final class GenotypeBuilder {
     }
 
     /**
-     * This genotype has these attributes.
+     * This genotype has these attributes. Attributes are added to previous ones.
      *
-     * Cannot contain inline attributes (DP, AD, GQ, PL)
+     * Cannot contain inline attributes (DP, AD, GQ, PL). Note: this is not checked
      * @return
      */
     public GenotypeBuilder attributes(final Map<String, Object> attributes) {
@@ -327,7 +354,7 @@ public final class GenotypeBuilder {
     /**
      * This genotype has this attribute key / value pair.
      *
-     * Cannot contain inline attributes (DP, AD, GQ, PL)
+     * Cannot contain inline attributes (DP, AD, GQ, PL). Note: this is not checked
      * @return
      */
     public GenotypeBuilder attribute(final String key, final Object value) {
--- a/src/test/java/htsjdk/variant/variantcontext/GenotypeBuilderTest.java
+++ b/src/test/java/htsjdk/variant/variantcontext/GenotypeBuilderTest.java
@@ -1,0 +1,75 @@
+/*
+* Copyright (c) 2016 The Broad Institute
+*
+* Permission is hereby granted, free of charge, to any person
+* obtaining a copy of this software and associated documentation
+* files (the "Software"), to deal in the Software without
+* restriction, including without limitation the rights to use,
+* copy, modify, merge, publish, distribute, sublicense, and/or sell
+* copies of the Software, and to permit persons to whom the
+* Software is furnished to do so, subject to the following
+* conditions:
+*
+* The above copyright notice and this permission notice shall be
+* included in all copies or substantial portions of the Software.
+*
+* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+* OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+* NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+* HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+* WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+* FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR
+* THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
+package htsjdk.variant.variantcontext;
+
+import htsjdk.variant.VariantBaseTest;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class GenotypeBuilderTest extends VariantBaseTest {
+
+    @Test
+    public void testMakeWithShallowCopy() {
+        final GenotypeBuilder gb = new GenotypeBuilder("test");
+        final List<Allele> alleles = new ArrayList<>(
+                Arrays.asList(Allele.create("A", true), Allele.create("T")));
+        final int[] ad = new int[]{1,5};
+        final int[] pl = new int[]{1,6};
+        final int[] first = new int[]{1, 2};
+        final int[] second = new int[]{3, 4};
+        final Genotype firstG = gb.alleles(alleles).attribute("first", first).makeWithShallowCopy();
+        final Genotype secondG = gb.AD(ad).PL(pl).attribute("second", second).makeWithShallowCopy();
+        // both genotypes have the first field
+        Assert.assertEquals(first, firstG.getExtendedAttribute("first"));
+        Assert.assertEquals(first, secondG.getExtendedAttribute("first"));
+        // both genotypes have the the alleles
+        Assert.assertEquals(alleles, firstG.getAlleles());
+        Assert.assertEquals(alleles, secondG.getAlleles());
+        // only the second genotype should have the AD field
+        Assert.assertNull(firstG.getAD());
+        Assert.assertEquals(ad, secondG.getAD());
+        // only the second genotype should have the PL field
+        Assert.assertNull(firstG.getPL());
+        Assert.assertEquals(pl, secondG.getPL());
+        // only the second genotype should have the second field
+        Assert.assertNull(firstG.getExtendedAttribute("second"));
+        Assert.assertEquals(second, secondG.getExtendedAttribute("second"));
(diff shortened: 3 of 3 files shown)


nh13:
The idea behind `FastGenotype` from what I can see is that values passed to it are [not copied for efficiency reasons](https://github.com/samtools/htsjdk/blob/master/src/java/htsjdk/variant/variantcontext/GenotypeBuilder.java#L186).  I think the doc for `GenotypeBuilder.make` should be improved to state this.  The changes you make may reduce the efficiency of `FastGenotype`.

To support your use case I would suggest a method added to `GenotypeBuilder` that allows for full re-use of the builder and it's members, either through this new `make` method performing the copy, or by using a different class that extends `Genotype` (`SlowGenotype`?).

magicDGS:
The javadoc for `FastGenotype` says that it is inmutable and it could be only be created by a `GenotypeBuilder`; on the other hand, `GenotypeBuilder` says that it could be used to create multiple `Genotypes`.

It seems that if the only way of create a `FastGenotype` is through a `GenotypeBuilder`, it does not make sense that it creates mutable `Genotypes`. Of course, instead of changing the constructor for `FastGenotype` the code in `GenotypeBuilder.make()` could be change, but I think that the result in terms of efficiency will be the same...

Another thing that I found when going through the code is that the `extendedAttributes` map is never checked by the builder to make sure that it not contains inline-coded int and int[] (GQ, DP, AD, PL). This could generate a `FastGenotype` that is not following the contract.

What do you suggest for all this, @nh13? Changing the `GenotypeBuilder.make()` method for copying the map solve the initial issue, and the second could be performed when attributes are added to the builder.

I can add tests for all this too.

nh13:
I am trying to divine the thinking of @depristo who likely wrote this originally.  Looking more closely, it seems that `FastGenotype` is unmodifiable not immutable for exactly the reasons you state.  

@droazen since you have looked at this code a lot (and git blames you for moving this code over), could you comment on this proposal?
1) update the doc in `FastGenotype` to state it is unmodifiable not immutable, since AD/PL/extendedAttributes/alleles could all be modified outside of `FastGenotype`.
2) create a new method in `GenotypeBuilder` called `makeImmutable` that either (a) makes new copies of objects that could be mutated prior to constructing `FastGenotype`, or (b) create a new class `ImmutableGenotype` that makes copies in its constructor.

depristo:
Hi all!

Originally there was a generally-purpose Genotype object that didn't have inline, native types for any fields, instead storing the map from field to values, so we had "PL" => List<Integer> or whatever it was at the time. This proved to be *ridiculous* expensive for actually working with the objects. FastGenotype was intended as high-performance version that parsed key genotype field values into native java data structures. Sounds like there are still a few rough edges in the codebase. Glad to see that bugs are still being weeded out after 5 years!

Hope everyone is doing well.

magicDGS:
@droazen, what should I do here after the review-party?

src/main/java/htsjdk/variant/variantcontext/FastGenotype.java
```diff
@@ -29,7 +29,7 @@
 import java.util.Map;
 
 /**
- * This class encompasses all the basic information about a genotype.  It is immutable.
+ * This class encompasses all the basic information about a genotype.  It is unmodifiable, but not immutable.
```
droazen: `"unmodifiable, but not immutable"` is a rather confusing phrase. I'd replace with something more explicit like `For the sake of performance, it does not make a copy of the Collections/arrays it's constructed from, and so subsequent changes to those Collections/arrays will be reflected in the FastGenotype object`

src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
```diff
@@ -185,14 +186,31 @@ public final void reset(final boolean keepSampleName) {
...
     /**
+     * Create a new Genotype object using the values set in this builder.
+     *
+     * After creation the values in this builder can be modified and more Genotypes
+     * created, although the contents of array values like PL should never be modified
+     * inline as they are not copied for efficiency reasons.
+     *
+     * @return a newly minted Genotype object with values provided from this builder
+     */
+    public Genotype makeImmutable() {
```
droazen: This name is also confusing -- recommend renaming to `makeWithShallowCopy()` to make it clear how it differs from `make()`

src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
```diff
@@ -185,14 +186,31 @@ public final void reset(final boolean keepSampleName) {
      * created, althrough the contents of array values like PL should never be modified
      * inline as they are not copied for efficiency reasons.
      *
+     * Note: if attributes are added, the new Genotype will be modified
```
droazen: `if attributes are added` -> `if attributes are added via this builder after a call to make()`

src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
```diff
@@ -28,6 +28,7 @@
 import htsjdk.tribble.util.ParsingUtils;
 import htsjdk.variant.vcf.VCFConstants;
```
droazen: Add disclaimer at the top of this class: `Re-using the same GenotypeBuilder to build multiple Genotype objects via calls to make() is dangerous, since reference types in the builder (eg., Collections/arrays) don't get copied when making each Genotype. To safely re-use the same builder object multiple times, use makeWithShallowCopy() instead of make().`

src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
```diff
@@ -185,14 +186,31 @@ public final void reset(final boolean keepSampleName) {
...
+        final Map<String, Object> ea = (extendedAttributes == null) ? NO_ATTRIBUTES : extendedAttributes;
         return new FastGenotype(sampleName, alleles, isPhased, GQ, DP, AD, PL, filters, ea);
     }
 
     /**
+     * Create a new Genotype object using the values set in this builder.
+     *
+     * After creation the values in this builder can be modified and more Genotypes
+     * created, although the contents of array values like PL should never be modified
+     * inline as they are not copied for efficiency reasons.
```
droazen: This comment is inconsistent with the actual code in the method (the PL array is copied)

src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
```diff
@@ -185,14 +186,31 @@ public final void reset(final boolean keepSampleName) {
...
      * @return a newly minted Genotype object with values provided from this builder
      */
     public Genotype make() {
-        final Map<String, Object> ea = extendedAttributes == null ? NO_ATTRIBUTES : extendedAttributes;
+        final Map<String, Object> ea = (extendedAttributes == null) ? NO_ATTRIBUTES : extendedAttributes;
         return new FastGenotype(sampleName, alleles, isPhased, GQ, DP, AD, PL, filters, ea);
     }
 
     /**
+     * Create a new Genotype object using the values set in this builder.
```
droazen: `Create a new Genotype object using the values set in this builder` -> `Create a new Genotype object using the values set in this builder, and perform a shallow copy of reference types to allow safer re-use of this builder`

src/test/java/htsjdk/variant/variantcontext/GenotypeBuilderTest.java
```diff
@@ -0,0 +1,49 @@
...
+package htsjdk.variant.variantcontext;
+
+import htsjdk.variant.VariantBaseTest;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class GenotypeBuilderTest extends VariantBaseTest {
+
+    @Test
+    public void testInmutableGenotypes() {
```
droazen: `testInmutableGenotypes` -> testMakeWithShallowCopy()` after doing the rename suggested above

src/test/java/htsjdk/variant/variantcontext/GenotypeBuilderTest.java
```diff
@@ -0,0 +1,49 @@
...
+        final GenotypeBuilder gb = new GenotypeBuilder("test");
+        final int[] first = new int[]{1, 2};
+        final int[] second = new int[]{3, 4};
+        final Genotype firstG = gb.attribute("first", first).makeImmutable();
+        final Genotype secondG = gb.attribute("second", second).makeImmutable();
+        // both genotypes have the first field
+        Assert.assertEquals(first, firstG.getExtendedAttribute("first"));
+        Assert.assertEquals(first, secondG.getExtendedAttribute("first"));
+        // only the second genotype should have the second field, but next line fails
```
droazen: Remove "but next line fails", since it doesn't fail with your changes, I assume

droazen:
Review complete -- back to @magicDGS for changes.

src/main/java/htsjdk/variant/variantcontext/FastGenotype.java
magicDGS: Done

src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
magicDGS: Done

src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
magicDGS: Done

src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
magicDGS: Removed

src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
magicDGS: Done

src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
magicDGS: Done

src/test/java/htsjdk/variant/variantcontext/GenotypeBuilderTest.java
magicDGS: Done

src/test/java/htsjdk/variant/variantcontext/GenotypeBuilderTest.java
magicDGS: Removed

magicDGS:
Done, back to @droazen. Once it's ok, I would squash/rebase before merging.

droazen:
@magicDGS Your test case is failing with a NPE:

```
Gradle suite > Gradle test > htsjdk.variant.variantcontext.GenotypeBuilderTest.testMakeWithShallowCopy FAILED
        at htsjdk.variant.variantcontext.GenotypeBuilder.makeWithShallowCopy(GenotypeBuilder.java:215)
        at htsjdk.variant.variantcontext.GenotypeBuilderTest.testMakeWithShallowCopy(GenotypeBuilderTest.java:39)
```

magicDGS:
Sorry, it was because of the AD/PL copy when they are `null`. I think that now it should pass, @droazen

src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
```diff
@@ -182,17 +188,36 @@ public final void reset(final boolean keepSampleName) {
...
+     * created
+     * inline as they are not copied for efficiency reasons.
+     *
+     * @return a newly minted Genotype object with values provided from this builder
+     */
+    public Genotype makeWithShallowCopy() {
+        final Map<String, Object> ea = (extendedAttributes == null) ? NO_ATTRIBUTES : new HashMap<>(extendedAttributes);
+        final List<Allele> al = new ArrayList<>(alleles);
+        final int[] copyAD = (AD == null) ? null : Arrays.copyOf(AD, AD.length);
+        final int[] copyPL = (AD == null) ? null : Arrays.copyOf(PL, PL.length);
```
droazen: Should be `PL == null`, I believe

src/test/java/htsjdk/variant/variantcontext/GenotypeBuilderTest.java
```diff
@@ -0,0 +1,49 @@
...
+        final int[] first = new int[]{1, 2};
+        final int[] second = new int[]{3, 4};
+        final Genotype firstG = gb.attribute("first", first).makeWithShallowCopy();
+        final Genotype secondG = gb.attribute("second", second).makeWithShallowCopy();
+        // both genotypes have the first field
+        Assert.assertEquals(first, firstG.getExtendedAttribute("first"));
+        Assert.assertEquals(first, secondG.getExtendedAttribute("first"));
+        // only the second genotype should have the second field
+        Assert.assertEquals(null, firstG.getExtendedAttribute("second"));
+        Assert.assertEquals(second, secondG.getExtendedAttribute("second"));
```
droazen: In your test for `makeWithShallowCopy()`, you should test all the reference fields that get shallow copied, including the alleles, AD, and PL, not just the attributes.

droazen:
Two remaining comments to address, then rebase and squash and we'll merge this.

src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
magicDGS: Sorry, copy-paste error. Fixed.

src/test/java/htsjdk/variant/variantcontext/GenotypeBuilderTest.java
magicDGS: Done.

magicDGS:
Ready for merge. Thanks a lot, @droazen.

src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
```diff
@@ -182,17 +188,36 @@ public final void reset(final boolean keepSampleName) {
...
         return new FastGenotype(sampleName, alleles, isPhased, GQ, DP, AD, PL, filters, ea);
     }
 
     /**
+     * Create a new Genotype object using the values set in this builder, and perform a
+     * shallow copy of reference types to allow safer re-use of this builder
+     *
+     * After creation the values in this builder can be modified and more Genotypes
+     * created
+     * inline as they are not copied for efficiency reasons.
```
droazen: Here the "inline as they are not copied for efficiency reasons." should be deleted, since it's not true for this method.

src/test/java/htsjdk/variant/variantcontext/GenotypeBuilderTest.java
```diff
@@ -0,0 +1,58 @@
...
+        final Genotype secondG = gb.AD(ad).PL(pl).attribute("second", second).makeWithShallowCopy();
+        // both genotypes have the first field
+        Assert.assertEquals(first, firstG.getExtendedAttribute("first"));
+        Assert.assertEquals(first, secondG.getExtendedAttribute("first"));
+        // only the second genotype should have the AD field
+        Assert.assertNull(firstG.getAD());
+        Assert.assertEquals(ad, secondG.getAD());
+        // only the second genotype should have the PL field
+        Assert.assertNull(firstG.getPL());
+        Assert.assertEquals(pl, secondG.getPL());
```
droazen: A better test is to modify the `ad` and `pl` arrays, and show that `secondG`'s ADs and PLs are unaffected.

droazen: And do the same for the alleles

droazen:
Found a few more issues -- let me know once they're fixed.

src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
```diff
@@ -182,17 +188,36 @@ public final void reset(final boolean keepSampleName) {
      * Create a new Genotype object using the values set in this builder.
      *
      * After creation the values in this builder can be modified and more Genotypes
-     * created, althrough the contents of array values like PL should never be modified
-     * inline as they are not copied for efficiency reasons.
+     * created
```
magicDGS: Done

src/main/java/htsjdk/variant/variantcontext/GenotypeBuilder.java
magicDGS: Done

src/test/java/htsjdk/variant/variantcontext/GenotypeBuilderTest.java
magicDGS: Done

magicDGS:
Comments addressed, @droazen. I will rebase/squash when you tell me.

droazen:
Can you squash before I review? Thanks!

magicDGS:
Done.

droazen:
:+1: merging

Issue #568, opened 2016-04-13
A proposal for improving the stability of htsjdk releases, and isolating production users from htsjdk development 


droazen:
Htsjdk development has entered a riskier phase in the past year or so, with major new subsystems like CRAM and SRA support hitting master, as well as efforts to improve performance by relying on third-party libraries like the Intel deflator and features like asynchronous I/O. The end result has been that the master branch is less stable than it was in the past, and the HEAD of master should definitely not be considered production-ready these days (if it ever was!). This has created a problem for Broad production, which does use the latest htsjdk HEAD when building Picard for production use. After meeting to discuss this problem, we came up with the following proposal:

* We will put together heavy-duty end-to-end acceptance tests for htsjdk that are stricter than the existing htsjdk unit tests. The initial version of these tests will involve running the full Picard, GATK4, and GATK3 test suites on the version of htsjdk being tested, but they may later evolve into something more sophisticated.

* We will only tag and release builds of htsjdk that pass these new heavy-duty acceptance tests.

* Passing htsjdk unit tests will continue to be a sufficient condition for merging a branch into htsjdk/master.

* After the first htsjdk release is made that passes the new heavy-duty tests, Picard will switch to depending on the latest tagged/released version of htsjdk, instead of HEAD, and we will make the necessary changes to its build system to enable this.

* If Broad production needs a hotfix to an existing htsjdk release, the procedure will be:
    * Create a branch based on the latest htsjdk release (not HEAD)
    * Implement the hotfix in this branch
    * Tag and release to maven the hotfixed version once it passes the heavy-duty testing, and satisfies
       the additional criteria that prompted the hotfix in the first place
    * Cherry-pick the hotfix into htsjdk/master as well, so that it will be part of the next release
       that is built from master
    * Open an issue in htsjdk's github describing the problem with the previous release that the heavy-duty 
       acceptance tests failed to detect, and suggest a test case that could detect it in the future.

@ktibbett @ndbolliger @bradtaylor @eitanbanks @akiezun @lbergelson @yfarjoun Does this adequately capture the consensus we arrived at in our meeting? Did I miss anything? Are there any lingering concerns about this plan?

yfarjoun:
Could you clarify what branch the releases will be on if not master?

It would be good to also solicit comment/concerns/suggestions from htsjdk's
many other users.

droazen:
[...]

PR #582, opened 2016-04-22
Adding downstream build badges to Readme

lbergelson:
Adding badges for the gatk4 and picard tests running on htsjdk master's branch.  This is intended to assist with #568.  Tests should also be able to be manually invoked on a given htsjdk branch by anyone with github push access to gatk.  In the future this should be amended to include htsjdk committers but may be a setup hassle.

See also https://github.com/broadinstitute/gatk/issues/1705

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,8 @@
-[![Build Status](https://travis-ci.org/samtools/htsjdk.svg?branch=master)](https://travis-ci.org/samtools/htsjdk)
+Status of master branch build: [![Build Status](https://travis-ci.org/samtools/htsjdk.svg?branch=master)](https://travis-ci.org/samtools/htsjdk)
+
+Status of downstream projects automatically built on top of the current htsjdk master branch. See [gatk-jenkins](https://gatk-jenkins.broadinstitute.org/view/HTSJDK%20Release%20Tests/) for detailed logs. Failure may indicate problems  in htsjdk, but may also be due to expected incompatibilities between versions, or unrelated failures in downstream projects.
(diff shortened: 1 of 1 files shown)


droazen:
:+1: merge when ready @lbergelson

nh13:
@lbergelson @droazen 

A I am torn on this commit.  I don't see this commit being proposed in #568 (and necessary for it).  Nonetheless, it is likely a good short term addition.  Could you please open an issue to *remove* these badges, as likely soon both Picard and GATK4 will not rely on the HEAD of htsjdk?  As I read #568, it was that there was a way to test master against Picard/GATK4/GATK3 before releasing, not that there would be status badges in `htsjdk` for this.

droazen:
@nh13 I think it's important that everyone be able to see the status of these tests, so that we know immediately when a change in master has broken a downstream project. If the badges are clearly labelled to avoid confusion, I don't see the problem? Can you explain your concerns?

droazen:
Also, this was part of the proposal in https://github.com/samtools/htsjdk/issues/568:

    Non-Broadies should have access to at least view the "heavy duty" test suite results (for master, run on every merge -- we're not sure we'll have the resources to run it for non-master branches as well). Ideally we'll be able to see their status via github badges.

nh13:
I'll submit a PR so you can see how I would have formatted it.  I am trying to balance the optics that htsjdk is purely built for Broad Institute tools, which historically it was, with it being under the samtools organization, and so public and independent.  It's definitely a tough balance.

These were just my two cents being an outsider from Picard and GATK (and Broad Institute), and the maintainers of `htsjdk` should feel free to disagree and ignore if there is consensus.  `¯\_(ツ)_/¯`.

nh13:
@droazen @lbergelson I was thinking something like this?  https://gist.github.com/nh13/531b766ee86f934694fd5e1143357cf4 
It's not much different, but it does say in bold that the integration tests **must** pass to consider an `htsjdk` release.

akiezun:
@nh13 i like it!

lbergelson:
@nh13 +1

PR #584, opened 2016-04-24
AsyncBufferedIterator implementation to improve throughput by perform…

d-cameron:
AsyncBufferedIterator implementation to improve throughput by performing read-ahead on a background thread

This implementation has feedback from #482 incorporated. Readability should be much improved compared to the last version.

### DIFF
--- a/src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
+++ b/src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
@@ -1,0 +1,261 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Daniel Cameron
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package htsjdk.samtools.util;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Iterator that uses a dedicated background thread to perform read-ahead to improve
+ * throughput at the expense of increased latency. This iterator will block
+ * until the background thread has read a full buffer of records.
+ * 
+ * Note that this implementation is not synchronized. If multiple threads
+ * access an instance concurrently, it must be synchronized externally. 
+ * 
+ * @author Daniel Cameron
+ *
+ */
+public class AsyncBufferedIterator<T> implements CloseableIterator<T> {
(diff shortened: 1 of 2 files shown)


src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,261 @@
...
+    private final int bufferSize;
+    /**
+     * A dedicated background thread is required since these iterators can be chained
+     * thus able to block on each other. Usage of a thread pool would result in
+     * a deadlock due to task dependencies.
+     */
+    private Thread backgroundThread;
+    private final Iterator<T> underlyingIterator;
+    private final BlockingQueue<RecordBlock> buffer;
+    private RecordBlock currentBlock = new RecordBlock(Collections.emptyList());
```
akiezun: I think `RecordBlock` is a strange name (both `block` and `record` have specific meanings that are not the meanings we use here). How about `BufferedList<T>`?  (I think we should make that class static and parameteric in `T` to make clear that it's a container`

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,261 @@
...
+    public boolean hasNext() {
+        if (backgroundThread == null) {
+            throw new IllegalStateException("iterator has been closed");
+        }
+        if (currentBlock.hasNext()) return true;
+        
+        // rethrow any exceptions raised on the background thread
+        // at the point the exception would have been encountered
+        // if we had performed synchronous iteration
+        Throwable t = currentBlock.getException();
```
akiezun: extract this rethrowing to a new method

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,261 @@
...
+            } else if (t instanceof RuntimeException) {
+                throw (RuntimeException) t;
+            } else {
+                throw new RuntimeException(t);
+            }
+        }
+        if (currentBlock.isEndOfStream()) return false;
+        try {
+            // try loading next block
+            currentBlock = buffer.take();
```
akiezun: what happens if the main thread is blocked here on an empty list and the worker thread blows up somewhere in `readAhead` or `backgroundRun` and dies? Will the main thread block forever?

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,261 @@
...
+        } catch (InterruptedException e) {
+            throw new RuntimeException("Error reading from background thread", e);
+        }
+        return hasNext();
```
akiezun: this recursive call is very unusual. Definitely document why it's done this way and why it will not cause an unbounded recursion.

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -0,0 +1,261 @@
...
+            do {
+                block = readAhead();
+                if (block.getException() instanceof InterruptedException) {
+                    // stop thread immediately if we've been told to stop
+                    return;
+                }
+                buffer.put(block);
+            } while (!block.isEndOfStream());
+        } catch (InterruptedException e) {
+            // stop thread
```
akiezun: as above: what happens if this thread dies while the buffer is empty and the main thread is blocking on the `take` call?

akiezun:
back to @d-cameron - I would like to understand better why the main thread does not potentially block forever if the worker dies when the queue is empty.

d-cameron:
Everything except InterruptedException is caught by the worker and put into the buffer the calling thread is waiting on to raise on the foreground thread. The only way to deadlock is for a third thread to interrupt the background thread (causing it to stop immediately) and not the blocked foreground thread.

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
```diff
@@ -129,17 +129,43 @@ public void close() {
             }
         }
     }
+    
+    private void ensureHasNext() {
+        if (!currentBlock.hasNext()) {
+            // Rethrow any exceptions raised on the background thread
+            // at the point the exception would have been encountered
+            // if we had performed synchronous iteration
+            raiseBackgroundThreadException();
```
akiezun: So any exception from the background thread will be thrown only at the **end** of processing the read-ahead block of `T`s, not at the beginning. Is that intended? I'd think that throwing as fast as we know there was a problem would be preferable.

akiezun:
Looks good @d-cameron. One last question about the timing of the rethrow.

src/java/htsjdk/samtools/util/AsyncBufferedIterator.java
d-cameron: By throwing at the end of the block, the exception timing will match the exception timing from iterating over the underlying iterator. If your underlying iterator blows up on a particular record with a 100 record async buffer, you want your aysnc iterator to also blow up on that same record - not a record somewhere between 0 and 99 records beforehand.

By delaying the exception timing, the async iterator will have the same exception behavior as using the iterator directly.

d-cameron: This works because once an exception is encountered, the read-ahead block is returned immediately, regardless of the number of records that were added.

akiezun:
ok, this looks good to me. Given that the code is not connected yet to rest of the codebase, the risk is minimal. Unless someone objects, I'll merge in 1 day (ie Friday). Thanks @d-cameron

PR #587, opened 2016-04-27
Changed VariantContextBuilder.attributes() to copy the supplied map

tfenne:
### Description

Changed VariantContextBuilder.attributes() to copy the supplied map in order to avoid:

1) Modification of the map after it is passed to VariantContextBuilder
2) Storing an unmodifiable map that cannot be added to later

Picard's LiftoverIntervalList currently fails if `WRITE_ORIGINAL_POSITION` is set to `true` because it calls `VariantContextBuilder.atttributes(oldContext.getAttributes())` which sets an `UnmodifiableMap` on the builder, which on subsequent calls to `attribute(key, value)` fails with this exception:

```
Exception in thread "main" java.lang.UnsupportedOperationException
	at java.util.Collections$UnmodifiableMap.put(Collections.java:1457)
	at htsjdk.variant.variantcontext.VariantContextBuilder.attribute(VariantContextBuilder.java:212)
	at picard.vcf.LiftoverVcf.doWork(LiftoverVcf.java:249)
```
### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
+++ b/src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
@@ -178,7 +178,10 @@ public class VariantContextBuilder {
     }
 
     /**
-     * Tells this builder to use this map of attributes alleles for the resulting <code>VariantContext</code>
+     * Tells this builder to use this map of attributes for the resulting <code>VariantContext</code>. The
+     * contents of the Map are copied to a new Map to ensure that modifications to the provided Map post-invocation
+     * don't affect the VariantContext and also to ensure additional attributes can be added in case the provided
+     * map doesn't support changes (e.g. UnmodifiableMap).
      *
      * Attributes can be <code>null</code> -&gt; meaning there are no attributes.  After
      * calling this routine the builder assumes it can modify the attributes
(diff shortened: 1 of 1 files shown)


nh13:
@tfenne I think you mean `LiftoverVcf` not `LiftoverIntervalList`.  

Another possibility is to check if the supplied `Map` is an instance of `ImmutableMap` or `UnmodifiableMap` and handle it appropriately by setting `this.attributesCanBeModified = false`?

tfenne:
@nh13 I did mean `LiftoverVcf`, you're correct.  

I don't really want to test for specific types of unmodifable maps - what if it is passed another implementation that doesn't support `put()`?  I think the defensive copy is the right route.

src/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
```diff
@@ -178,7 +178,7 @@ public VariantContextBuilder alleles(final String ... alleleStrings) {
     }
 
     /**
-     * Tells this builder to use this map of attributes alleles for the resulting <code>VariantContext</code>
+     * Tells this builder to use this map of attributes for the resulting <code>VariantContext</code>
      *
```
droazen: Document the fact that this method now makes a defensive copy of the map it's passed.

droazen:
@tfenne This does seem like the right thing to do, though we reserve the right to revisit this if it shows up in the profiler for GATK, etc., since this is a very frequently-called method. One comment to address then rebase/merge at will.

tfenne:
Sounds good @droazen .  If performance becomes a concern, let's look for an alternative solution before straight up reverting the change.

PR #589, opened 2016-04-27
Utility methods on CloseableIterator to make life a little easier.

tfenne:
### Description

Scala has spoiled me, and I'm tired of Java's iterators being second class citizens.  This commit adds a pair of default methods to `CloseableIterator` which will benefit most iterators in htsjdk, and allows easy transformation to a `List` and a `Stream`.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/java/htsjdk/samtools/util/CloseableIterator.java
+++ b/src/java/htsjdk/samtools/util/CloseableIterator.java
@@ -42,8 +42,6 @@ import java.util.stream.StreamSupport;
  * 2) When hasNext() returns false, the iterator implementation should automatically close itself.
  *    The latter makes it somewhat safer for consumers to use the for loop syntax for iteration:
  *    for (Type obj : getCloseableIterator()) { ... }
- * 
- * We do not inherit from java.io.Closeable because IOExceptions are a pain to deal with.
  */
 public interface CloseableIterator<T> extends Iterator<T>, Closeable {
     /** Should be implemented to close/release any underlying resources. */
@@ -53,12 +51,13 @@ public interface CloseableIterator<T> extends Iterator<T>, Closeable {
     default List<T> toList() {
         final List<T> list = new ArrayList<>();
         while (hasNext()) list.add(next());
+        close();
         return list;
     }
 
     /** Returns a Stream that will consume from the underlying iterator. */
     default Stream<T> stream() {
         final Spliterator<T> s = Spliterators.spliteratorUnknownSize(this, Spliterator.ORDERED);
-        return StreamSupport.stream(s, false);
+        return StreamSupport.stream(s, false).onClose(this::close);
     }
 }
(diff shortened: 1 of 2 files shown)


src/java/htsjdk/samtools/util/CloseableIterator.java
```diff
@@ -40,6 +46,19 @@
...
+    /** Consumes the contents of the iterator and returns it as a List. */
+    default List<T> toList() {
+        final List<T> list = new ArrayList<>();
+        while (hasNext()) list.add(next());
+        return list;
+    }
+
+    /** Returns a Stream that will consume from the underlying iterator. */
+    default Stream<T> stream() {
+        final Spliterator<T> s = Spliterators.spliteratorUnknownSize(this, Spliterator.ORDERED);
```
lbergelson: @tfenne Is there any way that we can set it up so the iterator is automatically closed when the stream is consumed?

src/java/htsjdk/samtools/util/CloseableIterator.java
```diff
@@ -40,6 +46,19 @@
...
 public interface CloseableIterator<T> extends Iterator<T>, Closeable {
+    /** Should be implemented to close/release any underlying resources. */
+    void close();
 
-    public void close();
+    /** Consumes the contents of the iterator and returns it as a List. */
+    default List<T> toList() {
+        final List<T> list = new ArrayList<>();
+        while (hasNext()) list.add(next());
+        return list;
```
lbergelson: Shouldn't this close the iterator before returning the stream?

tfenne: Good point. I'll fix that.

src/java/htsjdk/samtools/util/CloseableIterator.java
```diff
@@ -40,6 +46,19 @@
  * We do not inherit from java.io.Closeable because IOExceptions are a pain to deal with.
```
lbergelson: This comment seems out of date since it does inherit from `java.io.Closable`

src/java/htsjdk/samtools/util/CloseableIterator.java
tfenne: I agree it would be nice, but I'm not sure how possible it is without a bunch of wrapper-ing of the iterators.  Let me see if I can come up with something that's not too ugly.

lbergelson: I don't see an easy way unless stream offers some sort of on-closed callback... I was hoping you'd know some magic...

lbergelson: Ahah, there is exactly such a method.  `Stream.onClose(Runnable)`

tfenne: Yup - I just found that too.  Updated to use it.

tfenne:
@lbergelson Changes are all made, can you take a quick final look please?

lbergelson:
👍 Thanks for this.  I definitely seem myself using the stream function in the future.  Merge when ready.

Issue #549, opened 2016-04-04
CRAMCompressionHeader skips unrecognized encoding tags


cmnbroad:
I've not seen this actually happen, but it looks like CRAMCompressionHeader [just discards](https://github.com/samtools/htsjdk/blob/master/src/java/htsjdk/samtools/cram/structure/CompressionHeader.java#L173) any encoding key that it doesn't recognize, which seems like it could lead to data loss.

@vadimzalunin Can you take a look at this ? Shouldn't the code throw instead of log when this happpens?

vadimzalunin:
I think this was done deliberately for compatibility some time ago. But I agree, default behaviour in htsjdk should be more strict.

cmnbroad:
@vadimzalunin 4 of the tests fail when I change the code to throw when it detects these. Can you take a look ?

PR #593, opened 2016-04-27
Throw on unrecognized CRAM encoding tag.

cmnbroad:
### Description
Fixes https://github.com/samtools/htsjdk/issues/549. Change the CRAM reader to throw when it sees an encoding tag it doesn't recognize.

The CRAM code currently logs and ignores unrecognized encoding tags, but we really should throw if we ever see one of these.

This is a purely defensive change - I don't really know of a way to force this condition.

### Checklist

- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/cram/encoding/reader/DataReaderFactory.java
+++ b/src/main/java/htsjdk/samtools/cram/encoding/reader/DataReaderFactory.java
@@ -56,9 +56,7 @@ public class DataReaderFactory {
                 final DataSeries dataSeries = field.getAnnotation(DataSeries.class);
                 final EncodingKey key = dataSeries.key();
                 final DataSeriesType type = dataSeries.type();
-                if (header.encodingMap.get(key) == null) {
-                    log.debug("Encoding not found for key: " + key);
-                } else {
+                if (header.encodingMap.get(key) != null) {
                     try {
                         field.set(reader,
                                 createReader(type, header.encodingMap.get(key), bitInputStream, inputMap));
--- a/src/main/java/htsjdk/samtools/cram/structure/CompressionHeader.java
+++ b/src/main/java/htsjdk/samtools/cram/structure/CompressionHeader.java
@@ -17,6 +17,7 @@
  */
 package htsjdk.samtools.cram.structure;
 
+import htsjdk.samtools.cram.CRAMException;
 import htsjdk.samtools.cram.encoding.ExternalCompressor;
 import htsjdk.samtools.cram.encoding.NullEncoding;
 import htsjdk.samtools.cram.io.ITF8;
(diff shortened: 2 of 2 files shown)


cmnbroad:
@vadimzalunin can you review this and make sure it makes sense ?

cmnbroad:
Hm. 4 of the tests fail with this change because the code detects an unrecognized encoding tag  "BB_bases".

test_processMultiContainer
testCRAMContainerStreamWithIndex
testMappedAndUnmappedContainer
testMultirefContainer

Deferring to @vadimzalunin .

droazen:
@vadimzalunin Can you have a look at this, and give us your thoughts on why this simple patch has caused tests to fail?

vadimzalunin:
Data series may be optional and this can be implemented in two way: 
1) data series is not mentioned in the  data series encoding map (part of compression header)
2) unused/empty data series encodings are explicitly set to NULL encoding.

The proposed change basically prohibits 1), so we require all data series to be present even if not used. But I think this does not solve any problems because data series can still be effectively missing via 2). I would suggest to replace the first exception with a warning.

Would it be useful to take into account validation stringency? 

The second exception on line 174 should be ok to throw as in most cases this means the impl is outdated.

PS The decision for silent decoding was due to TM encoding tag I think: C and Java impls conflicted on this some time ago and it was decided to ignore/warn about unknown data series. This provides some forward compatibility as well if new data series are added. 
In a way data series tags are similar to (generalization of) SAM tags, some of them have special meaning (like NM and MD) whereas by default SAM tags are not interpreted.

cmnbroad:
@vadimzalunin Thanks - that helps a bit. I was under the impression that the data series were all required (maybe I just made that assumption since the spec enumerates them and I don't recall it saying they were optional). Reverting the throw in the DataFactory code does allow the tests to pass - I'm not sure why we'd want to leave a warning there though if its really that case that ALL data series are optional - we'd be warning about something harmless. If on the other hand some data series really are required, maybe we should add an optional/required tag to the DataSeries annotation to use to decide how to handle the missing case.

It sounds like we should leave the throw in the CompressionHeader code path, though - thats the case where we see a data series in the input stream that the code can't handle. Is it the case that the list of data series are fixed for a given CRAM version ? I don't, for example, see a data series tag named "BB" in the spec anywhere, though the code seems to know about it.

vadimzalunin:
@cmnbroad the question about required data series is still open. Compare this for example with BAM file where all bases and quality scores are missing (replaced with a single star symbol). Another odd example would be all reads having the same name (effectively no name), perhaps user is not interested in read identities. These sort of tricks are ok for low-level BAM parsers but they may fail higher level validation checks.

yfarjoun:
@cmnbroad The ball seems to be in your court.

cmnbroad:
I think this reflects the changes we settled on and should be ready to merge now.

PR #594, opened 2016-04-28
Provide a method that for each source contig gives the set of destination contigs used in liftover.

nh13:
### Description

This method will give the destination contigs are used when mapping positions in a source contig.  This method allows `LiftoverVcf` to be more intelligent about memory usage (https://github.com/broadinstitute/picard/pull/525).

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/liftover/LiftOver.java
+++ b/src/main/java/htsjdk/samtools/liftover/LiftOver.java
@@ -32,7 +32,12 @@ import htsjdk.samtools.util.OverlapDetector;
 
 import java.io.File;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 /**
  * Java port of UCSC liftOver.  Only the most basic liftOver functionality is implemented.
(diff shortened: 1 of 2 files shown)


src/java/htsjdk/samtools/liftover/LiftOver.java
```diff
@@ -139,6 +143,29 @@ public Interval liftOver(final Interval interval, final double liftOverMinMatch)
         return ret;
     }
 
+    /**
+     * @return the set of destination contigs for each source contig in the chains file.
+     */
+    public Map<String, Set<String>> getContigMap() {
```
yfarjoun: since the chainfile is a member, wouldn't it make sense to store the contigMap as a member as well (initialize to null) ? otherwise you will recalculate this map every time someone calls getContigMap.

yfarjoun:
small comment regarding caching the result. looks fine otherwise!!

src/java/htsjdk/samtools/liftover/LiftOver.java
nh13: I would actually like to change this to be static rather than keeping a member to encourage folks to only call this only once.  Consider genomes that both have many chain entries and many contigs.

nh13: But alas, I need sleep, and the chain is a member.

nh13:
@yfarjoun thanks for the review, I will merge after the tests pass.

yfarjoun:
upon passing tests. 👍

PR #595, opened 2016-05-03
fix setter in SAMFileWiterFactory

lindenb:
### Description

small fix for factory-method: in SAMFileWiterFactory changed 

`public void setCompressionLevel(final int compressionLevel) `

to

`public SAMFileWriterFactory setCompressionLevel(final int compressionLevel) { `


droazen:
:+1: merging

Issue #597, opened 2016-05-06
Make interface changes to allow CRAM to work with IGV


droazen:
From @jrobinso, there are a few small adjustments we need to make to `CRAMFileReader` and  `ReferenceSource` to allow CRAM to work with IGV:

    Here’s what’s missing wrt IGV requirements:

    (1) At the time I last tries all the constructors for CRAMFileReader  took File,  rather than
    SeekableStream,  and therefore did not support http.  This might have changed
    (2) I would like to supply the ReferenceSource object from  IGV,  as IGV user already has this loaded
    and the htsjdk implementation is too memory intensive for IGV use.    Currently this is a concrete
    class,  it would be easier for me to implement and supply my own if it were an interface.  This 
    conversion (class -> interface) is fairly easy,  Intellij has tools do to this.  It looks like I would only have
     to implement one method,     
    public synchronized byte[] getReferenceBases(final SAMSequenceRecord record, final boolean tryNameVariants) {
  
    So in summary if you provided a constructor for CRAMFileReader along these lines I should be able 
    to do the rest, or at least take a first crack at it

    public CRAMFileReader(final Seekable inputStream, final SeekableStream indexInputStream,
                          final ReferenceSource referenceSource, final ValidationStringency validationStringency)

droazen:
For @cmnbroad. Could you please prioritize this ticket over any other non-emergency fixes in htsjdk? Thanks!

droazen:
This ticket also includes getting fixes in for https://github.com/samtools/htsjdk/issues/331 and https://github.com/samtools/htsjdk/issues/332. I believe that @jrobinso can supply the fixes for these, if we can supply unit tests for the fixes.

yfarjoun:
I think that the PR's were closed due to lack of tests...https://github.com/samtools/htsjdk/pull/335

jrobinso:
Yes that's correct,  but the fixes are there in the closed request.   As 
you can see its fairly minor.

droazen:
We've agreed to implement the tests necessary to get the patches merged.

yfarjoun:
yes. I just wanted to point out that there are fixes written that just need
tests.

jrobinso:
I just pulled the latest htsjdk and had a look.  I think the following constructor can be used by IGV without modification:  

    public CRAMFileReader(final InputStream inputStream, final SeekableStream indexInputStream,
                          final ReferenceSource referenceSource, final ValidationStringency validationStringency) throws IOException {
[...]

Issue #331, opened 2015-09-23
bug:  hasBlockCompressedExtension fails for URLs with parameters


jrobinso:
The method "AbstractFeatureReader.hasBlockCompressedExtension" does not take into account url parameters and thus fails for resources such as the following because it doesn't end in ".gz"

https://www.googleapis.com/download/storage/v1/b/deflaux-public-test/o/NA12877.vcf.gz?alt=media

jrobinso:
The following version of the method fixes the problem

    public static boolean hasBlockCompressedExtension(final String path) {

        final String fileName = (path.startsWith("http://") || path.startsWith("https://")) ?
                path.split("\\?", 2)[0] :
                path;

        for (final String extension : BLOCK_COMPRESSED_EXTENSIONS) {
            if (fileName.toLowerCase().endsWith(extension))
                return true;
        }
        return false;
    }

cmnbroad:
I made a somewhat different fix than the one proposed above (I added an overload that accepts a URI) but let me know if for some reason thats not adequate.

Issue #332, opened 2015-09-24
Test for gzip extension fails for URLs with parameters


jrobinso:
Thi similar to bug #331,  but in a different location:  class WFIterator in TribbleIndexedFeatureReader.  There is a test path.endsWith(".gz")  that will fail for URLs such as  
https://www.googleapis.com/storage/v1/b/deflaux-public-test/o/bed_chr_17.bed.gz?alt=media

cmnbroad:
https://github.com/samtools/htsjdk/pull/599 should fix this, but there are other places where TribbleIndexedFeatureReader and ParseUtils make [assumptions](https://github.com/samtools/htsjdk/blob/c847e260a351dd4ded3caf3c44c6ec7d0c037675/src/java/htsjdk/tribble/TribbleIndexedFeatureReader.java#L141) about the path not being a URL.

magicDGS:
Now the `WFIterator` uses `hasBlockCompressedExtension`, but I think that for solve completely this issue all the usages in htsjdk of `hasBlockCompressedExtension` should use the `URI` version. If so, let me know if I can help with a PR modifying all the usages...

tfenne:
Fixed by #1132. TribbleIndexedFeatureReader now uses IOUtil.hasBlockCompressedExtension, which strips the http query string before checking the extension, so URLs like ...bed.gz?alt=media are handled. Closing as resolved.

PR #599, opened 2016-05-09
Factor out CRAMReferenceSource as an interface and fix two URI bugs.

cmnbroad:
### Description

Fixes https://github.com/samtools/htsjdk/issues/597. Factors out CRAMReferenceSource as an interface so that alternate implementations can be provided by consumers such as IGV. (Now that we have this as an interface, there is additional factoring that can be done i.e., separate the ENA CRAM service ReferenceSource implementation from the the file-based ReferenceSource implementation but we'll save those for a separate PR).

Also fixes https://github.com/samtools/htsjdk/issues/331 and https://github.com/samtools/htsjdk/issues/332.

For  https://github.com/samtools/htsjdk/issues/331, this PR adds a method overload that accepts a URI (though this is really just syntactic sugar). This also fixes the specific issue described in https://github.com/samtools/htsjdk/issues/332, though TribbleIndexedFeatureReader may have other issues using parameterized URLs (details in that ticket - trying to keep this minimal for now in the interest of getting IGV unblocked).

### Checklist

Note: the changes to the CRAM ReferenceSource are not backward compatible, though for most consumers they should be mostly source compatible).

- [x] Code compiles correctly
- [x] All tests passing


droazen:
@jrobinso Could you have a look at this, and let us know if everything you requested is here?

jrobinso:
This looks complete for now.   When its merged I will do a pull and start the IGV part.

src/java/htsjdk/samtools/CRAMIterator.java
```diff
@@ -94,10 +93,10 @@ public CRAMIterator(final InputStream inputStream, final ReferenceSource referen
         parser = new ContainerParser(cramHeader.getSamFileHeader());
     }
 
-    public CRAMIterator(final SeekableStream seekableStream, final ReferenceSource referenceSource, final long[] coordinates, final ValidationStringency validationStringency)
+    public CRAMIterator(final SeekableStream seekableStream, final CRAMReferenceSource referenceSource, final long[] coordinates, final ValidationStringency validationStringency)
             throws IOException {
         if (null == referenceSource) {
-            throw new CRAMException("A reference source is required for CRAM files");
+        //    throw new CRAMException("A reference source is required for CRAM files");
```
droazen: Why did you comment out this line? It seems to me that we still want to throw if `referenceSource` is null.

src/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -78,7 +79,7 @@
     abstract public SamReaderFactory referenceSequence(File referenceSequence);
 
     /** Sets the specified reference sequence * */
-    abstract public SamReaderFactory referenceSource(ReferenceSource referenceSequence);
+    abstract public SamReaderFactory referenceSource(CRAMReferenceSource referenceSequence);
```
droazen: It seems awkward that this abstract class `SamReaderFactory` should have CRAM-specific parameters/return values like this -- though I suppose that was always the case, just disguised under the former, more generic name.

src/java/htsjdk/samtools/cram/ref/CRAMReferenceSource.java
```diff
@@ -0,0 +1,21 @@
...
+
+    /**
+     * getReferenceBases
+     * @param sequenceRecord the SAMSequenceRecord identifying the reference
+     *                       being requested
+     * @param tryNameVariants if true, attempt to match the requested sequence name
+     *                        against the reference by using common name variations,
+     *                        such as adding or removing a leading "chr" prefix
+     *                        from the requested name. if false, use exact match
+     * @return the bases representing the requested sequence
```
droazen: Document method behavior when the reference sequence can't be found.

src/java/htsjdk/samtools/cram/ref/ReferenceSource.java
```diff
@@ -40,14 +40,14 @@
 import java.util.Map;
 import java.util.regex.Pattern;
 
-public class ReferenceSource {
+public class ReferenceSource implements CRAMReferenceSource {
```
droazen: Argh...in a just world `CRAMReferenceSource` would implement `ReferenceSource`, not vice versa. Can we rename `ReferenceSource` to something like `CRAMReferenceSourceImpl` to avoid confusion?

src/java/htsjdk/tribble/TribbleIndexedFeatureReader.java
```diff
@@ -271,6 +273,27 @@ private void readHeader() throws IOException {
         return new WFIterator();
     }
 
+    //Visible for testing
+    static boolean isGZIPPath(final String path) {
```
droazen: Should this method be moved to a utils class, instead of being buried inside here?

src/tests/java/htsjdk/tribble/TribbleIndexFeatureReaderTest.java
```diff
@@ -0,0 +1,85 @@
...
+        return new Object[][]{
+                {TestUtils.DATA_DIR + "test.vcf", 5},
+                {TestUtils.DATA_DIR + "test.vcf.gz", 5}
+        };
+    }
+
+    @Test(dataProvider = "featureFileStrings")
+    public void testIndexedGZIPVCF(final String testPath, final int expectedCount) throws IOException {
+        final VCFCodec codec = new VCFCodec();
+        final TribbleIndexedFeatureReader<VariantContext, LineIterator> featureReaderHttp =
```
droazen: How is this an "http" feature reader? It's reading from local files, isn't it?

src/tests/java/htsjdk/tribble/TribbleIndexFeatureReaderTest.java
```diff
@@ -0,0 +1,85 @@
...
+                new TribbleIndexedFeatureReader(testPath, codec, false);
```
droazen: Close reader via try-with-resources

src/tests/java/htsjdk/tribble/TribbleIndexFeatureReaderTest.java
```diff
@@ -0,0 +1,85 @@
...
+        };
+    }
+
+    @Test(enabled = true, dataProvider = "blockCompressedExtensionExtensionURIStrings")
+    public void testGZExtension(final String testString, final boolean expected) throws URISyntaxException {
+        Assert.assertEquals(TribbleIndexedFeatureReader.isGZIPPath(testString), expected);
+    }
+
+    @DataProvider(name = "featureFileStrings")
+    public Object[][] createGZExtensionStringss() {
```
droazen: `Stringss` -> `Strings`

droazen:
Review complete, back to @cmnbroad, merge after addressing comments.

src/java/htsjdk/samtools/CRAMIterator.java
cmnbroad: Hmm. Done.

src/java/htsjdk/samtools/cram/ref/CRAMReferenceSource.java
cmnbroad: Documented to return null, since that is what the current sole implementation does.

src/java/htsjdk/samtools/cram/ref/ReferenceSource.java
cmnbroad: Agreed. we'll do this in a separate PR.

src/java/htsjdk/tribble/TribbleIndexedFeatureReader.java
cmnbroad: There are similar functions in IOUtil, but they accept another extension for gzip, and its unclear to me what to what extent these should overlap or not. So I'm leaving this here as a narrow extension to the local test that was previously being done.

PR #600, opened 2016-05-09
Proposed addition of a backward compatibility checkbox to the PR template.

cmnbroad:
### Description

Propose adding a checkbox for backward compatibility to the PR template.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,3 +9,5 @@ Explain the **motivation** for making this change. What existing problem does th
 - [ ] New tests covering changes and new functionality
 - [ ] All tests passing
 - [ ] Extended the README / documentation, if necessary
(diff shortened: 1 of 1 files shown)


lbergelson:
@cmnbroad Maybe we should add an automated check as well using something like [japicmp](https://github.com/siom79/japicmp).

akiezun:
agreed, manually determining if two versions are binary compatible is definitely non trivial: https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html

cmnbroad:
True that, but I think 99% of the time the coder knows they're making an incompatible change when they do it. I just wanted something to alert the reviewers so we can set the version tag on the PR and make the breaking ones easy to find when we have to go hunt them down.

droazen:
I think this is fine as long as it's understood that this is a flag set by a human to alert project maintainers of possible breakage, rather than a comprehensive means of determining which commits have broken compatibility. @cmnbroad merge when ready

PR #601, opened 2016-05-10
adding a useAsynchronousIO flag to SamReaderFactory

akiezun:
### Description

To help with implementing asynchronous BAM decompression (see https://github.com/samtools/htsjdk/pull/576), I added a flag to SamReaderFactory. Previously the synchronous/asynchronous status was always taken from `Defaults.USE_ASYNC_IO_FOR_SAMTOOLS` which did not provide enough control to clients. In particular, in GATK we do want asynchronous BAM decompression when reading from regular file system but not from HDFS.

This functionality setting on the factory already exists for `SAMFileWriterFactory` and `FastqWriterFactory`. 

This PR is a bit unusual in that it does not actually add asynchronous reading. It only enables the flag to be set on the factory rather than be always taken from `Defaults`. The actual implementation will be in https://github.com/samtools/htsjdk/pull/576. This PR enables clients like Hadoop-BAM to selectively pre-switch to synchronous IO to future-proof them.

@droazen please review. @tomwhite if this goes in, will you be able to switch to sync IO using this API?

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/java/htsjdk/samtools/BAMFileReader.java
+++ b/src/java/htsjdk/samtools/BAMFileReader.java
@@ -67,6 +67,10 @@ class BAMFileReader extends SamReader.ReaderImplementation {
     // If true, all SAMRecords are fully decoded as they are read.
     private boolean eagerDecode;
 
+    // If true, the BAMFileReader will use asynchronous IO.
+    // Note: this field currently has no effect (is not hooked up anywhere), but will be in the future. See https://github.com/samtools/htsjdk/pull/576
+    private final boolean useAsynchronousIO;
+
     // For error-checking.
     private ValidationStringency mValidationStringency;
 
@@ -97,11 +101,13 @@ class BAMFileReader extends SamReader.ReaderImplementation {
     BAMFileReader(final InputStream stream,
                   final File indexFile,
                   final boolean eagerDecode,
+                  final boolean useAsynchronousIO,
                   final ValidationStringency validationStringency,
                   final SAMRecordFactory factory)
         throws IOException {
         mIndexFile = indexFile;
         mIsSeekable = false;
+        this.useAsynchronousIO = useAsynchronousIO;
         mCompressedInputStream = new BlockCompressedInputStream(stream);
         mStream = new BinaryCodec(new DataInputStream(mCompressedInputStream));
         this.eagerDecode = eagerDecode;
@@ -119,10 +125,11 @@ class BAMFileReader extends SamReader.ReaderImplementation {
     BAMFileReader(final File file,
                   final File indexFile,
                   final boolean eagerDecode,
+                  final boolean useAsynchronousIO,
                   final ValidationStringency validationStringency,
                   final SAMRecordFactory factory)
         throws IOException {
-        this(new BlockCompressedInputStream(file), indexFile!=null ? indexFile : SamFiles.findIndex(file), eagerDecode, file.getAbsolutePath(), validationStringency, factory);
+        this(new BlockCompressedInputStream(file), indexFile!=null ? indexFile : SamFiles.findIndex(file), eagerDecode, useAsynchronousIO, file.getAbsolutePath(), validationStringency, factory);
         if (mIndexFile != null && mIndexFile.lastModified() < file.lastModified()) {
             System.err.println("WARNING: BAM index file " + mIndexFile.getAbsolutePath() +
                     " is older than BAM " + file.getAbsolutePath());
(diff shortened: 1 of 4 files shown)


src/java/htsjdk/samtools/BAMFileReader.java
```diff
@@ -67,6 +67,9 @@
     // If true, all SAMRecords are fully decoded as they are read.
     private boolean eagerDecode;
 
+    // If true, the BAMFileReader will use asynchronous IO.
+    private final boolean useAsynchronousIO;
```
droazen: Add a note here that this field currently has no effect (is not hooked up anywhere), but will be in the future.

droazen: Also, should this default to `Defaults.USE_ASYNC_IO_FOR_SAMTOOLS`?

src/java/htsjdk/samtools/SAMFileReader.java
```diff
@@ -203,6 +204,13 @@ public void close() {
     }
 
     /**
+     * If true, this reader will use asynchronous IO.
+     */
+    public void setUseAsyncIO(final boolean useAsyncIO) {
+        this.useAsyncIO = useAsyncIO;
```
droazen: With `BAMFileReader` you added a constructor argument for this setting. Here you added a setter method instead. Why the different approaches? Can you use the same interface in both classes?

src/java/htsjdk/samtools/BAMFileReader.java
```diff
@@ -97,11 +100,13 @@
     BAMFileReader(final InputStream stream,
                   final File indexFile,
                   final boolean eagerDecode,
+                  final boolean useAsynchronousIO,
```
droazen: Should you add `setUseAsyncIO()` to `SamReader.ReaderImplementation`, which already has methods like `enableIndexCaching()`, instead of adding a new constructor argument?

src/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -89,6 +89,11 @@
     /** Set this factory's {@link ValidationStringency} to the provided one, then returns itself. */
     abstract public SamReaderFactory validationStringency(final ValidationStringency validationStringency);
 
+    /** Set whether readers created by this factory will use asynchronous IO.
+     * If this methods is not called, this flag will default to the value of {@link Defaults#USE_ASYNC_IO_FOR_SAMTOOLS}.
+     * Returns the factory itself. */
+    abstract public SamReaderFactory setUseAsyncIo(final boolean asynchronousIO);
```
droazen: Does this work for all kinds of readers that could possibly be returned from this factory? Document.

src/java/htsjdk/samtools/SAMFileReader.java
droazen: Also, add note that this only affects bams

droazen:
First-pass review done, back to @akiezun

src/java/htsjdk/samtools/BAMFileReader.java
akiezun: Added comment.

The field is final so it can't default to anything - by design, it's the factory's responsibility to set the reader in a correct way and the reader should not reach out to the defaults.

src/java/htsjdk/samtools/SAMFileReader.java
akiezun: this class is deprecated - i don't care about it. We should delete it, see https://github.com/samtools/htsjdk/pull/602

The reason for a different interface is that this is a public class so changing the constructor is breaks API (unless I just add constructors but this class already has a ton of constructors). It's a dead class anyway, let's not worry about it.

src/java/htsjdk/samtools/BAMFileReader.java
akiezun: maybe but the problem is this: the Asynch decompression implemented in https://github.com/samtools/htsjdk/pull/576 creates an asynch stream **in the constructor** of `BAMFileReader` so calling a setter is too late.

src/java/htsjdk/samtools/SamReaderFactory.java
akiezun: added a simple note saying that it may not apply to all readers

akiezun:
added 2 small comments for the review. back to @droazen

droazen:
:+1: back to @akiezun, squash/merge when ready

PR #602, opened 2016-05-10
deleted 2 deprecated classes: SAMFileReader SAMFileTruncatedReader

akiezun:
### Description

These 2 classes SAMFileReader and SAMFileTruncatedReader have been deprecated since 2014 and should finally be deleted to reduce maintenance burden.

Note: this change is not backwards compatible. Picard uses `SAMFileReader` in 1 place as far as I can tell (FingerprintChecker) and will be improved by a trivial change to use the factory.


droazen:
You made changes to `SAMFileReader` in https://github.com/samtools/htsjdk/pull/601 -- how do you want to proceed here?

akiezun:
i want to just delete these 2 classes. i'll fix #601 accordingly

droazen:
@akiezun The GATK3 codebase still uses `SAMFileReader` in quite a few places -- should this deletion perhaps wait until the GATK3 codebase is officially retired? Back to you for a response.

akiezun:
I think we should kill it and deal with the loss in GATK3. It should be a trivial adjustment (108 occurrences but they are all very easily changed to use the factory. I would not want to keep a bogus class in htsjdk because a client like gatk3 is dragging their feet deleting the uses)

akiezun:
back to @droazen

droazen:
Let's give the htsjdk community a few more days to comment before pulling the plug on these classes. @nh13 @tfenne @yfarjoun do any of you object to the removal of these deprecated classes?

If we do merge this, please create a ticket in the GATK3 issue tracker with instructions on dealing with the breakage the next time they rev their htsjdk dependency, and assign to @vdauwera so that she's aware this will be an issue.

Back to @akiezun, merge by the end of the week if no one else objects.

vdauwera:
Can you document in a paragraph or so what a foot-dragging client should do to get with the times? Would help enormously to have a clear explanation to refer to (what to replace these classes with?), not just for GATK3 but also for any others in the community who are willing to make the changes but don't know exactly what that entails.

nh13:
@droazen ✂️ 👍:
Cut it out

jrobinso:
IGV also uses SAMFileReader.    Its simple, the factory looks complicated, is there a migration guide?  It would be helpful to know what to do to get the proper reader for all combinations of http / file  & sam / bam / cram, with and without indexes.

droazen:
@jrobinso Basic usage of the factory is:

    final SamReader reader = SamReaderFactory.makeDefault().open(input);

where `input` is a file or `SamInputResource`. You can also configure the returned reader using methods from the factory. Eg., for cram you typically need to set a reference:

    final SamReader reader = SamReaderFactory.makeDefault().referenceSequence(reference).open(input);

The migration happened several years ago (2014), so I'm not sure if a migration guide was written at that time.

akiezun:
@vdauwera see migration guide in @droazen's comment up

vdauwera:
@ronlevine Do you feel like when the time comes you would be able to migrate GATK3 from SAMFileReader to the SamReader/SamReaderFactory paradigm based on @droazen's description above?

ronlevine:
@vdauwera It looks pretty straightforward. It's using standard design patterns.

vdauwera:
Cool, thanks @ronlevine. Then I'm OK with this. I would just ask that this PR get held off from merging until *after* the release that we'll cut once Intel's GKL pull request gets merged, which we want in the imminent release of GATK 3.6, so we don't have to depend on the SamWhatever migration to be complete before releasing the new GATK.

Issue #605, opened 2016-05-17
Gradle: Travis-Ci failed with "The command "eval ./gradlew assemble" failed. Retrying, 2 of 3."


lindenb:
### Subject of the issue

I'm very new to gradle. When Travis-CI is executed, it fails at the beginning  https://travis-ci.org/lindenb/htsjdk/builds/130833473 . The error seems unrelated to the changes I made.

```
FAILURE: Build failed with an exception.

* Where:

Build file '/home/travis/build/lindenb/htsjdk/build.gradle' line: 35

* What went wrong:

A problem occurred evaluating root project 'htsjdk'.

> Cannot get property 'commitDistance' on null object

* Try:

Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 30.932 secs

The command "eval ./gradlew assemble" failed. Retrying, 2 of 3.

FAILURE: Build failed with an exception.

* Where:

Build file '/home/travis/build/lindenb/htsjdk/build.gradle' line: 35

* What went wrong:

A problem occurred evaluating root project 'htsjdk'.

> Cannot get property 'commitDistance' on null object
```

### Your environment
* version of htsjdk https://github.com/lindenb/htsjdk/commit/1d052a3e32db3e74b1bced00c31d726b39b384ee + commit
* version of java Travis/java 8
* which OS (Travis)

### Steps to reproduce

 https://travis-ci.org/lindenb/htsjdk/builds/130833473

### Expected behaviour
Compilation should be OK

### Actual behaviour

> Travis-CI error: Cannot get property 'commitDistance' on null object

https://travis-ci.org/lindenb/htsjdk/builds/130833473

lbergelson:
Thanks for reporting.  This is almost certainly unrelated to changes you've made.  I'm looking into it.

lbergelson:
@lindenb This seems to be a bug in the version plugin we're using.  If a repo has no tag it's returning a null.  I suspect it can be worked around by adding a tag somewhere in your branche's history.  I'm going to add a patch to work around this as well as reporting it to the plugin author since it looks like a trivial fix and seems like a bug.

lindenb:
fixed with https://github.com/samtools/htsjdk/pull/606

PR #606, opened 2016-05-17
adding null check when determining isRelease

lbergelson:
### Description
This should fix a gradle issue that can manifest in forks of htsjdk.  Fix for #605 

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

adding an explicit check for null when determining isRelease
this happens in the case of a fork of the repo with no tags

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,8 +31,7 @@ dependencies {
 sourceCompatibility = 1.8
 targetCompatibility = 1.8
 
-
-final isRelease = versionDetails().commitDistance == 0
+final isRelease = Boolean.getBoolean("release")
 version = isRelease ? gitVersion() : gitVersion() + "-SNAPSHOT"
 
 logger.info("build for version:" + version)


lbergelson:
@lindenb Could check if this resolves your issue?  It works for me locally but it's always possible something weird happens in travis.

lbergelson:
Hmn. I think there might be additional problems with untagged repos that would cause them to always consider every build "a release".  This will make it demand signing keys if trying to do an installation.  I'm going to just change this behavior to only consider it a release if it's explicitly told that it is one.  That will avoid this whole can of worms.

lbergelson:
Updated this to get rid of any automatic determination if something is a proper release or not.  This is better anyway.

lindenb:
@lbergelson it's fixed: thanks 

https://travis-ci.org/lindenb/htsjdk/builds/130882667

lbergelson:
Thanks for the update.  Sorry about that.  Didn't think to test it on a fork.

PR #608, opened 2016-05-18
Cleanup Compression Utilities

gdgib-bina:
### Description

This PR cleans up some issues with the compression libraries, in particular removing native code of uncertain utility and origin.  Testing is currently in progress (thanks to some merge conflicts) and I'll update the PR checklist below when it's complete.  Please speak with @tfenne for a longer explanation.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


yfarjoun:
thanks @gdgib-bina I think you probably want to disable IntelDeflaterTest until we have a compliant solution...

gdgib-bina:
Tests should be disabled now.

droazen:
@akiezun Can you please comment on this PR? Does this PR have your blessing?

src/main/java/htsjdk/samtools/util/zip/DeflaterFactory.java
```diff
@@ -43,36 +43,12 @@
  */
 public class DeflaterFactory {
 
-    private static Constructor<IntelDeflater> intelDeflaterConstructor;
```
tfenne: Javadoc for this class needs updating to reflect the changes that have been made.  Maybe just change it to read something like: `Constructs deflaters as needed by the library. Simply constructs a default JDK Deflater but provides a central point for changing the behavior in future.`

src/main/java/htsjdk/samtools/util/zip/DeflaterFactory.java
```diff
@@ -43,36 +43,12 @@
...
-            }
-        } else {
-            return new Deflater(compressionLevel, nowrap);
-        }
+        return new Deflater(compressionLevel, nowrap);
     }
 
     public static boolean usingIntelDeflater() {
-        return intelDeflaterConstructor != null;
+        return false;
```
tfenne: Would suggest just removing this method and removing the log statements in the ~2 places that call it.

tfenne:
Thanks @gdgib-bina .  Mostly looks good to me.  I did a quick search of the codebase and I think there are also references to/configuration for Intel deflator in `Defaults.java` that should be cleaned up.  That's a quick change though and then :+1: from me.

akiezun:
Hi all, we will have a fix for the underlying problem probably later this week. I suggest we wait with the merge until we can replace this old code with new code to minimize damage downstream.

gdgib-bina:
Based on @akiezun's comment, I'll hold off any further work on this for a day or two at least unless someone tells me there's a better plan.

droazen:
Superseded by https://github.com/samtools/htsjdk/pull/612 -- closing.

PR #610, opened 2016-05-21
badge for the latest release on maven central

akiezun:
### Description

Adds a badge to show the latest release on Maven Central

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,8 @@
-Status of master branch build: [![Coverage Status](https://coveralls.io/repos/github/samtools/htsjdk/badge.svg?branch=master)](https://coveralls.io/github/samtools/htsjdk?branch=master)
+[![Coverage Status](https://coveralls.io/repos/github/samtools/htsjdk/badge.svg?branch=master)](https://coveralls.io/github/samtools/htsjdk?branch=master)
 [![Build Status](https://travis-ci.org/samtools/htsjdk.svg?branch=master)](https://travis-ci.org/samtools/htsjdk)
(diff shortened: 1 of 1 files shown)


nh13:
Why not also get rid of "Status of master branch build: " at the front? It is no longer valid.

While you're at it, could you add the language and license badge?
```
[![License](http://img.shields.io/badge/license-MIT-blue.svg)](https://github.com/samtools/htsjdk)
[![Language](http://img.shields.io/badge/language-java-brightgreen.svg)](/url/to/java)
```
If you like them, make sure to update `/url/to/java`.

droazen:
@nh13 Can you explain what you mean by the "status of master branch build" badge being no longer valid? It links to https://travis-ci.org/samtools/htsjdk, which shows the results of the direct unit tests on htsjdk/master (something not displayed anywhere else, afaik).

nh13:
There is now more than one badge and so while the unit test badges is related, the latest maven release is not related to the status of the master branch. If the other badges are also added I think it makes even less sense. 

Also, the trend I have seen on other projects is to have the badges as a top banner in the README. That would be my preference but of course it is a matter of stylistic choice and what others have seen.

akiezun:
What's the value of the language badge?

nh13:
Tells a developer quickly the primary language of this project. It could be inferred from the name, but I have it generally useful when quickly browsing projects for the first time. For example, I am looking for a scala library for X, and the badge says JavaScript. Your call since it was only a suggestion.

vdauwera:
I like the idea of the language badge, fwiw

PR #612, opened 2016-05-24
Switch to using the Intel GKL (Genomics Kernel Library) to pull down the Intel deflater, and remove the checked-in copy

droazen:


droazen:
@akiezun please review

build.gradle
```diff
@@ -16,6 +16,10 @@ plugins {
 
 repositories {
     mavenCentral()
+
+    maven {
+        url "https://artifactory.broadinstitute.org/artifactory/libs-snapshot/" // for intel deflater snapshots
```
akiezun: add a comment that it's temporary until the GKL moves to maven

src/main/java/htsjdk/samtools/util/zip/DeflaterFactory.java
```diff
@@ -24,11 +24,11 @@
 package htsjdk.samtools.util.zip;
 
 import htsjdk.samtools.Defaults;
-import htsjdk.samtools.SAMException;
 
-import java.lang.reflect.Constructor;
 import java.util.zip.Deflater;
 
+import com.intel.gkl.compression.IntelDeflater;
+
```
akiezun: Can you fix the outdated javadoc for this class?

build.gradle
```diff
@@ -65,12 +70,14 @@ jar {
     }
 }
 
-processResources {
-    into "lib/jni/", {
-        from ("lib/jni/libIntelDeflater.so")
-    }
+task extractIntelGKL(type: Copy) {
```
akiezun: i think you can remove this task. Just make sure to check if the so/dylib is still there in the jar

build.gradle
```diff
@@ -65,12 +70,14 @@ jar {
...
+    from {zipTree {configurations.compile.filter{it.name.startsWith('intel-gkl')}.singleFile}}
+    include "com/intel/gkl/native/**"
+    into "$compileJava.destinationDir"
 }
 
+compileJava.dependsOn(extractIntelGKL)
```
akiezun: delete this line

akiezun:
something is wrong. This command used to work, now it fails:
```
java -cp ./build/libs/htsjdk-2.3.0-13-g9345cc5.dirty-SNAPSHOT.jar htsjdk.samtools.example.PrintReadsExample ./src/test/resources/htsjdk/samtools/compressed.bam true
INFO	2016-05-24 16:28:04	PrintReadsExample	Start with args:[./src/test/resources/htsjdk/samtools/compressed.bam, true]
Exception in thread "main" java.lang.NoClassDefFoundError: com/intel/gkl/compression/IntelDeflater
	at htsjdk.samtools.example.PrintReadsExample.printConfigurationInfo(PrintReadsExample.java:99)
	at htsjdk.samtools.example.PrintReadsExample.main(PrintReadsExample.java:70)
Caused by: java.lang.ClassNotFoundException: com.intel.gkl.compression.IntelDeflater
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 2 more
```

akiezun:
back to @droazen - something is not working properly yet

build.gradle
nh13: Looking at the POM for this dependency (https://artifactory.broadinstitute.org/artifactory/libs-snapshot/com/intel/intel-gkl/0.0.1-SNAPSHOT/intel-gkl-0.0.1-20160524.180210-1.pom), I see it uses the `gatk-native-bindings` dependency.  Is  that something we want included in htsjdk?  It looks fairly light-weight, but looks like it is meant for native calls for the HMM in GATK's haplotype caller.

nh13:
@droazen was successfully able to build (outside of Broad) with this branch, but got the same error whe running @akiezun's test.

tfenne:
Following up on @nh13's comment about the POM/transitive dependencies for `intel-gkl`, it looks like this will add the following transitive dependencies to htsjdk:
* log4j
* commons-io
* testng

The last at least should be move to the `test` scope within `intel-gkl`.  For the other two, is there a way to specify that `htsjdk` doesn't require (or want) those transitive dependencies since it only relies on a subset of the `intel-gkl` library?

droazen:
@akiezun @nh13 If you're using the htsjdk jar directly to run tools (a rare thing to do!), then you need to build a monolithic jar using `gradle shadowJar` and use the jar that ends in `-all.jar` in `build/libs` to run `PrintReadsExample` (instead of an htsjdk-only jar, which is what you get with `gradle jar`)

src/main/java/htsjdk/samtools/util/zip/DeflaterFactory.java
```diff
@@ -43,36 +43,21 @@
...
-                intelDeflaterConstructor = clazz.getConstructor(Integer.TYPE, Boolean.TYPE);
-            }
-        } catch (ClassNotFoundException e) {
-            intelDeflaterConstructor = null;
-        } catch (NoSuchMethodException e) {
-            intelDeflaterConstructor = null;
-        } catch (UnsatisfiedLinkError e) {
-            intelDeflaterConstructor = null;
-        }
+        isUsingIntelDeflater = Defaults.TRY_USE_INTEL_DEFLATER && (new IntelDeflater().load());
```
tfenne: I would like to keep the reflective lookup and instantiation in place since this allows users of htsjdk to just exclude the intel-gkl project when hooking in the htsjdk dependency and it will all still work.  Given this change users have no choice but to depend on this library and even if they chose to avoid using the custom deflator.

droazen:
@tfenne I don't think we can avoid the first two transitive dependencies, as they are required by the classes in gkl that we use. We can fix the TestNG dependency if we modify the pom for the intel-gkl library.

gdgib-bina:
@austindoupnik can you comment on whether this PR addresses our concerns?

build.gradle
droazen: `gatk-native-bindings` is the set of interfaces that GATK uses/will use to interact with native code in general (not just the PairHMM). It is super-lightweight (just interfaces, and perhaps eventually a very small number of utilities). I do agree that it is a bit awkward to have htsjdk depend on a project with gatk in the name, even transitively -- we are considering switching to a more generic name.

src/main/java/htsjdk/samtools/util/zip/DeflaterFactory.java
droazen: Can you explain the use case for excluding this (tiny) dependency rather than just disabling its use at runtime? If we start using gkl code more extensively in the codebase the requirement to avoid a static dependency on the gkl would force us to write some rather gross code everywhere....

build.gradle
nh13: I would vote for excluding it from the dependency:
https://docs.gradle.org/current/userguide/dependency_management.html#sub:exclude_transitive_dependencies

droazen: The classes we use in `intel-gkl` implement interfaces from `gatk-native-bindings`, so I don't think we can exclude it.

austindoupnik:
This is great!

@gdgib-bina 
It certainly removes the immediate concern. Though I can't seem to find any details about GKL or it's license, only prebuilt artifacts on Broad's artificatory (and no `<license>` section in the pom). @akiezun indicated above that GKL should make it onto maven central at some point, maybe we will get more visibility then.

droazen:
@austindoupnik I will add a license section to the pom for `intel-gkl` when I fixed the TestNG dependency for @tfenne. `intel-gkl` is distributed under the MIT license.

austindoupnik:
@droazen fantastic, thanks!

src/main/java/htsjdk/samtools/util/zip/DeflaterFactory.java
tfenne: I'm personally of the opinion that as of Java 7/8+ the utility of the Intel deflator is questionable.  In the vast majority of my uses of HTSJDK I'm either writing at a decent compression level (where intel deflator doesn't really outperform modern zlib) or streaming between processes with compression level 0.

While intel-glk is currently small, it may grow, and it introduces some new transitive dependencies.  My preference in these cases is to (in my projects) exclude the transitive dependencies for features I don't use in order to keep my JARs as small as possible.

The code was already here to instantiate this reflectively, so that if the Intel code isn't there (or fails to load cleanly for any reason) things could continue to function.  I'm not suggesting this be done in all downstream projects, just in HTSJDK as it was until this commit.

droazen: Thanks for the explanation @tfenne -- I'll restore the reflection code. It may cause minor issues for downstream projects (like GATK) that use static dependency analysis via maven shade or gradle shadow to produce a slimmed-down jar for their binary distributions, since the intel classes will no longer show up as a compile-time dependency. But that sort of thing is inherently brittle anyway, and always requires special-casing to get the desired results.

The Intel deflater is optimized for compression level 1, as you probably are aware. If you're only using level 0 or max compression I can see how it would not be as useful -- though I do believe that @akiezun showed that it does outperform zlib even at the maximum compression level (@akiezun is this correct?)

akiezun: yes, see comments in https://github.com/samtools/htsjdk/pull/515

droazen:
I believe I've addressed all comments -- I've added a license to the `intel-gkl` pom, removed an unnecessary step from `build.gradle`, and updated `DeflaterFactory` in accordance with the comments above.

@tfenne Can you have a look at the updated `DeflaterFactory` and tell me whether it satisfies your needs? I did have to include an `import` statement for the `IntelDeflater` class, but since I think your goal is to avoid a runtime dependency rather than a compile-time dependency I think this should be fine.

@akiezun Can you have a look at the updated `DeflaterFactory` docs, and also test out the revised branch to make sure everything still works for you?

src/main/java/htsjdk/samtools/util/zip/DeflaterFactory.java
droazen: @tfenne Actually, could you clarify something for me? If the goal is to not fail at runtime if the `IntelDeflater` class is not present (but still require that class at compile time, as it was before), couldn't we just do a `new IntelDeflater()` and trap any exceptions, include `NoClassDefFoundError`, etc.? What does the extra reflection dancing actually gain us here?

droazen: Nevermind @tfenne, I see the problem now with `new IntelDeflater()` even inside a try-catch that catches `NoClassDefFoundError` -- the `NoClassDefFoundError` happens when `DeflaterFactory` is first loaded rather than at the point where `IntelDeflater` is instantiated. I guess we're stuck with the reflection after all.

tfenne:
Thanks @droazen .  I think the code in the static block is fine, but the static reference may cause some problems despite erasure.  I just tested and changing the declaration to:

```java
    private static Constructor<? extends Deflater> intelDeflaterConstructor;
```

works too.

src/main/java/htsjdk/samtools/util/zip/DeflaterFactory.java
tfenne: Exactly - it would cause a linking error during class loading, which is not something that can be recovered from, and then the DeflatorFactory would fail to load, and things would go badly.

droazen:
@tfenne Changed to `Constructor<? extends Deflater>` as requested

@akiezun Can you have a final look at this as well?

akiezun:
nope, does not work properly. IntelDeflater ignores the compression level and always uses 1.
To see it, take any file and compress it at level 1 and 5 with intel and jdk. Then look at md5s.

@gspowley FYI, the compression level setting seems to be not hooked up. Looking at decompiled code from IntelDeflator:
```
    public IntelDeflater(int level, boolean nowrap) {
    }

    public IntelDeflater() {
    }
```

droazen:
@akiezun As a temporary stopgap until Intel fixes this, I've modified the factory to only use the IntelDeflater if the requested compression level is 1.

akiezun:
hmm, i don't like it that the log still says "IntelDeflater" when I'm running on level5 and using JdkDeflater.

droazen:
@akiezun With the patch in place you can't know whether the Intel Deflater will be used until you call `makeDeflater()` with a specific `compressionLevel` value -- unless you want us to assume that the compression level will always be `Defaults.COMPRESSION_LEVEL`, or are willing to change signature of `usingIntelDeflater()` to accept a `compressionLevel` parameter. Suggestions?

akiezun:
So there's no way to tell if the loading of the inteldeflater happened and
was succesful? That seems strange. I'll look at the code later tonight

On Thursday, May 26, 2016, droazen <notifications@github.com> wrote:

> @akiezun <https://github.com/akiezun> With the patch in place you can't
> know whether the Intel Deflater will be used until you call makeDeflater()
> with a specific compressionLevel value -- unless you want us to assume
> that the compression level will always be Defaults.COMPRESSION_LEVEL, or
> are willing to change signature of usingIntelDeflater() to accept a
> compressionLevel parameter. Suggestions?
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/samtools/htsjdk/pull/612#issuecomment-221989113>
>


-- 
Sent from Gmail Mobile

droazen:
@akiezun With the current interface, you can certainly tell whether the IntelDeflater library was successfully loaded or not -- this is `DeflaterFactory.usingIntelDeflater()`. What you can't tell (with the new patch to only use IntelDeflater when compressionLevel == 1) is whether the IntelDeflater will actually be used until you actually call `makeDeflater()` and provide an actual value for `compressionLevel`.

akiezun:
you're right, i see the code now. Hmm, that's not so great for debugging, it is? How about logging the actual class of the deflater in BlockCompressedOutputStream after calls to `deflater = DeflaterFactory.makeDeflater(compressionLevel, true);`

tfenne:
@akiezun @droazen I understand we want to get this merged ASAP, but maybe we should wait until the new intel code is fixed to respect the compression level?

Alternatively, if it only really offers significant benefits at level 1, perhaps we could permanently set it to be used only for level 1 and change the logging messages to indicate `IntelDeflator loaded and available for Level 1 compression` or something similar?

akiezun:
@tfenne inteldeflater offers speedup at all levels, see https://github.com/samtools/htsjdk/pull/515 where i tested levels 1 and 5

tfenne:
@akiezun So maybe we should just wait until GKL is fixed to respect the compression level passed in?  And maybe add tests that a small BAM compressed with the intel deflator at each level 0-9 is _approximately_ the same size as when compressed through the JDK deflator?  I feel like there have been enough issues around the native compression code that it would be better to do this slowly and carefully than rush it in.

droazen:
@tfenne Under normal circumstances I'd agree that we should wait. Unfortunately from what I understand the fix in GKL might be several weeks away (it's not just a simple bug fix), and GATK 3.x needs this htsjdk patch now for its upcoming release. I'll try to make the log messages as explicit as possible (`IntelDeflator loaded and available for Level 1 compression` is a good suggestion), and I'll add debugging output to print the actual class of the deflater. We could also turn it off by default in the `Defaults` class to make it opt-in for now, if you guys agree that's the right thing to do? (the default behavior is already changing anyway due to the limitation to compression level 1).

akiezun:
sounds good for me. GATK4 selects the intel deflater explicitly anyway.

droazen:
Ok -- we'll have to be sure to communicate this change in default behavior to the relevant interested parties (eg., Picard, Broad production, etc.)

droazen:
@akiezun Can I get a final ok on this? `TRY_USE_INTEL_DEFLATER` now defaults to false, and I've added additional logging messages as requested

akiezun:
modulo failing tests on travis, this works for me. The javadoc in PrintReadsExample is outdated now but that's a triviality.

droazen:
@akiezun Had to re-segregate the IntelDeflater tests to allow only those tests to run with the non-default `-Dsamjdk.try_use_intel_deflater=true`. We should be good now, hopefully.

gdgib-bina:
Thank you all for this!  Much appreciated...

droazen:
@akiezun In our zeal to get this merged we seem to have overlooked the (obvious-in-retrospect) fact that we can't publish htsjdk to maven central with a dependency on an intel-gkl snapshot that is not in central. I am going to have to disable the `IntelDeflater` code completely in master for now, since we need to do an htsjdk release today.

akiezun:
understood. thanks.

PR #616, opened 2016-05-30
Add a size method to SamLocusIterator.LocusInfo.

nh13:
### Description

This is useful if we want to quickly know how many records overlap this
position but do not want to get all those records.

### Checklist

- [X] Code compiles correctly
- [X] All tests passing
- [ ] ~~~Extended the README / documentation, if necessary~~~
- [ ] ~~~Is not backward compatible (breaks binary or source compatibility)~~~

### DIFF
--- a/src/main/java/htsjdk/samtools/util/SamLocusIterator.java
+++ b/src/main/java/htsjdk/samtools/util/SamLocusIterator.java
@@ -116,6 +116,8 @@ public class SamLocusIterator implements Iterable<SamLocusIterator.LocusInfo>, C
         public List<RecordAndOffset> getInsertedInRecord() {
             return (insertedInRecord == null) ? Collections.emptyList() : Collections.unmodifiableList(insertedInRecord);
         }
+        /** @return the number of records overlapping the position, with deletions included if they are being tracked. */
+        public int size() { return this.recordAndOffsets.size() + ((deletedInRecord == null) ? 0 : deletedInRecord.size()); }
 
         /**
          * @return <code>true</code> if all the RecordAndOffset lists are empty;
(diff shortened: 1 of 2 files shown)


magicDGS:
@nh13, I like your idea but in your implementation size is only the number of records overlapping the position without deletions. But if tracked, these reads are also overlapping that position, and the `LocusInfo` object size is the sum of the two (the ones in the insertion should be in one of the two sets).

nh13:
@magicDGS look at that, I didn't notice that indel support got added in after my rebase, thank-you for that by the way.  Take a look again?

magicDGS:
Now for me is fine, @nh13. Probably a simple test with two reads (with/without deletion) for the method will be great.

src/main/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -116,6 +116,7 @@ public void addInserted(final SAMRecord read, int firstPosition) {
         public List<RecordAndOffset> getInsertedInRecord() {
             return (insertedInRecord == null) ? Collections.emptyList() : Collections.unmodifiableList(insertedInRecord);
         }
+        public int size() { return this.recordAndOffsets.size() + ((deletedInRecord == null) ? 0 : deletedInRecord.size()); }
```
yfarjoun: This needs a java-doc explaining what this size means. and a test.

nh13:
@yfarjoun I added the javadoc, and inserted asserts for `size()` in all the tests rather than writing my own.  Tests passing and a 👍  and I will merge.

yfarjoun:
👍

PR #617, opened 2016-05-30
Move a cram to test to its proper place.

nh13:
I think this test file is not in the right place.  Likely we will have some of these issues if we don't force folks to rebase onto master prior to merging, even if github says it is able to merge, due to changing the location of source and test files with the gradle move.

@droazen @akiezun @yfarjoun ?


cmnbroad:
Its kind of hard to see the paths from github, but I pulled it and it looks right to me.

jmarshall:
Also _testdata/htsjdk/samtools/io/bzip2-test.bz2_, which is referenced within _ExternalCompressionTest.java_.  Cf #533.

bpow:
Thanks for finding and fixing, @nh13.

In my defense, I had rebased against master several times before pull request that created this dangling test was merged, including right before it was merged, but rebase across the `git mv` commit that was part of the gradle transition didn't rebase the location of these files (consistent with `git mv` acting on a file rather than directory level).

nh13:
Not a problem. My move of this repo from SVN to GIT also left a few dangling sources.

Issue #611, opened 2016-05-24
SamLocusIterator with missing quality


dariober:
Hello- I think there is a bug in SamLocusIterator causing an ArrayIndexOutOfBoundsException when a SAMRecord has the QUAL string missing (i.e. set to "*"). The test case below should reproduce the issue.

I think it can be fixed by editing the method SamLocusIterator.accumulateSamRecord from:

```
if (dontCheckQualities || baseQualities[readOffset] >= minQuality) {
...
```
To:

```
if (baseQualities.length == 0 || dontCheckQualities || baseQualities[readOffset] >= minQuality) {
...
```

I.e. avoiding calling baseQualities[readOffset] on a zero-length array.

This is the test case causing problems:

```
import static org.junit.Assert.*;

import java.io.ByteArrayInputStream;

import org.junit.Test;

import htsjdk.samtools.SamInputResource;
import htsjdk.samtools.SamReader;
import htsjdk.samtools.SamReaderFactory;
import htsjdk.samtools.util.SamLocusIterator;

public class SamLocusIteratorTest {
    private SamReader createSamFileReader(final String samExample) {
        final ByteArrayInputStream inputStream = new ByteArrayInputStream(samExample.getBytes());
        return SamReaderFactory.makeDefault().open(SamInputResource.of(inputStream));
    }

    private SamLocusIterator createSamLocusIterator(final SamReader samReader) {
        final SamLocusIterator ret = new SamLocusIterator(samReader);
        ret.setEmitUncoveredLoci(false);
        return ret;
    }

    @Test
    public void testBasicIterator() {

        final String sqHeader = "@HD\tSO:coordinate\tVN:1.0\n@SQ\tSN:chrM\tAS:HG18\tLN:100000\n";
        final String seq1 = "ACCTACGTTCAATATTACAGGCGAACATACTTACTA";

        final String qual1 = "*"; // Quality string missing!

        final String s1 = "3851612\t16\tchrM\t165\t255\t36M\t*\t0\t0\t" + seq1 + "\t" + qual1 + "\n";
        final String exampleSam = sqHeader + s1 + s1;

        final SamReader samReader = createSamFileReader(exampleSam);
        final SamLocusIterator sli = createSamLocusIterator(samReader);


        // make sure we accumulated depth of 2 for each position
        int pos = 165;
        for (final SamLocusIterator.LocusInfo li : sli) {
            assertEquals(pos++, li.getPosition());
[...]

PR #626, opened 2016-05-31
Fix SamLocusIterator when missing base quality string is provided

brendanofallon:
### Description

This PR fixes issue #611 by checking for the missing base quality string "*" when using SamLocusIterator to iterate over SAMRecords. There's also a new test that targets this scenario.


droazen:
:+1: merging

PR #631, opened 2016-06-03
Histogram - made not extend TreeSet, make Bin static, add tests

### LINKED ISSUE #630: Histogram class is pathologically not serializable, despite being marked as Serializable
`Histogram` cannot be serialized via either Java or Kryo, despite being marked as `Serializable`, for two reasons:

-It has a non-static inner class (`Bin`). According to the Java Object Serialization Specification, serializing non-static inner classes has problems and should be avoided. Kryo does not appear to support this either (see discussion in, eg., https://groups.google.com/forum/#!topic/kryo-users/jgAU18bLyN0).

-It extends `TreeMap`, which has many transient fields and custom Java serialization routines. In extending this class, it appears that `Histogram` did not properly hook itself up with its superclass's custom serialization routines, causing major problems during both Java and Kryo serialization.

This is a blocking issue for porting htsjdk/Picard metrics to Spark. @yfarjoun @tfenne is there any appetite for a slight breaking change to `Histogram` to fix the serialization issues (likely just making the inner `Bin` class static and add a type parameter to it)?

akiezun:
### Description
Fixes #630. Fixes to Histogram - made it not extend TreeSet but use a TreeSet, made Bin static, add many tests, cleanup generics use.

Not backward compatible because Histogram does not extend TreeSet anymore and Bin is a static class. Some changes in clients look like this `Histogram<Integer>.Bin` -> `Histogram.Bin<Integer>`

### Checklist

- [X] Code compiles correctly
- [X] All tests passing
- [X] Is not backward compatible (breaks binary or source compatibility)

### DIFF
--- a/src/main/java/htsjdk/samtools/metrics/MetricsFile.java
+++ b/src/main/java/htsjdk/samtools/metrics/MetricsFile.java
@@ -49,16 +49,18 @@ import java.util.TreeSet;
  * @author Tim Fennell
  */
 public class MetricsFile<BEAN extends MetricBase, HKEY extends Comparable> implements Serializable {
+    private static final long serialVersionUID = 1L;
+
     public static final String MAJOR_HEADER_PREFIX = "## ";
     public static final String MINOR_HEADER_PREFIX = "# ";
     public static final String SEPARATOR = "\t";
     public static final String HISTO_HEADER = "## HISTOGRAM\t";
     public static final String METRIC_HEADER = "## METRICS CLASS\t";
 
-    private final Set<String> columnLabels = new HashSet<String>();
-    private final List<Header> headers = new ArrayList<Header>();
-    private final List<BEAN> metrics = new ArrayList<BEAN>();
-    private final List<Histogram<HKEY>> histograms = new ArrayList<Histogram<HKEY>>();
+    private final Set<String> columnLabels = new HashSet<>();
+    private final List<Header> headers = new ArrayList<>();
+    private final List<BEAN> metrics = new ArrayList<>();
+    private final List<Histogram<HKEY>> histograms = new ArrayList<>();
 
     /** Adds a header to the collection of metrics. */
     public void addHeader(Header h) { this.headers.add(h); }
@@ -267,7 +269,7 @@ public class MetricsFile<BEAN extends MetricBase, HKEY extends Comparable> imple
             out.append(key.toString());
 
             for (final Histogram<HKEY> histo : nonEmptyHistograms) {
-                final Histogram<HKEY>.Bin bin = histo.get(key);
+                final Histogram.Bin<HKEY> bin = histo.get(key);
                 final double value = (bin == null ? 0 : bin.getValue());
 
                 out.append(SEPARATOR);
@@ -537,7 +539,7 @@ public class MetricsFile<BEAN extends MetricBase, HKEY extends Comparable> imple
      * @return list of beans from the file.
      */
     public static <T extends MetricBase> List<T> readBeans(final File file) {
-        final MetricsFile<T, Comparable<?>> metricsFile = new MetricsFile<T, Comparable<?>>();
+        final MetricsFile<T, ?> metricsFile = new MetricsFile<>();
         final Reader in = IOUtil.openFileForBufferedReading(file);
         metricsFile.read(in);
         CloserUtil.close(in);
@@ -549,7 +551,7 @@ public class MetricsFile<BEAN extends MetricBase, HKEY extends Comparable> imple
      */
     public static List<Header> readHeaders(final File file) {
         try {
-            final MetricsFile<MetricBase, Comparable<?>> metricsFile = new MetricsFile<MetricBase, Comparable<?>>();
+            final MetricsFile<MetricBase, ?> metricsFile = new MetricsFile<>();
             metricsFile.read(new FileReader(file));
             return metricsFile.getHeaders();
         } catch (FileNotFoundException e) {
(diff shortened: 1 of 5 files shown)


akiezun:
TODO: needs a test that demonstrates that what used to be non serializable is now serializable. @cmnbroad do you have an example Histogram instance that fails to serialize?

src/main/java/htsjdk/samtools/util/Histogram.java
```diff
@@ -42,40 +35,46 @@
  *
  * @author Tim Fennell
  */
-public class Histogram<K extends Comparable> extends TreeMap<K, Bin> {
+public final class Histogram<K extends Comparable<K>> implements Serializable {
```
lbergelson: I don't know if you want to enforce that K is also serializable.  It seems like a reasonable thing to add to me.

tfenne: That would be a bigger breaking change - I'm sure I have a bunch of uses of this class where I don't care about serialization and am using non-serializable key types.

droazen: Recommend that we not cause any more breakage here than absolutely necessary to fix the original issue(s). We'll be using Kryo anyway to serialize this in GATK, which does not require strict Java serializability -- as long as we fix the non-static inner class and the issue related to inheriting from TreeMap we should be fine.

akiezun: I agree with @tfenne - you may want a histogram of things that are not serializable. Look at all java collections - none specifies keys/values to be serializable.

src/main/java/htsjdk/samtools/util/Histogram.java
```diff
@@ -153,12 +154,26 @@ public void increment(final K id, final double increment) {
...
+    }
+
+    @Override
+    public String toString() {
+        return map.toString();
+    }
+
+    @Override
+    public int hashCode() {
+        return java.util.Objects.hash(binLabel, valueLabel, map);
```
akiezun: in case anyone wonders - Objects needs to be fully qualified here because it clashes with htsjdk.samtools.util.Objects (which should be removed btw)

src/main/java/htsjdk/samtools/metrics/MetricsFile.java
```diff
@@ -48,17 +48,19 @@
  *
  * @author Tim Fennell
  */
-public class MetricsFile<BEAN extends MetricBase, HKEY extends Comparable> implements Serializable {
+public class MetricsFile<BEAN extends MetricBase, HKEY extends Comparable<HKEY>> implements Serializable {
+    private static final long serialVersionUID = 1L;
+
```
cmnbroad: I'm not aware of a case where we've needed to serialize a MetricsFile, but we should probably add a test to make sure it works..

src/test/java/htsjdk/samtools/util/HistogramTest.java
```diff
@@ -5,17 +5,18 @@
...
+import java.util.Comparator;
+import java.util.HashSet;
 
 import static java.lang.Math.abs;
+import static java.lang.StrictMath.pow;
 
-/**
- *
- */
 public class HistogramTest {
```
cmnbroad: As far as serialization tests go, I think any one of the test ones in this file (or [these](https://github.com/broadinstitute/gatk/blob/4864c3d5fe7ea78245ea03904b27a2ecbb95e5f5/src/test/java/org/broadinstitute/hellbender/tools/spark/sv/InsertSizeMetricsCollectorSparkUnitTest.java#L35)) would do, since I don't think any of them serialized properly before. As long as we get back a proper treemap and bin and value labels.

akiezun: there is a test though `testSerializeHistogram` that checks serialization and it passes on master. Can you give me an example histogram that fails to serialize on master?

src/main/java/htsjdk/samtools/metrics/MetricsFile.java
akiezun: it's tested in `MetricsFileTest.testWriteMetricsFile`

akiezun:
@cmnbroad @droazen I need a testcase to show that this fix works. Can you send me one historgram that fails to serialize on the master code?

akiezun:
assigning to @droazen to provide a testcase

src/test/java/htsjdk/samtools/util/HistogramTest.java
cmnbroad: Hm - I thought even java serialization was failing before - but it may have just been kryo. Ill dig one up.

droazen:
@cmnbroad Can you supply the test case Adam is requesting?

src/test/java/htsjdk/samtools/util/HistogramTest.java
cmnbroad: The java serialization issue I was seeing was due to a Histogram with a non-serializable anonymous comparator; it is only kryo serialization that fails always, so I don't have a case that fails without a kryo dependency. FWIW I did test this branch against my gatk branch that uses a custom kryo serializer to work around the kryo issues, and with these changes kryo is able to serialize Histograms without the custom serializer.

akiezun:
@droazen good to merge?

lbergelson:
@akiezun Did you make sure that all the TreeMap functions that picard uses are still available in the modified histogram or easily workaroundable?

cmnbroad:
The one I know for sure missing now is containsKey, but its easy to work around. Other than that all of the GATK metrics code (at least in my metrics branch) compiled with these changes. Picard may have others. The thing that required the most changes in GATK were due to the MetricsFile type parameter bounds changes.

akiezun:
add a containsKey and merge then? 
btw, how to compile picard with a custom htsjdk?

lbergelson:
@akiezun At the moment I think you can just cd into picard's htsjdk clone, checkout the htsjdk branch you want, and then build picard with it.  Going to change in the future though.

lbergelson:
If you just want to run the tests, you can launch a build from jenkins.  Log into jenkins, go to the picard-on-htsjdk-master job, select "build with parameters" and then fill in the htsjdk and picard branches you want to build together.

droazen:
I defer to @cmnbroad to give the final ok on this branch when it's ready

cmnbroad:
Its probably worth building with Picard to see if we're missing anything else. I would only expose things that are absolutely necessary though, i.e., to me containsKey seems like an unnecessary luxury since there is an easy workaround. Otherwise LGTM.

akiezun:
ugh, so the bound on the type parameter in MatricsFile (`HKEY extends Comparable<HKEY>`)  is a massive pain in picard because the generics are used very loosely.  I could just forget about it and revert it to `HKEY extends Comparable`, ugly as it is. WDYT?

cmnbroad:
I think the other changes are the important ones; that one is nice but optional (though in GATK the changes were pretty mechanical - mostly replacing "Comparable<?>" with "? extends Comparable<?>") I think. There were a bunch of them though.

PR #632, opened 2016-06-03
Add the ability to publish to local maven repository

jacarey:
### Description

Added maven publish plugin. This allows us to use publishToLocalMaven which in turn will allow use to use mavenLocal() in Picard to pick up snapshot builds from the local repository.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,6 +7,7 @@ buildscript {
 plugins {
     id "java"
     id 'maven'
+    id 'maven-publish'
     id 'signing'
     id 'jacoco'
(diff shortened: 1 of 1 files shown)


lbergelson:
@jacarey ./gradlew install should do this already.  We could switch to using the maven-publish plugin instead of the maven one, but we shouldn't do both.

jacarey:
Ah ok good to know. closing this PR.

lbergelson:
it should probably be added to the readme actually

Issue #462, opened 2016-02-04
Remove/Clarify Error: Could not find or load main class message


lbergelson:
An error message `Error: Could not find or load main class` has started appearing when starting htsjdk and software that includes htsjdk.  That message doesn't appear anywhere in htsjdk itself, and seems to have started happening around the time we introduced SRA support. 

This error message should be suppressed, or at the bare minimum be more descriptive.

lbergelson:
@a-nikitiuk @kwrodarmer Any idea if this is coming from SRA somehow?  It's the obvious candidate, but I can't find that message in the SRA code.  It's a worrying message that has been confusing people.  It's also possible it's unrelated to SRA.

a-nikitiuk:
This message comes from SRA when it tries to launch a new process to perform a version check of installed native libraries. That message should disappear after first execution of SRA, since it records version of a found/downloaded library to special properties file and does not launch version checking next time. That error should already be fixed in ngs-java-1.2.3.jar (currently HTSJDK uses ngs-java-1.2.2.jar).

Also, I am currently working a newer version of ngs-java.jar which supports an option to not download native libraries and has much better error reporting using different exception types and messages and that version (it is not released yet) should not have that issue either.

lbergelson:
Thanks @a-nikitiuk  for the information!

tfenne:
SRA support was fully removed from htsjdk in #1774. Since this warning originated entirely from the SRA/ngs-java native-library version check, it can no longer occur. Closing as obsolete.

Issue #414, opened 2015-12-22
Remove SRA related code and replace with a plugin API


tfenne:
The SRA related code that was recently accepted as part of #323 has caused too many issues, is tarnishing the reputation of HTSJDK and projects which depend upon it and is a big step in the wrong direction.  My problems with #323 are as follows:

1. It adds an extra dependency to HTSJDK.  For a core library this is to be frowned upon.  While I _think_ it is possible for someone using HTSJDK for non-SRA formatted data to run without the `ngs-java` jar present, there is no easy way to enforce this going forward.
2. Some of the things that `ngs-java` does are, IMHO, unacceptable, including a) searching a variety of system paths and classpaths for a native library and b) downloading a native library over a non-secure internet connection!!
3. Adding this code directly to HTSJDK sets the wrong precedent - are we really going to accept code and maintenance responsibility for any and every format?  I realize that NCBI has a lot of data in SRA format, but the format is basically unused outside of NCBI, so this has very limited general utility.

I can understand that users of various downstream tools (GATK, Picard, IGV, etc.) want to be able to use their own formats, and that HTSJDK is a convenient place to add that support, but the way this has been done for SRA is a big step in the wrong direction.  Instead I think we should discuss adding a plugin or service provide interface (SPI) mechanism to allow _runtime_ loading of alternative service providers to allow folks like the SRA to provide SRA support _outside_ of the core HTSJDK project.

In light of the problems users have already had, and the misbehavior of the `ngs-java` library, I would suggest that we excise the SRA changes from master while we work on a better solution.

nh13:
I would whole-heartedly agree, as the commit goes against the spirit of htsjdk, and generally has bad behavior as you mentioned above.  The error message every time it would try to load various libraries were also confusing.  I also looked at the source [`ngs-java`](https://github.com/ncbi/ngs/blob/master/ngs-java/ngs), which has no tests and adds in a lot of duplicate functionality, so it is not something I think htsjdk should rely upon or have access to.  Finally, given that it is not widely used outside of NCBI (could the original committer verify this?), then the benefit does not outweigh the debt that it incurs and the precedent it sets.

I support excising the SRA changes, including making `BAMFileSpan.java` private again.  I think how to integrate it is a separate discussion, but I like the ideas you propose.

Given the various holidays, I would recommend making a pull request that can be reviewed and merged after folks agree here.  I propose January 15th 2015 is the last day we comments for discussion if no further objections are made.

[...]

PR #634, opened 2016-06-03
Provided an option to disable auto-download of SRA native libraries

a-nikitiuk:
### Description

This PR implements optional SRA native libraries auto-download, as requested in #414
In case, when ngs-java fails to locate/load native libraries, we throw a subclass of java.lang.Error with details of what went wrong and a recommendation on how to fix the problem.

In ngs-java 1.2.4 we also improved version checking of native libraries before loading them and addressed issue #462 

We also fixed an issue with keeping references to SRA objects forever, which prevented them from being garbage collected and sometimes produced issues with reaching a limit of opened file handlers.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,7 +40,7 @@ dependencies {
     compile "org.xerial.snappy:snappy-java:1.0.3-rc3"
     compile "org.apache.commons:commons-compress:1.4.1"
     compile "org.tukaani:xz:1.5"
-    compile "gov.nih.nlm.ncbi:ngs-java:1.2.2"
+    compile "gov.nih.nlm.ncbi:ngs-java:1.2.4"
 
     testCompile "org.testng:testng:6.9.9"
 }
@@ -125,6 +125,8 @@ test {
 }
 
 task testSRA(type: Test){
+    jvmArgs '-Dsamjdk.sra_libraries_download=true'
+
     description "Run the SRA tests"
     useTestNG{
         includeGroups "sra"
--- a/build.gradle
+++ b/build.gradle
@@ -124,11 +124,11 @@ test {
     }
 }
 
-task testSRA(type: Test){
+task testSRA(type: Test) {
     jvmArgs '-Dsamjdk.sra_libraries_download=true'
 
     description "Run the SRA tests"
-    useTestNG{
+    useTestNG {
         includeGroups "sra"
     }
 }
--- a/src/main/java/htsjdk/samtools/Defaults.java
+++ b/src/main/java/htsjdk/samtools/Defaults.java
@@ -83,6 +83,10 @@ public class Defaults {
      */
     public static final String EBI_REFERENCE_SERVICE_URL_MASK;
 
+    /**
+     * Boolean describing whether downloading of SRA native libraries is allowed,
+     * in case such native libraries are not found locally
+     */
     public static final boolean SRA_LIBRARIES_DOWNLOAD;
 
 
--- a/src/main/java/htsjdk/samtools/Defaults.java
+++ b/src/main/java/htsjdk/samtools/Defaults.java
@@ -83,6 +83,8 @@ public class Defaults {
      */
     public static final String EBI_REFERENCE_SERVICE_URL_MASK;
 
+    public static final boolean SRA_LIBRARIES_DOWNLOAD;
+
 
     static {
         CREATE_INDEX = getBooleanProperty("create_index", false);
@@ -108,6 +110,7 @@ public class Defaults {
         EBI_REFERENCE_SERVICE_URL_MASK = "http://www.ebi.ac.uk/ena/cram/md5/%s";
         CUSTOM_READER_FACTORY = getStringProperty("custom_reader", "");
         SAM_FLAG_FIELD_FORMAT = SamFlagField.valueOf(getStringProperty("sam_flag_field_format", SamFlagField.DECIMAL.name()));
+        SRA_LIBRARIES_DOWNLOAD = getBooleanProperty("sra_libraries_download", false);
     }
 
     /**
--- a/src/main/java/htsjdk/samtools/SamReaderFactory.java
+++ b/src/main/java/htsjdk/samtools/SamReaderFactory.java
@@ -337,6 +337,10 @@ public abstract class SamReaderFactory {
         /** Attempts to detect whether the file is an SRA accessioned file. If SRA support is not available, returns false. */
         private boolean isSra(final File sourceFile) {
             try {
+                // if SRA fails to initialize (the most common reason is a failure to find/load native libraries),
+                // it will throw a subclass of java.lang.Error and here we only catch subclasses of java.lang.Exception
+                //
+                // Note: SRA initialization errors should not be ignored, but rather shown to user
                 return SRAAccession.isValid(sourceFile.getPath());
             } catch (final Exception e) {
                 return false;
--- a/src/main/java/htsjdk/samtools/sra/SRAAccession.java
+++ b/src/main/java/htsjdk/samtools/sra/SRAAccession.java
@@ -26,6 +26,8 @@
 
 package htsjdk.samtools.sra;
 
+import gov.nih.nlm.ncbi.ngs.error.LibraryLoadError;
+import htsjdk.samtools.Defaults;
 import htsjdk.samtools.util.Log;
 import gov.nih.nlm.ncbi.ngs.NGS;
 
@@ -46,12 +48,21 @@ import java.util.Arrays;
 public class SRAAccession implements Serializable {
     private static final Log log = Log.getInstance(SRAAccession.class);
 
-    private static Boolean isSupportedCached = null;
+    private static boolean noLibraryDownload;
+    private static boolean initTried = false;
+    private static ExceptionInInitializerError ngsInitError;
     private static String appVersionString = null;
     private final static String defaultAppVersionString = "[unknown software]";
     private final static String htsJdkVersionString = "HTSJDK-NGS";
 
     private String acc;
+
+    static {
+        noLibraryDownload = !Defaults.SRA_LIBRARIES_DOWNLOAD;
+        if (noLibraryDownload) {
+            System.setProperty("vdb.System.noLibraryDownload", "1");
+        }
+    }
 
     /**
      * Sets an app version string which will let SRA know which software uses it.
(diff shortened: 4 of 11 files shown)


droazen:
@yfarjoun Could you review this one?

build.gradle
```diff
@@ -143,6 +143,8 @@ task testIntelDeflater(type: Test){
 }
 
 task testSRA(type: Test){
+    jvmArgs '-Dsamjdk.sra_libraries_download=true'
```
yfarjoun: can you also test that without any jvmArguments there is no downloading?

build.gradle
```diff
@@ -143,6 +143,8 @@ task testIntelDeflater(type: Test){
...
+
     description "Run the SRA tests"
     useTestNG{
```
yfarjoun: space before { (also above)

src/main/java/htsjdk/samtools/Defaults.java
```diff
@@ -83,6 +83,8 @@
      */
     public static final String EBI_REFERENCE_SEVICE_URL_MASK;
 
+    public static final boolean SRA_LIBRARIES_DOWNLOAD;
```
yfarjoun: Please add a comment before this variable explaining what it's for.

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -320,6 +320,10 @@ public SamReader open(final SamInputResource resource) {
         /** Attempts to detect whether the file is an SRA accessioned file. If SRA support is not available, returns false. */
         private boolean isSra(final File sourceFile) {
             try {
+                // if SRA fails to initialize (the most common reason is a failure to find/load native libraries),
+                // it will throw a subclass of java.lang.Error and here we only catch subclasses of java.lang.Exception
+                //
+                // Note: SRA initialization errors should not be ignored, but rather shown to user
                 return SRAAccession.isValid(sourceFile.getPath());
             } catch (final Exception e) {
```
yfarjoun: Is there a test that this works? i.e. that the libraries are not available, download is set to false (by default) and the test attempts to open a SRA accession. the standard error should contain the error message that we would like to see.

src/main/java/htsjdk/samtools/sra/SRAAccession.java
```diff
@@ -46,13 +48,22 @@
 public class SRAAccession implements Serializable {
     private static final Log log = Log.getInstance(SRAAccession.class);
 
-    private static Boolean isSupportedCached = null;
+    private static boolean noLibraryDownload;
+    private static boolean initTried = false;
+    private static ExceptionInInitializerError ngsInitError;
```
yfarjoun: what is the value of saving this exception here? why not just 

```
return getInitializationError()
```
below?

src/main/java/htsjdk/samtools/sra/SRAAccession.java
```diff
@@ -107,7 +130,20 @@ public static boolean isValid(String acc) {
 
         if (!looksLikeSRA) return false;
 
-        return isSupported() && NGS.isValid(acc);
+        ExceptionInInitializerError initError = checkIfInitialized();
+        if (initError != null) {
+            if (noLibraryDownload && initError instanceof LibraryLoadError) {
+                throw new LinkageError(
+                        "Failed to load SRA native libraries and auto-download is disabled. " +
+                        "Please set property samjdk.sra_libraries_download=true to enable auto-download of native libraries",
```
yfarjoun: you might want to be explicit about the -D as you'll find that many users will not know how to do this unless you are very explicit...

src/main/java/htsjdk/samtools/sra/SRAAccession.java
```diff
@@ -107,7 +130,20 @@ public static boolean isValid(String acc) {
 
         if (!looksLikeSRA) return false;
 
-        return isSupported() && NGS.isValid(acc);
+        ExceptionInInitializerError initError = checkIfInitialized();
```
yfarjoun: final

src/main/java/htsjdk/samtools/sra/SRAAlignmentIterator.java
```diff
@@ -149,16 +153,17 @@ private void nextReference() {
         }
 
         try {
+            alignedIterator = null;
+            hasMoreAlignments = false;
+
             hasMoreReferences = referencesChunksIterator.hasNext();
```
yfarjoun: why is calling nextReference() when referencesChunksIterator.hasNext()==false not an error? I know that this isn't the code you are changing, but I'm curious to understand...

yfarjoun:
does this work on an empty accession? the initialization seems to depend on at-least one record passing through....is there a way to write a SRA "wrapper" to a file so that we can see some local SAM files being accessed using this library?

src/main/java/htsjdk/samtools/sra/SRAAlignmentIterator.java
```diff
@@ -191,4 +196,14 @@ private void nextReference() {
 
         return referencesChunks;
     }
+
+    @Override
+    public void close() {
+        if (lastRecord != null) {
+            lastRecord.detachFromIterator();
```
yfarjoun: why is this important? can you explain? from what I see detach... simply sets some variables to null, given that you are nullifying lastRecord in the next step, what good does it do?

src/main/java/htsjdk/samtools/sra/SRAIndexedSequenceFile.java
```diff
@@ -84,18 +78,15 @@ public ReferenceSequence getSubsequenceAt(String contig, long start, long stop)
         byte[] bases;
 
         try {
-            HashMap<Integer, Reference> localRefs = cachedReferences.get();
-            if (localRefs == null) {
-                localRefs = new HashMap<Integer, Reference>();
-                cachedReferences.set(localRefs);
+            Reference reference;
+            synchronized (this) {
+                if (cachedReference == null || !cachedReference.getCanonicalName().equals(contig)) {
```
yfarjoun: will different threads accessing this object and looking at different contigs cause it to thrash memory?

yfarjoun: can this be tested please?

yfarjoun:
back to you with some comments and questions @a-nikitiuk

build.gradle
a-nikitiuk: download of native libraries only happen when ngs-java fails to find native libraries locally. unfortunately, it depends on user environment and cannot be tested here

src/main/java/htsjdk/samtools/SamReaderFactory.java
a-nikitiuk: We have such tests for that in ngs-java, but those tests change ngs-java package local variables (and nobody else from outside should touch them) to simulate cases, when libraries are not found locally. Unfortunately, without doing so tests will depend on the environment and cannot be mocked

src/main/java/htsjdk/samtools/sra/SRAAlignmentIterator.java
a-nikitiuk: `lastRecord` object is an instance of SRALazyRecord class, which keeps a reference to NGS.AlignmentIterator for getting fields which were not loaded yet. That means that if somebody holds a reference to lastRecord outside of this class, he will also hold a reference to AlignmentIterator and that will prevent it from being garbage collected

src/main/java/htsjdk/samtools/sra/SRAAlignmentIterator.java
a-nikitiuk: This is just a way `nextReference()` is implemented. It tries to advance to the next reference and sets `hasMoreReferences` member variable indicating if it reached to the last reference in the list and whether future call to `nextReference()` is permitted or not.
That is somewhat similar to implementation of iterators in ngs-java

src/main/java/htsjdk/samtools/sra/SRAIndexedSequenceFile.java
a-nikitiuk: nothing bad will happen, just re-initialization of some data structures and re-open of file(s)

a-nikitiuk: added `SRAReferenceTest.testReferenceMt`

src/main/java/htsjdk/samtools/sra/SRAAccession.java
a-nikitiuk: updated the text

src/main/java/htsjdk/samtools/sra/SRAAccession.java
a-nikitiuk: added

src/main/java/htsjdk/samtools/sra/SRAAccession.java
a-nikitiuk: no value.. removed

src/main/java/htsjdk/samtools/Defaults.java
a-nikitiuk: added

build.gradle
a-nikitiuk: added, although that code was written by somebody else

a-nikitiuk:
@yfarjoun, we don't have empty accessions in SRA, however, we do have ones with only unaligned reads. Such accessions will not return a single record from `SRAAlignmentIterator`, which I suppose what you are interested in. I added an extra test to `SRATest.testCounts` which tests such accession.

As for an ability to feed SAM files to SRA, we don't have that.

@yfarjoun I answered all of your comments. Please review

yfarjoun:
Thanks @a-nikitiuk 👍 

There are conflicts now...so please rebase.

a-nikitiuk:
Done

vdauwera:
@a-nikitiuk Was this the PR you were asking us to wait for for the GATK SRA PR? Meaning we should take in the next release of htsjdk that includes this changeset before processing your PR against GATK?

a-nikitiuk:
Yes, that is the one. But I also wanted to add an extra change to GATK to re-enable SRA native libraries auto-download (if needed) by default. That will allow GATK users have a smooth experience and up-to-date libraries when passing SRA accessions as input

vdauwera:
Ah, OK. Let us know when that is ready then.

PR #443, opened 2016-01-20
add default methods to Locatable

lbergelson:
adding some default methods to `Locatable` as well as associated tests

moving these down from gatk since they're of general use

I think that most of them should be uncontroversial.  

The `getGA4GHStart/End` may want name changes.

PR #637, opened 2016-06-04
new algorithm for locus iteration and Picard's CollectWgsMetrics

ZLyanov:
### Description

We propose the following changes to Htsjdk to optimize CollectWgsMetrics work  in Picard tools:
SamLocusIterator traverses the input BAM file and creates for each reference locus position a list of aligned to this position bases. Each aligned read base is represented as one object, thus, if read has length of alignments blocks of L, then L objects will be created. All in all, for a BAM file holding N reads, O(N \* L) objects will be created.

Our idea is not to represent each aligned read base as a single object, but to create objects only for the start and the end of each aligned block of a read. In this way, number of created objects is reduced to O(N \* M) (N = number of reads, M = number of alignment blocks of a read). Usually M is much smaller (~ 1 – 5 per read), than length of aligned blocks.

In order to implement this approach we made some refactoring:
1.  Abstract class AbstractLcusIterator was extracted from SamLocusIterator to provide common interface and opportunity to override SAMRecord accumulation.
2.  Classes AbstractLocusInfo and AbstractRecordAndOffset also were extracted from inner classes of SamLocusIterator to allow their extending
3.  Class SamLocusIterator still represents the default algorithm for methods accumulateSamRecord() and accumulateIndels()
4.  Classes ReadEndsIterator and TypedRecordAndOffset implement our new approach. 

These changes are backward compatible: all classes, that use SamLocusIterator and its inner classes SamLocusIterator.RecordAndOffset and SamLocusIterator.LocusInfo will work the same without any adaptation. 
### Checklist
- [X] Code compiles correctly
- [X] All tests passing


yfarjoun:
- Have you implemented CollectWGSMetrics using this? Is it faster? Uses less memory? Can you provide some incentive for this added complication to the code?
- It seems that you have adapted the tests from SamLocusIterator to test the ReadEndsIterator. I think it would be better to follow the example of MArkDuplicates (in Piacrd) and abstract the testing framework so that the different implementations would share the same underlying tests and we will be guaranteed compatibility of the two implementations. 
- A cursory glance spotted many spelling mistakes, please correct these.

jabbarish:
@yfarjoun yes, we have implemented the described approach, but to make pull request for Picard Tools project we first need to make this proposed changes.

Get details about the implementation of the new CollectWGSMetric from the following pdf, please:
[CollectWgsMetricOptimization.pdf](https://github.com/samtools/htsjdk/files/299024/CollectWgsMetricOptimization.pdf)

magicDGS:
If this is going to break the compatibility, could it be possible for `Locus` to implement `Locatable`?

yfarjoun:
I think it might be helpful to see this class "in action". You can feel
free to submit a PR to picard even though that will clearly fail the tests.

jabbarish:
@magicDGS sorry, could you clarify?  Our changes do not break the backward compatibility.

jabbarish:
@yfarjoun sure, we will do it tomorrow, on Monday.

magicDGS:
Sorry, @jabbarish, what I mean it's that if you're including new interfaces and changing some inheritance stuff, it would be nice to have `Locus` (or at least `AbstractLocusInfo`) implementing [`Locatable`](https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/samtools/util/Locatable.java). Like that, default methods (#443) could be used and will be helpful.

jabbarish:
@magicDGS thank you for the details, we will think what we can do with it.

jabbarish:
@yfarjoun we have corrected some spelling mistakes and refactored tests according to your comments regarding "abstract the testing framework so that the different implementations would share the same underlying tests and we will be guaranteed compatibility of the two implementations".  Check, please.

jabbarish:
@yfarjoun please have a look at https://github.com/broadinstitute/picard/pull/556 with related changes to Picard Tools

yfarjoun:
@jacarey I am soliciting your help in reviewing this code. I am happy to look at the algorithmic part, but I I do not feel a good fit to review the code-structure and the implementation. Can you assist? I know this is  a big PR.

src/main/java/htsjdk/samtools/util/AbstractLocusInfo.java
```diff
@@ -0,0 +1,128 @@
...
+     * @return 1-based reference position
+     */
+    public int getPosition() {
+        return position;
+    }
+
+    /**
+     * @return unmodifiable list of aligned to the reference position <code>recordsAndOffsets</code>
+     */
+    public List<E> getRecordAndPositions() {
```
yfarjoun: shouldn't this be called getRecordAndOffsets ?

src/main/java/htsjdk/samtools/util/AbstractLocusInfo.java
```diff
@@ -0,0 +1,128 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2010 The Broad Institute
```
yfarjoun: 2016

src/main/java/htsjdk/samtools/util/AbstractLocusInfo.java
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * @return the length of reference sequence
+     */
+    public int getSequenceLength() {
+        return referenceSequence.getSequenceLength();
+    }
+
+    /**
+     * @return <code>true</code> if all the RecordAndOffset lists are empty;
```
yfarjoun: there's only one list, no? comment is confusing.

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
```diff
@@ -0,0 +1,572 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2010 The Broad Institute
```
yfarjoun: 2016

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
```diff
@@ -0,0 +1,572 @@
...
+import java.util.List;
+
+/**
+ * Iterator that traverses a SAM File, accumulating information on a per-locus basis.
+ * Optionally takes a target interval list, in which case the loci returned are the ones covered by
+ * the interval list.  If no target interval list, whatever loci are covered by the input reads are returned.
+ * By default duplicate reads and non-primary alignments are filtered out.  Filtering may be changed
+ * via setSamFilters().
+ *
+ * @author alecw@broadinstitute.org
```
yfarjoun: new author

src/main/java/htsjdk/samtools/util/AbstractLocusInfo.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.List;
+
+/**
+ * The unit of iteration for AbstractLocusIterator.
+ * Holds information about the locus (the SAMSequenceRecord and 1-based position on the reference),
+ * plus list of AbstractRecordAndOffset objects,
+ * If <code>RecordAndOffset</code> class is used, one object represents one aligned read that overlaps the locus.
+ * If <code>TypedRecordAndOffset</code> class is used, one object represents one aligned read,
+ * that starts or ends at the locus.
+ */
```
yfarjoun: add '@author'

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
```diff
@@ -0,0 +1,572 @@
...
+ */
+
+public abstract class AbstractLocusIterator<T extends AbstractRecordAndOffset, K extends AbstractLocusInfo<T>> implements Iterable<K>, CloseableIterator<K> {
```
yfarjoun: Given that this class uses 2 different class types, I think it would be easier to read if the names were more informative. 

How about RecordAndOffsetTYPE and LocusInfoTYPE ?

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
```diff
@@ -0,0 +1,572 @@
...
+     * It is no longer the case the useIndex==true can make performance worse.  It should always perform at least
+     * as well as useIndex==false, and generally will be much faster.
+     */
+    private final boolean useIndex;
+
+    /**
+     * LocusInfos on this list are ready to be returned by iterator.  All reads that overlap
+     * the locus have been accumulated before the AbstractLocusInfo is moved into this list.
+     */
+    private final ArrayList<K> complete = new ArrayList<>(100);
```
yfarjoun: should this be <? extends K> ?

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
```diff
@@ -0,0 +1,572 @@
...
+    private boolean enforcedAccumulationLimit = false;
+
+    /**
+     * If true, include indels in the LocusInfo
+     */
+    protected boolean includeIndels = false;
+
+    /**
+     * When there is a target mask, these members remember the last locus for which a AbstractLocusInfo has been
+     * returned, so that any uncovered locus in the target mask can be covered by a 0-coverage AbstractLocusInfo
```
yfarjoun: will actually be covered by a `K` type, right?

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
```diff
@@ -0,0 +1,572 @@
...
+     * @param intervalList Either the list of desired intervals, or null.  Note that if an intervalList is
+     *                     passed in that is not coordinate sorted, it will eventually be coordinated sorted by this class.
+     * @param useIndex     If true, do indexed lookup to improve performance.  Not relevant if intervalList == null.
+     *                     It is no longer the case the useIndex==true can make performance worse.  It should always perform at least
+     *                     as well as useIndex==false, and generally will be much faster.
+     */
+
+    public AbstractLocusIterator(final SamReader samReader, final IntervalList intervalList, final boolean useIndex) {
+        final String className = this.getClass().getSimpleName();
+        if (samReader.getFileHeader().getSortOrder() == null || samReader.getFileHeader().getSortOrder() == SAMFileHeader.SortOrder.unsorted) {
```
yfarjoun: second part of predicate should be 
samReader.getFileHeader().getSortOrder() != SAMFileHeader.SortOrder.coordinate

rather than ==unsorted, since order could also be queryname and that would also be bad.

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
```diff
@@ -0,0 +1,572 @@
...
+
+            LOG.warn(className + " constructed with samReader that has SortOrder == unsorted.  ", "" +
```
yfarjoun: fix error message to emit the order that was found.

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
yfarjoun: NM. didn't see the part below.

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
yfarjoun: ignore.

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
```diff
@@ -0,0 +1,572 @@
...
+        }
+    }
+
+    /**
+     * @return iterator over all/all covered locus position in reference according to <code>emitUncoveredLoci</code>
+     * value.
+     */
+    public Iterator<K> iterator() {
+        if (samIterator != null) {
+            throw new IllegalStateException("Cannot call iterator() more than once on SamLocusIterator");
```
yfarjoun: fix error message. this isn't a SamLocusIterator.

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
```diff
@@ -0,0 +1,572 @@
...
+        }
+        CloseableIterator<SAMRecord> tempIterator;
```
yfarjoun: final

yfarjoun: can you make it final?

src/main/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -495,91 +155,10 @@ private void accumulateIndels(final SAMRecord rec) {
...
-            complete.add(locusInfo);
-        }
-
-        lastReferenceSequence = sequenceIndex;
-        lastPosition = locusInfo.getPosition();
-    }
-
-    /**
-     * Ensure that the queue is populated and get the accumulator offset for the current record
+     * 582
```
yfarjoun: 582?

src/main/java/htsjdk/samtools/util/SamLocusIterator.java
```diff
@@ -495,91 +155,10 @@ private void accumulateIndels(final SAMRecord rec) {
...
+     * +     * Ensure that the queue is populated and get the accumulator offset for the current record
+     * 583
```
yfarjoun: 583?

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
```diff
@@ -0,0 +1,164 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2010 The Broad Institute
```
yfarjoun: 2016

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
```diff
@@ -0,0 +1,164 @@
...
+    private int hash = 0;
+
+    /**
+     * @param record inner SAMRecord
+     * @param offset from the start of the read
+     * @param length of alignment block
+     * @param refPos corresponding to read offset reference position
+     */
+    public AbstractRecordAndOffset(final SAMRecord record, final int offset, int length, int refPos) {
+        validateIndex(offset, record.getBaseQualities());
```
yfarjoun: confusing name, shouldn't this be validateOffset?

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
```diff
@@ -0,0 +1,164 @@
...
+        if (length > record.getReadLength()) {
+            throw new IllegalArgumentException("Block length cannot be larger than whole read length");
+        }
+        this.offset = offset;
+        this.record = record;
+        this.length = length;
```
yfarjoun: Including length even for the default implementation makes the memory requirement even greater for the "slow" implementation. is this required? couldn't you only add the length variable to the Typed implementation and return 1 here in getLength() ?

yfarjoun: Similarly with refPos. It would be better if the default implementation didn't require 2x the amount of memory

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
```diff
@@ -0,0 +1,164 @@
...
+        return record.getReadName();
+    }
+
+    /**
+     * @return array of base qualities of inner SAMRecord.
+     */
+    public byte[] getBaseQualities() {
+        return record.getBaseQualities();
+    }
+
```
yfarjoun: extra newline

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
```diff
@@ -0,0 +1,164 @@
...
+     * @return the read base according to <code>offset</code>.
+     */
+    public byte getReadBase() {
+        return record.getReadBases()[offset];
+    }
+
+    /**
+     * @return the base quality according to <code>offset</code>.
+     */
+    public byte getBaseQuality() {
```
yfarjoun: please move this method lower so it will be next to getBaseQuality(int)

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
```diff
@@ -0,0 +1,164 @@
...
+    public byte[] getBaseQualities() {
+        return record.getBaseQualities();
+    }
+
+
+    /**
+     * @param position in the reference
+     * @return base quality of a read base, corresponding to a given reference position
+     */
+    public byte getBaseQuality(int position) {
```
yfarjoun: where is this function used?

src/main/java/htsjdk/samtools/util/ReadEndsIterator.java
```diff
@@ -0,0 +1,207 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2010 The Broad Institute
```
yfarjoun: 2016

src/main/java/htsjdk/samtools/util/ReadEndsIterator.java
```diff
@@ -0,0 +1,207 @@
...
+     */
+    @Override
+    public void setMaxReadsToAccumulatePerLocus(int maxReadsToAccumulatePerLocus) {
+        throw new UnsupportedOperationException("Locus cap is not supported for " + getClass().getSimpleName() + ".");
+    }
+
+    /**
+     * This method isn't supported in current implementation.
+     *
+     * @param qualityScoreCutoff the minimum bae quality to include in <code>AbstractLocusInfo</code>.
```
yfarjoun: base

src/main/java/htsjdk/samtools/util/ReadEndsIterator.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    /**
+     * This method isn't supported in current implementation.
+     *
+     * @param maxReadsToAccumulatePerLocus maximum number of <code>RecordAndOffset</code> objects to store for
+     *                                     one loci in reference sequence.
+     */
+    @Override
+    public void setMaxReadsToAccumulatePerLocus(int maxReadsToAccumulatePerLocus) {
```
yfarjoun: hmm. this will change behaviour. is there a plan to make this work?

src/main/java/htsjdk/samtools/util/ReadEndsIterator.java
```diff
@@ -0,0 +1,207 @@
...
+        throw new UnsupportedOperationException("Locus cap is not supported for " + getClass().getSimpleName() + ".");
+    }
+
+    /**
+     * This method isn't supported in current implementation.
+     *
+     * @param qualityScoreCutoff the minimum bae quality to include in <code>AbstractLocusInfo</code>.
+     */
+    @Override
+    public void setQualityScoreCutoff(int qualityScoreCutoff) {
```
yfarjoun: same here...how can one filter on quality then?

src/main/java/htsjdk/samtools/util/ReadEndsIterator.java
```diff
@@ -0,0 +1,207 @@
...
+
+    /**
+     * For correct work of <code>ReadEndsIterator</code> value <code>emitUncoveredLoci</code> must be true.
+     *
+     * @param emitUncoveredLoci if false, iterator will skip uncovered loci in reference sequence, otherwise
+     *                          empty <code>AbstractLocusInfo</code> will be created for each loci.
+     */
+    @Override
+    public void setEmitUncoveredLoci(boolean emitUncoveredLoci) {
+        throw new UnsupportedOperationException(getClass().getSimpleName() + " doesn't support work with skipping " +
```
yfarjoun: shouldn't the exception only be thrown if emitUncoveredLoci=false ?

src/main/java/htsjdk/samtools/util/ReadEndsIterator.java
```diff
@@ -0,0 +1,207 @@
...
+                "uncovered bases.");
+    }
+
+
+    @Override
+    public void setIncludeIndels(boolean includeIndels) {
+        throw new UnsupportedOperationException("Indels accumulation is not supported for " + getClass().getSimpleName() + ".");
```
yfarjoun: only if true, yes?

src/main/java/htsjdk/samtools/util/ReadEndsIterator.java
```diff
@@ -0,0 +1,207 @@
...
+            // Ensure there are LocusInfos up to and including this position
+            for (int j = accumulator.size(); j <= refOffsetEnd; ++j) {
+                accumulator.add(createLocusInfo(getReferenceSequence(rec.getReferenceIndex()),
+                        rec.getAlignmentStart() + j));
+            }
+
+            int refOffsetInterval = refOffset;
+            int refOffsetEndInterval = refOffsetEnd;
+            int startShift = 0;
+
```
yfarjoun: can the general idea be written here in a comment? it would be very helpful for the review process and for later additions. Also, this block of code is very dense and seems the most important part of the PR, could you add comments to the code to explain what is happening?

src/main/java/htsjdk/samtools/util/TypedRecordAndOffset.java
```diff
@@ -0,0 +1,100 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2010 The Broad Institute
```
yfarjoun: 2016

src/main/java/htsjdk/samtools/util/TypedRecordAndOffset.java
```diff
@@ -0,0 +1,100 @@
...
+import htsjdk.samtools.SAMRecord;
+
+import static htsjdk.samtools.util.TypedRecordAndOffset.Type.BEGIN;
+import static htsjdk.samtools.util.TypedRecordAndOffset.Type.END;
+
+/**
+ * Holds a SAMRecord plus the zero-based offset into that SAMRecord's bases and quality scores that corresponds
+ * to the base and quality for the start of alignment block at the genomic position described by the AbstractLocusInfo.
+ * This is implementation for ReadEndsIterator, field <code>type</code> added to indicate whether object
+ * represents the start or the end of an alignment block.
```
yfarjoun: please also explain what the difference between the start and the end is and how that is to be used.

src/test/java/htsjdk/samtools/util/AbstractLocusInfoTest.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2010 The Broad Institute
```
yfarjoun: 2016

src/main/java/htsjdk/samtools/util/ReadEndsIterator.java
```diff
@@ -0,0 +1,207 @@
...
+/**
+ * Iterator that traverses a SAM File, accumulating information on a per-locus basis.
+ * Optionally takes a target interval list, in which case the loci returned are the ones covered by
+ * the interval list.  If no target interval list, whatever loci are covered by the input reads are returned.
+ * By default duplicate reads and non-primary alignments are filtered out.  Filtering may be changed
+ * via setSamFilters(). Difference from SamLocusIterator is that this implementation accumulates data
+ * only about start and end of alignment blocks from reads, not about each aligned base.
+ */
+public class ReadEndsIterator extends AbstractLocusIterator<TypedRecordAndOffset, AbstractLocusInfo<TypedRecordAndOffset>> {
```
yfarjoun: ReadEnds is not a good name. it's reminiscent of the ReadEnds object in MarkDuplicates, but seems to have nothing to do with it...please come up with a better name...one that has, perhaps `AlignmentBlocks` in it?

src/main/java/htsjdk/samtools/util/ReadEndsIterator.java
```diff
@@ -0,0 +1,207 @@
...
+     *
+     * @param samReader must be coordinate sorted
+     */
+    public ReadEndsIterator(final SamReader samReader) {
+        this(samReader, null);
+    }
+
+    /**
+     * Prepare to iterate through the given SAM records, skipping non-primary alignments.  Do not use
+     * BAM index even if available.
```
yfarjoun: comment about index doesn't match code.

src/main/java/htsjdk/samtools/util/ReadEndsIterator.java
```diff
@@ -0,0 +1,207 @@
...
+     *                     It is no longer the case the useIndex==true can make performance worse.  It should always perform at least
+     *                     as well as useIndex==false, and generally will be much faster.
+     */
+    public ReadEndsIterator(final SamReader samReader, final IntervalList intervalList, final boolean useIndex) {
+        super(samReader, intervalList, useIndex);
+    }
+
+    /**
+     * Capture the loci covered by the given SAMRecord in the LocusInfos in the accumulator,
+     * creating new LocusInfos as needed. TypedRecordAndOffset object are created only for start
```
yfarjoun: LocusInfo slipped in here.

yfarjoun:
First pass review is done, though I haven't looked carefully at the heart of the new algorithm as I would like there to be more comments explaining how it works. In particular the use of the typed "ReadEnds seems superfluous as the length of the alignment block is embedded within each read-end, and so it seems that the other end should not be required.

- The work `index` is used inappropriately. Especially referenceIndex is normally used to indicate a reference sequence out of the possible O(20), not the position within the reference. For that I think one should either use `position` or `offset`
- I compiled and ran the new version (and the old version) of CollectWgsMetrics on a 30x WGS bam (5x for each option). I see the following runtimes:
   - new CollectWgsMetrics with USE_FAST_ALGORITHM=true: ~3h
   - new CollectWgsMetrics with USE_FAST_ALGORITHM=false: ~6h
   - old CollectWgsMetrics : ~4h.

This indicates to me that while the new algorithm is 25% faster that the original, the reimplementation of SamLocusInfo underwent a significant slowdown, possibly due to the addition of 2 extra variables (length and referencePosition), which means much more memory allocation and garbage-collection. I would much rather see no change in the original version.

The resulting metrics are identical, which is great!

Thanks for this effort to make locus-walking faster! and sorry for the lengthy review process.

Back to you @mzueva

yfarjoun:
hmmm. a rerun of the algorithms shows slighly different results:

image

As you can see the old implementation looks better. I'm not sure why I got different results before, but now I'm questioning the utility of this change. 

This was run on a 33x WGS sample.

jabbarish:
@yfarjoun could we get an access to the sample?

yfarjoun:
Sure. let me know over email how to give you the file.

yfarjoun:
still waiting on a reply here...

jabbarish:
@yfarjoun we are running tests using the data you passed to us.  By the way, could you send me the script you used to get the picture?

yfarjoun:
```code
#!/bin/bash
BAM=input.bam
reference=Homo_sapiens_assembly19.fasta
PICARD=PicardWithFastCollectWgsMetrics.jar

i=0
for fast in true false ; do
        for iteration in {1..5}; do

                echo "{ time java -jar $PICARD CollectWgsMetrics USE_FAST_ALGORITHM=$fast INCLUDE_BQ_HISTOGRAM=true I=$BAM R=$reference O=fast_is_$fast.results.$iteration.wgs_metrics; } 2> fast_is_$fast.time.$iteration.txt" > temp.$i.sh

                chmod a+x ./temp.$i.sh
                submit_job ./temp.$i.sh 
i=$((i + 1))

        done

done 

exit;

PICARD=OldPicard.jar



for iteration in {1..5}; do

        echo "{ time java -jar $PICARD CollectWgsMetrics INCLUDE_BQ_HISTOGRAM=true I=$BAM R=$reference O=old.results.$iteration.wgs_metrics; } 2> old.time.$iteration.txt" > temp.$i.sh

        chmod a+x ./temp.$i.sh

        submit_job ./temp.$i.sh 

        i=$((i + 1))

done

**Comment by yfarjoun** [comment]
then to plot I used R:
```code
library(ggplot2)

files = dir('.','*.txt')

read_output<-function(file){
  str=str_split(rev(readLines(file))[3],'\t')[[1]][2]
  m=as.numeric(gsub('([:alpha:]*)m.*',"\\1",x=str))
  s=as.numeric(gsub('^.*m([0-9.]+)s',"\\1",x=str))
  data.frame(time=m+s/60)
}

data=ldply(files,function(x) data.frame(type=gsub("\\..*","",x), number=gsub(".*\\.([1-5])\\..*","\\1",x),read_output(x)))
ggplot(data)+geom_boxplot(aes(x=type,y=time,color=type))
```
I hope this helps.

jabbarish:
Sure, thx.

jabbarish:
@yfarjoun could you clarify,  what does submit_job in the script?  Do all jobs run sequentially, one by one, or simultaneously? If they work in parallel, do they all read the very same bam and reference file?  If so, it may have influence on the results.

We couldn't get a picture like yours where current master version (original) runs faster then our fast=true version.
But fast=false version is really 10 to 15 % slower than the original. We are trying to solve this issue now.

yfarjoun:
Yes. all the jobs were running on our farm at the same time. They were
accessing the same bam and reference at the same time. the files are so
large that I doubt that caching is helping one process more than another,
but I'm happy to rerun the jobs locally and without using the farm.

jabbarish:
Thank you, it would be very helpful.  Also, it would be nice to know specs of the environment in which the metrics  were launched: CPU, available RAM for JVM, Storage (SSD/HDD).

jabbarish:
Here is our chart (Linux, i5 / 4 cores, 4gb RAM for JVM, HDD;  all runs were done sequentially):
image

yfarjoun:
OK. I'll run sequentially.

yfarjoun:
I reran serially, on a single node with no other jobs running on that node. I got results that look a lot like your @jabbarish:
image

However, I'm unhappy that the old way becomes slower with this implementation. would it be possible to avoid that?

jabbarish:
@yfarjoun we found the origin of the problem, we introduced some regression during refactoring. We already have a solution, but our fast-false version is yet a little bit slower (3%) then the original one. I will send you the jar later this weekend.  We still hope to find a better solution, though.

yfarjoun:
I'm not concerned about 3% slowdown of the original version given that this
gives identical results in WGS metrics. a 40% slowdown was more worrisome.

magicDGS:
@jabbarish, does the refactoring introduce regression into indel-tracking? If so, how severe is the slowdown?

jabbarish:
image

jabbarish:
@magicDGS we believe our refactoring doesn't introduce any regression into indel-tracking.

magicDGS:
Thank you, @jabbarish.

yfarjoun:
This looks better @jabbarish. Can you fix the merge conflicts and then we can merge. Thanks!

yfarjoun:
Of-course, the actual changes that made the improvements are not yet committed...I'm assuming they'll be included in the final solution.

jabbarish:
@yfarjoun we have updated the branch, but 1 test failed, for FTP existence.  What to do then? )

yfarjoun:
that's probably transient, I restarted the test.

yfarjoun:
FYI: I'm kicked off the timing code again with the new commit.

.gitignore
```diff
@@ -17,3 +17,4 @@ src/htsjdk.iml
 #gradle stuff
 .gradle/
 build/
+/bin/
```
yfarjoun: is this needed?

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
```diff
@@ -59,14 +53,16 @@
...
-        validateIndex(offset, record.getBaseQualities());
-        if (length > record.getReadLength()) {
-            throw new IllegalArgumentException("Block length cannot be larger than whole read length");
-        }
+        this(record, offset);
+    }
+
+    /**
+     * @param record inner SAMRecord
+     * @param offset from the start of the read
```
yfarjoun: the code and comments are somewhat contradictory: what is offset? here it says "offset from the strart of read" but in getBaseQuality it talks about the reference....

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
```diff
@@ -91,24 +87,17 @@ public byte getReadBase() {
...
-        return length;
+        return 1;
     }
 
     /**
      * @return the position in reference sequence, to which the start of alignment block is aligned.
      */
     public int getRefPos() {
-        return refPos;
+        return 0;
```
yfarjoun: zero as an error code? better use -1 to avoid confusing with zero-based indexing.

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
```diff
@@ -124,41 +113,20 @@ public String getReadName() {
     public byte[] getBaseQualities() {
         return record.getBaseQualities();
     }
-
+    
+    /**
+     * @return the base quality according to <code>offset</code>.
+     */
+    public byte getBaseQuality() {
+        return record.getBaseQualities()[offset];
```
yfarjoun: so this is the base quality of the base at the "start" of the AbstractRecordAndOffset, yes?

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
```diff
@@ -124,41 +113,20 @@ public String getReadName() {
...
-
-    @Override
-    public int hashCode() {
-        if (hash != 0) return hash;
-        hash = record.hashCode();
-        hash = 31 * hash + length;
-        hash = 31 * hash + offset;
-        hash = 31 * hash + refPos;
-        return hash;
+        return record.getBaseQualities()[position - offset];
```
yfarjoun: why is position-offset a reasonable number to look into the BaseQ array? could it not be negative? or too large?

src/main/java/htsjdk/samtools/util/EdgeReadIterator.java
```diff
@@ -98,40 +100,52 @@ void accumulateSamRecord(SAMRecord rec) {
...
+             * We put two records to the accumulator. The first one has the "begin" type which corresponds to the locus 
+             * where the block starts. The second one has the "end" type which corresponds to the other locus where the block ends. 
+            */
+            int refOffsetInterval = refOffset; // corresponds to the beginning of the alignment block 
             int refOffsetEndInterval = refOffsetEnd;
             int startShift = 0;
 
+            // intersect intervals and alignment block
             if (getIntervals() != null) {
+                // get the current interval we're proceeding
```
yfarjoun: proceeding -> processing?

yfarjoun:
I have the latest timing results: 
timingsep19 which still shows a 30% increase in run-time for the "fast=FALSE" way of running. I would want the "old" SamLocusIterator to continue working without a slowdown. 

Could you try to see why the old way suffers so much from these changes?

jabbarish:
@yfarjoun hm... thank you!  It is very strange...  Could you send me the .jar you built?

yfarjoun:
I retract my statment!! stay tuned!

yfarjoun:
I'm finally convinced that the most recent implementation doesn't create any loss of performance for the old SamLocusIterator, while the new implementation results in a 30% reduction in runtime! This is great!!

I've made some stylistic comments and I think that when these are addressed we will be very close to being able to merge this. 

Thanks!!

image

jabbarish:
Thanks!

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
jabbarish: We think it's OK to use just K there

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
jabbarish: Right.  We think it is not reasonable to change it to K, because K extends AbstractLocusInfo.

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
jabbarish: This method is called by picard.analysis.FastWgsMetricsCollector.excludeByQuality() only (see changes in Picard)

.gitignore
jabbarish: sorry, removed.

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
jabbarish: From our point of view this text is correct:  take the read, skip 'offset' bases - and you'll get the start of the RecordAndOffset object.

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
jabbarish: fixed

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
jabbarish: yes

src/main/java/htsjdk/samtools/util/EdgeReadIterator.java
jabbarish: fixed

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
jabbarish: fixed

jabbarish:
@yfarjoun We have made some changes according to your comments. Could you have a look, please?

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
```diff
@@ -0,0 +1,132 @@
...
+    
+    /**
+     * @return the base quality according to <code>offset</code>.
+     */
+    public byte getBaseQuality() {
+        return record.getBaseQualities()[offset];
+    }
+
+    /**
+     * @param position in the reference
```
yfarjoun: this must be position in the read, not reference, yes?

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
```diff
@@ -0,0 +1,574 @@
...
+                    referenceSequenceMask.getMaxPosition() + 1);
+            // In this case... we're past the last read from SAM so see if we can
+            // fill out any more (zero coverage) entries from the mask
+            return createNextUncoveredLocusInfo(afterLastMaskPositionLocus);
+        } else {
+            return null;
+        }
+    }
+
+
```
yfarjoun: spare newlines (also below)

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
```diff
@@ -0,0 +1,132 @@
...
+     */
+    public int getLength() {
+        return 1;
+    }
+
+    /**
+     * @return the position in reference sequence, to which the start of alignment block is aligned.
+     */
+    public int getRefPos() {
+        return 0;
```
yfarjoun: this is an error code again, right? -1 please.

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * @param position in the reference
+     * @return base quality of a read base, corresponding to a given reference position
+     */
+    public byte getBaseQuality(int position) {
+        return record.getBaseQualities()[position - offset];
+    }
+
```
yfarjoun: superfluous newline

src/main/java/htsjdk/samtools/util/EdgeReadIterator.java
```diff
@@ -0,0 +1,227 @@
...
+    /**
+     * This method isn't supported in current implementation.
+     *
+     * @param maxReadsToAccumulatePerLocus maximum number of <code>RecordAndOffset</code> objects to store for
+     *                                     one loci in reference sequence.
+     */
+    @Override
+    public void setMaxReadsToAccumulatePerLocus(int maxReadsToAccumulatePerLocus) {
+        if (getMaxReadsToAccumulatePerLocus() != maxReadsToAccumulatePerLocus) {
+            throw new UnsupportedOperationException("Locus cap is not supported for " + getClass().getSimpleName() + ".");
```
yfarjoun: I'm confused. We can set a Cap (as per the picard branch that does this) so what's this about?

src/main/java/htsjdk/samtools/util/EdgeReadIterator.java
```diff
@@ -0,0 +1,227 @@
...
+     *                          empty <code>AbstractLocusInfo</code> will be created for each loci.
+     */
+    @Override
+    public void setEmitUncoveredLoci(boolean emitUncoveredLoci) {
+        if (isEmitUncoveredLoci() != emitUncoveredLoci) {
+            throw new UnsupportedOperationException(getClass().getSimpleName() + " doesn't support work with skipping " +
+                    "uncovered bases.");
+        }
+    }
+
```
yfarjoun: one of these two newline should move to the end of the file

src/main/java/htsjdk/samtools/util/EdgingRecordAndOffset.java
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * @return the position in reference sequence, to which the start of alignment block is aligned.
+     */
+    @Override
+    public int getRefPos() {
+        return refPos;
+    }
+
+
```
yfarjoun: extra nl

yfarjoun:
any progress on responding here? @ZLyanov

ZLyanov:
@yfarjoun The changes you requested are significant and numerous, so it will take some time to implement them. @mzueva could you please describe the current state?

yfarjoun:
no problem. It wasn't clear what you thought of me suggestions. happy to
hear that you've accepted the challenge and are responding :-)

src/main/java/htsjdk/samtools/util/EdgingRecordAndOffset.java
```diff
@@ -0,0 +1,171 @@
...
+        return start;
+    }
+
+
+    /**
+     * Method can be called only for an object with type END.
+     * @param start <code>TypedRecordAndOffset</code> that represents the start of alignment block of the read,
+     *              must have type <code>BEGIN</code>.
+     */
+    public void setStart(EdgingRecordAndOffset start) {
```
mzueva: I've implemented this part using one abstract class EdgingRecordAndOffset and two it subclasses for BEGIN and END cases. I think code looks cleaner now and we save some memory for END records, as you suggested.

src/main/java/htsjdk/samtools/util/EdgeReadIterator.java
mzueva: Yes, this part is unclear, if we are trying to set 0 value for this cap we shouldn't throw an exception in any other way - we do, as we do not support this kind of capping for this implementation

src/main/java/htsjdk/samtools/util/EdgeReadIterator.java
```diff
@@ -0,0 +1,227 @@
...
+                        continue;
+                    }
+                    // if the alignment block starts out of an interval shift the starting position
+                    if (refPos < intervalStart) {
+                        startShift = intervalStart - refPos;
+                        refOffsetInterval = refOffsetInterval + startShift;
+                    }
+                    // if the alignment block ends out of an interval shift the ending position
+                    if (refPos + alignmentBlock.getLength() > intervalEnd) {
+                        refOffsetEndInterval = refOffsetEndInterval - (refPos + alignmentBlock.getLength() - intervalEnd) + 1;
```
mzueva: These lines are not equivalent, I've tried several variants, but unfortunately I failed to make it final. I hope at least the comment clarifies the code.

mzueva:
I've updated the branch according to requested changes and gave some comments. @yfarjoun , please take a look.

src/main/java/htsjdk/samtools/util/AbstractLocusInfo.java
yfarjoun: could you respond to the renaming request?

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
yfarjoun: please add the name of the new author

src/main/java/htsjdk/samtools/util/AbstractRecordAndOffset.java
yfarjoun: I still do not understand the comment here. if `position` is a number in the reference (i.e. can be as big as the length of the reference), and offset is a size smaller than the length of the read then how can you access record.getBaseQualities()[position - offset] ? I think that position must be a position in the read, not the reference.

mzueva:
@yfarjoun, I'm sorry, but I can't post comments on the outdated diffs, so I'll answer here.
1. Renaming getRecordsAndPositions(). I hope, I understood your request correctly: I've implemented the corresponding changes in a new class AbstractRecordAndOffset (added a new method with a correct name) and marked an old one as deprecated.
2. AbstractLocusIterator.java: authors were fixed
3. AbstractRecordAndOffset.java getBaseQuality(int position): you are completely right, it is incorrect. This method has correct implementation in EdgingRecordAndOffset class and we don't need in the superclass, so i removed it.

yfarjoun:
Thanks for bearing with me and my nit-picking comments! I'll take this!

👍

yfarjoun:
🎉 🎈

ZLyanov:
Thank you!

Issue #545, opened 2016-03-31
SRA tests fail with missing accession


nh13:
### Description
See the travis build log: https://s3.amazonaws.com/archive.travis-ci.org/jobs/119821702/log.txt

```
   [testng] FAILED: ("SRR1063272", 0, 99, "SRR1063272.R.1", "ACTCGACATTCTGCCTTCGACCTATCTTTCTCCTCTCCCAGTCATCGCCCAGTAGAATTACCAGGCAATGAACCAGGGCCTTCCATCCCAACGGCACAGCA", "@@CDDBDFFBFHFIEEFGIGGHIEHIGIGGFGEGAFDHIIIIIGGGDFHII;=BF@FEHGIEEH?AHHFHFFFFDC5'5=?CC?ADCD@AC??9BDDCDB<", 86, "101M", "supercont2.1", 60, true, false)
   [testng] java.lang.IllegalArgumentException: Invalid SRA accession was passed to SRA reader: SRR1063272
```

This blocks https://github.com/samtools/htsjdk/pull/516 and likely any other PR that encounters these issues

### Desired Actions
I would like to know why the test failed and  what are the plans to ensure that the test does not fail in the future.   This may not be a new issue, see https://github.com/samtools/htsjdk/issues/461, so I am looking for a detailed response and robust solution if this is a re-occurrence.

@a-nikitiuk could you take a look or route to the appropriate person?  

Thank-you in advance to anyone who has the chance to take a look.

a-nikitiuk:
NCBI had network connectivity issues today that resulted in bad database connections. Some of the issues have been cleared up already, but we continue to work on the problem. It has nothing to do with the NGS code and it is simply a network problem which resulted in a failure to resolve SRA accession that you saw in the tests

nh13:
@a-nikitiuk that makes a lot more sense.  Is there somewhere we could check the status of SRA network connectivity in case this happens again?  A status page (ex https://status.github.com/) would be great to see the current and past uptime.

The error message should be improved.  The message states that an invalid accession was given but really it was a problem querying SRA due to network issues.  Perhaps something like "network connection timeout".  @a-nikitiuk would you be able to update the error message?  I can imagine a user who doesn't have their internet connected would be confused by such a message for example, in addition to when there is network connection problems at SRA.   What do you think about modifying the error message?

eitanbanks:
I agree that a better error message would be infinitely more helpful here.

a-nikitiuk:
[...]

Issue #460, opened 2016-02-04
Improve error messages when instantiating an SRAFileReader


lbergelson:
Currently constructing a new `SRAFileReader` calls `SRAAccession.isValid()` and throws an exception saying `Invalid SRA accession was passed to SRA reader` if `isValid` returns false.

Since is valid can fail for a number of reasons, this is not a very helpful message.  It's particularly not helpful because it could actually be a valid accession, but SRA is just not supported.

This should be improved so that someone instantiating an SRAReader can tell WHY it was considered invalid.

lbergelson:
@a-nikitiuk @kwrodarmer  Can you take this one please.

kwrodarmer:
Yes, we have already been working on a new reporting framework for this, as you have suggested, based upon various exception types.

cmnbroad:
Can this be closed (fixed in https://github.com/samtools/htsjdk/pull/638 ?)

a-nikitiuk:
Yes, it should be

cmnbroad:
Closed via https://github.com/samtools/htsjdk/pull/638.

PR #638, opened 2016-06-07
Improved SRA tests, corrected error message

a-nikitiuk:
### Description

These changes are intended to resolve most of previously reported SRA-related issues:

- #442 - updated tests to use much smaller files, so that footprint by SRA cache on Mac OS X is about 30 Mb (on other OS issue does not exist, since we use sparse files there and footprint is even smaller)
- #459, #461 - now each network-related test checks if it can resolve an accession, and skips the test in case of failure
- #460, #545 - updated error message in SRAFileReader, so that it tells about a possibility of a connection problem

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SRAFileReader.java
+++ b/src/main/java/htsjdk/samtools/SRAFileReader.java
@@ -61,7 +61,8 @@ public class SRAFileReader extends SamReader.ReaderImplementation implements Sam
         this.acc = acc;
 
         if (!acc.isValid()) {
-            throw new IllegalArgumentException("Invalid SRA accession was passed to SRA reader: " + acc);
+            throw new IllegalArgumentException("SRAFileReader: cannot resolve SRA accession '" + acc + "'\n" +
+                "Possible causes are an invalid SRA accession or a connection problem.");
         }
 
         try {
--- a/src/test/java/htsjdk/samtools/sra/AbstractSRATest.java
+++ b/src/test/java/htsjdk/samtools/sra/AbstractSRATest.java
@@ -7,6 +7,7 @@ import org.testng.SkipException;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
+import java.lang.reflect.Method;
 import java.util.NoSuchElementException;
 
 @Test(groups = "sra")
@@ -16,6 +17,25 @@ public abstract class AbstractSRATest {
     public final void assertSRAIsSupported(){
         if(!SRAAccession.isSupported()){
             throw new SkipException("Skipping SRA Test because SRA native code is unavailable.");
+        }
+    }
+
+    @BeforeMethod
+    public final void skipIfCantResolve(Method method, Object[] params) {
+        String accession = null;
+
+        if (params.length > 0) {
+            Object firstParam = params[0];
+            if (firstParam instanceof String) {
+                accession = (String)firstParam;
+            } else if (firstParam instanceof SRAAccession) {
+                accession = firstParam.toString();
+            }
+        }
+
+        if (accession != null &&
+                accession.matches(SRAAccession.REMOTE_ACCESSION_PATTERN) && !SRAAccession.isValid(accession)) {
+            throw new SkipException("Skipping network SRA Test because cannot resolve SRA accession.");
         }
     }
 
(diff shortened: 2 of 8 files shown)


src/main/java/htsjdk/samtools/SRAFileReader.java
```diff
@@ -61,7 +61,8 @@ public SRAFileReader(final SRAAccession acc) {
         this.acc = acc;
 
         if (!acc.isValid()) {
-            throw new IllegalArgumentException("Invalid SRA accession was passed to SRA reader: " + acc);
+            throw new IllegalArgumentException("SRAFileReader: cannot resolve SRA accession '" + acc + "'\n" +
```
yfarjoun: isn't it clear at some point in the code that a connection has been established but that the accession is bad? why not inform the user?

src/test/java/htsjdk/samtools/sra/SRAIndexTest.java
```diff
@@ -46,7 +46,7 @@
  * Created by andrii.nikitiuk on 10/28/15.
  */
 public class SRAIndexTest extends AbstractSRATest {
-    private static final SRAAccession DEFAULT_ACCESSION = new SRAAccession("SRR1298981");
+    private static final SRAAccession DEFAULT_ACCESSION = new SRAAccession("SRR2096940");
```
yfarjoun: why were these changed? what's wrong with the old ones?

yfarjoun: sorry, I see the comment now: I supposed these are the smaller files.

yfarjoun:
@a-nikitiuk back to you. only one comment really: can you not distinguish between a bad connection and a bad accesssion string? It seems that some part of the code must know, so perhaps you can give an return code to isValid() that will be more than a boolean and work with that. The problem is that if the accession becomes "bad" because of code change we will never find out because we will skip the test.

src/test/java/htsjdk/samtools/sra/SRAIndexTest.java
a-nikitiuk: yes, that is the reason

a-nikitiuk:
Unfortunately, it is not possible in current version of NGS API. We are working on improvement, but that is not an easy and quick fix.

yfarjoun:
In that case @a-nikitiuk I think it would be better to test connectivity with a single `@BeforeGroup` using an accession that is bound to exist (perhaps you have some sort of testing object that can be used for that?) and then remove the `@BeforeTest` so that each of the individual tests will fail if their accession is not valid. while this is fragile wrt losing connectivity in the middle of running the tests, it doesn't require a separate check that all the accessions are valid.

yfarjoun:
@a-nikitiuk This needs to be rebased, and also did you have time to think about my suggestion of using `@BeforeGroup` rather than `@BeforeTest` ?

yfarjoun:
@a-nikitiuk will you have time to rebase and replace `@BeforeMethod` with `@BeforeTest` ?

We would love to have these changes....

a-nikitiuk:
Yeah, will do that. Sorry, did not have much time lately

a-nikitiuk:
Rebased and added `@BeforeGroups` method

src/test/java/htsjdk/samtools/sra/AbstractSRATest.java
```diff
@@ -4,21 +4,51 @@
...
+            Object firstParam = params[0];
+            if (firstParam instanceof String) {
+                accession = (String)firstParam;
+            } else if (firstParam instanceof SRAAccession) {
+                accession = firstParam.toString();
+            }
+        }
+
+        if (accession != null &&
+                accession.matches(SRAAccession.REMOTE_ACCESSION_PATTERN) && !canResolveNetworkAccession) {
```
yfarjoun: Since you are no longer asking to connect for every accession, wouldn't it be better to remove the first condition, so that if someone puts in a non-matching accession to a new test, it will fail?

a-nikitiuk: If someone puts a non-matching accession, then this code will do nothing. However, it detects **remote** accessions from its first argument, and if remote accession is detected and `canResolveNetworkAccession` is `false` then it skips the test.

yfarjoun: Oh, OK. can you then simply change the text in the Exception to:
1. note the accession that cannot be resolved, and 
2. change "SRA accession" with "remote SRA accession"?

yfarjoun:
@a-nikitiuk are you planing to make these changes, or should I do them?

a-nikitiuk:
You can do them if you would like to. Or I will try to find time for that today-tomorrow

yfarjoun:
OK. I'll leave it to you then! thanks!

a-nikitiuk:
Done

yfarjoun:
thanks! 👍

PR #639, opened 2016-06-10
Move the last dangling test file out of the old testdata directory.

cmnbroad:
### Description

One more test file left over from https://github.com/samtools/htsjdk/pull/617


nh13:
👍

PR #641, opened 2016-06-16
split async flags for read and write

akiezun:
### Description

This PR splits the flags that control the async behavior to read/write to give more fine-grained control to clients. This will be important when hooking up #576 and make multi-threaded reading an opt-in option.

Not backward compatible because I removed the master option `USE_ASYNC_IO` - having that option conflicts with the opt-in requirement. I think an uber  `USE_ASYNC_IO` is just too broad. Note that there's no `USE_ASYNC_IO_READ_FOR_TRIBBLE` because we have no asynch tribble readers right now.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

@droazen can you review?

### DIFF
--- a/src/main/java/htsjdk/samtools/Defaults.java
+++ b/src/main/java/htsjdk/samtools/Defaults.java
@@ -22,25 +22,20 @@ public class Defaults {
     /** Should MD5 files be created when writing out SAM and BAM files?  Default = false. */
     public static final boolean CREATE_MD5;
 
-    /** Should asynchronous I/O be used where supported throughout all of htsjdk (one thread per file).
-     *  Note: this option takes precedence over {@link #USE_ASYNC_IO_FOR_SAMTOOLS} and {@link #USE_ASYNC_IO_FOR_TRIBBLE}.
+    /** Should asynchronous read I/O be used where supported by the samtools package (one thread per file).
      *  Default = false.
      */
-    public static final boolean USE_ASYNC_IO;
+    public static final boolean USE_ASYNC_IO_READ_FOR_SAMTOOLS;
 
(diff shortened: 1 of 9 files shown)


src/main/java/htsjdk/samtools/Defaults.java
```diff
@@ -126,9 +115,9 @@
         final SortedMap<String, Object> result = new TreeMap<>();
         result.put("CREATE_INDEX", CREATE_INDEX);
         result.put("CREATE_MD5", CREATE_MD5);
-        result.put("USE_ASYNC_IO", USE_ASYNC_IO);
-        result.put("USE_ASYNC_IO_FOR_SAMTOOLS", USE_ASYNC_IO_FOR_SAMTOOLS);
-        result.put("USE_ASYNC_IO_FOR_TRIBBLE", USE_ASYNC_IO_FOR_TRIBBLE);
+        result.put("USE_ASYNC_IO_READ_FOR_SAMTOOLS", USE_ASYNC_IO_READ_FOR_SAMTOOLS);
+        result.put("USE_ASYNC_IO_WRITE_FOR_SAMTOOLS", USE_ASYNC_IO_WRITE_FOR_SAMTOOLS);
+        result.put("USE_ASYNC_IO_WRTIE_FOR_TRIBBLE", USE_ASYNC_IO_WRITE_FOR_TRIBBLE);
```
droazen: `WRTIE` -> `WRITE`

droazen:
:+1: Looks good to me. One typo to fix, then merge when ready. Back to @akiezun

src/main/java/htsjdk/samtools/Defaults.java
akiezun: done

PR #642, opened 2016-06-16
Removed PollableTreeSet in SortingCollection as it was no longer necessary.

tfenne:
### Description

Replaced with direct usage of TreeSet.pollFirst(). One advantage (other than removing unneeded code) is that this now only has to search the tree once, instead of twice for poll() then remove().

This also resolved a bug I was experiencing where the `remove()` call was getting a NoSuchElementException that I couldn't figure out.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


tfenne:
@nh13 want to take a quick look?  Basically just drop-in replacing custom code with functionality that is now in the JDK.

nh13:
+1

nh13:
You probably removed more code that was covered than uncovered.

Issue #506, opened 2016-02-29
SAMFileHeader in CRAM files is missing attributes required by the CRAM spec


cmnbroad:
AFAICT, CRAM files created with htsjdk contain the samFileHeader that was passed in to the CRAM writer, whether or not it has the attributes required by the CRAM spec.The following are missing:
- The SQ:MD5 checksum is required unless the reference sequence has been embedded into the file.
- At least one RG record is required.
- UR (URI for the sequence fasta) field is strongly advised

I think we should fix this before we merge in https://github.com/samtools/htsjdk/pull/478, and add corresponding checks to the CRAMFileReader to validate these on input for >= v3.0 files.

vadimzalunin:
ah, yes, this is done in cramtools by i missed it here...

cmnbroad:
This should be easy to fix. We really should do it. See https://github.com/samtools/htslib/issues/769.

cmnbroad:
I don't see any reference in the spec about a read group requirement, so I assume thats been removed, and will only implement the MD5 requirement.

PR #644, opened 2016-06-20
Throw exception on MD5 mismatch when constructing SAMRecords from CRA…

jrobinso:
### Description

 If the sequences don't match there is a 100% certainty the reconstructed records will have an incorrect read sequence.   This is a very serious error, qualitatively a different level  than the typical "validation" errors that ValidationStringency settings control.

### DIFF
--- a/src/main/java/htsjdk/samtools/CRAMIterator.java
+++ b/src/main/java/htsjdk/samtools/CRAMIterator.java
@@ -177,10 +177,12 @@ public class CRAMIterator implements SAMRecordIterator {
             final Slice slice = container.slices[i];
             if (slice.sequenceId < 0)
                 continue;
-            if (validationStringency != ValidationStringency.SILENT && !slice.validateRefMD5(refs)) {
-                log.error(String
+            if (!slice.validateRefMD5(refs)) {
+                final String msg = String
                         .format("Reference sequence MD5 mismatch for slice: seq id %d, start %d, span %d, expected MD5 %s", slice.sequenceId,
-                                slice.alignmentStart, slice.alignmentSpan, String.format("%032x", new BigInteger(1, slice.refMD5))));
+                                slice.alignmentStart, slice.alignmentSpan, String.format("%032x", new BigInteger(1, slice.refMD5)));
+                log.error(msg);
+                throw new CRAMException(msg);
             }
         }
 
--- a/src/main/java/htsjdk/samtools/CRAMIterator.java
+++ b/src/main/java/htsjdk/samtools/CRAMIterator.java
@@ -181,7 +181,6 @@ public class CRAMIterator implements SAMRecordIterator {
                 final String msg = String
                         .format("Reference sequence MD5 mismatch for slice: seq id %d, start %d, span %d, expected MD5 %s", slice.sequenceId,
                                 slice.alignmentStart, slice.alignmentSpan, String.format("%032x", new BigInteger(1, slice.refMD5)));
-                log.error(msg);
                 throw new CRAMException(msg);
             }
         }


droazen:
@vadimzalunin please review

src/main/java/htsjdk/samtools/CRAMIterator.java
```diff
@@ -177,10 +177,12 @@ void nextContainer() throws IOException, IllegalArgumentException,
...
             if (slice.sequenceId < 0)
                 continue;
-            if (validationStringency != ValidationStringency.SILENT && !slice.validateRefMD5(refs)) {
-                log.error(String
+            if (!slice.validateRefMD5(refs)) {
+                final String msg = String
                         .format("Reference sequence MD5 mismatch for slice: seq id %d, start %d, span %d, expected MD5 %s", slice.sequenceId,
-                                slice.alignmentStart, slice.alignmentSpan, String.format("%032x", new BigInteger(1, slice.refMD5))));
+                                slice.alignmentStart, slice.alignmentSpan, String.format("%032x", new BigInteger(1, slice.refMD5)));
+                log.error(msg);
```
vadimzalunin: not needed given the exception below

vadimzalunin:
@jrobinso Please remove 184: log.error(msg);
Otherwise :+1:

cmnbroad:
I have mixed feeling about this - although I think the change itself is good, due to https://github.com/samtools/htsjdk/issues/506), we are not including any evidence in the file itself about what reference was used to create it. So basically, if you created this cram file with htsjdk, you'll need to have out-of-band evidence of what reference to use, or you will never be able to process it with htsjdk at all. Do we need an opt-out switch for this ?

vadimzalunin:
@cmnbroad I just realized what's missing: cramtools injects ref md5s into header if they are missing. In my head the same applied to htsjdk but apparently I was wrong.

jrobinso:
So its possible the ref (expected) MD5 is missing in a cram file?  If that's the case I suppose the test should be skipped.    That's concerning however.   I don't object to an opt-out but also don't think it should be based on ValidationStringency.   Many bam files are invalid in trivial un-important ways,  requiring that setting to be very lenient (SILENT in the case of IGV).   If an opt-out is included perhaps it should be explicit for opting out of cram file MD5 checks.

vadimzalunin:
I think CRAM should be explicitly strict about ref MD5 when ref-based compression is used. 
For odd cases we could introduce a recovery mode which would blindly trust any reference given but certainly not for normal use cases.

jrobinso:
I agree, but if CRAM is strict about ref MD5 I don't understand the need for an opt-out.   Decoding a record with the wrong reference is really bad because there is no indication anything is wrong.

vadimzalunin:
main difficulty with calculating md5s is to read all references from a fasta - there is no way to know in advance which reference is used by the following records.

cmnbroad:
@jrobinso There are two separate md5 calculations here. One is in the SAM header for the CRAM (thats the one thats missing per #506) that I believe has the md5 for each reference contig. The code you modified is validating against the md5 for a specific slice; AFAIK that will always be there.

I agree that its pretty sketchy to proceed with the wrong reference, but my take would be that we should be strict by default (independent of any validation stringency setting), and include a separate opt-out switch that allows you to bypass the guard if you need to.

jrobinso:
RE the broken unit test, htsjdk.samtools.util.AsyncBufferedIteratorTest.testBackgroundBlocks,  it passes from my local branch.

droazen:
@jrobinso Should this include a fix for https://github.com/samtools/htsjdk/issues/645 as well? It seems like we shouldn't merge this until https://github.com/samtools/htsjdk/issues/645 is resolved.

jrobinso:
@droazen Yes I think they should be fixed together.   Throwing the exception without #645 could make some valid CRAMS unreadable, or more precisely valid reference fastas unusable with CRAM files.

droazen:
@jrobinso Agreed -- would you be able to add the fix for https://github.com/samtools/htsjdk/issues/645 to this branch so that we can merge both together, since it's just a few additional lines?

jrobinso:
Yes, but I think we should step back and look at this again, briefly.    In review, the issue is around soft-masked references, i.e. references with lower case "actg".  The rules for computing MD5 for CRAM state that the reference should be upper-cased first.  So we can do the upper casing when computing MD5.   However, the reference byte array is also used to fill in the read sequence in the expanded SAM record.  I've never seen lower case actg in a read sequence, it seems weird.   So,  there's an argument for upper-casing the reference sequence when it is set (obtained from referencesSource).    Or further, making an upper case sequence part of the contract for ReferenceSource.

droazen:
Uppercasing at the `ReferenceSource` level sounds reasonable to me, but we should get feedback from others first. @yfarjoun @tfenne @vadimzalunin what do you think of this idea?

yfarjoun:
sound good to me....but would like to hear from @vadimzalunin
<https://github.com/vadimzalunin>

vadimzalunin:
Imo uppercasing should be part of ReferenceSource contract because it looks bases up by MD5. 

PS Ideally ReferenceSource should provide view objects that hold bases plus metadata like name, checksum and coordinates in case of subsequence. This was the plan with htsjdk.samtools.cram.ref.ReferenceRegion

In cramtools it's functionality has been expanded but I didn't have time to propagate this to htsjdk yet: 
https://github.com/enasequence/cramtools/blob/aba8c766acc75b3ef3042ba9de46724f8551639e/src/main/java/net/sf/cram/ref/ReferenceRegion.java

cmnbroad:
Should [base normalization](https://github.com/samtools/htsjdk/blob/8485225f488ac25b508f5876472d55a7844b5af7/src/main/java/htsjdk/samtools/cram/build/Utils.java#L30) (as in anything not ACGTN -> N) also be part of this ?

droazen:
@jrobinso Seems like there is consensus that uppercasing should become part of the contract for `ReferenceSource`, as you proposed.

jrobinso:
OK, the IGV reference source already upper-cases.   So I think this branch is complete.

jrobinso:
@cmnbroad  I think yes, since that's done when creating the CRAM.   It should be symmetrical, otherwise you can create a CRAM that cannot be read.   So this would also be part of the ReferenceSource contract.    (I think its rare to have anything other than atcgn in a fasta file).

droazen:
@jrobinso Don't we still need changes to `ReferenceSource` itself to reflect the new contract before we can call this branch complete?

jrobinso:
ReferenceSource is the concrete htsjdk implementation of CRAMReferenceSource,  so yes it is out of spec with the contract.   However, I would view that as a separate bug and enter a new issue for it.    I would rather not make changes to your concrete implementation,  IGV does not use ReferenceSource.

droazen:
Closing in favor of #671

PR #646, opened 2016-06-23
Fix self-suppression problem with exception handling in AbstractAsyncWriter

Lenbok:
Any exception thrown during synchronousWrite() by the underlying thread
is typically rethrown during write()'s invocation of checkAndRethrow().
However, subsequent calling of close() (e.g. by try-with-resources) results
in an attempt to throw the same exception instance while it is already 
being thrown, which triggers an additional exception from Java (Self-suppression 
not permitted).

This commit ensures that the underlying exception is only thrown once,
by clearing the reference to the Throwable before throwing.

### DIFF
--- a/src/test/java/htsjdk/samtools/util/AsyncWriterTest.java
+++ b/src/test/java/htsjdk/samtools/util/AsyncWriterTest.java
@@ -1,0 +1,77 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016 Len Trigg
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package htsjdk.samtools.util;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class AsyncWriterTest {
+    private static class MyException extends RuntimeException {
+        final Integer item;
+        public MyException(Integer item) {
+            this.item = item;
+        }
+    }
+    private static class TestAsyncWriter extends AbstractAsyncWriter<Integer> {
+        protected TestAsyncWriter() {
+            super(1); // Queue size of 1 to give us more control over the order of events
+        }
+
+        @Override
+        protected String getThreadNamePrefix() {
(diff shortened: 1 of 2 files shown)


src/test/java/htsjdk/samtools/util/AsyncWriterTest.java
```diff
@@ -0,0 +1,71 @@
...
+    public void testNoSelfSuppression() {
+        try {
+            try (TestAsyncWriter t = new TestAsyncWriter()) {
+                for (int i = 0; i < 1000; i++) {
+                    t.write(i);
+                }
+                Assert.fail("Expected exception");
+            }
+        } catch (Exception e) {
+            Assert.assertFalse(e.getMessage().contains("Self-suppression"));
```
droazen: This seems redundant, and bad practice to assert based on the text of exception message. The `e instanceof MyException` assert below should be sufficient.

src/test/java/htsjdk/samtools/util/AsyncWriterTest.java
```diff
@@ -0,0 +1,71 @@
...
+        @Override
+        protected void synchronouslyClose() {
+            // Nothing
+        }
+    }
+    @Test
+    public void testNoSelfSuppression() {
+        try {
+            try (TestAsyncWriter t = new TestAsyncWriter()) {
+                for (int i = 0; i < 1000; i++) {
```
droazen: Is this test timing-dependent? Could it fail intermittently in the future?

droazen:
Two comments, back to @Lenbok -- looks good, should be able to merge once comments are addressed. Please rebase and squash into a single commit after addressing comments.

Lenbok:
@droazen Thanks for the review comments, an updated version has been pushed.

vadimzalunin:
A suggestion is in checkAndRethrow() method to throw 
`new RuntimeException(t.get())` or `new Error(t.get())`

This way JVM will not suppress exceptions and the error state of the writer instance is preserved. 

It is a caller's responsibility to ensure errors are reported only once. The contract basically should be that writer becomes nonoperational if an exception occurred. Any subsequent calls to write() or close() methods will throw an exception.

Also I would consider getError() method to return the original exception object if needed.

Lenbok:
I'm really not a fan of adding uninformative layers of exception wrapping that make it even harder for the (potentially non-immediate) caller to take appropriate action. (As a side note, the else block currently containing an exception wrap can never be triggered, as checked exceptions can't occur in the WriterRunnable).

Having to jump through a getError() hoop would make the difference between async and non-async writers needlessly larger from a caller perspective.

Regarding your concern that a caller might continue to use the instance after the exception has been thrown (and presumably ignored by the caller), checkAndRethrow() could call this.isClosed.set(true) prior prior to the throw, so that any subsequent write attempts will cause a (new) exception.

vadimzalunin:
@Lenbok What might appropriate actions be? Perhaps I don't understand the idea behind this. 

We don't know how the API will be used. My concern was that async writer can clear it's error state as if it has magically recovered. Imo writer should preserve the error state somehow because it is implicit.

Lenbok:
@vadimzalunin In our case the primary distinctions are for causes that the end-user of the should be able to do something about since the problem is environmental (IO exceptions, SAM/VCF format parsing problems - essentially those things which are typically checked exceptions), vs errors that indicate a programmatic problem has occurred and these may get automatically reported to us. These are presented differently to the user and depending on the code, different decisions made as to whether the execution should continue. Surely this isn't a foreign concept?

In my opinion the pull request is sufficient as is to support any use case where the caller is behaving appropriately w.r.t exceptions (i.e. do not attempt to use the instance after the exception has been thrown), so there is very little extra utility in explicitly preserving error state of the writer.

vadimzalunin:
@Lenbok I see. The exception analysis then may not work on some implementations of AbstractAsyncWriter but this is not about this PR. 

... I just noticed the NOTE on "single thread only" in the class javadoc and this kinda invalidates my comments :D 

It would help to describe the new behaviour in javadocs.

Lenbok:
@vadimzalunin I have updated the PR with a couple of changes. Firstly, as you suggested, I just  inserted an extra note in the class javadoc describing the behaviour.

The other change is to mark the writer as closed once the exception is being thrown. While this does have the nice side effect of ensuring that any attempts by the caller to write() will throw, the main reason is to ensure that the *first* exception produced by the underlying writer is the one that gets thrown. Without this, there is a race condition where WriterRunnable sets the exception, write() throws the exception, close() is invoked while this exception is in play but the queue.clear() back in WriterRunnable has not yet been executed, and this would trigger close to directly call synchronouslyWrite() (which would typically trigger another exception, depending on the underlying writer behaviour).

vadimzalunin:
:+1:

PR #647, opened 2016-06-23
Fixed bug where SamPairUtil.setMateInfo() would omit the MQ tag if on…

tfenne:
### Description

Fix a small but annoying bug that was causing unmapped reads with mapped mates to get most mate information set _except_ `MQ` which is set in all other cases.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

…e read was mapped and the other unmapped.

### DIFF
--- a/src/main/java/htsjdk/samtools/SamPairUtil.java
+++ b/src/main/java/htsjdk/samtools/SamPairUtil.java
@@ -244,7 +244,7 @@ public class SamPairUtil {
             mapped.setMateAlignmentStart(unmapped.getAlignmentStart());
             mapped.setMateNegativeStrandFlag(unmapped.getReadNegativeStrandFlag());
             mapped.setMateUnmappedFlag(true);
-            // For the mapped read, set it's mateCigar to null, since the other read must be unmapped
+            mapped.setAttribute(SAMTag.MQ.name(), null);
             mapped.setAttribute(SAMTag.MC.name(), null);
             mapped.setInferredInsertSize(0);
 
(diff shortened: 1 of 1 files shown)


tfenne:
@ktibbett Want to take a quick look at this?  This will effect `FixMateInformation` and `MergeBamAlignment` results, but in a good way.

src/main/java/htsjdk/samtools/SamPairUtil.java
```diff
@@ -246,12 +246,14 @@ else if (rec1.getReadUnmappedFlag() && rec2.getReadUnmappedFlag()) {
             mapped.setMateUnmappedFlag(true);
             // For the mapped read, set it's mateCigar to null, since the other read must be unmapped
             mapped.setAttribute(SAMTag.MC.name(), null);
+            mapped.setAttribute(SAMTag.MQ.name(), null);
```
nh13: move to before setting the `MC` since that's the order elsewhere: it makes it easier to find.

tfenne:
@yfarjoun Want to return the favor and take a look here since @ktibbett seems busy?

src/main/java/htsjdk/samtools/SamPairUtil.java
```diff
@@ -246,12 +246,14 @@ else if (rec1.getReadUnmappedFlag() && rec2.getReadUnmappedFlag()) {
...
             mapped.setAttribute(SAMTag.MC.name(), null);
+            mapped.setAttribute(SAMTag.MQ.name(), null);
             mapped.setInferredInsertSize(0);
 
             unmapped.setMateReferenceIndex(mapped.getReferenceIndex());
             unmapped.setMateAlignmentStart(mapped.getAlignmentStart());
             unmapped.setMateNegativeStrandFlag(mapped.getReadNegativeStrandFlag());
             unmapped.setMateUnmappedFlag(false);
+            unmapped.setAttribute(SAMTag.MQ.name(), mapped.getMappingQuality());
             // For the unmapped read, set it's mateCigar to the mate's Cigar, since the mate must be mapped
```
yfarjoun: I know it was like that before...can you remove the `'` from "it's"

yfarjoun:
👍 though I protest that this class lacks tests...we should at least open a ticket to record this fact...

tfenne:
Updates applied.  @yfarjoun If we logged issues for everything that needed testing... we'd have a lot of issues 😞 .

yfarjoun:
👍

PR #650, opened 2016-06-24
Make unzipBlock in BlockGunzipper public.

nh13:
This is a simple change to expose the `unzipBlock` method, and is useful for folks that are de-compressing gzip blocks, for example specialty tools built for BAM files.

@tfenne @akiezun care to take a look?


tfenne:
Seems silly that the class was public but that the main method on it was not.  Merging!

PR #654, opened 2016-06-30
Number tag in header line for GENOTYPE_FILTER

SHuang-Broad:
### Description

User [reports] (https://github.com/broadinstitute/gsa-unstable/issues/1413) that for `VariantFiltration` in GATK3 with genotype filters, the Number tag in the description of FT in the output VCF should be "." rather than 1.
This is actually caused by htsjdk.
This fix tries to deal with that.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


SHuang-Broad:
@lbergelson would you please review? This should be very quick (and similar to [this](https://github.com/samtools/htsjdk/pull/546)).

lbergelson:
👍 Looks good to me.

droazen:
No test?

SHuang-Broad:
Will add one (and maybe more considering this is the second time such an error has been caught) over the weekend.

PR #656, opened 2016-07-07
As per request https://github.com/broadinstitute/picard/issues/519 th…

kasont:
### Description

Please explain the changes you made here.
Explain the **motivation** for making this change. What existing problem does the pull request solve?

**As per https://github.com/broadinstitute/picard/issues/519 wants, the output is changed from exception to null.**

### Checklist

- [ ] Code compiles correctly
- [ ] New tests covering changes and new functionality
- [ ] All tests passing
- [ ] Extended the README / documentation, if necessary
- [ ] Is not backward compatible (breaks binary or source compatibility)

…e output is null.

PR #657, opened 2016-07-08
RE:As per request https://github.com/broadinstitute/picard/issues/519…

kasont:
Fix for  https://github.com/broadinstitute/picard/issues/519.

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/VariantContext.java
+++ b/src/main/java/htsjdk/variant/variantcontext/VariantContext.java
@@ -1701,7 +1701,7 @@ public class VariantContext implements Feature, Serializable {
         return getAlternateAlleles().stream()
                 .map(allele -> new Tuple<>(allele, getCalledChrCount(allele)))
                 .max((alleleAndCount1, alleleAndCount2) -> Integer.compare(alleleAndCount1.b, alleleAndCount2.b))
-                .get()
+                .orElse(null)
                 .a;
     }
 


yfarjoun:
can you add a test to show that this works?

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1701,7 +1701,7 @@ public Allele getAltAlleleWithHighestAlleleCount() {
         return getAlternateAlleles().stream()
                 .map(allele -> new Tuple<>(allele, getCalledChrCount(allele)))
                 .max((alleleAndCount1, alleleAndCount2) -> Integer.compare(alleleAndCount1.b, alleleAndCount2.b))
-                .get()
+                .orElse(null)
                 .a;
```
droazen: Potential NullPointerException here, as mentioned in comments on your closed PR https://github.com/samtools/htsjdk/pull/656

droazen: @kasont Do you have time to fix this? The problem is that `orElse(null)` can return null, and you're dereferencing the result of that expression directly via the `.a`. You need to protect against dereference of null, and have the method return null instead.

droazen:
You also need to add a unit test proving that the original issue is fixed.

droazen:
Back to @kasont for changes.

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
lbergelson: I think something like this may be a more direct way to get the value you want:

```
     return getAlternateAlleles().stream()
                .max(Comparator.comparing(this::getCalledChrCount))
                .orElse(null);
```

yfarjoun:
What is this waiting for @kasont ??

yfarjoun:
closing in favor of #733

PR #663, opened 2016-07-13
Genotype attributes

bbimber:
### Description

A Genotype always has filters. If there are no non-PASS filters, it's PASS. Genotype.getAnyAttribute() currently will never return a value for "FT", which I dont think is the right behavior.  This also means you cant use GATK's JEXL select/filtering based on the genotype's filters. This is a minor change to Genotype to support this.  It includes a simple unit test.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

…ibute

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/Genotype.java
+++ b/src/main/java/htsjdk/variant/variantcontext/Genotype.java
@@ -528,7 +528,7 @@ public abstract class Genotype implements Comparable<Genotype>, Serializable {
     }
 
     /**
-     * A totally generic getter, that allows you to specific keys that correspond
+     * A totally generic getter, that allows you to get specific keys that correspond
      * to even inline values (GQ, for example).  Can be very expensive.  Additionally,
      * all <code>int[]</code> are converted inline into <code>List&lt;Integer&gt;</code> for convenience.
      *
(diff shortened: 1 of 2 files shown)


droazen:
@ldgauthier Can you review?

src/main/java/htsjdk/variant/variantcontext/Genotype.java
```diff
@@ -556,6 +556,8 @@ public Object getAnyAttribute(final String key) {
             return Collections.EMPTY_LIST;
```
ldgauthier: While you're here, can you fix the javadoc for this method?  There's a verb missing.  "allows you to [what?] specific keys"

ldgauthier:
Good catch @bbimber -- thanks!  We've only recently started dabbling in genotype filters and hadn't run into this yet.

Fix looks good to me.  Thanks for the tests.  Just one easy request for docs.

src/main/java/htsjdk/variant/variantcontext/Genotype.java
bbimber: i suppose the getter allows you to get specific keys?

ldgauthier: Seems reasonable.  :)

ldgauthier:
:+1:  Please address the remaining items in the checklist as necessary.

bbimber:
was there anything specific you think is missing on the checklist that needs to be addressed?  i added passing tests, and this is just a bugfix to existing functionality, so i dont think the others apply.

ldgauthier:
Sounds good.  I don't see a check next to "passing tests" so I wasn't sure.  Do we rebase and squash in htsjdk?

bbimber:
i guess I assumed travis-ci ran the tests?

bbimber:
That sounds good.  Do you need me to squash, or does github let you do this at time of merge?

ldgauthier:
If you could rebase and squash that would be great.

bbimber:
ok, should be all set.

PR #668, opened 2016-07-24
Fix JexlMap and GenotypeJEXLContext NullPtrException

SHuang-Broad:
### Description

* `GenotypeJEXLContext.get()` currently potentially throws `NullPointerException`, triggered by a design choice in `Genotype.getFilters()`, which explicitly returns `null` (genotype considered to be filter passing in such case according to documentation). This fix temporarily fixes that exception (1st commit, by Louis).

* Class `CommonInfo`, which holds "Common utility routines for VariantContext and Genotype", contains such a constant 
        
        private static Set<String> NO_FILTERS = Collections.emptySet();

    `VariantContext` uses this constant in its own `getFilters()` (it has a separate method `getFiltersMaybeNull()`), but `Genotype` doesn't use such constant. A choice has to be made to whether make `Genotype` resemble `VariantContext`'s behavior (a backward non-compatible change) or keep things as they are.

* Some code re-organizing and whole bunch of todo's in `JEXLMap`, which must be resolved in the review rounds.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing
- [PartiallyDone] Extended the README / documentation, if necessary
- [Possibly] Is not backward compatible (breaks binary or source compatibility)

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
+++ b/src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
@@ -17,12 +17,12 @@ import java.util.Set;
  */
 
 class JEXLMap implements Map<JexlVCMatchExp, Boolean> {
-
-    // our context
-    private JexlContext jContext = null;
     // our variant context and/or Genotype
     private final VariantContext vc;
     private final Genotype g;
+
+    // our context
+    private JexlContext jContext = null;
 
     /**
      * our mapping from {@link JexlVCMatchExp} to {@link Boolean}s, which will be set to {@code NULL}
@@ -47,12 +47,14 @@ class JEXLMap implements Map<JexlVCMatchExp, Boolean> {
      *                                  when any of the JexlVCMatchExp (i.e. keys) contains invalid Jexl expressions.
      */
     public Boolean get(Object o) {
-        if(o == null){
+        if (o == null) {
             throw new IllegalArgumentException("Query key is null");
         }
 
         // if we've already determined the value, return it
-        if (jexl.containsKey(o) && jexl.get(o) != null) return jexl.get(o);
+        if (jexl.containsKey(o) && jexl.get(o) != null) {
+            return jexl.get(o);
+        }
 
         // otherwise cast the expression and try again
         final JexlVCMatchExp e = (JexlVCMatchExp) o;
@@ -131,7 +133,7 @@ class JEXLMap implements Map<JexlVCMatchExp, Boolean> {
      */
     private void evaluateExpression(final JexlVCMatchExp exp) {
         // if the context is null, we need to create it to evaluate the JEXL expression
-        if (this.jContext == null){
+        if (this.jContext == null) {
             createContext();
         }
 
@@ -139,13 +141,12 @@ class JEXLMap implements Map<JexlVCMatchExp, Boolean> {
             final Boolean value = (Boolean) exp.exp.evaluate(jContext);
             // treat errors as no match
             jexl.put(exp, value == null ? false : value);
-        } catch (final JexlException.Variable e){
+        } catch (final JexlException.Variable e) {
             // if exception happens because variable is undefined (i.e. field in expression is not present), evaluate to FALSE
             jexl.put(exp,false);
         } catch (final JexlException e) {
             // todo - might be better if no exception is caught here but let's user decide how to deal with them; note this will propagate to get() and values()
-            throw new IllegalArgumentException(String.format("Invalid JEXL expression detected for %s with message %s",
-                    exp.name, e.getMessage()));
+            throw new IllegalArgumentException(String.format("Invalid JEXL expression detected for %s", exp.name), e);
         }
     }
 
@@ -154,13 +155,11 @@ class JEXLMap implements Map<JexlVCMatchExp, Boolean> {
      * This code is where new JEXL context variables should get added.
      */
     private void createContext() {
-        if ( vc == null ) {
+        if (vc == null) {
             jContext = new MapContext(Collections.emptyMap());
-        }
-        else if (g == null) {
+        } else if (g == null) {
             jContext = new VariantJEXLContext(vc);
-        }
-        else {
+        } else {
             jContext = new GenotypeJEXLContext(vc, g);
         }
     }
--- a/src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
+++ b/src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
@@ -54,6 +54,7 @@ class JEXLMap implements Map<JexlVCMatchExp, Boolean> {
     }
 
     /**
+     * Note: due to laziness, this accessor actually modifies the instance by possibly forcing evaluation of an Jexl expression.
      * @throws IllegalArgumentException when {@code o} is {@code null}
      */
     public Boolean get(Object o) {
@@ -87,6 +88,7 @@ class JEXLMap implements Map<JexlVCMatchExp, Boolean> {
      * This is fine if you truly want all the keys, but if you only want a portion, or  know
      * the keys you want, you would be better off using get() to get them by name.
      *
+     * Note: due to laziness, this accessor actually modifies the instance by possibly forcing evaluation of an Jexl expression.
      * @return a collection of boolean values, representing the results of all the variants evaluated
      *
      * @throws IllegalArgumentException
--- a/src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
+++ b/src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
@@ -14,16 +14,6 @@ import java.util.Set;
 /**
  * This is an implementation of a Map of {@link JexlVCMatchExp} to true or false values.
  * It lazily initializes each value as requested to save as much processing time as possible.
- *
- * TODO: must resolve the following outdated comment in PR review rounds
- * Compatible with JEXL 1.1 (this code will be easier if we move to 2.0, all of the functionality can go into the
- * JexlContext's get()
- *
- * TODO: there is some troubling design choices in this class: the laziness and exception behavior:
- * TODO:   because the class is designed to be lazy, when an instance is created, some invalid expressions may be provided
- * TODO:   later when user requests some/all values, those invalid expressions will result in IllegalArgumentException,
- * TODO:   especially when calling the function values(), which is the last thing users expect when calling such accessors.
- * TODO:   but there may be no good choice here. could decide to "unsupport" the values() function.
  */
 
 class JEXLMap implements Map<JexlVCMatchExp, Boolean> {
@@ -40,11 +30,8 @@ class JEXLMap implements Map<JexlVCMatchExp, Boolean> {
      */
     private Map<JexlVCMatchExp,Boolean> jexl;
 
-    // -----------------------------------------------------------------------------------------------
-    // Initializers and accessors
-    // -----------------------------------------------------------------------------------------------
     public JEXLMap(final Collection<JexlVCMatchExp> jexlCollection, final VariantContext vc, final Genotype g) {
-        lazyInitialize(jexlCollection);
+        initialize(jexlCollection);
         this.vc = vc;
         this.g = g;
     }
@@ -55,7 +42,9 @@ class JEXLMap implements Map<JexlVCMatchExp, Boolean> {
 
     /**
      * Note: due to laziness, this accessor actually modifies the instance by possibly forcing evaluation of an Jexl expression.
-     * @throws IllegalArgumentException when {@code o} is {@code null}
+     *
+     * @throws IllegalArgumentException when {@code o} is {@code null} or
+     *                                  when any of the JexlVCMatchExp (i.e. keys) contains invalid Jexl expressions.
      */
     public Boolean get(Object o) {
         if(o == null){
@@ -89,9 +78,10 @@ class JEXLMap implements Map<JexlVCMatchExp, Boolean> {
      * the keys you want, you would be better off using get() to get them by name.
      *
      * Note: due to laziness, this accessor actually modifies the instance by possibly forcing evaluation of an Jexl expression.
+     *
      * @return a collection of boolean values, representing the results of all the variants evaluated
      *
-     * @throws IllegalArgumentException
+     * @throws IllegalArgumentException when any of the JexlVCMatchExp (i.e. keys) contains invalid Jexl expressions.
      */
     public Collection<Boolean> values() {
         for (final JexlVCMatchExp exp : jexl.keySet()) {
@@ -103,18 +93,13 @@ class JEXLMap implements Map<JexlVCMatchExp, Boolean> {
     }
 
     /**
-     * TODO: the number may count invalid entries, is this good?
-     * @return the number of keys, i.e. {@link JexlVCMatchExp}'s hold by this mapping.
+     * @return the number of keys, i.e. {@link JexlVCMatchExp}'s held by this mapping.
      */
     public int size() {
         return jexl.size();
     }
 
     public boolean isEmpty() { return this.jexl.isEmpty(); }
-
-    // -----------------------------------------------------------------------------------------------
-    // Modifiers
-    // -----------------------------------------------------------------------------------------------
 
     public Boolean put(JexlVCMatchExp jexlVCMatchExp, Boolean aBoolean) {
         return jexl.put(jexlVCMatchExp, aBoolean);
@@ -124,15 +109,11 @@ class JEXLMap implements Map<JexlVCMatchExp, Boolean> {
         jexl.putAll(map);
     }
 
-    // -----------------------------------------------------------------------------------------------
-    // Utilities
-    // -----------------------------------------------------------------------------------------------
-
     /**
-     * Lazily initializes {@link #jexl}, in the sense that all keys in {@code jexlCollection} are associated with {@code null}.
-     * @param jexlCollection
+     * Initializes all keys with null values indicating that they have not yet been evaluated.
+     * The actual value will be computed only when the key is requested via {@link #get(Object)} or {@link #values()}.
      */
-    private void lazyInitialize(Collection<JexlVCMatchExp> jexlCollection) {
+    private void initialize(Collection<JexlVCMatchExp> jexlCollection) {
         jexl = new HashMap<>();
         for (final JexlVCMatchExp exp: jexlCollection) {
             jexl.put(exp, null);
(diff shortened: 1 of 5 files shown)


SHuang-Broad:
@lbergelson , would you mind open discussions on all these?
I don't think this is something that I should decide on my own.

SHuang-Broad:
Sorry @lbergelson , I accidentally pushed another improvements that I'm working on to this branch. Please ignore the third commit first.
Sorry.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,152 +12,188 @@
...
- * this is an implementation of a Map of JexlVCMatchExp to true or false values.  It lazy initializes each value
- * as requested to save as much processing time as possible.
+ * This is an implementation of a Map of {@link JexlVCMatchExp} to true or false values.
+ * It lazily initializes each value as requested to save as much processing time as possible.
  *
+ * TODO: must resolve the following outdated comment in PR review rounds
  * Compatible with JEXL 1.1 (this code will be easier if we move to 2.0, all of the functionality can go into the
  * JexlContext's get()
  *
+ * TODO: there is some troubling design choices in this class: the laziness and exception behavior:
```
lbergelson: @SHuang-Broad I think it would be better to open an issue with this comment instead of putting it into the code.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,152 +12,188 @@
...
  */
 
 class JEXLMap implements Map<JexlVCMatchExp, Boolean> {
-    // our variant context and/or Genotype
-    private final VariantContext vc;
-    private final Genotype g;
 
     // our context
     private JexlContext jContext = null;
+    // our variant context and/or Genotype
```
lbergelson: Why move these down like this?  Not a big deal either way but I don't see the reason to change them.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,152 +12,188 @@
...
-    // our mapping from JEXLVCMatchExp to Booleans, which will be set to NULL for previously uncached JexlVCMatchExp
+    /**
+     * our mapping from {@link JexlVCMatchExp} to {@link Boolean}s, which will be set to {@code NULL}
+     * for previously un-cached {@link JexlVCMatchExp}.
+     */
     private Map<JexlVCMatchExp,Boolean> jexl;
 
-
-    public JEXLMap(Collection<JexlVCMatchExp> jexlCollection, VariantContext vc, Genotype g) {
+    // -----------------------------------------------------------------------------------------------
```
lbergelson: We don't typically have these sorts of big signpost comments.  It seems useful for dividing the unsupported operations from the rest, but I'm not sure you need tthem for the rest of the class.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: Yes, I just wasn't sure if people would care.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,152 +12,188 @@
...
-            jexl.put(exp, null);
-        }
-    }
-
     /**
-     * create the internal JexlContext, only when required.  This code is where new JEXL context variables
-     * should get added.
-     *
+     * Note: due to laziness, this accessor actually modifies the instance by possibly forcing evaluation of an Jexl expression.
+     * @throws IllegalArgumentException when {@code o} is {@code null}
```
lbergelson: Does this also throw IllegalArgumentException in other cases like you mentioned in the todo?

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,152 +12,188 @@
...
-     * been evaluated yet.  This is fine if you truely want all the keys, but if you only want a portion, or  know
+     * Get all the values of the map, i.e. the {@link Boolean} values.
+     * This is an expensive call, since it evaluates all keys that haven't been evaluated yet.
+     * This is fine if you truly want all the keys, but if you only want a portion, or  know
      * the keys you want, you would be better off using get() to get them by name.
+     *
+     * Note: due to laziness, this accessor actually modifies the instance by possibly forcing evaluation of an Jexl expression.
      * @return a collection of boolean values, representing the results of all the variants evaluated
+     *
+     * @throws IllegalArgumentException
```
lbergelson: Can you describe when it throws IllegalArgumentException, which is strange since it takes no arguments.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,152 +12,188 @@
...
+            }
+        }
         return jexl.values();
     }
 
     /**
-     * evaulate a JexlVCMatchExp's expression, given the current context (and setup the context if it's null)
-     * @param exp the JexlVCMatchExp to evaluate
+     * TODO: the number may count invalid entries, is this good?
+     * @return the number of keys, i.e. {@link JexlVCMatchExp}'s hold by this mapping.
```
lbergelson: hold -> held

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,152 +12,188 @@
...
                 evaluateExpression(exp);
+            }
+        }
         return jexl.values();
     }
 
     /**
-     * evaulate a JexlVCMatchExp's expression, given the current context (and setup the context if it's null)
-     * @param exp the JexlVCMatchExp to evaluate
+     * TODO: the number may count invalid entries, is this good?
```
lbergelson: How can you know if they're invalid? By evaluating them and seeing if they throw?

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: For me personally, it serves the purpose of grouping methods providing similar functionalities in proximity, so that I would jump around to see how one function works. The second reason that I put in these "block segregators" is such that the interfaces exposed and the implementations can be better separated with a quick glance.

If you think this clusters the class instead of helping, I'll remove them.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: Will change back, thanks!

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,152 +12,188 @@
...
+
+    // -----------------------------------------------------------------------------------------------
+    // Utilities
+    // -----------------------------------------------------------------------------------------------
+
+    /**
+     * Lazily initializes {@link #jexl}, in the sense that all keys in {@code jexlCollection} are associated with {@code null}.
+     * @param jexlCollection
+     */
+    private void lazyInitialize(Collection<JexlVCMatchExp> jexlCollection) {
```
lbergelson: I'm not sure this is any clearer than just initialize. Comment might read better as 
"Initializes all keys with null values indicating that they have not yet been evaluated.  The actual value will be computed only when the key is requested"

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,152 +12,188 @@
...
+                    exp.name,
+                        /*(e.getMessage() == null ? */"no message" /*: e.getMessage())*/));
+        }
+    }
+
+    /**
+     * Create the internal JexlContext, only when required.
+     * This code is where new JEXL context variables should get added.
+     */
+    private void createContext() {
```
lbergelson: I think this was clearer with three distinct cases instead of collapsing one of them.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: This echoes the "todo" comment at the beginning of this class. There might be no way of knowing if they invalid unless they are evaluated.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,152 +12,188 @@
...
 
     /**
-     * helper function: adds the list of attributes to the information map we're building
+     * TODO: function is not used, remove or not? resolve in PR review rounds.
+     * adds the list of attributes to the information map we're building
      * @param infoMap the map
      * @param attributes the attributes
      */
-    private static void addAttributesToMap(Map<String, Object> infoMap, Map<String, ?> attributes ) {
+    private static void addAttributesToMap(final Map<String, Object> infoMap, final Map<String, ?> attributes ) {
```
lbergelson: delete this one if it's unused, it's private so nothing else could be using it.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,152 +12,188 @@
...
-        return jexl.put(jexlVCMatchExp,aBoolean);
-    }
-
-    public void putAll(Map<? extends JexlVCMatchExp, ? extends Boolean> map) {
-        jexl.putAll(map);
-    }
-
     // //////////////////////////////////////////////////////////////////////////////////////
+    // TODO: is the following comment still true? must resolve in PR review rounds
+    // TODO: is not supported, any other way to resolve at compile time rather than blow up in your face?
```
lbergelson: These are fine I think.  There isn't a good way to convert these to a compile errors since java doesn't have separate interfaces for immutable collections.  It's pretty standard to implement some collection functions to just throw `UnsupportedOperationException`

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,152 +12,188 @@
...
+        if (this.jContext == null){
+            createContext();
+        }
+
         try {
             final Boolean value = (Boolean) exp.exp.evaluate(jContext);
             // treat errors as no match
             jexl.put(exp, value == null ? false : value);
-        } catch (Exception e) {
+        } catch (final JexlException e) {
```
lbergelson: In the previous version this handled exceptions that had "undefined variable" differently than other exceptions.  I think you should have a separate catch block for `JexlException.Variable` that does `jexl.put(exp,false)` and then the catch for generic `JexlException` that get translated to `IllegalArgumentException`

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: Done, see #686.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: Removed, but kept the structure of "`public` first, `private` later".

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: Done.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: Done.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: Done

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: Done.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: Done.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: Done.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: Got it. Thanks!

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: Good point! Done.

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,153 +12,161 @@
...
             // if exception happens because variable is undefined (i.e. field in expression is not present), evaluate to FALSE
-            // todo - might be safer if we explicitly checked for an exception type, but Apache's API doesn't seem to have that ability
-            if (e.getMessage() != null && e.getMessage().contains("undefined variable"))
-                jexl.put(exp,false);
-            else
-                throw new IllegalArgumentException(String.format("Invalid JEXL expression detected for %s with message %s", exp.name, (e.getMessage() == null ? "no message" : e.getMessage())));
+            jexl.put(exp,false);
+        } catch (final JexlException e) {
+            // todo - might be better if no exception is caught here but let's user decide how to deal with them; note this will propagate to get() and values()
+            throw new IllegalArgumentException(String.format("Invalid JEXL expression detected for %s with message %s",
```
lbergelson: could you replace this throw with 

```
throw new IllegalArgumentexception(String.format("Invalid JEXL expression detected for %s", exp.name), e)
```

 so that the underlying cause is propagated more cleanly?

SHuang-Broad: done

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,153 +12,161 @@
...
+     * Create the internal JexlContext, only when required.
+     * This code is where new JEXL context variables should get added.
      */
-    private static void addAttributesToMap(Map<String, Object> infoMap, Map<String, ?> attributes ) {
-        for (Entry<String, ?> e : attributes.entrySet()) {
-            infoMap.put(e.getKey(), String.valueOf(e.getValue()));
+    private void createContext() {
+        if ( vc == null ) {
+            jContext = new MapContext(Collections.emptyMap());
+        }
```
yfarjoun: we like to use 
```
} else if {
```
(here and elsewhere)

yfarjoun: and 
```
} else {
```

src/main/java/htsjdk/variant/variantcontext/VariantContextUtils.java
```diff
@@ -212,8 +212,12 @@ public final static VCFCompoundHeaderLine getMetaDataForField(final VCFHeader he
          * Create a new matcher expression with name and JEXL expression exp
          * @param name name
          * @param exp  expression
+         * @throws IllegalArgumentException if either argument is {@code null}
          */
         public JexlVCMatchExp(String name, Expression exp) {
+            if(name==null){ throw new IllegalArgumentException("Cannot create JexlVCMatchExp with null name."); }
```
yfarjoun: spaces:
```
if␣(name␣==␣null)␣{....
```

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,153 +12,161 @@
...
-        if ( vc == null ) {
-            jContext = new MapContext(Collections.emptyMap());
-        }
-        else if (g == null) {
-            jContext = new VariantJEXLContext(vc);
-        }
-        else {
-            jContext = new GenotypeJEXLContext(vc, g);
+    public Boolean get(Object o) {
+        if(o == null){
```
yfarjoun: spaces here too

lbergelson:
@SHuang-Broad Can you add a unit test that used to fail with nullptr?

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,153 +12,161 @@
...
+     *
+     * @throws IllegalArgumentException when {@code exp} is {@code null}, or
+     *                                  when the Jexl expression in {@code exp} fails to evaluate the JexlContext
+     *                                  constructed with the input VC or genotype.
      */
-    private void evaluateExpression(JexlVCMatchExp exp) {
+    private void evaluateExpression(final JexlVCMatchExp exp) {
         // if the context is null, we need to create it to evaluate the JEXL expression
-        if (this.jContext == null) createContext();
+        if (this.jContext == null){
```
yfarjoun: space before brace

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
```diff
@@ -11,153 +12,161 @@
...
+            createContext();
+        }
+
         try {
             final Boolean value = (Boolean) exp.exp.evaluate(jContext);
             // treat errors as no match
             jexl.put(exp, value == null ? false : value);
-        } catch (Exception e) {
+        } catch (final JexlException.Variable e){
```
yfarjoun: space

SHuang-Broad: done

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: done

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: done

src/main/java/htsjdk/variant/variantcontext/VariantContextUtils.java
SHuang-Broad: done

src/main/java/htsjdk/variant/variantcontext/JEXLMap.java
SHuang-Broad: done

SHuang-Broad:
@lbergelson , yes, will add a test.

SHuang-Broad:
There you go, @lbergelson , the unit test were cleaned up a little and more tests are added.
Thanks!

SHuang-Broad:
Friendly poke on @lbergelson, is the unit test covering what's needed?

lbergelson:
@SHuang-Broad This looks good.  Sorry for slowness. 👍

SHuang-Broad:
Thank you very much! @lbergelson

PR #672, opened 2016-07-27
Fix #653

### LINKED ISSUE #653: TribbleIndexedFeatureReader cannot read header from .bgz files
`TribbleIndexedFeatureReader.isGZIPPath` only recognizes paths ending in _.gz_, so block gzip-compressed files are not picked up.

The fix is to check for any of the extensions defined in `AbstractFeatureReader.BLOCK_COMPRESSED_EXTENSIONS` rather than just _.gz_.

magicDGS:
Simple patch as suggested in #653. `TribbleIndexedFeatureReader.isGZIPPath()` is not used anymore and it could be removed.


src/main/java/htsjdk/tribble/TribbleIndexedFeatureReader.java
```diff
@@ -217,7 +217,7 @@ private void readHeader() throws IOException {
         PositionalBufferedStream pbs = null;
         try {
             is = ParsingUtils.openInputStream(path);
-            if (isGZIPPath(path)) {
+            if (hasBlockCompressedExtension(path) || hasBlockCompressedExtension(new URI(path))) {
```
yfarjoun: it seems that the first part of if statement is only good for improving performance, but this isn't a performance bottleneck. is it needed for anything else?

magicDGS: I don't think so, so I will change it. I was following the same approach as in `TribbleIndexedFeatureReader.readHeader()`. I can change it also there or even better, modify the `AbstractFeatureReader.hasBlockCompressedExtension (final File file)` to be something like:

```java
public static boolean hasBlockCompressedExtension (final File file) {
    return hasBlockCompressedExtension(new URI(file));
}
``

**Comment by magicDGS** [comment]
I addressed your comment and added other stuff:

* Bgzipped test file (__.bgz__ extension) + test 
* Fixing `WFIterator` (I forgot to do it in the previous commit)
* Deprecation of `isGZIPPath(final String path)`

Because the deprecated method is not used anymore, should I remove it and get rid of the tests too?

Back to you, @yfarjoun!

File: src/main/java/htsjdk/tribble/TribbleIndexedFeatureReader.java [hunk 2]
```diff
@@ -274,6 +275,7 @@ private void readHeader() throws IOException {
     }
 
     //Visible for testing
+    @Deprecated
```
yfarjoun: if deprecating, please comment what should be used instead.

magicDGS: Done, but because it is package specific I wasn't sure if it will be better to remove...

magicDGS:
@yfarjoun, I added the comment for the deprecated method.

yfarjoun:
💯

PR #678, opened 2016-08-04
Removed artificial cap of 99 on GQ from older VCF versions.

hwangy:
### Description

Removes the GQ limit of 99 for VCF files. This was a limitation posed by older standards, which is no longer the case (page 6, <a href="http://samtools.github.io/hts-specs/VCFv4.2.pdf">VCF Spec</a>).

This fixes the problem where importing then exporting a VCF with GQ values greater than 99 results in a different VCF file.
### Checklist
- [X] Code compiles correctly
- [X] New tests covering changes and new functionality (<b>no new tests</b>)
- [X] All tests passing

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/writer/IntGenotypeFieldAccessors.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/IntGenotypeFieldAccessors.java
@@ -80,7 +80,7 @@ public class IntGenotypeFieldAccessors {
     }
 
     public static class GQAccessor extends AtomicAccessor {
-        @Override public int getValue(final Genotype g) { return Math.min(g.getGQ(), VCFConstants.MAX_GENOTYPE_QUAL); }
+        @Override public int getValue(final Genotype g) { return g.getGQ(); }
     }
 
     public static class DPAccessor extends AtomicAccessor {


hwangy:
The failing tests seem to be in FTP requests; unless there's something really strange going on, I don't think this is related to my pull request.

pgrosu:
I can get to the FTP site - try rerunning the build:

broad_ftp

hwangy:
How do I re-run the build in Github? I'm getting the same failure when doing a local test as well.

pgrosu:
@hwangy You do realize that 99 means the following phred-encoded error and accuracy:

Error:

```
> options(digits=20)
> 1-(10^(-99/10))
[1] 0.99999999987410748
```

Accuracy:

```
> (10^(-99/10))*3.234830000*10^9
[1] 0.40724096898341228
```

So this is very good so the upper limit of 99 is fine to keep, since it is so rare for the genotype call to be wrong given that the site is a variant - usually we work lower on the scale than 99 :)

~p

pgrosu:
@hwangy  I've updated the numbers above, since I copied them wrongly in a hurry from R :)

pgrosu:
@hwangy If you have write access to the repo there should be a circle with an arrow - as shown below - otherwise just close and reopen the PR:

rebuild_travis

pgrosu:
I have access to this file:

[ftp://ftp.broadinstitute.org/distribution/igv/TEST/cpgIslands with spaces.hg18.bed](ftp://ftp.broadinstitute.org/distribution/igv/TEST/cpgIslands%20with%20spaces.hg18.bed)

The second test is to see if the following is non-existent, which of course it does not :)

[ftp://ftp.broadinstitute.org/pub/igv/TEST/fileDoesntExist.txt](ftp://ftp.broadinstitute.org/pub/igv/TEST/fileDoesntExist.txt)

droazen:
@hwangy As part of this PR, could you also delete `VCFConstants.MAX_GENOTYPE_QUAL`?

droazen:
@hwangy had time to look at this? We should be able to merge once you address the one comment I had above about `VCFConstants.MAX_GENOTYPE_QUAL`

yfarjoun:
We're concerned about overflow in some dark corner of the codebase that uses htsjdk. 

if there's a compelling reason why one needs to use GQ>99 please reopen and explain.

PR #682, opened 2016-08-11
Made SAMTextReader.RecordIterator close itself properly

### LINKED ISSUE #681: SAMTextReader should allow a second iterator to be opened after closing the first one
The docs for `SamReader.iterator()` state that a second iterator can be opened on a reader, provided that you close the first iterator:

```
     * Only a single open iterator on a SAM or BAM file may be extant at any one time.  If you want to start
     * a second iteration, the first one must be closed first.
```

`SAMTextReader` violates this contract by closing the reader itself whenever an iterator on the reader is closed, making it impossible to open a second iterator even when you dutifully close the first. This is also inconsistent with the behavior of `BAMFileReader`, which does allow a second iterator to be opened after the first one is closed.

jamesemery:
I fixed the bug that cause SAMTextReaders to fail to get a new iterator after closing one properly.

resolves #681

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMTextReader.java
+++ b/src/main/java/htsjdk/samtools/SAMTextReader.java
@@ -221,10 +221,16 @@ class SAMTextReader extends SamReader.ReaderImplementation {
         }
 
         public void close() {
-            SAMTextReader.this.close();
+            if ( mReader != null) {
+                if (mIterator != null && this != mIterator) {
+                    throw new IllegalStateException("Attempt to close non-current iterator");
+                }
+                mIterator = null;
+            }
         }
 
         public boolean hasNext() {
+            if (mIterator == null) {throw new IllegalStateException("Iterator has been closed");}
             return mCurrentLine != null;
         }
 
--- a/src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
+++ b/src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
@@ -434,4 +434,25 @@ public class SamReaderFactoryTest {
         countRecords(reader);
     }
 
+    @Test
+    public void testGettingNewIteratorFromSamTextFileAfterClosing() throws IOException {
+        final String samFilePath = new File(TEST_DATA_DIR, "unsorted.sam").getAbsolutePath();
+        final URL samURL = new URL("file://" + samFilePath);
+        final SamReaderFactory factory = SamReaderFactory.makeDefault()
+                .validationStringency(ValidationStringency.SILENT);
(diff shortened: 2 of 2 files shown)


jamesemery:
@droazen

src/main/java/htsjdk/samtools/SAMTextReader.java
```diff
@@ -221,7 +221,12 @@ private RecordIterator() {
         }
 
         public void close() {
-            SAMTextReader.this.close();
```
lbergelson: I think this will end up with weird situations still.  I.e. we close the iterator but calling hasNext() still returns true

lbergelson:
@jamesemery I think there's more wrong with this class than we first suspected.  Could you check that `hasNext() == false` on a closed iterator, that `next` on a closed iterator throws `IllegalStateException`?

jamesemery:
I added an exception on hasNext() for a closed iterator to make the behavior more closely match BAMFileReader. It cannot completely match BAMFileReader however because a SAMTextFile is not seekable and consequently there is no guaranteed way of resetting the iterator the start of the file as there is with an indexed BAM file. The PR still resolves the bug reported in #681.

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -434,4 +434,25 @@ public void testSamReaderFromMalformedSeekableStream() throws IOException {
...
+        final SamReaderFactory factory = SamReaderFactory.makeDefault()
+                .validationStringency(ValidationStringency.SILENT);
+        final SamReader reader = factory.open(SamInputResource.of(samURL));
+        CloseableIterator<SAMRecord> iterator = reader.iterator();
+        Assert.assertTrue(iterator.hasNext());
+        iterator.close();
+        //Testing that an empty iterator throws
+        boolean hasThrown = false;
+        try {
+            iterator.hasNext();
```
yfarjoun: s/hasNext/next/

lbergelson:
So after more careful investigation, we decided that the current direction of this pull request would be worse than the status quo.  What we decided to do:

1. add gross `instanceof SeekableStream` checks and the ability to restart iteration over a sam file when possible to `SAMTextReader` 
  if restarting iteration is not possible it will throw
2. update the bam and cram readers so that the they throw in the case the restarting iteration is not possible.  currently in some instances they will start a new iterator at the same position that it left off.  this is surprising and potentially bug causing behavior

yfarjoun:
it seems that we should close this PR....if I'm wrong about this, please reopen.

Issue #578, opened 2016-04-18
nondeterministic failure of RelativeIso8601DateTest.testLazyInstance


akiezun:
I had a build fail due to this:

```
   [testng] FAILED: testLazyInstance
   [testng] java.lang.AssertionError: expected [1.46099929E12] but found [1.46099942E12]
   [testng] 	at org.testng.Assert.fail(Assert.java:94)
   [testng] 	at org.testng.Assert.failNotEquals(Assert.java:513)
   [testng] 	at org.testng.Assert.assertEquals(Assert.java:260)
   [testng] 	at org.testng.Assert.assertEquals(Assert.java:273)
   [testng] 	at htsjdk.samtools.util.RelativeIso8601DateTest.testLazyInstance(RelativeIso8601DateTest.java:19)
   [testng] 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
   [testng] 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
   [testng] 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
   [testng] 	at java.lang.reflect.Method.invoke(Method.java:497)
   [testng] 	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:86)
   [testng] 	at org.testng.internal.Invoker.invokeMethod(Invoker.java:643)
   [testng] 	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:820)
   [testng] 	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1128)
   [testng] 	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:129)
   [testng] 	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:112)
   [testng] 	at org.testng.TestRunner.privateRun(TestRunner.java:782)
   [testng] 	at org.testng.TestRunner.run(TestRunner.java:632)
   [testng] 	at org.testng.SuiteRunner.runTest(SuiteRunner.java:366)
   [testng] 	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:361)
   [testng] 	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:319)
   [testng] 	at org.testng.SuiteRunner.run(SuiteRunner.java:268)
[...]

PR #683, opened 2016-08-12
Fix IndexFactory.writeIndex for tabix indexes

### LINKED ISSUE #430: IndexFactory.writeIndex is incorrect for tabix indexes, missing BlockCompressedOutputStream
IndexFactory.writeIndex is incorrect for tabix indexes (does not use BlockCompressedOutputStream) and thus loading the index will blow up (reading does use BlockCompressedInoutStream).

Courtesy of @droazen, we know that compressing the index is indeed the right thing:
"the tbi index itself is supposed to be block compressed. Cf., eg., the code for writing tbi indices in htslib (https://github.com/samtools/htslib/blob/1e2ed48935a58845cecb64880e8cf7553c7db852/hts.c):

    } else if (fmt == HTS_FMT_TBI) {
        BGZF *fp = bgzf_open(fnidx, "w");
        if (fp == NULL) return -1;
        bgzf_write(fp, "TBI\1", 4);
        hts_idx_save_core(idx, fp, HTS_FMT_TBI);
        bgzf_close(fp);
"

gatk has a workaround: https://github.com/broadinstitute/gatk/pull/1422

magicDGS:

### DIFF
--- a/src/main/java/htsjdk/tribble/index/AbstractIndex.java
+++ b/src/main/java/htsjdk/tribble/index/AbstractIndex.java
@@ -343,13 +343,20 @@ public abstract class AbstractIndex implements MutableIndex {
     }
 
     @Override
+    public void write(final File idxFile) throws IOException {
+        final LittleEndianOutputStream idxStream =
+                new LittleEndianOutputStream(new BufferedOutputStream(new FileOutputStream(idxFile)));
+        try {
+            write(idxStream);
+        } finally {
+            idxStream.close();
+        }
+    }
+
+    @Override
     public void writeBasedOnFeatureFile(final File featureFile) throws IOException {
         if (!featureFile.isFile()) return;
-        final LittleEndianOutputStream idxStream =
-                new LittleEndianOutputStream(new BufferedOutputStream(new FileOutputStream(Tribble.indexFile(featureFile))));
-        write(idxStream);
-        idxStream.close();
-
+        write(Tribble.indexFile(featureFile));
     }
 
     public void read(final LittleEndianInputStream dis) throws IOException {
--- a/src/main/java/htsjdk/tribble/index/AbstractIndex.java
+++ b/src/main/java/htsjdk/tribble/index/AbstractIndex.java
@@ -344,8 +344,7 @@ public abstract class AbstractIndex implements MutableIndex {
 
     @Override
     public void write(final File idxFile) throws IOException {
-        final LittleEndianOutputStream idxStream =
-                new LittleEndianOutputStream(new BufferedOutputStream(new FileOutputStream(idxFile)));
+        final LittleEndianOutputStream idxStream = new LittleEndianOutputStream(new BufferedOutputStream(new FileOutputStream(idxFile)));
         try {
             write(idxStream);
         } finally {
--- a/src/main/java/htsjdk/tribble/index/AbstractIndex.java
+++ b/src/main/java/htsjdk/tribble/index/AbstractIndex.java
@@ -344,11 +344,8 @@ public abstract class AbstractIndex implements MutableIndex {
 
     @Override
     public void write(final File idxFile) throws IOException {
-        final LittleEndianOutputStream idxStream = new LittleEndianOutputStream(new BufferedOutputStream(new FileOutputStream(idxFile)));
-        try {
+        try(final LittleEndianOutputStream idxStream = new LittleEndianOutputStream(new BufferedOutputStream(new FileOutputStream(idxFile)))) {
             write(idxStream);
-        } finally {
-            idxStream.close();
         }
     }
 
--- a/src/main/java/htsjdk/tribble/index/Index.java
+++ b/src/main/java/htsjdk/tribble/index/Index.java
@@ -70,6 +70,14 @@ public interface Index {
     public void write(LittleEndianOutputStream stream) throws IOException;
 
     /**
+     * Writes the index into a file.
+     *
+     * @param idxFile Where to write the index.
+     * @throws IOException if the index is unable to write to the specified file
+     */
+    public void write(final File idxFile) throws IOException;
+
+    /**
      * Write an appropriately named and located Index file based on the name and location of the featureFile.
      * If featureFile is not a normal file, the index will silently not be written.
      * @param featureFile
--- a/src/main/java/htsjdk/tribble/index/IndexFactory.java
+++ b/src/main/java/htsjdk/tribble/index/IndexFactory.java
@@ -285,7 +285,9 @@ public class IndexFactory {
      * @param idx
      * @param idxFile
      * @throws IOException
+     * @deprecated use {@link Index#write(File)} instead
      */
+    @Deprecated
     public static void writeIndex(final Index idx, final File idxFile) throws IOException {
         idx.write(idxFile);
     }
--- a/src/main/java/htsjdk/tribble/index/IndexFactory.java
+++ b/src/main/java/htsjdk/tribble/index/IndexFactory.java
@@ -41,16 +41,13 @@ import htsjdk.tribble.index.tabix.TabixIndex;
 import htsjdk.tribble.index.tabix.TabixIndexCreator;
 import htsjdk.tribble.readers.PositionalBufferedStream;
 import htsjdk.tribble.util.LittleEndianInputStream;
-import htsjdk.tribble.util.LittleEndianOutputStream;
 import htsjdk.tribble.util.ParsingUtils;
 import htsjdk.tribble.util.TabixUtils;
 
 import java.io.BufferedInputStream;
-import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.Constructor;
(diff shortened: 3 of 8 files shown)


src/main/java/htsjdk/tribble/index/IndexFactory.java
```diff
@@ -292,7 +294,9 @@ public static LinearIndex createLinearIndex(final File inputFile, final FeatureC
     public static void writeIndex(final Index idx, final File idxFile) throws IOException {
         LittleEndianOutputStream stream = null;
         try {
-            stream = new LittleEndianOutputStream(new BufferedOutputStream(new FileOutputStream(idxFile)));
+            final OutputStream underlyingStream = (idx instanceof TabixIndex) ?
+                    new BlockCompressedOutputStream(idxFile) : new FileOutputStream(idxFile);
```
yfarjoun: would it be possible for the index to know what type of outputstream it needs and either open it for the factory, or write itself to that stream (given a file). it seems that it's more appropriate to give that decision to the index rather than to the factory.

also: tests please.

magicDGS: I was thinking about this before, because actually it makes more sense to have a `Index.write(File idxFile)` method instead of the factory. But I decided to go for this to maintain compatibility. In addition, the javadoc said that "It is the responsibility of this class to determine and create the correct index type from the input file or stream.", so that's why I did it like this.

magicDGS:
@yfarjoun, I don't know how to do tests for checking this...

src/main/java/htsjdk/tribble/index/Index.java
```diff
@@ -26,6 +26,7 @@
 import htsjdk.tribble.util.LittleEndianOutputStream;
 
 import java.io.File;
+import java.io.FileNotFoundException;
```
yfarjoun: where is this used?

src/main/java/htsjdk/tribble/index/AbstractIndex.java
```diff
@@ -25,6 +25,7 @@
 
 import java.io.BufferedOutputStream;
 import java.io.File;
+import java.io.FileNotFoundException;
```
yfarjoun: doesn't seem to be used

yfarjoun:
couldn't you test that you can write to a file (and an index is created) and write to a stream (and no exception is thrown)?

magicDGS:
@yfarjoun, I added a test to assert that `write(File idxFile)` is writing a file, and that `write( LittleEndianOutputStream stream)` to a null stream is not throwing an error. This was done for both `LinearIndex` and `TabixIndex`. Are they enough?

magicDGS:
I also squashed the commits. I think that now that the writing to the file is inside the index, the method `IndexFatory.writeIndex(final Index idx, final File idxFile)` could be deprecated. If someone already have the index file, `idx.write(idxFile)` could be used...

magicDGS:
Can you have a look to this, @yfarjoun? It is important for other PR that I'm preparing...

magicDGS:
Could you have a look, @yfarjoun? Thanks in advance!

src/main/java/htsjdk/tribble/index/AbstractIndex.java
```diff
@@ -343,13 +343,20 @@ public void write(final LittleEndianOutputStream stream) throws IOException {
...
 
     @Override
-    public void writeBasedOnFeatureFile(final File featureFile) throws IOException {
-        if (!featureFile.isFile()) return;
+    public void write(final File idxFile) throws IOException {
         final LittleEndianOutputStream idxStream =
-                new LittleEndianOutputStream(new BufferedOutputStream(new FileOutputStream(Tribble.indexFile(featureFile))));
-        write(idxStream);
-        idxStream.close();
+                new LittleEndianOutputStream(new BufferedOutputStream(new FileOutputStream(idxFile)));
```
yfarjoun: extra indent?

yfarjoun:
@lbergelson could you take a look? we originally reviewed this together.

src/main/java/htsjdk/tribble/index/AbstractIndex.java
magicDGS: It comes from the previous version, but I've just put everything in a single line.

src/test/java/htsjdk/tribble/index/IndexFactoryTest.java
```diff
@@ -112,4 +117,45 @@ public void testCreateTabixIndexOnBlockCompressed() {
                     "Tabix indexed (bgzipped) VCF does not contain sequence: " + samSequenceRecord.getSequenceName());
         }
     }
+
+    @Test
+    public void testWriteIndex() throws Exception {
+        // temp directory for this test
+        final File tempDir = IOUtil.createTempDir("writeIndex", null);
```
lbergelson: I think the standard in htsjdk is to use `File output = File.createTempFile()` to get your output file, and then call `output.deleteOnExit()` to make sure it's cleaned up afterwards.

src/test/java/htsjdk/tribble/index/IndexFactoryTest.java
```diff
@@ -112,4 +117,45 @@ public void testCreateTabixIndexOnBlockCompressed() {
                     "Tabix indexed (bgzipped) VCF does not contain sequence: " + samSequenceRecord.getSequenceName());
         }
     }
+
+    @Test
+    public void testWriteIndex() throws Exception {
```
lbergelson: Since this test is in indexFactory.writeIndex() it should probably be testing `indexFactory.writeIndex()`...  

Test for `write` on the index files themselves should live in TabixIndexTest and LinearIndexTest respectively. 

Either way (or both might be the best at the risk of redundantly testing things), `TabixIndexTest.readWriteTest` provides a good model for how the test might be made.  Instead of repeating the code for the different cases you can write a test function that takes a dataprovider with several inputs.  It might be a pain though for linear indexes since I don't see a read or equals method on them.

src/test/java/htsjdk/tribble/index/IndexFactoryTest.java
```diff
@@ -112,4 +117,45 @@ public void testCreateTabixIndexOnBlockCompressed() {
...
+        // write to an stream does not thrown an error
+        linearIndex.write(new LittleEndianOutputStream(nullOutputStrem));
+
+        // TABIX INDEX
+        final File inputFileVcfGz =
+                new File("src/test/resources/htsjdk/tribble/tabix/testTabixIndex.vcf.gz");
+        final TabixIndex tabixIndexVcfGz =
+                IndexFactory
+                        .createTabixIndex(inputFileVcfGz, new VCFCodec(), TabixFormat.VCF, null);
+        // wirte to a file
```
lbergelson: typo wirte -> write

src/main/java/htsjdk/tribble/index/AbstractIndex.java
```diff
@@ -343,13 +343,19 @@ public void write(final LittleEndianOutputStream stream) throws IOException {
     }
 
     @Override
+    public void write(final File idxFile) throws IOException {
```
magicDGS: Done

src/main/java/htsjdk/tribble/index/IndexFactory.java
```diff
@@ -290,16 +287,7 @@ public static LinearIndex createLinearIndex(final File inputFile, final FeatureC
      * @throws IOException
      */
     public static void writeIndex(final Index idx, final File idxFile) throws IOException {
```
magicDGS: Done

src/main/java/htsjdk/tribble/index/tabix/TabixIndex.java
```diff
@@ -201,6 +201,7 @@ public TabixFormat getFormatSpec() {
      *
      * @param tabixFile Where to write the index.
      */
+    @Override
     public void write(final File tabixFile) {
```
magicDGS: Done. Now throwing `IOException`

src/test/java/htsjdk/tribble/index/IndexFactoryTest.java
magicDGS: I changed exactly the same method (with a data provider instead) to `IndexTest`, because I think that it could live in that package (it is testing the interface method for different implementations).

I added a check for the sequence names and properties loaded and written, and a test bed file too.

src/test/java/htsjdk/tribble/index/IndexFactoryTest.java
magicDGS: I change it in the new test.

src/test/java/htsjdk/tribble/index/IndexFactoryTest.java
magicDGS: Changed in the new test.

magicDGS:
Addressed comments, @lbergelson. I changed the method to throw `IOException`, although I'm waiting for @yfarjoun comments about it.

Now that I'm looking at the tribble package, I that an issue should be open to review exceptions thrown by the classes. It will be useful for distinguish them. For instance, when I implemented `getTabixFormat()` it throws a generic `TribbleException`, but I can create another extension of the class that will be more informative.

magicDGS:
Tests are failing for FTP issues, @yfarjoun and @lbergelson...

lbergelson:
@magicDGS I restarted the tests.  I hate those ftp tests. 

This gets my approval.  This is a breaking change with the changes to the exceptions, but I think it's a positive one to make them more consistent.  Lets hear what @yfarjoun thinks before merging though. 

I agree with you that the current state of exceptions is pretty random.  It would be good to do an overhaul of them.

src/test/java/htsjdk/tribble/index/IndexFactoryTest.java
lbergelson: Good decision 👍

yfarjoun:
a strange test is failing again. otherwise 👍

lbergelson:
Oh yeah, that one.  #578 strikes again. Restarted.

magicDGS:
Thanks for review and accept this change, @yfarjoun and @lbergelson. I also opened an issue to have a look to the exceptions thrown in the package.

PR #685, opened 2016-08-16
SAMUtils:getOtherCanonicalAlignments extract 'SA' tag  and return a list of supplementary alignments

lindenb:
### Description

I've added a new function `getOtherCanonicalAlignments` in SAMUtils. This function is used to extract the 'SA' tag of a SAMRecord as a `List<SAMRecord>`of supplementary alignements. 

* 
```   
    /**
     * Extract a List of 'other canonical alignments' from a SAM record. Those alignments are stored as a string in the 'SA' tag as defined
     * in the SAM specification.
     * Each record in the List is a non-paired read.
     * The name, sequence and qualities are copied from the original record
     * The SAM flag is set to <code>SUPPLEMENTARY_ALIGNMENT (+ READ_REVERSE_STRAND )</code>
     * @param record must be non null and must have a non-null associated header.
     * @return a list of 'other canonical alignments' SAMRecords. The list is empty if the 'SA' attribute is missing.
     */
public static List<SAMRecord> getOtherCanonicalAlignments(final SAMRecord record)
```
* `testOtherCanonicalAlignments` was added to SAMUtilsTest ( create one read with SA flag, check the values of the supplementary alignments)

PS: The SA tag is defined in the spec as 

> Other canonical alignments in a chimeric alignment, formatted as a   semicolon-delimited list: (rname,pos,strand,CIGAR,mapQ,NM;)+.  Each element in the list represents a part of the chimeric alignment.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMUtils.java
+++ b/src/main/java/htsjdk/samtools/SAMUtils.java
@@ -41,6 +41,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.regex.Pattern;
 
 
 /**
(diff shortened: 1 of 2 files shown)


droazen:
@jamesemery could you please review?

src/main/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -1096,4 +1097,79 @@ public static SAMRecord clipOverlappingAlignedBases(final SAMRecord record, fina
...
+            otherRec.setReferenceIndex( tid );
+            
+            /* fill other fields */
+            otherRec.setAlignmentStart( Integer.parseInt(commaStrs[1]) ); 
+            otherRec.setFlags(
+                SAMFlag.SUPPLEMENTARY_ALIGNMENT.flag + 
+                (commaStrs[2].equals("+") ? 0 : SAMFlag.READ_REVERSE_STRAND.flag) );
+            otherRec.setCigar( TextCigarCodec.decode( commaStrs[3] ) );
+            otherRec.setMappingQuality( Integer.parseInt(commaStrs[4]) );
+            otherRec.setAttribute( SAMTagUtil.getSingleton().NM , Integer.parseInt(commaStrs[5]) );
```
jamesemery: All of these .parseInt() calls open the possibility of breaking on bad input and exposing a stack trace to the user. Perhaps you should emit a SAMException if any of these fields don't parse correctly?

src/main/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -1096,4 +1097,79 @@ public static SAMRecord clipOverlappingAlignedBases(final SAMRecord record, fina
...
+     * in the SAM specification.
+     * Each record in the List is a non-paired read.
+     * The name, sequence and qualities are copied from the original record.
+     * The SAM flag is set to <code>SUPPLEMENTARY_ALIGNMENT (+ READ_REVERSE_STRAND )</code>
+     * @param record must be non null and must have a non-null associated header.
+     * @return a list of 'other canonical alignments' SAMRecords. The list is empty if the 'SA' attribute is missing.
+     */
+    public static List<SAMRecord> getOtherCanonicalAlignments(final SAMRecord record) {
+        final Pattern semiColonPattern = Pattern.compile("[;]");
+        final Pattern commaPattern = Pattern.compile("[,]");
```
jamesemery: Do you need to compile these with every call of this method? Perhaps these could be extracted into a static field that you pre-compile so you don't have to remake these  patterns every time.

src/test/java/htsjdk/samtools/SAMUtilsTest.java
```diff
@@ -173,4 +173,59 @@ public void testClippingOfRecordWithMateAtSamePosition() {
...
+        Assert.assertEquals(other.getAttribute(SAMTagUtil.getSingleton().NM),1);
+        Assert.assertEquals(other.getCigarString(),"3S2=1X2=2S");
+        
+        other = suppl.get(1);
+        Assert.assertEquals(other.getReferenceName(),"1");
+        Assert.assertEquals(other.getAlignmentStart(),191);
+        Assert.assertTrue(other.getReadNegativeStrandFlag());
+        Assert.assertEquals(other.getMappingQuality(), 60);
+        Assert.assertEquals(other.getAttribute(SAMTagUtil.getSingleton().NM),0);
+        Assert.assertEquals(other.getCigarString(),"8M2S");
```
jamesemery: You should at least add a test that a record with no SA tag returns an empty list properly

src/main/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -1096,4 +1097,79 @@ public static SAMRecord clipOverlappingAlignedBases(final SAMRecord record, fina
...
+            final String commaStrs[] = commaPattern.split(semiColonStr);
+            if( commaStrs.length != 6 )  throw new SAMException("Bad 'SA' attribute in " + semiColonStr);
+            
+            /* create the new record */
+            final SAMRecord otherRec = samReaderFactory.createSAMRecord( record.getHeader() );
+            
+            /* copy fields from the original record */
+            otherRec.setReadName( record.getReadName() );
+            otherRec.setReadBases( record.getReadBases() );
+            otherRec.setBaseQualities( record.getBaseQualities() );
```
jamesemery: Perhaps you might also set the supplementary alignment flag in the new read? It should be predictable according to convention

jamesemery: Furthermore, what about the mate information?

src/main/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -1096,4 +1097,79 @@ public static SAMRecord clipOverlappingAlignedBases(final SAMRecord record, fina
...
+            
+            /* get reference sequence */
+            final int tid = record.getHeader().getSequenceIndex( commaStrs[0] );
+            if( tid == -1 ) throw new SAMException("Unknown contig in " + semiColonStr);
+            otherRec.setReferenceIndex( tid );
+            
+            /* fill other fields */
+            otherRec.setAlignmentStart( Integer.parseInt(commaStrs[1]) ); 
+            otherRec.setFlags(
+                SAMFlag.SUPPLEMENTARY_ALIGNMENT.flag +
```
jamesemery: Is it correct to set the Supplementary Alignment flag for every SA read? What if record is supplementary? In that case wouldn't you want to set the the first flag to primary according to convention?

lindenb:
Thank you for the review and back to you @jamesemery

  * I've considered the mate
  * surrounded parseInt with try catch
  * considered the first suppl alignment if the read is supplementary
  * moved the Pattern out of the function.
  * updated the test

NB: I'll be away from github for a few days.

jamesemery:
@droazen These changes look good to me

lbergelson:
Thanks @lindenb and @jamesemery

PR #688, opened 2016-08-19
Added a "RANDOM" duplicate scoring strategy.

tfenne:
### Description

When evaluating error rates in data with UMIs and contrasting different approaches to duplicate marking, it is useful to have an unbiased read picked as the representative for a duplicate set instead of some heuristic for "best".

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
+++ b/src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
@@ -24,6 +24,8 @@
 
 package htsjdk.samtools;
 
+import htsjdk.samtools.util.Murmur3;
+
 /**
  * This class helps us compute and compare duplicate scores, which are used for selecting the non-duplicate
  * during duplicate marking (see MarkDuplicates).
@@ -33,8 +35,12 @@ public class DuplicateScoringStrategy {
 
     public enum ScoringStrategy {
         SUM_OF_BASE_QUALITIES,
-        TOTAL_MAPPED_REFERENCE_LENGTH
+        TOTAL_MAPPED_REFERENCE_LENGTH,
+        RANDOM,
     }
+
+    /** Hash used for the RANDOM scoring strategy. */
+    private static final Murmur3 hasher = new Murmur3(1);
 
     /** An enum to use for storing temporary attributes on SAMRecords. */
     private static enum Attr { DuplicateScore }
@@ -80,6 +86,8 @@ public class DuplicateScoringStrategy {
                         score += SAMUtils.getMateCigar(record).getReferenceLength();
                     }
                     break;
+                case RANDOM:
+                    score += (short) (hasher.hashUnencodedChars(record.getReadName()) >> 16);
             }
 
             storedScore = score;


tfenne:
@yfarjoun This is a trivial little addition.  There don't appear to be standalone tests for the DuplicateScoringStrategy, and I'm not really up for writing comprehensive tests to enable my ~3 line addition.  Can you take a look and let me know what you think?

src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
```diff
@@ -80,6 +81,8 @@ public static short computeDuplicateScore(final SAMRecord record, final ScoringS
                         score += SAMUtils.getMateCigar(record).getReferenceLength();
                     }
                     break;
+                case RANDOM:
+                    score += (short) Math.sqrt(record.getReadName().hashCode());
```
yfarjoun: Since we already have the murmur3 hash in htsjdk, and given that we already agreed for other parts of the code that it is better than the default String.hashCode implementation, could you use murmur3 instead?

tfenne: I can make that change if you'd like.  I felt a bit funny about it because the `enum` would have to have an instance of Murmur3, which felt funny to me.  But other than for the seed, `Murmur3` is stateless so it would work.  Thoughts?

yfarjoun: Since the seed is final I think you can do it however you like, in the
enum, or in DuplicateScoringStrategy. Not sure why it needs to go in the
enum.

tfenne: Doh - thanks for pointing that out.  For some reason I was thinking `DuplicateScoringStrategy` _was_ the enum.  Will fix.

tfenne:
Requested change made @yfarjoun.  Also since Murmur3 is more random, I switched to just right shifting 16 bits to get a short out, instead of sqrt'ing.

src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
```diff
@@ -80,6 +86,8 @@ public static short computeDuplicateScore(final SAMRecord record, final ScoringS
                         score += SAMUtils.getMateCigar(record).getReferenceLength();
                     }
                     break;
+                case RANDOM:
+                    score += (short) (hasher.hashUnencodedChars(record.getReadName()) >> 16);
```
yfarjoun: I'm curious if there's a reason you didn't opt to take the low order bits using & 0xFFFF ? Speed is not an issue here, I presume, but clarity?

src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
```diff
@@ -33,9 +35,13 @@
 
     public enum ScoringStrategy {
         SUM_OF_BASE_QUALITIES,
-        TOTAL_MAPPED_REFERENCE_LENGTH
+        TOTAL_MAPPED_REFERENCE_LENGTH,
+        RANDOM,
     }
 
+    /** Hash used for the RANDOM scoring strategy. */
+    private static Murmur3 hasher = new Murmur3(1);
```
yfarjoun: final?

src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
tfenne: No particular reason other than that `>> 16` came to mind for me first.

src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
tfenne: Done.

yfarjoun:
looks good. merge when tests pass. 👍

src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
yfarjoun: I think that there should be no difference in the randomness, due to how
murmur works (lots of >>>'s, which mixes up low and high order bits), so
I'm happy.

PR #689, opened 2016-08-21
Yf add tag aware merge samdictionary

yfarjoun:
### Description

This adds a static function that can merge two SamSequenceDictionaries into one. The problem it's solving is that in MergeBamAlignment the reference dictionary and the aligned Bam dictionary, while having the same sequences, may have different tags. In particular, the reference dictionary has lots of interesting tags like MD, SP, UR etc. and the (bwa) dictionary in the aligned bam may contain AH which informs us that the sequence is an Alternative Haplotype. 

This PR enables MergeBamAlignment to merge the two dictionaries rather than only use the reference one (as it currently does)

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
+++ b/src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
@@ -29,13 +29,15 @@ import java.io.Serializable;
 import java.math.BigInteger;
 import java.security.MessageDigest;
 import java.util.*;
+import java.util.stream.Collector;
 import java.util.stream.Collectors;
 
 import javax.xml.bind.annotation.XmlElement;
 import javax.xml.bind.annotation.XmlRootElement;
 import javax.xml.bind.annotation.XmlTransient;
 
-import static htsjdk.samtools.SAMSequenceRecord.UNKNOWN_SEQUENCE_LENGTH;
+import static htsjdk.samtools.SAMSequenceRecord.*;
+import static java.util.stream.Collectors.toList;
 
 /**
  * Collection of SAMSequenceRecords.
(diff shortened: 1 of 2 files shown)


yfarjoun:
@tfenne This is the htsjdk part we discussed in https://github.com/broadinstitute/picard/pull/627 

Can you take a look?

yfarjoun:
friendly ping. @tfenne.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -266,5 +268,100 @@ public String toString() {
...
+            SAMSequenceRecord.MD5_TAG,
+            SAMSequenceRecord.SEQUENCE_LENGTH_TAG);
+
+    /**
+     * Will merge dictionaryTags from two dictionaries into one focusing on merging the tags rather than the sequences.
+     *
+     * Requires that dictionaries have the same SAMSequence records in the same order.
+     * For each sequenceIndex, the union of the tags from both sequences will be added to the new sequence, mismatching
+     * values (for tags that are in both) will generate a warning, and the value from dict1 will be used.
+     * For tags that are in tagsToEquate. an unequal value will generate an error (an IllegalArgumentException will
```
tfenne: Extra period after `tagsToEquate`

tfenne: Also it's called `tagsToMatch`

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -266,5 +268,100 @@ public String toString() {
...
+     * be thrown.)
+     *
+     * @param dict1 first dictionary
+     * @param dict2 first dictionary
+     * @param tagsToMatch list of tags that must be equal if present in both sequence. Typical values will be MD, and LN
+     * @return dictionary consisting of the same sequences as the two inputs with the merged values of tags.
+     */
+    static public SAMSequenceDictionary mergeDictionaries(final SAMSequenceDictionary dict1,
```
tfenne: Function name should be more descriptive given all the different ways in which one could mergeDictionaries.  Perhaps `mergeTags` instead?

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -266,5 +268,100 @@ public String toString() {
...
+                                                          final SAMSequenceDictionary dict2,
+                                                          final List<String> tagsToMatch) {
+
+        if(dict1.getSequences().size()!=dict2.getSequences().size()) {
```
tfenne: `if(` -> `if (`

tfenne: Spaces around `!=` please.

tfenne: If you're going to check this, why not go all the way and check that the two dictionaries contain the exact same set of names up front?

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -266,5 +268,100 @@ public String toString() {
...
+            throw new IllegalArgumentException(String.format("Do not use this function to merge dictionaries with " +
+                    "different number of sequences in them. Found %d and %d sequences",
+                    dict1.getSequences().size(), dict2.getSequences().size()));
+        }
+        final SAMSequenceDictionary finalDict = new SAMSequenceDictionary();
+        for (final SAMSequenceRecord sequence : dict1.getSequences()) {
+            final SAMSequenceRecord mergedSequence = new SAMSequenceRecord(sequence.getSequenceName(),
+                    UNKNOWN_SEQUENCE_LENGTH);
```
tfenne: `SAMSequenceRecord` already has a constructor that takes just the name - no need to define `UNKNOWN_SEQUENCE_LENGTH` and do this.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -266,5 +268,100 @@ public String toString() {
...
+            finalDict.addSequence(mergedSequence);
+            final int sequenceIndex = sequence.getSequenceIndex();
+            if (!dict2.getSequence(sequenceIndex).getSequenceName().equals(sequence.getSequenceName())) {
```
tfenne: I think it would be better to validate up front, and then here you can assume everything is going to line up.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -266,5 +268,100 @@ public String toString() {
...
+                throw new IllegalArgumentException(String.format("Non-equal sequence names (%s and %s) at index %d in " +
+                        "the dictionaries.",
+                        sequence.getSequenceName(), dict2.getSequence(sequenceIndex).getSequenceName(), sequenceIndex));
+            }
+
+            final Set<String> allTags = new HashSet<>(dict1.getSequence(sequenceIndex).getAttributes()
```
tfenne: I think this would be cleaner as:
```java
final Set<String> allTags = new HashSet<>;
dict1.getSequence(sequenceIndex).getAttributes().forEach(a -> allTags.add(a.getKey());
dict2.getSequence(sequenceIndex).getAttributes().forEach(a -> allTags.add(a.getKey());
```

Though I would also suggest renaming `sequence` to `s1` and defining an `s2` at line 305 and then using them througout instead of repeatedly accessing through the dictionaries.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -266,5 +268,100 @@ public String toString() {
...
+            allTags.addAll(dict2.getSequence(sequenceIndex).getAttributes()
+                    .stream().map(Map.Entry::getKey).collect(Collectors.toSet()));
+
+            for (final String tag : allTags) {
+                final String value1 = dict1.getSequence(sequenceIndex).getAttribute(tag);
+                final String value2 = dict2.getSequence(sequenceIndex).getAttribute(tag);
+
+                if (value1 != null && value2 != null && !value1.equals(value2)) {
+                    if (tagsToMatch.contains(tag)) {
+                        final String error = String.format("Cannot merge the two dictionaries. Found sequence entry for which " +
```
tfenne: No need for `String.format` here, since log.error() can just be fed a comma separated list of things to concatenate and log.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -266,5 +268,100 @@ public String toString() {
...
+
+            if (length1 != UNKNOWN_SEQUENCE_LENGTH && length2 != UNKNOWN_SEQUENCE_LENGTH && length1 != length2) {
+                if (tagsToMatch.contains(SAMSequenceRecord.SEQUENCE_LENGTH_TAG)) {
+                    final String error = String.format("Cannot merge the two dictionaries. Found sequence entry for which " +
+                                    "lengths differ: %s has lengths %s and %s",
+                            dict1.getSequence(sequenceIndex).getSequenceName(),
+                            length1, length2);
+
+                    log.error(error);
+                    throw new IllegalArgumentException(error);
```
tfenne: Logging and then throwing is just going to lead to the same error getting printed multiple times.  Please pick one!

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -266,5 +268,100 @@ public String toString() {
...
+                } else {
+                    log.warn("Found sequence entry for which " +
```
tfenne: a) String concat when not needed.  b) I don't think log.warn() will String.format() for you.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -266,5 +268,100 @@ public String toString() {
...
+                    }
+                }
+                mergedSequence.setAttribute(tag, value1 == null ? value2 : value1);
+            }
+
+            final int length1 = dict1.getSequence(sequenceIndex).getSequenceLength();
+            final int length2 = dict2.getSequence(sequenceIndex).getSequenceLength();
+
+            if (length1 != UNKNOWN_SEQUENCE_LENGTH && length2 != UNKNOWN_SEQUENCE_LENGTH && length1 != length2) {
+                if (tagsToMatch.contains(SAMSequenceRecord.SEQUENCE_LENGTH_TAG)) {
```
tfenne: I don't think you should allow merging to happen if the lengths differ, regardless.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -266,5 +268,100 @@ public String toString() {
...
+     * be thrown.)
+     *
+     * @param dict1 first dictionary
+     * @param dict2 first dictionary
+     * @param tagsToMatch list of tags that must be equal if present in both sequence. Typical values will be MD, and LN
+     * @return dictionary consisting of the same sequences as the two inputs with the merged values of tags.
+     */
+    static public SAMSequenceDictionary mergeDictionaries(final SAMSequenceDictionary dict1,
+                                                          final SAMSequenceDictionary dict2,
+                                                          final List<String> tagsToMatch) {
```
tfenne: Circling back, I think this parameter should probably go away.  Length (if present) and MD5 (if present) should be _required_ to match in all cases, since by definition they should match.  Other tags, I'm not sure it matters if they match.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -266,5 +268,100 @@ public String toString() {
...
+                            dict1.getSequence(sequenceIndex).getSequenceName(),
+                            length1, length2);
+
+                    log.error(error);
+                    throw new IllegalArgumentException(error);
+                } else {
+                    log.warn("Found sequence entry for which " +
+                                    "lengths differ: Sequence %s has lengths %s and %s. Using Value %s",
+                            dict1.getSequence(sequenceIndex).getSequenceName(),
+                            length1, length2, length1);
```
magicDGS: If I understand correctly, the logging of the used value is not always `length1` regarding the next line.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
yfarjoun: I agree about Length and MD5.

Other tags there are two options when they do not match: take the first, or throw an exception. 

My idea is that if the tags are in the list, an exception will be thrown, otherwise the value from the first dictionary will be taken. Sounds good?

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
yfarjoun: what's the value of doing that upfront?

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
yfarjoun: It's deprecated for some reason...

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
yfarjoun: ok.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
yfarjoun: OK. I thought we tend to do both to protect against someone catching the exception...

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
yfarjoun: ok.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
yfarjoun: I removed this if statement. length must be equal.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
yfarjoun: tx

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
yfarjoun: good idea. done.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
yfarjoun: ok. I'll do that.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
tfenne: I think it's much clearer to write methods where possible that:
a) validate their inputs up front
b) proceed to implement the intended function assuming/knowing they have valid inputs

yfarjoun:
@tfenne I responded to your review requests, could you take another look?

yfarjoun:
@tfenne bump 🙏

PR #690, opened 2016-08-22
Adds a function isWithinHammingDistance that checks if two strings ar…

fleharty:
### Description

The motivation for this code is to add functionality to compare two strings and determine
their Hamming distance.  Hamming distance is a common metric that is used to compare
strings.  The two functions, hammingDistance, and isWithinHammingDistance provide a way
to check the actual Hamming distance, and whether or not two strings are within a given
Hamming distance of each other.  isWithinHammingDistance is provided because there are
many cases where it is important to know only if two strings are within a given Hamming distance
and we can terminate early if the two strings are sufficiently dissimilar.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

…e within a given Hamming distance

### DIFF
--- a/src/main/java/htsjdk/samtools/util/StringUtil.java
+++ b/src/main/java/htsjdk/samtools/util/StringUtil.java
@@ -545,4 +545,33 @@ public class StringUtil {
 
         return i;
     }
+
+    /**
+     * Determines if two strings s1 and s2 are within maxHammingDistance of ecah other using the Hamming Distance metric.
+     * Since Hamming distance is not defined for strings of differing lengths, we throw an exception if
+     * the two strings are of different lengths.
+     *
+     * @param s1 The first string to compare
+     * @param s2 The second string to compare, note that if s1 and s2 are swapped the value returned will be identical.
+     * @param maxHammingDistance The largest Hamming distance the strings can have for this function to return true.
+     * @return true if the two strings are within maxHammingDistance of each other, false otherwise.
+     * @throws IllegalArgumentException If the two strings have differing lengths.
+     */
+    public static boolean isWithinHammingDistance(final String s1, final String s2, final int maxHammingDistance) {
+        if (s1.length() != s2.length()) {
+            throw new IllegalArgumentException("Attempted to determine if two strings of different length were within a specified edit distance.");
+        }
+        int measuredDistance = 0;
+        for (int i = 0;i < s1.length();i++) {
+            if (s1.charAt(i) != s2.charAt(i)) {
+                measuredDistance++;
+                // If the measuredDistance is larger than the maxHammingDistance we can short circuit and return
+                // false, there is no need to continue evaluating the distance.
+                if (measuredDistance > maxHammingDistance) {
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
 }
--- a/src/main/java/htsjdk/samtools/util/StringUtil.java
+++ b/src/main/java/htsjdk/samtools/util/StringUtil.java
@@ -549,8 +549,8 @@ public class StringUtil {
     /**
      * Calculates the Hamming distance (number of character mismatches) between two strings s1 and s2.
      * Since Hamming distance is not defined for strings of differing lengths, we throw an exception if
-     * the two strings are of different lengths.  Hamming distance is case sensitive.  Also note that when
-     * used to compare two DNA strings that contain Ns mathing Ns will not be counted as mismatches.
+     * the two strings are of different lengths.  Hamming distance is case sensitive and does not have
+     * any special treatment for DNA.
      *
      * @param s1 The first string to compare
      * @param s2 The second string to compare, note that if s1 and s2 are swapped the value returned will be identical.
@@ -574,8 +574,8 @@ public class StringUtil {
     /**
      * Determines if two strings s1 and s2 are within maxHammingDistance of each other using the Hamming distance metric.
      * Since Hamming distance is not defined for strings of differing lengths, we throw an exception if
-     * the two strings are of different lengths.  Hamming distance is case sensitive.  Also note that when
-     * used to compare two DNA strings that contain Ns mathing Ns will not be counted as mismatches.
+     * the two strings are of different lengths.  Hamming distance is case sensitive and does not have any
+     * special treatment for DNA.
      *
      * @param s1 The first string to compare
      * @param s2 The second string to compare, note that if s1 and s2 are swapped the value returned will be identical.
--- a/src/main/java/htsjdk/samtools/util/StringUtil.java
+++ b/src/main/java/htsjdk/samtools/util/StringUtil.java
@@ -547,9 +547,10 @@ public class StringUtil {
     }
 
     /**
-     * Calculates the hamming distance between two strings s1 and s2.
+     * Calculates the Hamming distance (number of character mismatches) between two strings s1 and s2.
      * Since Hamming distance is not defined for strings of differing lengths, we throw an exception if
-     * the two strings are of different lengths.
+     * the two strings are of different lengths.  Hamming distance is case sensitive.  Also note that when
+     * used to compare two DNA strings that contain Ns mathing Ns will not be counted as mismatches.
      *
      * @param s1 The first string to compare
      * @param s2 The second string to compare, note that if s1 and s2 are swapped the value returned will be identical.
@@ -558,7 +559,8 @@ public class StringUtil {
      */
     public static int hammingDistance(final String s1, final String s2) {
         if (s1.length() != s2.length()) {
-            throw new IllegalArgumentException("Attempted to determine Hamming distance of strings with differing lengths.");
+            throw new IllegalArgumentException("Attempted to determine Hamming distance of strings with differing lengths. " +
+                    "The first string has length " + s1.length() + " and the second string has length " + s2.length() + ".");
         }
         int measuredDistance = 0;
         for (int i = 0;i < s1.length();i++) {
@@ -570,9 +572,10 @@ public class StringUtil {
     }
 
     /**
-     * Determines if two strings s1 and s2 are within maxHammingDistance of ecah other using the Hamming Distance metric.
+     * Determines if two strings s1 and s2 are within maxHammingDistance of each other using the Hamming distance metric.
      * Since Hamming distance is not defined for strings of differing lengths, we throw an exception if
-     * the two strings are of different lengths.
+     * the two strings are of different lengths.  Hamming distance is case sensitive.  Also note that when
+     * used to compare two DNA strings that contain Ns mathing Ns will not be counted as mismatches.
      *
      * @param s1 The first string to compare
      * @param s2 The second string to compare, note that if s1 and s2 are swapped the value returned will be identical.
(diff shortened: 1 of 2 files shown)


magicDGS:
Why don't extract also the method for computing the distance, @fleharty? It will be useful for other purposes...

fleharty:
@magicDGS I agree, so I added the hammingDistance function.
I personally need the isWithinHammingDistance to be a little bit faster than calculating the hammingDistance and then checking if the the value is within a particular value, which is why I have that function.

cmnbroad:
I'm wondering if these belong in htsjdk (I know there are similar things like this in the code, like levenshteinDistance, but...) they're not used by htsjdk anywhere, AFAICT and have no dependence on any htsjdk classes.

fleharty:
@cmnbroad @yfarjoun Picard depends on htsjdk and there is a PR now in Picard that will make use of this function.  Since htsjdk already had a stringutils, I was trying to reduce the possibility of code duplication in the future given that things like levenshteinDistance was there, but I see your point that levenshteinDistance isn't ever called within htsjdk.

yfarjoun:
I think it's fine to have general-purpose utility functions like this in
htsjdk. there are many functions in htsjdk that are not called internally
(they should ofcourse be tested...) but they are used from other projects,
e.g. picard and GATK.

src/main/java/htsjdk/samtools/util/StringUtil.java
```diff
@@ -545,4 +545,56 @@ public static int levenshteinDistance(final String string1, final String string2
 
         return i;
     }
+
+    /**
+     * Calculates the hamming distance between two strings s1 and s2.
```
lbergelson: should hamming always be capitalized in the comments?  It's inconsistent.

lbergelson: Could you add a brief explanation of what the hamming distance is to the javadoc.  I.e. "Calculates the Hamming distance (number of character mismatches) between two strings"

src/main/java/htsjdk/samtools/util/StringUtil.java
```diff
@@ -545,4 +545,56 @@ public static int levenshteinDistance(final String string1, final String string2
...
+        for (int i = 0;i < s1.length();i++) {
+            if (s1.charAt(i) != s2.charAt(i)) {
+                measuredDistance++;
+            }
+        }
+        return measuredDistance;
+    }
+
+    /**
+     * Determines if two strings s1 and s2 are within maxHammingDistance of ecah other using the Hamming Distance metric.
```
lbergelson: typo ecah

src/main/java/htsjdk/samtools/util/StringUtil.java
```diff
@@ -545,4 +545,56 @@ public static int levenshteinDistance(final String string1, final String string2
...
+     * the two strings are of different lengths.
+     *
+     * @param s1 The first string to compare
+     * @param s2 The second string to compare, note that if s1 and s2 are swapped the value returned will be identical.
+     * @return Hamming distance between s1 and s2.
+     * @throws IllegalArgumentException If the two strings have differing lengths.
+     */
+    public static int hammingDistance(final String s1, final String s2) {
+        if (s1.length() != s2.length()) {
+            throw new IllegalArgumentException("Attempted to determine Hamming distance of strings with differing lengths.");
```
lbergelson: Could you include the actual length of the strings in the error message so it's easier to debug what happened?

src/test/java/htsjdk/samtools/util/StringUtilTest.java
```diff
@@ -67,4 +67,53 @@ public void testSplit(final String input, final String[] expectedResult, final b
...
+                {"ATAC", "T", 1, false},
+                {"", "GCAT", 0, false}
+        };
+    }
+
+    @Test(dataProvider = "hammingDistanceExceptionProvider", expectedExceptions = IllegalArgumentException.class)
+    public void testIsWithinHammingDistanceExceptions(final String s1, final String s2, final int maxHammingDistance) {
+        // We assert hammingDistance = 0, and isWithinHammingDistance = true because the values don't matter
+        // and we are checking to ensure that the IllegalArgumentException is thrown
+        Assert.assertEquals(StringUtil.hammingDistance(s1, s2), 0);
```
lbergelson: This won't check the second case if the first one throws an exception. You should either have two different tests one for each case, (and don't use the asserts, just call the method directly), or you should have calls like this one after another.
```
 Assert.assertThrows(IllegalArgumentException.class,  () -> {StringUtil.hammingDistance(s1, s2);});
```

lbergelson:
@yfarjoun @cmnbroad What's the resolution here?  Sounds like an ok for this to go forward?  
@fleharty I made a few minor comments for you to address.

src/main/java/htsjdk/samtools/util/StringUtil.java
fleharty: fixed

src/main/java/htsjdk/samtools/util/StringUtil.java
fleharty: fixed

cmnbroad:
It sounds like I'm the only one who thinks this is out of place here, so I defer to the community sentiment.

src/main/java/htsjdk/samtools/util/StringUtil.java
fleharty: fixed

src/main/java/htsjdk/samtools/util/StringUtil.java
```diff
@@ -545,4 +545,56 @@ public static int levenshteinDistance(final String string1, final String string2
...
+     * Calculates the hamming distance between two strings s1 and s2.
+     * Since Hamming distance is not defined for strings of differing lengths, we throw an exception if
+     * the two strings are of different lengths.
+     *
+     * @param s1 The first string to compare
+     * @param s2 The second string to compare, note that if s1 and s2 are swapped the value returned will be identical.
+     * @return Hamming distance between s1 and s2.
+     * @throws IllegalArgumentException If the two strings have differing lengths.
+     */
+    public static int hammingDistance(final String s1, final String s2) {
```
nh13: Since likely this is going to be used to compare DNA sequences, you probably want in the doc stating that the comparison is case sensitive.  Also, do you have a need to count or not count `N`s as mismatches?

src/main/java/htsjdk/samtools/util/StringUtil.java
```diff
@@ -545,4 +545,56 @@ public static int levenshteinDistance(final String string1, final String string2
...
+     * Determines if two strings s1 and s2 are within maxHammingDistance of ecah other using the Hamming Distance metric.
+     * Since Hamming distance is not defined for strings of differing lengths, we throw an exception if
+     * the two strings are of different lengths.
+     *
+     * @param s1 The first string to compare
+     * @param s2 The second string to compare, note that if s1 and s2 are swapped the value returned will be identical.
+     * @param maxHammingDistance The largest Hamming distance the strings can have for this function to return true.
+     * @return true if the two strings are within maxHammingDistance of each other, false otherwise.
+     * @throws IllegalArgumentException If the two strings have differing lengths.
+     */
```
nh13: Ditto about the doc usage for DNA bases.

src/test/java/htsjdk/samtools/util/StringUtilTest.java
```diff
@@ -67,4 +67,53 @@ public void testSplit(final String input, final String[] expectedResult, final b
...
+        Assert.assertEquals(StringUtil.isWithinHammingDistance(s1, s2, maxHammingDistance), true);
+    }
+
+    @DataProvider(name="hammingDistanceProvider")
+    public Object[][] hammingDistance() {
+        return new Object[][] {
+                {"ATAC", "GCAT", 3},
+                {"ATAGC", "ATAGC", 0},
+                {"ATAC", "atac", 4}, // Hamming distance is case sensitive
```
nh13: One test for non-DNA bases?

src/test/java/htsjdk/samtools/util/StringUtilTest.java
```diff
@@ -67,4 +67,53 @@ public void testSplit(final String input, final String[] expectedResult, final b
...
+        Assert.assertEquals(StringUtil.isWithinHammingDistance(s1, s2, maxHammingDistance), expectedResult);
+    }
+
+    @DataProvider(name="withinHammingDistanceExceptionProvider")
+    public Object[][] isWithinHammingDistanceException() {
+        return new Object[][] {
+                {"ATAC", "GCT", 3, true},
+                {"ATAC", "AT", 2, false},
+                {"ATAC", "T", 1, false},
+                {"", "GCAT", 0, false}
```
nh13: The params are so nicely aligned above, how about here and below ;) ?

src/test/java/htsjdk/samtools/util/StringUtilTest.java
fleharty: Fixed.

src/main/java/htsjdk/samtools/util/StringUtil.java
fleharty: @nh13 Added comment about case sensitive.
I'm also adding a note that matching Ns will not be counted as mismatches, since that isn't desired here, but I can see that someone else might want that.

src/main/java/htsjdk/samtools/util/StringUtil.java
fleharty: fixed

src/test/java/htsjdk/samtools/util/StringUtilTest.java
fleharty: fixed

src/test/java/htsjdk/samtools/util/StringUtilTest.java
fleharty: fixed

fleharty:
@lbergelson @nh13 Thanks for the comments, back to you.

src/test/java/htsjdk/samtools/util/StringUtilTest.java
fleharty: @nh13 I changed this to test for Hamming distance alone and not account for the fact that most the time people are interested in comparing DNA, and we might want to count matching Ns as mismatches.  I'd just like to get your opinion on what you think should be done here.

For the UmiAwareMarkDupliates, I don't ever expect to encounter an N, so I think checking only Hamming distance is okay, but I can see that others might not be okay with this.  If I set different behavior for matching Ns, I should probably also check to see if it is A, T, C, or G as well and throw an exception if an unexpected character is encountered.

Thoughts?

nh13: Lets defer adding support for matching valid bases for now, but just know that it may become useful in the future.

src/main/java/htsjdk/samtools/util/StringUtil.java
lbergelson: @fleharty Sorry for the very slow response.  I think this note about N's is more confusing that it is clarifying.  I would say something more like "This implementation is case sensitive and does not have any special treatment for DNA."

fleharty: @lbergelson fixed

lbergelson:
👍

PR #691, opened 2016-08-23
Removed unused .bcf test files.

cmnbroad:
### Description

These two files are not referenced anywhere. (They're also of questionable utility since they're BCF2.2 files, which is not yet supported by htsjdk, that contain embedded version 4.1 VCF headers (two point versions older than when BCF 2.2 was introduced). I can't tell from the spec if thats allowed, but either way I think these files are not helpful.


lbergelson:
👍

PR #693, opened 2016-08-31
Added support for an empty NM tag in SA tag fields using '*' character

jamesemery:
### Description

Previously when encountered with a '*' in the last field of an SA tag for getOtherCannonicalAlignments() method would throw an exception. It will now gracefully allow for an empty NM field in the new field.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


yfarjoun:
👍 thanks @jamesemery

lbergelson:
this may effect https://github.com/broadinstitute/gatk/pull/2133

PR #703, opened 2016-09-12
Added a secondary constructor to IntervalList that takes a dict.

tfenne:
### Description

I find myself writing these three lines of code again and again and again to initialize an IntervalList since it requires a `SAMFileHeader` with a useful `SAMSequenceDictionary` and there's no way to create a header from a dictionary as an expression.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMFileHeader.java
+++ b/src/main/java/htsjdk/samtools/SAMFileHeader.java
@@ -120,6 +120,12 @@ public class SAMFileHeader extends AbstractSAMHeaderRecord
         setAttribute(VERSION_TAG, CURRENT_VERSION);
     }
 
+    /** Constructor that initializes the sequence dictionary with the provided one. */
+    public SAMFileHeader(final SAMSequenceDictionary dict) {
+        this();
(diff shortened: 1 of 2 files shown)


magicDGS:
Why a private method in `IntervalList` and not a public static in `IntervalUtil`. It may be useful not only for construct an interval list...

tfenne:
@magicDGS I wanted the secondary constructor there.  It's ~3 lines of simple code - I don't think it warrants extracting to a separate utility class.

tfenne:
Also, yay, SRA tests are failing.

droazen:
I've found myself sometimes having to go from `SAMSequenceDictionary` -> `SAMFileHeader` as well in other contexts -- perhaps you could add a new constructor to `SAMFileHeader` that takes a `SAMSequenceDictionary`, and call into that constructor from your new `IntervalList` constructor?

droazen:
Review complete, back to @tfenne

tfenne:
Sure @droazen I can do that instead.  I wasn't sure there were other uses for it.  I'll swap that around then merge.

PR #708, opened 2016-09-16
Yf add test empty string sam tag

yfarjoun:
### Description
Enables the reading and writing of sam/bam/cram records with empty string tags as required by https://github.com/samtools/hts-specs/issues/135

added tests to cover this functionality
### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/TextTagCodec.java
+++ b/src/main/java/htsjdk/samtools/TextTagCodec.java
@@ -41,6 +41,7 @@ import java.util.Map;
  * instance is used in multiple threads.
  */
 public class TextTagCodec {
+    // 3 fields for non-empty strings 2 fields if the string is empty.
     private static final int NUM_TAG_FIELDS = 3;
 
     /**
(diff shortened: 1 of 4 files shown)


yfarjoun:
This is deep in the codec world where I've rarely been... @tfenne @alecw I think one of you should take a look....but if not, please help me identify a good candidate for this.

src/main/java/htsjdk/samtools/TextTagCodec.java
```diff
@@ -149,12 +150,12 @@ public String encodeUntypedTag(final String tagName, final Object value) {
      */
     public Map.Entry<String, Object> decode(final String tag) {
         final int numFields = StringUtil.splitConcatenateExcessTokens(tag, fields, ':');
-        if (numFields != TextTagCodec.NUM_TAG_FIELDS) {
+        if (numFields != TextTagCodec.NUM_TAG_FIELDS && numFields != TextTagCodec.NUM_TAG_FIELDS - 1) {
```
alecw: Not a big deal, but might be simpler to test
if (numFields < TextTagCodec.NUM_TAG_FIELDS - 1)

yfarjoun: I wanted to protect against a possibility of numFields>3, given that the original code protected against that.

alecw: Hi Yossi, that's not possible given the contract of StringUtil.splitConcatenateExcessTokens, but again, not a big deal.

yfarjoun:
Thanks for looking.

PR #574, opened 2016-04-15
workaround to make publishLocal work again

lbergelson:
workaround whatever is causing #573 following advice from https://github.com/sbt/sbt-testng-interface/issues/9

fixes #573

PR #716, opened 2016-09-29
Fix for picard issue 574: cram reader should look for the index file automatically

dnikolaeva:
Fix for issue [#574](https://github.com/broadinstitute/picard/issues/574) (picard):

ViewSam works very slowly with CRAM files even if index file .crai exists. It turns out, in the case the index is not set explicitly CRAMFileReader does not try to find CRAM index file. We propose improvements in htsjdk project that make CRAMFileReader works with CRAM index automatically. To achieve this, we have expanded constructors in CRAMFileReader as follows: if indexFile is null try to find it using method findIndex in SamFiles class. All proposed improvements include unit tests to validate the results.

### DIFF
--- a/src/main/java/htsjdk/samtools/CRAMFileReader.java
+++ b/src/main/java/htsjdk/samtools/CRAMFileReader.java
@@ -26,7 +26,7 @@ import htsjdk.samtools.seekablestream.SeekableFileStream;
 import htsjdk.samtools.seekablestream.SeekableStream;
 import htsjdk.samtools.util.CloseableIterator;
 import htsjdk.samtools.util.CloserUtil;
-import htsjdk.samtools.util.CoordMath;
+import htsjdk.samtools.util.Log;
 import htsjdk.samtools.util.RuntimeEOFException;
 
 import java.io.File;
@@ -56,6 +56,8 @@ public class CRAMFileReader extends SamReader.ReaderImplementation implements Sa
     private boolean mEnableIndexMemoryMapping;
 
     private ValidationStringency validationStringency;
+
+    private final static Log log = Log.getInstance(CRAMFileReader.class);
 
     /**
      * Create a CRAMFileReader from either a file or input stream using the reference source returned by
@@ -95,6 +97,9 @@ public class CRAMFileReader extends SamReader.ReaderImplementation implements Sa
         this.cramFile = cramFile;
         this.inputStream = inputStream;
         this.referenceSource = referenceSource;
+        if (cramFile != null) {
+            mIndexFile = findIndexForFile(null, cramFile);
+        }
         getIterator();
     }
 
@@ -117,7 +122,7 @@ public class CRAMFileReader extends SamReader.ReaderImplementation implements Sa
         }
 
         this.cramFile = cramFile;
-        this.mIndexFile = indexFile;
+        mIndexFile = findIndexForFile(indexFile, cramFile);
         this.referenceSource = referenceSource;
 
         getIterator();
@@ -140,6 +145,7 @@ public class CRAMFileReader extends SamReader.ReaderImplementation implements Sa
 
         this.cramFile = cramFile;
         this.referenceSource = referenceSource;
+        mIndexFile = findIndexForFile(null, cramFile);
 
         getIterator();
     }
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -95,6 +97,7 @@ public CRAMFileReader(final File cramFile, final InputStream inputStream,
         this.cramFile = cramFile;
         this.inputStream = inputStream;
         this.referenceSource = referenceSource;
+        if (cramFile != null) initIndexFile(null);
```
cmnbroad: Since initIndexFile is called in the middle of the constructors, I would make it static (you'll have to pass in cramFile as well) and just have it return the updated indexFile, which the constructors can use to initialize the member variable. You might rename it (maybe "findIndexForFile" ?).

src/main/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -214,8 +218,25 @@ public CRAMFileReader(final InputStream stream,
     public CRAMFileReader(final File cramFile,
                           final File indexFile, final CRAMReferenceSource referenceSource,
                           final ValidationStringency validationStringency) throws IOException {
-        this(new FileInputStream(cramFile), indexFile, referenceSource, validationStringency);
+        this(new FileInputStream(cramFile),
+                wrapIndexFile(cramFile, indexFile),
```
cmnbroad: Rather than having this somewhat redundant single-use static method, I think it would be cleaner to remove it, and instead factor out the common initialization code from the constructor it's delegating to, and call that method from both constructors (this method would no longer delegate to the other one).

src/test/java/htsjdk/samtools/CRAMFileReaderTest.java
```diff
@@ -42,13 +43,17 @@
 public class CRAMFileReaderTest {
 
     private static final File TEST_DATA_DIR = new File("src/test/resources/htsjdk/samtools");
+    private static final File CRAM_FILE = new File(TEST_DATA_DIR, "cram_with_crai_index.cram");
```
cmnbroad: Can you rename this to something like CRAM_WITH_CRAI, to make the tests read a little easier.

src/test/java/htsjdk/samtools/CRAMFileReaderTest.java
```diff
@@ -146,18 +177,27 @@ public void testCRAMReader6_ReferenceRequired() throws IOException {
...
+        File indexFile = null;
+        CRAMFileReader reader = new CRAMFileReader(CRAM_FILE, indexFile, REFERENCE, ValidationStringency.STRICT);
+        Assert.assertTrue(reader.hasIndex(), "Can't find existing CRAM index.");
+    }
+
+    @Test
+    public void testCRAMReader7_ShouldUseCRAMIndex() throws IOException {
+        CRAMFileReader reader = new CRAMFileReader(CRAM_FILE, INDEX_FILE, REFERENCE, ValidationStringency.STRICT);
+        Assert.assertTrue(reader.hasIndex(), "Can't find existing CRAM index.");
+    }
```
cmnbroad: It would be good to add a negative test for each constructor that got updated just to make sure there are no problems in those code paths. There are plenty of cram files in the test folder with no index.

cmnbroad:
@dnikolaeva Thanks for the PR - a couple of requests for changes - back to you.

src/main/java/htsjdk/samtools/CRAMFileReader.java
jabbarish: done

src/test/java/htsjdk/samtools/CRAMFileReaderTest.java
jabbarish: done

src/main/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: It would be an incremental step in minimizing the number of code paths (and exception propagation strategies) for creating SeekableStreams in this code. There are several bugs in the index/iterator state management in this code (some of which can be seen [here](https://github.com/samtools/htsjdk/issues?q=is%3Aissue+is%3Aopen+label%3Acram)). Ultimately, this needs a bigger refactoring anyway. At a minimum, if you keep this method, can you rename it to something more specific - maybe getStreamForIndex ? Thx.

jabbarish: fixed. see the updated version.  We have extracted common initialization code from the two constructors into the 'initWithStreams' method and inlined mentioned above 'wrapIndexFile' method.

src/main/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -95,6 +97,7 @@ public CRAMFileReader(final File cramFile, final InputStream inputStream,
         this.cramFile = cramFile;
         this.inputStream = inputStream;
         this.referenceSource = referenceSource;
+        if (cramFile != null) mIndexFile = findIndexForFile(null, cramFile);
```
cmnbroad: Add braces and move the if-stmt body to a separate line.

jabbarish:
@cmnbroad   Sorry, forgot to place comments..  Thanks!

PR #717, opened 2016-09-30
Adds option to deep copy attributes in ReverseComplement

meganshand:
### Description

Adds a "safe" option to reverseComplement that will clone all tags as well as bases and qualities instead of reverse complementing (or reversing) them in place. The default remains "unsafe" so clones will not be made unless they are requested. 

The motivation is from https://github.com/broadinstitute/picard/pull/663 in order to avoid having to deep copy every secondary and supplementary SAMRecord in MergeBamAlignment. We need a deep copy of the attributes any time there is a single read with two alignments on different strands.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2112,7 +2112,8 @@ private String buildMessage(final String baseMessage, final boolean isMate) {
     /**
      * Note that this does a shallow copy of everything, except for the attribute list, for which a copy of the list
      * is made, but the attributes themselves are copied by reference.  This should be safe because callers should
-     * never modify a mutable value returned by any of the get() methods anyway.
+     * never modify a mutable value returned by any of the get() methods anyway. If one of the cloned records SEQ or
```
nh13: records -> record's

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2112,7 +2112,8 @@ private String buildMessage(final String baseMessage, final boolean isMate) {
...
+     * QUAL needs to be modified, a deeper copy should be made, or other precautions taken (Eg. Reverse Complement).
```
nh13: Eg -> e.g.

nh13: Remove "other precautions should be taken" since it is open ended.

src/main/java/htsjdk/samtools/SAMRecordUtil.java
```diff
@@ -39,23 +39,38 @@
 
     /**
      * Reverse-complement bases and reverse quality scores along with known optional attributes that
-     * need the same treatment.  See {@link #TAGS_TO_REVERSE_COMPLEMENT} {@link #TAGS_TO_REVERSE}
+     * need the same treatment. This does not make a deep copy of the bases, qualities or attributes:
```
nh13: comma after qualities.

nh13: Rather than talking about "deep copy", perhaps talk about "in-place" versus "makes a copy of".

src/main/java/htsjdk/samtools/SAMRecordUtil.java
```diff
@@ -39,23 +39,38 @@
...
+     * for safe reverseComplement see {@link #reverseComplement(SAMRecord, boolean)}.
```
nh13: I think in-place versus copy is better understood here than "safe".

src/main/java/htsjdk/samtools/SAMRecordUtil.java
```diff
@@ -39,23 +39,38 @@
...
     public static void reverseComplement(final SAMRecord rec) {
-        reverseComplement(rec, TAGS_TO_REVERSE_COMPLEMENT, TAGS_TO_REVERSE);
+        reverseComplement(rec, TAGS_TO_REVERSE_COMPLEMENT, TAGS_TO_REVERSE, true);
+    }
+
+    /**
+     * Reverse-complement bases and reverse quality scores along with known optional attributes that
+     * need the same treatment. Optionally makes a deep copy of the bases, qualities or attributes.
+     * See {@link #TAGS_TO_REVERSE_COMPLEMENT} {@link #TAGS_TO_REVERSE}
+     * for the default set of tags that are handled.
```
nh13: ditto for my comments above.

src/main/java/htsjdk/samtools/SAMRecordUtil.java
```diff
@@ -39,23 +39,38 @@
...
+     *
+     * @param rec Record to reverse complement.
+     * @param unsafe Setting this to false will clone all attributes, bases and qualities before changing the values.
+     */
+    public static void reverseComplement(final SAMRecord rec, boolean unsafe) {
```
nh13: unsafe -> inplace.

src/main/java/htsjdk/samtools/SAMRecordUtil.java
```diff
@@ -39,23 +39,38 @@
...
     }
 
     /**
      * Reverse complement bases and reverse quality scores. In addition reverse complement any
      * non-null attributes specified by tagsToRevcomp and reverse and non-null attributes
      * specified by tagsToReverse.
      */
-    public static void reverseComplement(final SAMRecord rec, final Collection<String> tagsToRevcomp, final Collection<String> tagsToReverse) {
-        final byte[] readBases = rec.getReadBases();
+    public static void reverseComplement(final SAMRecord rec, final Collection<String> tagsToRevcomp, final Collection<String> tagsToReverse, boolean unsafe) {
```
nh13: unsafe -> inplace.

src/main/java/htsjdk/samtools/SAMRecordUtil.java
```diff
@@ -64,8 +79,13 @@ public static void reverseComplement(final SAMRecord rec, final Collection<Strin
...
                 Object value = rec.getAttribute(tag);
                 if (value != null) {
-                    if (value instanceof byte[]) SequenceUtil.reverseComplement((byte[]) value);
-                    else if (value instanceof String) value = SequenceUtil.reverseComplement((String) value);
+                    if (value instanceof byte[]) {
+                        value = unsafe ? value : ((byte[]) value).clone();
+                        SequenceUtil.reverseComplement((byte[]) value);
+                    }
+                    else if (value instanceof String) {
+                        value = SequenceUtil.reverseComplement((String) value);
```
nh13: perhaps a comment here for the reader that `SequenceUtil.reverseComplement` is in-place with bytes and copies with `String`.

src/test/java/htsjdk/samtools/SAMRecordUtilTest.java
```diff
@@ -17,13 +57,7 @@
...
-        Assert.assertEquals(rec.getReadString(), "GTGTGTGTGT");
-        Assert.assertEquals(rec.getBaseQualityString(), "IIIIIHHHHH");
-        Assert.assertEquals(rec.getByteArrayAttribute("X1"), new byte[] {5,4,3,2,1});
-        Assert.assertEquals(rec.getSignedShortArrayAttribute("X2"), new short[] {5,4,3,2,1});
-        Assert.assertEquals(rec.getSignedIntArrayAttribute("X3"), new int[] {5,4,3,2,1});
-        Assert.assertEquals(rec.getFloatArrayAttribute("X4"), new float[] {5.0f,4.0f,3.0f,2.0f,1.0f});
-        Assert.assertEquals(rec.getStringAttribute("Y1"), "GTTTTCTTTT");
+        return(rec);
     }
+
```
nh13: extra newline

src/test/java/htsjdk/samtools/SAMRecordUtilTest.java
```diff
@@ -3,10 +3,50 @@
...
+        Assert.assertEquals(cloneOfOriginal.getByteArrayAttribute("X1"), new byte[] {5,4,3,2,1});
+        Assert.assertEquals(cloneOfOriginal.getSignedShortArrayAttribute("X2"), new short[] {5,4,3,2,1});
+        Assert.assertEquals(cloneOfOriginal.getSignedIntArrayAttribute("X3"), new int[] {5,4,3,2,1});
+        Assert.assertEquals(cloneOfOriginal.getFloatArrayAttribute("X4"), new float[] {5.0f,4.0f,3.0f,2.0f,1.0f});
+        Assert.assertEquals(cloneOfOriginal.getStringAttribute("Y1"), "GTTTTCTTTT");
+
+    }
+
+
+    public SAMRecord createTestSamRec() {
```
nh13: extra newline

src/test/java/htsjdk/samtools/SAMRecordUtilTest.java
```diff
@@ -3,10 +3,50 @@
...
+        Assert.assertEquals(rec.getSignedShortArrayAttribute("X2"), new short[] {5,4,3,2,1});
+        Assert.assertEquals(rec.getSignedIntArrayAttribute("X3"), new int[] {5,4,3,2,1});
+        Assert.assertEquals(rec.getFloatArrayAttribute("X4"), new float[] {5.0f,4.0f,3.0f,2.0f,1.0f});
+        Assert.assertEquals(rec.getStringAttribute("Y1"), "GTTTTCTTTT");
+    }
+
+    @Test public void testSafeReverseComplement() throws CloneNotSupportedException {
+        final SAMRecord original = createTestSamRec();
+        final SAMRecord cloneOfOriginal = (SAMRecord) original.clone();
+        //Runs a safe reverseComplement
```
nh13: copy vs in-place.

cmnbroad:
My 2c: I think we should err on the side of making apis like this be safe-by-default. Since reverseComplement is a SAMRecord mutator anyway, it could be moved to SAMRecord as a method with reversed polarity so it defaults to safe, with an overload with an opt-in arg for "in place". The existing method could then delegate to SAMRecord (and preferably be marked as deprecated - after which the entire class could be removed.)

nh13:
@cmnbroad it's unsafe for performance reasons, and I would be concerned about such a change impacting downstream tools which require performance.  It's a good philosophy in general, but may be beyond the scope of this PR.

cmnbroad:
@nh13 Agreed - I was suggesting that we keep the existing behavior as is, but deprecate the method and change the implementation to call into the new (SAMRecord method) that implements in-place as opt-in (and have the existing method opt-in to in-place to maintain the current behavior). Its a pretty small incremental changed from the current PR (just moving the code). Anyway, just a suggestion that I think cleans up the api nicely.

meganshand:
@nh13 I addressed your comments in the first commit and refactored according to @cmnbroad's suggestion in the second commit.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2249,5 +2254,127 @@ public final Object removeTransientAttribute(final Object key) {
         if (this.transientAttributes != null) return this.transientAttributes.remove(key);
         else return null;
     }
-}
 
+    /**
+     * Reverse-complement bases and reverse quality scores along with known optional attributes that
+     * need the same treatment. Changes made after making a copy of the bases, qualities,
+     * and attributes. If in-place update is needed use {@link #reverseComplement(SAMRecord, boolean)}.
```
yfarjoun: make sure it's clear that only some of the attributes are copied

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2248,5 +2254,127 @@ public final Object removeTransientAttribute(final Object key) {
...
+
+        // Deal with tags that need to be reverse complemented
+        if (tagsToRevcomp != null) {
+            for (final String tag: tagsToRevcomp) {
+                Object value = rec.getAttribute(tag);
+                if (value != null) {
+                    if (value instanceof byte[]) {
+                        value = inplace ? value : ((byte[]) value).clone();
+                        SequenceUtil.reverseComplement((byte[]) value);
+                    }
```
yfarjoun: } else if {

here and below

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2248,5 +2254,127 @@ public final Object removeTransientAttribute(final Object key) {
...
+                    else if (value instanceof String) {
+                        //SequenceUtil.reverseComplement is in-place for bytes but copies Strings since they are immutable.
+                        value = SequenceUtil.reverseComplement((String) value);
+                    }
+                    else throw new UnsupportedOperationException("Don't know how to reverse complement: " + value);
```
yfarjoun: need { }

src/test/java/htsjdk/samtools/SAMRecordUtilTest.java
```diff
@@ -24,7 +24,7 @@
     @Test public void testSafeReverseComplement() throws CloneNotSupportedException {
         final SAMRecord original = createTestSamRec();
         final SAMRecord cloneOfOriginal = (SAMRecord) original.clone();
-        //Runs a safe reverseComplement
+        //Runs a copy (rather than in-place) reverseComplement
         SAMRecordUtil.reverseComplement(cloneOfOriginal, Arrays.asList("Y1"), Arrays.asList("X1", "X2", "X3", "X4", "X5"), false);
 
         Assert.assertEquals(original.getReadString(), "ACACACACAC");
```
yfarjoun: could you expand this test using a data-provider to show that if false->true then the assertion failes?

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2249,5 +2254,127 @@ public final Object removeTransientAttribute(final Object key) {
...
+    /**
+     * Reverse-complement bases and reverse quality scores along with known optional attributes that
+     * need the same treatment. Optionally makes a copy of the bases, qualities or attributes instead
+     * of altering them in-place. See {@link #TAGS_TO_REVERSE_COMPLEMENT} {@link #TAGS_TO_REVERSE}
+     * for the default set of tags that are handled.
+     *
+     * @param rec Record to reverse complement.
+     * @param inplace Setting this to false will clone all attributes, bases and qualities before changing the values.
+     */
+    public static void reverseComplement(final SAMRecord rec, boolean inplace) {
```
yfarjoun: similarly here, shouldn't it be non-static and only take inplace as an argument?

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2249,5 +2254,127 @@ public final Object removeTransientAttribute(final Object key) {
...
+     * See {@link #TAGS_TO_REVERSE_COMPLEMENT} {@link #TAGS_TO_REVERSE}
+     * for the default set of tags that are handled.
+     */
+    public static void reverseComplement(final SAMRecord rec) {
+        reverseComplement(rec, false);
+    }
+
+    /**
+     * Reverse-complement bases and reverse quality scores along with known optional attributes that
+     * need the same treatment. Optionally makes a copy of the bases, qualities or attributes instead
```
yfarjoun: scratch that. I misunderstood that comment.

src/main/java/htsjdk/samtools/SAMRecord.java
yfarjoun: still needs {} for else.

src/main/java/htsjdk/samtools/SAMRecordUtil.java
```diff
@@ -33,97 +33,41 @@
 /**
  * @author alecw@broadinstitute.org
  */
+@Deprecated
```
yfarjoun: in a comment, point users to what they should be using.

yfarjoun:
back to you @meganshand. this looks nice, just a few comments and then I think we are good to go. Given that @nh13 looked at it first, it would be good if he could sign-off on it as well...

meganshand:
@yfarjoun @nh13 I think I've caught all of your comments.

yfarjoun:
I think this is fine. :+1: from me. I'll take @nh13's "this makes sense" to be a tentative thumbs up (conditional on all the work you did afterwards). but to be sure, let's give him till end of next week to respond, and if we don't hear from him till then, merge.

yfarjoun:
Let's move forward with this PR @meganshand as you have resolved the issues that @nh13 raised. and he has been silent for a while.

yfarjoun:
done.

PR #724, opened 2016-10-12
Support use of Path in ParsingUtils, SeekableStreamFactory, AbstractVCFCodec

tomwhite:
### Description

Added support for java.util.nio.Path for some more code pathways.

This is to support loading of Features with Spark in GATK.
### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/IOUtil.java
+++ b/src/main/java/htsjdk/samtools/util/IOUtil.java
@@ -48,14 +48,20 @@ import java.io.OutputStreamWriter;
 import java.io.Reader;
 import java.io.Writer;
 import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.Charset;
+import java.nio.file.FileSystemNotFoundException;
+import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Scanner;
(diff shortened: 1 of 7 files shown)


src/test/java/htsjdk/tribble/util/ParsingUtilsTest.java
```diff
@@ -143,6 +164,18 @@ private void tstExists(String path, boolean expectExists) throws IOException{
     }
 
     @Test
```
lbergelson: I think we need a way of testing non-file URI's.  I'm not sure the best way to go about it.  One possibility would be to include one of the alternative filesystem providers as a test dependency.  Either that or maybe we could mock one.  There is a [google in memory filesystem] that might make a good test for the non-file based file systems.

lbergelson: @droazen What do you think about including a new test dependency for providing alternative paths.  It might help us avoid problems like the index one we saw before.

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -943,6 +949,43 @@ public static String slurp(final InputStream is, final Charset charSet) {
...
+    /**
+     * Converts the given URI to a {@link Path} object. If the filesystem cannot be found in the usual way, then attempt
+     * to load the filesystem provider using the thread context classloader. This is needed when the filesystem
+     * provider is loaded using a URL classloader (e.g. in spark-submit).
+     *
+     * @param uriString the URI to convert
+     * @return the resulting {@code Path}
+     * @throws IOException an I/O error occurs creating the file system
+     */
+    public static Path getPath(String uriString) throws IOException {
```
lbergelson: could you replace the use of `Paths.get` in [`SamInputResource.UrlInputResource`](https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/samtools/SamInputResource.java#L351) with a call to this instead?  Just in case...

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -943,6 +949,43 @@ public static String slurp(final InputStream is, final Charset charSet) {
...
+        URI uri = URI.create(uriString);
+        try {
+            return uri.getScheme() == null ? Paths.get(uriString) : Paths.get(uri);
+        } catch (FileSystemNotFoundException e) {
+            ClassLoader cl = Thread.currentThread().getContextClassLoader();
+            if (cl == null) {
+                throw e;
+            }
+            return FileSystems.newFileSystem(uri, new HashMap<>(), cl).provider().getPath(uri);
```
tomwhite: I did try running this on HDFS with Spark a while ago and it worked. There's a fix for the GCS issue from @jean-philippe-martin here: https://github.com/GoogleCloudPlatform/google-cloud-java/pull/1470

src/main/java/htsjdk/samtools/util/IOUtil.java
tomwhite: Done

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -943,6 +949,43 @@ public static String slurp(final InputStream is, final Charset charSet) {
...
+     * provider is loaded using a URL classloader (e.g. in spark-submit).
+     *
+     * @param uriString the URI to convert
+     * @return the resulting {@code Path}
+     * @throws IOException an I/O error occurs creating the file system
+     */
+    public static Path getPath(String uriString) throws IOException {
+        URI uri = URI.create(uriString);
+        try {
+            return uri.getScheme() == null ? Paths.get(uriString) : Paths.get(uri);
```
tomwhite: Done

src/test/java/htsjdk/tribble/util/ParsingUtilsTest.java
tomwhite: Good idea. I've added a test dependency on https://github.com/google/jimfs, and a few tests that use non-file URIs that exercise the NIO file path.

tomwhite:
@lbergelson thanks for the review. I've addressed all your comments, so please take another look.  (See https://github.com/broadinstitute/gatk/pull/2344 for the companion GATK issue.)

src/main/java/htsjdk/samtools/util/IOUtil.java
jean-philippe-martin: @lbergerson: yes this was a bug, it's fixed now.

@tomwhite: Does this mean that on Spark after someone calls getPath on a gs:// URI, the fs provider will be loaded and Path.get will work correctly? Because that would be sweet.

src/test/java/htsjdk/tribble/util/ParsingUtilsTest.java
jean-philippe-martin: :thumbsup:

lbergelson:
@tomwhite Changes look good.  Tests with jimfs are a good addition.  It would be good to have a test that reads a vcf with an index from jimfs so we make sure we avoid the issue we had with bam indexes.  I want to get this in though, so I'm going to merge and we can add one after the fact.

PR #725, opened 2016-10-12
seek() checks for negative position argument value

typekpb:


src/main/java/htsjdk/samtools/seekablestream/ByteArraySeekableStream.java
```diff
@@ -27,7 +27,11 @@ public long position() throws IOException {
 
     @Override
     public void seek(long position) throws IOException {
-        this.position = position;
+        if (position < 0) {
+			throw new IOException("Negative seek offset");
+		} else {
+			this.position = position;
+		}
```
lbergelson: This is very nitpicky, but could you use spaces instead tabs so that the formatting matches the rest of the file?

src/main/java/htsjdk/samtools/seekablestream/ByteArraySeekableStream.java
```diff
@@ -27,7 +27,11 @@ public long position() throws IOException {
 
     @Override
     public void seek(long position) throws IOException {
-        this.position = position;
+        if (position < 0) {
+			throw new IOException("Negative seek offset");
```
lbergelson: This should probably be an `IllegalArgumentException`, and I would make the exception include the requested position as well.

```java
throw new IllegalArgumentException("Cannot seek to a negative position, position=" + position + ".");
```

lbergelson:
@typekpb Thanks for this bug fix!  I had a few comments.  👍  to merge when those are addressed.

yfarjoun:
@lbergelson could you take this PR over the finishline?

yfarjoun:
closing in favor of #883

PR #726, opened 2016-10-18
Added getDescription() in VCFFilterHeaderLine

lindenb:
### Description

Using a VCFFilterHeaderLine as a java-bean, I was surprised to see that there is no (?) public/easy way to get the "Description" of a VCFFilterHeaderLine . This PR adds the method `getDescription()` ; It returns the ID if the description is empty. 
I've also added some `final` keywords and the `serialVersionUID`
### Checklist
- [X] Code compiles correctly
- [X] All tests passing

### DIFF
--- a/src/main/java/htsjdk/variant/vcf/VCFFilterHeaderLine.java
+++ b/src/main/java/htsjdk/variant/vcf/VCFFilterHeaderLine.java
@@ -27,12 +27,16 @@ package htsjdk.variant.vcf;
 
 import java.util.Arrays;
 
+import htsjdk.samtools.util.StringUtil;
+
 /**
  * @author ebanks
  * 
  * A class representing a key=value entry for FILTER fields in the VCF header
  */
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/variant/vcf/VCFFilterHeaderLine.java
```diff
@@ -58,12 +62,21 @@ public VCFFilterHeaderLine(String name) {
...
         return true;
     }
+    
+    /**
+     * get the "Description" field
+     * @return the "Description" or the Filter ID if the description is null or empty 
+     */
+    public String getDescription() {
+        final String desc = getGenericFieldValue("Description");
+        return ( StringUtil.isBlank(desc) ? getID() : desc );
```
yfarjoun: I think that this logic should not be in htsjdk. getDescription() should return null or blank if there's no decription. the user can choose what to do with that information.

yfarjoun:
This looks fine to me. I'll leave it for a few more days to see if there are other comments and merge it otherwise.

PR #727, opened 2016-10-21
updating the gradle version from 2.13 to 3.2.1

lbergelson:
### Description

updating from gradle 2.13 -> 3.1
adding a settings.gradle file to define the project name 

these changes make it possible to use htsjdk in a composite gradle build with other projects

This will make changes to downstream projects that require updates to htsjdk easier to develop and test.  Experimental builds of intellij are including a composite build feature and these changes are necessary to use it.
### Checklist
- [x] Code compiles correctly
- [x] All tests passing


yfarjoun:
will this affect downstream projects (like picard) that have not upgraded their gradle?

lbergelson:
@yfarjoun It won't affect any downstream projects that use the artifacts produced by the build.  If they actually call into the gradle build directly there's a possibility that it will be affected.  I would be surprised if there were any problems though, I don't think we make use of any features that would have  changed between 2 and 3.  

Once all the projects in the ecosystem are on 3.1+ it will let us do some nice things, like have intellij manage them as if they were a single build.  (I.e. both projects open in intellij, change something in htsjdk and see immediate changes in picard like it used to be, but still keep the releases separated.)

lbergelson:
I'm running a build of picard master against this branch just to be check . https://gatk-jenkins.broadinstitute.org/job/picard-on-htsjdk-master/390/

lbergelson:
@yfarjoun 👍 / 👎 ?

yfarjoun:
is 3.1 "cutting edge" I want to make sure we are using a stable version of gradle.

lbergelson:
3.1 is actually outdated now.  We should probably update to 3.2.1 which is supposed to make IDE integration performance better.  3.1 is the minimum version necessary to use the incredibly useful composite build support though.

yfarjoun:
why did coverage go up?

lbergelson:
@yfarjoun Not sure... Coveralls bug is my assumption.  I think coveralls is showing coverage based on the push branch instead of the pull request one.  I wouldn't worry about.

lbergelson:
(part of why I want to switch to codecov, #764 )

lbergelson:
@yfarjoun picard-private tests pass with this

PR #728, opened 2016-10-21
Yf fix test removed deprecated methods relating to nm

yfarjoun:
### Description

**Bug** in SequenceUtil.calculateMdAndNmTags() whereby bases with different case (upper/lower) in reference or read are considered different despite the sam-spec saying that regarding SEQ: "No assumptions can be made on the letter cases." This has caused validatino problems for files that had their NM values set via SequenceUtil.calculateMdAndNmTags() (which takes case into account) and validated via SequenceUtil.calculateSamNmTag() (which, correctly, ignores case)

This PR fixes calculateMdAndNmTags by the same mechanism as used by calculateSamNmTag, adds a test to both functions (also testing MD at the same time) and gives variables in calculateMdAndNmTags more meaningful names.

Also, while I'm there I removed several deprecated functions that are not used.

I would like to nominate @jacarey for review as he seems to be the original author of calculateSamNmTag.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


jacarey:
:+1:

yfarjoun:
waiting for tests to pass, then merging.

PR #732, opened 2016-10-27
Enable reading of CRAM Header with no (explicit) reference

### LINKED ISSUE #731: SAMSequenceDictionaryExtractor doesn't handle CRAMs
Just adding the cram extension to the list of extensions used by the SAM handler won't work unless a method overload is added that accepts a reference. Alternative it would be possible to have a custom handler that bypasses the SAMReader and just retrieves the header directly, since the reference isn't needed to just do that.

yfarjoun:
### Description

fixes #731 by allowing a CRAM file to be opened using a trivial referenceSource which has been added to the SamReaderFactory default settings.
- added test cases to show that DictionaryExtractor works and also that reading the header of a cram file works.
- fixed old tests that broke as they were getting a new (=unexpected) exception when opening a cram file for reading without a reference.
- continued the march of java8 into the code
- suppressed some intelliJ code inspection warnings and fixed a javadoc error.
### Checklist
- [x] Code compiles correctly
- [x] All tests passing


src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -135,6 +137,7 @@ private SamReaderFactoryImpl(final EnumSet<Option> enabledOptions, final Validat
             this.samRecordFactory = samRecordFactory;
             this.validationStringency = validationStringency;
             this.customReaderFactory = CustomReaderFactory.getInstance();
+            this.referenceSource = EMPTY_REFERENCE;
```
cmnbroad: @yfarjoun I think this will mask the fallback chain thats in place in the default reference source implementation (the code that uses environment variables to try the ENA download service or DEFAULT_FASTA settings - I don't think we have tests for those). Its currently established through the calls in SamReaderFactory to getDefaultCRAMReferenceSource. When we switched to using an interface based pattern for reference source for IGV, I intended to break those out into a separate implementations for each type (ENA ref service, DEFAULT_FASTA, user-supplied file) so we could establish which one to use once up front, but never did. For now they're still all mashed together in the one implementation.

Issue #366, opened 2015-11-04
SAMWriterFactory.makeSAMOrBAMWriter creates BAM files with a ".cram" extension


cmnbroad:
If you pass in a filename with a .cram extension, makeSAMOrBAMWriter will create a BAM file with a .cram extension and return a BAMFileWriter, which is probably not what was intended. This was causing many Hellbender tools to create BAM files with the .cram extension. Using makeWriter instead will remedy this (as it will create a real .cram file) but makeSAMOrBAMWriter should check the extension and either redirect to makeCRAMWriter or reject the request.

cmnbroad:
This should probably throw if its handed a .cram extension since it doesn't accept a reference, so in the CRAM case it would only be useful with the download service or an embedded reference.

tfenne:
I admit I haven't followed this all too closely, but I thought the reason for the entry in Defaults for `reference_fasta` was so that there was a way to specify a default reference fasta for CRAM so as not to need to update every single API?  I.e. could one call this to make a CRAM file in a JVM that was started with -Dsamjdk.reference_fasta=my.fasta?

cmnbroad:
Yes, we could fix this so that if you had the environment variable set (or
the download service enabled) it could pick those up. We'd still need to
make the change to go down the CRAM path though. However, in the (common ?)
case where neither of those is true, the downstream failures may not be
obvious. So maybe we should have this check for a .cram extension, and
throw only if neither of default reference or download are set.

PR #734, opened 2016-10-27
CRAM format support

SilinPavel:
**1. Picard Tools cannot use CRAM file as input for metrics**

Current version of Picard Tools cannot use CRAM file as input for metrics that don’t pass reference sequence directly to SamReaderFactory, despite the fact that each CommandLineProgram has an input parameter REFERENCE_SEQUENCE. 
For example, MarkDuplicates experiences such a problem: (https://github.com/broadinstitute/picard/issues/433, https://github.com/broadinstitute/picard/issues/361) 

To resolve this problem we suggest several changes in htsjdk: as @tfenne suggested (https://github.com/broadinstitute/picard/issues/433#issuecomment-183141150)  we propose to add a static field  referenceSequenceFromCmd with a static setter in SamReaderFactory. It allows Picard CommandLineProgram  to set field referenceSequenceFromCmd when REFERENCE_SEQUENCE is non-null and/or Defaults.REFERENCE_FASTA is set.

**2. CRAMFileReader throws exceptions when there is no reference.**

In current implementation CRAMFileReader throws exceptions "A valid CRAM reference was not supplied" or "A reference is required for CRAM readers" exceptions when there is no reference. However, there can be cases when reference is not really needed (CRAM with all reads unmapped).

Our proposition is not to throw any exception when no reference was defined in getDefaultCRAMReferenceSource, but just return empty ReferenceSource object. With this approach CramFileReader will only throw exception in runtime and only if there is no reference when it is really needed (CramFileReader’s constructors do not throw any exception when null reference is passed). Tests were fixed to fit these cases.

All these changes are needed for corresponding changes in PicardTools, that will provide option to handle CRAM file as input/output correctly (https://github.com/samtools/htsjdk/issues/366). We are going to publish the corresponding pull request to Picard Tools shortly.  
- [x] Code compiles correctly
- [x] All tests passing


jabbarish:
@yfarjoun @tfenne Take a look at this PR from our team, please.

cmnbroad:
I'm unclear on the need for the SamReaderFactory changes outlined for #1 above. Picard has to be changed one way or another to make CRAM work; those changes should be trivial using the existing htsjdk APIs. That seems preferable to me to adding a new API and new state to SamReaderFactory. As for #2, we probably need something like that based on all the recent discussion around not requiring a reference.

bmkodoo:
@cmnbroad
Yes we agree that it's unclear. There are alternative ways:

1. To pass reference using existing htsjdk API is `SamReaderFactory#referenceSequence` factory chain method: to add this call to every metric and oblige to use it in future each time `SAMReader` opens.
2. Our current modification call `getDefaultCRAMReferenceSource` fallback chain if no reference is passed. So, reference can be passed to SamReaderFactory by `samjdk.reference_fasta` property.

@yfarjoun @tfenne, what do you think about those alternatives?

ekazachkova:
As we can see, Picard has 11 classes which instance SamReader without reference set. They are:

- AbstractMarkDuplicatesCommandLineProgram.java
- BamToBfqWriter.java
- CheckFingerprint.java
- CollectIndependentReplicateMetrics.java
- CollectJumpingLibraryMetrics.java
- CollectOxoGMetrics.java
- CollectRrbsMetrics.java
- CollectWgsMetricsFromQuerySorted.java
- FingerprintChecker.java
- LiftOverIntervalList.java
- SplitSamByLibrary.java

These classes have to be changed by adding reference sequence call:
```
SamReader in = SamReaderFactory.makeDefault()
            .referenceSequence(REFERENCE_SEQUENCE)
            .open(SamInputResource.of(INPUT));
```

Also, convention calling `referenceSequence` each time SamReader is opened will be needed. 

If those conditions are acceptable, there is no need for the SamReaderFactory changes.

yfarjoun:
I think that this is fine. we might find a better solution later, but for now, this is fine.

ZLyanov:
@yfarjoun Now I'm confused :)   Could you clarify please, what you agreed with:  the changes in this PR or the other approach, described above in the comments by @bmkodoo  & @ekazachkova?

yfarjoun:
I think that it would be better to start with the small changes  to the
tools that need it before making such a big change to the defaults... I
agree with @chrisnorman about needing to make the changes required to read
a cram header.

ZLyanov:
Hi!  Ok, let's summarize the discussion.

First, we still think that using default static reference source in `SamReaderFactory` has some quite important advantages: 

- this approach allows any tool that uses htsjdk not to care about setting reference source each time a reader is opened, 
- it's fully backward compatible and it requires only few lines changed in Picard tools (in `CommandLineProgram` class). 

But if our changes are not acceptable, we'll prepare new changes in Picard tools and htsjdk, as @ekazachkova mentioned above. 

There are two points to consider for these changes:

- We have already published the [PR](https://github.com/broadinstitute/picard/pull/683) that adds CRAM writing support to all relevant metrics. We can update this PR or create a new one, but the majority of the classes in these PRs will intersect. Which way would you prefer: a new PR or updating the existing one?
- Some picard's metrics create readers via htsjdk classes, for instance `DownsampleSam` uses `htsjdk.samtools.DownsamplingIteratorFactory` to open an iterator. So some changes for reference source setting will be required in htsjdk too.

lbergelson:
I'm not a fan of the static defaults.  They can cause big problems in a multiprocess environment like spark.  They also tend to be opaque to the user and don't tend to interact well with other ways of injecting arguments into programs.

yfarjoun:
I think that this PR should be closed at this point since the picard PRs supercede this.

please reopen if I'm mistaken.

Issue #513, opened 2016-03-04
NPE when validating a CRAM file with a .fasta reference with no .dict file


cmnbroad:
If you pass in a .fasta file for which there is no corresponding .dict file to SamFileValidator you'll get an NPE during NM tag validation. Even though the bug isn't specific to CRAM files, I'm tagging it with the CRAM tag since its way more likely to happen when you're required to pass a reference, as is the case for CRAM files.

java.lang.NullPointerException
    at htsjdk.samtools.reference.ReferenceSequenceFileWalker.get(ReferenceSequenceFileWalker.java:87)
    at htsjdk.samtools.SamFileValidator.validateNmTag(SamFileValidator.java:462)
    at htsjdk.samtools.SamFileValidator.validateSamRecordsAndQualityFormat(SamFileValidator.java:288)
    at htsjdk.samtools.SamFileValidator.validateSamFile(SamFileValidator.java:200)
    at htsjdk.samtools.SamFileValidator.validateSamFileVerbose(SamFileValidator.java:160)
    at org.broadinstitute.hellbender.tools.picard.sam.ValidateSamFile.doWork(ValidateSamFile.java:131)
    at org.broadinstitute.hellbender.cmdline.CommandLineProgram.runTool(CommandLineProgram.java:98)
    at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMainPostParseArgs(CommandLineProgram.java:146)
    at org.broadinstitute.hellbender.cmdline.PicardCommandLineProgram.instanceMain(PicardCommandLineProgram.java:61)
    at org.broadinstitute.hellbender.Main.instanceMain(Main.java:66)
    at org.broadinstitute.hellbender.Main.instanceMain(Main.java:73)
    at org.broadinstitute.hellbender.CommandLineProgramTest.runCommandLine(CommandLineProgramTest.java:72)
    at org.broadinstitute.hellbender.tools.picard.sam.ValidateSamFileIntegrationTest.testJunkFile(ValidateSamFileIntegrationTest.java:76)

vadimzalunin:
fixed https://github.com/samtools/htsjdk/pull/735

PR #735, opened 2016-10-28
Fix for issue #513: NPE in ValidateSamFile with CRAM

ekazachkova:
# **Fix for issue [#513](https://github.com/samtools/htsjdk/issues/513):**

Current Htsjdk implementation throws NullPointerException during NM tag validation in SamFileValidator because getSequenceDictionary() method in ReferenceFileWalker returns null without .dict file. We propose check result of getSequenceDictionary() for null. If this method returns null don't throw any exceptions and continue processing as if reference isn't indexed.  Also, we added new unit test to check code correctness.
### Checklist
- [x] Code compiles correctly
- [x] All tests passing


ZLyanov:
@yfarjoun could you take a look, please?

droazen:
Can you review?

src/test/java/htsjdk/samtools/ValidateSamFileTest.java
```diff
@@ -64,6 +65,18 @@ public void testValidSamFile() throws Exception {
     }
 
     @Test
+    public void testValidCRAMFileWithoutSeqDict() throws Exception {
+        final SamReader samReader = SamReaderFactory.
+                makeDefault().
+                validationStringency(ValidationStringency.SILENT).
+                referenceSequence(new File(TEST_DATA_DIR, "nm_tag_validation.fa")).
+                open(new File(TEST_DATA_DIR, "nm_tag_validation.cram"));
+        final ReferenceSequenceFile reference = new FastaSequenceFile(new File(TEST_DATA_DIR, "nm_tag_validation.fa"), true);
```
ekazachkova: @vadimzalunin I can not understand why do we need this check: this test will always fail since cram reader can't work with reference without index. Could you please clarify your comment?

ekazachkova:
Also, we have discovered that our test data aren't valid: cram file is not paired with reference. So we have regenerated them.

vadimzalunin:
Could you please check the CRAM file? Reference MD5 checksums don't seem to match.

> cat nm_tag_validation.fa | tail -n +2 | tr -d '\n' | md5sum
f8c08a4411f07717451464d546b3706d  -

> samtools view -H nm_tag_validation.cram | grep '^@SQ' | grep -P 'M5:\w+' -o
M5:351f64d4f4f9ddd45b35336ad97aa6de

ekazachkova:
ooops, @vadimzalunin thanks for remark

yfarjoun:
verbal thumbs from vadim. 👍

PR #743, opened 2016-11-16
Increased test coverage for SamFileValidator

ekazachkova:
### Increased test coverage for SamFileValidator

We represent PR that contains increased test coverage for `SamFileValidator` class with the following changes in `ValidateSamFileTest`:
 - Make simple refactoring for `executeValidation` method
 - Add new test for `validateBamFileTermination` method
 - Add new test and new data provider for validating E2 tag and U2 tag process correctness.

Also, we suggest the following changes in `SamFileValidator` class:
 - Initialize `SamFileValidator`'s fields by constructor instead of using default values
 - Add getter for `errorsByType` field with package access for testing.

### Checklist

- [x] Code compiles correctly
- [x] New tests covering changes
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SamFileValidator.java
+++ b/src/main/java/htsjdk/samtools/SamFileValidator.java
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2009 The Broad Institute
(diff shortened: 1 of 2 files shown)


yfarjoun:
Thanks! 👍

PR #744, opened 2016-11-17
Fix for issue #432 in Picard Tools. CreateSequenceDictionary stalls indefinitely with large genomes 

SilinPavel:
We introduce two pull requests to Htsjdk and Picard Tools(https://github.com/broadinstitute/picard/pull/687) to resolve the issue https://github.com/broadinstitute/picard/issues/432 .

Original version of CreateSequenceDictionary creates and stores all reference sequences entirely in memory in a SAMHeader object and then writes it to a file.

In order to check sequence's names for uniqueness, they are added to a HashSet. This might lead to "Out of memory" exceptions for files with a large number of sequences .

**Changes in Htsjdk:**

In our implementation we introduce a new "on fly" codec SAMSequenceDictionaryCodec, which encodes each sequence and directly writes it to the Dictionary file. This allows us not to store sequences in memory.

**Changes in Picard Tools:**

CreateSequenceDictionary now uses new SAMSequenceDictionaryCodec to write dictionary file.

SortingCollection is used instead of a HashSet to verify the sequences names uniqueness , this allows us not to keep all the names in memory since identical names will be stored in the collection one after another and we can find duplicates in one pass. 

Please pay attention, that now CreateSequenceDictionary doesn't write HD line, since we think that it is not required, SequenceDictionary must contain only SQ line, not HD. If our assumption is wrong, correct us, please.

 

 

So, if we look at the plot "GC Times" from JMC: FlightRecorder, we will see that standard version of CreateSequenceDictionary have a big trouble with a large amount of records (several millions):

**The plot "GC Times" for a  Standard version:**
 
selection_041
selection_043

**The plot "GC Times" for a fixed version:**
selection_040
selection_042

How we can see, now CreateSequenceDictionary processes a big reference without troubles, for example time of execution for file from issue https://github.com/broadinstitute/picard/issues/432 is:

```
real 762.83
user 626.94
sys 67.02
```

Full GC reports (jmc flight recorder):
[JMC FR logs.zip](https://github.com/broadinstitute/picard/files/597773/JMC.FR.logs.zip)

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMSequenceDictionaryCodec.java
+++ b/src/main/java/htsjdk/samtools/SAMSequenceDictionaryCodec.java
@@ -1,17 +1,19 @@
 package htsjdk.samtools;
 
 import htsjdk.samtools.util.LineReader;
+import htsjdk.samtools.util.RuntimeIOException;
 
 import java.io.BufferedWriter;
+import java.io.IOException;
 import java.io.Writer;
 
 public class SAMSequenceDictionaryCodec {
 
     private final SAMTextHeaderCodec codec;
 
-    public SAMSequenceDictionaryCodec(final Writer writer) {
+    public SAMSequenceDictionaryCodec(final BufferedWriter writer) {
         codec = new SAMTextHeaderCodec();
-        codec.setWriter(new BufferedWriter(writer));
+        codec.setWriter(writer);
     }
 
     public void encodeSQLine(final SAMSequenceRecord sequenceRecord) {
@@ -24,6 +26,7 @@ public class SAMSequenceDictionaryCodec {
 
     public void encode(final SAMSequenceDictionary dictionary) {
         dictionary.getSequences().forEach(this::encodeSQLine);
+
     }
 
     public void setValidationStringency(final ValidationStringency validationStringency) {
--- a/src/main/java/htsjdk/samtools/SAMSequenceDictionaryCodec.java
+++ b/src/main/java/htsjdk/samtools/SAMSequenceDictionaryCodec.java
@@ -55,8 +55,7 @@ import java.io.BufferedWriter;
  * try(BufferedWriter writer = new BufferedWriter(new FileWriter("path/to/file"))) {
  *      SAMSequenceDictionaryCodec codec = new SAMSequenceDictionaryCodec(writer);
  *
- *      //we have complete SAMSequenceDictionary, so encode header line and {@link SAMSequenceDictionary}.
- *      codec.encodeHeaderLine(false);
+ *      //we have complete {@link SAMSequenceDictionary}, so just encode it.
  *      codec.encode(dict);
  *}
  *
@@ -105,6 +104,7 @@ public class SAMSequenceDictionaryCodec {
      * @param dictionary object to be converted to text.
      */
     public void encode(final SAMSequenceDictionary dictionary) {
+        codec.encodeHeaderLine(false);
         dictionary.getSequences().forEach(this::encodeSequenceRecord);
     }
 
--- a/src/main/java/htsjdk/samtools/SAMSequenceDictionaryCodec.java
+++ b/src/main/java/htsjdk/samtools/SAMSequenceDictionaryCodec.java
@@ -31,14 +31,45 @@ import java.io.BufferedWriter;
  * "On the fly" codec SAMSequenceDictionaryCodec.
  * Encodes each sequence and directly writes it to the Dictionary file.
  *
+ * To use this class you should provide BufferedWriter to it, and so you should close it as you stop using this class.
+ * You can work with this class as shown below.
+ *
+ * Example of using this class:
+ *
+ * List<SAMSequenceRecord> dict = ...;
+ *
+ * //open BufferedReader and close in try-with-resources
+ * try(BufferedWriter writer = new BufferedWriter(new FileWriter("path/to/file"))) {
+ *      SAMSequenceDictionaryCodec codec = new SAMSequenceDictionaryCodec(writer);
+ *
+ *      //we have list of sequences, so encode header line and after that encode each sequence
+ *      codec.encodeHeaderLine(false);
+ *      dict.forEach(codec::encodeSequenceRecord);
+ *}
+ *
+ * or
+ *
+ * SAMSequenceDictionary dict = ...;
+ *
+ * //open BufferedReader and close in try-with-resources
+ * try(BufferedWriter writer = new BufferedWriter(new FileWriter("path/to/file"))) {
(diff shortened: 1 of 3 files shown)


yfarjoun:
1. This is great performance! Thanks for fixing this.

2. The sequence dictionary should indeed contain the @HD line. please reinstate it.

src/main/java/htsjdk/samtools/SAMSequenceDictionaryCodec.java
```diff
@@ -0,0 +1,79 @@
...
+
+import htsjdk.samtools.util.LineReader;
+import htsjdk.samtools.util.RuntimeIOException;
+
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.io.Writer;
+
+/**
+ * "On fly" codec SAMSequenceDictionaryCodec.
```
yfarjoun: "on the fly"

src/main/java/htsjdk/samtools/SAMTextHeaderCodec.java
```diff
@@ -437,8 +441,8 @@ private void writeHDLine(final boolean keepExistingVersionNumber) {
         println(StringUtil.join(FIELD_SEPARATOR, fields));
     }
 
-    private void writeSQLine(final SAMSequenceRecord sequenceRecord) {
-        final int numAttributes =sequenceRecord.getAttributes() != null ? sequenceRecord.getAttributes().size() : 0;
+    void writeSQLine(final SAMSequenceRecord sequenceRecord) {
```
yfarjoun: hmmm. I'm not so happy about the access change of this internal function. could you perhaps wrap it with a public function that also checks that the writer is not null and is more cognizant of it's package-protected status? For example, a java-doc would be good that would inform the assumptions in place (like the existence of a writer..)

SilinPavel:
@yfarjoun I have done what you asked for. Take a look, please!

src/test/java/htsjdk/samtools/SAMSequenceDictionaryCodecTest.java
```diff
@@ -0,0 +1,76 @@
...
+        writer = new StringWriter(10);
+        bufferedWriter = new BufferedWriter(writer);
+        codec = new SAMSequenceDictionaryCodec(bufferedWriter);
+    }
+
+    @Test
+    public void testEncodeDecode() throws Exception {
+        LineReader reader = null;
+        try {
+            codec.encodeHeaderLine(false);
```
yfarjoun: I think that this line is throwing a NPE since mFileHeader is null...how does encodeHeaderLine when mFileHeader hasn't been initialized?

src/test/java/htsjdk/samtools/SAMSequenceDictionaryCodecTest.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.Random;
+
+import static org.testng.Assert.*;
+
+/**
+ * @author Pavel_Silin@epam.com, EPAM Systems, Inc. <www.epam.com>
+ */
+public class SAMSequenceDictionaryCodecTest {
+
+    private static final Random RANDOM = new Random();
```
yfarjoun: change to lowercase random

src/main/java/htsjdk/samtools/SAMSequenceDictionaryCodec.java
```diff
@@ -0,0 +1,84 @@
...
+package htsjdk.samtools;
+
+import htsjdk.samtools.util.LineReader;
+import java.io.BufferedWriter;
+
+/**
+ * "On the fly" codec SAMSequenceDictionaryCodec.
+ * Encodes each sequence and directly writes it to the Dictionary file.
+ *
+ * @author Pavel_Silin@epam.com, EPAM Systems, Inc. <www.epam.com>
```
yfarjoun: Could you include a short paragraph on how to use this class? something that would describe the various steps needed (open a writer, set the writer, encode the header, encode each sequence record, close the writer). Also include whatever fix you put in-place for the possible NPE of mFileHeader.

src/test/java/htsjdk/samtools/SAMSequenceDictionaryCodecTest.java
SilinPavel: Header is set to `SAMTextHeaderCodec`  in `SAMSequenceDictionaryCodec` constructor now.  It fixes the problem with NPE.

SilinPavel:
@yfarjoun We have done what you asked for. Take a look, please!

src/main/java/htsjdk/samtools/SAMSequenceDictionaryCodec.java
```diff
@@ -31,14 +31,45 @@
...
+ * or
+ *
+ * SAMSequenceDictionary dict = ...;
+ *
+ * //open BufferedReader and close in try-with-resources
+ * try(BufferedWriter writer = new BufferedWriter(new FileWriter("path/to/file"))) {
+ *      SAMSequenceDictionaryCodec codec = new SAMSequenceDictionaryCodec(writer);
+ *
+ *      //we have complete SAMSequenceDictionary, so just encode it.
+ *      codec.encode(dict);
```
SilinPavel: Method encode from `SAMSequenceDictionaryCodec` encodes the header, so we don't need to do it ourself:

```
public void encode(final SAMSequenceDictionary dictionary) { 
        encodeHeaderLine(false);
        dictionary.getSequences().forEach(this::encodeSequenceRecord);
}
```
As you have noticed, in the test I encode @HD line 2 times, I will fix it, or maybe it would be better to delete `encodeHeaderLine(false);` from `encode()` method and do it manually every time, when we use this class, what is your opinion?

yfarjoun: I would not break backwards compatibility. I think that encode(SAMSequenceDictionary) should encode the header line and I would change that. simply do the right thing in the test and make sure that the usage comment up top is correct. also, could you please, in your test verify that the encoded file reads correctly?

SilinPavel: @yfarjoun Done.

yfarjoun:
Sorry, @SilinPavel I misunderstood you. encode on a SAMSequenceDictionary should behave like SAMTextHeaderCodec::encode  which **does** emit the @HD line on its own. Please change the behavior and the example in the comment back to what you (rightly) did before.

SilinPavel:
@yfarjoun I did the rollback to the previous version. Could you check please?

yfarjoun:
great! thanks and sorry for the back and forth. 👍

SilinPavel:
@yfarjoun Thanks!

PR #745, opened 2016-11-19
Discounts reads that are non-filter-passing in DuplicateScoringStrategy 

yfarjoun:
### Description

DuplicateScoringStrategy is a class used by Picard's MarkDuplicates for the purpose of choosing the "best" read/read-pair from a duplicate set. It has multiple strategies, but they all ignore the filter-flag. This means (as mentioned in https://github.com/broadinstitute/picard/issues/690) that in some cases a duplicate-set will end up being represented by a filter-failing read, which will then be ignored by most downstream tools. 

To fix this problem, this PR discounts the score of reads if they are failing the vendor flag. The other scores are capped to guarantee that the discount is large enough. 

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
+++ b/src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
@@ -47,10 +47,9 @@ public class DuplicateScoringStrategy {
 
     /** Calculates a score for the read which is the sum of scores over Q15. */
     private static int getSumOfBaseQualities(final SAMRecord rec) {
-
         int score = 0;
         for (final byte b : rec.getBaseQualities()) {
-            if (b >= 15 ) score += b;
+            if (b >= 15) score += b;
         }
 
         return score;
@@ -65,7 +64,7 @@ public class DuplicateScoringStrategy {
 
     /**
      * Returns the duplicate score computed from the given fragment.
-     * value should be capped by Short.MAX_VALUE since the score from two reads will be
+     * value should be capped by Short.MAX_VALUE/2 since the score from two reads will be
      * added and an overflow will be
      *
      * If true is given to assumeMateCigar, then any score that can use the mate cigar to compute the mate's score will return the score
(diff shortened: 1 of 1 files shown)


yfarjoun:
@tfenne This changes the heart of MarkDuplicates, so I'd really appreciate a careful look, if you can.

yfarjoun:
@eitanbanks could you comment if you think this could introduce batch effects? the "phenotype" will be a little more depth than otherwise. Possibly concentrated in reads that are harder to read.

eitanbanks:
I just don't see this being enough of a phenomenon to induce batch effects.  How often does this change the best pair in a random bam?  Seems like it would happen a small handful of times, so it shouldn't be enough to affect calls.

yfarjoun:
I tend to agree. just making sure that there isn't anything I haven't
thought about.

src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
```diff
@@ -46,10 +46,11 @@
     private static enum Attr { DuplicateScore }
 
     /** Calculates a score for the read which is the sum of scores over Q15. */
-    private static short getSumOfBaseQualities(final SAMRecord rec) {
-        short score = 0;
+    private static int getSumOfBaseQualities(final SAMRecord rec) {
+
```
tfenne: Bonus newline

src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
```diff
@@ -46,10 +46,11 @@
...
+        int score = 0;
         for (final byte b : rec.getBaseQualities()) {
-            if (b >= 15) score += b;
+            if (b >= 15 ) score += b;
```
tfenne: Extra space before `)`

src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
```diff
@@ -64,6 +65,8 @@ public static short computeDuplicateScore(final SAMRecord record, final ScoringS
 
     /**
      * Returns the duplicate score computed from the given fragment.
+     * value should be capped by Short.MAX_VALUE since the score from two reads will be
```
tfenne: Capped to `Short.MAX_VALUE / 2`?  Maybe where they are summed it should just use an int?

src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
```diff
@@ -87,9 +91,23 @@ public static short computeDuplicateScore(final SAMRecord record, final ScoringS
                     }
                     break;
                 case RANDOM:
-                    score += (short) (hasher.hashUnencodedChars(record.getReadName()) >> 16);
+                    // start with a random number between Short.MIN_VALUE/4 and Short.MAX_VALUE/4
+                    score += (short) (hasher.hashUnencodedChars(record.getReadName()) & 0b11_1111_1111_1111);
```
tfenne: This seems unecessarily complicated.  Why not just use `java.util.Random::nextInt(Short.MAX_VALUE / 2)`?

src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
```diff
@@ -87,9 +91,23 @@ public static short computeDuplicateScore(final SAMRecord record, final ScoringS
...
+                    // subtract Short.MIN_VALUE/4 from it to end up with a number between
+                    // 0 and Short.MAX_VALUE/2. This number can be then discounted in case the read is
+                    // not passing filters. We need to stay far from overflow so that when we add the two
+                    // scores from a read-pair we do not overflow since that could cause us to chose the
+                    // Failing read-pair instead of a passing one.
+                    score -= Short.MIN_VALUE/4;
             }
 
+            // make sure that filter-failing records are heavily discounted. Dividing by 2 to be far from
```
tfenne: This comment really applies to lines 109, but there is the handling of overflow and a comment regarding that between this comment and line 109, which is confusing.  Can you move this comment down to immediately before line 109?

src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
```diff
@@ -87,9 +91,23 @@ public static short computeDuplicateScore(final SAMRecord record, final ScoringS
...
+            // overflow when adding score from both fragments of a read-pair
+
+            // a long read with high quality bases may overflow and
+            score = (short) Math.max(Short.MIN_VALUE, score);
+            score += record.getReadFailsVendorQualityCheckFlag() ? (short) (Short.MIN_VALUE / 2) : 0;
```
tfenne: Also, from reading the comment I thought you were just going to do `score = score / 2`, but you're not.  Either the comment needs to explain a little better, or the code needs to just divide the score.

src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
yfarjoun: and then what? the sorting collection holds a short and I don't want to pay the memory cost of changing that to an int...

src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
yfarjoun: I was simply keeping the logic of using mumor hash, as it's a better source of randomness than java.util.Random.

yfarjoun: Oh, I remember why *you* did this :smile:, you wanted the same score for both mate of the same query name....so nextInt will not do. I'll add a comment to this effect.

yfarjoun:
is this better now @tfenne ?

tfenne:
Works for me @yfarjoun.  Merge away.

PR #746, opened 2016-11-20
Public CustomGzipOutputStream

magicDGS:
### Description

There is a `CustomGzipOutputStream` in HTSJDK for setting the compression level at construction, which is used in `IOUtils.openGzipFileForWriting(File file, boolean append)`. But if you want to wrap an `OutputStream` from other sources (like a `Path`) using the `Defaults.COMPRESSION_LEVEL` or other compression, it is impossible to set unless this class is coded again.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


lbergelson:
👍 seems reasonable to me

magicDGS:
Thanks @lbergelson!

Issue #563, opened 2016-04-09
CRAM reader allows simultaneous iterators, but with inconsistent state


cmnbroad:
SamReader enforces that only one iterator exist on a given reader at a time, but this isn't enforced by CRAMFileReader, and can result in inconsistent state if the original source of the CRAM is a stream since the same stream can get re-positioned by both iterators. We should add the same guards (requiring the previous iterator to be closed) on new iterator creation as are implemented in BAMFileReader.

PR #748, opened 2016-11-21
Fix for issue 563: make CRAMFileReader work with only one iterator simultaneously

ekazachkova:
### Fix for issue [#563](https://github.com/samtools/htsjdk/issues/563):

`CRAMFileReader`can keep several iterators opened and this may lead to inconsistent iterator's state if the source of the `CRAMFileReader`is a stream. @cmnbroad suggests to implement the same iterator handling for `CRAMFileReader`as in `BAMFileReader`.  `BAMFileReader` produces iterator as inner class `AbstractBAMIterator`  and iterator is set to reader's field `mCurrentIterator`. `mCurrentIterator` is set to null when `close()` method is called on the iterator and it allows us to check the state of the reader and iterator when creating a new one.

But there are several difficulties with implementing this approach in `CRAMFileReader` and it's iterators:
 - There are two types of iterators: `CRAMFileReader.CRAMIntervalIterator` (private inner) and `CRAMIterator` - public class, that doesn't have access to `CRAMFileReader` fields. `CRAMIterator`'s public access prevents us from implementing the desired approach, but making it private inner class of `CRAMFileReader` will cause lost of backward compatibility. So we suggest another solution: we introduce the new Interface `CloseCheckable` extends `SAMRecordIterator` for the inner `CRAMFileReader` iterator, which allows us to check the iterator's state from the `CRAMFileReader` and restrict simultaneous iterators creation.
 - Besides, there is no way to get several iterators from reader on a stream with a consistent state in the current implementation as `CRAMIterator` reads header by itself. But we need to read it by `CRAMFileReader` and pass to `CRAMIterator`, otherwise, we can open only a single `CRAMIterator` on a stream. We've made changes to the constructors to fix this issue.

We propose the following changes:
 - New constructors in `CRAMIterator` and `CramContainerIterator` that receive CRAM header as an argument. These changes were caused by re-instancing iterator in `CRAMFileReader`.
 - Mark `CRAMIterator` public constructors deprecated since new constructors with header and package access were  implemented. `CRAMIterator` should be instanced only in `CRAMFileReader`, so it is more appropriate not to make its constructors public. 
 - Create inner interface `CRAMFileReader.CloseCheckableSAMRecordIterator` that extends `SAMRecordIterator` by `isClose()` method to check if existing iterator is closed; `CRAMIterator` and `CRAMFileReader.CRAMIntervalIterator` implement this interface to allow `CRAMFileReader` check its state.
 - Throw `IllegalStateException` in `CRAMFileReader` if iterator isn't closed, but a new one is created.
 - Add new fields in `CRAMFileReader` (`cramFileHeader`, `firstContainerOffset`, `mReader`) to keep information about header, first container offset and source.
 - Do some refactoring for code readability: create `initHeaderAndFirstOffset(CRAMIterator)` and `initCRAMIterator()` methods in `CRAMFileReader` class.
 - Fix some tests (`CRAMCRAIIndexerTest`, `CRAMFileBAIIndexTest`, `CRAMFileCRAIIndexTest`, `CRAMFileWriterWithIndexTest`) since it should be impossible to hold several iterators opened at the same time.

Also, we updated license and added new tests in `CRAMFileReaderTest` to check code correctness.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/CRAMFileReader.java
+++ b/src/main/java/htsjdk/samtools/CRAMFileReader.java
@@ -22,12 +22,14 @@ import htsjdk.samtools.cram.ref.CRAMReferenceSource;
 import htsjdk.samtools.cram.ref.ReferenceSource;
 import htsjdk.samtools.cram.structure.Container;
 import htsjdk.samtools.cram.structure.ContainerIO;
+import htsjdk.samtools.cram.structure.CramHeader;
 import htsjdk.samtools.seekablestream.SeekableFileStream;
 import htsjdk.samtools.seekablestream.SeekableStream;
 import htsjdk.samtools.util.CloseableIterator;
 import htsjdk.samtools.util.CloserUtil;
 import htsjdk.samtools.util.Log;
 import htsjdk.samtools.util.RuntimeEOFException;
+import htsjdk.samtools.util.RuntimeIOException;
 
 import java.io.File;
 import java.io.FileInputStream;
@@ -49,13 +51,16 @@ public class CRAMFileReader extends SamReader.ReaderImplementation implements Sa
     private File cramFile;
     private final CRAMReferenceSource referenceSource;
     private InputStream inputStream;
-    private CRAMIterator iterator;
+    private IsClosableSAMRecordIterator iterator;
     private BAMIndex mIndex;
     private File mIndexFile;
     private boolean mEnableIndexCaching;
     private boolean mEnableIndexMemoryMapping;
 
     private ValidationStringency validationStringency;
+    private CramHeader cramFileHeader;
+    private long firstContainerOffset;
+    private SamReader mReader = null;
 
     private final static Log log = Log.getInstance(CRAMFileReader.class);
 
@@ -100,7 +105,7 @@ public class CRAMFileReader extends SamReader.ReaderImplementation implements Sa
         if (cramFile != null) {
             mIndexFile = findIndexForFile(null, cramFile);
         }
-        getIterator();
+        initHeaderAndFirstOffset(initCRAMIterator());
     }
 
     /**
@@ -125,7 +130,9 @@ public class CRAMFileReader extends SamReader.ReaderImplementation implements Sa
         mIndexFile = findIndexForFile(indexFile, cramFile);
         this.referenceSource = referenceSource;
 
-        getIterator();
+        initInputStreamOnFile(cramFile);
+
+        initHeaderAndFirstOffset(initCRAMIterator());
     }
 
     /**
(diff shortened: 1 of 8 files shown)


droazen:
@vadimzalunin Can you review?

vadimzalunin:
@ekazachkova this test fails: 
`new CRAMFileReader(CRAM_WITH_CRAI, createReferenceSource()).getIterator().hasNext();`

I'd suggest to avoid uncertainty in stream positioning somehow. CRAMIterator constructor now assumes the state of the stream based on header value but it is not obvious when making the call.

yfarjoun:
@ekazachkova  please rebase, and add the test mentioned by @vadimzalunin.

yfarjoun:
The BamFileReader doesn't allow for a second open iterator, which agrees with the SamFileReader contract. Could you replicate that behavior instead of allowing for a second open iterator?

bhandsaker:
I'd like to vote for getting this change into htsjdk.
We have hit this problem and it is difficult for us to work around.

yfarjoun:
@vadimzalunin would you be ale to finish reviewing this PR?

@ekazachkova in the future, please refrain from squashing your commits. In github, you can rebase and squash as you merge, and so there's no need to do this in advance (rebasing is sometimes useful, but not squashing) it is much easier to go through the review back-and-forth if you keep the different commits distinct.

cmnbroad:
@ekazachkova I don't think we should let the fact that CRAMIterator is currently public stop us from implementing the simplest solution. Ideally, we would just make it an inner class of CRAMFileReader, but its already quite a large class, mostly due to the fact that much of it's content should be delegated to Container/ContainerParser, etc. So for now, I would propose adding an inner class to CRAMFileReader that acts as a CRAMIterator adapter/wrapper. It would then have access to the CRAMFileReader state; could handle close; and could forward everything else to the real iterator class. I think that would be simpler; it would eliminate the need for some of the new state, and would eliminate the circular dependency now introduced between CRAMFileReader and CRAMIterator. Please let me know if you think thats workable. Also, I think it would be worthwhile to separate any other code cleanup, preferably into a separate PR, but at least into a separate commit

denklewer:
@cmnbroad , sorry for delay and thank you for your suggestion. Now we are ready to continue working on this issue. According to your comments we are going to add a new inner class in `CRAMFileReader` that wrapps `CRAMIterator` using the same approach as implemented in `BAMFileReader`.
  
We suggest to close this PR since proposed changes are not relevant now.

cmnbroad:
Closing this since it has lots of conflicts, and the issue is also being addressed in my CRAM refactoring branch.

PR #707, opened 2016-09-16
Clean some deprecated classes/methods usages in htsjdk

magicDGS:
### Description

I explore some of the deprecated classes/methods when cleaning some of my own code, and I found that there are usages of some of this methods in other parts of htsjdk. This is a PR to clean the usages and to ask for removal of some deprecated classes/methods no longer used. This includes:
- ~~Fixes #330~~ Already fixed in other PR
- Changed all `Feature.getChr()` to `Feature.getContig()` usages in htsjdk.
- Remove `SRAAccession.isSupported()` usages in the tests and change for the recomended one.
- Annotating some deprecated methods with `@Deprecated` and `/** @deprecated */`
- Adding javadoc `{@link }` for pointing to new/correct usage.
- ~~Asking for removal of classes and methods for different packages, including gelitext support, `VariantContextWriterFactory` and others.~~ Edited: open a PR for removal and discussion.

I divided the PR in several commits according to common changes, to make it easier to review. I think that this will help in the maintenance of the htsjdk library. If the removal of some classes is made effective, this will break compatibility with previous versions.

### Checklist
- [x] Code compiles correctly
- [ ] New tests covering changes and new functionality
- [x] All tests passing
- [x] Extended the README / documentation, if necessary
- [ ] Is not backward compatible (breaks binary or source compatibility)

PR #755, opened 2016-11-24
Removed unused deprecated URLHelper.openInputStreamForRange

magicDGS:
### Description

Removed unused and deprecated method in URLHelper. Found in #707.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/tribble/util/HTTPHelper.java
+++ b/src/main/java/htsjdk/tribble/util/HTTPHelper.java
@@ -90,24 +90,6 @@ public class HTTPHelper implements URLHelper {
         return new WrapperInputStream(connection, connection.getInputStream());
     }
 
-
-    /**
-     * Open an input stream for the requested byte range.  Its the client's responsibility to close the stream.
-     *
-     * @param start start of range in bytes
-     * @param end   end of range ni bytes
-     * @return
-     * @throws IOException
-     */
-    @Deprecated
-    public InputStream openInputStreamForRange(long start, long end) throws IOException {
-
-        HttpURLConnection connection = openConnection();
-        String byteRange = "bytes=" + start + "-" + end;
-        connection.setRequestProperty("Range", byteRange);
-        return new WrapperInputStream(connection, connection.getInputStream());
-    }
-
     private HttpURLConnection openConnection() throws IOException {
         HttpURLConnection connection;
         if (proxy == null) {
--- a/src/main/java/htsjdk/tribble/util/URLHelper.java
+++ b/src/main/java/htsjdk/tribble/util/URLHelper.java
@@ -21,7 +21,6 @@ package htsjdk.tribble.util;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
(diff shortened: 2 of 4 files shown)


src/main/java/htsjdk/tribble/util/HTTPHelper.java
```diff
@@ -90,24 +90,6 @@ public InputStream openInputStream() throws IOException {
...
-     * @throws IOException
-     */
-    @Deprecated
-    public InputStream openInputStreamForRange(long start, long end) throws IOException {
-
-        HttpURLConnection connection = openConnection();
-        String byteRange = "bytes=" + start + "-" + end;
-        connection.setRequestProperty("Range", byteRange);
-        return new WrapperInputStream(connection, connection.getInputStream());
-    }
```
droazen: @jrobinso Would you object to our removing this long-deprecated method? It does appear to be used in one or two places in IGV (eg., `IGVUrlHelper`)

jrobinso: It looks save to delete,  IGV doesn't call that method.    It does use openInputStreamForPath,  but that can probably be factored out as well.    Its been many years but I think the idea of that helper class was to allow clients to supply, in essence, a factory object to create custom streams.   Specifically there was a period when IGV used the Apache http library.  

An editorial comment on the title of this issue and reference to "Unused".   Unused by whom?   All public methods of htsjdk are potentially used.

droazen: Yes, I agree that these PRs to remove deprecated methods/classes should involve a check of major downstream projects for usages (GATK, Picard, IGV). We have been doing this before merging them, even if the title of this PR doesn't reflect that.

src/main/java/htsjdk/tribble/util/URLHelper.java
```diff
@@ -21,6 +21,7 @@
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
+import javax.naming.OperationNotSupportedException;
```
droazen: Remove stray import

src/main/java/htsjdk/tribble/util/HTTPHelper.java
magicDGS: Sorry for the word "Unused". I usually check GATK and Picard if I add a PR to remove deprecated methods/classes.

src/main/java/htsjdk/tribble/util/URLHelper.java
magicDGS: Done.

src/main/java/htsjdk/tribble/util/HTTPHelper.java
jrobinso: No problem.   My impression is the htsjdk is widely used in the community at large, at least the SAM / BAM reading part.  Tribble maybe not so much.

vadimzalunin:
i'm working on a streaming PR that should take care of this issue.

Issue #749, opened 2016-11-21
Index associated with GCS-path bam not working properly


kcibul:
When attempting to access a BAM in GCS using GATK PrintReads, although the index is found initially it appears to be lost in SamReaderFactory.

### Your environment
GATK 4.alpha.2-103-g4d29cf7

### Steps to reproduce

`./gatk-launch PrintReads --input gs://public-dream-data/synthetic.challenge.set1.normal.v2.bam  --output k.bam --intervals 1:1000000-1001000 `

### Expected behaviour
k.bam would contain the reads in the specified interval

### Actual behaviour
```
***********************************************************************
A USER ERROR has occurred: Traversal by intervals was requested but some input files are not indexed.
Please index all input files:
samtools index /synthetic.challenge.set1.normal.v2.bam
***********************************************************************
```

This area may be part of the problem in SamReaderFactory as it tries to treat the index like a file

```
                    final File indexFile = indexMaybe == null ? null : indexMaybe.asFile();
                    if (SamStreams.isBAMFile(bufferedStream)) {
                        if (sourceFile == null || !sourceFile.isFile()) {
                            // Handle case in which file is a named pipe, e.g. /dev/stdin or created by mkfifo
                            primitiveSamReader = new BAMFileReader(bufferedStream, indexFile, false, asynchronousIO, validationStringency, this.samRecordFactory);
                        } else {
                            bufferedStream.close();
                            primitiveSamReader = new BAMFileReader(sourceFile, indexFile, false, asynchronousIO, validationStringency, this.samRecordFactory);
                        }
```

droazen:
@tomwhite Can you have a look? Should be a simple patch.

jean-philippe-martin:
Fixed by #762

tfenne:
Fixed by #762 ('Path index', merged Dec 2016), which reworked index handling in SamReaderFactory to support non-File (nio Path) indexes. Closing as resolved.

PR #762, opened 2016-11-29
Path index

jean-philippe-martin:
### Description

Fix a bug that was preventing inputs from being specified as Paths when they are not also Files
(issue #749).

### Checklist

- [X] Code compiles correctly
(I wrote one, it passed, but then I had to take it out to avoid adding a dependency to htsjdk)
- [X] All tests passing
 except for AbstractFeatureReaderTest, testFTPOpenInputStream, sourceLikeBam, and SeekableFTPStreamTest.setUp showing a (presumably normal): Could not connect to ftp.broadinstitute.org, for input source: ftp://ftp.broadinstitute.org/distribution/igv/TEST/cpgIslands with spaces.hg18.bed
- [X] Extended the README / documentation, if necessary (not necessary for a bugfix)
- [X] Is backward compatible

### DIFF
--- a/src/main/java/htsjdk/samtools/BAMFileReader.java
+++ b/src/main/java/htsjdk/samtools/BAMFileReader.java
@@ -116,6 +116,24 @@ class BAMFileReader extends SamReader.ReaderImplementation {
         this.mFileHeader = readHeader(this.mStream, this.mValidationStringency, null);
     }
 
+    BAMFileReader(final InputStream stream,
+        final SeekableStream indexStream,
+        final boolean eagerDecode,
+        final boolean useAsynchronousIO,
+        final ValidationStringency validationStringency,
+        final SAMRecordFactory factory)
+        throws IOException {
+        this.useAsynchronousIO = useAsynchronousIO;
+        mIndexStream = indexStream;
+        mIsSeekable = true;
+        mCompressedInputStream = new BlockCompressedInputStream(stream);
+        mStream = new BinaryCodec(new DataInputStream(mCompressedInputStream));
+        this.eagerDecode = eagerDecode;
+        this.mValidationStringency = validationStringency;
+        this.samRecordFactory = factory;
+        this.mFileHeader = readHeader(this.mStream, this.mValidationStringency, null);
+    }
+
     /**
      * Prepare to read BAM from a file (seekable)
      * @param file source of bytes.
@@ -136,6 +154,31 @@ class BAMFileReader extends SamReader.ReaderImplementation {
         }
         // Provide better error message when there is an error reading.
         mStream.setInputFileName(file.getAbsolutePath());
+    }
+
+
+
+    private BAMFileReader(final File file,
+        final SeekableStream indexStream,
+        final boolean eagerDecode,
+        final boolean useAsynchronousIO,
+        final ValidationStringency validationStringency,
+        final SAMRecordFactory factory,
+        String onlyThereSoTheSignatureIsntAmbiguous)
+        throws IOException {
+        this(new BlockCompressedInputStream(file), indexStream, eagerDecode, useAsynchronousIO, indexStream.getSource(), validationStringency, factory);
+        // Provide better error message when there is an error reading.
+        mStream.setInputFileName(file.getAbsolutePath());
+    }
+
+    static BAMFileReader fromFileAndSeekable(final File file,
+        final SeekableStream indexStream,
+        final boolean eagerDecode,
+        final boolean useAsynchronousIO,
+        final ValidationStringency validationStringency,
+        final SAMRecordFactory factory)
+        throws IOException {
+        return new BAMFileReader(file, indexStream, eagerDecode, useAsynchronousIO, validationStringency, factory, "please");
     }
 
     BAMFileReader(final SeekableStream strm,
--- a/src/main/java/htsjdk/samtools/BAMFileReader.java
+++ b/src/main/java/htsjdk/samtools/BAMFileReader.java
@@ -116,24 +116,6 @@ class BAMFileReader extends SamReader.ReaderImplementation {
         this.mFileHeader = readHeader(this.mStream, this.mValidationStringency, null);
     }
 
-    BAMFileReader(final InputStream stream,
-        final SeekableStream indexStream,
-        final boolean eagerDecode,
-        final boolean useAsynchronousIO,
-        final ValidationStringency validationStringency,
-        final SAMRecordFactory factory)
-        throws IOException {
-        this.useAsynchronousIO = useAsynchronousIO;
-        mIndexStream = indexStream;
-        mIsSeekable = true;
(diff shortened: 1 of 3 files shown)


jean-philippe-martin:
cc: @droazen @lbergelson

droazen:
Tests failed -- re-running to try to determine if the failure is real.

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -311,7 +311,24 @@ public SamReader open(final SamInputResource resource) {
                     if (SamStreams.isBAMFile(bufferedStream)) {
```
droazen: Is it ok to leave the `indexMaybe.asFile()` call intact here (even if the index might be in a bucket)?

src/main/java/htsjdk/samtools/BAMFileReader.java
```diff
@@ -138,6 +138,31 @@
...
+        final ValidationStringency validationStringency,
+        final SAMRecordFactory factory,
+        String onlyThereSoTheSignatureIsntAmbiguous)
+        throws IOException {
+        this(new BlockCompressedInputStream(file), indexStream, eagerDecode, useAsynchronousIO, indexStream.getSource(), validationStringency, factory);
+        // Provide better error message when there is an error reading.
+        mStream.setInputFileName(file.getAbsolutePath());
+    }
+
+    static BAMFileReader fromFileAndSeekable(final File file,
```
droazen: This method `fromFileAndSeekable` is unused

src/main/java/htsjdk/samtools/BAMFileReader.java
```diff
@@ -138,6 +138,31 @@
         mStream.setInputFileName(file.getAbsolutePath());
     }
 
+
+
+    private BAMFileReader(final File file,
```
droazen: This new private constructor appears to be unused.

src/main/java/htsjdk/samtools/BAMFileReader.java
```diff
@@ -187,7 +212,7 @@ private BAMFileReader(final BlockCompressedInputStream compressedInputStream,
                           final SAMRecordFactory factory)
         throws IOException {
         mIndexStream = indexStream;
-        mIsSeekable = true;
+        mIsSeekable = null != indexStream;
```
droazen: This may fix the immediate problem, but I'm concerned that the treatment of `mIsSeekable` is now inconsistent across constructors in this class. Eg., the constructor that takes a `BlockCompressedInputStream` and a `File` for the index sets `mIsSeekable` to true unconditionally, regardless of whether the index file is null. We should probably attempt to impose consistency across constructors here rather than leave the class in an inconsistent state.

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -311,7 +311,24 @@ public SamReader open(final SamInputResource resource) {
...
+                                SeekableStream sourceSeekable = data.asUnbufferedSeekableStream();
+                                if (null != sourceSeekable && null != indexSeekable) {
+                                    // let's make sure the user can seek in this file.
+                                    // need to return to the beginning because it's the same stream we used earlier
+                                    // and read a bit from.
+                                    sourceSeekable.seek(0);
+                                    primitiveSamReader = new BAMFileReader(
+                                        sourceSeekable, indexSeekable, false, asynchronousIO, validationStringency, this.samRecordFactory);
+                                } else {
+                                    primitiveSamReader = new BAMFileReader(
```
droazen: This else clause is the same as the `if (indexFile != null)` clause above -- is it possible to simplify this nested if/else to combine them into a single case?

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -311,7 +311,24 @@ public SamReader open(final SamInputResource resource) {
...
+                                primitiveSamReader = new BAMFileReader(bufferedStream, indexFile, false, asynchronousIO, validationStringency, this.samRecordFactory);
+                            } else {
+                                final SeekableStream indexSeekable = indexMaybe == null ? null : indexMaybe.asUnbufferedSeekableStream();
+                                // do not close bufferedStream, it's the same stream we're getting here.
+                                SeekableStream sourceSeekable = data.asUnbufferedSeekableStream();
+                                if (null != sourceSeekable && null != indexSeekable) {
+                                    // let's make sure the user can seek in this file.
+                                    // need to return to the beginning because it's the same stream we used earlier
+                                    // and read a bit from.
+                                    sourceSeekable.seek(0);
```
droazen: Can you explain why we don't have to seek to the beginning in the other places where `bufferedStream` are used (assuming `bufferedStream` and `sourceSeekable` are the same underlying stream)?

droazen:
Tests passed on the second run -- I assume the initial failure was a temporary ftp connectivity issue.

src/main/java/htsjdk/samtools/BAMFileReader.java
jean-philippe-martin: I think you're right and I can leave it to "true". I ran SamReaderFactoryTest and all 84 tests pass.

src/main/java/htsjdk/samtools/SamReaderFactory.java
jean-philippe-martin: There is no such call on this line so I'm guessing you mean the call just before? Yes, calling toFile on a bucket has no harmful effect, it just returns null. We use other ways to get to the index inside this block, except for the case where there's no index at all (this matches the previous behavior: we pass null and the code deals with the fact there's no index). I've added a comment to this effect.

src/main/java/htsjdk/samtools/SamReaderFactory.java
jean-philippe-martin: I think that's because the other places were written before this patch and so they expect the non-seekable stream to start from wherever isBAMFile leaves it. You can see in the [previous version of the code](https://github.com/samtools/htsjdk/blob/8e7fe7a0b204096c6d0736d473f20fb144908572/src/main/java/htsjdk/samtools/SamReaderFactory.java#L314) how the stream is passed to this variant of the constructor without rewinding. The seekable code path, instead, expects the stream to start from the beginning.

src/main/java/htsjdk/samtools/BAMFileReader.java
jean-philippe-martin: removed.

jean-philippe-martin:
@droazen this should address all your comments. I ran the test suite and all 23992 tests report passing.

jean-philippe-martin:
Travis reports 2 failures, both with "Could not connect to ftp.broadinstitute.org". Sounds like the same transient we've had issues with before.

jean-philippe-martin:
@droazen OK for me to merge?

PR #764, opened 2016-11-29
switching from coverals to codecov

lbergelson:
codecov seems to be more accurate since it incorporates a notion of "partially covered lines"
it also makes reviewing pr's easier since coverage can be shown inline during reviews with a browser plugin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,9 @@ jdk:
   - oraclejdk8
 script: ./gradlew jacocoTestReport testSRA;
 after_success:
+  - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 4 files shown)


lbergelson:
don't merge this, it isn't capturing the sra test coverage yet

yfarjoun:
I love the report that Codecov creates...makes me want to start writing
tests....:-)

lbergelson:
@yfarjoun Have you seen the browser plugin that lets you see which lines of a pull request are covered and which aren't?

lbergelson:
@yfarjoun I'm not certain but I suspect that if the SRA tests started working the coverage would be included now.  Could you review this?

PR #769, opened 2016-12-06
Fixed bug disallowing file-based index and seekable path based source

kcibul:
### Description

Added support for use case where index is a file, and the source stream is a seekable path.  This is important to allow users to pull down an index locally while querying a BAM in GCS.  Currently the performance of leaving the index in GCS in unacceptable.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SamReaderFactory.java
+++ b/src/main/java/htsjdk/samtools/SamReaderFactory.java
@@ -316,23 +316,17 @@ public abstract class SamReaderFactory {
                             final SeekableStream indexSeekable = indexMaybe == null ? null : indexMaybe.asUnbufferedSeekableStream();
                             // do not close bufferedStream, it's the same stream we're getting here.
                             SeekableStream sourceSeekable = data.asUnbufferedSeekableStream();
-                            if (indexFile!=null || null == sourceSeekable || null == indexSeekable) {
-                                if (null == sourceSeekable || null == indexSeekable) {
-                                    // not seekable.
-                                    // it's OK that we consumed a bit of the stream already, this ctor expects it.
-                                    primitiveSamReader = new BAMFileReader(bufferedStream, indexFile, false, asynchronousIO, validationStringency, this.samRecordFactory);
-                                } else {
-                                    sourceSeekable.seek(0);
-                                    primitiveSamReader = new BAMFileReader(
-                                            sourceSeekable, indexSeekable, false, asynchronousIO, validationStringency, this.samRecordFactory);
-                                }
+                            if (null == sourceSeekable || null == indexSeekable) {
+                                // not seekable.
+                                // it's OK that we consumed a bit of the stream already, this ctor expects it.
+                                primitiveSamReader = new BAMFileReader(bufferedStream, indexFile, false, asynchronousIO, validationStringency, this.samRecordFactory);
                             } else {
                                 // seekable.
                                 // need to return to the beginning because it's the same stream we used earlier
                                 // and read a bit from, and that form of the ctor expects the stream to start at 0.
                                 sourceSeekable.seek(0);
                                 primitiveSamReader = new BAMFileReader(
-                                    sourceSeekable, indexSeekable, false, asynchronousIO, validationStringency, this.samRecordFactory);
+                                        sourceSeekable, indexSeekable, false, asynchronousIO, validationStringency, this.samRecordFactory);
                             }
                         } else {
                             bufferedStream.close();
--- a/src/main/java/htsjdk/samtools/SamReaderFactory.java
+++ b/src/main/java/htsjdk/samtools/SamReaderFactory.java
@@ -317,9 +317,15 @@ public abstract class SamReaderFactory {
                             // do not close bufferedStream, it's the same stream we're getting here.
                             SeekableStream sourceSeekable = data.asUnbufferedSeekableStream();
                             if (indexFile!=null || null == sourceSeekable || null == indexSeekable) {
-                                // not seekable.
-                                // it's OK that we consumed a bit of the stream already, this ctor expects it.
-                                primitiveSamReader = new BAMFileReader(bufferedStream, indexFile, false, asynchronousIO, validationStringency, this.samRecordFactory);
+                                if (null == sourceSeekable || null == indexSeekable) {
+                                    // not seekable.
+                                    // it's OK that we consumed a bit of the stream already, this ctor expects it.
+                                    primitiveSamReader = new BAMFileReader(bufferedStream, indexFile, false, asynchronousIO, validationStringency, this.samRecordFactory);
+                                } else {
+                                    sourceSeekable.seek(0);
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -317,9 +317,15 @@ public SamReader open(final SamInputResource resource) {
...
+                                if (null == sourceSeekable || null == indexSeekable) {
+                                    // not seekable.
+                                    // it's OK that we consumed a bit of the stream already, this ctor expects it.
+                                    primitiveSamReader = new BAMFileReader(bufferedStream, indexFile, false, asynchronousIO, validationStringency, this.samRecordFactory);
+                                } else {
+                                    sourceSeekable.seek(0);
+                                    primitiveSamReader = new BAMFileReader(
+                                            sourceSeekable, indexSeekable, false, asynchronousIO, validationStringency, this.samRecordFactory);
+                                }
                             } else {
```
droazen: I think that this entire nested if statement from line 319 up to line 336 can be simplified from this:

```
                            if (indexFile!=null || null == sourceSeekable || null == indexSeekable) {
                                if (null == sourceSeekable || null == indexSeekable) {
                                    // not seekable.
                                    // it's OK that we consumed a bit of the stream already, this ctor expects it.
                                    primitiveSamReader = new BAMFileReader(bufferedStream, indexFile, false, asynchronousIO, validationStringency, this.samRecordFactory);
                                } else {
                                    sourceSeekable.seek(0);
                                    primitiveSamReader = new BAMFileReader(
                                            sourceSeekable, indexSeekable, false, asynchronousIO, validationStringency, this.samRecordFactory);
                                }
                            } else {
                                // seekable.
                                // need to return to the beginning because it's the same stream we used earlier
                                // and read a bit from, and that form of the ctor expects the stream to start at 0.
                                sourceSeekable.seek(0);
                                primitiveSamReader = new BAMFileReader(
                                    sourceSeekable, indexSeekable, false, asynchronousIO, validationStringency, this.samRecordFactory);
                            }
```

to this:

```
                                if (null == sourceSeekable || null == indexSeekable) {
                                    // not seekable.
                                    // it's OK that we consumed a bit of the stream already, this ctor expects it.
                                    primitiveSamReader = new BAMFileReader(bufferedStream, indexFile, false, asynchronousIO, validationStringency, this.samRecordFactory);
                                } else {
                                    // seekable.
                                    // need to return to the beginning because it's the same stream we used earlier
                                    // and read a bit from, and that form of the ctor expects the stream to start at 0.
                                    sourceSeekable.seek(0);
                                    primitiveSamReader = new BAMFileReader(
                                            sourceSeekable, indexSeekable, false, asynchronousIO, validationStringency, this.samRecordFactory);
                                }
```

Since the two `else` clauses in the original code are identical.

In other words, what you seem to have done with this patch is just to say "if you have an index that's a file, you can now use the constructor that takes two seekable streams instead of the (non-seekable-stream, file) constructor."

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -284,7 +284,37 @@ public void queryInputResourcePermutation(final SamInputResource resource) throw
         }
         reader.close();
     }
-    
+
+    public class NeverFilePathInpurResource extends PathInputResource {
```
droazen: This class should definitely be `private static`, and come with a comment explaining what it's for.

Also, typo: `Inpur` -> `Input`

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -284,7 +284,37 @@ public void queryInputResourcePermutation(final SamInputResource resource) throw
...
+        }
+    }
+
+    @Test
+    public void streamingPathBamWithFileIndex() throws IOException {
+        InputResource bam = new NeverFilePathInpurResource(localBam.toPath());
+        InputResource index = new FileInputResource(localBamIndex);
+
+        // ensure that the index is being used, not checked in queryInputResourcePermutation
+        final SamReader reader = SamReaderFactory.makeDefault().open(new SamInputResource(bam, index));
```
droazen: Close your reader when done using a try-with-resources statement here.

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -284,7 +284,37 @@ public void queryInputResourcePermutation(final SamInputResource resource) throw
...
+        }
+
+        @Override
+        public File asFile() {
+            return null;
+        }
+    }
+
+    @Test
+    public void streamingPathBamWithFileIndex() throws IOException {
```
droazen: Rename test case to something like `checkHasIndexForStreamingPathBamWithFileIndex()`, to make it clearer how it differs from the test case below it.

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -284,7 +284,37 @@ public void queryInputResourcePermutation(final SamInputResource resource) throw
...
+            super(pathResource);
+        }
+
+        @Override
+        public File asFile() {
+            return null;
+        }
+    }
+
+    @Test
```
droazen: Can you confirm that these tests both failed before this patch?

droazen:
@vdauwera This is the PR that needs to be merged before the next htsjdk release, if you want to monitor its progress.

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
kcibul: The query test actually doesn't fail if it can't use the index, it just falls back to not using the index and emits a warning.  That seemed strange (to log a warning in a test rathr than fail) but I didn't want to change it's behavior as it predates me.  So I added the second test instead to assert the index is available which combined with the first test ensure it's both available AND does the right thing in query

Techically, the Then I patched SamFileReader

src/main/java/htsjdk/samtools/SamReaderFactory.java
kcibul: 👍

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -284,7 +284,43 @@ public void queryInputResourcePermutation(final SamInputResource resource) throw
...
         reader.close();
     }
-    
+
+
+    /**
+     * A path that pretends it's not based upon a file.  This helps in cases where we want to test branches
+     * that apply to non-file based paths without actually having to use non-file based resources (like cloud urls)
+     */
+    public static class NeverFilePathInputResource extends PathInputResource {
```
droazen: Still need to make this class `private`

droazen:
Ok, looks good now -- I'll hit merge once tests pass, then we're going to do an htsjdk release.

PR #775, opened 2016-12-15
Add optional wrapper for the underlying SeekableByteChannel

jean-philippe-martin:
### Description

This allows users to provide their own buffering or prefetching, without them having to change htsjdk.

### Checklist

- [x] Code compiles correctly
- [x] New tests covering changes and new functionality (added a wrapper just for tests that adds required headers)
- [x] All tests passing
- [x] Extended the README / documentation, if necessary (not necessary aside from comments)
- [x] Is backward compatible (does not breaks binary or source compatibility)

cc: @droazen @lbergelson


src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -74,11 +76,13 @@
 public abstract class SamReaderFactory {
 
     private static ValidationStringency defaultValidationStringency = ValidationStringency.DEFAULT_STRINGENCY;
-    
+
+    protected Function<SeekableByteChannel, SeekableByteChannel> pathWrapper;
```
lbergelson: If we're going to include this in the base class with accessors, lets just make it private.

lbergelson: Can you give this a default of `Function.identity()`?

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
```diff
@@ -34,6 +35,16 @@ public SeekablePathStream(final Path path) throws IOException {
         ALL_INSTANCES.add(this);
```
lbergelson: Should this  constructor one just delegate to the new one with Function.identity()?

src/main/java/htsjdk/samtools/SamInputResource.java
```diff
@@ -89,7 +91,9 @@ public String toString() {
     public static SamInputResource of(final File file) { return new SamInputResource(new FileInputResource(file)); }
 
     /** Creates a {@link SamInputResource} reading from the provided resource, with no index. */
-    public static SamInputResource of(final Path path) { return new SamInputResource(new PathInputResource(path)); }
```
lbergelson: Could you add javadoc explaining what the wrapper is for on all these constructors that take one?

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -74,11 +76,13 @@
...
+
     abstract public SamReader open(final File file);
 
     public SamReader open(final Path path) {
-        final SamInputResource r = SamInputResource.of(path);
+        final SamInputResource r = SamInputResource.of(path, getPathWrapper());
         final Path indexMaybe = SamFiles.findIndex(path);
         if (indexMaybe != null) r.index(indexMaybe);
```
lbergelson: I just noticed, this isn't going to wrap the index.  That seems like a problem.  

We might actually want the ability to treat the index and the main path separately?  Do you think that's necessary?  I.e. download the index in it's entirety and store it in on disk vs cache chunks bam in memory as we stream over them.

src/main/java/htsjdk/samtools/SamReaderFactory.java
jean-philippe-martin: OK

src/main/java/htsjdk/samtools/seekablestream/SeekablePathStream.java
jean-philippe-martin: Yes, done (though with null since we also want to accept that)

src/main/java/htsjdk/samtools/SamInputResource.java
```diff
@@ -281,8 +286,27 @@ public SeekableStream make() {
...
+        this.wrapper = wrapper;
+    }
+
+    /** Returns a modified PathInputResource with the specific wrapper. After calling this,
+     * do NOT use the original object anymore.
+     *
+     * @param wrapper
+     * @return wrapped PathInputResource
+     */
+    PathInputResource wrap(Function<SeekableByteChannel, SeekableByteChannel> wrapper) {
```
jean-philippe-martin: This had something to do with how GATK uses it. I don't remember now, it has been too long and I lost the context. If this function's a problem then I should also push the GATK PR and we can work them in concert.

src/main/java/htsjdk/samtools/SamInputResource.java
jean-philippe-martin: Kept overload, added comment here. Which other ctors would you like to add a comment to? I think this one's the only public one so it would be the right place for the comment.

src/main/java/htsjdk/samtools/SamReaderFactory.java
jean-philippe-martin: Not wrapping the index is exactly the right thing to do. The index is small, we already read the whole thing into memory at startup anyways.
We could optionally add a way for the user to specify their own wrapper for the index, if we see a need for it later.

lbergelson: You're certain the index is always read into memory and cached?  I could have sworn @kcibul was having issues where there was slowdown due to repeatedly accessing the index over the network.  He ended up copying the index files locally.  It looks like there's an option `CACHE_FILE_BASED_INDEXES` maybe he wasn't using that while you are?

src/main/java/htsjdk/samtools/SamInputResource.java
lbergelson: If we need it for gatk then we can keep it.  It's not a public method, so I was assuming no one is using it if they're not in htsjdk.  Can you check if your gatk branch needs it for something?

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -102,6 +106,21 @@ public SamReader open(final Path path) {
...
+     *
+     * @param wrapper how to modify the SeekableByteChannel (Function.identity to unset)
+     * @return this
+     */
+    public SamReaderFactory setPathWrapper(Function<SeekableByteChannel, SeekableByteChannel> wrapper) {
+        this.pathWrapper = wrapper;
+        return this;
+    }
+
+    public Function<SeekableByteChannel, SeekableByteChannel> getPathWrapper() {
```
lbergelson: This should probably have javadoc while we're here.

src/main/java/htsjdk/samtools/SamInputResource.java
lbergelson: I thought it might be worth it to add an explanation to the PathInputResource constructor in case someone wanders into the code there and wants to know why path inputs have this special thing that the others don't.

src/main/java/htsjdk/samtools/SamReaderFactory.java
jean-philippe-martin: It's up to user code; reading the whole thing is the right thing to do so I expect performance-oriented code will do that already (by "we" earlier I meant my own code).

jean-philippe-martin:
Unfortunately, doubling the buffer results in an invalid SAM file (breaks the header rules).

jean-philippe-martin:
OK, I got a test with a wrapper that adds the necessary headers (reading would fail if the wrapper weren't invoked).

src/main/java/htsjdk/samtools/SamInputResource.java
jean-philippe-martin: Done

src/main/java/htsjdk/samtools/SamInputResource.java
jean-philippe-martin: Let me get back to you on that.

jean-philippe-martin: I looked into it and it looks like I can do without wrap(). So I removed it.

lbergelson:
@jean-philippe-martin 👍

jean-philippe-martin:
Thank you @lbergelson!

PR #780, opened 2017-01-05
refactoring Lazy to take a Supplier()

lbergelson:
### Description
Lazy previously accepted a custom interface called LazyInitializer which is redundant with Supplier since java 8

made LazyInitializer extend Supplier
changed the constructor to take generic Supplier instead of only LazyInitializers

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/VariantContextUtils.java
+++ b/src/main/java/htsjdk/variant/variantcontext/VariantContextUtils.java
@@ -52,15 +52,12 @@ public class VariantContextUtils {
 
     /** Use a {@link Lazy} {@link JexlEngine} instance to avoid class-loading issues. (Applications that access this class are otherwise
      * forced to build a {@link JexlEngine} instance, which depends on some apache logging libraries that mightn't be packaged.) */
-    final public static Lazy<JexlEngine> engine = new Lazy<JexlEngine>(new Lazy.LazyInitializer<JexlEngine>() {
-        @Override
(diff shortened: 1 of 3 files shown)


lbergelson:
@yfarjoun I noticed that we had this pointless class.  I deprecated it. It should be backwards compatible with existing code.

src/main/java/htsjdk/variant/variantcontext/VariantContextUtils.java
```diff
@@ -52,15 +52,12 @@
...
-    final public static Lazy<JexlEngine> engine = new Lazy<JexlEngine>(new Lazy.LazyInitializer<JexlEngine>() {
-        @Override
-        public JexlEngine make() {
-            final JexlEngine jexl = new JexlEngine();
-            jexl.setSilent(false); // will throw errors now for selects that don't evaluate properly
-            jexl.setLenient(false);
-            jexl.setDebug(false);
-            return jexl;
-        }
+    final public static Lazy<JexlEngine> engine = new Lazy<>(() -> {
```
yfarjoun: I think you can drop the `new Lazy<>(` and only have 
```
final public static Lazy<JexlEngine> engine = () -> {
...
}
```

yfarjoun: nm. got confused.

yfarjoun:
👍

PR #782, opened 2017-01-10
Shl edit samrecord docs

sooheelee:
Clarify SAMRecord documentation so that users are not confused by thinking the opposite of what is meant.

- Three variable descriptions changed: getAlignmentStart, setAlignmentStart and getAlignmentEnd. Before, each read, "position of the clipped sequence," and now each reads, "position of the sequence remaining after clipping," for all three variable descriptions. 
- In addition, took liberty to clarify "or 0 if there is no position" further. Now reads, "If no position, as would occur e.g. for the unmapped mate of a singly mapping pair, returns 0." Talked briefly to LB and agreed _not_ to link to GATK documentation within htsjdk given it is a standalone project.

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMRecord.java
+++ b/src/main/java/htsjdk/samtools/SAMRecord.java
@@ -572,14 +572,14 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
     }
 
     /**
-     * @return 1-based inclusive leftmost position of the clipped sequence, or 0 if there is no position.
+     * @return 1-based inclusive leftmost position of the sequence remaining after clipping. If no position, as would occur e.g. for the unmapped mate of a singly mapping pair, returns 0.
      */
     public int getAlignmentStart() {
         return mAlignmentStart;
     }
 
     /**
-     * @param value 1-based inclusive leftmost position of the clipped sequence, or 0 if there is no position.
+     * @param value 1-based inclusive leftmost position of the sequence remaining after clipping. If no position, e.g. for unmapped mate of a singly mapping pair, returns 0.
      */
     public void setAlignmentStart(final int value) {
         mAlignmentStart = value;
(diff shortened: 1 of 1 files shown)


sooheelee:
The commit with comments contain the actual documentation updates. The later commit (with no comments, sorry about that) just formats line spacing. Newbie here so please forgive.

nh13:
I am confused by the second line in each (singly mapped pair).  I am not sure about the need for this change, but if you are adamant, then how about: `1-based inclusive leftmost position of the sequence remaining after clipping, or 0 if there is no position (ex. the read is unmapped)`

sooheelee:
@nh13, I believe the top of the SAMRecord doc talks about the types of unmapped reads that the GATK documentation refer to as part of a singly mapping pair [here](https://software.broadinstitute.org/gatk/guide/article?id=6976). I don't think the SAMRecord doc means to imply any unmapped read. My conclusion is that it must refer to those with coordinates within the SAM record but with `*` in the CIGAR field as happens for unmapped reads of singly mapping pairs.

Please correct if I'm mistaken.

If you feel strongly about merging the two sentences, and keeping it simple, then I can do:

> 1-based inclusive leftmost position of the sequence remaining after clipping, or 0 if there is no position, e.g. for an unmapped read in a singly mapping pair.

julianhess:
It will return 0 for a totally unmapped read — it simply regurgitates what's in the BAM.  According to SAM spec, a totally unmapped read will have `RNAME = *` and `POS = 0`.  I agree with @nh13's more general wording.

lbergelson:
@vdauwera The newlines don't make any difference to the doc output.

nh13:
@sooheelee let me know when you make the change and I can look again.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -572,14 +572,16 @@ protected static String resolveNameFromIndex(final int referenceIndex, final SAM
     }
 
     /**
-     * @return 1-based inclusive leftmost position of the clipped sequence, or 0 if there is no position.
+     * @return 1-based inclusive leftmost position of the sequence remaining after clipping. 
+     * If no position, as would occur e.g. for the unmapped mate of a singly mapping pair, returns 0.
```
droazen: singly mapping -> singly-mapped (here and below

sooheelee:
@nh13 Now reads "1-based inclusive leftmost position of the sequence remaining after clipping, or 0 if there is no position, e.g. for unmapped read."

sooheelee:
Thanks for the feedback @julianhess @nh13

vdauwera:
Sounds like we have 👍 from everyone then?

lbergelson:
Congrats @sooheelee on your first htsjdk commit!

sooheelee:
Thank you @lbergelson!

PR #785, opened 2017-01-11
wrap both path and index

jean-philippe-martin:
### Description

Extend optional wrapping to the index.

### Checklist

- [X] Code compiles correctly
- [X] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SamReaderFactory.java
+++ b/src/main/java/htsjdk/samtools/SamReaderFactory.java
@@ -82,9 +82,15 @@ public abstract class SamReaderFactory {
     abstract public SamReader open(final File file);
 
     public SamReader open(final Path path) {
-        final SamInputResource r = SamInputResource.of(path, getPathWrapper());
+        return open(path, getPathWrapper(), Function.identity());
+    }
+
+    public SamReader open(final Path path,
+            Function<SeekableByteChannel, SeekableByteChannel> pathWrapper,
+            Function<SeekableByteChannel, SeekableByteChannel> indexWrapper) {
+        final SamInputResource r = SamInputResource.of(path, pathWrapper);
         final Path indexMaybe = SamFiles.findIndex(path);
-        if (indexMaybe != null) r.index(indexMaybe);
+        if (indexMaybe != null) r.index(indexMaybe, indexWrapper);
         return open(r);
     }
 
--- a/src/main/java/htsjdk/samtools/SamReaderFactory.java
+++ b/src/main/java/htsjdk/samtools/SamReaderFactory.java
@@ -81,14 +81,26 @@ public abstract class SamReaderFactory {
 
     abstract public SamReader open(final File file);
 
+    /**
+     * Open the specified path (without using any wrappers).
+     *
+     * @param path the SAM or BAM file to open.
+     */
     public SamReader open(final Path path) {
-        return open(path, getPathWrapper(), Function.identity());
+        return open(path, null, null);
     }
 
+    /**
+     * Open the specified path, using the specified wrappers for prefetching/caching.
+     *
+     * @param path the SAM or BAM file to open
+     * @param dataWrapper the wrapper for the data (or null for none)
+     * @param indexWrapper the wrapper for the index (or null for none)
+     */
     public SamReader open(final Path path,
-            Function<SeekableByteChannel, SeekableByteChannel> pathWrapper,
+            Function<SeekableByteChannel, SeekableByteChannel> dataWrapper,
             Function<SeekableByteChannel, SeekableByteChannel> indexWrapper) {
-        final SamInputResource r = SamInputResource.of(path, pathWrapper);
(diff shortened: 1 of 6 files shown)


src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -82,9 +82,15 @@
     abstract public SamReader open(final File file);
 
     public SamReader open(final Path path) {
-        final SamInputResource r = SamInputResource.of(path, getPathWrapper());
+        return open(path, getPathWrapper(), Function.identity());
```
droazen: It seems like we should use the wrapper returned by `getPathWrapper()` as the default for both the bam and its index, instead of just the bam?

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -82,9 +82,15 @@
...
+    }
+
+    public SamReader open(final Path path,
+            Function<SeekableByteChannel, SeekableByteChannel> pathWrapper,
+            Function<SeekableByteChannel, SeekableByteChannel> indexWrapper) {
```
droazen: Is there a use case for having different wrappers for the bam vs. the index? If not, should we just have a single wrapper argument to this constructor that is applied to both (assuming it's not problematic to share wrapper)?

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -82,9 +82,15 @@
     abstract public SamReader open(final File file);
 
     public SamReader open(final Path path) {
-        final SamInputResource r = SamInputResource.of(path, getPathWrapper());
+        return open(path, getPathWrapper(), Function.identity());
+    }
+
+    public SamReader open(final Path path,
```
droazen: Add javadoc for both overloads of `open()` touched in this PR.

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -82,9 +82,15 @@
...
+            Function<SeekableByteChannel, SeekableByteChannel> pathWrapper,
+            Function<SeekableByteChannel, SeekableByteChannel> indexWrapper) {
+        final SamInputResource r = SamInputResource.of(path, pathWrapper);
         final Path indexMaybe = SamFiles.findIndex(path);
-        if (indexMaybe != null) r.index(indexMaybe);
+        if (indexMaybe != null) r.index(indexMaybe, indexWrapper);
         return open(r);
     }
```
droazen: Is it possible to add a simple unit test for this method that verifies that the provided wrapper is applied (perhaps using a dummy wrapper of some kind)?

src/main/java/htsjdk/samtools/SamReaderFactory.java
jean-philippe-martin: Actually since we've now moved to passing the wrappers explicitly to open, I suggest we remove setPathWrapper. The open call without wrappers will not apply any wrapper, behaving like the other methods that take optional wrapper arguments.

src/main/java/htsjdk/samtools/SamReaderFactory.java
jean-philippe-martin: done

src/main/java/htsjdk/samtools/SamReaderFactory.java
jean-philippe-martin: It's not problematic to share wrappers (they have no state). Those files are accessed differently, so it would be a great coincidence if the best caching&prefetching for one were also the best for the other.

src/main/java/htsjdk/samtools/SamReaderFactory.java
jean-philippe-martin: One way would be to copy the xor wrapper from GATK and check that we can read an XOR'd index.

jean-philippe-martin: Actually that wouldn't work because the factory only uses the wrappers for things that are not files. Instead I'm doing something simpler and directly calling "asUnbufferedSeekableStream" and checking that the wrapper was accessed.

jean-philippe-martin:
@droazen the code is ready! (Mysteriously, github isn't letting me assign the bug back to you. Perhaps something needs to be done to grant me bug-assigner permissions?)

src/test/java/htsjdk/samtools/PathInputResourceTest.java
```diff
@@ -0,0 +1,46 @@
+package htsjdk.samtools;
+
+import java.nio.channels.SeekableByteChannel;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.function.Function;
+import junit.framework.Assert;
```
droazen: Should be `org.testng.Assert`

droazen:
@jean-philippe-martin Looks good, just one import in the tests needs fixing, then I'll hit merge!

PR #786, opened 2017-01-19
Make BAMFileReader and some related classes public

tomwhite:
... and expose methods for iterating over a part of a BAM file (needed for Hadoop-BAM,
which processes BAM files in parallel, see https://github.com/HadoopGenomics/Hadoop-BAM/issues/91).

Also, add BAMFileSpan#removeContentsAfter method to mirror
removeContentsBefore.

### Description

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/BAMFileReader.java
+++ b/src/main/java/htsjdk/samtools/BAMFileReader.java
@@ -40,7 +40,7 @@ import java.util.NoSuchElementException;
 /**
  * Class for reading and querying BAM files.
  */
-class BAMFileReader extends SamReader.ReaderImplementation {
+public class BAMFileReader extends SamReader.ReaderImplementation {
     // True if reading from a File rather than an InputStream
     private boolean mIsSeekable = false;
 
@@ -869,25 +869,56 @@ class BAMFileReader extends SamReader.ReaderImplementation {
         }
     }
 
-    private CloseableIterator<SAMRecord> createIndexIterator(final QueryInterval[] intervals,
-                                                             final boolean contained) {
-
-        assertIntervalsOptimized(intervals);
-
-        // Hit the index to determine the chunk boundaries for the required data.
+    /**
+     * Use the index to determine the chunk boundaries for the required intervals.
+     * @param intervals the intervals to restrict reads to
+     * @param fileIndex the BAM index to use
+     * @return file pointer pairs corresponding to chunk boundaries
+     */
+    public static BAMFileSpan getFileSpan(QueryInterval[] intervals, BAMIndex fileIndex) {
         final BAMFileSpan[] inputSpans = new BAMFileSpan[intervals.length];
-        final BAMIndex fileIndex = getIndex();
         for (int i = 0; i < intervals.length; ++i) {
             final QueryInterval interval = intervals[i];
             final BAMFileSpan span = fileIndex.getSpanOverlapping(interval.referenceIndex, interval.start, interval.end);
             inputSpans[i] = span;
         }
-        final long[] filePointers;
+        final BAMFileSpan span;
         if (inputSpans.length > 0) {
-            filePointers = BAMFileSpan.merge(inputSpans).toCoordinateArray();
+            span = BAMFileSpan.merge(inputSpans);
         } else {
-            filePointers = null;
+            span = null;
         }
+        return span;
+    }
+
+    private CloseableIterator<SAMRecord> createIndexIterator(final QueryInterval[] intervals,
+                                                             final boolean contained) {
+
+        assertIntervalsOptimized(intervals);
+
+        BAMFileSpan span = getFileSpan(intervals, getIndex());
+
+        // Create an iterator over the above chunk boundaries.
+        final BAMFileIndexIterator iterator = new BAMFileIndexIterator(span == null ? null : span.toCoordinateArray());
+
+        // Add some preprocessing filters for edge-case reads that don't fit into this
+        // query type.
+        return new BAMQueryFilteringIterator(iterator, new BAMQueryMultipleIntervalsIteratorFilter(intervals, contained));
+    }
+
+    /**
+     * Prepare to iterate through SAMRecords that match the given intervals.
+     * @param intervals the intervals to restrict reads to
+     * @param contained if <code>true</code>, return records that are strictly
+     *                  contained in the intervals, otherwise return records that overlap
+     * @param filePointers file pointer pairs corresponding to chunk boundaries for the
+     *                     intervals
+     */
+    public CloseableIterator<SAMRecord> createIndexIterator(final QueryInterval[] intervals,
+                                                            final boolean contained,
+                                                            final long[] filePointers) {
+
+        assertIntervalsOptimized(intervals);
 
         // Create an iterator over the above chunk boundaries.
         final BAMFileIndexIterator iterator = new BAMFileIndexIterator(filePointers);
--- a/src/main/java/htsjdk/samtools/BAMFileSpan.java
+++ b/src/main/java/htsjdk/samtools/BAMFileSpan.java
@@ -115,15 +115,55 @@ public class BAMFileSpan implements SAMFileSpan, Serializable {
         validateSorted();
 
         final BAMFileSpan trimmedChunkList = new BAMFileSpan();
+        final long chunkStart = bamFileSpan.chunks.get(0).getChunkStart();
         for(final Chunk chunkToTrim: chunks) {
-            if(chunkToTrim.getChunkEnd() > chunkToTrim.getChunkStart()) {
-                if(chunkToTrim.getChunkStart() >= bamFileSpan.chunks.get(0).getChunkStart()) {
+            if(chunkToTrim.getChunkEnd() > chunkStart) {
+                if(chunkToTrim.getChunkStart() >= chunkStart) {
                     // This chunk from the list is completely beyond the start of the filtering chunk.
                     trimmedChunkList.add(chunkToTrim.clone());
                 }
                 else {
                     // This chunk from the list partially overlaps the filtering chunk and must be trimmed.
-                    trimmedChunkList.add(new Chunk(bamFileSpan.chunks.get(0).getChunkStart(),chunkToTrim.getChunkEnd()));
+                    trimmedChunkList.add(new Chunk(chunkStart,chunkToTrim.getChunkEnd()));
+                }
+            }
+        }
+        return trimmedChunkList;
+    }
+
+    /**
+     * Creates a new file span by removing all chunks after the given file span ends.
+     * If a chunk in the chunk list starts before and ends after the given
+     * chunk, the second portion of the chunk will be deleted.
+     * @param fileSpan The filespan after which to eliminate.
+     * @return A new BAMFileSpan which contains the portion of the chunk list before the
+     * given chunk.
+     */
+    public SAMFileSpan removeContentsAfter(final SAMFileSpan fileSpan) {
+        if(fileSpan == null)
+            return clone();
+
+        if(!(fileSpan instanceof BAMFileSpan))
+            throw new SAMException("Unable to compare ");
+
+        final BAMFileSpan bamFileSpan = (BAMFileSpan)fileSpan;
+
+        if(bamFileSpan.isEmpty())
+            return clone();
+
+        validateSorted();
+
+        final BAMFileSpan trimmedChunkList = new BAMFileSpan();
+        final long chunkEnd = bamFileSpan.chunks.get(bamFileSpan.chunks.size() - 1).getChunkEnd();
+        for(final Chunk chunkToTrim: chunks) {
+            if(chunkToTrim.getChunkStart() < chunkEnd) {
+                if(chunkToTrim.getChunkEnd() <= chunkEnd) {
+                    // This chunk from the list is completely before the end of the filtering chunk.
+                    trimmedChunkList.add(chunkToTrim.clone());
+                }
+                else {
+                    // This chunk from the list partially overlaps the filtering chunk and must be trimmed.
+                    trimmedChunkList.add(new Chunk(chunkToTrim.getChunkStart(),chunkEnd));
                 }
             }
         }
--- a/src/main/java/htsjdk/samtools/SamReader.java
+++ b/src/main/java/htsjdk/samtools/SamReader.java
@@ -381,7 +381,11 @@ public interface SamReader extends Iterable<SAMRecord>, Closeable {
             this.resource = resource;
         }
 
-        PrimitiveSamReader underlyingReader() {
+        /**
+         * Access the underlying {@link PrimitiveSamReader} used by this adapter.
+         * @return the {@link PrimitiveSamReader} used by this adapter.
+         */
+        public PrimitiveSamReader underlyingReader() {
             return p;
         }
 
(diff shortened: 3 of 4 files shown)


tomwhite:
@droazen, can you take a look?

src/main/java/htsjdk/samtools/BAMFileReader.java
```diff
@@ -329,6 +329,29 @@ public ValidationStringency getValidationStringency() {
...
+     * file pointer. The restrictions described for {@link #getIterator()} apply here.
+     */
+    public CloseableIterator<SAMRecord> getIterator(long firstRecordPointer) {
+        if (mStream == null) {
+            throw new IllegalStateException("File reader is closed");
+        }
+        if (mCurrentIterator != null) {
+            throw new IllegalStateException("Iteration in progress");
+        }
+        if (mIsSeekable) {
```
lbergelson: this should should throw if it's not seekable, not silently give you back the wrong iterator

src/main/java/htsjdk/samtools/BAMFileReader.java
```diff
@@ -329,6 +329,29 @@ public ValidationStringency getValidationStringency() {
         return mCurrentIterator;
     }
 
+    /**
+     * Prepare to iterate through the SAMRecords in file order, starting at the given
+     * file pointer. The restrictions described for {@link #getIterator()} apply here.
```
droazen: It seems like this is unused in the GATK/Hadoop-BAM client code in favor of `getIterator(final SAMFileSpan chunks)` -- can we remove?

src/main/java/htsjdk/samtools/BAMFileReader.java
```diff
@@ -732,25 +755,40 @@ private void assertIntervalsOptimized(final QueryInterval[] intervals) {
...
+    }
+
+    private CloseableIterator<SAMRecord> createIndexIterator(final QueryInterval[] intervals,
+                                                             final boolean contained) {
+
+        BAMFileSpan span = getFileSpan(intervals, getIndex());
+        return createIndexIterator(intervals, contained, span == null ? null : span.toCoordinateArray());
+    }
+
+    public CloseableIterator<SAMRecord> createIndexIterator(final QueryInterval[] intervals,
```
droazen: Add javadoc for this method

src/main/java/htsjdk/samtools/BAMFileReader.java
```diff
@@ -763,7 +801,7 @@ private void assertIntervalsOptimized(final QueryInterval[] intervals) {
     /**
      * Iterate over the SAMRecords defined by the sections of the file described in the ctor argument.
      */
-    private class BAMFileIndexIterator extends BAMFileIterator {
+    public class BAMFileIndexIterator extends BAMFileIterator {
```
droazen: Why does this need to be made public? It seems like there are no direct usages in any downstream PRs

src/main/java/htsjdk/samtools/BAMFileSpan.java
```diff
@@ -115,15 +115,55 @@ public SAMFileSpan removeContentsBefore(final SAMFileSpan fileSpan) {
         validateSorted();
 
         final BAMFileSpan trimmedChunkList = new BAMFileSpan();
+        long chunkStart = bamFileSpan.chunks.get(0).getChunkStart();
```
lbergelson: make this final I think

src/main/java/htsjdk/samtools/BAMFileReader.java
nh13: @droazen can you use the review feature so I don't get a notification per comment?

src/main/java/htsjdk/samtools/BAMFileSpan.java
```diff
@@ -115,15 +115,55 @@ public SAMFileSpan removeContentsBefore(final SAMFileSpan fileSpan) {
...
         for(final Chunk chunkToTrim: chunks) {
-            if(chunkToTrim.getChunkEnd() > chunkToTrim.getChunkStart()) {
-                if(chunkToTrim.getChunkStart() >= bamFileSpan.chunks.get(0).getChunkStart()) {
+            if(chunkToTrim.getChunkEnd() > chunkStart) {
```
droazen: This does not seem equivalent to the previous implementation -- it used to be comparing against the start of the current chunk (`chunkToTrim`), but now it's comparing against the start of the first chunk (`chunkStart`)

src/test/java/htsjdk/samtools/BAMFileSpanTest.java
```diff
@@ -0,0 +1,66 @@
...
+        ((BAMFileSpan) originalSpan.removeContentsBefore(cutoff)).getChunks(),
+        expectedSpan.getChunks());
+  }
+
+  @DataProvider(name = "testRemoveContentsBeforeProvider")
+  private Object[][] testRemoveContentsBeforeProvider() {
+    return new Object[][] {
+        { span(chunk(6,10), chunk(11,15)), null, span(chunk(6,10), chunk(11,15)) },
+        { span(chunk(6,10), chunk(11,15)), span(), span(chunk(6,10), chunk(11,15)) },
+        { span(chunk(6,10), chunk(11,15)), span(chunk(6, 0)), span(chunk(6,10), chunk(11,15)) },
```
droazen: You are creating chunks with end < start -- is this allowed?

src/main/java/htsjdk/samtools/SamReader.java
```diff
@@ -381,7 +381,7 @@ public PrimitiveSamReaderToSamReaderAdapter(final PrimitiveSamReader p, final Sa
             this.resource = resource;
         }
 
-        PrimitiveSamReader underlyingReader() {
+        public PrimitiveSamReader underlyingReader() {
```
droazen: Can you explain why you can't just use the existing `iterator(final SAMFileSpan chunks)` on the `PrimitiveSamReaderToSamReaderAdapter`, instead of making `underlyingReader()` public, calling it on the adapter, and then calling your new `getIterator(long firstRecordPointer)` method on it.

src/main/java/htsjdk/samtools/BAMFileReader.java
droazen: We were doing a collaborative review on this branch, so we needed our comments to be visible to everyone right away rather than at the end of the review. Sorry for the inconvenience.

droazen:
First-pass review complete, back to @tomwhite to answer some of our questions.

src/main/java/htsjdk/samtools/BAMFileReader.java
tomwhite: Removed this method as it was not needed.

src/main/java/htsjdk/samtools/BAMFileReader.java
tomwhite: Yes - not needed in the latest version of the Hadoop-BAM client code. Well spotted!

src/main/java/htsjdk/samtools/BAMFileSpan.java
tomwhite: Yes, this was intentional as the previous behaviour was incorrect. If a chunk entirely preceded the cutoff, it would never be removed.

There were no tests for this before, and the method is not used internally by htsjdk, which is possibly how this was missed. I've now added unit tests.

src/test/java/htsjdk/samtools/BAMFileSpanTest.java
tomwhite: Yes. End is not used, so it can be any value.

src/main/java/htsjdk/samtools/SamReader.java
tomwhite: `PrimitiveSamReaderToSamReaderAdapter` is not public - but even if it was, the underlying reader is still needed to get the `BAMFileReader` so its newly-exposed `getFileSpan()` and `createIndexIterator()` methods can be called.

tomwhite:
Thanks for the detailed reviews @droazen and @lbergelson. I've addressed all the comments.

src/main/java/htsjdk/samtools/BAMFileSpan.java
```diff
@@ -115,15 +115,55 @@ public SAMFileSpan removeContentsBefore(final SAMFileSpan fileSpan) {
...
+
+        final BAMFileSpan bamFileSpan = (BAMFileSpan)fileSpan;
+
+        if(bamFileSpan.isEmpty())
+            return clone();
+
+        validateSorted();
+
+        final BAMFileSpan trimmedChunkList = new BAMFileSpan();
+        final long chunkEnd = bamFileSpan.chunks.get(0).getChunkEnd();
```
droazen: This doesn't seem right -- the logical "end" of the `BAMFileSpan` is the end of the **last** chunk, not the end of the **first** chunk, isn't it, since the chunks list is sorted? 

Can you fix this, and add a unit test that fails before the change and passes after?

src/main/java/htsjdk/samtools/BAMFileReader.java
```diff
@@ -732,25 +732,48 @@ private void assertIntervalsOptimized(final QueryInterval[] intervals) {
...
+     * @param contained if <code>true</code>, return records that are strictly
+     *                  contained in the intervals, otherwise return records that overlap
+     * @param filePointers file pointer pairs corresponding to chunk boundaries for the
+     *                     intervals
+     */
+    public CloseableIterator<SAMRecord> createIndexIterator(final QueryInterval[] intervals,
+                                                            final boolean contained,
+                                                            final long[] filePointers) {
+
+        assertIntervalsOptimized(intervals);
```
droazen: This sanity check (that the intervals are sorted, and that overlapping/adjacent intervals are properly merged) now happens much later than previously when calling the two-argument overload of this method (which is what the `query()` methods all call). Specifically, the check happens after the index is actually queried in `getFileSpan()`. Are we certain that the process of querying the index is guaranteed not to blow up if the intervals were not optimized?

src/main/java/htsjdk/samtools/SamReader.java
droazen: Add docs for this method, since you've made it public

src/main/java/htsjdk/samtools/BAMFileSpan.java
```diff
@@ -115,15 +115,55 @@ public SAMFileSpan removeContentsBefore(final SAMFileSpan fileSpan) {
         validateSorted();
 
         final BAMFileSpan trimmedChunkList = new BAMFileSpan();
+        final long chunkStart = bamFileSpan.chunks.get(0).getChunkStart();
         for(final Chunk chunkToTrim: chunks) {
-            if(chunkToTrim.getChunkEnd() > chunkToTrim.getChunkStart()) {
-                if(chunkToTrim.getChunkStart() >= bamFileSpan.chunks.get(0).getChunkStart()) {
+            if(chunkToTrim.getChunkEnd() > chunkStart) {
```
droazen: Just want to confirm that you now have at least one unit test that fails before this bug fix (`chunkToTrim.getChunkEnd() > chunkToTrim.getChunkStart()` -> `chunkToTrim.getChunkEnd() > chunkStart`) and passes after it.

src/main/java/htsjdk/samtools/BAMFileSpan.java
```diff
@@ -115,15 +115,55 @@ public SAMFileSpan removeContentsBefore(final SAMFileSpan fileSpan) {
...
+        final long chunkEnd = bamFileSpan.chunks.get(0).getChunkEnd();
+        for(final Chunk chunkToTrim: chunks) {
+            if(chunkToTrim.getChunkStart() < chunkEnd) {
+                if(chunkToTrim.getChunkEnd() <= chunkEnd) {
+                    // This chunk from the list is completely before the end of the filtering chunk.
+                    trimmedChunkList.add(chunkToTrim.clone());
+                }
+                else {
+                    // This chunk from the list partially overlaps the filtering chunk and must be trimmed.
+                    trimmedChunkList.add(new Chunk(chunkToTrim.getChunkStart(),chunkEnd));
```
droazen: Do we have a test to confirm that the chunks in the returned filespan are still sorted after trimming? (applies to `removeContentsBefore()` as well)

src/test/java/htsjdk/samtools/BAMFileSpanTest.java
```diff
@@ -0,0 +1,66 @@
...
+        { span(chunk(6,10), chunk(11,15)), null, span(chunk(6,10), chunk(11,15)) },
+        { span(chunk(6,10), chunk(11,15)), span(), span(chunk(6,10), chunk(11,15)) },
+        { span(chunk(6,10), chunk(11,15)), span(chunk(0,6)), span() },
+        { span(chunk(6,10), chunk(11,15)), span(chunk(0,7)), span(chunk(6,7)) },
+        { span(chunk(6,10), chunk(11,15)), span(chunk(0,9)), span(chunk(6,9)) },
+        { span(chunk(6,10), chunk(11,15)), span(chunk(0,10)), span(chunk(6,10)) },
+        { span(chunk(6,10), chunk(11,15)), span(chunk(0,11)), span(chunk(6,10)) },
+        { span(chunk(6,10), chunk(11,15)), span(chunk(0,12)), span(chunk(6,10), chunk(11,12)) },
+        { span(chunk(6,10), chunk(11,15)), span(chunk(0,15)), span(chunk(6,10), chunk(11,15)) },
+        { span(chunk(6,10), chunk(11,15)), span(chunk(0,16)), span(chunk(6,10), chunk(11,15)) },
```
droazen: What's missing here (and in the test cases for `removeContentsBefore()` as well) are test cases in which the `cutoff` `BAMFileSpan` consists of multiple chunks. As mentioned above, I think you have a bug in `removeContentsAfter()` in that case.

src/main/java/htsjdk/samtools/BAMFileReader.java
droazen: Probably we should ensure that `assertIntervalsOptimized()` is called as the very first step in both overloads of `createIndexIterator()` to be safe, even if that means that the two-argument version can't call directly into the three-argument version.

src/main/java/htsjdk/samtools/BAMFileSpan.java
tomwhite: Yes, five new unit tests fail without this change.

src/main/java/htsjdk/samtools/BAMFileSpan.java
tomwhite: I was assuming there was only one chunk, but you're right - there could be more. Fixed, and added more tests to cover these cases.

src/main/java/htsjdk/samtools/BAMFileSpan.java
tomwhite: Yes, the test checks that the chunks are in the order specified (which is sorted).

src/main/java/htsjdk/samtools/BAMFileReader.java
tomwhite: Done.

src/main/java/htsjdk/samtools/SamReader.java
tomwhite: Done

src/test/java/htsjdk/samtools/BAMFileSpanTest.java
tomwhite: Done.

tomwhite:
Thanks for the review @droazen. I've addressed all your comments.

droazen:
:+1: latest version looks good -- merging

Issue #776, opened 2017-01-04
non-deterministic test failure in AsyncBufferedIteratorTest


lbergelson:
I'm seeing a random failure of AsynchBufferedIteratorTest.testBackground blocks.  

```
htsjdk.samtools.util.AsyncBufferedIteratorTest.testBackgroundBlocks FAILED
    java.lang.AssertionError: expected [9] but found [0]
        at org.testng.Assert.fail(Assert.java:94)
        at org.testng.Assert.failNotEquals(Assert.java:496)
        at org.testng.Assert.assertEquals(Assert.java:125)
        at org.testng.Assert.assertEquals(Assert.java:372)
        at org.testng.Assert.assertEquals(Assert.java:382)
        at htsjdk.samtools.util.AsyncBufferedIteratorTest.testBackgroundBlocks(AsyncBufferedIteratorTest.java:79)
```

The test has a comment saying that it's subject to race conditions.  

```
  Thread.sleep(10); // how do we write this test and not be subject to race conditions?
        // should have read 9 records: 2*3 in the buffers, and another 3 read but
        // blocking waiting to be added 
```

tfenne:
Mitigated by #790 (2.10.0), which replaced the fixed sleep with a poll; no failures reported since.

PR #790, opened 2017-01-20
Defensive programming around test case race condition to address #776

d-cameron:
### Description

AsyncBufferedIteratorTest.testBackgroundBlocks() is subject to a race condition as it is a testing the async loading of records in AsyncBufferedIterator, but is not privy to the internal state of the iterator. Additional logic has been added such that the test case race condition is now extremely unlikely to be encountered.


d-cameron:
@lbergelson is the codecov failure here due to the non-deterministic nature of the test case as it is now, or is there something else I'm missing?

lbergelson:
@d-cameron I'm honestly not quite sure what's going on with that.  We've been seeing it a lot.  I've been attributing it to the pull request not being rebased on the head of master, but yours is.  It shows you gaining a few lines in AsyncBlockCompressedInputStream and losing Md5CalculatingOutputStream.java.  Both seem plausibly caused by test non-determinism. 

I'll just disable that check since we're basically treating it as a false positive and I'm not sure it would ever have added value over the total project coverage test.

d-cameron:
Is there any way to retrigger the coverage checks? As it stand, it's causing the PR to incorrectly appear as if it's not suitable for merging yet.

lbergelson:
@d-cameron Sorry, I keep meaning to reply to this and then getting distracted.  Could we change this to use a CountDownLatch instead of the extended sleep and check?  It would be cleaner and avoid any sleeping.

yfarjoun:
I think that this PR is fine as is.  👍

PR #795, opened 2017-02-01
Don't set a custom default InflaterFactory when testing passing in an explicit InflaterFactory

droazen:


droazen:
Quickie for @lbergelson

PR #803, opened 2017-02-10
add a coverage drop threshold to limit codecov failure spam

lbergelson:
Codecov is adding failure messages for very small coverage changes that aren't worrisome.  I'm adding a threshold that should hopefully fix it so only more significant coverage drops are reported as failures.


droazen:
:+1: merge when ready

PR #804, opened 2017-02-10
Changed parameter names and doc to be consistent in use of "offset" a…

tfenne:
…nd "position".

### Description

It's been driving me crazy that all the methods to translate from read position to reference position and vice versa in SAMRecord sometimes refer to position (or 1-based position) and sometimes refer to offset.  They all work the same way (take and return 1-based positions) but some of them refer to the parameters and returns as "offsets".  I'd argue that "offset" is inherently 0-based, and thus this causes some confusion.  I've just tidied up the javadoc, and rename some parameters from offset to position.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMRecord.java
+++ b/src/main/java/htsjdk/samtools/SAMRecord.java
@@ -633,7 +633,8 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
      * Non static version of the static function with the same name.
      *
      * @param position 1-based location within the unclipped sequence
-     * @return 1-based reference position of the unclipped sequence at a given read position
+     * @return 1-based reference position of the unclipped sequence at a given read position,
+     *         or 0 if there is no position.
      */
     public int getReferencePositionAtReadPosition(final int position) {
         return getReferencePositionAtReadPosition(this, position);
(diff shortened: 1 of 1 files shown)


src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -630,47 +630,53 @@ public int getUnclippedEnd() {
 
 
     /**
-     * @param offset 1-based location within the unclipped sequence or 0 if there is no position.
-     * <p/>
      * Non static version of the static function with the same name.
-     * @return 1-based inclusive reference position of the unclipped sequence at a given offset,
+     *
+     * @param position 1-based location within the unclipped sequence
+     * @return 1-based reference position of the unclipped sequence at a given read position
```
nh13: what happened to the "0 if there is no position"?

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -681,37 +687,43 @@ public int getReadPositionAtReferencePosition(final int pos) {
...
     /**
-     * @param rec record to use
-     * @param pos 1-based reference position
-     * @param returnLastBaseIfDeleted if positive, and reference position matches a deleted base in the read, function will
-     * return the offset
-     * @return 1-based (to match getReferencePositionAtReadPosition behavior) inclusive position into the
-     * unclipped sequence at a given reference position,
-     * or 0 if there is no such position. If returnLastBaseIfDeleted is true deletions are assumed to "live" on the last read base
-     * in the preceding block.
+     * Returns the 1-based position in the read of the provided reference position, or zero if no
```
nh13: `zero` -> `0`

src/main/java/htsjdk/samtools/SAMRecord.java
tfenne: Whoops.

PR #808, opened 2017-02-21
adding CloseableIterator.asyncMap(f) for Yossi to play with, don't merge

lbergelson:
@yfarjoun This is the thing we were talking about on friday.  From playing around with it it sped up my tests on the uncompressed bam from ~1 minute to ~35s.  Just reading the raw bytes of the uncompressed bam takes ~30s so it's doing a pretty decent job of distributing work.

```
    @Test
    public void testAsyncBlockBuffer(){
        final SamReader reader = SamReaderFactory.makeDefault().setUseAsyncIo(false).validationStringency(ValidationStringency.SILENT)
                .open(new File("/Users/louisb/data/uncompressed.bam"));
        Assert.assertEquals(reader.iterator().asyncMap(record -> {record.eagerDecode(); return record;}).stream().count(), 26028704);
    }

```

This returns a new iterator which applies f to the contents of the
previous iterator in a separate set of threads.

It pulls the underlying iterator repeatedly to build chunks, and then
executes those chunks asynchronously on a thread pool.  The
implementation is pretty rough and could probably be made more
efficient. (and no tests are attached... use at your
own risk...)

### DIFF
--- a/src/main/java/htsjdk/samtools/util/AsyncChunkedOperationIterator.java
+++ b/src/main/java/htsjdk/samtools/util/AsyncChunkedOperationIterator.java
@@ -1,0 +1,82 @@
+package htsjdk.samtools.util;
+
+import java.util.*;
+import java.util.concurrent.*;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class AsyncChunkedOperationIterator<T,R> implements CloseableIterator<R> {
+
+    private final CloseableIterator<T> input;
+    private final Function<T,R> f;
+    private final ExecutorService pool = Executors.newFixedThreadPool(4);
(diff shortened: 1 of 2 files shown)


yfarjoun:
Yeah, I changed my implementation to use chunks...and it's faster now. the problem is that encoding is the show-stopper. even with async IO, the writing thread is using all the CPU, but it's spending it encoding the bamrecords...I think we need the same functionality for _encoding_....I'll try it out. I have an idea how to do it. using the async iterator.

lbergelson:
@yfarjoun I think we've accidentally reimplemented AsyncBufferedIterator in a less thorough manner...

PR #809, opened 2017-02-27
fixing vararg warnings on test compilation

lbergelson:
fixing the 4 annoying warnings that pop up whenever you compile tests

I also added a test case that seems to have been accidentally forgotten, the variable was constructed but not added to the dataprovider.

the first commit just fixes the extraneous <type>, the second commit has the casts and the extra test case


droazen:
:+1: merge when ready

PR #810, opened 2017-03-01
Tribble/Tabix index path support

magicDGS:
#### Motivation
While trying to index a `Path` (no convertible to `File`) using the `Index` interface, the client should:
- Create the `LittleEndianOutputStream`by its own. This may lead to wrong indexes because they aren't block-compressed.
- Get the index file name by its own if it is based on a feature file. This may lead to incorrect index names.
In addition, this is part of a bigger change for include support of writing in whatever filesystem (and also reading).
#### Changes
- Clean up (first commit):  removing mentions to vcf in javadoc for `Tribble` and unused imports.
- Path support for `Index` interface: methods with default values use `Path.toFile()`
- Implementation of the new methods in `AbstractIndex` and `TabixIndex`

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/tribble/Tribble.java
+++ b/src/main/java/htsjdk/tribble/Tribble.java
@@ -37,9 +37,9 @@ public class Tribble {
     public final static String STANDARD_INDEX_EXTENSION = ".idx";
 
     /**
-     * Return the name of the index file for the provided vcf {@code filename}
+     * Return the name of the index file for the provided {@code filename}
      * Does not actually create an index
-     * @param filename  name of the vcf file
+     * @param filename  name of the file
      * @return non-null String representing the index filename
      */
     public static String indexFile(final String filename) {
@@ -47,9 +47,9 @@ public class Tribble {
     }
 
     /**
-     * Return the File of the index file for the provided vcf {@code file}
+     * Return the File of the index file for the provided {@code file}
      * Does not actually create an index
-     * @param file  the vcf file
+     * @param file  the file
      * @return a non-null File representing the index
      */
     public static File indexFile(final File file) {
@@ -57,9 +57,9 @@ public class Tribble {
     }
 
     /**
-     * Return the name of the tabix index file for the provided vcf {@code filename}
+     * Return the name of the tabix index file for the provided {@code filename}
      * Does not actually create an index
-     * @param filename  name of the vcf file
+     * @param filename  name of the file
      * @return non-null String representing the index filename
      */
     public static String tabixIndexFile(final String filename) {
@@ -67,9 +67,9 @@ public class Tribble {
     }
 
     /**
-     * Return the File of the tabix index file for the provided vcf {@code file}
+     * Return the File of the tabix index file for the provided {@code file}
      * Does not actually create an index
-     * @param file  the vcf file
+     * @param file  the file
      * @return a non-null File representing the index
      */
     public static File tabixIndexFile(final File file) {
@@ -77,9 +77,9 @@ public class Tribble {
     }
 
     /**
-     * Return the name of the index file for the provided vcf {@code filename} and {@code extension}
+     * Return the name of the index file for the provided {@code filename} and {@code extension}
      * Does not actually create an index
-     * @param filename  name of the vcf file
+     * @param filename  name of the file
      * @param extension the extension to use for the index
      * @return non-null String representing the index filename
      */
@@ -88,9 +88,9 @@ public class Tribble {
     }
 
     /**
-     * Return the File of the index file for the provided vcf {@code file} and {@code extension}
+     * Return the File of the index file for the provided {@code file} and {@code extension}
      * Does not actually create an index
-     * @param file  the vcf file
+     * @param file  the file
      * @param extension the extension to use for the index
      * @return a non-null File representing the index
      */
--- a/src/main/java/htsjdk/tribble/Tribble.java
+++ b/src/main/java/htsjdk/tribble/Tribble.java
@@ -58,13 +58,13 @@ public class Tribble {
     }
 
     /**
-     * Return the name of the index file for the provided {@code filename}
+     * Return the name of the index file for the provided {@code path}
      * Does not actually create an index
-     * @param path  name of the path
-     * @return non-null String representing the index filename
+     * @param path the path
+     * @return Path representing the index filename
      */
-    public static String indexPath(final Path path) {
-        return indexFile(path.toAbsolutePath().toString());
+    public static Path indexPath(final Path path) {
+        return path.getFileSystem().getPath(indexFile(path.toAbsolutePath().toString()));
     }
 
     /**
@@ -90,7 +90,7 @@ public class Tribble {
     /**
      * Return the name of the tabix index file for the provided {@code path}
      * Does not actually create an index
-     * @param path  name of the path
+     * @param path the path
      * @return non-null String representing the index filename
      */
     public static String tabixIndexPath(final Path path) {
--- a/src/main/java/htsjdk/tribble/Tribble.java
+++ b/src/main/java/htsjdk/tribble/Tribble.java
@@ -91,10 +91,10 @@ public class Tribble {
      * Return the name of the tabix index file for the provided {@code path}
      * Does not actually create an index
      * @param path the path
-     * @return non-null String representing the index filename
+     * @return Path representing the index filename
      */
-    public static String tabixIndexPath(final Path path) {
-        return indexFile(path.toAbsolutePath().toString());
+    public static Path tabixIndexPath(final Path path) {
+        return path.getFileSystem().getPath(tabixIndexFile(path.toAbsolutePath().toString()));
     }
 
     /**
--- a/src/main/java/htsjdk/tribble/Tribble.java
+++ b/src/main/java/htsjdk/tribble/Tribble.java
@@ -27,6 +27,7 @@ import htsjdk.tribble.util.ParsingUtils;
 import htsjdk.tribble.util.TabixUtils;
 
 import java.io.File;
+import java.nio.file.Path;
 
 /**
  * Common, tribble wide constants and static functions
@@ -57,6 +58,16 @@ public class Tribble {
     }
 
     /**
+     * Return the name of the index file for the provided {@code filename}
+     * Does not actually create an index
+     * @param path  name of the path
+     * @return non-null String representing the index filename
+     */
+    public static String indexPath(final Path path) {
+        return indexFile(path.toAbsolutePath().toString());
+    }
+
+    /**
      * Return the name of the tabix index file for the provided {@code filename}
      * Does not actually create an index
      * @param filename  name of the file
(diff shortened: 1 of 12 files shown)


lbergelson:
@magicDGS Something went wrong here, tests are failing all over.

magicDGS:
I figured it out already, I'm trying to find the problem.

magicDGS:
Very awful error from my side, sorry.  The `writeBasedOnFeatureFile` was overwriting the input file instead of using the name for making the index.

Now I guess that it's solved. Could you have a quick review, @lbergelson?

magicDGS:
This is still WIP, because I found some other paths in the indexing code that should use Path for fully support. So this should wait for review, @lbergelson.

magicDGS:
This should be ok now, @lbergelson. Previously failing tests are passing in my computer.

magicDGS:
Can you have a look to this one @lbergelson? This is going to be important for my upcoming PR with tribble writing support with indexing on the fly.

Thanks a lot!

src/main/java/htsjdk/tribble/Tribble.java
```diff
@@ -37,49 +38,69 @@ private Tribble() { } // can't be instantiated
...
+     * @param file  the file
      * @return a non-null File representing the index
      */
     public static File indexFile(final File file) {
         return indexFile(file.getAbsoluteFile(), STANDARD_INDEX_EXTENSION);
     }
 
     /**
-     * Return the name of the tabix index file for the provided vcf {@code filename}
+     * Return the name of the index file for the provided {@code filename}
```
magicDGS: Done.

src/main/java/htsjdk/tribble/Tribble.java
```diff
@@ -37,49 +38,69 @@ private Tribble() { } // can't be instantiated
...
      * Does not actually create an index
-     * @param filename  name of the vcf file
+     * @param path  name of the path
```
magicDGS: Done.

src/main/java/htsjdk/tribble/Tribble.java
```diff
@@ -37,49 +38,69 @@ private Tribble() { } // can't be instantiated
...
+     * @return non-null String representing the index filename
+     */
+    public static String indexPath(final Path path) {
```
magicDGS: I changed it.

src/main/java/htsjdk/tribble/index/AbstractIndex.java
```diff
@@ -67,7 +70,7 @@
     private final static long NO_TS = -1L;
 
     protected int version;                    // Our version value
-    protected File indexedFile = null;         // The file we've created this index for
+    protected Path indexedPath = null;         // The file we've created this index for
```
magicDGS: Oh, I haven't realize that it is breaking compatibility. If we do not want to break compatibility, I can kept the field with the deprecated annotation and set it when the used constructor is a `File`. Let me know if I should do that.

src/main/java/htsjdk/tribble/index/AbstractIndex.java
```diff
@@ -201,7 +213,11 @@ public boolean isCurrentVersion() {
     }
 
     public File getIndexedFile() {
-        return indexedFile;
+        return getIndexedPath().toFile();
```
magicDGS: Done.

src/main/java/htsjdk/tribble/index/AbstractIndex.java
```diff
@@ -251,7 +271,7 @@ private void writeHeader(final LittleEndianOutputStream dos) throws IOException
         dos.writeInt(MAGIC_NUMBER);
         dos.writeInt(getType());
         dos.writeInt(version);
-        dos.writeString(indexedFile.getAbsolutePath());
+        dos.writeString(indexedPath.toString());
```
magicDGS: Done.

src/main/java/htsjdk/tribble/index/DynamicIndexCreator.java
```diff
@@ -56,13 +57,18 @@
     MathUtils.RunningStat stats = new MathUtils.RunningStat();
     long basesSeen = 0;
     Feature lastFeature = null;
-    File inputFile;
+    // TODO: actually this field is not needed
+    Path inputPath;
```
magicDGS: In this case, because it is package protected, I do not think that is going to harm at all (unless someone is subclassing in the same package outside HTSJDK).

If we are breaking compatibiity anyway, as it is unused, why don't we just remove the field (no the constructor arg)? It is only used for get the index creators...

src/main/java/htsjdk/tribble/index/Index.java
```diff
@@ -78,13 +79,37 @@
...
 
     /**
+     * Writes the index into a path.
+     *
+     * Default implementation converts the path to a file and uses {@link #write(File)}.
+     *
+     * @param indexPath Where to write the index.
+     * @throws IOException if the index is unable to write to the specified path.
+     */
+    public default void write(final Path indexPath) throws IOException {
```
magicDGS: Done.

src/main/java/htsjdk/tribble/index/Index.java
```diff
@@ -78,13 +79,37 @@
...
 
     /**
+     * Write an appropriately named and located Index file based on the name and location of the featureFile.
+     * If featureFile is not a normal file, the index will silently not be written.
+     *
+     * Default implementation converts the path to a file and uses {@link #writeBasedOnFeatureFile(File)}.
+     *
+     * @param featurePath
+     */
+    public default void writeBasedOnFeaturePath(Path featurePath) throws IOException {
```
magicDGS: Using OperationUnsupportedException for now.

src/main/java/htsjdk/tribble/index/tabix/TabixIndex.java
```diff
@@ -209,6 +213,18 @@ public void write(final File tabixFile) throws IOException {
     }
 
     /**
+     * Writes the index with BGZF.
+     *
+     * @param tabixPath Where to write the index.
+     */
+    @Override
```
magicDGS: Done.

src/main/java/htsjdk/tribble/index/tabix/TabixIndex.java
```diff
@@ -220,6 +236,17 @@ public void writeBasedOnFeatureFile(final File featureFile) throws IOException {
     }
 
     /**
+     * Writes to a path with appropriate name and directory based on feature path.
+     *
+     * @param featurePath Path being indexed.
+     */
+    @Override
+    public void writeBasedOnFeaturePath(final Path featurePath) throws IOException {
```
magicDGS: Done.

src/main/java/htsjdk/tribble/index/DynamicIndexCreator.java
magicDGS: Removed the unused field in the commit for breaking compatibility.

src/main/java/htsjdk/tribble/index/Index.java
magicDGS: Changed the interface in the commit for breaking compatibility.

magicDGS:
I added two commits, @lbergelson:

1. Addressing most of your comments that does not imply backwards incompatibility.
2. Breaking compatibility for making simpler the interfaces. This could be removed if @droazen thinks that it is better to keep backwards compatibility for subclassing indexes.

Regarding the test, I don't know if I will have time today. I will try, but if not I will add them by tomorrow. You can review the current changes anyway, in case that something else should be changed.

magicDGS:
I've just realized that something is broken with the changes, @lbergelson. Do you have an idea of what is happening? It looks that is breaking indexing on the fly for gzip VCF result (not the indexing itself). The most likely reson is the change to delegate the file constructor to the path one for files, because `BlockCompressedOutputStream` does not have `Path` support (see #811) and the filename is not set correctly in the binary codec.

I can add a simple constructor to `BlockCompressedOutputStream` with a String filename argument, but the change of that class is out of the scope of this commit.

src/main/java/htsjdk/tribble/index/tabix/TabixIndex.java
```diff
@@ -219,8 +219,8 @@ public void write(final Path tabixPath) throws IOException {
      */
     @Override
     public void writeBasedOnFeaturePath(final Path featurePath) throws IOException {
-        if (Files.isRegularFile(featurePath)) return;
-        write(IOUtil.getPath(Tribble.tabixIndexPath(featurePath)));
+        if (!Files.isRegularFile(featurePath)) return;
```
lbergelson: Ack, good catch.  That's terrible behavior.  I wonder why it was done that way?  Definitely needs to at least log a warning.  I would also prefer your suggestion to fail with a TribbleException, but I think we'd need to investigate if that would break existing downstream code somehow.  For now, could you have it log a warning, and open a new ticket to change the behavior?

src/main/java/htsjdk/tribble/index/Index.java
```diff
@@ -73,41 +73,42 @@
     /**
      * Writes the index into a file.
      *
+     * Default implementation delegates in {@link #write(Path)}
```
lbergelson: Sorry to nitpick the language, but delegates **to** instead of delegates **in** is more idiomatic.

src/main/java/htsjdk/tribble/index/Index.java
```diff
@@ -73,41 +73,42 @@
...
-    public default void write(final Path indexPath) throws IOException {
-        write(indexPath.toFile());
-    }
+    public void write(final Path indexPath) throws IOException;
 
     /**
      * Write an appropriately named and located Index file based on the name and location of the featureFile.
      * If featureFile is not a normal file, the index will silently not be written.
+     *
+     * Default implementation delegates in {@link #writeBasedOnFeaturePath(Path)}
```
lbergelson: same as above

src/main/java/htsjdk/tribble/index/tabix/TabixIndex.java
lbergelson: @magicDGS I've opened the issue here: #821

src/main/java/htsjdk/tribble/index/AbstractIndex.java
lbergelson: Yeah, it's tricky.  Anything field that's visible to a subclass of a non-final class is technically a breaking change since someone could be relying on it if they implement a subclass.  It's why I'm so strongly in favor of making classes final and of making variable private with accessors if they need to be accessed.

lbergelson:
@magicDGS Looks good.  👍 Once we have the tests for the two index types.

src/main/java/htsjdk/tribble/index/tabix/TabixIndex.java
magicDGS: Added warning log here and in `AbstractIndex`.

src/main/java/htsjdk/tribble/index/Index.java
magicDGS: Thanks for this, I really appreciate to have comments to improve my English skills. Done!

src/main/java/htsjdk/tribble/index/Index.java
magicDGS: Done.

magicDGS:
Three more commits:

1. Minor changes from review.
2. A couple of missing constructors for Path (to test)
3. Small test for using path in writing and loading.

Thanks for reviewing, @lbergelson - Back to you!

lbergelson:
👍

magicDGS:
Thanks!

Issue #812, opened 2017-03-02
Suggestion: use Locatable in tribble instead of Feature


magicDGS:
Currently, the only method in `Feature` is deprecated, and `Locatable` represents the same structure. Tribble is supposed to handle "features" which maps to the genome, and thus using `Locatable` will help to provide a broader scope of the classes contained here.

I propose the following changes to minimize the backwards compatibility:
- Deprecate `Feature` in favor of `Locatable`and setting a date/release to remove `Feature`to allow API clients to change their code (I'm also affected by that). When the deadline is reached, `Feature` will be removed from HTSJDK and/or all implemented features will become `Locatable`s.
- Change tribble generic classes and interfaces to use `Locatable` instead of `Feature`. This is backwards compatible, because `Feature`extends `Locatable`. API client code implementing/extending this classes are already bounded with a `Feature`, and this won't introduce any change in their code.

This will reduce the complexity and duplicated code for managing genomic locations, and will benefit from `Locatable` code as the one implemented in #443. What are the community thoughts about this? I will open a pull request with the changes proposed to start rolling the ball....

PR #813, opened 2017-03-02
Use Locatable in tribble instead of Feature (DO NOT MERGE: discussion)

magicDGS:
### Description

- Deprecate `Feature` in favor of `Locatable`.
- Change tribble interfaces and generic classes to use `Locatable`. This is backwards compatible, because `Feature` extends `Locatable`. API client code implementing/extending this classes are already bounded with a `Feature`, and this won't introduce any change in their code (except for deprecation warnings).

### For  discussion
- [ ] Set a date/release for removal of `Feature` 
- [ ] Set a date/release for change HTSJDK `Feature` classes to implement `Locatable`: `BEDFeature`, `VariantContext` and `SimpleFeature`.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/tribble/AbstractFeatureCodec.java
+++ b/src/main/java/htsjdk/tribble/AbstractFeatureCodec.java
@@ -23,6 +23,8 @@
  */
 package htsjdk.tribble;
 
+import htsjdk.samtools.util.Locatable;
+
 import java.io.IOException;
 
 /**
@@ -31,7 +33,7 @@ import java.io.IOException;
  * Note that that method is the only way that the right codec for a file is identified and that <bold>only one</bold> codec
  * is allowed to identify itself as being able to decode any given file.
  */
-public abstract class AbstractFeatureCodec<FEATURE_TYPE extends Feature, SOURCE> implements FeatureCodec<FEATURE_TYPE, SOURCE> {
+public abstract class AbstractFeatureCodec<FEATURE_TYPE extends Locatable, SOURCE> implements FeatureCodec<FEATURE_TYPE, SOURCE> {
     private final Class<FEATURE_TYPE> myClass;
 
     protected AbstractFeatureCodec(final Class<FEATURE_TYPE> myClass) {
@@ -39,7 +41,7 @@ public abstract class AbstractFeatureCodec<FEATURE_TYPE extends Feature, SOURCE>
     }
     
     @Override
-    public Feature decodeLoc(final SOURCE source) throws IOException {
+    public Locatable decodeLoc(final SOURCE source) throws IOException {
         return decode(source);
     }
 
--- a/src/main/java/htsjdk/tribble/AbstractFeatureReader.java
+++ b/src/main/java/htsjdk/tribble/AbstractFeatureReader.java
@@ -18,6 +18,7 @@
 
 package htsjdk.tribble;
 
+import htsjdk.samtools.util.Locatable;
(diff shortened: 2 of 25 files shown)


magicDGS:
Discussion for #812.

magicDGS:
@lbergelson, @yfarjoun, @tfenne, @vadimzalunin, @cmnbroad, @droazen -- because this discussion is related with the evolving API one (https://github.com/samtools/htsjdk/issues/378), can you contribute to this one and discuss about it?

I also would like some feedback for @jrobinso, because IGV depends strongly on the tribble package.

If you know other people who may be interested in chime in, could you please include them here? Thanks in advance for your help!

jrobinso:
What is the motivation for deprecating Feature?  The bar should be pretty high for such a high level interface.
 
I appreciate the ping,  a change of getChr to getContig caused major unexpected disruption some time ago.   That one still causes hundreds of deprecated warnings during compile.

magicDGS:
Thanks for chime in, @jrobinso! The motivation is that `Feature` is a single-method interface, and that method is deprecated for the `Locatable.getContig()`. Thus, I think that everything that requires a `Feature` will just use methods from `Locatable` instead. In addition, I think that an evolving API should get rid of the deprecated methods at some point, and that requires to make some changes that the users should take into account when updating their libraries. 

Regarding the warnings about `getChr()` in IGV, I will contribute with a PR there to remove its usage to help in the transition. If the `Feature` is deprecated, the API users should move their implementations to `Locatable`as soon as possible, and I can also contribute in case the IGV team needs help on that.

magicDGS:
Sorry, I forgot to ask you about the change to use the `Locatable` interface in tribble instead of `Feature`. Is the rest of the PR, apart of the deprecation, alright for IGV @jrobinso?

jrobinso:
To be honest I wasn't involved with the rename of getChr ot getContig, it
seemed a bit gratuitous to be honest and didn't add anything useful.    Now
it seems that change is motivating another disruptive change.

Don't submit a PR yet,  I might break IGV's dependence on tribble by
forking and copying files.  IGV's needs for feature classes differ from the
analysis packages,  and perhaps they should evolve independently.  Tribble
started as an IGV package,  I developed most of it,  and was moved out in a
restructuring of libraries some years ago.   Its gotten very heavy over
time for IGV needs,  particularly VariantContext.   I have very lightweight
implementations in javascript that are all IGV requires and would save
significant memory,  which is at a premium for the desktop.   So I'm
seriously considering porting these implementations to Java as sort of a
"tribble lite".   I would still use the indexing code,  but not the Feature
/ Locatable packages.

droazen:
The problem with this proposal is that downstream code (like GATK3/4) really needs to be able to distinguish between `Features` (ie., record formats for which tribble has support) and other kinds of `Locatables`. We often need to use type bounds like `? extends Feature` in our generic classes to express things like "this class must deal with tribble-compatible records only", and we would lose the ability to do this if we eliminated `Feature`. We also sometimes use reflection to do things like "find all supported tribble formats on the classpath" by searching for subclasses of `Feature`. In other words, `Feature` is valuable for us simply as a marker interface.

For these reasons, I have to oppose this change as both removing useful existing functionality and causing unnecessary breakage.

magicDGS:
Ok, now I see the point of the `Feature` as a marker. Maybe it would be nice to have this purpose in the javadoc to keep things clear. Thanks for the feedback!

PR #815, opened 2017-03-02
Fixed bug where calling IntervalList.uniqued(false) would fail if the…

tfenne:
… interval names were null.

### Description

`IntervalList.uniqued(false)` would throw an exception if called on Interval objects that had null names (which is allowed).

### Checklist

- [x] Code compiles correctly
- [ ] All tests passing


tfenne:
@yfarjoun Tiny PR, mind taking a look for me?

PR #817, opened 2017-03-07
WIP: Cram compatibility and roundtrip tests

vadimzalunin:
### Description

A patch to ensure htsjdk CRAM is compatible with samtools and roundtrip tests work as expected given the version. 
Main changes: 
- ReadTag removed in favour of SAMBinaryTagAndValue
- added CIGAR tests
- added custom version asserts

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


vadimzalunin:
@cmnbroad please have a look and let me know what to change - this somehow is a big patch. I still need to figure out the last failing test with insert size and improve the other things like javadoc etc. A more intelligent SAMRecord comparator will follow shortly.

cmnbroad:
@vadimzalunin I was just taking a quick look at this - if you haven't already squashed the commits I'd say to leave them since, as you say, they help illustrate the reasons for some of the changes. Let me know when this is ready for review.

vadimzalunin:
@cmnbroad I'm done with the changes, please review. I'm happy to slice the PR into a few commits after that.

cmnbroad:
@vadimzalunin - a couple of questions to start:

- It looks like you added support for specifying the version when writing CRAMs, and added tests that write/verify v2.1 CRAMs. Is there a reason we need to support either of those (writing 2.1, and allowing the caller to specify the version ?)
- I’m still not clear on why the test files had to be regenerated (or why we need a samtools script to do it). Did we have to do that (will the new code work on the old files), or is there some other dependency there ?
- I noticed you added a new, public method (Sam2CramRecordFactory.setPresevationPolicy), that is both deprecated, and used internally ? Not sure if that was intentional.

vadimzalunin:
@cmnbroad 
1. i did this for ease of testing, but perhaps you are right -- version support in writer doesn't have to be public.
2. reverted the test files back - they work fine any way!
3. intentional but a bit hard to explain :) 
Currently Sam2CramRecordFactory exposes captureTags/ignoreTags as public fieds. I've changed the implementation, so the fields are no longer used/exist. To provide some compatibility (who knows) I've added a new method htsjdk.samtools.cram.build.Sam2CramRecordFactory#setTagPresevationPolicy that accepts old-style arguments (aka fields). It is marked deprecated to express that this is an old way of tag capture language, it is public because of usage in htsjdk.samtools.CRAMContainerStreamWriter#flushContainer.

vadimzalunin:
@cmnbroad 
Test files reverted and visibility of the constructor in CRAMFileWriter that accepts version is reduced to protected. Please review.

src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
```diff
@@ -437,6 +457,12 @@ protected void flushContainer() throws IllegalArgumentException, IllegalAccessEx
                     final SAMRecord restoredSamRecord = f.create(cramRecords.get(i));
                     assert (restoredSamRecord.getAlignmentStart() == samRecords.get(i).getAlignmentStart());
                     assert (restoredSamRecord.getReferenceName().equals(samRecords.get(i).getReferenceName()));
+                    if (!restoredSamRecord.getReadString().equals(samRecords.get(i).getReadString())) {
```
yfarjoun: What is this `if` statement for? you're dumping a stacktrace, but continuing to run...what should actually happen here?

vadimzalunin: debug leftovers

yfarjoun:
@cmnbroad claims that there's nothing of value left here.

PR #818, opened 2017-03-08
Update javadoc for Feature

magicDGS:
### Description

As pointed out in #813, the `Feature` interface is meant to be a marker for a `Locatable` with tribble support. This is a simple documentation patch to:
1.  point out what is the purpose of this class.
2. add a date note to the deprecation of  `getChr()`. Because of the move to the gradle structure, I cannot pinpoint the exact date of deprecation, but at least we know that is before 2016 May.

The first point is to clarify why tribble is using `Feature` instead of `Locatable` (and avoid proposals like #812), and the second is to help in the API evolution (to be able to remove at some point deprecated methods).

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/tribble/Feature.java
+++ b/src/main/java/htsjdk/tribble/Feature.java
@@ -27,13 +27,14 @@ package htsjdk.tribble;
 import htsjdk.samtools.util.Locatable;
 
 /**
- * Represents a locus on a reference sequence.   All Features are expected to return 1-based closed-ended intervals.
+ * Marker interface for locatables with Tribble support.
+ * As {@link Locatable}, represents a locus on a reference sequence and is expected to return 1-based closed-ended intervals.
  */
 public interface Feature extends Locatable {
 
     /**
      * Return the features reference sequence name, e.g chromosome or contig
-     * @deprecated use getContig() instead
+     * @deprecated before 05-2017. Use getContig() instead.
      */
     @Deprecated
     default public String getChr() {
--- a/src/main/java/htsjdk/tribble/Feature.java
+++ b/src/main/java/htsjdk/tribble/Feature.java
@@ -27,14 +27,14 @@ package htsjdk.tribble;
 import htsjdk.samtools.util.Locatable;
 
 /**
- * Marker interface for locatables with Tribble support.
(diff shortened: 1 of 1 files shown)


magicDGS:
Can you have a look to this, @droazen? I think that with this change it is clear what is the meaning of the `Feature` class, and we can avoid proposals like mine. Thank you very much!

src/main/java/htsjdk/tribble/Feature.java
```diff
@@ -27,13 +27,14 @@
...
- * Represents a locus on a reference sequence.   All Features are expected to return 1-based closed-ended intervals.
+ * Marker interface for locatables with Tribble support.
+ * As {@link Locatable}, represents a locus on a reference sequence and is expected to return 1-based closed-ended intervals.
  */
 public interface Feature extends Locatable {
 
     /**
      * Return the features reference sequence name, e.g chromosome or contig
-     * @deprecated use getContig() instead
+     * @deprecated before 05-2017. Use getContig() instead.
```
droazen: "before 05-2017" -> "on 03/2015"

src/main/java/htsjdk/tribble/Feature.java
```diff
@@ -27,13 +27,14 @@
 import htsjdk.samtools.util.Locatable;
 
 /**
- * Represents a locus on a reference sequence.   All Features are expected to return 1-based closed-ended intervals.
+ * Marker interface for locatables with Tribble support.
```
droazen: Might make things clearer to add an extra sentence here: "Marker interface for Locatables with Tribble support. A Feature represents a record in a tribble-supported file format."

droazen:
Thanks for this @magicDGS -- two comments to address, then we can merge this.

src/main/java/htsjdk/tribble/Feature.java
magicDGS: Done.

src/main/java/htsjdk/tribble/Feature.java
magicDGS: Done.

magicDGS:
Thanks for reviewing, @droazen. Back to you!

droazen:
:+1:

magicDGS:
Thnaks!

Issue #393, opened 2015-11-23
Corrupted tabix index


dariober:
Hello,

It seems to me that tabix indexes created with `writeBasedOnFeatureFile` are corrupted. Am I doing something wrong?

Here's a bed test file, it has 1M rows on 10 chroms. I bgzipped with tabix 1.2.1:
https://dl.dropboxusercontent.com/u/53487723/tmp.bedGraph.gz

I created an index for it with the following program (java 8, htsjdk-1.141):

```
import java.io.File;
import java.io.IOException;
import htsjdk.tribble.bed.BEDCodec;
import htsjdk.tribble.index.IndexFactory;
import htsjdk.tribble.index.tabix.TabixFormat;
import htsjdk.tribble.index.tabix.TabixIndex;

public class TestTabix {

    public static void main(String[] args) throws IOException {

        String bgzfOut= "tmp.bedGraph.gz";
        TabixIndex tabixIndexGz =
                IndexFactory.createTabixIndex(new File(bgzfOut), new BEDCodec(), TabixFormat.BED,
                        null);
        tabixIndexGz.writeBasedOnFeatureFile(new File(bgzfOut));
    }

}
```

Querying the resulting index returns only the first records of chr1:

```
tabix tmp.bedGraph.gz chr1 | wc -l  # Should be 100000
    4500
tabix tmp.bedGraph.gz chr2 | wc -l  # Should be 100000
       0
```

magicDGS:
Can you try with the latest master version, @dariober? I fixed some bugs in the Tabix/Tribble indexing recently...

dariober:
Hi - Thanks for looking into this and apologies I kept quiet. I tested the code above on htsjdk-2.7.0 an the problem persists. 

Also, after indexing the following file (call it chrom.bed.gz):

    chr1	1	10
    chr1	100	1000000

The query

    tabix chrom.bed.gz chr1

skips the first record and returns:

    chr1	100	1000000

However, this file is correctly queried:

    chr1	1	10
    chr1	100	1000

It seems to me there is something going on when features are in the first bin 0-16384.

(Just to make sure... Can anybody reproduce this bug or it's just me?)

magicDGS:
This is also happening to me, @dariober. I will create a PR with a failing index to point out that this is happening and ask how can be fixed.

cmnbroad:
[...]

PR #820, opened 2017-03-09
Failing test for BED tabix indexing

magicDGS:
### Description

This is a failing test for tabix indexing as reported in #393.  I tried to fix it, but I cannot find the way to do it. This is a PR for the community to help us to solve the issue while creating a tabix index from a file:

* The tabix index is broken while retrieving with __htslib tabix__
* In addition, if used with the `FeatureReader` query method, it is also broken.

Because the index created with __htslib tabix__ is working with the `FeatureReader` query method, that means that there is something happening in index creation. This may be very important for some downstream tools like `IndexFeatureFile` from GATK.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


cmnbroad:
@magicDGS I updated #393 with info on why this is happening. Should have a fix fairly soon.

magicDGS:
Thanks for the feedback, @cmnbroad! Feel free to include this code in the test cases for the fix. If not, we can include this PR with the test disabled, and enable it once it is fixed. Just let me know what to do with this PR...

cmnbroad:
Closing in favor of https://github.com/samtools/htsjdk/pull/906.

Issue #701, opened 2016-09-07
Request: Tribble writing support


magicDGS:
### Tribble writing support
Now the tribble support is only regarding reading through implemented `FeatureCodec` and `Feature`, and the only support for writing is for __VCF/BCF__. Implementing an interface and some abstract `FeatureWriter` and `IndexingFeatureWriter` could improve the usage of the codecs implemented in htsjdk or other software (like GATK), for indexing on the fly.

Two possibilities come to my mind:
* Creates a `Encoder<Feature>` interface, similar to the way as the __VCF/BCF__ format is handled.
* Encapsulate the logic for the encoding in the `Codec`. 

The latest will break compatibility, but it makes more sense than separate input/output. The normal behaviour will be that read/write/read a `Feature` and it is easier to maintain in sync a unique codec to take into account the possible changes. In the case of __VCF/BCF__, the codec could contain an instance of the encoder, because the code is more complex. But this classes could be package-specific to avoid usage outside the codecs.

I would like to know what the community thinks about it. Actually, it would be nice for refactoring if #520 is done.

lbergelson:
It's always bothered me that tribble codecs are actually only "decs". I would be a fan of adding a mandatory encode to the codec API along with decode, but we should see how much an impact that would have on users. 

@jrobinson likely has the most custom codecs implemented, so we should get his opinion before making any changes.  It looks like he has his own interfaces for encoding on top of the existing tribble ones.  

I think @cmnbroad may also have opinions about this.  We've been discussing ways to handle tribble headers for generic features, and this might be something that could be worked into any api changes here.

magicDGS:
Hi @lbergelson, I was looking to this issue (work in progress at [magicDGS@`dgs_feature_writers`]( https://github.com/magicDGS/htsjdk/tree/dgs_feature_writers). I did a basic structure for the writers following the `VariantContextWriter` interface and the implementations. The abstract implementation is using the same approach as the `VCFWriter` for high performance and managing both block-compressed and plain text, and a extended class is indexing on the fly.

I tried to do very little change to the codecs, but can you have a look to see if this is going in the good direction?

PR #822, opened 2017-03-10
Tribble writing support

magicDGS:
### Description

As described in #701, it will be nice to have a simple mechanism to write features in tribble not only for `VariantContext`. This is a port from one of my project that I think that a lot of API users will benefit. It includes the following:

* `FeatureWriter` interface to encapsulate adding features and writing headers. There is no support for comments yet, but it could be something added in the future.
* `FeatureEncoder` interface to encapsulate feature-specific output formats. Based on a very nice suggestion from @lindenb in one of my previous PRs, the purpose of this interface is to format the feature/header and add it to an `Appendable`. This allow to write directly the features without storing the `String` format, and give more flexibility. It also have a method to encapsulate the tabix format, to allow indexing on the fly.
* Example of `FeatureEncoder` to convert any `Feature` into a simple BED format (tab-delimited chromosome, start and end). It will be used also in tests.
* Package protected implementation of `FeatureWriter` similar to the VCF writer, which uses an encoder and an output stream to create the feature file. This class can be extended for other purposes as noted bellow.
* Package protected sub-class of `FeatureWriterImpl` to index on the fly. This one will require eventually #810 for use the `java.nio.Path` index method, but it is very useful for writing feature files already indexed.
* Wrapper for `FeatureWriter` to use asynchronous writing. This is a very simple wrapper extending the `AbstractAsyncWriter`. The only flaw is that the header could not be written asynchronously with the current implementation (and thus, it's not safe). I do not really need that, and if it is a problem that it's not safe, this support could be removed.
* Finally, for construction of feature writers, I decided to use a factory pattern instead of the static methods used in the tribble reader classes. This factory have support: 1) index on the fly, 2) block-gzip or gzip, 3) generate MD5 diggest files, 4) asynchronous wriitng and 5) sequence dictionary addition.

All the writing support use `java.nio.Path` to allow other filesystems. There are no tests for the writers yet (only for the simple encoder), because I want to know your opinion about this addition beforehand. I know that this is a very big PR, but I think that it may be useful for others and will be nice to included in HTSJDK for easier development of new bioinformatic tools.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/tribble/bed/BEDSimpleEncoder.java
+++ b/src/main/java/htsjdk/tribble/bed/BEDSimpleEncoder.java
@@ -42,7 +42,7 @@ public class BEDSimpleEncoder implements FeatureEncoder<Feature> {
 
     /**
      * Writes as a BED feature. Note: BED coordinates are 0-based, so the start position returned by
-     * {@link Feature#getStart()} may be different thant the output start position.
+     * {@link Feature#getStart()} may be different than the output start position.
      */
     @Override
     public void write(final Appendable out, final Feature feature) throws IOException {
--- a/src/main/java/htsjdk/tribble/bed/BEDSimpleEncoder.java
+++ b/src/main/java/htsjdk/tribble/bed/BEDSimpleEncoder.java
@@ -1,0 +1,65 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Daniel Gomez-Sanchez
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package htsjdk.tribble.bed;
+
+import htsjdk.tribble.Feature;
+import htsjdk.tribble.writer.FeatureEncoder;
+import htsjdk.tribble.index.tabix.TabixFormat;
+
+import java.io.IOException;
+
+/**
+ * Encode any kind of feature into a simple BED format.
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class BEDSimpleEncoder implements FeatureEncoder<Feature> {
+
+    // default separator for BED files written by this encoder
+    private static final char BED_SEPARATOR = '\t';
+
+    /**
+     * Writes as a BED feature. Note: BED coordinates are 0-based, so the start position returned by
+     * {@link Feature#getStart()} may be different thant the output start position.
+     */
+    @Override
+    public void write(final Appendable out, final Feature feature) throws IOException {
+        out.append(feature.getContig())
+                // The BED format uses a first-base-is-zero convention, Tribble features use 1 => subtract 1.
+                .append(BED_SEPARATOR).append(Integer.toString(feature.getStart() - 1))
+                .append(BED_SEPARATOR).append(Integer.toString(feature.getEnd()));
+    }
+
+    /** Simple BED encoder does not write any header. */
+    @Override
+    public void writeHeader(final Appendable out, final Object header) throws IOException {
+        // does not write anything
+    }
+
+    @Override
+    public TabixFormat getTabixFormat() {
+        return TabixFormat.BED;
+    }
+}
--- a/src/main/java/htsjdk/tribble/writer/FeatureWriter.java
+++ b/src/main/java/htsjdk/tribble/writer/FeatureWriter.java
@@ -1,0 +1,58 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Daniel Gomez-Sanchez
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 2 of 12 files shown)


magicDGS:
Because you were interested in the open issue, can you have a look to this one, @lbergelson? It still needs #810 for simplify some code paths, but I think that this could be cool for others here. I'm already using this code in my own tools, and it works amazingly!

lbergelson:
@magicDGS Sorry I haven't gotten back to you on this.  I think we're very interested in this but I haven't had time to take a proper look.  It's a big update and there's an upcoming deadline for some of my own work that's taking priority.

magicDGS:
No worries @lbergelson. I'm already using this code in another project, but I though that it would be nice to have support for this in HTSJDK. Whenever you have time I'm up for working in your suggestions...

magicDGS:
Friendly ping here @lbergelson!

droazen:
@jonn-smith has volunteered to review this old branch, as he is finding the lack of tribble writing support quite awkward in his current project.

jonn-smith:
@magicDGS Can you rebase this and I'll start taking a look?

magicDGS:
@jonn-smith - I rebased, and I hope that it works with the current implementation of bgzip, which changed a lot after I had the last look to this branch.

Thanks for looking into this, will be really, really helpful!

src/main/java/htsjdk/tribble/writer/FeatureWriterFactory.java
```diff
@@ -0,0 +1,209 @@
...
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * Factory for creating {@link FeatureWriter}.
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public final class FeatureWriterFactory {
```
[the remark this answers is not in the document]
magicDGS: I was working slowly on the tests until someone takes a look to the design, because I didn't want to invest time on something that might change completely the design after review. This class will have tests after the rest of components are properly tested and reviewed.

src/main/java/htsjdk/tribble/writer/IndexingFeatureWriter.java
```diff
@@ -0,0 +1,105 @@
...
+import java.nio.file.Path;
+
+/**
+ * Feature writer class for indexing on the fly.
+ *
+ * @param <F> feature type.
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+final class IndexingFeatureWriter<F extends Feature> extends FeatureWriterImpl<F> {
```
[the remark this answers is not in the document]
magicDGS: In codecov it appears with 87.5% coverage (only sequence dictionary added to the index is  uncovered)

I checked it here: https://codecov.io/gh/samtools/htsjdk/compare/88edbf3e906666f22aa1e759d91208296ab503f9...852c84273543579d9d20225ceb354c254c46144c/diff#diff-c3JjL21haW4vamF2YS9odHNqZGsvdHJpYmJsZS93cml0ZXIvSW5kZXhpbmdGZWF0dXJlV3JpdGVyLmphdmE=

src/main/java/htsjdk/tribble/bed/BEDSimpleEncoder.java
```diff
@@ -0,0 +1,65 @@
...
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class BEDSimpleEncoder implements FeatureEncoder<Feature> {
+
+    // default separator for BED files written by this encoder
+    private static final char BED_SEPARATOR = '\t';
+
+    /**
+     * Writes as a BED feature. Note: BED coordinates are 0-based, so the start position returned by
+     * {@link Feature#getStart()} may be different thant the output start position.
```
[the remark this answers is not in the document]
magicDGS: Done.

src/main/java/htsjdk/tribble/writer/IndexingFeatureWriter.java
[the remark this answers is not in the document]
magicDGS: I'll rebase to check if it is a problem of the diff

yfarjoun:
@jonn-smith are you going to get around to reviewing this?

yfarjoun:
Another friendly ping @jonn-smith. This was rebased as you asked a year ago....

src/main/java/htsjdk/tribble/writer/FeatureWriterImpl.java
```diff
@@ -0,0 +1,113 @@
...
+        writer.close();
+        outputStream.close();
+    }
+
+    /**
+     * Actually write the line buffer contents to the destination output stream. After calling this
+     * function the line buffer is reset so the contents of the buffer can be reused
+     */
+    private void writeAndResetBuffer() throws IOException {
+        writer.flush();
```
yfarjoun: do we really want to flush on every record? seems wasteful

src/main/java/htsjdk/tribble/writer/FeatureWriterImpl.java
```diff
@@ -0,0 +1,113 @@
...
+import java.io.Writer;
+
+/**
+ * Feature writer class, which uses codecs to write out Tribble file formats.
+ *
+ * @param <F> feature type.
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+class FeatureWriterImpl<F extends Feature> implements FeatureWriter<F> {
```
yfarjoun: rename `LineFeatureWriter`

src/main/java/htsjdk/tribble/writer/IndexingFeatureWriter.java
```diff
@@ -0,0 +1,105 @@
...
+        super(encoder, asLocationAwareStream(outputStream));
+        this.indexer = idxCreator;
+        this.indexPath = indexPath;
+        this.refDict = refDict;
+        // this is already a LocationAware
+        this.location = (LocationAware) getOutputStream();
+    }
+
+    /** Wraps if necessary the output stream. */
+    private static OutputStream asLocationAwareStream(final OutputStream outputStream) {
```
yfarjoun: why not return OutputStream

src/main/java/htsjdk/tribble/writer/IndexingFeatureWriter.java
```diff
@@ -0,0 +1,105 @@
...
+     */
+    public IndexingFeatureWriter(final FeatureEncoder<F> encoder, final OutputStream outputStream,
+            final IndexCreator idxCreator, final Path indexPath,
+            final SAMSequenceDictionary refDict) {
+        super(encoder, asLocationAwareStream(outputStream));
+        this.indexer = idxCreator;
+        this.indexPath = indexPath;
+        this.refDict = refDict;
+        // this is already a LocationAware
+        this.location = (LocationAware) getOutputStream();
```
yfarjoun: perhaps override getOutputStream() and do the cast there.

src/main/java/htsjdk/tribble/writer/IndexingFeatureWriter.java
```diff
@@ -0,0 +1,105 @@
...
+        // should be added to the indexer first
+        indexer.addFeature(feature, location.getPosition());
+        super.add(feature);
+    }
+
+    /** Adds the reference dictionary (if not {@code null}), finalize/write the index and close. */
+    @Override
+    public void close() throws IOException {
+        if (refDict != null) {
+            indexer.setIndexSequenceDictionary(refDict);
```
yfarjoun: any reason for this to be here rather than the constructor?

src/main/java/htsjdk/tribble/writer/FeatureWriter.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Writes the header.
+     *
+     * @param header object to be written. May have restrictions depending on the implementation.
+     *
+     * @throws IOException                     if an IO error occurs.
+     * @throws htsjdk.tribble.TribbleException if an encoding error occurs.
+     */
+    public void writeHeader(final Object header) throws IOException;
```
yfarjoun: Should be H type rather than Object

src/main/java/htsjdk/tribble/writer/FeatureWriterFactory.java
```diff
@@ -0,0 +1,211 @@
...
+
+    /** Default approach for create tribble indexes. */
+    public static final IndexFactory.IndexBalanceApproach DEFAULT_INDEX_BALANCE_APPROACH = IndexFactory.IndexBalanceApproach.FOR_SEEK_TIME;
```
yfarjoun: woudl be ince to have a way to inject implementation-specific options. 

for example: compression level for gzip-compressed files. indexing options, specific version of writer if applicable.

src/main/java/htsjdk/tribble/writer/FeatureWriterFactory.java
```diff
@@ -0,0 +1,211 @@
...
+            }
+
+            // if we should not index in the fly
+            if (!createIndex) {
+                return maybeAsync(new FeatureWriterImpl<>(encoder, outputStream));
+            }
+
+            final IndexCreator idxCreator;
+            final Path indexPath;
+            if (blockCompressed) {
```
yfarjoun: would it make sense to have a factory for the index itself?

PR #824, opened 2017-03-14
core function of htsjdk.variant.vcf.VCFEncoder using java.lang.Appendable instead of StringBuilder

lindenb:
### Description

In **htsjdk.variant.vcf.VCFEncoder**  when **streaming** a vcf with a large number of genotype, we avoid to create a large String in memory using StringBuilder, instead we use a java.lang.Appendable

the code is now

```java
	public String encode(final VariantContext context) {
    	try {
            return encode(context,new StringBuilder(1000)).toString();
        } catch (final IOException error) {
           throw new RuntimeIOException("Cannot encode variant", error);
        }
	}

   public Appendable encode(final VariantContext context,final Appendable vcfoutput) throws IOException {	
(...)
```

### Checklist

- [X] Code compiles correctly
- [ ] New tests covering changes and new functionality ( the old 'encode(vc)' function calls new 'encode(vcf,appendable)' )
- [X] All tests passing

### DIFF
--- a/src/main/java/htsjdk/variant/vcf/VCFEncoder.java
+++ b/src/main/java/htsjdk/variant/vcf/VCFEncoder.java
@@ -1,5 +1,6 @@
 package htsjdk.variant.vcf;
 
+import htsjdk.samtools.util.RuntimeIOException;
 import htsjdk.tribble.util.ParsingUtils;
 import htsjdk.variant.variantcontext.Allele;
 import htsjdk.variant.variantcontext.Genotype;
@@ -9,6 +10,7 @@ import htsjdk.variant.variantcontext.LazyGenotypesContext;
 import htsjdk.variant.variantcontext.VariantContext;
 import htsjdk.variant.variantcontext.writer.IntGenotypeFieldAccessors;
 
+import java.io.IOException;
 import java.lang.reflect.Array;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


src/main/java/htsjdk/variant/vcf/VCFEncoder.java
```diff
@@ -64,15 +66,32 @@ public void setAllowMissingFieldsInHeader(final boolean allow) {
...
 
+   /** 
+    * encodes a new Variant context as VCF, writes it to a java.lang.Appendable 
+    * 
+    * @param context the variant
+    * @param vcfoutput the java.lang.Appendable
+    * @return the java.lang.Appendable 'vcfoutput'
+    * @throws IOException
+    */
+   public Appendable encode(final VariantContext context,final Appendable vcfoutput) throws IOException {
```
magicDGS: Can you change the signature to be the other way around? Will be useful for #822

lindenb:
@magicDGS I've changed the signature to `void write(appendable,ctx)`.

> If you can add a similar method for the header,

as far as I can see there is no VCFHeader encoder. The code is in https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java#L147 .

magicDGS:
Ok, I see... I thought that the header was also encoded with the `VCFEncoder`. Thanks for changing the signature and for doing this, it is useful work for me too!

lbergelson:
@lindenb This is now merged as pr #1110.  I did the rebase myself rather than force you through another round of update-review.  Thank you for your contribution!

PR #826, opened 2017-03-16
Obey the tmpDir setting in several constructors that currently ignore it

Lenbok:
### Description

Several constructors don't do what they should regarding tmpDir.

Ensured that both initWriter and initializeBAMWriter both call setTempDirectory() when needed, and made one of the existing constructors make use of initWriter() (This private method had an unused parameter, which has been removed)

We encountered this by running out of disk space while sorting a file during output.

### Checklist

- [X] Code compiles correctly
- [X] All tests passing

### DIFF
--- a/src/test/java/htsjdk/samtools/SAMFileWriterFactoryTest.java
+++ b/src/test/java/htsjdk/samtools/SAMFileWriterFactoryTest.java
@@ -167,7 +167,29 @@ public class SAMFileWriterFactoryTest extends HtsjdkTest {
         fillSmallBam(writer);
         writer.close();
     }
-   
+
+    @Test(description="check that factory settings are propagated to writer")
+    public void testFactorySettings()  throws Exception {
+        final SAMFileWriterFactory factory = new SAMFileWriterFactory();
+        factory.setCreateIndex(false);
+        factory.setCreateMd5File(false);
+        final File wontBeUsed = new File("wontBeUsed.tmp");
+        final int maxRecsInRam = 271828;
+        factory.setMaxRecordsInRam(maxRecsInRam);
+        factory.setTempDirectory(wontBeUsed);
+        final SAMFileHeader header = new SAMFileHeader();
+        header.setSortOrder(SAMFileHeader.SortOrder.coordinate);
+        header.addSequence(new SAMSequenceRecord("chr1", 123));
+        try (final SAMFileWriter writer = factory.makeBAMWriter(header, false, new ByteArrayOutputStream())) {
+            Assert.assertEquals(maxRecsInRam, ((SAMFileWriterImpl) writer).getMaxRecordsInRam());
+            Assert.assertEquals(wontBeUsed, ((SAMFileWriterImpl) writer).getTempDirectory());
+        }
+        try (final SAMFileWriter writer = factory.makeSAMWriter(header, false, new ByteArrayOutputStream())) {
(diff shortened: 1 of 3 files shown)


cmnbroad:
Hi @Lenbok - The fix looks good, but we'd like to have tests to verify these changes. Would you be able to add some getters on the writer for the properties such as tmpdir, maxRecordsInRam, etc., and then verify that they're propagated from the factory to the writer instance for the affected code paths ? Thanks.

Lenbok:
@cmnbroad No problems, will do.

Lenbok:
@cmnbroad Added the accessors and augmented an existing helper method that is used for a couple of SAM and BAM writing tests to check the properties are passed into the writers. Also put a couple of getters on the factory as well.

Lenbok:
@cmnbroad status check?

src/test/java/htsjdk/samtools/SAMFileWriterFactoryTest.java
```diff
@@ -155,6 +155,10 @@ private void createSmallBamToOutputStream(final OutputStream outputStream,boolea
         final SAMFileWriterFactory factory = new SAMFileWriterFactory();
         factory.setCreateIndex(false);
         factory.setCreateMd5File(false);
+        final File wontBeUsed = new File("wontBeUsed.tmp");
+        final int maxRecsInRam = 271828;
+        factory.setMaxRecordsInRam(maxRecsInRam);
+        factory.setTempDirectory(wontBeUsed);
```
droazen: Can you make a separate test case to test this functionality instead of modifying an existing one?

yfarjoun: ( you can copy the old test verbatim and put in your changes )

Lenbok:
@yfarjoun Have pulled out a separate test case and rebased to current master. Should be good to go now.

src/test/java/htsjdk/samtools/SAMFileWriterFactoryTest.java
```diff
@@ -167,7 +167,29 @@ private void createSmallBamToOutputStream(final OutputStream outputStream,boolea
...
+        final SAMFileHeader header = new SAMFileHeader();
+        header.setSortOrder(SAMFileHeader.SortOrder.coordinate);
+        header.addSequence(new SAMSequenceRecord("chr1", 123));
+        try (final SAMFileWriter writer = factory.makeBAMWriter(header, false, new ByteArrayOutputStream())) {
+            Assert.assertEquals(maxRecsInRam, ((SAMFileWriterImpl) writer).getMaxRecordsInRam());
+            Assert.assertEquals(wontBeUsed, ((SAMFileWriterImpl) writer).getTempDirectory());
+        }
+        try (final SAMFileWriter writer = factory.makeSAMWriter(header, false, new ByteArrayOutputStream())) {
```
yfarjoun: is this code duplicated on purpose?

Lenbok: Notice that one is testing SAM, and one is testing BAM. It didn't seem worth pulling out the two duplicated assertions to a separate method.

yfarjoun: ah. OK. sorry. I didn't see that.

yfarjoun:
👍

PR #827, opened 2017-03-16
Allow the use of getSAMString on SAMReadGroupRecord objects

Lenbok:
### Description

Fairly minor, this just allows programmatic access to the string representation of a read group, in much the same way as SAMRecord.getSAMString(). AFAIK there currently isn't really a non-convoluted way to get this. This is useful for displaying to the end user a problematic SAM read group.

### Checklist

- [X] Code compiles correctly
- [X] All tests passing

### DIFF
--- a/src/test/java/htsjdk/samtools/SAMReadGroupRecordTest.java
+++ b/src/test/java/htsjdk/samtools/SAMReadGroupRecordTest.java
@@ -1,0 +1,42 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 The Broad Institute
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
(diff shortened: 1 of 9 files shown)


src/test/java/htsjdk/samtools/SAMReadGroupRecordTest.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009 The Broad Institute
```
nh13: Did you build a time machine, out of a DeLorean? - `2009`

Lenbok: Haha! (Copied an existing test file from the same package as the starting point and the comment was folded in the editor off the bat! I can update if you want)

yfarjoun:
looks good @Lenbok. This is a good idea.  

Could you do the same refactoring for the 3 other classes that inherit from AbstractSAMHeaderRecord and add an abstract getSAMString() function to AbstractSAMHeaderRecord?

lbergelson:
@Lenbok This seems like a good idea and the implementation + test seems good.  It seems like a good idea for all of the other types of header lines as well.  If you don't mind, could you make `toSAMString` into an abstract method on `AbstractSAMHeaderRecord` and do analogous implementations for the other subclasses?

Lenbok:
@yfarjoun, @lbergelson No probs, done. I didn't construct an explicit new test for SAMFileHeader.getSAMString(), as that one is essentially refactored out of the existing clone() method, which already gets called all over the show.

lbergelson:
@Lenbok Thanks you!  👍

Issue #666, opened 2016-07-20
update to a newer snappy version


lbergelson:
HTSJDK currently is using `org.xerial.snappy:snappy-java:1.0.3-rc3`, the most recent version is `1.1.2.6`

Many other projects (spark, rely on newer versions of snappy which have eliminated the `LoadSnappy` class.  If these are required by other libraries they tend to replace the older htsjdk dependency causing htsjdk to break and fail with `NoClassDefinitionFound` if snappy is used.  A workaround is to disable snappy in htsjdk, but this removes any benefit of using snappy in the first place.

It looks like were using this old version of snappy because newer versions statically load themselves and can fail with an Error during static initialization.  

We may need to patch snappy in order to resolve this issue.

See https://github.com/broadinstitute/gatk/issues/2026 for an example of a failure due to loading multiple incompatible snappys.

maarten-k:
I checked the changelog of snappy and there were some speed-ups (although minor):

```
1.0.4 Speeded up the decompressor somewhat; typically about 2–8%
1.0.5 3–10% faster decompression for the base case (x86-64).
1.1.0 Snappy now uses 64 kB block size instead of 32 kB. On average,
    this means it compresses about 3% denser (more so for some
    inputs), at the same or better speeds. 
1.1.1  Speed up decompression by ~2%; much more so (~13-20%) on
    a few benchmarks on given compilers and CPUs.
1.1.4 Improve compression performance by 5%.
Improve decompression performance by 20%.
```

> It looks like were using this old version of snappy because newer versions statically load themselves and can fail with an Error during static initialization.

It looks that org.xerial.snappy:snappy-java is well maintained and I could not reproduce this error(n=1).

I think it is worth the shot to update  org.xerial.snappy:snappy-java  to a recent version. This will also solve some upstream problems.

lbergelson:
@maarten-k I'm definitely in favor of upgrading if we can be relatively sure it won't cause problems.  It's just silly that we're on an ancient beta version instead of a proper release.  

[...]

PR #829, opened 2017-03-21
Update snappy to newer version

maarten-k:
### Description

see issue #666 

please be gentle :This is my first pull request and not a full time  java developer ;) 

### Checklist

- [x] Code compiles correctly
- [ ] All tests passing (tests about testDate failed, but where untouched)

### DIFF
--- a/src/main/java/htsjdk/samtools/util/SnappyLoader.java
+++ b/src/main/java/htsjdk/samtools/util/SnappyLoader.java
@@ -42,9 +42,6 @@ public class SnappyLoader {
     // Force Snappy-java code to be loaded into executable jars.
     private final SnappyInputStream ignoreMe = null;
 
-    // Force bcel to load Snappy.
-    //private static final Class SnappyClass = SnappyInputStream.class;
-
     private static final boolean DefaultVerbosity = Boolean.valueOf(System.getProperty("snappy.loader.verbosity", "false"));
 
     public SnappyLoader() {
--- a/src/main/java/htsjdk/samtools/util/SnappyLoader.java
+++ b/src/main/java/htsjdk/samtools/util/SnappyLoader.java
@@ -24,7 +24,6 @@
 package htsjdk.samtools.util;
 
 import htsjdk.samtools.SAMException;
-import org.xerial.snappy.LoadSnappy;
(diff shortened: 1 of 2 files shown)


lbergelson:
@maarten-k Thanks for the pull request!  Welcome to the project!  Sorry I haven't gotten time to do a proper review yet.  

One note, I think that we should upgrade to a regular release if we're going to do an upgrade instead of a release candidate.  It looks like 1.1.2.6 is the last regular release.  (We can upgrade again easily once 1.1.4 is officially released.  It sounds like they don't have all of the native libraries available yet in the [1.1.4-M3 preview](https://github.com/xerial/snappy-java/blob/master/Milestone.md#snappy-java-114-will-be-available-early-2017).

maarten-k:
Sound like a good idea. I downgraded to the stable version

src/main/java/htsjdk/samtools/util/SnappyLoader.java
```diff
@@ -64,42 +60,13 @@ public SnappyLoader(final boolean verbose) {
...
-            } catch (Error e) {
-                if (e.getClass().equals(snappyErrorClass)) {
-                    if (verbose) System.err.println("Snappy dll failed to load: " + e.getMessage());
-                    tmpSnappyAvailable = false;
-                } else {
-                    throw e;
-                }
-            }
-            SnappyAvailable = tmpSnappyAvailable;
+            SnappyAvailable = true;
```
yfarjoun: codeCov shows that this line is never executed in tests. I suspect that this is not due to lack of tests using gzip. 

The constructor objects are always null and so the if statement is always false...please fix.

tfenne:
I think (as @yfarjoun points out) this PR doesn't work - it deletes much of the code to detect whether snappy is actually available or not.  I've prepared an alternative PR, #872, and would suggest closing this one.

yfarjoun:
Thanks for submitting this PR @maarten-k we are very happy for the contribution. it seems that the best way forward is to focus on @tfenne 's PR #872. So I'll close this one.

maarten-k:
@yfarjoun great to see someone picked this up!

PR #832, opened 2017-03-27
Fix DateParserTest unit tests that were failing in my locale.

Lenbok:
### Description

Several of the DateParserTest unit tests were failing in my locale due to their use of the deprecated Date.parse() method. It seems to me that the existing steps of Date.parse(date.toString()) weren't really accomplishing anything related to htsjdk code (perhaps it was required in past JREs in order to get unix timestamps), so have simplified the assertion to just compare the time stamps directly.

### Checklist

- [X] Code compiles correctly
- [X] All tests passing


droazen:
:+1:

Issue #730, opened 2016-10-26
VCFHeader drops sequence dictionary attributes


cmnbroad:
Both:

- VCFHeader.setSequenceDictionary
- the VCFContigHeaderLine constructor VCFContigHeaderLine(final SAMSequenceRecord sequenceRecord, final String assembly)

drop all SAMSequenceRecord attributes except ID, length, and assembly (i.e. they drop MD5/UR).  If these are allowed in the vcf (AFAICT they are) then both of these classes should be fixed up to propagate these in and out.

Issue #575, opened 2016-04-18
VCF4.3 constants and header lines


magicDGS:
### VCF4.3 support
New specifications for [VCF 4.3](http://samtools.github.io/hts-specs/VCFv4.3.pdf) suggest that a new `VCFCompoundHeaderLine` could be implemented for symbolic alternative alleles (`VCFAltHeaderLine`).

New reserved keys are included in the specs, which could be added to `VCFConstants` in conjuction with reserved structural variant alleles (DEL, INS, etc.). In addition, other headers that can be supported are "##SAMPLE", "##META" and "##PEDIGREE".

I can work on it if there is interest on this issue.

droazen:
@magicDGS I think we'd be happy to accept any patch that improves htsjdk support for VCF4.3, and the changes you mention sound good to me.

magicDGS:
I will work on this, but I found some related issues (#517, #500, #389, #277, #251, #250). Should I take into account that too, @droazen?

droazen:
You're welcome to take any or all of those tickets on if you like, @magicDGS, but I recommend making each individual branch/pull request as small and specific as possible, with good tests for each one. With htsjdk we have to be extra-careful about what we merge, and large monolithic branches with many features tend to take months to get in.

tfenne:
Implemented by #1359 (VCF v4.3 read support). VCFHeaderVersion.VCF4_3 exists and ALT/SAMPLE/META/PEDIGREE header lines are now modeled with dedicated VCFSimpleHeaderLine-derived classes (VCFAltHeaderLine, VCFSampleHeaderLine, VCFMetaHeaderLine, VCFPedigreeHeaderLine). The specific constants/header-line modeling this ticket asked for are done; the broader 'full VCF4.3 + BCF2.2' effort is tracked separately in #596/#628. Closing this one as resolved.

Issue #517, opened 2016-03-08
VCF4Parser.parseLine throws exception on optional Source and Version tags in INFO header


heuermh:
Given VCFv4.2 header with Source and Version tags in `##INFO` header line

```
##fileformat=VCFv4.2
...
##INFO=<ID=ENST,Number=.,Type=String,Description="Ensembl transcript identifier",Source="savant",Version="1.1.0">
```

we're getting an exception thrown from htsjdk via HadoopBam

```
Caused by: htsjdk.tribble.TribbleException$InvalidHeader: Your input file has a malformed header: unexpected tag count 6 in line <ID=ENST,Number=.,Type=String,Description="Ensembl transcript identifier",Source="savant",Version="1.1.0">
    at htsjdk.variant.vcf.VCF4Parser.parseLine(VCFHeaderLineTranslator.java:109)
    at htsjdk.variant.vcf.VCFHeaderLineTranslator.parseLine(VCFHeaderLineTranslator.java:51)
    at htsjdk.variant.vcf.VCFCompoundHeaderLine.<init>(VCFCompoundHeaderLine.java:164)
    at htsjdk.variant.vcf.VCFInfoHeaderLine.<init>(VCFInfoHeaderLine.java:48)
    at htsjdk.variant.vcf.AbstractVCFCodec.parseHeaderFromLines(AbstractVCFCodec.java:199)
    at htsjdk.variant.vcf.VCFCodec.readActualHeader(VCFCodec.java:111)
    at htsjdk.tribble.AsciiFeatureCodec.readHeader(AsciiFeatureCodec.java:88)
    at org.seqdoop.hadoop_bam.VCFRecordReader.initialize(VCFRecordReader.java:87)
    at org.seqdoop.hadoop_bam.VCFInputFormat.createRecordReader(VCFInputFormat.java:189)
    at org.apache.spark.rdd.NewHadoopRDD$$anon$1.<init>(NewHadoopRDD.scala:149)
    at org.apache.spark.rdd.NewHadoopRDD.compute(NewHadoopRDD.scala:122)
    at org.apache.spark.rdd.NewHadoopRDD.compute(NewHadoopRDD.scala:63)
```

The same stack trace can be replicated by Picard

``` bash
$ Picard SortVcf INPUT=info.v4.2.vcf OUTPUT=info.v4.2.sorted.vcf
[...]

Issue #500, opened 2016-02-23
VCF header lines get dropped


bhandsaker:
### Verify
Can you see anything in the logs? 
Make sure your issue is not already in the [htsjdk issue tracker](https://github.com/samtools/htsjdk/issues?q=)

### Subject of the issue
In htsjdk 2.1.0, the VCFHeader class will silently drop header lines with the same key but distinct IDs for all keys except INFO, FORMAT, FILTER and contig.  For example, ALT tags (other than the first one) get dropped, as do SAMPLE, META or any other VCF 4.3 "structured" tags.

### Your environment
* htsjdk 2.1.0
* java 1.8
* any OS

### Steps to reproduce
Try to add two
`##ALT=<ID=X,Description="x"> 
##ALT=<ID=Y,Description="y">`
header lines to a VCFHeader and write out the VCF file. One of the ALT tags will get dropped.

### Expected behaviour
Per the VCF spec (clarified in v4.3) structured tags must have unique ID fields within each key field (e.g. "ALT" in the example above).

### Actual behaviour
The ALT tags are overwriting each other.
These are currently being stored in a hash map keyed by the key field (e.g. ALT), but they either need to be put in a Map<String, Map<String, VCFHeaderLine>> (first key is "key", second key is ID) or not tracked. There is a method shouldBeAddedToDictionary() in VCFHeaderLine but the intent of this method is unclear to me and VCFHeader makes a dictionary by key regardless.

bartcharbon:
We are running into this issue as well. Do you have an idea if/when this might be fixed?

PR #835, opened 2017-03-29
VCFHeader, VCFCodec and VCFHeaderLine refactoring for VCF4.3/BCF2.2

cmnbroad:
This is the first of 3 PRs for adding support for VCF4.3 and BCF 2.2; the first PR is mostly a refactoring of VCFHeader and the VCFHeaderLine class hierarchy in order to fix a bunch of bugs, and to provide the foundation for supporting new VCF versions.

There are a number of TODO comments embedded in these changes that are there either to:

- underscore differences from prior behavior
- ask a question about how some case should be handled
- provide an explanation to help the reviewer understand the changes
- note other changes that need to be fixed independent of this PR.

Fixes:

- https://github.com/samtools/htsjdk/issues/250 VCFHeader class design is unsafe
- https://github.com/samtools/htsjdk/issues/730 VCFHeader drops sequence dictionary attributes
- https://github.com/samtools/htsjdk/issues/251 VCFHeader should not strip fileformat/version lines
- https://github.com/samtools/htsjdk/issues/277 VCFHeader should allow multiple other metadata lines (with the same key)
- https://github.com/samtools/htsjdk/issues/500 VCFHeader lines dropped by addMetaDataLine()
- https://github.com/samtools/htsjdk/issues/575 (partial fix) VCF 4.3 constants and header lines (ALT, META, SAMPLE, PEDIGREE)
- https://github.com/samtools/htsjdk/issues/517 VCF4Parser.parse throws on Source and Version tags in INFO lines
- https://github.com/samtools/htsjdk/issues/389 VCFContigHeaderLine throws if no length tag is present

Changes to VCFHeader:

- Always maintains a VCFHeaderVersion, and validates any header lines against the validation rules for that version.
- Delegates management of the header line list to the VCFMetaDataLines class, which uses a single LinkedHashMap of header lines. The VCFMetaDataLines class is also used to merge headers.
- No longer removes/restores file format lines from the header line list.

Changes to the VCFHeaderLine hierarchy:

- All structured header lines classes maintain and round-trips all tags
- All header lines are now immutable
- All VCFHeaderLine classes have a validateForVersion method that validates the line against a particular version (this is called by VCFHeader when the line is added to a header, or when a header version is changed)
- Replaced VCFSimpleHeaderLine/VCFIDHeaderLine with VCFStructuredHeaderLine, which is the base class for all “ID” header lines. Its no longer necessary to use “instanceof” tests to determine header line type. The VCFHeaderLine base class now has an “isStructuredHeaderLine” and a “getID” method; getID is valid iff isStructuredHeaderLine returns true.

VcfUtils.smartMergeHeaders delegates to VCFHeader, which in turn sometimes delegates to VCFHeaderLine classes to resolve conflicts when merging headers

Non-backward compatible API changes:

- Removed VCFSimpleHeaderLine class, VCFIDHeaderLine interface
- Removed the SupportedLineType enum from VCFCompoundHeaderLine
- Removed public sameLineTypeAndName and equalExcludingDescription methods from VCFCompoundHeaderLine
- Removed the VCFContigHeaderLine constructor that allowed the caller to specify a key
- Also deprecated VCFUtils.makeContigHeaderLine

Notes: These changes have been run against the Picard and GATK test suites. All of the tests pass, though some require minor changes; these are mostly tests that do line-by-line comparisons against expected results files that have duplicate ID header lines. These now fail because the new implementation no longer retains/roundtrips the duplicate lines. One tool (LiftOverVcf) requires a minor code change because the version validation framework when it attempts to set a header version on a header that contains a conflicting file format line (which would previously have been stripped out by VCFHeader).

Outstanding question: There are several places where code nw either logs or throws on input that is not conformant for the version; was previously silently dropped; was previously silently retained when it shouldn't have been, or has correct behavior that differs from the previous implementation. For now, the throwing/logging is based on “strict reporting” mode setting that is controlled by a global variable. We need to determine the right level and mechanism for opt-out/opt-in should be for these cases.


cmnbroad:
This branch also fixes https://github.com/samtools/htsjdk/issues/934.

src/main/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -98,135 +92,261 @@
...
+        if (!newVCFVersion.equals(vcfVersion)) {
+            logger.warn(String.format("Changing VCFHeader version from %s to %s",
+                    vcfVersion.getVersionString(),
+                    newVCFVersion.getVersionString()));
+            // TODO: This can cause failures in code that used to succeed (i.e. Picard LiftOverVcf tests fail
+            // if they're not modified to remove the embedded version line) since we now retain ##fileformat
+            // lines in the metadata list; the validation code recognizes and validates the embedded ##fileformat
+            // lines against the new version, and throws if they conflict.
+            //
+            // We might want to add a removeHeaderLine method to VCFHeader so that consumers with this problem
```
magicDGS: +1 for adding that method

src/main/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -98,135 +92,261 @@
...
+
+    /**
+     * Adds a new line to the VCFHeader. If there is an existing header line of the
+     * same type with the same key, and the header version is pre-4.3, the new line is added
+     * using a modified key to make it unique for BWC; otherwise (in strict validation mode), an
+     * exception is thrown since duplicates are not allowed.
+     *
+     * @param headerLine header line to attempt to add
+     */
+    public void addMetaDataLine(final VCFHeaderLine headerLine) {
```
magicDGS: Maybe a method to override a meta data line will be useful, to being able to remove the previous associated key. It can return the previous line to check if it was present...

src/main/java/htsjdk/variant/vcf/VCFHeaderLineCount.java
```diff
@@ -25,9 +25,78 @@
 
 package htsjdk.variant.vcf;
 
+import htsjdk.utils.Utils;
+
 /**
  * the count encodings we use for fields in VCF header lines
  */
+//TODO: this should really be called VCHFHeaderCountType
```
magicDGS: Because this PR is introducing breaking changes anyway, I vote for change the name here too.

src/main/java/htsjdk/variant/vcf/VCFHeaderLineCount.java
```diff
@@ -25,9 +25,78 @@
...
 public enum VCFHeaderLineCount {
     INTEGER, A, R, G, UNBOUNDED;
+
+    // A default int value used to represent an integral count value (not a count *type*) when the
+    // actual count is derived and not a fixed integer (i.e., when isFixedCount()==false)
```
magicDGS: I will make this a javadoc and not a comment, because it is public.

src/main/java/htsjdk/variant/vcf/VCFMetaDataLines.java
```diff
@@ -0,0 +1,298 @@
+package htsjdk.variant.vcf;
+
+import htsjdk.samtools.util.Log;
+import htsjdk.tribble.TribbleException;
+
+import java.io.Serializable;
+import java.util.*;
+import java.util.stream.Collectors;
+
+//TODO: should this class be public so consumers can use in place of Set<VCFHeaderLine>
```
magicDGS: Is any other change planned for the next PRs? If not, maybe it is worthy to make it public in this PR to make people familiar with the API from the beginning...

src/main/java/htsjdk/variant/vcf/VCFMetaDataLines.java
```diff
@@ -0,0 +1,298 @@
...
+ */
+class VCFMetaDataLines implements Serializable {
+    public static final long serialVersionUID = 1L;
+    protected final static Log logger = Log.getInstance(VCFMetaDataLines.class);
+
+    // Require a unique key for each ID line
+    // Allow duplicate non-ID keys, unless fileformat, reference ?
+
+    // TODO: Should we reject attempts to add two contig header lines with the same contigIndex ?
+    // TODO: GATK VcfUtilsUnitTest.createHeaderLines test creates headers with contig lines with identical (0) indices
```
magicDGS: Again, I think that it is not a good idea to block htsjdk development for failing tests in downstream projects...

src/main/java/htsjdk/variant/vcf/VCFMetaDataLines.java
```diff
@@ -0,0 +1,298 @@
...
+                // TODO: them in the master header line list maintained by VCFHeader, but would silently drop them
+                // TODO: from the typed header line lists, so the duplicates would not surface in queries (i.e. via
+                // TODO: getInfoHeaderLines()).
+                //
+                // TODO: This implementation doesn't allow the duplicates to be preserved (the 4.2/4.3 specs expressly
+                // TODO: forbid it), so they're dropped here. There are several Picard and GATK test files around with
+                // TODO: duplicate ID lines, which will have to be modified in order for the tests to pass.
+                //
+                // TODO: Is it sufficient to log a warning for this case (duplicate key) ? Should we add code
+                // TODO: here to throw only for v4.3+ ? Only if VCFUtils.getStrictVCFVersionValidation is set ? Or both ?
```
magicDGS: After reading through the whole PR, I start to think that VCF requires a `ValidationStringency` enum similar to SAM/BAM parsing to use in different parts, with STRICT value by default (maybe SILENT until the 3 PRs are in to keep the actual behaviour).

src/main/java/htsjdk/variant/vcf/VCFMetaDataLines.java
```diff
@@ -0,0 +1,298 @@
...
+     * @param id the id of the requested header line
+     * @return the meta data line, or null if there is none
+     */
+    public VCFFilterHeaderLine getFilterHeaderLine(final String id) {
+        return (VCFFilterHeaderLine) mMetaData.get(makeKey(VCFConstants.FILTER_HEADER_KEY, id));
+    }
+
+    //TODO: Is this useful ? It returns the first match for the given key, no matter how many
+    //TODO: there are. Should we deprecate it (and add a new one that returns a collection)
+    //TODO: or just change this one to return a collection ?
```
magicDGS: I vote for a breaking change to return a Collection, because anyway this PR will introduce lots of backwards incompatible changes...

src/main/java/htsjdk/variant/bcf2/BCF2Codec.java
```diff
@@ -150,6 +150,10 @@ public FeatureCodecHeader readHeader( final PositionalBufferedStream inputStream
 
             if ( bcfVersion.getMajorVersion() != ALLOWED_MAJOR_VERSION )
                 error("BCF2Codec can only process BCF2 files, this file has major version " + bcfVersion.getMajorVersion());
+
+            // TODO: fixing this breaks GATK GenomicsDB integration/tests
```
cmnbroad: Agreed. It needs to be fixed, but that will go into the next PR. This is one several places in this PR where I added notes like this for things like this.

src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
```diff
@@ -141,7 +146,7 @@ public void writeHeader(final VCFHeader header) {
     }
 
     public static String getVersionLine() {
-        return VERSION_LINE;
+        return DEFAULT_VERSION_LINE;
     }
 
     public static VCFHeader writeHeader(VCFHeader header,
```
cmnbroad: There are already a lot of changes in this PR, so I'd prefer to keep this focused for now.
(I like the the general idea of #822, but that one also will need some discussion).

src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
```diff
@@ -195,6 +201,32 @@ public static VCFHeader writeHeader(VCFHeader header,
...
+            final String message = String.format("Attempt to write a version %s VCF header to a version %s VCF output",
+                    vcfVersion.getVersionString(),
+                    versionLine);
+            if (VCFHeaderVersion.versionsAreCompatible(VCFHeaderVersion.getHeaderVersion(versionLine), vcfVersion)) {
+                if (VCFUtils.getStrictVCFVersionValidation()) {
+                    throw new TribbleException(message);
+                }
+            }
+            if (VCFUtils.getVerboseVCFLogging()) {
+                logger.warn(message);
```
cmnbroad: Not quite sure I get what you're saying. Are you saying to only log this message if the versions are incompatible ?

src/main/java/htsjdk/variant/vcf/VCF3Codec.java
```diff
@@ -129,8 +108,20 @@ else if (line.startsWith(VCFHeader.HEADER_INDICATOR)) {
...
+    protected void reportDuplicateFilterIDs(final String duplicateFilterMessage) {
+        // no-op since this codec's parseFilters method doesn't check for them
     }
+
+    /**
+     * Handle report of duplicate info field values
+     * @param key
+     * @param infoLine
+     */
+    public void reportDuplicateInfoKeyValue(final String key, final String infoLine) {}
```
cmnbroad: Yes, there should be a comment here. I made this a no-op because that was the previous behavior.

src/main/java/htsjdk/variant/vcf/VCFAltHeaderLine.java
```diff
@@ -0,0 +1,75 @@
+package htsjdk.variant.vcf;
+
+import htsjdk.samtools.util.Log;
+import htsjdk.tribble.TribbleException;
+
+import java.util.*;
+
+//TODO: Should we validate these alt allele types ?
```
cmnbroad: I agree that sounds right, but it could break backward compatibility. So perhaps we should only do it starting with v4.3.

src/main/java/htsjdk/variant/vcf/VCFCodec.java
```diff
@@ -72,91 +65,68 @@
...
-            fFields.addAll(Arrays.asList(filterString.split(VCFConstants.FILTER_CODE_SEPARATOR)));
+    protected void reportDuplicateFilterIDs(final String duplicateFilterMessage) {
+        // older versions of htsjdk have been silently dropping these for a while, but we can at least warn
+        if (VCFUtils.getVerboseVCFLogging()) {
+            logger.warn(duplicateFilterMessage);
+        }
+    }
 
-        filterHash.put(filterString, Collections.unmodifiableList(fFields));
+    public void reportDuplicateInfoKeyValue(final String key, final String infoLine) {}
```
cmnbroad: Excellent question. Historically, htsjdk reads multiple VCF versions, but always writes only the current version. That always worked because we could always coerce old versions to the current version. But not all existing VCFs can be auto-converted to V4.3 in a non-destructive way since 4.3 is somewhat more restrictive. So we have to make a decision about to integrate v4.3. i.e., do we require opt-in to 4.3 ? IIRC, this method is a no-op based on the idea that we would implement a new VCF43Codec that WOULD report these, but thats part of the philosophical discussion that we need to resolve for this PR. I think there multiple places in this PR that will be affected by how we resolve that issue.

src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
```diff
@@ -101,153 +257,79 @@ public int getCount(final VariantContext vc) {
...
-        if (type == VCFHeaderLineType.Flag && !allowFlagValues())
-            throw new IllegalArgumentException("Flag is an unsupported type for this kind of field");
+    public static VCFCompoundHeaderLine getSmartMergedCompoundHeaderLine(
+            final VCFCompoundHeaderLine line1,
+            final VCFCompoundHeaderLine line2,
+            final VCFHeader.HeaderConflictWarner conflictWarner,
+            BiFunction<VCFCompoundHeaderLine, VCFCompoundHeaderLine, VCFCompoundHeaderLine> compoundHeaderLineResolver)
+    {
+        Utils. nonNull(line1);
+        Utils. nonNull(line2);
```
cmnbroad: Oh yeah.

src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
```diff
@@ -101,153 +257,79 @@ public int getCount(final VariantContext vc) {
...
-                number = VCFConstants.PER_GENOTYPE_COUNT;
-                break;
-            case UNBOUNDED:
-                number = VCFConstants.UNBOUNDED_ENCODING_v4;
-                break;
-            case INTEGER:
-            default:
-                number = count;
+        if (!line1.getDescription().equals(line2.getDescription())) {
+            conflictWarner.warn(line1, "Allowing unequal description fields through: keeping " + line2 + " excluding " + line1);
```
cmnbroad: Yeah, not sure what the right thing is here. IIRC, both lines are retained in the current implementation, but that results in duplicate IDs. And there are a lot of files with duplicates like this around in both Picard and GATK.

src/main/java/htsjdk/variant/vcf/VCFHeaderLineCount.java
cmnbroad: Yes, good point.

src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
magicDGS: I mean that the message could include information about the compatibility...

src/main/java/htsjdk/variant/vcf/VCFCodec.java
magicDGS: Thanks for the explanation, @cmnbroad! Maybe a comment about this in the method will be good for discussion and knowing what is happening...

src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
magicDGS: I prefer not to keep duplicate IDs either, but it should be a way to choose which line description is preferred and/or merge both descriptions.

It is true that the sencond option can end in a concatenation of similar descriptions with period/no-period at the end of the line. But I believe that it is better than using one "randomly picked" description if they are not the same at all, leading to the same ID referring to different things depending on the variant/genotype/etc.

For example, An user looking at a concrete ID description as following  "x-statistic OR y-statistic" will catch that there is different information encoded in the same ID; otherwise, they will think that the value represents only the x-statistic.

yfarjoun:
Closing due complexity. 

author will work with htsjdk-next-beta to include these great features.

PR #837, opened 2017-03-30
new Interface VCFIterator extends CloseableIterator<VariantContext>

lindenb:
### Description

As far as I know, there is no existing solution to handle a VCF file from an InputStream like System.in.
Here I introduce :
+ VCFIterator an `Iterator<VariantContext>` with the method `VCFheader getFileHeader()`
+ VCFIteratorBuilder a simple builder for the interface above (builds from uri,file and InputStream ). Decodes BCF,VCF, VCF.gz.

```
  VCFIterator r = new VCFIteratorBuilder().open(System.in);
  while (r.hasNext()) {
     System.out.println(r.next());
  }
  r.close();
```
```

### Checklist

- [X] Code compiles correctly
- [X] All tests passing

### CODE REVIEW

File: src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java [hunk 1]
```diff
@@ -0,0 +1,210 @@
...
+    /**
+     * creates a VCF iterator from an URI It detects if the stream is a BCF
+     * stream or a GZipped stream.
+     * 
+     * @param uri
+     *            the URI
+     * @return the VCFIterator
+     * @throws OException
+     */
+    public VCFIterator open(final String uri) throws IOException {
```
magicDGS: I think that this method should use `ParsingUtils.openInputStream()` instead, to deal with `java.nio.Path`, `URL` and `File`.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,210 @@
...
+
+    /**
+     * creates a VCF iterator from a File
+     * 
+     * @param file
+     *            the file (can be bcf, vcf, vcf.gz)
+     * @return the VCFIterator
+     * @throws IOException
+     */
+    public VCFIterator open(final File file) throws IOException {
```
magicDGS: I would love to have a `java.nio.Path` method too!

src/test/java/htsjdk/variant/vcf/VCFIteratorTest.java
```diff
@@ -0,0 +1,76 @@
...
+    public void testUsingFile(String uri, int nVariants) throws IOException {
+        VCFIterator r = new VCFIteratorBuilder().open(new File(uri));
+        Assert.assertNotNull(r.getFileHeader());
+        Assert.assertEquals(countVariants(r), nVariants);
+        r.close();
+    }
+
+    @Test(dataProvider = "VcfFiles")
+    public void testUsingStreams(String uri, int nVariants) throws IOException {
+        VCFIterator r = new VCFIteratorBuilder().open(uri);
```
yfarjoun: how is this test different from testUsingUri?

lindenb: @yfarjoun ah you're right; I've changed this with a FileInputStream

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,210 @@
...
+     * @param in
+     *            inputstream
+     * @return the VCFIterator
+     * @throws RuntimeIOException
+     */
+    @SuppressWarnings("resource")
+    public VCFIterator open(InputStream in) throws IOException {
+        if (in == null) {
+            throw new IllegalArgumentException("input stream is null");
+            }
```
yfarjoun: remove spaces before `}`

src/test/java/htsjdk/variant/vcf/VCFIteratorTest.java
```diff
@@ -0,0 +1,80 @@
...
+    @Test(dataProvider = "VcfFiles")
+    public void testUsingStreams(String file, int nVariants) throws IOException {
+        InputStream in = new FileInputStream(file); 
+        VCFIterator r = new VCFIteratorBuilder().open(in);
+        Assert.assertNotNull(r.getFileHeader());
+        Assert.assertEquals(countVariants(r), nVariants);
+        r.close();
+        in.close();
+    }
+
```
yfarjoun: extra nl

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,210 @@
...
+                    null :
+                    this.codec.decode(this.inputStream);
+        }
+
+        @Override
+        public void close() {
+            this.inputStream.close();
+        }
+    }
+
```
yfarjoun: extra nl

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,210 @@
...
+     * @return the VCFIterator
+     * @throws IOException
+     */
+    public VCFIterator open(final File file) throws IOException {
+        IOUtil.assertFileIsReadable(file);
+        return open(new FileInputStream(file));
+    }
+
+    /** implementation of VCFIterator, reading VCF */
+    private static class VCFReaderIterator extends AbstractIterator<VariantContext>implements VCFIterator {
```
yfarjoun: space before `implements`

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,210 @@
...
+            throw new IllegalArgumentException("input stream is null");
+            }
+        // try to read the BCF header
+        final BufferedInputStream bufferedinput = new BufferedInputStream(in, 100);
+        bufferedinput.mark(100);
+        final BCFVersion bcfVersion = BCFVersion.readBCFVersion(bufferedinput);
+        bufferedinput.reset();
+
+        //this is VCF or VCF.gz
+        if (bcfVersion == null) {
```
yfarjoun: since you just tried reading the bcf header, coudl you flip the if statement and open the bcf if ! null?

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,210 @@
...
+            // try to read GZIP header
+            // see http://stackoverflow.com/questions/4818468/how-to-check-if-inputstream-is-gzipped
+            final byte[] signature = new byte[2];
```
yfarjoun: Please refactor this useful piece of code to a function. Perhaps a static public function in BlockCompressedInputStream or IOUtil.

yfarjoun: oh, and test it _please_.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
lindenb: @yfarjoun back to you:

in Ioutils, I've added:
```

public static boolean isGZIPInputStream(BufferedInputStream in);
public static InputStream mayBeGZippedInputStream(InputStream in);
```

and I've added some tests for those two functions in IoUtilTest.

magicDGS:
I still think that `java.nio.Path` support is a good adition to this PR, and it will be fairly simple. If not, I could open a PR after this one to add that support.

lindenb:
> , I could open a PR after this one to add that support.

@magicDGS yes please, I'm not familiar enough whith java.nio

magicDGS:
Perfect @lindenb. Could you please add a ticket after this is accepted (and maybe @yfarjoun could assign it to me) for that purpose? Just for tracking the work that should be done... Thanks in advance!

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -988,4 +988,45 @@ public static Path getPath(String uriString) throws IOException {
...
+     * @return a decompressed input stream
+     * @throws IOException
+     */
+    public static InputStream mayBeGZippedInputStream(final InputStream in) throws IOException {
+        if(in == null) throw new IllegalArgumentException("mayBeGZippedInputStream(null)");
+        /* BufferedInputStream is needed to decode the first bytes of the stream */
+        final BufferedInputStream bufferedinput;
+        /* it's already a buffered input stream, no need to wrap it again */
+        if(in instanceof BufferedInputStream) {
+            bufferedinput = BufferedInputStream.class.cast(in);
```
yfarjoun: I think I'd avoid the premature optimization of trying to figure out if it's buffered "enough" and just wrap it in a buffer to be sure.

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -988,4 +988,45 @@ public static Path getPath(String uriString) throws IOException {
...
+        } else {
+            bufferedinput = new BufferedInputStream(in, 4);
+        }
+        if(isGZIPInputStream(bufferedinput)) {
```
yfarjoun: space after if (and before parenthesis)

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,200 @@
...
+     * @param uri
+     *            the URI
+     * @return the VCFIterator
+     * @throws OException
+     */
+    public VCFIterator open(final String uri) throws IOException {
+        if (IOUtil.isUrl(uri)) {
+            try {
+                final URL url = new URL(uri); // this will throw if its not a
+                                              // url
```
yfarjoun: no need for a single word in a newline

src/test/java/htsjdk/samtools/util/IoUtilTest.java
```diff
@@ -187,4 +191,54 @@ public void testFileTypeUnix(final String path, boolean expectedIsRegularFile) {
                 {"/non/existent/file", Boolean.TRUE},
         };
     }
+    
+    @Test
+    public void isGZIPInputStreamTest() throws IOException {
```
yfarjoun: please use a dataprovider here too instead of testing two different scenarios in a single test.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,200 @@
...
+     * creates a VCF iterator from an URI It detects if the stream is a BCF
+     * stream or a GZipped stream.
+     * 
+     * @param uri
+     *            the URI
+     * @return the VCFIterator
+     * @throws OException
+     */
+    public VCFIterator open(final String uri) throws IOException {
+        if (IOUtil.isUrl(uri)) {
```
magicDGS: Sorry, althought I said that I can add the `java.nio.Path` support, I realized that my change will change the behaviour of this method. To still support urls and files, but with the addition of `java.nio.Path` comming from a `String`, I think that here the best approach will be:

```java
 /**
  * creates a VCF iterator from an Path.
  * It detects if the stream is a BCF stream or a GZipped stream.
  * 
  * @param path a uri like string
  * @return the VCFIterator
  * @throws OException
  */
public VCFIterator open(final String path) throws IOException {
    return open(ParsingUtils.openInputStream(path, null));
}
```

This will simplify the code and do not change the behaviour in a later PR.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
magicDGS: For the `java.nio.Path` method, if the previous change is made, the simpler option will be:

```java
public VCFIterator open(final Path path) throws IOException {
    return open(new SeekablePathStream(path, null));
}
```

src/main/java/htsjdk/samtools/util/IOUtil.java
magicDGS: I suppport @yfarjoun comment, and I will suggest to use the original input stream for the return, and use only the buffered one for the determination of the gzip signature...

yfarjoun:
@magicDGS, that idea (wrapping and returning the unwrapped stream) seems "dangerous": If the underlying stream cannot reset back to the starting position (buffer is too small) it will not give you the beginning of the file anymore. Barring a way to detect how many byte can be read without irreversibly consuming data from the stream, you have to wrap it yourself.

Perhaps what we need is an override boolean (to mayBeGZIP) because the main use-case will be in the VCFIterator constructor and there we know that we have enough buffer that we do not want to re-wrap with another BufferedStream.

src/main/java/htsjdk/samtools/util/IOUtil.java
lindenb: @magicDGS I'm currently looking at this. Unless I'm misunderstanding there is no way to return the original InputStream because you need to consume the first bytes of the stream.

lindenb: @magicDGS ah.. sorry for the duplicate: what @yfarjoun said below.

lindenb:
@magicDGS @yfarjoun @lbergelson thank you all for the reviews. Back to you. I've added @magicDGS 's suggestions for Path. (while I was here , I've closed some opened Streams in IoUtils.assertFilesEqual )

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -1009,21 +1011,15 @@ public static boolean isGZIPInputStream(final BufferedInputStream bufferedinput)
...
         /* BufferedInputStream is needed to decode the first bytes of the stream */
-        final BufferedInputStream bufferedinput;
-        /* it's already a buffered input stream, no need to wrap it again */
-        if(in instanceof BufferedInputStream) {
-            bufferedinput = BufferedInputStream.class.cast(in);
-        } else {
-            bufferedinput = new BufferedInputStream(in, 4);
-        }
-        if(isGZIPInputStream(bufferedinput)) {
+        final BufferedInputStream bufferedinput = new BufferedInputStream(in, 4);
```
yfarjoun: I'm annoyed that we will have this extra layer around all our streamed inputs....

how about making a protected wrapper class around BufferedInputStream that is only used in `VCFIteratorBuilder` and then you can have a different signature function that doesn't wrap the Stream with another buffered Stream. Since this will be the main use-case, we will generally not have the extra layer.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,200 @@
...
+import htsjdk.tribble.readers.AsciiLineReader;
+import htsjdk.tribble.readers.AsciiLineReaderIterator;
+import htsjdk.tribble.readers.PositionalBufferedStream;
+import htsjdk.tribble.util.ParsingUtils;
+import htsjdk.variant.bcf2.BCF2Codec;
+import htsjdk.variant.bcf2.BCFVersion;
+import htsjdk.variant.variantcontext.VariantContext;
+
+/**
+ * An Class building {@link VCFReaderIterator}
```
ronlevine: Change `An` to `A`.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,200 @@
...
+     * @return the VCFIterator
+     * @throws RuntimeIOException
+     */
+    public VCFIterator open(final InputStream in) throws IOException {
+        if (in == null) {
+            throw new IllegalArgumentException("input stream is null");
+            }
+        // try to read the BCF header
+        final BufferedInputStream bufferedinput = new BufferedInputStream(in, 100);
+        bufferedinput.mark(100);
```
ronlevine: Can you use a static constant i.e. private static BCF_VERFSION_SIZE = 100 instead of a magic number? Can it be a smaller value?

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -988,4 +991,38 @@ public static Path getPath(String uriString) throws IOException {
...
+    
+    /**
+     * If `in` is a GZipped stream, wrap it into a {@link GZIPInputStream}.
+     * @param in the input stream
+     * @return a decompressed input stream
+     * @throws IOException
+     */
+    public static InputStream gunZipIfNeeded(final InputStream in) throws IOException {
+        /* BufferedInputStream is needed to decode the first bytes of the stream */
+        final BufferedInputStream bufferedinput = new BufferedInputStream(in, 4);
```
ronlevine: Why is this magic number 4 bytes? Can it be changed to a named static variable?

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -988,4 +991,38 @@ public static Path getPath(String uriString) throws IOException {
...
+    
+    /**
+     * Test whether a input stream looks like a GZIP input.
+     * @param bufferedinput the input stream. The buffer must be large enough to contain the GZIP signature
+     * @return true if `bufferedinput` starts with a gzip signature 
+     * @throws IOException
+     */
+    public static boolean isGZIPInputStream(final BufferedInputStream bufferedinput) throws IOException {
+        // see http://stackoverflow.com/questions/4818468/how-to-check-if-inputstream-is-gzipped
+        final byte[] signature = new byte[2];
```
ronlevine: Can 2 be changed to static variable?

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,200 @@
...
+    public VCFIterator open(final String path) throws IOException {
+        return open(ParsingUtils.openInputStream(path, null));
+    }
+
+    /**
+     * creates a VCF iterator from a Path
+     * 
+     * @param path the path
+     * @return the VCFIterator
+     * @throws OException
```
ronlevine: Change to `IOException`. Tell when it will be thrown, same comment throughout file.

src/test/java/htsjdk/variant/vcf/VCFIteratorTest.java
```diff
@@ -0,0 +1,90 @@
...
+    @DataProvider(name = "VcfFiles")
+    public Object[][] getVcfFiles() {
+        return new Object[][] { 
+                new Object[] { "src/test/resources/htsjdk/tribble/tabix/testTabixIndex.vcf", 25 },
+                new Object[] { "src/test/resources/htsjdk/tribble/tabix/testTabixIndex.vcf.gz", 25 },
+                new Object[] { "src/test/resources/htsjdk/variant/serialization_test.bcf", 12 } };
+    }
+
+    private int countVariants(VCFIterator r) {
```
ronlevine: Make input variables final. Same comment for all methods in file.

src/test/java/htsjdk/variant/vcf/VCFIteratorTest.java
```diff
@@ -0,0 +1,90 @@
...
+        return (int) StreamSupport.stream(Spliterators.spliteratorUnknownSize(r, Spliterator.ORDERED), true).count();
+    }
+
+    @Test(dataProvider = "VcfFiles")
+    public void testUsingUri(String uri, int nVariants) throws IOException {
```
ronlevine: Should `url` be `path`? Same comment got next test.

src/test/java/htsjdk/variant/vcf/VCFIteratorTest.java
```diff
@@ -0,0 +1,90 @@
...
+        VCFIterator r = new VCFIteratorBuilder().open(in);
+        Assert.assertNotNull(r.getFileHeader());
+        Assert.assertEquals(countVariants(r), nVariants);
+        r.close();
+        in.close();
+    }
+    
+    @Test(dataProvider = "VcfFiles")
+    public void testUsingPath(String file, int nVariants) throws IOException {
+        Path path =Paths.get(file);
```
ronlevine: space after `=`.

src/test/java/htsjdk/variant/vcf/VCFIteratorTest.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Test(dataProvider = "VcfFiles")
+    public void testUsingStreams(String file, int nVariants) throws IOException {
+        InputStream in = new FileInputStream(file); 
+        VCFIterator r = new VCFIteratorBuilder().open(in);
+        Assert.assertNotNull(r.getFileHeader());
+        Assert.assertEquals(countVariants(r), nVariants);
+        r.close();
+        in.close();
```
ronlevine: I don't think this is needed, handled by `r.close()`.

ronlevine: Can these 3 test be combined into one? Can there be round trip validation of the input file to ensure that what comes in is the same as what comes out?

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,200 @@
...
+     */
+    public VCFIterator open(final InputStream in) throws IOException {
+        if (in == null) {
+            throw new IllegalArgumentException("input stream is null");
+            }
+        // try to read the BCF header
+        final BufferedInputStream bufferedinput = new BufferedInputStream(in, 100);
+        bufferedinput.mark(100);
+        final BCFVersion bcfVersion = BCFVersion.readBCFVersion(bufferedinput);
+        bufferedinput.reset();
```
ronlevine: Actually, this can be replaced by `BCFVersion#readBCFVersion(final InputStream stream)`.

lindenb:
@yfarjoun @lbergelson @ronlevine thank you for the reviews.

* I've fixed the typos, added `final`
* removed gunzipIfNeeded from IOUtils and moved it back to VCFIteratorBuilder
* added ` IOUtil.GZIP_SIGNATURE`
* using constants to create a bufferèsize for `new BufferedInputStream` in VCFIteratorBuilder
* simplified VCFIteratorBuilderTest

thanks

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -988,4 +996,20 @@ public static Path getPath(String uriString) throws IOException {
             return FileSystems.newFileSystem(uri, new HashMap<>(), cl).provider().getPath(uri);
         }
     }
+    
+    /**
+     * Test whether a input stream looks like a GZIP input.
```
ronlevine: Add a blank line.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,204 @@
...
+        final BufferedInputStream bufferedinput = new BufferedInputStream(in,bufferSize);
+        bufferedinput.mark(bufferSize);
+        final BCFVersion bcfVersion = BCFVersion.readBCFVersion(bufferedinput);
+        bufferedinput.reset();
+
+        if (bcfVersion != null) {
+            //this is BCF
+            return new BCFInputStreamIterator(bufferedinput);
+        } else {
+            //
```
ronlevine: extra comment line

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,204 @@
...
+     * @param in inputstream
+     * @return the VCFIterator
+     * @throws IOException
+     */
+    public VCFIterator open(final InputStream in) throws IOException {
+        if (in == null) {
+            throw new IllegalArgumentException("input stream is null");
+            }
+        // buffer must be large enough to contain the BCF header (+ min/max version) and/or GZIP signature
+        final int bufferSize = Math.max( BCFVersion.MAGIC_HEADER_START.length + 2*Byte.BYTES , IOUtil.GZIP_SIGNATURE.length);
```
ronlevine: extra space after `Math.max(`.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,204 @@
...
+        } else {
+            //
+            //this is VCF or VCF.gz, gunzip the input stream if needed
+            return new VCFReaderIterator( IOUtil.isGZIPInputStream(bufferedinput) ? 
+                    new GZIPInputStream(bufferedinput) : bufferedinput );
+        }
+    }
+
+    /**
+     * creates a VCF iterator from a URI It detects if the stream is a BCF
```
ronlevine: Change URI to file path

droazen: Why? `ParsingUtils.openInputStream()` handles more than just files (it handles http URIs, as well as URIs with schemes supported by NIO plugins)

ronlevine: Correct. Ignore my comment.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,204 @@
...
+
+        if (bcfVersion != null) {
+            //this is BCF
+            return new BCFInputStreamIterator(bufferedinput);
+        } else {
+            //
+            //this is VCF or VCF.gz, gunzip the input stream if needed
+            return new VCFReaderIterator( IOUtil.isGZIPInputStream(bufferedinput) ? 
+                    new GZIPInputStream(bufferedinput) : bufferedinput );
+        }
```
ronlevine: Should return null or throw an excpetion if `in` is not BCF, VCF or VCF.gz.

lindenb:
@yfarjoun @lbergelson @ronlevine thanks & back to you. Sorry for the late response, I do this PR on my 'workflow is running' time :-)

lindenb:
friendly ping for this one please.

yfarjoun:
@lbergelson could you take a look as well?

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -97,6 +96,11 @@
 
     private static int compressionLevel = Defaults.COMPRESSION_LEVEL;
 
+    /** signature of a gzip stream used by {@link isGZIPInputStream} 
+     * @see {@linkplain http://stackoverflow.com/questions/4818468/how-to-check-if-inputstream-is-gzipped}
+     */
+    public static final byte GZIP_SIGNATURE[]= {(byte) 0x1f, (byte) 0x8b};
```
lbergelson: It would be better I think to use the `GZIPInputStream.GZIP_MAGIC` constant directly instead of reproducing it here?

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -97,6 +96,11 @@
 
     private static int compressionLevel = Defaults.COMPRESSION_LEVEL;
 
+    /** signature of a gzip stream used by {@link isGZIPInputStream}
```
lbergelson: I think there are a few issues with the javadoc that my IDE pointed out to me.

 1.  to make that link work you need it to be `{@link #isGZIPInputStream}`
 2.  the `@see` doesn't need a `{@linkplain}`, it should be `* @see "http://stackoverflow.com/questions/4818468/how-to-check-if-inputstream-is-gzipped"` I believe.

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -475,9 +479,13 @@ public static void assertFilesEqual(final File f1, final File f2) {
             while ((len1 = s1.read(buf1)) != -1) {
                 final int len2 = s2.read(buf2);
                 if (len1 != len2) {
+                    s1.close();
```
lbergelson: Good catch on these.  Instead of putting closes at all the exit points, I think it would make more sense to lift the creation of the FileInputStreams into a try-with-resources block and then they'll be guaranteed to be closed no matter where the function exits.

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -988,4 +996,20 @@ public static Path getPath(String uriString) throws IOException {
             return FileSystems.newFileSystem(uri, new HashMap<>(), cl).provider().getPath(uri);
         }
     }
+
+    /**
+     * Test whether a input stream looks like a GZIP input.
+     * @param bufferedinput the input stream. The buffer must be large enough to contain {@link GZIP_SIGNATURE}
```
lbergelson: similar javadoc issues to above

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -988,4 +996,20 @@ public static Path getPath(String uriString) throws IOException {
...
+     * @return true if `bufferedinput` starts with a gzip signature 
+     * @throws IOException if `bufferedinput` cannot mark or reset the stream
+     * @see {@linkplain http://stackoverflow.com/questions/4818468/how-to-check-if-inputstream-is-gzipped}
+     */
+    public static boolean isGZIPInputStream(final BufferedInputStream bufferedinput) throws IOException {
```
lbergelson: Should this be just a simple `InputStream` but check markSupported and then throw if it isn't?  In any case it should probably explicitly check `markSupported()` and throw a helpful error message if it isn't.

src/main/java/htsjdk/variant/vcf/VCFIterator.java
```diff
@@ -0,0 +1,56 @@
...
+ * r.close();
+ * </pre>
+ * 
+ * @author Pierre Lindenbaum / @yokofakun
+ * @see {@link VCFIteratorBuilder}
+ *
+ */
+public interface VCFIterator extends CloseableIterator<VariantContext> {
+    /** Returns the VCFHeader associated with this VCF/BCF file. */
+    public VCFHeader getFileHeader();
```
lbergelson: I would rename this to `getHeader()` since it's not necessarily read from a file.

src/main/java/htsjdk/variant/vcf/VCFIterator.java
```diff
@@ -0,0 +1,56 @@
...
+* FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR
+* THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+package htsjdk.variant.vcf;
+
+import htsjdk.samtools.util.CloseableIterator;
+import htsjdk.variant.variantcontext.VariantContext;
+
+/**
+ * An iterator used to read variants from an InputStream. It can be used to
```
lbergelson: This description seems misplaced, since the iterator itself doesn't have any of that functionality.  This comment seems like it should be on the `VCFIteratorBuilder()` instead.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,207 @@
...
+ *     System.out.println(r.next());
+ * }
+ * r.close();
+ * </pre>
+ * 
+ * @author Pierre Lindenbaum / @yokofakun
+ * @see {@link VCFIterator}
+ *
+ */
+
```
lbergelson: accidental linebreak here

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,207 @@
...
+        // buffer must be large enough to contain the BCF header (+ min/max version) and/or GZIP signature
+        final int bufferSize = Math.max(BCFVersion.MAGIC_HEADER_START.length + 2*Byte.BYTES , IOUtil.GZIP_SIGNATURE.length);
+        // try to read the BCF header
+        final BufferedInputStream bufferedinput = new BufferedInputStream(in,bufferSize);
+        bufferedinput.mark(bufferSize);
+        final BCFVersion bcfVersion = BCFVersion.readBCFVersion(bufferedinput);
+        bufferedinput.reset();
+
+        try {
+            if (bcfVersion != null) {
```
lbergelson: what about gzipped bcf?

src/main/java/htsjdk/samtools/util/IOUtil.java
lbergelson: I noticed there's a similar method `SamStreams.isGzippedSAMFile`. Instead of just checking the header it actually tries to decompress the file and catches exceptions.  Your way may be slightly more elegant, but seems a little more brittle.  In either case, it should probably be refactored to use this method if that's possible.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,207 @@
...
+                //this is BCF
+                return new BCFInputStreamIterator(bufferedinput);
+            } else {
+                //this is VCF or VCF.gz, gunzip the input stream if needed
+                return new VCFReaderIterator( IOUtil.isGZIPInputStream(bufferedinput) ?
```
lbergelson: You're not special casing bgzip files here, but I assume that's fine since you're not allowing seeking?

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,207 @@
...
+    /**
+     * creates a VCF iterator from a URI It detects if the stream is a BCF
+     * stream or a GZipped stream.
+     * 
+     * @param path the Path
+     * @return the VCFIterator
+     * @throws IOException
+     */
+    public VCFIterator open(final String path) throws IOException {
+        return open(ParsingUtils.openInputStream(path, null));
```
lbergelson: Could you add an overload that take the wrapper parameter as well?

src/test/java/htsjdk/variant/vcf/VCFIteratorTest.java
```diff
@@ -0,0 +1,87 @@
...
+    public Object[][] getVcfFiles() {
+        return new Object[][] { 
+                new Object[] { "src/test/resources/htsjdk/tribble/tabix/testTabixIndex.vcf", 25 },
+                new Object[] { "src/test/resources/htsjdk/tribble/tabix/testTabixIndex.vcf.gz", 25 },
+                new Object[] { "src/test/resources/htsjdk/variant/serialization_test.bcf", 12 } };
+    }
+
+    private void countVariants(final VCFIterator r, final int expectVariants) {
+        Assert.assertNotNull(r.getFileHeader());
+        final int nVariants =  (int) StreamSupport.stream(Spliterators.spliteratorUnknownSize(r, Spliterator.ORDERED), true).count();
```
lbergelson: As a `CloseableIterator`, `VCFIterator` already has this operation baked in as the inherited method `.stream()`

src/test/java/htsjdk/variant/vcf/VCFIteratorTest.java
```diff
@@ -0,0 +1,87 @@
...
+        Assert.assertNotNull(r.getFileHeader());
+        Assert.assertEquals(nVariants, expectVariants);
+        r.close();
```
lbergelson: it's good practice to put a close like this in the finally of a try-finally block so it's executed even when the test fails assertions

src/test/java/htsjdk/samtools/util/IoUtilTest.java
```diff
@@ -187,4 +191,35 @@ public void testFileTypeUnix(final String path, boolean expectedIsRegularFile) {
...
+        gzout.close();
+        //and now test for compression
+        in = new BufferedInputStream(new ByteArrayInputStream(bos.toByteArray()),10);
+        Assert.assertTrue(IOUtil.isGZIPInputStream(in));
+        // call twice to verify 'in.reset()' was called
+        Assert.assertTrue(IOUtil.isGZIPInputStream(in));
+        in.close();
+    }
+    @DataProvider(name = "mayBeGZippedInputStreamTestCases")
+    private Object[][] mayBeGZippedInputStreamTestCases() {
```
lbergelson: you added this dataprovider but you didn't actually do anything with it I think

src/test/java/htsjdk/samtools/util/IoUtilTest.java
```diff
@@ -187,4 +191,35 @@ public void testFileTypeUnix(final String path, boolean expectedIsRegularFile) {
...
+        return new Object[][]{
+                {"Hello World", Boolean.FALSE},
```
lbergelson: you can just use the constants `false` and `true` instead `Boolean.FALSE`,
also no need to put the hello world in the dataprovider if it's always the same.

src/test/java/htsjdk/variant/vcf/VCFIteratorTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @DataProvider(name = "VcfFiles")
+    public Object[][] getVcfFiles() {
+        return new Object[][] { 
+                new Object[] { "src/test/resources/htsjdk/tribble/tabix/testTabixIndex.vcf", 25 },
+                new Object[] { "src/test/resources/htsjdk/tribble/tabix/testTabixIndex.vcf.gz", 25 },
+                new Object[] { "src/test/resources/htsjdk/variant/serialization_test.bcf", 12 } };
+    }
+
+    private void countVariants(final VCFIterator r, final int expectVariants) {
```
lbergelson: This is very nitpicky, but I would name this method something like `assertExpectedNumberOfVariants` since it performs an assertion.  I would expect a method called `countVariants` to return the number of variants, not to be a void method that sometimes blows up.

src/test/java/htsjdk/variant/vcf/VCFIteratorTest.java
lbergelson: (although switching to use the baked in .stream() method closes it for you.)

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
lindenb: >  since you're not allowing seeking?

true

lindenb:
> @lindenb Sorry this sat for so long

@lbergelson no worries & thanks for the review !

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
lindenb: I dont get that one: did you mean  adding an `open(String path,Function<SeekableByteChannel, SeekableByteChannel> wrapper)` ?

lbergelson: Yes, that's exactly what I mean.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
lindenb: @lbergelson I need some advice on this point: I don't have a great experience with BCF. I've been working with `resources/test/htsjdk/variant/serialization_test.bcf `which I thought was a valid bcf file. But it's not compressed:

```
htsjdk$ find . -name "*.bcf" | xargs file
(...)
./build/resources/test/htsjdk/variant/serialization_test.bcf: data
```

on the other side, John Marshall says that a bcf is de-facto a bgz file   https://twitter.com/jomarnz/status/908768415971766272 .  It seems that there is no '*.bcf.gz' file ? Who's wrong ? Is there a true-valid bcf file under the 'resources/test' directory ?

Thanks

lindenb: https://twitter.com/jomarnz/status/908775081236680705

> There is also “raw BCF” which bcftools outputs with -Ou. Historically HTSJDK has only supported raw BCF not real BCF

so, in htsjdk, the suffix '*.bcf' doesn't imply that the file is bgzipped ?

lbergelson:
@lindenb This seems to have some compilation errors since the rebase.

lindenb:
@lbergelson yes, thanks ,  it saw it. I would like to have your feedback about the way htsjdk handles bcf. see https://github.com/samtools/htsjdk/pull/837#discussion_r139231337

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
lbergelson: So, it turns out I was confused.  I had assumed that it was always bgzipped, but it turns out htsjdk only handles uncompressed bcf at the moment. The spec recommends but doesn't require that bcf be compressed, but I think all real usage is with compressed files.  There's a ticket to fix it so that it handle proper block compressed bcfs https://github.com/samtools/htsjdk/issues/946.  No need for you to do anything I think.

lbergelson:
@lindenb I replied, turns out htsjdk bcf handling is worse than I thought it was, but you don't need to make any changes.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
lindenb: @lbergelson ok thanks. I've worked on this, I'll push my latest commit tomorrow morning.

lindenb:
@lbergelson thanks &  back to you.

I hope I have made all the changes you wanted.  

* i'm now calling  IOUtil.isGZIPInputStream(stream)  moved and invoked from  SamStreams.isGzippedSAMFile() 
* I've created a 3rd implementation of VCFIterator in VCFIteratorBuilder that wraps a VCFFFileReader. This new simple class avoids to wrap the InputStream into the set of BufferedInputStream  of `open(InputStream)`.
* I've added some comments with links to this PR to explain why we only handle "vcf" / "vcf.gz" files at some points.
* I've added some tests that compress a VCF using gzip or bzip and  test if we can re-read the new compressed file.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,282 @@
...
+            bufferedinput = new BufferedInputStream(
+                    new GZIPInputStream(bufferedinput),
+                    sizeof_bcf_header
+                    );
+        }
+
+        // try to read a BCF header
+        bufferedinput.mark(sizeof_bcf_header);
+        final BCFVersion bcfVersion = BCFVersion.readBCFVersion(bufferedinput);
+        bufferedinput.reset();
```
cmnbroad: I'd lobby to move this BCF-sniffing logic out into a public method somewhere, probably in BCFCodec; maybe to a static method that returns the BCFCodec or null on failure. In addition to keeping the knowledge of the BCF internals inside the BCFCodec, it would be immediately useful to solve [this problem](https://github.com/HadoopGenomics/Hadoop-BAM/issues/132) in Hadoop-BAM.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,282 @@
...
+    public VCFIterator open(final File file) throws IOException {
+        return new FileBasedVCFIterator(file);
+    }
+
+    /** 
+     * implementation of VCFIterator, wrapping a VCFFileReader.
+     * using this class instead of VCFReaderIterator below,
+     * avoid to wrap an InputStream into a set of BufferedInputStream 
+     * */
+    private static class FileBasedVCFIterator
```
cmnbroad: I'm not sure I understand the advantage of having this implementation class - why is this better than just having the open(File) method open a stream and delegate to open(InputStream). I THINK the input stream will get wrapped in (at least 1!) buffered stream either way.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,282 @@
...
+        /** VCF codec */
+        private final VCFCodec codec = new VCFCodec();
+        /** VCF header */
+        private final VCFHeader vcfHeader;
+        /** Iterator over the lines of the VCF */
+        private final AsciiLineReaderIterator lineIterator;
+
+        VCFReaderIterator(final InputStream inputStream) {
+            this.inputStream = inputStream;
+            this.lineIterator = new AsciiLineReaderIterator(AsciiLineReader.from(inputStream));
```
cmnbroad: It would be better to delegate to the codec's makeSourceFromStream method here.

src/main/java/htsjdk/variant/vcf/VCFIteratorBuilder.java
```diff
@@ -0,0 +1,282 @@
...
+            implements VCFIterator {
+        /** the underlying input stream */
+        private final PositionalBufferedStream inputStream;
+        /** BCF codec */
+        private final BCF2Codec codec = new BCF2Codec();
+        /** the VCF header */
+        private final VCFHeader vcfHeader;
+
+        BCFInputStreamIterator(final InputStream inputStream) {
+            this.inputStream = new PositionalBufferedStream(inputStream);
```
cmnbroad: Same here; delegate to the codec 's makeSourceFromStream method.

src/main/java/htsjdk/samtools/SamStreams.java
```diff
@@ -66,26 +67,11 @@ public static boolean isBAMFile(final InputStream stream)
     /**
      * Checks whether the file is a gzipped sam file.  Returns true if it
      * is and false otherwise.
+     * @see {@link IOUtil#isGZIPInputStream(InputStream)}
      */
     public static boolean isGzippedSAMFile(final InputStream stream) {
```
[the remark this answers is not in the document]
lbergelson: 👍

lindenb:
@lbergelson @cmnbroad thank you for the reviews !

- I've deprecated isGzippedSAMFile
- created BCF2Codec.tryReadBCFVersion
- fixed the typo
- deleted FileBasedVCFIterator
- used the codec.makeSourceFromStream
- etc..

lindenb:
friendly ping for this one :-)

romanzenka:
Hello, I could really use this code. What can I do to make it happen?

lindenb:
I gave up for this PR , I'm now using my private src : https://github.com/lindenb/jvarkit/blob/master/src/main/java/com/github/lindenb/jvarkit/util/vcf/VcfIteratorImpl.java & https://github.com/lindenb/jvarkit/blob/master/src/main/java/com/github/lindenb/jvarkit/util/vcf/VcfIterator.java

romanzenka:
@lindenb I can totally use your local fork for a while - but maybe if you could let me know what is the holdup here, I could try to "carry the torch"?
Is it fair that what is left are @cmnbroad and @lbergelson change requests, and then you need to figure out that weird "coverage changes not visible in diff" codecov error?

lindenb:
>  lindenb:vcfiter 

https://github.com/lindenb/htsjdk/tree/vcfiter

lbergelson:
Ack.. this was one of those totally reasonable and helpful pr's that get dogpiled on by every reviewer in the world and then we all just gave up on...  Lets fix that...

romanzenka:
I made PR #1245, will continue my work there!

Many thanks to @lindenb for getting the implementation 99% there!

lbergelson:
@lindenb Fair enough... sorry it became such a mess.  Thanks for doing this work.  Should we close this then?

lindenb:
closing, moved to https://github.com/samtools/htsjdk/pull/1245

Issue #830, opened 2017-03-23
Java implementation of .fai index creation


magicDGS:
### Motivation

The `.fai` index from samtools is used in some parts of the code in HTSJDK, but there is no way to generate it within the library. Because it is well documented [here](http://www.htslib.org/doc/faidx.html) and it looks simple to implement, I think that HTSJDK could benefit for a java-based implementation of this indexing.

tfenne:
I agree, this would be nice.  It would also be nice to have the capability to create and use the indices that are generated on bgzipped fasta files which samtools supports.

dariober:
If any useful... I have the code to create `.fai` index from fasta (not bgzipped though). The code it's here: https://github.com/dariober/ASCIIGenome/tree/master/src/faidx

The tests are here: https://github.com/dariober/ASCIIGenome/blob/master/test/faidx/

Usage is pretty straightforward:

```
File fasta= new File("genome.fa"); 
new Faidx(fasta); // will create genome.fa.fai
```

Feel free to grab anything that maybe useful.

magicDGS:
I've already started an implementation, but it may be useful to refine some parts of my code. Thank you very much, @dariober!

magicDGS:
Closed by #842

magicDGS:
@tfenne, I open a different issue regarding the capability to fully support compressed FASTA.

PR #842, opened 2017-04-05
Add .fai index creation support

magicDGS:
### Description

As described in #830, it would be nice to have a way of generate a __.fai__ index from a provided reference file. This PR adds the `FastaSequenceIndexCreator` utility class, which contains:

* Method for create a __.fai__ file from a reference one. This will be useful for a bundle-tool in Picard/GATK for obtaining all required files within a folder, as the sequence dictionary.
* Method for build an in-memory index for a non-indexed file. This may be useful for allow using `IndexedFastaSequenceFile` without requiring a file-based index.

Moreover, other useful changes are included:
* Method in `FastaSequenceIndex` to write to a file in the __.fai__ format.
* Move `IndexedFastaSequenceFile.getFastaIndexFileName` private method to `ReferenceSequenceFileFactory`

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


magicDGS:
Can you have a look, @tfenne? I'm asking you because you were involved in the conversation .

Maybe you will be also interested on this @dariober...

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
```diff
@@ -0,0 +1,179 @@
...
+import htsjdk.samtools.SAMException;
+import htsjdk.samtools.util.IOUtil;
+import htsjdk.tribble.readers.AsciiLineReader;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * Static methods for create an {@link FastaSequenceIndex}.
```
tfenne: Either "Static methods to create an ..." or "Static methods for creating {@link FastaSequenceIndex}s"

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
```diff
@@ -0,0 +1,179 @@
...
+        }
+        // build the index
+        final FastaSequenceIndex index = buildFromFasta(fastaFile);
+        index.write(indexFile);
+    }
+
+    /**
+     * Builds a FastaSequenceIndex on the fly from a FASTA file.
+     *
+     * Note: this also alows to create an index for a compressed file, but does not generate the
```
tfenne: I tend to think that if we're not going to create the `.gzi` file too, that this method should probably fail with an exception on a compressed fasta file.  I can easily imagine scenarios where code uses this to build a .fai, then turns around to try and query the fasta file, only to find that it cannot because the fasta file is compressed.

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
```diff
@@ -0,0 +1,179 @@
...
+     *
+     * @param fastaFile the file to build the index from.
+     *
+     * @throws SAMException if the fai file already exists or the file is malformed.
+     * @throws IOException  if an IO error occurs.
+     */
+    public static void create(final Path fastaFile) throws IOException {
+        // get the index to write the file in
+        final Path indexFile = ReferenceSequenceFileFactory.getFastaIndexFileName(fastaFile);
+        if (Files.exists(indexFile)) {
```
tfenne: I'm curious about this decision.  I run into this kind of thing a lot when trying to do something, attempt 1 fails but creates an empty/truncated file at the output location, and then attempt 2 fails because the tool is unwilling to overwrite the broken file.  I would suggest either logging a warning and then over-writing, _or_ having a boolean parameter to the method that allows the user to force creation even if the file already exists.

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
```diff
@@ -0,0 +1,179 @@
...
+            for (String line = in.readLine(); previous != null; line = in.readLine()) {
+                // in this case, the previous line contains a header and the current line the first sequence
+                if (previous.charAt(0) == '>') {
+                    // first entry should be skipped
+                    if (entry != null) {
+                        // add the entry to the index
+                        index.add(entry.build());
+                    }
+                    // update the sequence index
+                    sequenceIndex++;
```
tfenne: Could replace this line with adding `++` after `sequenceIndex` on the line below.

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
```diff
@@ -0,0 +1,179 @@
...
+            // initialize an empty index and the entry builder to null
+            final FastaSequenceIndex index = new FastaSequenceIndex();
+            FaiEntryBuilder entry = null;
+
+            // read the lines two by two
+            for (String line = in.readLine(); previous != null; line = in.readLine()) {
+                // in this case, the previous line contains a header and the current line the first sequence
+                if (previous.charAt(0) == '>') {
+                    // first entry should be skipped
+                    if (entry != null) {
```
tfenne: Could be simplified to a one-liner:
`if (entry != null) index.add(entry.build())`

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
```diff
@@ -0,0 +1,179 @@
...
+            return index;
+        }
+    }
+
+    // utility class for building the FastaSequenceIndexEntry
+    private static class FaiEntryBuilder {
+        private final int index;
+        private final String contig;
+        private final long location;
+        // the bytes per line is the bases per line plus the end of the line
```
tfenne: I don't think this will always work - using `basesPerLine + 1`.  It's possible this code will encounter files that use `CR+LF` formatted files, in which case the bytes per line would be `basesPerLine + 2`.  Admittedly noone should do that, but that doesn't mean it doesn't happen.

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
```diff
@@ -0,0 +1,179 @@
...
+
+        private FaiEntryBuilder(final int index, final String header, final String firstSequenceLine, final long location) {
+            if (header == null || header.charAt(0) != '>') {
+                throw new SAMException("Wrong sequence header: " + header);
+            } else if (firstSequenceLine == null) {
+                throw new SAMException("Empty sequences could not be indexed");
+            }
+            this.index = index;
+            // parse the contig name (without the starting '>')
+            this.contig =  header.substring(1);
```
tfenne: This doesn't appropriately handle headers that have spaces in the name.  With `samtools faidx` if you have a sequence with a header like:
`>chr1 Human chromosome 1`
The name that is written to the `.fai` file is just `chr1`.

I think you want to do something like:
```java
SAMSequenceRecord.truncateSequenceName(header.substring(1).trim());
```

in order to handle both spaces between the `>` and the contig name, and then comments after the name.

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
```diff
@@ -0,0 +1,179 @@
...
+            this.location = location;
+            this.basesPerLine = firstSequenceLine.length();
+
+            this.size = firstSequenceLine.length();
+            this.lessBasesFound = false;
+        }
+
+        private void updateWithSequence(final String sequence) {
+            if (sequence.length() > basesPerLine) {
+                throw new SAMException(String.format("Sequence bases for '{}' exceed the maximum length ({}): {}",
```
tfenne: The message here is technically correct, but I think I would find it confusing, since other than the last line all lines should be the same length, not variable lengths with a maximum.  How about "Sequence line for {} was longer than the expected length ({}): {}"

src/main/java/htsjdk/samtools/reference/FastaSequenceIndex.java
```diff
@@ -159,6 +162,27 @@ private void parseIndexFile(Path indexFile) {
     }
 
     /**
+     * Writes this index to the specified path.
```
magicDGS: Done.

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
magicDGS: Done

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
magicDGS: I always prefer to do not overwrite files unless I know that I'm doing it in purpose. But I agree that a parameter to the method is a good addition, so added.

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
magicDGS: I added this note because it is a difference with respect to the `samtools faidx` command.

I can also imagine scenarios where where a compressed fasta file is already indexed with bgzip (`.gzi` index), and the only requirement is to create the `.fai` index. Anyway, if there is a way to create with htsjdk a `.gzi` index I can integrate it here in the case of block-compressed input.

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
magicDGS: This is my personal taste, but I always prefer brackets. Anyway I changed it, because the style for htsjdk allows it.

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
magicDGS: True!

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
magicDGS: That's true, it could happen. But it is quite difficult to fix with the current implementation and even throwing it's difficult because there is no way of checking how many end-of-line are unless we iterate over the bytes. That requires a complete re-implementation.

I've just added a TODO with date to keep track, and I can open a ticket if this gets accepted as it is now.

tfenne: I think you could probably catch this quite easily at the end for any file with multiple sequences.  You could check to ensure that the start position you have for each sequence (which is the correct physical start location) matches what you get if you calculate the start of the last sequence plus the number of bytes expected given bytesPerLine.

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
magicDGS: True, I think that I removed it when I was working on it. I fixed it and added a test.

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
magicDGS: Good suggestion. Done!

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
magicDGS: That's a good suggestion, but what will happen with this:

```
>chr1\r\n
ACTG\r\n
```

tfenne: I just created this PR that I think solves the problem: https://github.com/samtools/htsjdk/pull/843 .  Want to review for me?

magicDGS: Thanks, I did it already!

magicDGS:
I pushed the changes for first round of comments @tfenne. I will update when #843 is accepted.

tfenne:
Looks good to me @magicDGS - I think you should rebase on top of master now #843 is merged, use that change to throw an exception if you encounter varying line-endings, and then merge!

magicDGS:
I rebased and included the check for the end of line concordance for each sequence. Could you please have a look, @tfenne?

tfenne:
👍  from me @magicDGS.  Are you able to merge or do you need me to do it?

magicDGS:
I cannot merge, @tfenne. Let me know if I should squash beforehand. Thanks for the review!

magicDGS:
Thanks @tfenne!

PR #843, opened 2017-04-05
Added the ability to query the length of the line terminator to Ascii…

tfenne:
…LineReader.

### Description

I wanted @magicDGS to have access to the line terminator length so that `.fai` creation could be more robust 😉 

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
+++ b/src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
@@ -40,8 +40,9 @@ public class AsciiLineReader implements LineReader, LocationAware {
     private static final byte LINEFEED = (byte) ('\n' & 0xff);
     private static final byte CARRIAGE_RETURN = (byte) ('\r' & 0xff);
 
-    PositionalBufferedStream is;
-    char[] lineBuffer;
+    private final PositionalBufferedStream is;
+    private char[] lineBuffer;
+    private int lineTerminatorLength = -1;
 
     public AsciiLineReader(final InputStream is){
         this(new PositionalBufferedStream(is));
@@ -65,6 +66,16 @@ public class AsciiLineReader implements LineReader, LocationAware {
         return is.getPosition();
     }
 
+    /** Returns the length of the line terminator read after the last read line.  Returns either:
+     * -1 if no line has been read
+     * 0  after the last line if the last line in the file had no CR or LF line ending
+     * 1  if the line ended with CR or LF
+     * 2  if the line ended with CR and LF
+     */
+    public int getLineTerminatorLength() {
+        return this.lineTerminatorLength;
+    }
+
     /**
      * Read a line of text.  A line is considered to be terminated by any one
      * of a line feed ('\n'), a carriage return ('\r'), or a carriage return
@@ -83,6 +94,7 @@ public class AsciiLineReader implements LineReader, LocationAware {
             if (b == -1) {
                 // eof reached.  Return the last line, or null if this is a new line
                 if (linePosition > 0) {
+                    this.lineTerminatorLength = 0;
                     return new String(lineBuffer, 0, linePosition);
                 } else {
                     return null;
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
```diff
@@ -93,6 +105,10 @@ public final String readLine(final PositionalBufferedStream stream) throws IOExc
             if (c == LINEFEED || c == CARRIAGE_RETURN) {
                 if (c == CARRIAGE_RETURN && stream.peek() == LINEFEED) {
                     stream.read(); // <= skip the trailing \n in case of \r\n termination
+                    this.lineTerminatorLength = 2;
+                }
+                else {
```
yfarjoun: I detect an extra `\n` here :-)

tfenne: How meta ;)  In all seriousness though I thought putting the else onto the next line was user-choice in htsjdk & picard.  Certainly the rule used to be the opposite (always put it on a new line).

src/test/java/htsjdk/tribble/readers/AsciiLineReaderTest.java
```diff
@@ -49,4 +41,20 @@ public void testReadLines() throws Exception {
         assertEquals(expectedNumber, actualLines);
 
     }
+
+    @Test public void voidTestLineEndingLength() throws Exception {
+        final String input = "Hello\nThis\rIs A Silly Test\r\nSo There";
+        final InputStream is = new ByteArrayInputStream(input.getBytes());
```
yfarjoun: What should happen if there's a stray carriage return with no linefeed? Could you add that to the test?

yfarjoun: also, I don't know what happens with the wrong order `\n\r` but it should be tested...

src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
yfarjoun: I don't know about "used to be" I was trained by @nh13 to always have `} else {` I'm not super picky, but I thought it appropriate due to the subject matter.

tfenne: Since I've been around since the inception of htsjdk, I'm pretty sure I know about "used to be" 😛

src/test/java/htsjdk/tribble/readers/AsciiLineReaderTest.java
tfenne: Seriously?  Valid line endings in use today are `CR`, `LF`, `CR+LF`.  `AsciiLineReader` specifically looks for those three endings for lines. The test already covers all three valid line ending (a solo CR, a solo LF and a CR+LF).

If they are in the wrong order, they will be treated as multiple lines (existing behaviour).  I.e. `foo\n\r` would be tokenized as `<foo><eol><empty><eol>`.

src/test/java/htsjdk/tribble/readers/AsciiLineReaderTest.java
```diff
@@ -49,4 +41,20 @@ public void testReadLines() throws Exception {
         assertEquals(expectedNumber, actualLines);
 
     }
+
+    @Test public void voidTestLineEndingLength() throws Exception {
+        final String input = "Hello\nThis\rIs A Silly Test\r\nSo There";
```
magicDGS: I think that will be nice to have also the ' \r\n' at the end for testing.

magicDGS:
Great @tfenne! This is perfect for the __.fai__ index. Thank you!

src/test/java/htsjdk/tribble/readers/AsciiLineReaderTest.java
yfarjoun: I'm not asking to modify the valid line-endings...I just want the tests to be explicit about it.

So the test will show an empty line and the correct size of line ending.

nh13: I don't think this test is necessary.

src/test/java/htsjdk/tribble/readers/AsciiLineReaderTest.java
tfenne: Good point, added.

magicDGS:
Could this be merged for using it for the fai creator? Thanks in advance! :)

yfarjoun:
👍 ...I'll add the test I want myself :-p

yfarjoun:
needs a rebase..sorry.

tfenne:
@yfarjoun Rebased and squashed.  Ok to merge once the tests pass?

yfarjoun:
💯

Issue #844, opened 2017-04-06
Indexing very large VCF files causes out of memory exception


noamBarkai:
Am using htsjdk 2.9.1, trying to index a 4gb VCF file, via IndexFactory.createLinearIndex()
It crashes on out of memory. This was somewhat surprising as the indexing is done via iteration on the input file, and hence the memory footprint should remain rather low.

Looking at the heap dump created it seems the "culprit" for the OOM is: 
htsjdk.variant.vcf.AbstractVCFCodec#stringCache

defined thus:
    protected Map<String, String> stringCache = new HashMap<String, String>();

It seems this cache is only used to "intern" the strings, presumably to make the comparisons faster? 
usage is here:
htsjdk.variant.vcf.AbstractVCFCodec#getCachedString

The cache doesn't seem to serve any other purpose.
If that's the case, I think a simple fix to the situation would be to use Weak References in the map, thus allowing the JVM to remove strings from the cache if heap grows too much.
This can be done by replacing the map above with:
    protected WeakHashMap<String, WeakReference<String>> stringCache = new WeakHashMap<>();

I can offer a pull request with the needed change

noamBarkai:
I've created the following PR:
https://github.com/samtools/htsjdk/pull/845
tested locally - OOM issue gone
all tests pass locally (except two, also failed on master I checked out, and which seem to be completely unrelated)

tfenne:
The unbounded cache behind this was removed in 2.19.0 by #1282: `AbstractVCFCodec` no longer caches REF/ALT strings, only contig and filter strings, which are bounded. Please reopen if indexing a large VCF still runs out of memory on a current release.

PR #845, opened 2017-04-07
fix issue 844: Indexing very large VCF files causes OOM exception

noamBarkai:
changed internal strings cache to weak reference based map

### DIFF
--- a/src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
+++ b/src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
@@ -42,7 +42,6 @@ import htsjdk.variant.variantcontext.LazyGenotypesContext;
 import htsjdk.variant.variantcontext.VariantContext;
 import htsjdk.variant.variantcontext.VariantContextBuilder;
 
-import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
@@ -96,6 +95,7 @@ public abstract class AbstractVCFCodec extends AsciiFeatureCodec<VariantContext>
      * only for single-sample VCFs.
      */
     protected String remappedSampleName = null;
+    protected static boolean vcfIndexUseStringsCache = "true".equals(System.getProperty("VCF_INDEX_USE_STRINGS_CACHE"));
 
     protected AbstractVCFCodec() {
         super(VariantContext.class);
@@ -381,12 +381,18 @@ public abstract class AbstractVCFCodec extends AsciiFeatureCodec<VariantContext>
      * @return interned string
      */
     protected String getCachedString(String str) {
-        WeakReference<String> internedString = stringCache.get(str);
-        if ( internedString == null || internedString.get() == null) {
-            internedString = new WeakReference<>(str);
-            stringCache.put(str, internedString);
+        if (!vcfIndexUseStringsCache)
+            return str;
+        WeakReference<String> stringWeakReference = stringCache.get(str);
+        if ( stringWeakReference != null) {
+            String ret = stringWeakReference.get();
+            if (ret != null)
+                return ret;
         }
-        return internedString.get();
+        // not found in cache - str will be put in cache
+        stringWeakReference = new WeakReference<>(str);
+        stringCache.put(str, stringWeakReference);
+        return str;
     }
 
     /**
--- a/src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
+++ b/src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
@@ -382,7 +382,7 @@ public abstract class AbstractVCFCodec extends AsciiFeatureCodec<VariantContext>
      */
     protected String getCachedString(String str) {
         WeakReference<String> internedString = stringCache.get(str);
-        if ( internedString == null ) {
+        if ( internedString == null || internedString.get() == null) {
             internedString = new WeakReference<>(str);
             stringCache.put(str, internedString);
         }
--- a/src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
+++ b/src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
@@ -95,7 +95,7 @@ public abstract class AbstractVCFCodec extends AsciiFeatureCodec<VariantContext>
      * only for single-sample VCFs.
      */
     protected String remappedSampleName = null;
-    protected static boolean vcfIndexUseStringsCache = "true".equals(System.getProperty("VCF_INDEX_USE_STRINGS_CACHE"));
+    protected static boolean vcfCodecUseStringsCache = "true".equals(System.getProperty("VCF_CODEC_USE_STRINGS_CACHE"));
 
     protected AbstractVCFCodec() {
         super(VariantContext.class);
@@ -381,7 +381,7 @@ public abstract class AbstractVCFCodec extends AsciiFeatureCodec<VariantContext>
      * @return interned string
      */
     protected String getCachedString(String str) {
-        if (!vcfIndexUseStringsCache)
+        if (!vcfCodecUseStringsCache)
             return str;
         WeakReference<String> stringWeakReference = stringCache.get(str);
         if ( stringWeakReference != null) {
--- a/src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
+++ b/src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
@@ -46,18 +46,10 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.ref.WeakReference;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.StringTokenizer;
+import java.util.*;
 import java.util.zip.GZIPInputStream;
 
 
@@ -90,7 +82,7 @@ public abstract class AbstractVCFCodec extends AsciiFeatureCodec<VariantContext>
 
     protected int lineNo = 0;
 
-    protected Map<String, String> stringCache = new HashMap<String, String>();
+    protected Map<String, WeakReference<String>> stringCache = new WeakHashMap<>();
 
     protected boolean warnedAboutNoEqualsForNonFlag = false;
 
@@ -389,12 +381,12 @@ public abstract class AbstractVCFCodec extends AsciiFeatureCodec<VariantContext>
      * @return interned string
      */
     protected String getCachedString(String str) {
-        String internedString = stringCache.get(str);
+        WeakReference<String> internedString = stringCache.get(str);
         if ( internedString == null ) {
-            internedString = new String(str);
-            stringCache.put(internedString, internedString);
+            internedString = new WeakReference<>(str);
+            stringCache.put(str, internedString);
         }
-        return internedString;
+        return internedString.get();
     }
 
     /**
--- a/build.gradle
+++ b/build.gradle
@@ -42,9 +42,6 @@ dependencies {
     compile "org.tukaani:xz:1.5"
     compile "gov.nih.nlm.ncbi:ngs-java:1.2.4"
 
-//    compile "org.openjdk.jmh:jmh-core:1.0'"
-//    compile "org.openjdk.jmh:jmh-generator-annprocess:1.0"
-
     testCompile "org.testng:testng:6.9.9"
     testCompile "com.google.jimfs:jimfs:1.1"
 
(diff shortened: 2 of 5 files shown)


src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -389,12 +381,12 @@ public void setName(String name) {
      * @return interned string
      */
     protected String getCachedString(String str) {
-        String internedString = stringCache.get(str);
+        WeakReference<String> internedString = stringCache.get(str);
```
yfarjoun: I thought you need to also check that the referenced string inside of the WeakReference hasn't been GC'ed...shouldn't the if statement be something like
```
if ( internedString == null || internedString.get()==null) 
```
???

yfarjoun:
I am very worried by this PR. 

either the interning is important, and replacing it with WeakReferences will break the interning,
or the interning is not important in which case, it should be removed wholesale.

@tfenne ? @nh13 ? @droazen ? Do you know why we have this weird map from a string to itself in VcfCodec?

noamBarkai:
@yfarjoun you're absolutely right about the weak reference check for nullity - fixed it - good catch!
 
The original code isn't just interning the strings, but also holding references to them, hence bloating memory. Re utility, I can make some very basic benchmarks w/ an w/o this "cache", but that would only be on a particular dataset, whereas I would think the effectiveness of this technique would depend on overall number of strings and their cardinality, no?

noamBarkai:
one more comment: from what I know, string interning is done for two main reasons: 1) lowering memory footprint; and 2) allowing use of == to compare strings. from what I see (albeit, only via manual inspection of the code) there's no usage of 2) and 1) is obviously not achieved this way, on the contrary. so maybe indeed removing this map altogether is the proper way to go?

yfarjoun:
@noamBarkai, I think that the solution you propose is only a half-way solution. I would prefer a replacement with a Cache. Perhaps the Guava one since the library is already in htsjdk?
 
I suspect that your problem is that you have many many different filter strings or alleles and that is filling up the memory. I imagine that using a capacity of 200 should be enough for the high use string, and the low use strings do not matter anyway.

I would much prefer a solution that gives us control over how many things we keep in memory, then a magical "WeakReference" implementation....

yfarjoun:
To answer your comments: 

1. If you want to hold a large collection of variantContexts in memory, storing a separate copy of the contig string and alleles in memory isn't a good idea...so the interning will help there. 

2. `==` is used automatically inside String.equals, so it will help there too. the fact that you do not see the usage doesn't mean anything since htsjdk is used within picard and gatk, igv, etc.

tfenne:
@yfarjoun This is my first time ever looking at that class, and .... I'm a little horrified.  I don't think the `getCachedString()` method does anything other than fill memory up with copies of the string.  Unless I'm mistaken, despite referring to `internedString`, there is no actual interning!  But 'stringCache` is protected, so in theory there could be accesses outside of the class and the `getCachedString` method.  Ugh.

From my reading it would be a 100% improvement to simply remove the `stringCache` and `getCachedString()` and fix any usages to just use the strings directly.

tfenne:
Also, regarding your point 2, while it's always acceptable to test String equality with `a == b || a.equals(b)`, if any code is relying on `a == b` exclusively without controlling the intern itself... well, it should not do that!

lbergelson:
@tfenne Unless I'm reading it wrong, if you call `getCachedString` and an equal string is already in the cache, then you get back canonical copy of it.  So it would have the effect of reducing memory usage after garbage collection, you'd only end up holding a reference to the canonical string instead of many duplicates.

It's not using the explicit string interning mechansim, but I think it's doing effectively the same thing.  Although there seems to be an extra string allocation in there that I'm not clear on the reason for...

tfenne:
Good point @lbergelson.  Though the overhead of keeping the `Map<String,String>` may outweigh the savings of caching the strings depending on usage.  Maybe it makes sense to actually intern FILTER strings, and then get rid of the caching mechanism?  I think caching every ref and alt seen is going to cause problems eventually since with indels in large call sets you'll end up caching half the genome or more!

yfarjoun:
which is why i suggested replacing it with a cache of about 200 strings. the most recently used ones will be kept, and the rest will be released.

tfenne:
Perhaps @droazen can comment since he is the original author?  @yfarjoun I'm just not convinced that caching is at all necessary, so am advocating for simplifying vs. adding yet more complication (I  would see MRU as a complication - and also I would then argue that MFU would be more helpful, yada yada).

droazen:
I am not actually the original author @tfenne, and object strongly to that characterization :)  I just moved this code from GATK to htsjdk a long time ago, and so am git blamed for it.

tfenne:
Lol, apologies @droazen !  I should have thought twice before trusting the git-blame output!

yfarjoun:
Oh, I'm sure it's a certain MDP...

noamBarkai:
@yfarjoun I can replace this with a Guava cache, sure. 
But of course getting MRU or MFU behaviour isn't free, and could potentially offset the gains in string comparison no? At least for the immediate use-case I'm looking at right now, I doubt it would be very beneficial. 
I could parameterize the Guvava cache size, possibly allowing a "zero size" that would mean "bypass the whole mechanism" - but as @tfenne pointed out, this would further complicate the code, now with an additional configuration.
If there's some "canonical" dataset I can use to benchmark the alternatives I might be able to do that. 
I also think my code is still not completely immune to NPE - in theory if the weakly referenced string got gc'ed right before line 389 my code will return a null.. so I think if you decide to keep the weak reference option I'll have to further modify it a bit.

noamBarkai:
BTW, I ran a very simple test-run on my local dev machine (Intel(R) Core(TM) i5-4300U CPU @ 1.90GHz) using a 2.3Gb VCF file as input, calling IndexFactory.createLinearIndex(input, new VCFCodec()) (i.e. using default bin-size). 
With my Weak Reference code it took ~80 seconds
Whereas having the method just return the original string took ~70 seconds.
Obviously this can't count as anything nearing a serious benchmark, but nevertheless worth mentioning I guess

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -96,6 +95,7 @@
      * only for single-sample VCFs.
      */
     protected String remappedSampleName = null;
+    protected static boolean vcfIndexUseStringsCache = "true".equals(System.getProperty("VCF_INDEX_USE_STRINGS_CACHE"));
```
yfarjoun: I think that the property name is rather confusing, this has nothing to do with indexes. it affects the entire vcf codec system, which is why this PR potentially has very wide-reaching effects and should not be taken lightly.

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -381,12 +381,18 @@ public void setName(String name) {
      * @return interned string
      */
     protected String getCachedString(String str) {
-        WeakReference<String> internedString = stringCache.get(str);
-        if ( internedString == null || internedString.get() == null) {
-            internedString = new WeakReference<>(str);
-            stringCache.put(str, internedString);
+        if (!vcfIndexUseStringsCache)
+            return str;
+        WeakReference<String> stringWeakReference = stringCache.get(str);
```
yfarjoun: As I said before, I dislike the use of WeakReferences for this for two reasons:
1. It leaves half the problem there (the keys). We had a memory "explosion" with a map that has keys and values. Removing the values will only make 1/2 the problem go away
2. A WeakReference gives us very little control over when the objects are GCd. I'd prefer to see a LFU (or even preferably LFUDA) cache be used instead.

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
noamBarkai: indeed, will change.

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
noamBarkai: re 1 - note the map is a WeakHashMap - to my understanding that will solve the issue for the keys as well, not just the values, no?
re 2 - indeed, but the flip side is that maintaining an LFU (LFUDA) isn't free either. 
I think this whole things pivots around one basic thing: benchmarks. without conducting some, preferably on varied data-sets, I can't see how one can "apriori" conclude which is preferable: no-caching / weak references / interning / LRU's etc.

tfenne: I totally agree with you @noamBarkai - we all have opinions, but a well-defined benchmark would be very handy.  To my way of thinking there are a couple of different things to benchmark:
1) A VCF with a a very high density of variants (maybe the exome VCF and one of the chromosome VCF for the genomes from http://gnomad.broadinstitute.org/downloads ?)
2) A VCF with a lot of genotyped samples - I'm not really sure what's out there for this other than some 1KG VCFs

Then the question is what is the benchmark on these files?  I think I'd suggest two benchmarks:
1) Reading/decoding the entire file and doing something very simple, e.g. counting the total number of genotypes 
2) Read the file and re-writing it - perhaps even sorting it?

If you're up for doing some of these I'm happy to work with you @yfarjoun to try and specify some benchmarks more clearly.

yfarjoun: My main concern is that if we do this, I'd like there to be a test-target that covers this, but it would have to live in GATK or Picard and would then not be directly testing htsjdk, and it would be difficutl to test branches/PRs in htsjdk...also, it would be a long-running test and so should only be invoked "on occasion"...I'm not sure I know how to set something like this up....

tfenne: Why would the test have to live in GATK or Picard?  I'm a little confused.

yfarjoun: Maybe I'm confused, but I was thinking of running SortVcf. Looking at the code, it seems that most of it has to do with collecting multiple inputs...so i retract my statement. it can easily be done within htsjdk.

noamBarkai:
I pushed integration of [jmh](http://openjdk.java.net/projects/code-tools/jmh/) benchmarking with gradle
am using [this plugin](https://github.com/melix/jmh-gradle-plugin)
to run:
./gradlew clean jmh
this is really just a "skeleton" for any actual benchmarking

yfarjoun:
@noamBarkai Do you mind telling us what program you were running that led you to the OOM error? how much memory was given?

noamBarkai:
@yfarjoun - gladly: as I wrote in the original issue (#844) I was trying to index a 2.5Gb VCF file using a JVM instance with 1Gb RAM. 
I would expect this to easily be doable, which is indeed the case after either completely circumventing or at least limiting the aforementioned cache.

yfarjoun:
I'm surprised that indexing a 2.5 Gb vcf with 1Gb of ram failed. Could you comment about the reference you are using and also the "shape" of the vcf (rows/samples)? This will not affect this PR, but I"m trying to understand when this would fail...

Regarding the testing, we will be looking at the effects of removing the interning altogether (both in terms of correctness and in terms of speed). if correctness is still OK, and the speed becomes a problem, I think I'd prefer a cache than a WeakMap of WeakReferences.

noamBarkai:
@yfarjoun the aforementioned file has:
8M rows
4 samples
avg sequence length: 180 chars
max sequence length: 9000 chars
the organism in question is maize, which usually has a relatively high level of structural variations

yfarjoun:
I see. So the OOM you saw was likely due to the large number of **alleles** in your vcf. 

I suggest that we keep contigs, filters,  and **one character** alleles interned with the current implementation but forgo entirely the interning of longer alleles. 

@tfenne ? @lbergelson ? What do you think? Do you see any value in interning long alleles?

tfenne:
@yfarjoun I think I'm on record as favoring getting rid of all this caching (I refuse to call it `intern`ing since it's not actually using `String.intern()`).  I'm still not convinced that, without benchmarks, doing a map lookup on each of these strings is a good tradeoff for reducing (in theory) the memory footprint.

At the very least, yes, I'm in favor of not caching the allele strings.

yfarjoun:
I am happy to work with you to define good benchmarks, and didn't mean to imply that we will pull an idea out of out collective 🎩 and run with it. 

I just wanted to record another solution that might alleviate the problem observed, without forgoing all the potential benefit of the "interning", and also without the added complication of using a LRU or LFU cache.

droazen:
I share @tfenne 's skepticism that we're making a good tradeoff either performance-wise or memory-wise with this pseudo-interning scheme -- whatever benchmarks we end up doing should definitely include a "no-interning" implementation as a baseline sanity check.

We should also bear in mind that most tools that process VariantContexts don't accumulate very many in memory at once, and those few that do require some degree of manual memory management anyway. Based on my knowledge of the psychology of the people who likely implemented this interning scheme, I am pretty sure that it was originally implemented to speed up `equals()` operations rather than to control memory use -- and it's unclear to me whether the extra hash lookups don't actually make things worse in that respect for a typical tool.

yfarjoun:
So, @tfenne @jacarey @lbergelson, what's a reasonable way forward here? if we have a PR that removes the interning altogether, what tests do we perform on it?

yfarjoun:
Given the push towards hts-next, I'm closing this PR. if there is renewed interest in it, it could be opened.

PR #846, opened 2017-04-07
Add fromPath to IntervalList.

nh13:
I added a `fromPath` method that is analogous to the `fromFile` static method in `IntervalList`.  I added a simple test to make both behave similarly.

### DIFF
--- a/src/main/java/htsjdk/samtools/util/IntervalList.java
+++ b/src/main/java/htsjdk/samtools/util/IntervalList.java
@@ -401,13 +401,18 @@ public class IntervalList implements Iterable<Interval> {
         return fromPath(file.toPath());
     }
 
+    /**
+     * Parses an interval list from a path.
+     * @param path the path containing the intervals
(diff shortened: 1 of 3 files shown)


nh13:
@yfarjoun or @droazen care to review?

src/main/java/htsjdk/samtools/util/IntervalList.java
```diff
@@ -397,12 +398,16 @@ public static IntervalList copyOf(final IntervalList list){
...
+        return fromPath(file.toPath());
+    }
+
+    public static IntervalList fromPath(final Path path) {
+        final BufferedReader reader = IOUtil.openFileForBufferedReading(path);
         final IntervalList list = fromReader(reader);
         try {
             reader.close();
         } catch (final IOException e) {
-            throw new SAMException(String.format("Failed to close file %s after reading",file));
```
lbergelson: You might want to use the more ugly path.toUri().toString() because the path.toString() tends to drop the scheme.

Issue #828, opened 2017-03-17
SamLocusIterator.RecordAndOffset.getRefPos always returns -1


nh13:
@ZLyanov could you take a look since you added `AbstractRecordAndOffset`?

It's also weird that `public AbstractRecordAndOffset(final SAMRecord record, final int offset, int length, int refPos)` is never used, even by the constructor in `StartEdgingRecordAndOffset`

ZLyanov:
@nh13 this changes were made as a support for the https://github.com/broadinstitute/picard/pull/556 which is still under review/modification. We will try to finish it ASAP. Meanwhile, you may check the proposed changes to see how these classes and methods are used.

nh13:
@ZLyanov unfortunately the changes have already made it into the master of htsjdk prior to the above change.  My main concern is that folks will call `getRefPos` on a `SamLocusIterator.RecordAndOffset` and always return `-1`, which seems like an oversight of the original commit.  Could you make a PR and link this issue?

tfenne:
Fixed by #849 (2.10.0): `getRefPos()` lives only on `EdgingRecordAndOffset`, which sets it, and the unused constructor was removed from `AbstractRecordAndOffset`.

PR #849, opened 2017-04-10
Fix for SamLocusIterator.RecordAndOffset

kamyshova:
### Fix for issue: https://github.com/samtools/htsjdk/issues/828
In current version of master branch there is a problem with SamLocusIterator.RecordAndOffset class. This class inherits public method getRefPos that always returned -1.

We suggest to remove this method from AbstractRecordAndOffset class and replace it in abstract class EdgingRecordAndOffset, which heritors use this method.


yfarjoun:
simpler is better! 

👍

PR #850, opened 2017-04-11
Fix for filtering of redundant containers

kamyshova:
The current 2.9.1 htsjdk version has dramatically lower performance in contrast with the version 2.2.4 while working with CRAM data on query intervals. As we could see, the one of the reasons is redundant containers processing. We suggest filter out the container in `CRAMIterator` class before its processing if that container doesn't match to query intervals. Our fragmental tests show that these changes increase the performance up to 2-2.5 times. 

However we believe there is a more fundamental and efficient solution, namely to avoid the case that `CRAMFileReader` doesn't work with native CRAI indices (it finds and reads CRAI file but then converts it internally to a BAI stream and, as a result, processes extra containers). We've created the following issue: #851

### DIFF
--- a/src/main/java/htsjdk/samtools/CRAMFileReader.java
+++ b/src/main/java/htsjdk/samtools/CRAMFileReader.java
@@ -484,6 +484,7 @@ public class CRAMFileReader extends SamReader.ReaderImplementation implements Sa
                             getSeekableStreamOrFailWithRTE(),
                             referenceSource,
                             coordinates,
+                            queries,
                             validationStringency
                     );
                 } catch (final IOException e) {
--- a/src/main/java/htsjdk/samtools/CRAMIterator.java
+++ b/src/main/java/htsjdk/samtools/CRAMIterator.java
@@ -35,6 +35,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.math.BigInteger;
 import java.util.ArrayList;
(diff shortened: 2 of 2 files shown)


cmnbroad:
Hi @kamyshova. Sorry for the long delay on this - we're trying to get around to some of these old PRs - and would like to resurrect this one. I've started to take a look at the changes, but have some questions about the underlying premise, which as I understand it is that containers are being unnecessarily processed during index queries due to the use of .bai rather than native .crai.

It would be really helpful to have a test case that demonstrates this, since for both .bai and .crai, the resolution of the index is the `slice` (which is basically the `container`, given 1 slice/container). Its not clear to me that filtering containers at the CRAMIterator level is the right fix - having a test case would help pinpoint the issue. Can you provide a CRAM, index, and query that shows the problem ? Thx.

yfarjoun:
@kamyshova I appologize again for the lengthy delay, are you still interested in following up on this PR?

cmnbroad:
Without having a test that illustrates this (processing redundant containers) its hard to say if this is the correct fix (though I suspect that if there is a perf issue that is fixed by post-filtering containers, the actual issue somewhere in the indexing code, and should be fixed there). At any rate, Im closing this as obsolete since its out of sync with the current implementation, an will soon be replaced by the upcoming CRAM refactoring branch.

PR #853, opened 2017-04-14
Expose a couple of protected methods and replace hard coded strings w…

tfenne:
…ith an enum.

### Description

I have some programs that use `FastqReader` that are failing when they encounter fastqs with empty seq or quality lines.  Since there is no complete formal spec for fastq, it's hard to argue that this is invalid (though is silly IMHO) - and this situation seems to be created by tools that quality trim paired fastq files when one read is all low quality (e.g. R2) and the other (e.g. R1) is fine.

I've made the simplest change possible here, which is to expose the method that asserts the lines are non-empty, so that I can subclass and allow empty Seq and Qual lines, rather than trying to bake that into the reader which would require potentially a lot more changes.

Functionality is unchanged; a couple of methods go from private to protected, and I've replace four hard-coded strings for the line types with an enum so that subclasses don't have to depend on string matching.

### Checklist

- [x] Code compiles correctly
- [ ] All tests passing


tfenne:
@yfarjoun Could I get a 👍  on this little change?

src/main/java/htsjdk/samtools/fastq/FastqReader.java
```diff
@@ -41,6 +41,22 @@
...
+        QualityHeader("Quality Header"),
+        QualityLine("Quality Line");
+
+        private String printable;
+
+        LineType(String printable) {
+            this.printable = printable;
+        }
+
+        @Override public String toString() { return this.printable; }
```
yfarjoun: code-cov shows that this `toString` doesn't get called, at all. 

Can I get a small test from you?

yfarjoun:
Looks good. A tiny test that would actually exercise the newly added toString would be appreciated.
and there are unspaced `+` nearby....perhaps you can give them some air?
👍

src/main/java/htsjdk/samtools/fastq/FastqReader.java
nh13: Seeing the extra requests for tests like this seems like a burden to the submitter, even if the test guards against a regression.  It's a fine line, but I'd be fine without the test.  Just my 2 cents.

nh13: @yfarjoun also see line 191 and 194 that come back red on code-cov.

yfarjoun: I saw that. but on those I chose not to comment. when new functionality fails to be tested, I think that it's warrants a request for a test. the other two lines are not new functionality. Also I suspected that a test of the toString would pass via one one these.

Anyhow, I gave it 👍 ...so @tfenne can chose what he does.

tfenne:
I'm surprised there is any test coverage as I could not find a `FastqReaderTest`.  But apparently the `FastqWriterTest` is really a combined test of both reading and writing.  Sigh.  How about this @yfarjoun - you help me get #789 merged quickly, and then I'll beef up the tests for fastq reading/writing in general, in scala?

yfarjoun:
deal. :+1:

src/test/java/htsjdk/samtools/fastq/FastqReaderWriterTest.scala
```diff
@@ -0,0 +1,121 @@
...
+    out.close()
+    IOUtil.slurpLines(path.toFile).get(1) should have length 1
+  }
+
+  it should "write a record with zero-length bases and quals" in {
+    val path = makeTempFile("test.", ".fastq")
+    val out = new FastqWriterFactory().newWriter(path.toFile)
+    out.write(fq("q1", 0))
+    out.close()
+    IOUtil.slurpLines(path.toFile).get(1) should have length 0
```
yfarjoun: are you only looking at the bases here, or am I misunderstanding something?

src/test/java/htsjdk/samtools/fastq/FastqReaderWriterTest.scala
```diff
@@ -0,0 +1,121 @@
...
+    lines.get(8) shouldBe "@q3"
+    lines.get(9) shouldBe q3.getReadString
+  }
+
+  it should "write a record with only a single base" in {
+    val path = makeTempFile("test.", ".fastq")
+    val out = new FastqWriterFactory().newWriter(path.toFile)
+    out.write(fq("q1", 1))
+    out.close()
+    IOUtil.slurpLines(path.toFile).get(1) should have length 1
```
yfarjoun: as below, shouldn't you also have 
```scala
IOUtil.slurpLines(path.toFile).get(3) should have length 1
```
?

src/test/java/htsjdk/samtools/fastq/FastqWriterTest.java
```diff
@@ -1,74 +0,0 @@
...
-    @DataProvider(name = "fastqsource")
-    public Object[][] createTestData() {
-        return new Object[][]{
-                {"solexa_full_range_as_solexa.fastq"},
-                {"5k-30BB2AAXX.3.aligned.sam.fastq"}
-        };
-    }
-
-    @Test(dataProvider = "fastqsource")
-    public void testReadReadWriteFastq(final String basename) throws Exception {
```
yfarjoun: why eliminate these tests?

src/test/java/htsjdk/samtools/fastq/FastqReaderWriterTest.scala
```diff
@@ -0,0 +1,151 @@
...
+        |AACCGGTT
+        |+
+        |########
+        |@q2
+        |ACGT
+        |####
+      """.stripMargin.trim
+
+    val in = new FastqReader(null, new BufferedReader(new StringReader(fastq)))
+    an[Exception] shouldBe thrownBy { in.next() }
```
yfarjoun: Can you be more specific about the exception here? I was expecting a SamException, but the coverage indicates that none of the `throw new SamException` lines are covered..so I'm confused as to what is being thrown and why...

src/main/java/htsjdk/samtools/fastq/FastqReader.java
```diff
@@ -169,16 +181,18 @@ public void close() {
         }
     }
 
-    private void checkLine(final String line, final String kind) {
+    /** Checks that the line is neither null (representing EOF) or empty (blank line in file). */
+    protected void checkLine(final String line, final LineType kind) {
         if (line == null) {
-            throw new SAMException(error("File is too short - missing "+kind+" line"));
+            throw new SAMException(error("File is too short - missing " + kind));
```
yfarjoun: why isn't this thrown for the truncated file?

src/test/java/htsjdk/samtools/fastq/FastqReaderWriterTest.scala
tfenne: Yes - I'm assuming that if a) the writer didn't explode and b) the sequence line is the write length, we're probably in good shape.

src/test/java/htsjdk/samtools/fastq/FastqReaderWriterTest.scala
tfenne: I think something very funny is going on with codecov.  Both running in IntelliJ, and running `./gradlew test jacocoTestReport` locally show those lines getting covered.  I'm going to try making the other recommended changes, squashing and then opening up a new PR to see if that will get codecov to recognize the changes.

src/test/java/htsjdk/samtools/fastq/FastqWriterTest.java
tfenne: Because:
1. They weren't actually testing anything beyond "doesn't throw an exception".  This is the worst kind of test because it gives the impression of test coverage, but could have been doing totally the wrong thing.
2. They depended on testdata from another package/test, which makes them fairly brittle should those tests ever need to change
3. They are superceded by better tests in the new FastqReaderWriterTest
4. The serialization test I moved into the `FastqRecordTest` where it more properly belongs as it has nothing to do with the FastqWriter!

tfenne:
Closing and will open a new PR to see if that corrects CodeCov.

PR #854, opened 2017-04-16
Expose a couple of protected methods and replace hard coded strings w…

tfenne:
…ith an enum.  Take 2!

### Description

I have some programs that use `FastqReader` that are failing when they encounter fastqs with empty seq or quality lines.  Since there is no complete formal spec for fastq, it's hard to argue that this is invalid (though is silly IMHO) - and this situation seems to be created by tools that quality trim paired fastq files when one read is all low quality (e.g. R2) and the other (e.g. R1) is fine.

I've made the simplest change possible here, which is to expose the method that asserts the lines are non-empty, so that I can subclass and allow empty Seq and Qual lines, rather than trying to bake that into the reader which would require potentially a lot more changes.

Functionality is unchanged; a couple of methods go from private to protected, and I've replace four hard-coded strings for the line types with an enum so that subclasses don't have to depend on string matching.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


tfenne:
Fixed it @yfarjoun - seems like I'd managed to create the new test class in the wrong source root (java vs. scala) and somehow it got compiled and run locally, but not on travis.

yfarjoun:
So the lesson here is that scala-tests should be placed under the scala root and java-tests under java, and also that not doing so seems to skip the incorrectly placed tests, correct?

Should we have a target that explicitly tests that so that we do not get cases like this in the future?

src/test/scala/htsjdk/samtools/fastq/FastqReaderWriterTest.scala
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  it should "write a record with only a single base" in {
+    val path = makeTempFile("test.", ".fastq")
+    val out = new FastqWriterFactory().newWriter(path.toFile)
+    out.write(fq("q1", 1))
+    out.close()
+    val lines = IOUtil.slurpLines(path.toFile)
+    lines.get(1) should have length 1
+    lines.get(3) should have length 1
```
yfarjoun: thanks.

src/test/scala/htsjdk/samtools/fastq/FastqReaderWriterTest.scala
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  it should "write a record with zero-length bases and quals" in {
+    val path = makeTempFile("test.", ".fastq")
+    val out = new FastqWriterFactory().newWriter(path.toFile)
+    out.write(fq("q1", 0))
+    out.close()
+    IOUtil.slurpLines(path.toFile).get(1) should have length 0
```
yfarjoun: here too, just to be sure. also 
```
IOUtil.slurpLines(path.toFile) should have size 4
```

tfenne:
I'm not sure if there's a good way to test for scala files under the Java root.  This was caused (in my case) by IntelliJ trying to be too smart.  If you have multiple source roots (java and scala) and one of them already contains a package and the other doesn't it does funny things when you try to create a new class (either auto-using the one root that has the package, or offering a drop-down that contains only that root).  We could probably try and add something to the build that catches `.scala` files under src/test/java.

yfarjoun:
I think we could add a script that scans for .scala files anywhere except
the test/scala subdir and .java files there. if it finds any it fails. Then
we can add this target to the test targets, so that it will fail tests.

tfenne:
I agree, I’m just not really sure how to do that in Gradle.

PR #858, opened 2017-04-18
removing scripts/release_picard.sh

lbergelson:
this `scripts/release_picard.sh` is no longer used and should be removed


droazen:
:+1: merge when ready

PR #862, opened 2017-04-20
Yf add unknown so

yfarjoun:
### Description

Adds validation to the values in the SO and GO tags in the SAM header.

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMFileHeader.java
+++ b/src/main/java/htsjdk/samtools/SAMFileHeader.java
@@ -24,6 +24,7 @@
 package htsjdk.samtools;
 
 
+import htsjdk.samtools.util.Log;
 import htsjdk.samtools.util.StringLineReader;
 
 import java.io.StringWriter;
@@ -50,6 +51,7 @@ public class SAMFileHeader extends AbstractSAMHeaderRecord
     public static final Set<String> ACCEPTABLE_VERSIONS =
             new HashSet<String>(Arrays.asList("1.0", "1.3", "1.4", "1.5"));
 
+    private static final Log log = Log.getInstance(SAMFileHeader.class);
     /**
      * These tags are of known type, so don't need a type field in the text representation.
      */
@@ -65,8 +67,8 @@ public class SAMFileHeader extends AbstractSAMHeaderRecord
      * Ways in which a SAM or BAM may be sorted.
      */
     public enum SortOrder {
-
         unsorted(null),
+        unknown(null),
         queryname(SAMRecordQueryNameComparator.class),
         coordinate(SAMRecordCoordinateComparator.class),
         duplicate(SAMRecordDuplicateComparator.class); // NB: this is not in the SAM spec!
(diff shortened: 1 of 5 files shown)


src/main/java/htsjdk/samtools/SAMFileHeader.java
```diff
@@ -48,13 +49,14 @@
     public static final String GROUP_ORDER_TAG = "GO";
     public static final String CURRENT_VERSION = "1.5";
     public static final Set<String> ACCEPTABLE_VERSIONS =
-            new HashSet<String>(Arrays.asList("1.0", "1.3", "1.4", "1.5"));
+            new HashSet<>(Arrays.asList("1.0", "1.3", "1.4", "1.5"));
```
tfenne: Well, if you're going to start cleaning up, this could be `CollectionUtil.makeSet(....)`

src/main/java/htsjdk/samtools/SAMFileHeader.java
```diff
@@ -65,8 +67,8 @@
      * Ways in which a SAM or BAM may be sorted.
      */
     public enum SortOrder {
-
         unsorted(null),
+        unknown(null),
```
tfenne: While it's probably not important, it's good practice to add new enum values at the end, in case anyone relied on the previous ordering/ordinal values.

src/main/java/htsjdk/samtools/SAMFileHeader.java
```diff
@@ -107,15 +109,15 @@ public SAMRecordComparator getComparatorInstance() {
...
 
     private List<SAMReadGroupRecord> mReadGroups =
-        new ArrayList<SAMReadGroupRecord>();
-    private List<SAMProgramRecord> mProgramRecords = new ArrayList<SAMProgramRecord>();
+            new ArrayList<>();
+    private List<SAMProgramRecord> mProgramRecords = new ArrayList<>();
     private final Map<String, SAMReadGroupRecord> mReadGroupMap =
-        new HashMap<String, SAMReadGroupRecord>();
-    private final Map<String, SAMProgramRecord> mProgramRecordMap = new HashMap<String, SAMProgramRecord>();
+            new HashMap<>();
```
tfenne: Could hoist back onto the prior line now.

src/main/java/htsjdk/samtools/SAMFileHeader.java
```diff
@@ -253,7 +255,12 @@ public SortOrder getSortOrder() {
         if (so == null || so.equals("unknown")) {
```
tfenne: Should probably get rid of the `so.equals("unknown")` here since that's now a valid value?

src/main/java/htsjdk/samtools/SAMFileHeader.java
```diff
@@ -253,7 +255,12 @@ public SortOrder getSortOrder() {
...
             return SortOrder.unsorted;
         }
-        return SortOrder.valueOf((String) so);
+        try {
+            return SortOrder.valueOf(so);
+        } catch (IllegalArgumentException e) {
+            log.warn("Found non conforming header SO tag: "+ so + ". Treating as 'unknown'.");
```
tfenne: The existing implementation here is poor - the fact that it translates from the String on every access.  I hadn't realized it was doing that.  This also means though, that it's going to log your warning on every access.  What do you think about either changing the class to cache the SortOrder enum value _or_ adding `setAttribute("SO", "unknown")` in the catch?

yfarjoun: I don't like changing the header, as that might have side-effects (the header might be used later to write a file.)

I'll try to cache the resulting enum.

yfarjoun:
merging under the assumption that the minor comments were addressed in my latest PR. @tfenne please let me know if I was being overly presumptuous and I'll be happy to revert and discuss further.

PR #863, opened 2017-04-21
Support Filter For Genotype.getAnyAttribute ()

bbimber:
### Description

I ran into an NPE running GATK's VariantsToTable when requesting the genotype filter field.  They call Genotype.getAnyAttribute("FT").  The current code special-cases all the other values in PRIMARY_KEYS (GT, GQ, DP, AD, etc), but forgets FT.  This is a small addition to include it.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/Genotype.java
+++ b/src/main/java/htsjdk/variant/variantcontext/Genotype.java
@@ -540,6 +540,8 @@ public abstract class Genotype implements Comparable<Genotype>, Serializable {
             return getAlleles();
         } else if (key.equals(VCFConstants.GENOTYPE_QUALITY_KEY)) {
             return getGQ();
+        } else if (key.equals(VCFConstants.GENOTYPE_FILTER_KEY)) {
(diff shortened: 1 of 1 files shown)


yfarjoun:
This is very confusing @bbimber. I think that your fork is not uptodate or something as the reference version of the code already has a special case for GENOTYPE_FILTER_KEY:
https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/variant/variantcontext/Genotype.java#L559

yfarjoun:
even more confusing is that implicated in the line changes is.....you!

yfarjoun:
here's the commit: https://github.com/samtools/htsjdk/commit/6141de3e8885cb3491fef598db19bdf342a94060

yfarjoun:
👎

bbimber:
yes, sorry, my local enlistment is screwed up.  i do remember that initial patch now.  at least it's fixed...

PR #868, opened 2017-04-29
Improved naming and documentation for flag field accessors and mutators.

tfenne:
### Description

The accessor and mutator methods for flag fields in `SAMRecord` have had really long, ugly names forever; names that don't even follow java standards very well.  I feel like I/we should have done this a lot sooner!  I've deprecated all the existing flag field methods and re-routed them to new methods with much cleaner names.  In addition for some flag fields I've added inverted accessors (e.g. `isUnmapped()` and `isMapped()` together replace `getReadUnmappedFlag()`).

I suspect this might  be a little controversial, but since I think we're likely stuck with the existing implementation for SAMRecord it would be nice to start tidying up by improving things and deprecating things we don't like.

While I don't think we want to have a long conversation about the exact names of the new methods, if there are any that really don't sit well with anyone, I'm happy to change them.  I struggled to come up with good names for the replacements for `getReadFailsVendorQualityCheck()` which ended up with `failsQc()` and `passesQc()` since all other phrasings sounded weird to me (e.g. `isFailingQc()`, `isQcFailure()`, etc.).

Lastly, if/when we are happy with this, I will take on the task of finding/replacing calls to all the newly deprecated methods within HTSJDK, so that upon-merge there will be no direct use of the deprecated methods within HTSJDK.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMRecord.java
+++ b/src/main/java/htsjdk/samtools/SAMRecord.java
@@ -862,9 +862,7 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
         }
     }
 
-    /**
-     * It is preferable to use the get*Flag() methods that handle the flag word symbolically.
-     */
+    /** It is preferable to use the get*Flag() methods that handle the flag word symbolically. */
     public int getFlags() {
         return mFlags;
     }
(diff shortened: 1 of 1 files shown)


tfenne:
I feel like this is a pretty significant step, so would like fairly broad review if possible.  @nh13, @yfarjoun, @jacarey, @lbergelson, @droazen: want to share your 2c each?

tfenne:
Also, FYI to any reviewers, my apologies for making the review slightly more awkward that needs be.  The original code had a block of all the getters followed by a block with all the setters.  I've re-arranged to bring the accessor/mutator pair together for each flag, but that makes the diff a little weird looking.

lindenb:
ok for the changes, but please, please keep the java bean nomenclature ! :-)

when injecting a SAMRecord in a _javascript_ , _servlet_ , _velocity_ context, etc... we can _defacto_ use the standard APIs.

`if(record.failingQC) { ... }`

If my voice matters,  I would be happy with:  `getReadFailsVendorQualityCheck` -> `isFailingQC()`  but not with  `getReadFailsVendorQualityCheck` -> `failsQC()` :-)

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -875,233 +873,192 @@ public void setFlags(final int value) {
...
+    public boolean isProperlyPaired() { requireReadPaired(); return getFlag(SAMFlag.PROPER_PAIR); }
 
-    /**
-     * strand of the query (false for forward; true for reverse strand).
-     */
-    public boolean getReadNegativeStrandFlag() {
-        return (mFlags & SAMFlag.READ_REVERSE_STRAND.flag) != 0;
-    }
+    /** Sets whether both reads in the read pair are aligned as expected. */
+    public void setProperlyPaired(final boolean paired) { setFlag(SAMFlag.PROPER_PAIR, paired); }
```
tfenne: I think that ship has sailed, in general - it would be a breaking change to add those kinds of checks into the setter, and I don't want to be the one who does that ;)  More specifically, I think things are the way they are so that records can be cleaned up without having to worry about ordering.  E.g. if you want to make a record "unpaired", you don't have to worry about calling `setProperlyPaired(false)` before `setPaired(false)` etc.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -875,233 +873,192 @@ public void setFlags(final int value) {
...
-        } else {
-            mFlags &= ~bit;
-        }
+    /** @deprecated Since May 2017; use {@link #isSupplementaryAlignment()} instead. */
+    @Deprecated public boolean getSupplementaryAlignmentFlag() { return isSupplementaryAlignment(); }
+    /** @deprecated Since May 2017; use {@link #()} instead. */
+    @Deprecated public void setSupplementaryAlignmentFlag(final boolean flag) { setSupplementaryAlignment(flag); }
+
+    /** Returns true if the read fails vendor quality checks. */
+    public boolean failsQc() { return getFlag(SAMFlag.READ_FAILS_VENDOR_QUALITY_CHECK); }
```
tfenne: That works for me.  I'll change it.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -875,233 +873,192 @@ public void setFlags(final int value) {
...
+    public boolean isPaired() { return getFlag(SAMFlag.READ_PAIRED); }
 
-    /**
-     * the read is mapped in a proper pair (depends on the protocol, normally inferred during alignment).
-     */
-    public boolean getProperPairFlag() {
-        requireReadPaired();
-        return getProperPairFlagUnchecked();
-    }
+    /** Sets the "read paired in sequencing" flag to the given value. */
```
tfenne: I've taken a pass through and tried to be more consistent.  Am happy to take suggestions for anywhere you think could still be improved.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -863,245 +863,211 @@ public SAMReadGroupRecord getReadGroup() {
...
+    /** Sets the first of pair flag. */
+    public void setFirstOfPair(final boolean first) { setFlag(SAMFlag.FIRST_OF_PAIR, first); }
+
+    /** @deprecated Since May 2017; use {@link #isFirstOfPair()} instead. */
+    @Deprecated public boolean getFirstOfPairFlag() { return isFirstOfPair(); }
+    /** @deprecated Since May 2017; use {@link #setFirstOfPair(boolean)} instead. */
+    @Deprecated public void setFirstOfPairFlag(final boolean flag) { setFirstOfPair(flag); }
+
+
+    /** Returns true if the second of pair flag is set, false otherwise. Illegal to call on unpaired reads. */
```
yfarjoun: "...if the read is second in a pair,...."

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -863,245 +863,211 @@ public SAMReadGroupRecord getReadGroup() {
...
+    public boolean isSecondOfPair() { requireReadPaired(); return getFlag(SAMFlag.SECOND_OF_PAIR); }
+
+    /** Sets the second of pair flag. */
```
yfarjoun: "Sets whether the read is second in a pair"

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -863,245 +863,211 @@ public SAMReadGroupRecord getReadGroup() {
...
-        setFlag(flag, SAMFlag.SUPPLEMENTARY_ALIGNMENT.flag);
-    }
 
-    /**
-     * the read fails platform/vendor quality checks.
-     */
-    public void setReadFailsVendorQualityCheckFlag(final boolean flag) {
-        setFlag(flag, SAMFlag.READ_FAILS_VENDOR_QUALITY_CHECK.flag);
-    }
+    /** Returns true if the first of pair flag is set, false otherwise. Illegal to call on unpaired reads. */
```
yfarjoun: "...if the read is the first in a pair,...."

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -863,245 +863,211 @@ public SAMReadGroupRecord getReadGroup() {
...
+    public boolean isFirstOfPair() { requireReadPaired(); return getFlag(SAMFlag.FIRST_OF_PAIR); }
+
+    /** Sets the first of pair flag. */
```
yfarjoun: "Sets whether the read is first in a pair."

lbergelson:
@tfenne It might make sense to make these changes in the context of introducing a new read interface.  (We happen to have a new read interface that we think is pretty good https://github.com/broadinstitute/gatk/blob/master/src/main/java/org/broadinstitute/hellbender/utils/read/GATKRead.java)

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -863,245 +863,262 @@ public SAMReadGroupRecord getReadGroup() {
...
+    @Deprecated public boolean getFirstOfPairFlag() { return isFirstOfPair(); }
+    /** @deprecated Since May 2017; use {@link #setFirstOfPair(boolean)} instead. */
+    @Deprecated public void setFirstOfPairFlag(final boolean flag) { setFirstOfPair(flag); }
+
 
     /**
-     * strand of the mate (false for forward; true for reverse strand).
+     * True if the read is the second read in a read pair, false otherwise. Illegal to call on unpaired reads.
+     * Returns the value of flag bit 0x80.
      */
```
lbergelson: If we're already making changes changes we might want to update these to match the language in the sam spec.  I.e. firstSegment, lastSegment.

droazen:
@tfenne Completely agree with @lbergelson -- it would be extremely useful to finally have a `Read` interface in htsjdk! If we could propose these naming adjustments in the context of a discussion around a new `Read` interface, without changing `SAMRecord` proper (yet), it would be less immediately disruptive to downstream code, and we'd likely end up with something less constrained by the existing structure of the `SAMRecord` class.

tfenne:
@lbergelson & @droazen thanks for the feedback.  I'm trying to understand the goal of introducing a `Read` interface above `SAMRecord` in htsjdk.  I'm not against it, but don't yet understand what it would buy us.  Specifically, I think the interface would have to be constrained somewhat by not requiring breaking changes in `SAMRecord` - which may or may not make it palatable to you vs. your approach in GATK.  Also, unless there are multiple SAM-like things that satisfy the interface, it seems like overhead to have an interface and a single implementation.  Would the goal be to have other implementations such as the google genomics one in GATK make their way down in htsjdk?

I did take a look at the linked `GATKRead` interface, and note that most of the names given to flag-like methods either match or are pretty close to where I ended up 😄 .  But there are also constraints on that interface that would be hard to impose retro-actively on `SAMRecord` - e.g. the deep copying of any mutable structure returned.  Still, I could imagine an interface in htsjdk that doesn't impose those constraints, that is sub-typed in the GATK to impose them - it would still require either extending or wrappering `SAMRecord` there though.  Lastly, it's a small thing, but the `setIs*` are somewhat unconventional - for booleans the JavaBean standard is to have `isX` and `setX`, not `setIsX`.  Are you open to changing things like that for an interface in htsjdk?

I don't think the changes proposed are immediately disruptive - the old methods are deprecated, but not removed.  That said, changing them multiple times to align with a future interface would be more disruptive than doing it all at once.

My goal is to clean up some of the cruft in SAMRecord and make interacting with nicer, without the need to wrapper it since I'm somewhat afraid of the performance overhead of putting another level of indirection between application code and reads.  I'd also like, over time, to do things like get rid of `indexingBin` and compute it on the fly, and investigate improving the storage, performance and access to extended attributes.  If an interface is helpful, I'm open to that.  But I'd rather that not lead this down a path that takes us months to make these kinds of changes.

droazen:
@tfenne Deprecating these methods in `SAMRecord` directly would be immediately disruptive to us -- it would break all of our builds, which treat use of deprecated code as an error (a policy which has proved necessary to prevent usages of deprecated code from accumulating and persisting over time).

We've had a need for a Read (and Variant!) interface in htsjdk for a long time -- having these interfaces in the GATK, and then having to adapt or convert the underlying htsjdk types has been painful for us. It would be great if we could take this opportunity to finally design a Read interface in htsjdk that we are all happy with. The design process need not take months!

I'm sure I don't need to convince you of the advantages and value of coding against interfaces rather than concrete types in a general sense. The specific use case that prompted us to create the `GATKRead` interface was the need to deal with GA4GH data -- the records returned by GA4GH queries were sufficiently different from `SAMRecord` that we needed an interface with clear, well-defined semantics not tied to any particular file format to write code against.

yfarjoun:
@droazen I'm not sure that idiosyncrasies of the GATK4 build can be used as an argument to increase the scope of a PR, or significantly slow it's acceptance. We should be able to accept/reject a PR on its merits and make the build systems/CI we use in projects that depend on htsjdk flexible enough to handle the occasional deprecation. Heaven knows we cause this pain to others....

Regarding the the specific ideas being floated around. 

1. Is the GATKRead interface close enough to being the Read interface we would like to have? 
2. Are the changes proposed here compatible with the eventual interface? 
3. I think that enabling various sources of reads to be used interchangeably in htsjdk and libraries that use it would be a positive move and so I support the creation of an interface that would support that. However, that might indeed prove to be a longer effort than what is needed to get this PR though. 

Can we agree on the names/javadoc of the methods in this PR and open a separate issue dedicated to the discussion of the read interface?

src/main/java/htsjdk/samtools/SAMRecord.java
yfarjoun: hmmm. While first/last is indeed the language of the sam-spec, the vast majority of the actual usage pair-ended reads. If we are to go this route (of pretending to have more then 2 reads), we would also need to avoid the use of the term "mate" and start using "next" everywhere...I don't think that the confusion that would cause is worth the adherance to the language in the spec. What we can do is specify in the javadoc precisely what flag is being accessed here (and in other places...)

magicDGS:
I think that the `Read` interface is something nice to add, but more in the scope of a new API (https://github.com/samtools/htsjdk/issues/520). It will also include the `Variant` interface, and I'm open to contribute to that one.

What's about doing something similar as in GATK and create a new HTSJDK repository for version 3, including interfaces and solving known problems with the API? New issues could be back-ported in the meanwhile, and the new SemVer system could be apply from version 3 onwards.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -863,245 +863,262 @@ public SAMReadGroupRecord getReadGroup() {
...
+    @Deprecated public boolean getDuplicateReadFlag() { return isDuplicate(); }
+    /** @deprecated Since May 2017; use {@link #setDuplicate(boolean)} instead. */
+    @Deprecated public void setDuplicateReadFlag(final boolean flag) { setDuplicate(flag); }
+
+
+    /** Returns true if the record represents either a secondary or supplementary alignment. */
+    public boolean isSecondaryOrSupplementary() { return isSecondaryAlignment() || isSupplementaryAlignment(); }
+
+    /** Sets a specific flag bit to the provided value. */
+    private void setFlag(final SAMFlag flag, final boolean value) {
```
magicDGS: To have more control and do not rely on the new syntax, could this and the getter be public?

droazen:
@yfarjoun If we do this rename now, and then introduce a Read interface later, we'll have to make two passes through our code base to migrate method calls instead of one. In the past, as a courtesy to downstream projects like IGV, Picard, and GATK, we've made an effort to try to minimize such unnecessary pain. And to be clear, this change likely affects thousands or tens of thousands of lines of code in GATK alone, so this is on a somewhat different level from typical htsjdk breakage :)  Also, consider this: if we did this rename now, and then wanted different names in the final `Read` interface, for whatever reason, we'd then have to do a *second* round of renaming/deprecation in `SAMRecord` itself when we make it implement that interface, which would be gross.

It also seems more productive and ultimately useful to work together to design a generalized `Read` interface agnostic to file format than to just fiddle with method names in `SAMRecord`. `GATKRead` could act as a starting point for that discussion, but we are very open to changes!

yfarjoun:
well. I guess that part I disagree with is the "have to". You could also change GATK's build system (temporarily) to allow deprecated calls.

droazen:
@yfarjoun Not really a viable option for us, given the number of our dependencies and the frequency with which APIs change out from under us (particularly on the Spark side of things).

But there's a bigger problem here: if we rename a bunch of methods in `SAMRecord`, and then separately design a `Read` interface, we might have to do a second round of renaming in `SAMRecord` to make it agree with the names we settle on for the interface. This is why it makes little sense to me to de-couple the tasks of improving naming in `SAMRecord` from introducing a `Read` interface.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -863,245 +863,262 @@ public SAMReadGroupRecord getReadGroup() {
...
-        return (mFlags & SAMFlag.READ_PAIRED.flag) != 0;
-    }
+    public boolean isPaired() { return getFlag(SAMFlag.READ_PAIRED); }
 
-    private void requireReadPaired() {
-        if (!getReadPairedFlag()) {
-            throw new IllegalStateException("Inappropriate call if not paired read");
-        }
-    }
+    /** Sets whether or not the read is part of a read pair. Sets flag bit 0x1. */
```
nh13: Sets _the_, or "Sets the 0x1 flag bit"

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -863,245 +863,262 @@ public SAMReadGroupRecord getReadGroup() {
...
+    public void setPaired(boolean paired) { setFlag(SAMFlag.READ_PAIRED, paired); }
```
nh13: final

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -863,245 +863,262 @@ public SAMReadGroupRecord getReadGroup() {
...
-    }
+    public final boolean isMapped() { return !isUnmapped(); }
 
-    private boolean getFirstOfPairFlagUnchecked() {
-        return (mFlags & SAMFlag.FIRST_OF_PAIR.flag) != 0;
+    /** Sets the read's unmapped flag. Sets flag bit 0x4. */
+    public void setUnmapped(final boolean unmapped) {
+        setFlag(SAMFlag.READ_UNMAPPED, unmapped);
+        setIndexingBin(null);
     }
```
nh13: It would be nice to have a `setMapped` method too.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -863,245 +863,262 @@ public SAMReadGroupRecord getReadGroup() {
...
+    public final boolean isMateMapped() { return !isMateUnmapped(); }
 
-    /**
-     * the read fails platform/vendor quality checks.
-     */
-    public boolean getReadFailsVendorQualityCheckFlag() {
-        return (mFlags & SAMFlag.READ_FAILS_VENDOR_QUALITY_CHECK.flag) != 0;
-    }
+    /** Sets whether the read's mate is unmapped. Sets flag bit 0x8. */
+    public void setMateUnmapped(final boolean unmapped) { setFlag(SAMFlag.MATE_UNMAPPED, unmapped); }
```
nh13: ditto about a `setMateMapped`method.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -863,245 +863,262 @@ public SAMReadGroupRecord getReadGroup() {
...
+     * True if the read is mapped to the positive strand of the genome, false otherwise.
+     * Returns the negatation of flag bit 0x10.
      */
-    @Deprecated
-    public void setReadUmappedFlag(final boolean flag) {
-        setReadUnmappedFlag(flag);
+    public final boolean isPositiveStrand() { return !isNegativeStrand(); }
+
+    /** Sets whether the read is mapped to the negative strand of the genome. Sets flag bit 0x10. */
+    public void setNegativeStrand(final boolean negative) { setFlag(SAMFlag.READ_REVERSE_STRAND, negative); }
```
nh13: I don't know why, but I want a `setPositiveStrand` too.

nh13: I think it's for symmetry.

magicDGS:
I repeat my suggestion about creating a non-backwards compatible API and ask to the community for a clean-up effort of the library. Doing that, this PR could get in and the discussion about what is a good `Read` interface could be moved to the new htsjdk3 repository. I'm in for contribution!

droazen:
@nh13 Well, it's not really practical to push down alternate read implementations into htsjdk, given how strictly we control project dependencies here (eg., I assume you would not want to see a new Google Genomics dependency added to htsjdk, right?). But the fact is that alternate read representations do exist, have been standardized, and the lack of a `Read` interface in htsjdk is a big problem for projects that have to deal with these. We end up having to write all sorts of adapter classes to interface with htsjdk code.

yfarjoun:
@tfenne I think that this is totally a good direction and I too would love to have better API for sam-record. I also think that a Read API is a good way forward and would hate to change the names twice. I think we are all on the same page. The only question is timing. As we are working towards a presentation in BioIT we are too short-staffed to deal with the breakage that would result (due to the way gatk4 is currently set-up) but would love to be part of designing the Read API. There are important decisions to be made (for example regarding using first / mate / second versus first / next / last ) and we do not have the bandwidth right now to give it the attention it deserves. 

Can we agree to put this on hold for about a month and then hold a day-long meeting where we will work on the new API including changing names? We're happy to host. 

I hope I was able to convey our reluctance to making these changes **right now** and at the same time, our full commitment to getting to them in the near future, and not letting them languish like some htsjdk PRs...

tfenne:
@yfarjoun: I appreciate your candour.  As you know @nh13 and I do most of our development in scala these days, and we discussed two options.  The first was to do what this PR is trying to do - i.e. evolve HTSJDK APIs, in this case `SAMRecord`, in a better direction.  We feel like this is both a better option for the community at large, and also more effort for us.  The other option is to encapsulate our use of HTSJDK classes and provide ourselves with a better API in scala.  The end result of that option is actually a better result for us, but doesn't help any other HTSJDK users.

Waiting a month to have the discussion about starting a longer process of specifying a super-interface for `SAMRecord` is likely not compatible with our plans.  The best answer I can give is that you should ask again in a month, but likely we'll have moved on by then.  I'm not even fully bought into the idea that there should be a `Read` interface in HTSJDK unless either a) there is more than one implementation or b) it's a more radical departure and is not based on [paving the cow path](https://english.stackexchange.com/questions/44800/what-does-don-t-pave-the-cow-path-mean-in-this-context) that is the current SAMRecord API give or take some re-naming.  I.e. if we're going to introduce an interface we should be open to broader change as @magicDGS suggests.

I also echo your own earlier comment that it should not be encumbent on htsjdk to deal with the consequences of choices made in downstream projects.  If htsjdk were to be completely beholden to the GATK development team's choices that would be a real problem.  Treating use of deprecated code as compile breaks is a choice, but should not mean that we now have to treat deprecation as a breaking change in HTSJDK.  If that were true there would be literally no way to evolve the library!  In addition, nobody is forcing GATK to use the latest HTSJDK snapshot - if if that's disruptive why not simply pin the dependency at 2.9.1 until you're ready?  Also, how is it that GATK is not insulated from changes in SAMRecord via it's use of [`GATKRead`](https://github.com/broadinstitute/gatk/blob/master/src/main/java/org/broadinstitute/hellbender/utils/read/GATKRead.java) and [`SAMRecordToGATKReadAdapter`](https://github.com/broadinstitute/gatk/blob/master/src/main/java/org/broadinstitute/hellbender/utils/read/SAMRecordToGATKReadAdapter.java)?

droazen:
@tfenne Surely changes to a core class like `SAMRecord` require some coordination among stakeholders and maintainers of the project -- and if one stakeholder would be inconvenienced and requests that a change be delayed by a few weeks, in the spirit of collegiality I'd hope that request would be honored unless the change is particularly urgent. "Making method names nicer" does not seem to me to rise to that level of urgency. I feel like there have been many occasions when we've been asked to hold back much more substantial changes for weeks or months out of deference to the Picard project.

tfenne:
@yfarjoun Thanks for the suggestion.  I honestly didn't think this PR was going to generate as much discussion as it did.  My goal here was to spruce up the API to SAMRecord here, and then create custom sub-classes in fgbio for use with a `SAMRecordFactory`.

The names I picked here are already something of a compromise w.r.t. idiomatic scala.  Since the PR generated the response it did, I spent a bit more time on the scala side and have found a neat way of totally encapsulating the API to SAMRecord _without_ having to wrapper it(!).  For anyone interested (it's all in scala), the prototype of that work is [here](https://github.com/fulcrumgenomics/fgbio/pull/219/files#diff-c9293df9e63c77546b186e59188ff82dR33).

Having done that, I'm not much less dependent on the particulars of the SAMRecord API and am happy to wait.  I can either leave the PR open, or close the PR and leave the branch up here, so it's available as input to the future discussion.  Thoughts on which is preferable?

yfarjoun:
I would prefer that you kept it open and it would serve as the beginning of
a discussion for the HtsJdkRead interface (or whatever we will call it...)
 I'm glad that you have a solution that works for you. (if you close it,
the branch could be deleted to easily...)

I hope that you'll be involved in a few weeks when we would like to start
designing the interface.

yfarjoun:
can this PR be closed?

magicDGS:
I guess so. This would be a addressed in the [htsjdk-next-beta `Read` interface](https://github.com/samtools/htsjdk-next-beta/issues/1)

PR #872, opened 2017-05-06
Upgraded snappy-java to the latest released version.

tfenne:
### Description

Upgrade snappy-java to the latest released version.  The version in HTSJDK was no longer working on macs as best I can tell (I only have one to test on).  The new version appears to work fine on my mac, and also on a single linux machine I tested on.

I've also added the option in SnappyLoader to log failures but be quiet on successes.

I also manually tested the loader with and without native libraries available, and it works as expected, which is nice as I was able to simplify quite a bit.

### Checklist

- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/SnappyLoader.java
+++ b/src/main/java/htsjdk/samtools/util/SnappyLoader.java
@@ -24,27 +24,19 @@
 package htsjdk.samtools.util;
 
 import htsjdk.samtools.SAMException;
-import org.xerial.snappy.LoadSnappy;
 import org.xerial.snappy.SnappyInputStream;
+import org.xerial.snappy.SnappyOutputStream;
 
+import java.io.ByteArrayOutputStream;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.lang.reflect.Constructor;
 
 /**
  * If Snappy is available, obtain single-arg ctors for SnappyInputStream and SnappyOutputStream.
  */
 public class SnappyLoader {
     private static final int SNAPPY_BLOCK_SIZE = 32768;  // keep this as small as can be without hurting compression ratio.
-    private final Constructor<InputStream> SnappyInputStreamCtor;
-    private final Constructor<OutputStream> SnappyOutputStreamCtor;
     public final boolean SnappyAvailable;
-
-    // Force Snappy-java code to be loaded into executable jars.
-    private final SnappyInputStream ignoreMe = null;
-
-    // Force bcel to load Snappy.
-    //private static final Class SnappyClass = SnappyInputStream.class;
 
     private static final boolean DefaultVerbosity = Boolean.valueOf(System.getProperty("snappy.loader.verbosity", "false"));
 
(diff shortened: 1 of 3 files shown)


src/main/java/htsjdk/samtools/util/SnappyLoader.java
```diff
@@ -57,71 +49,47 @@ public SnappyLoader() {
...
         }
         else {
+            boolean tmpSnappyAvailable = false;
             try {
-                final Class<InputStream> snappyInputStreamClass = (Class<InputStream>)Class.forName("org.xerial.snappy.SnappyInputStream");
-                final Class<OutputStream> snappyOutputStreamClass = (Class<OutputStream>)Class.forName("org.xerial.snappy.SnappyOutputStream");
-                snappyErrorClass = (Class<Error>)Class.forName("org.xerial.snappy.SnappyError");
-                inputStreamCtor = snappyInputStreamClass.getConstructor(InputStream.class);
-                outputStreamCtor = snappyOutputStreamClass.getConstructor(OutputStream.class, Integer.TYPE);
+                final OutputStream test = new SnappyOutputStream(new ByteArrayOutputStream(1000));
```
yfarjoun: You actually have to try to compress something in-order to tests that snappy exists? what step here would fail if it doesn't?

src/main/java/htsjdk/samtools/util/SnappyLoader.java
```diff
@@ -24,28 +24,20 @@
...
  * If Snappy is available, obtain single-arg ctors for SnappyInputStream and SnappyOutputStream.
  */
 public class SnappyLoader {
     private static final int SNAPPY_BLOCK_SIZE = 32768;  // keep this as small as can be without hurting compression ratio.
-    private final Constructor<InputStream> SnappyInputStreamCtor;
-    private final Constructor<OutputStream> SnappyOutputStreamCtor;
     public final boolean SnappyAvailable;
 
-    // Force Snappy-java code to be loaded into executable jars.
-    private final SnappyInputStream ignoreMe = null;
```
yfarjoun: Is this required for builds that have automatic cleaning of "unused" libraries? I think that GATK does something like this... @lbergelson ?

tfenne: @yfarjoun - this used to be the case when a) downstream projects tried to be "smart" about only including necessary classes in their fat jars instead of just including all dependencies and b) the code in `SnappyLoader` didn't explicitly reference any classes from the snappy-java library.  With this latest release it works to just attempt to use the snappy-java classes, whereas (I think) in the old release we were using, you would get `ClassNotFoundError`s because static initializers would fail if the native libraries could not be loaded.

src/main/java/htsjdk/samtools/util/SnappyLoader.java
tfenne: Previously we tried to be smart about this, but I tend to think the best way to test if snappy compression is available is to actually try and compress something.  It's very hard to test though, since we'd need to be able to fabricate a test that _doesn't_ have access to the native libraries.  My expectation is that the process of opening a stream, writing to it and the closing it should be sufficient.

tfenne:
@lbergelson Any chance you can take a look at this some time soon?

src/main/java/htsjdk/samtools/util/SnappyLoader.java
```diff
@@ -57,71 +49,47 @@ public SnappyLoader() {
...
-        Class<Error> snappyErrorClass = null;
+        this(verbose, verbose);
+    }
 
+    /**
+     * Constructs a new SnappyLoader which will check to see if snappy is available in the JVM/library path.
+     * @param logFailures  if true output a message to stderr if snappy loading fails
+     * @param logSuccesses if true output a message to stderr if snappy loading succeeds
+     */
+    public SnappyLoader(final boolean logFailures, final boolean logSuccesses) {
```
lbergelson: Do we really need this level of configurability of error messages?  It seems like this should be using the logger instead of printing directly to standard.err anyway.

src/main/java/htsjdk/samtools/util/SnappyLoader.java
```diff
@@ -57,71 +49,47 @@ public SnappyLoader() {
...
-                    tmpSnappyAvailable = true;
-                }
-            } catch (Error e) {
-                if (e.getClass().equals(snappyErrorClass)) {
-                    if (verbose) System.err.println("Snappy dll failed to load: " + e.getMessage());
-                    tmpSnappyAvailable = false;
-                } else {
-                    throw e;
-                }
+            catch (Throwable e) {
```
lbergelson: I'm always worried about catching `Throwables` and eating them.  If we could enumerate the expected `Errors` that could be thrown by failure to load native code and then catch those we'd avoid potentially catching unrelated fatal errors.  `UnsatisfiedLinkError` and `SnappyError` comes to mind as the most obvious but I'm not sure if there are others.

src/main/java/htsjdk/samtools/util/SnappyLoader.java
```diff
@@ -57,71 +49,47 @@ public SnappyLoader() {
...
+
     /** Wrap an InputStream in a SnappyInputStream. If Snappy is not available will throw an exception. */
     public InputStream wrapInputStream(final InputStream inputStream) {
-        try {
-            return SnappyInputStreamCtor.newInstance(inputStream);
-        } catch (Exception e) {
-            throw new SAMException("Error instantiating SnappyInputStream", e);
-        }
+        try { return new SnappyInputStream(inputStream); }
+        catch (Exception e) { throw new SAMException("Error instantiating SnappyInputStream", e); }
```
lbergelson: I think this should probably check`isSnappyAvailable` and throw if it's false.  It looks like maybe it used to check it implicitly by getting a `NullPointerException` if the constructor wasn't initialized.   As this is now you could specify `snappy.disable` on the command line and then end up using snappy unintentionally if you didn't guard against it in your own code.

src/main/java/htsjdk/samtools/util/SnappyLoader.java
```diff
@@ -24,28 +24,20 @@
...
-import java.lang.reflect.Constructor;
 
 /**
  * If Snappy is available, obtain single-arg ctors for SnappyInputStream and SnappyOutputStream.
  */
 public class SnappyLoader {
     private static final int SNAPPY_BLOCK_SIZE = 32768;  // keep this as small as can be without hurting compression ratio.
-    private final Constructor<InputStream> SnappyInputStreamCtor;
-    private final Constructor<OutputStream> SnappyOutputStreamCtor;
     public final boolean SnappyAvailable;
```
lbergelson: Probably best to make this private since we're adding an accessor.  A breaking change, but not exactly a hard one to repair.

src/main/java/htsjdk/samtools/util/SnappyLoader.java
```diff
@@ -57,71 +49,47 @@ public SnappyLoader() {
...
+        this(verbose, verbose);
+    }
 
+    /**
+     * Constructs a new SnappyLoader which will check to see if snappy is available in the JVM/library path.
+     * @param logFailures  if true output a message to stderr if snappy loading fails
+     * @param logSuccesses if true output a message to stderr if snappy loading succeeds
+     */
+    public SnappyLoader(final boolean logFailures, final boolean logSuccesses) {
         if (java.lang.Boolean.valueOf(System.getProperty("snappy.disable", "false"))) {
```
lbergelson: While we're here, could we expose this as a public constant?

src/main/java/htsjdk/samtools/util/SnappyLoader.java
```diff
@@ -57,71 +49,47 @@ public SnappyLoader() {
...
     }
 
     /** Wrap an InputStream in a SnappyInputStream. If Snappy is not available will throw an exception. */
     public OutputStream wrapOutputStream(final OutputStream outputStream) {
-        try {
-            return SnappyOutputStreamCtor.newInstance(outputStream, SNAPPY_BLOCK_SIZE);
-        } catch (Exception e) {
-            throw new SAMException("Error instantiating SnappyOutputStream", e);
-        }
+        try { return new SnappyOutputStream(outputStream, SNAPPY_BLOCK_SIZE); }
```
lbergelson: same comment as above

src/main/java/htsjdk/samtools/util/SnappyLoader.java
```diff
@@ -57,71 +49,47 @@ public SnappyLoader() {
...
+     */
+    public SnappyLoader(final boolean logFailures, final boolean logSuccesses) {
         if (java.lang.Boolean.valueOf(System.getProperty("snappy.disable", "false"))) {
             System.err.println("Snappy is disabled via system property.");
+            SnappyAvailable = false;
         }
         else {
+            boolean tmpSnappyAvailable = false;
             try {
-                final Class<InputStream> snappyInputStreamClass = (Class<InputStream>)Class.forName("org.xerial.snappy.SnappyInputStream");
```
lbergelson: Good riddance to this.

src/main/java/htsjdk/samtools/util/SnappyLoader.java
lbergelson: 👍 to removing this.  GATK does not rely on it.

yfarjoun:
friendly ping to respond to comments. we'd love to merge this in!

tfenne:
@yfarjoun Yes - it's on my list.  I'd also like to get this merged.  I'll get to it soon-ish.

jacarey:
+1 for merging this. I've tracked down a memory leak in the version of snappy we are using. This is causing issues for use when we have lots of sorting collection temp files (since we wrap each one)

lbergelson:
These changes look good to me 👍  I have a few small comments, but I think I'll open them in a new PR rather than block this one in a confusing state of ownership.

lbergelson:
Thanks @tfenne @danxmoran @jacarey.

PR #877, opened 2017-05-22
Adding an overlaps function to IntervalList.

nh13:
Finds all intervals in the first list that overlap any interval in the
second list.

@yfarjoun  would you have a chance to review?  I know @tfenne is on vacation this week and I know you are familiar with this code.  Otherwise, I can ask @lbergelson or @jacarey to review?

Relates to https://github.com/nh13/picard/pull/7

### DIFF
--- a/src/main/java/htsjdk/samtools/util/IntervalList.java
+++ b/src/main/java/htsjdk/samtools/util/IntervalList.java
@@ -739,6 +739,59 @@ public class IntervalList implements Iterable<Interval> {
                 subtract(lists2, lists1));
     }
 
+    /**
+     * A utility function for finding the intervals in the first list that have at least 1bp overlap with any interval
+     * in the second list.
+     *
+     * @param lhs the first collection of IntervalLists
+     * @param lhs the second collection of IntervalLists
+     * @return an IntervalList comprising of all intervals in the first IntervalList that have at least 1bp overlap with
+     * any interval in the second.
+     */
+    public static IntervalList overlaps(final IntervalList lhs, final IntervalList rhs) {
+        return overlaps(Collections.singletonList(lhs), Collections.singletonList(rhs));
+    }
+
+    /**
+     * A utility function for finding the intervals in the first list that have at least 1bp overlap with any interval
+     * in the second list.
+     *
+     * @param lists1 the first collection of IntervalLists
+     * @param lists2 the second collection of IntervalLists
+     * @return an IntervalList comprising of all intervals in the first collection of lists that have at least 1bp
+     * overlap with any interval in the second lists.
+     */
+    public static IntervalList overlaps(final Collection<IntervalList> lists1, final Collection<IntervalList> lists2) {
+        final SAMFileHeader header = lists1.iterator().next().getHeader().clone();
+        header.setSortOrder(SAMFileHeader.SortOrder.unsorted);
+
+        // Create an overlap detector on list2
+        final IntervalList overlapIntervals = new IntervalList(header);
+        for (final IntervalList list : lists2) {
+            SequenceUtil.assertSequenceDictionariesEqual(header.getSequenceDictionary(),
+                    list.getHeader().getSequenceDictionary());
+            overlapIntervals.addall(list.getIntervals());
+        }
+        final OverlapDetector<Interval> detector = new OverlapDetector<>(0, 0);
+        for (final Interval interval : overlapIntervals.sorted().uniqued()) {
+            detector.addLhs(interval, interval);
+        }
+
+        // Go through each input interval in in lists1 and see if overlaps any interval in lists2
+        final IntervalList merged = new IntervalList(header);
+        for (final IntervalList list : lists1) {
+            SequenceUtil.assertSequenceDictionariesEqual(header.getSequenceDictionary(),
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/util/IntervalList.java
```diff
@@ -739,6 +739,59 @@ public static IntervalList difference(final Collection<IntervalList> lists1, fin
...
+     * A utility function for finding the intervals in the first list that have at least 1bp overlap with any interval
+     * in the second list.
+     *
+     * @param lists1 the first collection of IntervalLists
+     * @param lists2 the second collection of IntervalLists
+     * @return an IntervalList comprising of all intervals in the first collection of lists that have at least 1bp
+     * overlap with any interval in the second lists.
+     */
+    public static IntervalList overlaps(final Collection<IntervalList> lists1, final Collection<IntervalList> lists2) {
+        final SAMFileHeader header = lists1.iterator().next().getHeader().clone();
```
yfarjoun: will break on empty list....but then you'll have no header to work with...you could then grab the header from an intervalList in lists2...but that might be empty as well.. 

Do you really need Collections of IntervalLists? Wouldn't it be better to make this robust on `overlaps(final IntervalList lhs, final IntervalList rhs) ` and have the caller addAll outside?

src/main/java/htsjdk/samtools/util/IntervalList.java
```diff
@@ -739,6 +739,59 @@ public static IntervalList difference(final Collection<IntervalList> lists1, fin
...
+        // Create an overlap detector on list2
+        final IntervalList overlapIntervals = new IntervalList(header);
+        for (final IntervalList list : lists2) {
+            SequenceUtil.assertSequenceDictionariesEqual(header.getSequenceDictionary(),
+                    list.getHeader().getSequenceDictionary());
+            overlapIntervals.addall(list.getIntervals());
+        }
+        final OverlapDetector<Interval> detector = new OverlapDetector<>(0, 0);
+        for (final Interval interval : overlapIntervals.sorted().uniqued()) {
+            detector.addLhs(interval, interval);
```
yfarjoun: can put the same (dummy) Object in the second position of addLhs. Might be more efficient that way?

src/test/java/htsjdk/samtools/util/IntervalListTest.java
```diff
@@ -377,12 +377,83 @@ public void testSubtractSingletonIntervalLists(final IntervalList fromLists, fin
...
 
+    @DataProvider(name = "overlapsSingletonData")
+    public Object[][] overlapSingletonData() {
+        final IntervalList two_overlaps_one   = new IntervalList(fileHeader);
+        final IntervalList three_overlaps_two = new IntervalList(fileHeader);
+        final IntervalList three_overlaps_one = new IntervalList(fileHeader);
+        final IntervalList one_overlaps_three = new IntervalList(fileHeader);
+
+        two_overlaps_one.add(new Interval("1", 50, 150));
+        //two_overlaps_one.add(new Interval("1", 301, 500));
```
yfarjoun: why the commented lines?

src/test/java/htsjdk/samtools/util/IntervalListTest.java
```diff
@@ -377,12 +377,83 @@ public void testSubtractSingletonIntervalLists(final IntervalList fromLists, fin
...
+                new Object[]{list1, list1, list1}, // should return itself
+                new Object[]{list1, IntervalList.invert(list1), new IntervalList(list1.getHeader())}, // should be empty
+                new Object[]{list2, list1, two_overlaps_one},
+                new Object[]{list3, list2, three_overlaps_two},
+                new Object[]{list3, list1, three_overlaps_one},
+                new Object[]{list1, list3, one_overlaps_three}
+        };
+    }
+
+    @DataProvider(name = "overlapsData")
```
yfarjoun: please add a test with an empty collection on the lhs and another on the rhs.

src/main/java/htsjdk/samtools/util/IntervalList.java
nh13: If you look at the rest of the "actions", you'll see either they return `null` (`intersection`) or throw an exception (`concatenate`).  And then some of the other actions (ex. `union`) depend on these.  So I think it's fine the way it is (albeit I need to add a check for an empty list).

src/main/java/htsjdk/samtools/util/IntervalList.java
nh13: Ok, but it does feel like a pre-mature optimization.

src/test/java/htsjdk/samtools/util/IntervalListTest.java
nh13: Will add this comment:
```
 // NB: commented lines below are there to show hte intervals in the first list that will not be in the resulting list
```

src/test/java/htsjdk/samtools/util/IntervalListTest.java
nh13: Done.

nh13:
@yfarjoun back to you

nh13:
@yfarjoun bump

PR #884, opened 2017-05-28
Yf add tests

yfarjoun:
### Description
A PR containing tests for some woefully under-tested parts of the codebase. I also discovered some tests classes that do not extend HtsjdkTest...so they were not run....


src/main/java/htsjdk/samtools/filter/ReadNameFilter.java
```diff
@@ -81,17 +81,7 @@ public ReadNameFilter(final Set<String> readNameFilterSet, final boolean include
...
-                return false;
-            }
-        } else {
-            if (!readNameFilterSet.contains(record.getReadName())) {
-                return false;
-            }
-        }
-
-        return true;
+        return readNameFilterSet.contains(record.getReadName()) != includeReads;
```
droazen: While this version is more compact, I also think it is more confusing -- We had to draw a truth table to confirm that the behavior is the same in all cases.

src/main/java/htsjdk/samtools/filter/ReadNameFilter.java
```diff
@@ -104,19 +94,8 @@ public boolean filterOut(final SAMRecord record) {
...
-        } else {
-            if (!readNameFilterSet.contains(first.getReadName()) &&
-                !readNameFilterSet.contains(second.getReadName())) {
-                return false;
-            }
-        }
-
-        return true;
+        return (readNameFilterSet.contains(first.getReadName()) &&
+                readNameFilterSet.contains(second.getReadName())) != includeReads;
```
droazen: The negation of the expression `readNameFilterSet.contains(first.getReadName()) && readNameFilterSet.contains(second.getReadName())` is `! readNameFilterSet.contains(first.getReadName()) || ! readNameFilterSet.contains(second.getReadName())`, which doesn't match the old version from the else clause, which had `(!readNameFilterSet.contains(first.getReadName()) && !readNameFilterSet.contains(second.getReadName()))`

src/test/java/htsjdk/samtools/filter/IntervalFilterTest.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Test
+    public void testReads() {
+        SAMRecordSetBuilder builder = new SAMRecordSetBuilder();
+
+        int i = 0;
+        int expected = 0;
+        //reads are of length 36
+        builder.addPair("abutting" + i++, 0, 50 - 36, 151); //both abutting
```
lbergelson: I might just define a constant for something like this, ex: `50 - READ_LENGTH` which might be clearer and eliminate the need for the comment.

src/main/java/htsjdk/samtools/filter/ReadNameFilter.java
droazen: Drew out a truth table showing that the new version of this method is not the same as the old version for all cases (the two columns for contains correspond to the two tests for first and second):

```
Original version:

                     includeReads:   T     F
contains:
T     T                                      F     T    
T     F                                      T     T
F     T                                      T     T
F     F                                      T     F

New version:

                     includeReads:   T     F
contains:
T     T                                      F     T   
T     F                                      T     F
F     T                                      T     F
F     F                                      T     F
```

I also think the new version is more confusing / bug prone.

yfarjoun: I agree that for the two-read version this is changing the behavior, and while I should have advertised it more clearly, I think that it's the right thing. that said, I'll keep it as it was.

src/test/java/htsjdk/samtools/filter/IntervalFilterTest.java
yfarjoun: ok.

src/main/java/htsjdk/samtools/filter/ReadNameFilter.java
yfarjoun: ok.

yfarjoun:
@lbergelson @droazen, back to you.

src/test/java/htsjdk/tribble/readers/TabixReaderTest.java
```diff
@@ -14,8 +14,7 @@
 import java.util.List;
 import java.util.Set;
 
-import static org.testng.AssertJUnit.assertTrue;
-
+import static org.testng.Assert.assertTrue;
```
lbergelson: why this static import?

src/test/java/htsjdk/samtools/util/EdgingRecordAndOffsetTest.java
```diff
@@ -30,9 +30,8 @@
 import org.testng.annotations.BeforeTest;
 import org.testng.annotations.Test;
 
-import static org.junit.Assert.assertArrayEquals;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotSame;
+import static org.testng.Assert.assertEquals;
```
lbergelson: more statics, good to get rid of the accidental use of junit, but generally we don't use static imports

src/test/java/htsjdk/samtools/filter/IntervalFilterTest.java
```diff
@@ -0,0 +1,69 @@
...
+        fileHeader = IntervalList.fromFile(new File("src/test/resources/htsjdk/samtools/intervallist/IntervalListchr123_empty.interval_list")).getHeader();
+        fileHeader.setSortOrder(SAMFileHeader.SortOrder.unsorted);
+
+        list = new IntervalList(fileHeader);
+
+        list.add(new Interval("1", 50, 150));   //de-facto 1:50-150 1:301-500      2:1-150 2:250-270 2:290-400
+        list.add(new Interval("1", 301, 500));
+        list.add(new Interval("2", 1, 150));
+        list.add(new Interval("2", 250, 270));
+        list.add(new Interval("2", 300, 299)); // empty, but located.
```
droazen: Can you refactor to use a `DataProvider` instead of class-scope variables, which is a safer/better option as a general rule.

src/test/java/htsjdk/samtools/filter/NotPrimaryAlignmentFilterTest.java
```diff
@@ -0,0 +1,108 @@
...
+                                10,false).setSupplementaryAlignmentFlag(true);
+                    }
+                }
+            }
+        }
+
+        FilteringSamIterator filteringSamIterator = new FilteringSamIterator(builder.getRecords().iterator(),
+                new NotPrimaryAlignmentFilter());
+
+        Assert.assertEquals(filteringSamIterator.stream().count(),i*3);
```
droazen: Can you add a comment above each assert explaining why (for example) `i * 3` is the expected value?

yfarjoun:
@droazen and @lbergelson ...back to you.

PR #879, opened 2017-05-25
The existing code was throwing a NPE when using a CRAMFileReader for …

skashin:
…InputResource.Type.PATH

### Description

Please explain the changes you made here.
Explain the **motivation** for making this change. What existing problem does the pull request solve?

### Checklist

- [ ] Code compiles correctly
- [ ] New tests covering changes and new functionality
- [ ] All tests passing
- [ ] Extended the README / documentation, if necessary
- [ ] Is not backward compatible (breaks binary or source compatibility)

PR #885, opened 2017-05-31
Revert "The existing code was throwing a NPE when using a CRAMFileReader for …"

yfarjoun:
Reverts samtools/htsjdk#879

Somehow this broke the build. Reverting until further investigation.


skashin:
Hey, @yfarjoun, how can I reproduce the build error you mention above?
Thanks

PR #887, opened 2017-06-03
IntervalTreeMap: use Locatable interface

lindenb:
### Description

I've changed the signature of four functions in **IntervalTreeMap**. Instead of using a concrete class  'Interval', they will use 'Locatable'. It shouldn't break the code.
Furthermore it avoids to create a new Interval for each query. 

```
map.containsOverlapping( new Interval(variantContext.getContig(), variantContext.getStart(), variantContext.getEnd()));
```

it's now

```
map.containsOverlapping(variantContext)
```

### Checklist

- [x ] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/IntervalTreeMap.java
+++ b/src/main/java/htsjdk/samtools/util/IntervalTreeMap.java
@@ -165,16 +165,16 @@ public class IntervalTreeMap<T>
     }
     /**
      * Test overlapping interval 
-     * @param key the interval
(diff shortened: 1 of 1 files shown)


magicDGS:
Will be nice to have a `LocatableTreeMap` instead, but I like this idea too!

yfarjoun:
this is nice. i was just wishing for this the other day...thanks!

@tfenne any concerns?

PR #888, opened 2017-06-06
Cleaned up SAMUtils (mostly cosmetic)

yfarjoun:
- adding more information to some exceptions
- fixing java-doc
- java8 <>
- if/else alignment & braces
- removed use of @Deprecated SAMRecordUtils class
- @Deprecated all methods and members of SAMRecordUtils

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMRecordUtil.java
+++ b/src/main/java/htsjdk/samtools/SAMRecordUtil.java
@@ -31,18 +31,18 @@ import java.util.List;
  *
  * @author alecw@broadinstitute.org
  *
- * Deprecated 10/27/2016 Use {@link SAMRecord} constants and functions
+ * @deprecated 10/27/2016 Use {@link SAMRecord} constants and functions
  */
 @Deprecated
 public class SAMRecordUtil {
     /**
-     * 6/5/2017 Use {@link SAMRecord#TAGS_TO_REVERSE_COMPLEMENT}
+     * @deprecated 6/5/2017 Use {@link SAMRecord#TAGS_TO_REVERSE_COMPLEMENT}
      */
     @Deprecated
     public static List<String> TAGS_TO_REVERSE_COMPLEMENT = Arrays.asList(SAMTag.E2.name(), SAMTag.SQ.name());
 
     /**
-     * 6/5/2017 Use {@link SAMRecord#TAGS_TO_REVERSE}
+     * @deprecated 6/5/2017 Use {@link SAMRecord#TAGS_TO_REVERSE}
      */
     @Deprecated
     public static List<String> TAGS_TO_REVERSE            = Arrays.asList(SAMTag.OQ.name(), SAMTag.U2.name());
@@ -54,7 +54,7 @@ public class SAMRecordUtil {
      * See {@link #TAGS_TO_REVERSE_COMPLEMENT} {@link #TAGS_TO_REVERSE}
      * for the default set of tags that are handled.
      *
-     * Deprecated 6/5/2017 Use {@link SAMRecord#reverseComplement} but note that the default behavior there is different
+     * @deprecated  6/5/2017 Use {@link SAMRecord#reverseComplement} but note that the default behavior there is different
      * It will default to making a copy, not reverse-complementing in-place!
      */
     @Deprecated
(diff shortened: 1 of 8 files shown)


src/main/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -1002,7 +1007,7 @@ public static String getCanonicalRecordName(final SAMRecord record) {
      * or the given record's start position is greater than its mate's start position, zero is automatically returned.
      * NB: This method assumes that the record's mate is not contained within the given record's alignment.
      *
-     * @param rec
+     * @param rec SamRecord that needs clipping due to overlapping pairs.
```
nh13: `SamRecord` -> `SAMRecord`

src/main/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -1151,7 +1158,7 @@ public static boolean isValidUnsignedIntegerAttribute(long value) {
 
             /* break string using comma */
             final String commaStrs[] = COMMA_PAT.split(semiColonStr);
-            if( commaStrs.length != 6 )  throw new SAMException("Bad 'SA' attribute in " + semiColonStr);
+            if( commaStrs.length != 6 )  throw new SAMException("Bad 'SA' attribute in " + semiColonStr + ". Record: " + record.toString());
```
nh13: Fix the `if` formatting while here.

magicDGS:
@yfarjoun, could you add a javadoc String explaining why they are deprecated and what to use instead? In addition, it will be nice there is a comment specifying the date of deprecation...

src/main/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -87,27 +90,27 @@
     private static final byte COMPRESSED_K_HIGH = (byte) (COMPRESSED_K_LOW << 4);
     private static final byte COMPRESSED_D_HIGH = (byte) (COMPRESSED_D_LOW << 4);
     private static final byte COMPRESSED_B_HIGH = (byte) (COMPRESSED_B_LOW << 4);
-    
-    private static final byte [] COMPRESSED_LOOKUP_TABLE = 
+
+    private static final byte[] COMPRESSED_LOOKUP_TABLE =
             new byte[]{
```
pshapiro4broad: `new byte[]` is unnecessary here. If you remove it, I would pull the `{` back onto the previous line.

src/main/java/htsjdk/samtools/SAMRecordUtil.java
```diff
@@ -39,7 +39,9 @@
  */
 @Deprecated
```
pshapiro4broad: This is good but you also want to use the `@deprecated` Javadoc tag that includes the version number where the API became deprecated. I think that marking a class as deprecated should be sufficient and you don't need to also mark each member. But I guess it couldn't hurt to do that either.

FYI the Java doc on deprecation is here: http://docs.oracle.com/javase/7/docs/technotes/guides/javadoc/deprecation/deprecation.html

src/main/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -721,15 +730,15 @@ public static long findVirtualOffsetOfFirstRecordInBam(final File bamFile) {
                     refBase += length;
                     break;
                 default:
-                    throw new IllegalStateException("Case statement didn't deal with " + cigarTypeName + " op: " + e.getOperator());
+                    throw new IllegalStateException("Case statement didn't deal with " + cigarTypeName + " op: " + e.getOperator() + "in CIGAR: " + cigar.toString());
```
pshapiro4broad: `toString()` is unnecessary here and below since `"text" + obj` will automatically call `obj.toString()`. You can enable a setting in intellij to mark these in the editor and auto-fix them for you.

src/main/java/htsjdk/samtools/SAMRecordUtil.java
yfarjoun: hmm. I haven't seen this being used anywhere..not sure what that buys us...

lbergelson: I agree with @pshapiro4broad about using the javadoc tag.  We try to use it other places we deprecated things.

magicDGS: I also use it in some deprecation patches that I did in HTSJDK...

yfarjoun: Ok. I relent...:-) went back and fixed all the @Deprecation s

pshapiro4broad:
👍

yfarjoun:
I have a different PR that fixes the SRA issue...so I'll merge this despite the failures...

PR #890, opened 2017-06-06
overloading BlockCompressedInputStream.checkTerminator to support NIO

lbergelson:
BlockCompressedInputStream.checkTerminator only worked on files.  Gatk4 needs it to accept Paths and ideally SeekableByteChannels directly.  Adding overloads that take Path and SeekableByteChannel and refactoring the internals to use SeekableByteChannel instead of RandomAccessFile.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
+++ b/src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
@@ -586,51 +586,95 @@ public class BlockCompressedInputStream extends InputStream implements LocationA
 
     public enum FileTermination {HAS_TERMINATOR_BLOCK, HAS_HEALTHY_LAST_BLOCK, DEFECTIVE}
 
+    /**
+     *
+     * @param file the file to check
+     * @return status of the last compressed block
+     * @throws IOException
+     */
     public static FileTermination checkTermination(final File file) throws IOException {
         return checkTermination(file == null ? null : file.toPath());
     }
 
+    /**
+     *
+     * @param path to the file to check
+     * @return status of the last compressed block
+     * @throws IOException
+     */
     public static FileTermination checkTermination(final Path path) throws IOException {
         try( final SeekableByteChannel channel = Files.newByteChannel(path, StandardOpenOption.READ) ){
             return checkTermination(channel);
         }
     }
 
+    /**
+     * check the status of the final bzgipped block for the given bgzipped resource
+     *
+     * @param channel an open channel to read from,
+     * the channel will remain open and the initial position will be restored when the operation completes
+     * this makes no guarantee about the state of the channel if an exception is thrown during reading
+     *
+     * @return the status of the last compressed black
+     * @throws IOException
+     */
     public static FileTermination checkTermination(SeekableByteChannel channel) throws IOException {
         final long fileSize = channel.size();
         if (fileSize < BlockCompressedStreamConstants.EMPTY_GZIP_BLOCK.length) {
             return FileTermination.DEFECTIVE;
         }
-        channel.position(fileSize - BlockCompressedStreamConstants.EMPTY_GZIP_BLOCK.length);
+        final long initialPosition = channel.position();
+        boolean exceptionThrown = false;
+        try {
+            channel.position(fileSize - BlockCompressedStreamConstants.EMPTY_GZIP_BLOCK.length);
 
-        final ByteBuffer bytebuffer = ByteBuffer.allocate(BlockCompressedStreamConstants.EMPTY_GZIP_BLOCK.length);
-        readFully(channel, bytebuffer);
-        if (Arrays.equals(bytebuffer.array(), BlockCompressedStreamConstants.EMPTY_GZIP_BLOCK)) {
-            return FileTermination.HAS_TERMINATOR_BLOCK;
-        }
-        final int bufsize = (int)Math.min(fileSize, BlockCompressedStreamConstants.MAX_COMPRESSED_BLOCK_SIZE);
-        final byte[] buf = new byte[bufsize];
-        channel.position(fileSize - bufsize);
-        readFully(channel, ByteBuffer.wrap(buf));
-        for (int i = buf.length - BlockCompressedStreamConstants.EMPTY_GZIP_BLOCK.length;
-                i >= 0; --i) {
-            if (!preambleEqual(BlockCompressedStreamConstants.GZIP_BLOCK_PREAMBLE,
-                    buf, i, BlockCompressedStreamConstants.GZIP_BLOCK_PREAMBLE.length)) {
-                continue;
+            //Check if the end of the file is an empty gzip block which is used as the terminator for a bgzipped file
+            final ByteBuffer lastBlockBuffer = ByteBuffer.allocate(BlockCompressedStreamConstants.EMPTY_GZIP_BLOCK.length);
+            readFully(channel, lastBlockBuffer);
+            if (Arrays.equals(lastBlockBuffer.array(), BlockCompressedStreamConstants.EMPTY_GZIP_BLOCK)) {
+                return FileTermination.HAS_TERMINATOR_BLOCK;
             }
-            final ByteBuffer byteBuffer = ByteBuffer.wrap(buf, i + BlockCompressedStreamConstants.GZIP_BLOCK_PREAMBLE.length, 4);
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -588,40 +587,53 @@ private int unpackInt32(final byte[] buffer, final int offset) {
...
-                }
-                final ByteBuffer byteBuffer = ByteBuffer.wrap(buf, i + BlockCompressedStreamConstants.GZIP_BLOCK_PREAMBLE.length, 4);
-                byteBuffer.order(ByteOrder.LITTLE_ENDIAN);
-                final int totalBlockSizeMinusOne =  byteBuffer.getShort() & 0xFFFF;
-                if (buf.length - i == totalBlockSizeMinusOne + 1) {
-                    return FileTermination.HAS_HEALTHY_LAST_BLOCK;
-                } else {
-                    return FileTermination.DEFECTIVE;
-                }
+            final ByteBuffer byteBuffer = ByteBuffer.wrap(buf, i + BlockCompressedStreamConstants.GZIP_BLOCK_PREAMBLE.length, 4);
```
droazen: You might want to change the naming here to distinguish from `bytebuffer` above

src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -588,40 +587,53 @@ private int unpackInt32(final byte[] buffer, final int offset) {
...
-                return FileTermination.HAS_TERMINATOR_BLOCK;
+        channel.position(fileSize - BlockCompressedStreamConstants.EMPTY_GZIP_BLOCK.length);
+
+        final ByteBuffer bytebuffer = ByteBuffer.allocate(BlockCompressedStreamConstants.EMPTY_GZIP_BLOCK.length);
+        readFully(channel, bytebuffer);
+        if (Arrays.equals(bytebuffer.array(), BlockCompressedStreamConstants.EMPTY_GZIP_BLOCK)) {
+            return FileTermination.HAS_TERMINATOR_BLOCK;
+        }
+        final int bufsize = (int)Math.min(fileSize, BlockCompressedStreamConstants.MAX_COMPRESSED_BLOCK_SIZE);
+        final byte[] buf = new byte[bufsize];
```
droazen: While you're at it, you might want to rename `buf` as well to distinguish from the member variable

src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -588,40 +587,53 @@ private int unpackInt32(final byte[] buffer, final int offset) {
...
+        return checkTermination(file == null ? null : file.toPath());
+    }
+
+    public static FileTermination checkTermination(final Path path) throws IOException {
+        try( final SeekableByteChannel channel = Files.newByteChannel(path, StandardOpenOption.READ) ){
+            return checkTermination(channel);
+        }
+    }
+
+    public static FileTermination checkTermination(SeekableByteChannel channel) throws IOException {
```
droazen: Add javadoc for the new overloads, including stream closing behavior.

src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -588,40 +587,53 @@ private int unpackInt32(final byte[] buffer, final int offset) {
...
+                return FileTermination.DEFECTIVE;
             }
-            return FileTermination.DEFECTIVE;
-        } finally {
-            raFile.close();
+        }
+        return FileTermination.DEFECTIVE;
+    }
+
+    private static void readFully(SeekableByteChannel channel, ByteBuffer dst) throws IOException {
```
droazen: Add unit test for `readFully()`

src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -588,40 +587,53 @@ private int unpackInt32(final byte[] buffer, final int offset) {
     public enum FileTermination {HAS_TERMINATOR_BLOCK, HAS_HEALTHY_LAST_BLOCK, DEFECTIVE}
 
     public static FileTermination checkTermination(final File file) throws IOException {
-        final long fileSize = file.length();
+        return checkTermination(file == null ? null : file.toPath());
+    }
+
+    public static FileTermination checkTermination(final Path path) throws IOException {
```
droazen: Expand existing unit tests to call into these two new overloads as well.

lbergelson: done

src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
lbergelson: done

src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
lbergelson: renamed

src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
lbergelson: bytebuffer lowercase renamed

src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
lbergelson: done

src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -587,41 +586,98 @@ private int unpackInt32(final byte[] buffer, final int offset) {
...
 
+    /**
+     *
+     * @param file the file to check
+     * @return status of the last compressed block
+     * @throws IOException
+     */
     public static FileTermination checkTermination(final File file) throws IOException {
-        final long fileSize = file.length();
+        return checkTermination(file == null ? null : file.toPath());
```
pshapiro4broad: Won't this cause `checkTermination(Path)` to throw when `file == null`? If so avoiding the NPE here doesn't seem worthwhile. Maybe you want to return `FileTermination.DEFECTIVE` in that case?

src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -587,41 +586,98 @@ private int unpackInt32(final byte[] buffer, final int offset) {
...
             }
             return FileTermination.DEFECTIVE;
+        } catch (final Throwable e) {
+            exceptionThrown = true;
+            throw e;
         } finally {
-            raFile.close();
+            //if an exception was thrown we don't want to reset the position because that would be likely to throw again
+            //and suppress the initial exception
+            if(!exceptionThrown) {
```
pshapiro4broad: This code is not consistent with its use of spaces around `if` elements. Here there's no space between `if` and `(`. On line 679 there is a space there, but not after the `)`. On line 654 there are spaces both before and after.

The code should be consistent; putting a space before and after the `()`s is the java standard and is what most of the existing code does, so I would recommend formatting it that way.

The `try` on line 606 is also formatted inconsistently with respect to the surrounding code.

src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
lbergelson: @pshapiro4broad It absolutely just shifts the npr to later.  I wanted to just keep the exact behavior of the original function, which was to NPR somewhere later.  It does probably makes sense to check for null and throw a nice exception instead of an NPR.  Returning defective would be a bad idea for a null file I think since it would hide other errors.

src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
lbergelson: That's a good point.  I'm not sure it's worth opening another pr over though since I've already merged this one.

pshapiro4broad: no worries, something to keep in mind for next time :)

PR #892, opened 2017-06-08
updating artifactory link to point to new artifactory

lbergelson:
* the Broad artifactory moved from a self hosted repository at artifactory.broadinstitute.org
to a web based artifactory at broadinstitute.jfrog.io/broadinstitute/
this was causing our snapshot uploads to fail
* updating our snapshot repository to point to the new link
* the old artifactory has a redirect in place to the new artifactory so
artifacts relying on the old url should still resolve.  For unknown
reasons this redirect doesn't work with gradle uploads, probably a
gradle bug.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -169,7 +169,7 @@ uploadArchives {
                 authentication(userName: project.findProperty("sonatypeUsername"), password: project.findProperty("sonatypePassword"))
             }
 
-            snapshotRepository(url: "https://artifactory.broadinstitute.org/artifactory/libs-snapshot-local/") {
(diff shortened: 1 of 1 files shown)


magicDGS:
One question about this for downstream projects using HTSJDK/GATK/etc: will the old artifactory links continue working forever or should we change them in the near future? Thanks in advance!

lbergelson:
@magicDGS I was given a promise to keep the link up for at least a year, but I'm going to try to advocate for keeping the link around forever because it seems like really bad form to break old builds that way.  

It's probably prudent to update even if the old link stays around though.

lbergelson:
@droazen Simple change here.

Issue #297, opened 2015-08-06
SAMRecord with no bases blows up on toString


akiezun:
htsjdk 1.137
`toString()` should never blow up.
But it does blow up for a SAMRecord with no bases. The stacktrace is:
`java.lang.NullPointerException
    at htsjdk.samtools.SAMRecord.getReadLength(SAMRecord.java:232)
    at htsjdk.samtools.SAMRecord.toString(SAMRecord.java:1768)
`

And the culprit is `return getReadBases().length;`

PR #893, opened 2017-06-09
Fix SAMRecord.getReadLength() so it does not throw an exception if null bases

ronlevine:
### Description

Fixes https://github.com/samtools/htsjdk/issues/297.
Fixed SAMRecord.getReadLength by not calling getReadBases().length read bases is null.
Added a unit test.
Made private unit tests public so they can be run.

### Checklist

- [X] Code compiles correctly
- [X] All tests passing

### DIFF
--- a/src/test/java/htsjdk/samtools/SAMRecordUnitTest.java
+++ b/src/test/java/htsjdk/samtools/SAMRecordUnitTest.java
@@ -776,7 +776,7 @@ public class SAMRecordUnitTest extends HtsjdkTest {
     }
 
     @Test
-    private void testNullHeaderDeepCopy() {
+    public void testNullHeaderDeepCopy() {
         SAMRecord sam = createTestRecordHelper();
         sam.setHeader(null);
         final SAMRecord deepCopy = sam.deepCopy();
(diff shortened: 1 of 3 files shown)


ronlevine:
@yfarjoun Please review.

src/test/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -776,7 +776,7 @@ public void testNullHeaderRecordValidation() {
     }
 
     @Test
-    private void testNullHeaderDeepCopy() {
+    public void testNullHeaderDeepCopy() {
```
yfarjoun: oh snap, these were not being tested?

src/test/java/htsjdk/samtools/SAMRecordUnitTest.java
```diff
@@ -1039,4 +1039,20 @@ public SAMRecord createTestSamRec() {
 
         return(rec);
     }
+
+    @DataProvider
+    public Object [][] readBasesGetReadLengthData() {
```
yfarjoun: Can you also add tests that set the bases using `setReadString`? in particular try to break it with an empty string and the "*" string that means the empty string for read bases.

src/test/java/htsjdk/samtools/SAMRecordUnitTest.java
ronlevine: Yes. I assume snpp == &$%@.

src/test/java/htsjdk/samtools/SAMRecordUnitTest.java
ronlevine: Done

ronlevine:
@yfarjoun Responded to comments. Please take another look.

Issue #256, opened 2015-05-19
Bug in DuplicateScoringStrategy


mattsooknah:
DuplicateScoringStrategy.compare() compares two records, essentially returning score(rec2) - score(rec1). However, it has:

``` java
// always prefer paired over non-paired
if (rec1.getReadPairedFlag() != rec2.getReadPairedFlag()) return rec1.getReadPairedFlag() ? 1 : -1;
```

The signs are flipped. Luckily, the only place where this method is called (in MarkQueue) does the correct version of the comparison earlier on, so currently I think this is a no-op bug. But it should still be fixed.

nh13:
https://github.com/samtools/htsjdk/blob/919e901557258db71501d825a4aaeb031759e76e/src/java/htsjdk/samtools/DuplicateScoringStrategy.java#L94

PR #897, opened 2017-06-09
Fixed flipped sign when comparing read paired flags

ronlevine:
### Description

Fixes https://github.com/samtools/htsjdk/issues/256.
The read comparing duplicates, if one of the reds is not paired, the first read in the comparison should return -1 if it is paired, not 1. 

### Checklist

- [X] Code compiles correctly
- [X] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
+++ b/src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
@@ -36,7 +36,7 @@ public class DuplicateScoringStrategy {
     public enum ScoringStrategy {
         SUM_OF_BASE_QUALITIES,
         TOTAL_MAPPED_REFERENCE_LENGTH,
-        RANDOM,
+        RANDOM
     }
 
     /** Hash used for the RANDOM scoring strategy. */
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/DuplicateScoringStrategy.java
```diff
@@ -128,7 +128,7 @@ public static int compare(final SAMRecord rec1, final SAMRecord rec2, final Scor
         int cmp;
 
         // always prefer paired over non-paired
-        if (rec1.getReadPairedFlag() != rec2.getReadPairedFlag()) return rec1.getReadPairedFlag() ? 1 : -1;
+        if (rec1.getReadPairedFlag() != rec2.getReadPairedFlag()) return rec1.getReadPairedFlag() ? -1 : 1;
```
yfarjoun: yikes! 

that's a surprising bug. I guess we do not have many files that have paired and unpaired records together...

Can you add a test that covers this ?

ronlevine: Will add a test. The only place it's used picard's MarkQueue.compare but it covers same logic correctly before it uses this method, so the bug is masked.

nh13: Great find!!!

Issue #902, opened 2017-06-19
Support for partial VCF writing - htsget extension for VCF


cyenyxe:
### Motivation
The [htsget specification](https://samtools.github.io/hts-specs/htsget.html) allows to query multiple data blocks in parallel by providing links to different regions. When extending it to support VCF, it is necessary to return metadata and data blocks separately.

### Subject of the issue
The way the [VCFWriter class](https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java) currently works forces to initialize the header and inner `VCFEncoder` when the header is first written. The writer can't be fully initialized unless the header is written.

If the classes derived from `VariantContextWriter` supported setting and writing the header in 2 different steps, both the existing use cases and the htsget one would work.

lbergelson:
This was closed by #904   Thanks everyone for the work on this!

PR #904, opened 2017-06-20
Support for partial VCF writing

amilamanoj:
### Description

This PR addresses Issue #902.

Added support for setting and writing the header in two different steps for the classes derived from `VariantContextWriter`.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/writer/AsyncVariantContextWriter.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/AsyncVariantContextWriter.java
@@ -55,8 +55,8 @@ public class AsyncVariantContextWriter extends AbstractAsyncWriter<VariantContex
     }
 
     @Override
-    public void setVcfHeader(VCFHeader header) {
-        this.underlyingWriter.setVcfHeader(header);
+    public void setVCFHeader(final VCFHeader header) {
+        this.underlyingWriter.setVCFHeader(header);
     }
 
 }
--- a/src/main/java/htsjdk/variant/variantcontext/writer/AsyncVariantContextWriter.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/AsyncVariantContextWriter.java
@@ -53,4 +53,10 @@ public class AsyncVariantContextWriter extends AbstractAsyncWriter<VariantContex
     public boolean checkError() {
         return false;
     }
+
+    @Override
+    public void setVcfHeader(VCFHeader header) {
+        this.underlyingWriter.setVcfHeader(header);
+    }
+
 }
--- a/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
@@ -145,39 +145,13 @@ class BCF2Writer extends IndexingVariantContextWriter {
 
     @Override
     public void writeHeader(VCFHeader header) {
-        // make sure the header is sorted correctly
-        header = new VCFHeader(header.getMetaDataInSortedOrder(), header.getGenotypeSamples());
-
-        // create the config offsets map
-        if ( header.getContigLines().isEmpty() ) {
-            if ( ALLOW_MISSING_CONTIG_LINES ) {
-                if ( GeneralUtils.DEBUG_MODE_ENABLED ) {
-                    System.err.println("No contig dictionary found in header, falling back to reference sequence dictionary");
-                }
-                createContigDictionary(VCFUtils.makeContigHeaderLines(getRefDict(), null));
-            } else {
-                throw new IllegalStateException("Cannot write BCF2 file with missing contig lines");
-            }
-        } else {
-            createContigDictionary(header.getContigLines());
-        }
-
-        // set up the map from dictionary string values -> offset
-        final ArrayList<String> dict = BCF2Utils.makeDictionary(header);
-        for ( int i = 0; i < dict.size(); i++ ) {
-            stringDictionaryMap.put(dict.get(i), i);
-        }
-
-        sampleNames = header.getGenotypeSamples().toArray(new String[header.getNGenotypeSamples()]);
-
-        // setup the field encodings
-        fieldManager.setup(header, encoder, stringDictionaryMap);
+        setVcfHeader(header);
 
         try {
             // write out the header into a byte stream, get its length, and write everything to the file
             final ByteArrayOutputStream capture = new ByteArrayOutputStream();
             final OutputStreamWriter writer = new OutputStreamWriter(capture);
-            this.header = VCFWriter.writeHeader(header, writer, doNotWriteGenotypes, VCFWriter.getVersionLine(), "BCF2 stream");
+            this.header = VCFWriter.writeHeader(header, writer, VCFWriter.getVersionLine(), "BCF2 stream");
             writer.append('\0'); // the header is null terminated by a byte
             writer.close();
 
@@ -223,7 +197,9 @@ class BCF2Writer extends IndexingVariantContextWriter {
 
     @Override
     public void setVcfHeader(VCFHeader header) {
+        // make sure the header is sorted correctly
         header = new VCFHeader(header.getMetaDataInSortedOrder(), header.getGenotypeSamples());
+
         this.header = doNotWriteGenotypes ? new VCFHeader(header.getMetaDataInSortedOrder()) : header;
         // create the config offsets map
         if ( header.getContigLines().isEmpty() ) {
@@ -244,7 +220,7 @@ class BCF2Writer extends IndexingVariantContextWriter {
             stringDictionaryMap.put(dict.get(i), i);
         }
 
-        sampleNames = header.getGenotypeSamples().toArray(new String[header.getNGenotypeSamples()]);
+        sampleNames = this.header.getGenotypeSamples().toArray(new String[this.header.getNGenotypeSamples()]);
         // setup the field encodings
         fieldManager.setup(header, encoder, stringDictionaryMap);
 
--- a/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
@@ -221,6 +221,11 @@ class BCF2Writer extends IndexingVariantContextWriter {
         super.close();
     }
 
+    @Override
+    public void setVcfHeader(VCFHeader header) {
+        //no-op
+    }
+
     // --------------------------------------------------------------------------------
     //
     // implicit block
--- a/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
@@ -122,6 +122,12 @@ class BCF2Writer extends IndexingVariantContextWriter {
     private VCFHeader lastVCFHeaderOfUnparsedGenotypes = null;
     private boolean canPassOnUnparsedGenotypeDataForLastVCFHeader = false;
 
+    // is the header written to the output stream?
+    private boolean isHeaderWritten;
+
+    // is at least one part of body written to the output stream?
+    private boolean isBodyWritten;
+
     public BCF2Writer(final File location, final OutputStream output, final SAMSequenceDictionary refDict,
                       final boolean enableOnTheFlyIndexing, final boolean doNotWriteGenotypes) {
         super(writerName(location, output), location, output, refDict, enableOnTheFlyIndexing);
@@ -145,6 +151,9 @@ class BCF2Writer extends IndexingVariantContextWriter {
 
     @Override
     public void writeHeader(VCFHeader header) {
+        if (isHeaderWritten) {
+            throw new IllegalStateException("Header is already written");
+        }
         setVcfHeader(header);
 
         try {
@@ -159,6 +168,7 @@ class BCF2Writer extends IndexingVariantContextWriter {
             new BCFVersion(MAJOR_VERSION, MINOR_VERSION).write(outputStream);
             BCF2Type.INT32.write(headerBytes.length, outputStream);
             outputStream.write(headerBytes);
+            isHeaderWritten = true;
         } catch (IOException e) {
             throw new RuntimeIOException("BCF2 stream: Got IOException while trying to write BCF2 header", e);
         }
@@ -178,6 +188,7 @@ class BCF2Writer extends IndexingVariantContextWriter {
 
             // write the two blocks to disk
             writeBlock(infoBlock, genotypesBlock);
+            isBodyWritten = true;
         }
         catch ( IOException e ) {
             throw new RuntimeIOException("Error writing record to BCF2 file: " + vc.toString(), e);
@@ -197,6 +208,12 @@ class BCF2Writer extends IndexingVariantContextWriter {
 
     @Override
     public void setVcfHeader(VCFHeader header) {
+        if (isHeaderWritten) {
+            throw new IllegalStateException("Header is already written");
+        }
+        if (isBodyWritten) {
+            throw new IllegalStateException("Body is already written");
+        }
         // make sure the header is sorted correctly
         header = new VCFHeader(header.getMetaDataInSortedOrder(), header.getGenotypeSamples());
 
--- a/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
@@ -207,9 +207,8 @@ class BCF2Writer extends IndexingVariantContextWriter {
             throw new IllegalStateException("The header cannot be modified after the header or variants have been written to the output stream.");
         }
         // make sure the header is sorted correctly
-        this.header = new VCFHeader(header.getMetaDataInSortedOrder(), header.getGenotypeSamples());
-
-        this.header = doNotWriteGenotypes ? new VCFHeader(this.header.getMetaDataInSortedOrder()) : this.header;
+        this.header = doNotWriteGenotypes ? new VCFHeader(header.getMetaDataInSortedOrder()) : new VCFHeader(
+                header.getMetaDataInSortedOrder(), header.getGenotypeSamples());
         // create the config offsets map
         if ( header.getContigLines().isEmpty() ) {
             if ( ALLOW_MISSING_CONTIG_LINES ) {
--- a/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
@@ -238,6 +238,16 @@ class BCF2Writer extends IndexingVariantContextWriter {
         } else {
             createContigDictionary(header.getContigLines());
         }
+        // set up the map from dictionary string values -> offset
+        final ArrayList<String> dict = BCF2Utils.makeDictionary(header);
+        for ( int i = 0; i < dict.size(); i++ ) {
+            stringDictionaryMap.put(dict.get(i), i);
+        }
+
+        sampleNames = header.getGenotypeSamples().toArray(new String[header.getNGenotypeSamples()]);
+        // setup the field encodings
+        fieldManager.setup(header, encoder, stringDictionaryMap);
+
     }
 
     // --------------------------------------------------------------------------------
--- a/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
@@ -122,11 +122,9 @@ class BCF2Writer extends IndexingVariantContextWriter {
     private VCFHeader lastVCFHeaderOfUnparsedGenotypes = null;
     private boolean canPassOnUnparsedGenotypeDataForLastVCFHeader = false;
 
-    // is the header written to the output stream?
-    private boolean isHeaderWritten;
+    // is the header or body written to the output stream?
+    private boolean isWrittenToOutput;
 
-    // is at least one part of body written to the output stream?
-    private boolean isBodyWritten;
 
     public BCF2Writer(final File location, final OutputStream output, final SAMSequenceDictionary refDict,
                       final boolean enableOnTheFlyIndexing, final boolean doNotWriteGenotypes) {
@@ -151,16 +149,13 @@ class BCF2Writer extends IndexingVariantContextWriter {
 
     @Override
     public void writeHeader(VCFHeader header) {
-        if (isHeaderWritten) {
-            throw new IllegalStateException("Header is already written");
-        }
-        setVcfHeader(header);
+        setVCFHeader(header);
 
         try {
             // write out the header into a byte stream, get its length, and write everything to the file
             final ByteArrayOutputStream capture = new ByteArrayOutputStream();
             final OutputStreamWriter writer = new OutputStreamWriter(capture);
-            this.header = VCFWriter.writeHeader(header, writer, VCFWriter.getVersionLine(), "BCF2 stream");
+            this.header = VCFWriter.writeHeader(this.header, writer, VCFWriter.getVersionLine(), "BCF2 stream");
             writer.append('\0'); // the header is null terminated by a byte
             writer.close();
 
@@ -168,7 +163,7 @@ class BCF2Writer extends IndexingVariantContextWriter {
             new BCFVersion(MAJOR_VERSION, MINOR_VERSION).write(outputStream);
             BCF2Type.INT32.write(headerBytes.length, outputStream);
             outputStream.write(headerBytes);
-            isHeaderWritten = true;
+            isWrittenToOutput = true;
         } catch (IOException e) {
             throw new RuntimeIOException("BCF2 stream: Got IOException while trying to write BCF2 header", e);
         }
@@ -188,7 +183,7 @@ class BCF2Writer extends IndexingVariantContextWriter {
 
             // write the two blocks to disk
             writeBlock(infoBlock, genotypesBlock);
-            isBodyWritten = true;
+            isWrittenToOutput = true;
         }
         catch ( IOException e ) {
             throw new RuntimeIOException("Error writing record to BCF2 file: " + vc.toString(), e);
@@ -207,17 +202,14 @@ class BCF2Writer extends IndexingVariantContextWriter {
     }
 
     @Override
-    public void setVcfHeader(VCFHeader header) {
-        if (isHeaderWritten) {
-            throw new IllegalStateException("Header is already written");
-        }
-        if (isBodyWritten) {
-            throw new IllegalStateException("Body is already written");
+    public void setVCFHeader(final VCFHeader header) {
+        if (isWrittenToOutput) {
+            throw new IllegalStateException("The header cannot be modified after the header or variants have been written to the output stream.");
         }
         // make sure the header is sorted correctly
-        header = new VCFHeader(header.getMetaDataInSortedOrder(), header.getGenotypeSamples());
+        this.header = new VCFHeader(header.getMetaDataInSortedOrder(), header.getGenotypeSamples());
 
-        this.header = doNotWriteGenotypes ? new VCFHeader(header.getMetaDataInSortedOrder()) : header;
+        this.header = doNotWriteGenotypes ? new VCFHeader(this.header.getMetaDataInSortedOrder()) : this.header;
         // create the config offsets map
         if ( header.getContigLines().isEmpty() ) {
             if ( ALLOW_MISSING_CONTIG_LINES ) {
--- a/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
@@ -210,7 +210,7 @@ class BCF2Writer extends IndexingVariantContextWriter {
         this.header = doNotWriteGenotypes ? new VCFHeader(header.getMetaDataInSortedOrder()) : new VCFHeader(
                 header.getMetaDataInSortedOrder(), header.getGenotypeSamples());
         // create the config offsets map
-        if ( header.getContigLines().isEmpty() ) {
+        if ( this.header.getContigLines().isEmpty() ) {
             if ( ALLOW_MISSING_CONTIG_LINES ) {
                 if ( GeneralUtils.DEBUG_MODE_ENABLED ) {
                     System.err.println("No contig dictionary found in header, falling back to reference sequence dictionary");
@@ -220,17 +220,17 @@ class BCF2Writer extends IndexingVariantContextWriter {
                 throw new IllegalStateException("Cannot write BCF2 file with missing contig lines");
             }
         } else {
-            createContigDictionary(header.getContigLines());
+            createContigDictionary(this.header.getContigLines());
         }
         // set up the map from dictionary string values -> offset
-        final ArrayList<String> dict = BCF2Utils.makeDictionary(header);
+        final ArrayList<String> dict = BCF2Utils.makeDictionary(this.header);
         for ( int i = 0; i < dict.size(); i++ ) {
             stringDictionaryMap.put(dict.get(i), i);
         }
 
         sampleNames = this.header.getGenotypeSamples().toArray(new String[this.header.getNGenotypeSamples()]);
         // setup the field encodings
-        fieldManager.setup(header, encoder, stringDictionaryMap);
+        fieldManager.setup(this.header, encoder, stringDictionaryMap);
 
     }
 
--- a/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
@@ -223,7 +223,21 @@ class BCF2Writer extends IndexingVariantContextWriter {
 
     @Override
     public void setVcfHeader(VCFHeader header) {
-        //no-op
+        header = new VCFHeader(header.getMetaDataInSortedOrder(), header.getGenotypeSamples());
+        this.header = doNotWriteGenotypes ? new VCFHeader(header.getMetaDataInSortedOrder()) : header;
+        // create the config offsets map
+        if ( header.getContigLines().isEmpty() ) {
+            if ( ALLOW_MISSING_CONTIG_LINES ) {
+                if ( GeneralUtils.DEBUG_MODE_ENABLED ) {
+                    System.err.println("No contig dictionary found in header, falling back to reference sequence dictionary");
+                }
+                createContigDictionary(VCFUtils.makeContigHeaderLines(getRefDict(), null));
+            } else {
+                throw new IllegalStateException("Cannot write BCF2 file with missing contig lines");
+            }
+        } else {
+            createContigDictionary(header.getContigLines());
+        }
     }
 
     // --------------------------------------------------------------------------------
--- a/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
@@ -123,7 +123,7 @@ class BCF2Writer extends IndexingVariantContextWriter {
     private boolean canPassOnUnparsedGenotypeDataForLastVCFHeader = false;
 
     // is the header or body written to the output stream?
-    private boolean isWrittenToOutput;
+    private boolean outputHasBeenWritten;
 
 
     public BCF2Writer(final File location, final OutputStream output, final SAMSequenceDictionary refDict,
@@ -163,7 +163,7 @@ class BCF2Writer extends IndexingVariantContextWriter {
             new BCFVersion(MAJOR_VERSION, MINOR_VERSION).write(outputStream);
             BCF2Type.INT32.write(headerBytes.length, outputStream);
             outputStream.write(headerBytes);
-            isWrittenToOutput = true;
+            outputHasBeenWritten = true;
         } catch (IOException e) {
             throw new RuntimeIOException("BCF2 stream: Got IOException while trying to write BCF2 header", e);
         }
@@ -183,7 +183,7 @@ class BCF2Writer extends IndexingVariantContextWriter {
 
             // write the two blocks to disk
             writeBlock(infoBlock, genotypesBlock);
-            isWrittenToOutput = true;
+            outputHasBeenWritten = true;
         }
         catch ( IOException e ) {
             throw new RuntimeIOException("Error writing record to BCF2 file: " + vc.toString(), e);
@@ -203,7 +203,7 @@ class BCF2Writer extends IndexingVariantContextWriter {
 
     @Override
     public void setVCFHeader(final VCFHeader header) {
-        if (isWrittenToOutput) {
+        if (outputHasBeenWritten) {
             throw new IllegalStateException("The header cannot be modified after the header or variants have been written to the output stream.");
         }
         // make sure the header is sorted correctly
--- a/src/main/java/htsjdk/variant/variantcontext/writer/SortingVariantContextWriterBase.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/SortingVariantContextWriterBase.java
@@ -132,6 +132,11 @@ abstract class SortingVariantContextWriterBase implements VariantContextWriter {
         emitSafeRecords();
     }
 
+    @Override
+    public void setVcfHeader(VCFHeader header) {
+        innerWriter.setVcfHeader(header);
+    }
+
     /**
      * Gets a string representation of this object.
      * @return a string representation of this object
--- a/src/main/java/htsjdk/variant/variantcontext/writer/SortingVariantContextWriterBase.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/SortingVariantContextWriterBase.java
@@ -133,8 +133,8 @@ abstract class SortingVariantContextWriterBase implements VariantContextWriter {
     }
 
     @Override
-    public void setVcfHeader(VCFHeader header) {
-        innerWriter.setVcfHeader(header);
+    public void setVCFHeader(final VCFHeader header) {
+        innerWriter.setVCFHeader(header);
     }
 
     /**
--- a/src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
@@ -66,11 +66,8 @@ class VCFWriter extends IndexingVariantContextWriter {
     // should we always output a complete format record, even if we could drop trailing fields?
     private final boolean writeFullFormatField;
 
-    // is the header written to the output stream?
-    private boolean isHeaderWritten;
-
-    // is at least one part of body written to the output stream?
-    private boolean isBodyWritten;
+    // is the header or body written to the output stream?
+    private boolean isWrittenToOutput;
 
     /*
      * The VCF writer uses an internal Writer, based by the ByteArrayOutputStream lineBuffer,
@@ -135,16 +132,13 @@ class VCFWriter extends IndexingVariantContextWriter {
     @Override
     public void writeHeader(final VCFHeader header) {
 
-        if (isHeaderWritten) {
-            throw new IllegalStateException("Header is already written");
-        }
         // note we need to update the mHeader object after this call because they header
         // may have genotypes trimmed out of it, if doNotWriteGenotypes is true
         try {
-            setVcfHeader(header);
+            setVCFHeader(header);
             writeHeader(header, writer, getVersionLine(), getStreamName());
             writeAndResetBuffer();
-            isHeaderWritten = true;
+            isWrittenToOutput = true;
         } catch ( IOException e ) {
             throw new RuntimeIOException("Couldn't write file " + getStreamName(), e);
         }
@@ -231,19 +225,16 @@ class VCFWriter extends IndexingVariantContextWriter {
             write("\n");
 
             writeAndResetBuffer();
-            isBodyWritten = true;
+            isWrittenToOutput = true;
         } catch (IOException e) {
             throw new RuntimeIOException("Unable to write the VCF object to " + getStreamName(), e);
         }
     }
 
     @Override
-    public void setVcfHeader(VCFHeader header) {
-        if (isHeaderWritten) {
-            throw new IllegalStateException("Header is already written");
-        }
-        if (isBodyWritten) {
-            throw new IllegalStateException("Body is already written");
+    public void setVCFHeader(final VCFHeader header) {
+        if (isWrittenToOutput) {
+            throw new IllegalStateException("The header cannot be modified after the header or variants have been written to the output stream.");
         }
         this.mHeader = doNotWriteGenotypes ? new VCFHeader(header.getMetaDataInSortedOrder()) : header;
         this.vcfEncoder = new VCFEncoder(this.mHeader, this.allowMissingFieldsInHeader, this.writeFullFormatField);
--- a/src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
@@ -136,7 +136,7 @@ class VCFWriter extends IndexingVariantContextWriter {
         // may have genotypes trimmed out of it, if doNotWriteGenotypes is true
         try {
             setVCFHeader(header);
-            writeHeader(header, writer, getVersionLine(), getStreamName());
+            writeHeader(this.mHeader, writer, getVersionLine(), getStreamName());
             writeAndResetBuffer();
             isWrittenToOutput = true;
         } catch ( IOException e ) {
--- a/src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
@@ -67,7 +67,7 @@ class VCFWriter extends IndexingVariantContextWriter {
     private final boolean writeFullFormatField;
 
     // is the header or body written to the output stream?
-    private boolean isWrittenToOutput;
+    private boolean outputHasBeenWritten;
 
     /*
      * The VCF writer uses an internal Writer, based by the ByteArrayOutputStream lineBuffer,
@@ -138,7 +138,7 @@ class VCFWriter extends IndexingVariantContextWriter {
         try {
             writeHeader(this.mHeader, writer, getVersionLine(), getStreamName());
             writeAndResetBuffer();
-            isWrittenToOutput = true;
+            outputHasBeenWritten = true;
         } catch ( IOException e ) {
             throw new RuntimeIOException("Couldn't write file " + getStreamName(), e);
         }
@@ -225,7 +225,7 @@ class VCFWriter extends IndexingVariantContextWriter {
             write("\n");
 
             writeAndResetBuffer();
-            isWrittenToOutput = true;
+            outputHasBeenWritten = true;
         } catch (IOException e) {
             throw new RuntimeIOException("Unable to write the VCF object to " + getStreamName(), e);
         }
@@ -233,7 +233,7 @@ class VCFWriter extends IndexingVariantContextWriter {
 
     @Override
     public void setVCFHeader(final VCFHeader header) {
-        if (isWrittenToOutput) {
+        if (outputHasBeenWritten) {
             throw new IllegalStateException("The header cannot be modified after the header or variants have been written to the output stream.");
         }
         this.mHeader = doNotWriteGenotypes ? new VCFHeader(header.getMetaDataInSortedOrder()) : header;
--- a/src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
@@ -131,8 +131,8 @@ class VCFWriter extends IndexingVariantContextWriter {
         // note we need to update the mHeader object after this call because they header
         // may have genotypes trimmed out of it, if doNotWriteGenotypes is true
         try {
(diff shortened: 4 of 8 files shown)


src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
```diff
@@ -221,6 +221,11 @@ public void close() {
         super.close();
     }
 
+    @Override
+    public void setVcfHeader(VCFHeader header) {
+        //no-op
```
droazen: Why is this a "no-op"? This doesn't seem right. The `add()` method in `BCF2Writer` will blow up without a header, since it calls `vc = vc.fullyDecode(header, false);`

src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriter.java
```diff
@@ -49,4 +49,6 @@
     public boolean checkError();
     
     public void add(VariantContext vc);
+
+    void setVcfHeader(VCFHeader header);
```
droazen: Add javadoc for the new interface method.

src/main/java/htsjdk/variant/variantcontext/writer/AsyncVariantContextWriter.java
```diff
@@ -53,4 +53,10 @@ public void writeHeader(final VCFHeader header) {
     public boolean checkError() {
         return false;
     }
+
+    @Override
+    public void setVcfHeader(VCFHeader header) {
+        this.underlyingWriter.setVcfHeader(header);
```
yfarjoun: untested according to code cov

src/main/java/htsjdk/variant/variantcontext/writer/SortingVariantContextWriterBase.java
```diff
@@ -132,6 +132,11 @@ public synchronized void add(VariantContext vc) {
         emitSafeRecords();
     }
 
+    @Override
+    public void setVcfHeader(VCFHeader header) {
+        innerWriter.setVcfHeader(header);
```
yfarjoun: also untested

yfarjoun: but don't worry about this one since SortingVariantContextWriterBase is basically ununsed.

src/test/java/htsjdk/variant/variantcontext/writer/VCFWriterUnitTest.java
```diff
@@ -26,10 +26,13 @@
 package htsjdk.variant.variantcontext.writer;
```
droazen: You need tests to prove that the `setVcfHeader()` works for all concrete implementations of `VariantContextWriter`

src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
```diff
@@ -228,4 +228,10 @@ public void add(final VariantContext context) {
             throw new RuntimeIOException("Unable to write the VCF object to " + getStreamName(), e);
         }
     }
+
+    @Override
+    public void setVcfHeader(VCFHeader header) {
+        this.mHeader = doNotWriteGenotypes ? new VCFHeader(header.getMetaDataInSortedOrder()) : header;
```
droazen: Why are you changing the header on-the-fly like this when `doNotWriteGenotypes` is true? This at least requires an explanatory comment.

src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
```diff
@@ -228,4 +228,10 @@ public void add(final VariantContext context) {
...
+        this.vcfEncoder = new VCFEncoder(this.mHeader, this.allowMissingFieldsInHeader, this.writeFullFormatField);
```
droazen: Since this duplicates code in the existing `writeHeader()` method, would it be possible to have `writeHeader()` call into `setVcfHeader()`?

src/test/java/htsjdk/variant/variantcontext/writer/VCFWriterUnitTest.java
droazen: Should also test interaction between `setVcfHeader()` and `writeHeader()` (ie., test calling one and then the other for the various implementation).

src/test/java/htsjdk/variant/variantcontext/writer/VCFWriterUnitTest.java
```diff
@@ -133,6 +137,49 @@ public void testBasicWriteAndRead(final String extension) throws IOException {
 
     }
 
+    /** test, using the writer and reader, that we can output and input a VCF body without problems */
+    @Test(dataProvider = "vcfExtensionsDataProvider")
+    public void testWriteAndReadVCFBody(final String extension) throws IOException {
+        final File fakeVCFFile = File.createTempFile("testWriteAndReadVCFBody.", extension);
+        fakeVCFFile.deleteOnExit();
```
droazen: Use method `VariantBaseTest.createTempFile()` here, which does this for you.

src/test/java/htsjdk/variant/variantcontext/writer/VCFWriterUnitTest.java
```diff
@@ -133,6 +137,49 @@ public void testBasicWriteAndRead(final String extension) throws IOException {
...
+        final VCFHeader header = createFakeHeader(metaData, additionalColumns, sequenceDict);
+        final VariantContextWriter writer = new VariantContextWriterBuilder()
+                .setOutputFile(fakeVCFFile)
+                .setReferenceDictionary(sequenceDict)
+                .setOptions(EnumSet.of(Options.ALLOW_MISSING_FIELDS_IN_HEADER, Options.INDEX_ON_THE_FLY))
+                .build();
+        writer.setVcfHeader(header);
+        writer.add(createVC(header));
+        writer.add(createVC(header));
+        writer.close();
```
droazen: Use a `try-with-resources` block rather than an explicit `close()` call, to ensure that the writer gets closed even if there's an exception.

src/test/java/htsjdk/variant/variantcontext/writer/VCFWriterUnitTest.java
```diff
@@ -133,6 +137,49 @@ public void testBasicWriteAndRead(final String extension) throws IOException {
...
+
+        final VCFCodec codec = new VCFCodec();
+        codec.setVCFHeader(header, VCFHeaderVersion.VCF4_0);
```
droazen: Why 4.0? The writer above will write a 4.2 vcf?

src/test/java/htsjdk/variant/variantcontext/writer/VCFWriterUnitTest.java
```diff
@@ -133,6 +137,49 @@ public void testBasicWriteAndRead(final String extension) throws IOException {
...
+        AsciiLineReaderIterator iterator;
+        if (".vcf.gz".equals(extension)) {
+            iterator = new AsciiLineReaderIterator(new AsciiLineReader(new BlockCompressedInputStream(fakeVCFFile)));
+        } else {
+            iterator = new AsciiLineReaderIterator(new AsciiLineReader(new FileInputStream(fakeVCFFile)));
```
droazen: All input stream creations should be wrapped in a `try-with-resources` to ensure the streams get closed.

droazen: Also, you should just use `AbstractFeatureReader.getFeatureReader()` here to get a reader on the vcf file -- it will handle the block-gzipped case for you automatically.

amilamanoj: `AbstractFeatureReader.getFeatureReader()` cannot be used here because it includes a check for header is the vcf file. Since we are reading a data-block, calling this method throws `htsjdk.tribble.TribbleException$MalformedFeatureFile`.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,253 @@
...
+            counter++;
+        }
+        Assert.assertEquals(counter, 2);
+    }
+
+    /**
+     * test, using the writer and reader, that we can output and input a BCF body without header
+     */
+    @Test
+    public void testWriteAndReadBCFBody() throws IOException {
```
cmnbroad: Maybe call this testWriteAndReadBCFHeaderless to emphasize whats unique about it.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,253 @@
...
+                .setOptions(EnumSet.of(Options.ALLOW_MISSING_FIELDS_IN_HEADER, Options.INDEX_ON_THE_FLY))
+                .build()) {
+            fakeVCFFileWriter.writeHeader(header); // writes header
+            fakeVCFFileWriter.add(createVC(header));
+            fakeVCFFileWriter.add(createVC(header));
+        }
+
+        try (final VariantContextWriter fakeVCFBodyFileWriter = new VariantContextWriterBuilder()
+                .setOutputFile(fakeVCFBodyFile).setReferenceDictionary(sequenceDict)
+                .setOptions(EnumSet.of(Options.ALLOW_MISSING_FIELDS_IN_HEADER, Options.INDEX_ON_THE_FLY))
```
cmnbroad: Creating an index on one of these header-less files seems questionable. I wonder if we can/should prevent that, or if we should allow it.

src/test/java/htsjdk/variant/variantcontext/writer/VCFWriterUnitTest.java
```diff
@@ -133,6 +137,49 @@ public void testBasicWriteAndRead(final String extension) throws IOException {
...
 
+    /** test, using the writer and reader, that we can output and input a VCF body without problems */
+    @Test(dataProvider = "vcfExtensionsDataProvider")
+    public void testWriteAndReadVCFBody(final String extension) throws IOException {
+        final File fakeVCFFile = VariantBaseTest.createTempFile("testWriteAndReadVCFBody.", extension);
+        if (".vcf.gz".equals(extension)) {
+            new File(fakeVCFFile.getAbsolutePath() + ".tbi").deleteOnExit();
+        } else {
+            Tribble.indexFile(fakeVCFFile).deleteOnExit();
+        }
```
cmnbroad: It would be simpler to just create a temp directory (with deleteOnExit), and then just create the file in the temp dir. Then a lot of this code would be unnecessary.

src/test/java/htsjdk/variant/variantcontext/writer/VCFWriterUnitTest.java
```diff
@@ -133,6 +137,49 @@ public void testBasicWriteAndRead(final String extension) throws IOException {
 
     }
 
+    /** test, using the writer and reader, that we can output and input a VCF body without problems */
+    @Test(dataProvider = "vcfExtensionsDataProvider")
+    public void testWriteAndReadVCFBody(final String extension) throws IOException {
```
cmnbroad: testWriteAndReadVCFHeaderless

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
```diff
@@ -221,6 +221,35 @@ public void close() {
...
+                if ( GeneralUtils.DEBUG_MODE_ENABLED ) {
+                    System.err.println("No contig dictionary found in header, falling back to reference sequence dictionary");
+                }
+                createContigDictionary(VCFUtils.makeContigHeaderLines(getRefDict(), null));
+            } else {
+                throw new IllegalStateException("Cannot write BCF2 file with missing contig lines");
+            }
+        } else {
+            createContigDictionary(header.getContigLines());
+        }
```
cmnbroad: Much of this code is duplicated in writeHeader, and should be factored out into a common method called by both.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,253 @@
...
+        final File fakeVCFFile = VariantBaseTest.createTempFile("testWriteAndReadVCFBody.", ".bcf");
+
+        Tribble.indexFile(fakeVCFFile).deleteOnExit();
+        metaData = new HashSet<VCFHeaderLine>();
+        additionalColumns = new HashSet<String>();
+        final SAMSequenceDictionary sequenceDict = createArtificialSequenceDictionary();
+        final VCFHeader header = createFakeHeader(metaData, additionalColumns, sequenceDict);
+        try (final VariantContextWriter writer = new VariantContextWriterBuilder()
+                .setOutputFile(fakeVCFFile).setReferenceDictionary(sequenceDict)
+                .setOptions(EnumSet.of(Options.ALLOW_MISSING_FIELDS_IN_HEADER, Options.INDEX_ON_THE_FLY))
```
cmnbroad: BCF requires all fields to be present in the header. I'm a bit surprised this isn't rejected, but I think it should be removed for this test.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,253 @@
...
+
+        metaData = new HashSet<VCFHeaderLine>();
+        additionalColumns = new HashSet<String>();
+        final SAMSequenceDictionary sequenceDict = createArtificialSequenceDictionary();
+        final VCFHeader header = createFakeHeader(metaData, additionalColumns, sequenceDict);
+
+        // we write two files, fakeVCFFile with the header and body, and fakeVCFBodyFile with just the body
+        try (final VariantContextWriter fakeVCFFileWriter = new VariantContextWriterBuilder()
+                .setOutputFile(fakeVCFFile).setReferenceDictionary(sequenceDict)
+                .setOptions(EnumSet.of(Options.ALLOW_MISSING_FIELDS_IN_HEADER, Options.INDEX_ON_THE_FLY))
```
cmnbroad: Remove ALLOW_MISSING_FIELDS_IN_HEADER.

src/test/java/htsjdk/variant/variantcontext/writer/AsyncVariantContextWriterUnitTest.java
```diff
@@ -0,0 +1,169 @@
+/*
+* Copyright (c) 2012 The Broad Institute
```
cmnbroad: 2017.

src/test/java/htsjdk/variant/variantcontext/writer/AsyncVariantContextWriterUnitTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+    /**
+     * create a fake header of known quantity
+     * @param metaData           the header lines
+     * @param additionalColumns  the additional column names
+     * @return a fake VCF header
+     */
+    public static VCFHeader createFakeHeader(final Set<VCFHeaderLine> metaData, final Set<String> additionalColumns,
+                                             final SAMSequenceDictionary sequenceDict) {
+        metaData.add(new VCFHeaderLine(VCFHeaderVersion.VCF4_0.getFormatString(), VCFHeaderVersion.VCF4_0.getVersionString()));
```
cmnbroad: I think this can be removed, or at the least use VCF4_2.

src/test/java/htsjdk/variant/variantcontext/writer/AsyncVariantContextWriterUnitTest.java
```diff
@@ -0,0 +1,169 @@
...
+        tempDir = TestUtil.getTempDirectory("VCFWriter", "StaleIndex");
+    }
+
+    @AfterClass
+    private void deleteTemporaryDirectory() {
+        for (File f : tempDir.listFiles()) {
+            f.delete();
+        }
+        tempDir.delete();
+    }
```
cmnbroad: Same comment as above, use deleteOnExit.

src/test/java/htsjdk/variant/variantcontext/writer/VCFWriterUnitTest.java
```diff
@@ -133,6 +137,49 @@ public void testBasicWriteAndRead(final String extension) throws IOException {
...
+            AsciiLineReaderIterator iterator;
+            if (".vcf.gz".equals(extension)) {
+                iterator = new AsciiLineReaderIterator(new AsciiLineReader(bcis));
+            } else {
+                iterator = new AsciiLineReaderIterator(new AsciiLineReader(fis));
+            }
+            int counter = 0;
+            while (iterator.hasNext()) {
+                VariantContext context = codec.decode(iterator.next());
+                if (context != null) {
```
cmnbroad: Same comment as above. Can this be removed ?

src/test/java/htsjdk/variant/variantcontext/writer/AsyncVariantContextWriterUnitTest.java
```diff
@@ -0,0 +1,169 @@
...
+        codec.setVCFHeader(header, VCFHeaderVersion.VCF4_2);
+
+        try (FileInputStream fis = new FileInputStream(fakeVCFFile)) {
+            AsciiLineReaderIterator iterator;
+
+            iterator = new AsciiLineReaderIterator(new AsciiLineReader(fis));
+            int counter = 0;
+            while (iterator.hasNext()) {
+                VariantContext context = codec.decode(iterator.next());
+                if (context != null) {
```
cmnbroad: Not sure what this is guarding against ? If hasNet returns true I think next and decode should succeed. Can this be removed ?

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
cmnbroad: I just noticed a previous reviewer suggest the name "Body". I like Headerless a bit better since I don't think "body" is used this way in htsjdk.

src/main/java/htsjdk/variant/variantcontext/writer/AsyncVariantContextWriter.java
amilamanoj: Added a test case

amilamanoj: Added the test case

src/test/java/htsjdk/variant/variantcontext/writer/AsyncVariantContextWriterUnitTest.java
amilamanoj: Removed the null check.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,253 @@
...
+     * create a fake header of known quantity
+     *
+     * @param metaData          the header lines
+     * @param additionalColumns the additional column names
+     * @return a fake VCF header
+     */
+    public static VCFHeader createFakeHeader(final Set<VCFHeaderLine> metaData, final Set<String> additionalColumns,
+                                             final SAMSequenceDictionary sequenceDict) {
+        metaData.add(new VCFHeaderLine(VCFHeaderVersion.VCF4_0.getFormatString(),
+                                       VCFHeaderVersion.VCF4_0.getVersionString()));
```
amilamanoj: Removed.

src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
amilamanoj: @droazen I've changed it so that `writeHeader()` calls into `setVcfHeader()`.
@cmnbroad Yes that should work for our case, since we don't need to change the header once it's set. When rejecting, do we still allow repeated calls of `writeHeader()` and `setVcfHeader()` (individually, without calling them both) or just simply check if `header` has been set once before?

cmnbroad: @amilamanoj I originally thought the right thing would be for both methods to reject if the header is already set, but changing writeHeader to work that way could break existing code. So perhaps we should just make setVcfHeader enforce that, and leave writeHeader's current promiscuous behavior as is.

cmnbroad: @amilamanoj Actually, on giving this more thought, any code that currently calls writeHeader more than once is already on a bad path, so I would vote for rejecting in either case if a header exists.

amilamanoj: @cmnbroad Thanks for the feedback. I've also discussed this with @cyenyxe. We also think changing header should be prevented if the header (or just a part of body, in our use case) is already written to the output stream, since there's only one output stream per writer instance. But setting the header many times should not matter as long as nothing is written to the output stream yet. I've added some checks along with tests, please take a look and see if it works.

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
amilamanoj: Duplicates have bee removed.

src/test/java/htsjdk/variant/variantcontext/writer/VCFWriterUnitTest.java
amilamanoj: Done.

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
amilamanoj: Implemented.

src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriter.java
amilamanoj: Done.

src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
amilamanoj: @cmnbroad @droazen @yfarjoun Added the same checks for BCF2Writer as well. I think with this commit I've addressed all the comments so far. Waiting for further feedback.

src/test/java/htsjdk/variant/variantcontext/writer/VCFWriterUnitTest.java
amilamanoj: Done.

src/test/java/htsjdk/variant/variantcontext/writer/VCFWriterUnitTest.java
amilamanoj: Done.

src/test/java/htsjdk/variant/variantcontext/writer/VCFWriterUnitTest.java
amilamanoj: Used 4.2.

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
```diff
@@ -122,6 +122,12 @@
     private VCFHeader lastVCFHeaderOfUnparsedGenotypes = null;
     private boolean canPassOnUnparsedGenotypeDataForLastVCFHeader = false;
 
+    // is the header written to the output stream?
+    private boolean isHeaderWritten;
+
+    // is at least one part of body written to the output stream?
+    private boolean isBodyWritten;
+
```
cmnbroad: @amilamanoj Is there a reason that the writers need to keep two state variables to track what is essentially a single binary state (the header is frozen) ?

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
```diff
@@ -145,46 +151,24 @@ public BCF2Writer(final File location, final OutputStream output, final SAMSeque
...
-        } else {
-            createContigDictionary(header.getContigLines());
-        }
-
-        // set up the map from dictionary string values -> offset
-        final ArrayList<String> dict = BCF2Utils.makeDictionary(header);
-        for ( int i = 0; i < dict.size(); i++ ) {
-            stringDictionaryMap.put(dict.get(i), i);
+        if (isHeaderWritten) {
+            throw new IllegalStateException("Header is already written");
```
cmnbroad: This code block seems redundant - the next statement is a call to setVcfheader, which handles this.

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
```diff
@@ -145,46 +151,24 @@ public BCF2Writer(final File location, final OutputStream output, final SAMSeque
...
         }
-
-        sampleNames = header.getGenotypeSamples().toArray(new String[header.getNGenotypeSamples()]);
-
-        // setup the field encodings
-        fieldManager.setup(header, encoder, stringDictionaryMap);
+        setVcfHeader(header);
```
cmnbroad: Also, even though htsjdk is not completely consistent about this, can we name these methods "setVCFHeader" instead of "setVcfHeader"?

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
```diff
@@ -221,6 +206,43 @@ public void close() {
         super.close();
     }
 
+    @Override
+    public void setVcfHeader(VCFHeader header) {
```
cmnbroad: the param should be marked final

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
```diff
@@ -221,6 +206,43 @@ public void close() {
...
+        if (isHeaderWritten) {
+            throw new IllegalStateException("Header is already written");
+        }
+        if (isBodyWritten) {
+            throw new IllegalStateException("Body is already written");
+        }
+        // make sure the header is sorted correctly
+        header = new VCFHeader(header.getMetaDataInSortedOrder(), header.getGenotypeSamples());
+
+        this.header = doNotWriteGenotypes ? new VCFHeader(header.getMetaDataInSortedOrder()) : header;
```
cmnbroad: I don't think we should need to make a copy of the header twice here.

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
```diff
@@ -221,6 +206,43 @@ public void close() {
...
     }
 
+    @Override
+    public void setVcfHeader(VCFHeader header) {
+        if (isHeaderWritten) {
+            throw new IllegalStateException("Header is already written");
+        }
+        if (isBodyWritten) {
+            throw new IllegalStateException("Body is already written");
+        }
```
cmnbroad: As mentioned above, I think we need only one state variable. If we can get away with that, this message can say something like "The header cannot be modified after the header or variants have been written to the output stream."

src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
```diff
@@ -128,13 +134,17 @@ private void writeAndResetBuffer() throws IOException {
 
     @Override
     public void writeHeader(final VCFHeader header) {
+
+        if (isHeaderWritten) {
+            throw new IllegalStateException("Header is already written");
+        }
```
cmnbroad: Also redundant - setVcfHeader checks this.

src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
```diff
@@ -66,6 +66,12 @@
     // should we always output a complete format record, even if we could drop trailing fields?
     private final boolean writeFullFormatField;
 
+    // is the header written to the output stream?
+    private boolean isHeaderWritten;
+
+    // is at least one part of body written to the output stream?
+    private boolean isBodyWritten;
+
```
cmnbroad: Sam comment as above regarding using a single state variable.

src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
```diff
@@ -223,9 +231,21 @@ public void add(final VariantContext context) {
...
     }
+
+    @Override
+    public void setVcfHeader(VCFHeader header) {
+        if (isHeaderWritten) {
+            throw new IllegalStateException("Header is already written");
+        }
+        if (isBodyWritten) {
+            throw new IllegalStateException("Body is already written");
+        }
```
cmnbroad: Same comments as above about using a single state variable, more specific error message, and final args.

src/main/java/htsjdk/variant/variantcontext/writer/AsyncVariantContextWriter.java
```diff
@@ -53,4 +53,10 @@ public void writeHeader(final VCFHeader header) {
     public boolean checkError() {
         return false;
     }
+
+    @Override
+    public void setVcfHeader(VCFHeader header) {
```
cmnbroad: The existing code isn't consistent about this, but we generally mark arguments as final in new code. Thanks.

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
```diff
@@ -145,46 +151,24 @@ public BCF2Writer(final File location, final OutputStream output, final SAMSeque
...
-        // setup the field encodings
-        fieldManager.setup(header, encoder, stringDictionaryMap);
+        setVcfHeader(header);
 
         try {
             // write out the header into a byte stream, get its length, and write everything to the file
             final ByteArrayOutputStream capture = new ByteArrayOutputStream();
             final OutputStreamWriter writer = new OutputStreamWriter(capture);
-            this.header = VCFWriter.writeHeader(header, writer, doNotWriteGenotypes, VCFWriter.getVersionLine(), "BCF2 stream");
+            this.header = VCFWriter.writeHeader(header, writer, VCFWriter.getVersionLine(), "BCF2 stream");
```
cmnbroad: I believe this code is using the wrong header now. Previous to this PR, this method updated the "header" argument variable (questionable practice) with the updated header, and then passed that to the writer. With these changes, it call setVcfHeader, which updates the header member variable with the updated header, but then passes the (now stale) input argument "header" to the writer.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,283 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * @author amila
+ *         <p/>
+ *         Class BCF2WriterUnitTest
+ *         <p/>
+ *         This class tests out the ability of the VCF writer to correctly write VCF files
```
cmnbroad: I think this comment is still wrong - this is testing that the BCF writer works correctly.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,283 @@
...
+        final VariantContextWriter writer2 = new VariantContextWriterBuilder()
+                .setOutputFile(bcfOutputFile).setReferenceDictionary(sequenceDict)
+                .setOptions(EnumSet.of(Options.INDEX_ON_THE_FLY))
+                .build();
+        writer2.writeHeader(header);
+        try {
+            writer2.writeHeader(header);
+            Assert.fail("Should not allow writing header twice");
+        } catch (Exception e) {
+            Assert.assertTrue(e instanceof IllegalStateException);
```
cmnbroad: This test (prevent writing the header twice) should be factored out into a separate test method that declares IllegalStateException as an expectedException in the test annotation (see example [here](https://github.com/samtools/htsjdk/blob/656dc24930e1dc7fd0abd4c5e2abedf6f5074b35/src/test/java/htsjdk/samtools/BAMFileWriterTest.java#L146)). It eliminates the need to write a try/catch block and cleanly separates the individual tests from each other.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,283 @@
...
+                .setOutputFile(bcfOutputFile).setReferenceDictionary(sequenceDict)
+                .setOptions(EnumSet.of(Options.INDEX_ON_THE_FLY))
+                .build();
+        writer3.writeHeader(header);
+        try {
+            writer3.setVcfHeader(header);
+            Assert.fail("Should not allow changing header if header is already written");
+        } catch (Exception e) {
+            Assert.assertTrue(e instanceof IllegalStateException);
+        }
```
cmnbroad: Same comment here, and in the other test files. Each individual test should be a separate method.

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
amilamanoj: I thought of providing two distinct error messages based on which one (header/body) is written (In our use case, for example, we don't write the header). But I can merge the two variables if it's not important.

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
amilamanoj: Yes, will remove this.

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
amilamanoj: Sure.

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
magicDGS: What's about using an `int` variable for different error messages? If different to -1, there is an error, and that can be switch to different messages...

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
amilamanoj: Changed it to use the member variable.

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
amilamanoj: @magicDGS I didn't understand how exactly you meant using an `int` variable. For now I merged the two variables. We can change it later if needed.

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
amilamanoj: Done.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
amilamanoj: Done.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
amilamanoj: Done.

src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
```diff
@@ -128,13 +131,14 @@ private void writeAndResetBuffer() throws IOException {
...
     @Override
     public void writeHeader(final VCFHeader header) {
+
         // note we need to update the mHeader object after this call because they header
         // may have genotypes trimmed out of it, if doNotWriteGenotypes is true
         try {
-            this.mHeader = writeHeader(header, writer, doNotWriteGenotypes, getVersionLine(), getStreamName());
-            this.vcfEncoder = new VCFEncoder(this.mHeader, this.allowMissingFieldsInHeader, this.writeFullFormatField);
+            setVCFHeader(header);
+            writeHeader(header, writer, getVersionLine(), getStreamName());
```
cmnbroad: @amilamanoj This one also is using the stale (local) header. It should use the member variable.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,301 @@
...
+
+    @Test(expectedExceptions = IllegalStateException.class)
+    public void testChangeHeaderAfterWritingBody() {
+        final File bcfOutputFile = VariantBaseTest.createTempFile("testWriteAndReadVCF.", ".bcf");
+
+        Tribble.indexFile(bcfOutputFile).deleteOnExit();
+        Set<VCFHeaderLine> metaData = new HashSet<VCFHeaderLine>();
+        Set<String> additionalColumns = new HashSet<String>();
+        final SAMSequenceDictionary sequenceDict = createArtificialSequenceDictionary();
+        final VCFHeader header = createFakeHeader(metaData, additionalColumns, sequenceDict);
```
cmnbroad: It looks like all of the calls to createFakeHeader pass in the same empty metadata and additional columns lists, and fake seq dictionary, so this can be much simplified by just removing those and returning the fake header.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,301 @@
...
+
+    /**
+     * create a fake VCF record
+     *
+     * @param header the VCF header
+     * @return a VCFRecord
+     */
+    private VariantContext createVC(final VCFHeader header) {
+
+        return createVCGeneral(header, "1", 1);
```
cmnbroad: This method should just be collapsed into createVCGeneral (it seems like all calls are identical).

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,301 @@
...
+        final VCFHeader header = createFakeHeader(metaData, additionalColumns, sequenceDict);
+
+        // prevent changing header if it's already written
+        final VariantContextWriter writer2 = new VariantContextWriterBuilder()
+                .setOutputFile(bcfOutputFile).setReferenceDictionary(sequenceDict)
+                .setOptions(EnumSet.of(Options.INDEX_ON_THE_FLY))
+                .build();
+        writer2.writeHeader(header);
+        writer2.setVCFHeader(header);
+        writer2.add(createVC(header));
```
cmnbroad: This call is unnecessary for the test - the exception gets thrown in the previous stmt.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,301 @@
...
+        writer1.writeHeader(header);
+        writer1.add(createVC(header));
+        writer1.close();
+    }
+
+    @Test(expectedExceptions = IllegalStateException.class)
+    public void testChangeHeaderAfterWritingHeader() {
+        final File bcfOutputFile = VariantBaseTest.createTempFile("testWriteAndReadVCF.", ".bcf");
+
+        Tribble.indexFile(bcfOutputFile).deleteOnExit();
```
cmnbroad: The reason I suggested creating a temporary directory in the earlier review pass was so this kind of call would be unnecessary.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,301 @@
...
+        Set<VCFHeaderLine> metaData = new HashSet<VCFHeaderLine>();
+        Set<String> additionalColumns = new HashSet<String>();
+        final SAMSequenceDictionary sequenceDict = createArtificialSequenceDictionary();
+        final VCFHeader header = createFakeHeader(metaData, additionalColumns, sequenceDict);
+
+        // prevent changing header if it's already written
+        final VariantContextWriter writer2 = new VariantContextWriterBuilder()
+                .setOutputFile(bcfOutputFile).setReferenceDictionary(sequenceDict)
+                .setOptions(EnumSet.of(Options.INDEX_ON_THE_FLY))
```
cmnbroad: I think both the sequence dictionary and the INDEX_ON_THE_FLY options are unnecessary for this test, and can be removed. were just trying to test that we throw if you change the header.

cmnbroad:
@amilamanoj I added a couple more comments on the tests (most of which apply to both test files, though I only put comment on one of them), since I think they can be simplified a bit. In general we should only include the minimal code necessary for the specific test. Thanks in advance for your patience in seeing these through.

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
magicDGS: Nevermind @amilamanoj, it was just a proposal if you really want to be more verbose about the error and what happened. Just FYI, this was what I meant:

```java
// this integer indicates the status for writing the header down
// -1 indicates no problem to write the header
// 0 means header was already written
// 1 means body was already written
// otherwise, this is an unexpected error
private int shouldWriteHeader = -1;

@Override
public void setVCFHeader(final VCFHeader header) {
    if (shouldWriteHeader != -1) {
        final String msg;
        switch(shouldWriteHeader) {
            case 0: msg = "header already written"; break;
            case 1: msg = "body already written"; break;
            default: msg = "unexpected error code (" + shouldWriteHeader + ")";
        }
        throw new IllegalStateException("The header cannot be modified after the header or variants have been written to the output stream. Error was: " + msg);
    }

    ....

}
```

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
cyenyxe: In fact, JUnit supports that via rules: http://junit.org/junit4/javadoc/4.12/org/junit/rules/TemporaryFolder.html

src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
amilamanoj: Fixed.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
amilamanoj: Done.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
amilamanoj: Removed.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
amilamanoj: Fixed.

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
amilamanoj: Removed INDEX_ON_THE_FLY option. But sequence dictionary has to be set because INDEX_ON_THE_FLY is a default option anyway, and without setting the sequence dictionary, writer cannot be created (an exception gets thrown).

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
amilamanoj: Do you mean something like this? `this.header = doNotWriteGenotypes ? new VCFHeader(header.getMetaDataInSortedOrder()) : new VCFHeader(header.getMetaDataInSortedOrder(), header.getGenotypeSamples());`

cmnbroad: @amilamanoj Yes, thats what I meant.

amilamanoj: @cmnbroad Fixed.

amilamanoj:
@cmnbroad @droazen I think all comments so far have been addressed. Do you have more feedback or does the PR look OK now?

cmnbroad:
@amilamanoj I need to do another review pass on the updated changes. Sorry for the long turnaround time, but we have quite a backlog of PRs. I'll do it as soon as I can; hopefully it will be this week.

src/test/java/htsjdk/variant/variantcontext/writer/VCFWriterUnitTest.java
```diff
@@ -133,6 +129,95 @@ public void testBasicWriteAndRead(final String extension) throws IOException {
...
+        codec.setVCFHeader(header, VCFHeaderVersion.VCF4_2);
+
+        try (BlockCompressedInputStream bcis = new BlockCompressedInputStream(fakeVCFFile);
+                FileInputStream fis = new FileInputStream(fakeVCFFile)) {
+            AsciiLineReaderIterator iterator;
+            if (".vcf.gz".equals(extension)) {
+                iterator = new AsciiLineReaderIterator(new AsciiLineReader(bcis));
+            } else {
+                iterator = new AsciiLineReaderIterator(new AsciiLineReader(fis));
+            }
```
cmnbroad: Ternary operator cold make this all much cleaner and easier to read:
```
AsciiLineReaderIterator iterator =
    new AsciiLineReaderIterator(new AsciiLineReader(".vcf.gz".equals(extension) ? bcis : fis));

```

src/test/java/htsjdk/variant/variantcontext/writer/AsyncVariantContextWriterUnitTest.java
```diff
@@ -0,0 +1,155 @@
...
+            genotypes.add(gt);
+        }
+        return new VariantContextBuilder("RANDOM", chrom, position, position, alleles)
+                .genotypes(genotypes).attributes(attributes).make();
+    }
+
+
+
+
+}
```
cmnbroad: Can we get rid of all of the blank lines?

amilamanoj:
@cmnbroad Thanks for the feedback. I've addressed your comments.

amilamanoj:
@cmnbroad I'm waiting for any further feedback.

cmnbroad:
@amilamanoj I haven't forgotten that this needs another review pass, and will get to it as soon as possible. Realistically, that will not be until next week though.

cmnbroad:
This still needs a final review. In addition to enabling the original motivating (htsget) use case, it would also allow us to get rid of [this Hadoop-BAM hack](https://github.com/HadoopGenomics/Hadoop-BAM/blob/fd006477611935a3560ba33028a167fb31d12964/src/main/java/org/seqdoop/hadoop_bam/VCFRecordWriter.java#L152).

src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriter.java
```diff
@@ -49,4 +55,13 @@
     public boolean checkError();
     
     public void add(VariantContext vc);
+
+    /**
```
lbergelson: It seems like the javadoc should mention that you should call exactly one of `writeHeader()` / `setHeader()`

src/main/java/htsjdk/variant/variantcontext/writer/BCF2Writer.java
```diff
@@ -122,6 +122,10 @@
     private VCFHeader lastVCFHeaderOfUnparsedGenotypes = null;
     private boolean canPassOnUnparsedGenotypeDataForLastVCFHeader = false;
 
+    // is the header or body written to the output stream?
+    private boolean isWrittenToOutput;
```
lbergelson: I think the name `outputHasBeenWritten` would be more clear.   Can you change this in all places where this variable has been added?

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,275 @@
...
+        tempDir.deleteOnExit();
+    }
+
+
+    /**
+     * test, using the writer and reader, that we can output and input BCF without problems
+     */
+    @Test
+    public void testWriteAndReadBCF() throws IOException {
+        final File bcfOutputFile = File.createTempFile("testWriteAndReadVCF.", ".bcf", tempDir);
```
lbergelson: could you mark this for deletion?

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,275 @@
...
+
+    }
+
+
+    /**
+     * test, with index-on-the-fly option, that we can output and input BCF without problems
+     */
+    @Test
+    public void testWriteAndReadBCFWithIndex() throws IOException {
+        final File bcfOutputFile = File.createTempFile("testWriteAndReadVCF.", ".bcf", tempDir);
```
lbergelson: same with this

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,275 @@
...
+        }
+        Assert.assertEquals(counter, 2);
+    }
+
+    /**
+     * test, using the writer and reader, that we can output and input a BCF body without header
+     */
+    @Test
+    public void testWriteAndReadBCFHeaderless() throws IOException {
+        final File bcfOutputFile = File.createTempFile("testWriteAndReadBCFWithHeader.", ".bcf", tempDir);
```
lbergelson: same with these

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,275 @@
...
+                counter++;
+            }
+            Assert.assertEquals(counter, 2);
+        }
+
+    }
+
+    @Test(expectedExceptions = IllegalStateException.class)
+    public void testWriteHeaderTwice() throws IOException {
+        final File bcfOutputFile = File.createTempFile("testWriteAndReadVCF.", ".bcf", tempDir);
```
lbergelson: same here

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,275 @@
...
+                .unsetOption(Options.INDEX_ON_THE_FLY)
+                .build()) {
+            writer1.writeHeader(header);
+            writer1.writeHeader(header);
+        }
+    }
+
+    @Test(expectedExceptions = IllegalStateException.class)
+    public void testChangeHeaderAfterWritingHeader() throws IOException {
+        final File bcfOutputFile = File.createTempFile("testWriteAndReadVCF.", ".bcf", tempDir);
```
lbergelson: same here

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,275 @@
...
+                .unsetOption(Options.INDEX_ON_THE_FLY)
+                .build()) {
+            writer2.writeHeader(header);
+            writer2.setVCFHeader(header);
+        }
+    }
+
+    @Test(expectedExceptions = IllegalStateException.class)
+    public void testChangeHeaderAfterWritingBody() throws IOException {
+        final File bcfOutputFile = File.createTempFile("testWriteAndReadVCF.", ".bcf", tempDir);
```
lbergelson: same here

src/test/java/htsjdk/variant/bcf2/BCF2WriterUnitTest.java
```diff
@@ -0,0 +1,275 @@
...
+
+    }
+
+    @Test(expectedExceptions = IllegalStateException.class)
+    public void testWriteHeaderTwice() throws IOException {
+        final File bcfOutputFile = File.createTempFile("testWriteAndReadVCF.", ".bcf", tempDir);
+
+        final VCFHeader header = createFakeHeader();
+        // prevent writing header twice
+        try (final VariantContextWriter writer1 = new VariantContextWriterBuilder()
```
lbergelson: I would drop the number from these writers 1, 2, and 3

src/test/java/htsjdk/variant/variantcontext/writer/AsyncVariantContextWriterUnitTest.java
```diff
@@ -0,0 +1,145 @@
...
+    @BeforeClass
+    private void createTemporaryDirectory() {
+        File tempDir = TestUtil.getTempDirectory("VCFWriter", "StaleIndex");
+        tempDir.deleteOnExit();
+    }
+
+    /** test, using the writer and reader, that we can output and input a VCF body without problems */
+    @Test
+    public void testWriteAndReadAsyncVCFHeaderless() throws IOException {
+        final File fakeVCFFile = VariantBaseTest.createTempFile("testWriteAndReadAsyncVCFHeaderless.", ".vcf");
```
lbergelson: again, sadly all the test files have to be manually marked as deleteOnExit()

src/test/java/htsjdk/variant/variantcontext/writer/VCFWriterUnitTest.java
```diff
@@ -77,28 +80,21 @@
     @BeforeClass
     private void createTemporaryDirectory() {
         tempDir = TestUtil.getTempDirectory("VCFWriter", "StaleIndex");
+        tempDir.deleteOnExit();
```
lbergelson: I'm not sure this pattern works.  The older `deleteTemporaryDirectory` method that was called previously seems more likely to work.  I would just mark all of the temporary files as `deleteOnExit()`

amilamanoj:
@lbergelson Thanks for the feedback. I've addressed your comments.

lbergelson:
@amilamanoj Thank you for sticking through with this!  This is a useful change, I'm not sure how it became so complicated with so many reviewers.

lbergelson:
@tomwhite This is now merged into htjskd and we may want to update hadoop bam to take advantage of it.  Although that will probably require a release and a coordinated change with GATK since this breaks our GVCF writer.

PR #905, opened 2017-06-20
Sort and group order not updated when using setAttribute.

nh13:
@yfarjoun could you take a look since this fixes a bug you introduced, and @tfenne can a maintainer take a look?

The issue was that folks can call `setAttribute("SO", "queryname")` and the cached sort order introduced in ad8aa02aa2ed02b9cb2bb8801c3e6897e91ec3f6 would not get updated.  Similarly for the group order.

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMFileHeader.java
+++ b/src/main/java/htsjdk/samtools/SAMFileHeader.java
@@ -325,12 +325,11 @@ public class SAMFileHeader extends AbstractSAMHeaderRecord
     @Override
     public void setAttribute(final String key, final String value) {
         if (key.equals(SORT_ORDER_TAG)) {
-            this.setSortOrder(SortOrder.valueOf(value));
+            this.sortOrder = null;
         } else if (key.equals(GROUP_ORDER_TAG)) {
-            this.setGroupOrder(GroupOrder.valueOf(value));
-        } else {
-            super.setAttribute(key, value);
+            this.groupOrder = null;
         }
+        super.setAttribute(key, value);
     }
 
     /**
--- a/src/main/java/htsjdk/samtools/SAMFileHeader.java
+++ b/src/main/java/htsjdk/samtools/SAMFileHeader.java
@@ -307,7 +307,6 @@ public class SAMFileHeader extends AbstractSAMHeaderRecord
     @Deprecated
     @Override
     public void setAttribute(final String key, final Object value) {
-        super.setAttribute(key, value);
         if (key.equals(SORT_ORDER_TAG) || key.equals(GROUP_ORDER_TAG)) {
             this.setAttribute(key, value.toString());
         } else {
@@ -325,10 +324,11 @@ public class SAMFileHeader extends AbstractSAMHeaderRecord
      */
     @Override
     public void setAttribute(final String key, final String value) {
-        super.setAttribute(key, value);
-        if (key.equals(SORT_ORDER_TAG)) this.setSortOrder(SortOrder.valueOf(value));
-        else if (key.equals(GROUP_ORDER_TAG)) this.setGroupOrder(GroupOrder.valueOf(value));
-        else super.setAttribute(key, value);
+        if (key.equals(SORT_ORDER_TAG)) {
+            this.setSortOrder(SortOrder.valueOf(value));
+        } else if (key.equals(GROUP_ORDER_TAG)) {
+            this.setGroupOrder(GroupOrder.valueOf(value));
+        } else super.setAttribute(key, value);
     }
 
     /**
--- a/src/main/java/htsjdk/samtools/SAMFileHeader.java
+++ b/src/main/java/htsjdk/samtools/SAMFileHeader.java
@@ -328,7 +328,9 @@ public class SAMFileHeader extends AbstractSAMHeaderRecord
             this.setSortOrder(SortOrder.valueOf(value));
         } else if (key.equals(GROUP_ORDER_TAG)) {
             this.setGroupOrder(GroupOrder.valueOf(value));
-        } else super.setAttribute(key, value);
+        } else {
+            super.setAttribute(key, value);
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/htsjdk/samtools/SAMFileHeader.java
```diff
@@ -292,7 +292,43 @@ public GroupOrder getGroupOrder() {
...
+     * @deprecated Use {@link #setAttribute(String, String) instead
+     */
+    @Deprecated
+    @Override
+    public void setAttribute(final String key, final Object value) {
+        super.setAttribute(key, value);
+        if (key.equals(SORT_ORDER_TAG) || key.equals(GROUP_ORDER_TAG)) {
+            this.setAttribute(key, value.toString());
+        } else {
+            super.setAttribute(key, value);
```
pshapiro4broad: This is calling `super.setAttribute()` twice is that intentional?

src/main/java/htsjdk/samtools/SAMFileHeader.java
```diff
@@ -292,7 +292,43 @@ public GroupOrder getGroupOrder() {
...
+     * Set the given value for the attribute named 'key'.  Replaces an existing value, if any.
+     * If value is null, the attribute is removed.
+     * Supported types are Character, Integer, Float and String.  Byte and Short may also be
+     * passed in but they will be converted to Integer.
+     * @param key attribute name
+     * @param value attribute value
+     */
+    @Override
+    public void setAttribute(final String key, final String value) {
+        super.setAttribute(key, value);
```
pshapiro4broad: Also called twice here.

src/main/java/htsjdk/samtools/SAMFileHeader.java
```diff
@@ -292,7 +292,43 @@ public GroupOrder getGroupOrder() {
...
+        if (key.equals(SORT_ORDER_TAG)) this.setSortOrder(SortOrder.valueOf(value));
```
pshapiro4broad: This code should use `{` `}` around `if`/`else` as well.

src/main/java/htsjdk/samtools/SAMFileHeader.java
nh13: Ugh, left over from trying to make the first commit's test fail.  Good catch.

nh13:
@pshapiro4broad thanks for spotting those!

src/main/java/htsjdk/samtools/SAMFileHeader.java
```diff
@@ -325,10 +324,11 @@ public void setAttribute(final String key, final Object value) {
...
     public void setAttribute(final String key, final String value) {
-        super.setAttribute(key, value);
-        if (key.equals(SORT_ORDER_TAG)) this.setSortOrder(SortOrder.valueOf(value));
-        else if (key.equals(GROUP_ORDER_TAG)) this.setGroupOrder(GroupOrder.valueOf(value));
-        else super.setAttribute(key, value);
+        if (key.equals(SORT_ORDER_TAG)) {
+            this.setSortOrder(SortOrder.valueOf(value));
+        } else if (key.equals(GROUP_ORDER_TAG)) {
+            this.setGroupOrder(GroupOrder.valueOf(value));
+        } else super.setAttribute(key, value);
```
nh13: will @pshapiro4broad allow this, or should I put more brackets?

nh13: I put more brackets

pshapiro4broad: lol thanks @nh13 
I think it's 👍 but someone who actually understands the code should probably look at it too.

nh13:
@yfarjoun this is now breaking other tests (namely `tagCorrectlyProcessTest` in `ValidateSamFileTest`).  I feel like reverting your commit to remove the internal references to sort and group order.  What do you think?

nh13:
@yfarjoun or @tfenne care to take a quick final peak?

src/main/java/htsjdk/samtools/SAMFileHeader.java
```diff
@@ -292,7 +292,44 @@ public GroupOrder getGroupOrder() {
...
+            this.setAttribute(key, value.toString());
+        } else {
+            super.setAttribute(key, value);
+        }
+    }
+
+    /**
+     * Set the given value for the attribute named 'key'.  Replaces an existing value, if any.
+     * If value is null, the attribute is removed.
+     * Supported types are Character, Integer, Float and String.  Byte and Short may also be
```
tfenne: The parameter type is String though?

src/test/java/htsjdk/samtools/SAMFileHeaderTest.java
```diff
@@ -0,0 +1,52 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ */
+package htsjdk.samtools;
+
+import htsjdk.HtsjdkTest;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class SAMFileHeaderTest extends HtsjdkTest {
+
```
tfenne: Could you also just add lines to the test that ensure that header.getAttribute("SO"/"GO") do the right things after each of the setting methods?

src/main/java/htsjdk/samtools/SAMFileHeader.java
nh13: Will fix this here and in `AbstractSAMHeaderRecord` where I copied it from.

src/test/java/htsjdk/samtools/SAMFileHeaderTest.java
nh13: Of course!

nh13:
@tfenne I think this is ready for a squash and merge after your thumbs up.

PR #906, opened 2017-06-20
Fix bugs in IndexFactory and BinningIndexBuilder.

cmnbroad:
### Description

Fixes https://github.com/samtools/htsjdk/issues/393 and a few other issues when creating indices through IndexFactory.create*Index methods that take a file:

- Block gzipped inputs were wrapped with PositionalBufferedStreams, resulting in byte offsets being handed to the indexer rather than virtual file pointers.
- The first feature was always indexed as if it were at offset 0, even if it was preceded by a header. Most of the time this worked, since the feature iterator thats used to read the indexed file would scan right through the header. But it failed when the header was large (i.e. hg38 large), since the scanner would be short-circuited by the linear index before the header was exhausted.
- The BinningIndexBuilder assumed that 0 was never a legitimate feature location, but a file with no header at all (such as a header-less BED file) can have features at offset/virtual-file-pointer 0, even if gzipped.

 

### Checklist

- [ X] Code compiles correctly
- [ X] New tests covering changes and new functionality
- [ X] All tests passing


src/main/java/htsjdk/tribble/readers/BlockCompressedAsciiLineReader.java
```diff
@@ -0,0 +1,42 @@
...
+import htsjdk.samtools.util.BlockCompressedInputStream;
+import htsjdk.samtools.util.CloserUtil;
+
+import java.io.IOException;
+
+/**
+ * An AsciiLineReader implementation that wraps a BlockCompressedInputStream and provides no additional buffering.
+ * Useful for cases where we need to preserve virtual file pointers in the underlying stream, such as during indexing.
+ */
+public class BlockCompressedAsciiLineReader extends AsciiLineReader {
```
magicDGS: If the static method is done in `AsciiLineReader`, this could be a private inner class to avoid direct usages.

src/main/java/htsjdk/tribble/readers/PositionalBufferedStream.java
```diff
@@ -28,7 +28,11 @@
 
 /**
  * A wrapper around an {@code InputStream} which performs it's own buffering, and keeps track of the position.
- * 
+ *
+ * TODO: This class implements Positional, which in turn extends LocationAware, which requires preservation of
```
magicDGS: This could be fixed by throwing an `IllegalArgumentException` if the `InputStream` is a `BlockCompressedInputStream` as a temporary solution. It is much better than violate the contract and assume that it is working as expected...

src/test/java/htsjdk/tribble/index/IndexTest.java
```diff
@@ -92,6 +88,7 @@ public void testWriteIndex(final File inputFile, final IndexFactory.IndexType ty
         Assert.assertTrue(tempIndex.exists());
         // load the generated index
         final Index loadedIndex = IndexFactory.loadIndex(tempIndex.getAbsolutePath());
+        //TODO: This not very useful; the test can pass even if the generated index is unuseable for queries.
```
magicDGS: I implemented this test only for checking if writing the index does not blow up, and if the in-memory index and the one written have the same contents. For testing if it is usable for queries, it should be a different test (which actually does not require to write the index down, I guess).

src/test/java/htsjdk/tribble/index/tabix/TabixIndexTest.java
```diff
@@ -134,6 +148,73 @@ public void testQueryProvidedItemsAmount() throws IOException {
...
+                        new File(TestUtils.DATA_DIR, "bed/2featuresWithHeader.bed.gz"),
+                        Arrays.asList(
+                                new Interval("chr1", 1, 10),
+                                new Interval("chr1", 100, 1000000))
+                },
+        };
+    }
+
+    @Test(dataProvider = "bedTabixIndexTestData")
+    public void testBedTabixIndex(
```
magicDGS: Nice test!

src/test/java/htsjdk/tribble/index/tabix/TabixIndexTest.java
```diff
@@ -102,6 +114,8 @@ public void testQueryProvidedItemsAmount() throws IOException {
         ) // create TabixIndex straight from plaintext VCF
                 .write(plainTextVcfIndexFile); // write it
 
+        //TODO: you can pass in a .tbi file as the index for a plain .vcf, but if you *don't* pass in the file name and
```
magicDGS: But is it correct to use a **.tbi** index with a plain text file? As you noted in a previous comment, this is not allowed by tabix... I suggest to blow up in this case!

src/main/java/htsjdk/tribble/readers/PositionalBufferedStream.java
cmnbroad: I agree that would be the right thing to do, but htsjdk is kind of painted into a corner at this point. There are several other code paths that would have to change, but worse, BinaryFeatureCodec (and thus BCF2Codec) use PositionalBufferedStream for a SOURCE type parameter, even though BCF can be block compressed. So that codec actually depends on being able to wrap a PositionalBufferedStream around a block compressed stream.

src/main/java/htsjdk/tribble/AsciiFeatureCodec.java
```diff
@@ -49,14 +50,19 @@ public boolean isDone(final LineIterator lineIterator) {
     }
 
     @Override
-    public LocationAware makeIndexableSourceFromStream(final InputStream bufferedInputStream) {
-        final PositionalBufferedStream pbs;
-        if (bufferedInputStream instanceof PositionalBufferedStream) {
-            pbs = (PositionalBufferedStream) bufferedInputStream;
+    public LocationAware makeIndexableSourceFromStream(final InputStream inputStream) {
+        if (inputStream instanceof BlockCompressedInputStream) {
```
cmnbroad: I think that would put an unnecessary constraint on AsciiLineReader (it should be legal to wrap an AsciiLineReader around a BCS, though having said that, its confusing because AsciiLineReader confers both a line orientation as well as buffering, which is what messes up the positions). 

Also, given how brittle all this code is, I like the idea of having this behavior and the comments explicitly included right here, where the explicit control over the buffering/positioning matters (this method is specifically to support indexing).

src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
```diff
@@ -40,11 +40,19 @@
...
 
+    protected AsciiLineReader() {};
+
+    /**
+     * TODO: This class implements LocationAware, which requires preservation of virtual file pointers on BGZF inputs.
+     * However, if the inputStream wrapped by this class is a BlockCompressedInputStream, it violates that contract by
+     * wrapping the stream and returning positional file offsets instead.
+     */
     public AsciiLineReader(final InputStream is){
+        // NOTE: This will wrap the input stream in a PositionalBufferedStream even if its already a PositionalBufferedStream
```
cmnbroad: Not sure it should be illegal to pass a BCS to AsciiLineReader (my comment that it violates the contract might be an overstatement). Perhaps we should just change the doc to be more explicit about the fact that it provides buffering, which affects the positions returned.

src/main/java/htsjdk/tribble/AsciiFeatureCodec.java
```diff
@@ -49,14 +50,19 @@ public boolean isDone(final LineIterator lineIterator) {
...
+            // ensure that the correct (virtual file pointer) positions returned from this stream are preserved for
+            // the indexer. We can't used AsciiLineReader in this case since it wraps the input stream with a
+            // PositionalBufferedInputStream.
+            return new AsciiLineReaderIterator(new BlockCompressedAsciiLineReader((BlockCompressedInputStream) inputStream));
+        } else if (inputStream instanceof PositionalBufferedStream) {
+            // if this is already a PositionalBufferedStream, don't let AsciiLineReader wrap it with another one...
+            return new AsciiLineReaderIterator(new AsciiLineReader((PositionalBufferedStream)inputStream));
         } else {
-            pbs = new PositionalBufferedStream(bufferedInputStream);
+            return new AsciiLineReaderIterator(new AsciiLineReader(inputStream));
```
droazen: What about this case? Seems like you're not "indexable" if you reach this else clause.

src/main/java/htsjdk/tribble/readers/BlockCompressedAsciiLineReader.java
droazen: This class needs unit tests

src/main/java/htsjdk/tribble/readers/BlockCompressedAsciiLineReader.java
```diff
@@ -0,0 +1,42 @@
...
+
+    final private BlockCompressedInputStream bcs;
+
+    public BlockCompressedAsciiLineReader(final BlockCompressedInputStream inputBlockCompressedStream) {
+        bcs = inputBlockCompressedStream;
+    }
+
+    /**
+     * Read a single line of inpu, advnced the underlying stream only enough to read the line.
```
droazen: inpu -> input
advnced -> advanced

src/main/java/htsjdk/tribble/readers/BlockCompressedAsciiLineReader.java
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public String readLine() throws IOException {
+        return bcs.readLine();
+    };
+
+    public String readLine(final PositionalBufferedStream stream) {
+        throw new IllegalArgumentException("A BlockCompressedAsciiLineReader class cannot be used to read from a PositionalBufferedStream");
```
droazen: `UnsupportedOperationException`

src/main/java/htsjdk/samtools/BinningIndexBuilder.java
```diff
@@ -50,6 +52,10 @@
      */
     public BinningIndexBuilder(final int referenceSequence, final int sequenceLength) {
         this.referenceSequence = referenceSequence;
+        // Initially set each window to -1 so we can distinguish between windows that have no overlapping
+        // features, and those who's lowest offset is 0, which is a valid (virtual file) offset for feature
```
droazen: who's -> whose

src/main/java/htsjdk/samtools/BinningIndexBuilder.java
```diff
@@ -131,9 +137,11 @@ public void processFeature(final FeatureToBeIndexed feature) {
             largestIndexSeen = endWindow;
         }
 
-        // set linear index at every 16K window that this feature overlaps
+        // Set the linear index at every 16K window that this feature overlaps, but only if this chunk
+        // start is strictly earlier than any previously seen chunk start for the window, or if the window
+        // is uninitialized (this is the first feature overlapping this window).
         for (int win = startWindow; win <= endWindow; win++) {
-            if (index[win] == 0 || chunkStart < index[win]) {
+            if (index[win] == UNINITIALIZED_WINDOW || chunkStart < index[win]) {
```
droazen: Add comment that this needs to be this way to deal with records that legitimately start at position 0

src/main/java/htsjdk/samtools/BinningIndexBuilder.java
```diff
@@ -33,6 +34,7 @@
  * Builder for a BinningIndexContent object.
  */
 public class BinningIndexBuilder {
+    private static final int UNINITIALIZED_WINDOW = -1;
```
droazen: Can you add a unit test for this class that tests with a record at position 0.

src/main/java/htsjdk/tribble/TribbleIndexedFeatureReader.java
```diff
@@ -243,6 +243,7 @@ private void readHeader() throws IOException {
         try {
             is = ParsingUtils.openInputStream(path, wrapper);
             if (hasBlockCompressedExtension(new URI(URLEncoder.encode(path, "UTF-8")))) {
+                // TODO: TEST/FIX THIS!
```
droazen: Can you ensure that a github ticket is filed for this?

src/main/java/htsjdk/tribble/bed/BEDCodec.java
```diff
@@ -86,11 +86,38 @@ public BEDFeature decode(String line) {
...
 
+    /**
+     * The BED codec doesn't retain the actual header, but we need to parse through
+     * it and advance to the beginning of the first feature. This is especially true
+     * if we're indexing, since we want to underlying stream offset to be established
+     * correctly, but is also the case for when we're simply iterating to satisfy a
+     * feature query (otherwise the feature reader can terminate prematurely if the
+     * header is large).
+     * @param lineIterator
+     * @return
```
droazen: Fill in `@return`

src/main/java/htsjdk/tribble/bed/BEDCodec.java
```diff
@@ -86,11 +86,38 @@ public BEDFeature decode(String line) {
...
+     */
     @Override
-    public Object readActualHeader(LineIterator reader) {
+    public Object readActualHeader(final LineIterator lineIterator) {
```
droazen: Add unit test for this method

src/main/java/htsjdk/tribble/AsciiFeatureCodec.java
```diff
@@ -49,14 +50,19 @@ public boolean isDone(final LineIterator lineIterator) {
     }
 
     @Override
-    public LocationAware makeIndexableSourceFromStream(final InputStream bufferedInputStream) {
-        final PositionalBufferedStream pbs;
-        if (bufferedInputStream instanceof PositionalBufferedStream) {
-            pbs = (PositionalBufferedStream) bufferedInputStream;
+    public LocationAware makeIndexableSourceFromStream(final InputStream inputStream) {
```
droazen: Can you file a ticket against `LocationAware` to either make all implementation obey the interface contract, or change the interface to reflect the actual behavior?

cmnbroad:
This PR fixes https://github.com/samtools/htsjdk/issues/943, which is a tracking ticket for the BinningIndexBuilder issue fixed as part of this PR.

src/main/java/htsjdk/samtools/BinningIndexBuilder.java
cmnbroad: Done. There were previously no unit tests so no there are tests for a few basic cases.

src/main/java/htsjdk/samtools/BinningIndexBuilder.java
cmnbroad: Done.

src/main/java/htsjdk/tribble/AsciiFeatureCodec.java
cmnbroad: Possibly; it depends on the type of the inputStream. I left this in since it represents the previous behavior for this method. If inputStream isn't wrapping a block compressed stream, and reports it's position correctly, I think it would work. We could throw instead, but if we we're going to do that, we could deprecate this method and replace it method with two overloads with the correct stream argument types.

src/main/java/htsjdk/tribble/TribbleIndexedFeatureReader.java
cmnbroad: https://github.com/samtools/htsjdk/issues/944

src/main/java/htsjdk/tribble/bed/BEDCodec.java
cmnbroad: Done.

src/main/java/htsjdk/tribble/bed/BEDCodec.java
cmnbroad: Done.

src/main/java/htsjdk/tribble/readers/BlockCompressedAsciiLineReader.java
cmnbroad: Done.

src/main/java/htsjdk/tribble/readers/BlockCompressedAsciiLineReader.java
cmnbroad: Done.

cmnbroad: Done.

src/main/java/htsjdk/tribble/AsciiFeatureCodec.java
cmnbroad: https://github.com/samtools/htsjdk/issues/945

src/main/java/htsjdk/tribble/readers/BlockCompressedAsciiLineReader.java
cmnbroad: Done.

cmnbroad:
Round 1 done. Back to @droazen.

src/main/java/htsjdk/tribble/AsciiFeatureCodec.java
lbergelson: It seems like we should have a test case for this case.

lbergelson: Should there at least be a warning if we reach this?  It seems very possible that you're going to try indexing STDIN or something like that here.

src/main/java/htsjdk/tribble/index/IndexFactory.java
```diff
@@ -418,56 +418,48 @@ private static void checkSorted(final File inputFile, final Feature lastFeature,
...
 
-        private PositionalBufferedStream initStream(final File inputFile, final long skip) {
+        private PositionalBufferedStream initIndexablePositionalStream(final File inputFile) {
             try {
                 final FileInputStream fileStream = new FileInputStream(inputFile);
-                final InputStream is;
+                return new PositionalBufferedStream(fileStream);
+            } catch (final FileNotFoundException e) {
+                throw new TribbleException.FeatureFileDoesntExist("Unable to open the input file, most likely the file doesn't exist.", inputFile.getAbsolutePath());
+            } catch (final IOException e) {
```
lbergelson: lets make sure we don't get npr in these exceptions

src/test/java/htsjdk/tribble/index/IndexFactoryTest.java
```diff
@@ -112,5 +125,83 @@ public void testCreateTabixIndexOnBlockCompressed() {
...
+                },
+        };
+    }
+
+    @Test(dataProvider = "vcfDataProvider")
+    public void testCreateTabixIndexFromVCF(
+            final File inputVCF,
+            final Interval queryInterval) throws IOException {
+        // copy the original file and create the index for the copy
+        final File tmpVCF = new File(IOUtil.createTempDir("testCreateTabixIndexFromVCF", null), inputVCF.getName());
```
lbergelson: cleanup this temp dir afterwards

src/test/java/htsjdk/tribble/index/IndexFactoryTest.java
```diff
@@ -112,5 +125,83 @@ public void testCreateTabixIndexOnBlockCompressed() {
...
+                new Object[] {
+                        new File("src/test/resources/htsjdk/variant/serialization_test.bcf")
+                }
+        };
+    }
+
+    @Test(dataProvider = "bcfDataFactory")
+    public void testCreateLinearIndexFromBCF(final File inputBCF) throws IOException {
+        // copy the original file and create the index for the copy
+        final File tmpBCF = new File(IOUtil.createTempDir("testCreateIndexFromBCF", null), inputBCF.getName());
```
lbergelson: same cleanup issue here I think

src/test/java/htsjdk/tribble/index/IndexFactoryTest.java
```diff
@@ -112,5 +125,83 @@ public void testCreateTabixIndexOnBlockCompressed() {
...
+        Files.copy(inputBCF, tmpBCF);
+        final Index index = IndexFactory.createIndex(tmpBCF, new BCF2Codec(), IndexFactory.IndexType.LINEAR);
+        index.writeBasedOnFeatureFile(tmpBCF);
+
+        try(final VCFFileReader originalReader = new VCFFileReader(inputBCF,false);
+            final VCFFileReader tmpReader = new VCFFileReader(tmpBCF, new File(tmpBCF.getAbsolutePath()+ ".idx"),true)) {
```
lbergelson: hardcoded file extension

src/test/java/htsjdk/tribble/index/tabix/TabixIndexTest.java
```diff
@@ -134,6 +148,73 @@ public void testQueryProvidedItemsAmount() throws IOException {
...
+        };
+    }
+
+    @Test(dataProvider = "bedTabixIndexTestData")
+    public void testBedTabixIndex(
+            final File inputBed,
+            final List<Interval> queryIntervals
+    ) throws Exception {
+        // copy the input file and create an index for the copy
+        final File tmpBed = new File(IOUtil.createTempDir("testBedTabixIndex", null), inputBed.getName());
```
lbergelson: temp dirs all over the place

src/main/java/htsjdk/tribble/index/IndexFactory.java
cmnbroad: Done.

src/test/java/htsjdk/tribble/index/IndexFactoryTest.java
cmnbroad: Done.

src/test/java/htsjdk/tribble/index/IndexFactoryTest.java
cmnbroad: Done.

src/test/java/htsjdk/tribble/index/IndexFactoryTest.java
cmnbroad: Done.

src/test/java/htsjdk/tribble/index/tabix/TabixIndexTest.java
cmnbroad: Done.

src/test/java/htsjdk/samtools/BinningIndexBuilderTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.testng.annotations.Test;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+
+public class BinningIndexBuilderTest extends HtsjdkTest {
+
+    static int referenceSequenceIndex = 19;
```
pshapiro4broad: should be `private`, `final` and all uppercase, e.g. `REFERENCE_SEQUENCE_INDEX`

src/test/java/htsjdk/tribble/bed/BEDCodecTest.java
```diff
@@ -25,27 +25,63 @@
...
+    }
+
+    @Test(dataProvider = "gzippedBedTestData")
+    public void testReadActualHeader(final File gzippedBedFile, final int firstFeatureOffset) throws IOException {
+        // Given an indexable SOURCE on a BED file, test that readActualHeader retains the correct offset
+        // of the first feature, whether there is a header or not
+        BEDCodec bedCodec = new BEDCodec();
+        try (final InputStream is = ParsingUtils.openInputStream(gzippedBedFile.getPath());
+             final BlockCompressedInputStream bcis = new BlockCompressedInputStream(is))
+        {
```
lbergelson: For multiline resource blocks I think this way is clearer.

pshapiro4broad: A matter of taste :)
One data point, oracle's Java documentation: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

src/main/java/htsjdk/tribble/AsciiFeatureCodec.java
cmnbroad: Since we had to duplicate this code in GATK anyway, I added a "from" method per this suggestion.

cmnbroad:
@jrobinso We're planning to make some changes to the way indexing works in order to fix several bugs, and wanted to get in front of any issues this change might cause for IGV. Amongst other things, with this PR, when indexing we no longer close the input SOURCE/stream after reading the header (this affected indexing for at least one codec in GATK). Can you take a look and see if this will require IGV changes ? Thx.

jrobinso:
I don't see any IGV impact, and if there are I will deal with it in IGV.

src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
```diff
@@ -85,7 +116,10 @@ public int getLineTerminatorLength() {
      * @return A String containing the contents of the line or null if the
      *         end of the stream has been reached
      */
-    public final String readLine(final PositionalBufferedStream stream) throws IOException{
+    // TODO: This should be deprecated and made private/merged with readLine, or made
+    // static, as it reads from the provided input stream rather than the wrapped input stream
+    @Deprecated
```
lbergelson: missed this before, but could you add a deprecation date?

lbergelson: and a javadoc comment

src/main/java/htsjdk/tribble/readers/BlockCompressedAsciiLineReader.java
lbergelson: I think now that we have that static method we should make this either a package protected class or an inner class of AsciiLineReader.  I don't think we want to expose it unless we have a use case for that.

src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
```diff
@@ -36,15 +35,25 @@
...
     private int lineTerminatorLength = -1;
 
+    protected AsciiLineReader() {};
+
+    /**
+     * TODO: This class implements LocationAware, which requires preservation of virtual file pointers on BGZF inputs.
+     * However, if the inputStream wrapped by this class is a BlockCompressedInputStream, it violates that contract by
+     * wrapping the stream and returning positional file offsets instead.
+     */
     public AsciiLineReader(final InputStream is){
```
lbergelson: Should we deprecate this constructor and eventually make it private?  Is there a case outside of this class that we would want to call this constructor instead of `from()`?

src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
cmnbroad: I put an @deprecated inside the javadoc, with the date, and added an @see to the "from" method. I can't figure out how to anchor to the OTHER readLine method though. Maybe I could if I swapped the order of the methods....

src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
cmnbroad: Hm. We could do that. There are a bunch of internal uses of it, but thats easily fixed. But should we deprecate the other one too (the one that takes a PositionalBufferedInputStream), since "from" is the preferred method ? Otherwise its weird to leave the PBS constructor, but not have a BCS constructor (you'd have to use .from for that case). It would be most consistent to just have .from, but thats less BWC.

cmnbroad:
@lbergelson I may have added more questions than answers... See what you think.

src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
lbergelson: I think you're right that it would make sense to also deprecate the other one.  What's BWC stand for?

cmnbroad: Sorry...backward compatible. I'll deprecate the other one as well.

droazen:
@cmnbroad Before merging this, can we ensure that there are clear and complete instructions for codec authors on how to write a codec that is properly indexable? (eg., in a comment to `FeatureCodec`)

cmnbroad:
@droazen Done. We should let @lbergelson have one more look at the previous set of changes I made when he gets to it.

lbergelson:
@cmnbroad Looks good to me.  👍  Merge when ready.

src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
```diff
@@ -36,18 +35,36 @@
...
 
-    private final PositionalBufferedStream is;
+    private PositionalBufferedStream is;
     private char[] lineBuffer;
     private int lineTerminatorLength = -1;
 
+    protected AsciiLineReader() {};
+
+    /**
+     * TODO: This class implements LocationAware, which requires preservation of virtual file pointers on BGZF inputs.
```
pshapiro4broad: I don't think you want TODO here. This sounds more like a warning or a note to the user of the class.

cmnbroad:
Squashed and rebase once more. Will merge when tests pass.

src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
```diff
@@ -36,18 +35,36 @@
...
+     * However, if the inputStream wrapped by this class is a BlockCompressedInputStream, it violates that contract by
+     * wrapping the stream and returning positional file offsets instead.
+     *
+     * @deprecated 8/8/2017 see {@link #from}
```
pshapiro4broad: I think the format you want here is `use {@link #from(InputStream)}`. Saying "see" to me sounds like `@see` which has a different usage in javadoc. You can provide the types of the parameters if you want to specify an overridden method.

Also it's not great form to use a date, since the date when the software released is not yet fixed, or easy for a user to determine given a .jar they've got lying around. If you want to include this information, use the version number.

src/main/java/htsjdk/tribble/readers/BlockCompressedAsciiLineReader.java
```diff
@@ -0,0 +1,42 @@
...
+    final private BlockCompressedInputStream bcs;
+
+    public BlockCompressedAsciiLineReader(final BlockCompressedInputStream inputBlockCompressedStream) {
+        bcs = inputBlockCompressedStream;
+    }
+
+    /**
+     * Read a single line of input, advance the underlying stream only enough to read the line.
+     */
+    public String readLine() throws IOException {
```
pshapiro4broad: If these are overridden, they need an `@Override` annotation. Also there's no need to javadoc an overridden method, as javadoc will copy down the base class method doc.

src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
```diff
@@ -36,18 +35,36 @@
...
     private static final int BUFFER_OVERFLOW_INCREASE_FACTOR = 2;
     private static final byte LINEFEED = (byte) ('\n' & 0xff);
     private static final byte CARRIAGE_RETURN = (byte) ('\r' & 0xff);
 
-    private final PositionalBufferedStream is;
+    private PositionalBufferedStream is;
     private char[] lineBuffer;
     private int lineTerminatorLength = -1;
 
+    protected AsciiLineReader() {};
```
pshapiro4broad: This is not a great use of an empty constructor. If subclassed improperly a user ends up with a class that's in an invalid state. It will only work correctly if certain methods are overridden, which are not part of a documented API.

The proper solution is to create an abstract base class (or interface with default methods) that both `AsciiLineReader` and `BlockCompressedAsciiLineReader` inherit from. That makes it explicit which methods must be overridden and prevents having inherited fields in `BlockCompressedAsciiLineReader` that are invalid.

cmnbroad: Agreed thats a better solution, but it its not source compatible. We recognize that we need to do something more comprehensive to fix the underlying problems, and there are a couple of issue tickets to represent that. Ideally we'll only break compatibility once for these issues.

src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
cmnbroad: Well, it is a TODO, but you're right the TODO doesn't need to be in the javadoc, just the warning.

src/main/java/htsjdk/tribble/readers/BlockCompressedAsciiLineReader.java
cmnbroad: Done.

src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
cmnbroad: I changed the "see"s to "use". AFACIT, we've been putting deprecation dates, I think mostly for our own benefit, since we have existing deprecated stuff but we don't know when it was deprecated. As soon as the maintainer groups defines the deprecation rules, we'll need that info...

lbergelson: We've started adding the dates so we can get a rough sense of how long something has been deprecated.  We could always look in the blame which is more accurate, but it's a pain if you're scanning through lots of things.  We could provide the version number, but our deprecation removal policy (which doesn't really exist in any official form) is "has this been deprecated for a long time" rather than "has this been deprecated for N versions".

src/main/java/htsjdk/tribble/readers/BlockCompressedAsciiLineReader.java
lbergelson: Good catch.

src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
cmnbroad: Never mind - just saw your comment....

lbergelson: @cmnbroad Yeah, it's really not intended for the users except for them to say "oh, yeah, this is ancient and I shouldn't use it."  A version number wouldn't be a bad idea, but then we have to be sure we guess what version the commit is going into, which is awkward.  Checking the blame is the ultimate guide, but it's a pain to do in bulk.

src/main/java/htsjdk/tribble/readers/AsciiLineReader.java
pshapiro4broad: OK

Issue #908, opened 2017-06-22
TabixReader bug in query(String) method for coordiantes with 0 values


phanikishore2:
The following method breaks if coordinate value is 0.   

For example chr2:0-1

TabixReader.query("chr2:0-1") returns (tid,-1,1)
TabixReader.query("chr2",0,1) returns (tid,0,1)

Digging down in method parseReg(String) method  ret[1] value assignment does subtract 1.

public int[] parseReg(final String reg) { // FIXME: NOT working when the sequence name contains : or -.
        String chr;
        int colon, hyphen;
        int[] ret = new int[3];
        colon = reg.indexOf(':');
        hyphen = reg.indexOf('-');
        chr = colon >= 0 ? reg.substring(0, colon) : reg;
        _**ret[1] = colon >= 0 ? Integer.parseInt(reg.substring(colon + 1, hyphen >= 0 ? hyphen : reg.length())) - 1 : 0;**_
        ret[2] = hyphen >= 0 ? Integer.parseInt(reg.substring(hyphen + 1)) : 0x7fffffff;
        ret[0] = this.chr2tid(chr);
        return ret;
}

Thanks,
Phani

yfarjoun:
Genomic locations in the format that you use are interpreted using 1-based coordinate system.

this is not a bug.

Vernando:
Hello, 
so far I understand it there is a bug in the query(final String) and in the query(String, int, int) methods. 
With a 1-indexed file query(final String) fails with an IndexOutOfBoundsException with "chr2:0-0" which is in my eyes a valid query string with indexes outside of the allowed range and with that the found result should be just an empty iterator (or maybe null, but not a crashing function). The commandline tool handles this correctly.

So with that experience changing to the query(final String, int, int) method fails again, but differently. 
A valid search like ("chr3",178936275,178936275) delivers ZERO results instead of the actual findings in the used tabix files (which should be e.g. 1 in our case). 

The inner most used query(final int tid, final int beg, final int end) which is the same in both cases seems to need a range information versus a position information for beginPos and endPosition.
I switched to use only the query(final String) as entry point and extended the out of bound index checks there. I think that should fix the problem in (only) this method.


```
    public Iterator query(final String reg) {
        int[] x = parseReg(reg);
        if(x[0]<0
        // (jh, 6/23/17) I found bugs in this method and query(String, int, int) for two different scenarios. This
[...]

PR #911, opened 2017-06-24
Move bad coordinates check

ronlevine:
### Description

Follow up to #908.
Move the bad coordinates check to the root method, `public Iterator query(final String reg)`. If this public method is used by an application, the bad coordinates will return an `EOF_ITERATOR;`

### Checklist

- [X] Code compiles correctly
- [X] All tests passing

### DIFF
--- a/src/main/java/htsjdk/tribble/readers/TabixReader.java
+++ b/src/main/java/htsjdk/tribble/readers/TabixReader.java
@@ -446,7 +446,7 @@ public class TabixReader {
     }
 
     /**
-     * Return
+     * Get an iterator for an interval specified by the sequence id and begin and end coordinates
      * @param tid Sequence id
      * @param beg beginning of interval, genomic coords
      * @param end end of interval, genomic coords
@@ -455,7 +455,7 @@ public class TabixReader {
     public Iterator query(final int tid, final int beg, final int end) {
         TPair64[] off, chunks;
         long min_off;
-        if(tid< 0 || tid>=this.mIndex.length) return EOF_ITERATOR;
+        if (tid < 0 || beg < 0 || end < 0 || tid >= this.mIndex.length) return EOF_ITERATOR;
         TIndex idx = mIndex[tid];
         int[] bins = new int[MAX_BIN];
         int i, l, n_off, n_bins = reg2bins(beg, end, bins);
@@ -514,7 +514,6 @@ public class TabixReader {
      */
     public Iterator query(final String reg) {
         int[] x = parseReg(reg);
-        if(x[0]<0) return EOF_ITERATOR;
         return query(x[0], x[1], x[2]);
     }
 
@@ -528,7 +527,6 @@ public class TabixReader {
     */
    public Iterator query(final String reg,int start,int end) {
        int tid=this.chr2tid(reg);
-       if(tid==-1) return EOF_ITERATOR;
        return query(tid, start, end);
    }
 
--- a/src/main/java/htsjdk/tribble/readers/TabixReader.java
+++ b/src/main/java/htsjdk/tribble/readers/TabixReader.java
@@ -447,15 +447,15 @@ public class TabixReader {
 
     /**
      * Get an iterator for an interval specified by the sequence id and begin and end coordinates
-     * @param tid Sequence id
-     * @param beg beginning of interval, genomic coords
-     * @param end end of interval, genomic coords
-     * @return an iterator over the lines within the specified interval
+     * @param tid Sequence id, if non-existent returns EOF_ITERATOR
+     * @param beg beginning of interval, genomic coords (0-based, closed-open)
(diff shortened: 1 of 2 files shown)


ronlevine:
@yfarjoun Please review.

src/test/java/htsjdk/tribble/readers/TabixReaderTest.java
```diff
@@ -95,7 +98,15 @@ public void testIterators() throws IOException {
         iter=tabixReader.query("1",Integer.MAX_VALUE-1,Integer.MAX_VALUE);
         Assert.assertNotNull(iter);
         Assert.assertNull(iter.next());
-        
+
+        iter=tabixReader.query("1",-1,Integer.MAX_VALUE);
```
yfarjoun: spaces around operators

yfarjoun:
@ronlevine are you still available to process this? 

if not @lbergelson what's the procedure?

yfarjoun:
@cmnbroad can you take over this?

cmnbroad:
@yfarjoun Sure.

cmnbroad:
@yfarjoun I don't see anything in this PR that addresses the second issue raised in the ticket. A couple of questions:

Do you think we should we retain the current behavior where querying a non-existent sequence name silently returns an empty iterator ? It seems to me it should throw. Same for start index <= 0. Such queries seem incoherent, but the caller has no way to notify the user. It would be a behavior change though.
It looks like the parseReg method is converting com 1-based to 0-based, and subtracts 1 from the start, but not the end. Unless its intentionally converting from a fully closed interval to half-closed, half-open, that seems wrong. Do you know if thats intentional ? I'd like to add some comments/javadoc and boundary condition tests, but I'm not sure what the intentions are.

src/main/java/htsjdk/tribble/readers/TabixReader.java
```diff
@@ -439,7 +439,7 @@ public String next() throws IOException {
     public Iterator query(final int tid, final int beg, final int end) {
         TPair64[] off, chunks;
         long min_off;
-        if(tid< 0 || tid>=this.mIndex.length) return EOF_ITERATOR;
+        if (tid < 0 || beg < 0 || end < 0 || tid >= this.mIndex.length) return EOF_ITERATOR;
```
droazen: I think this should be `end <= 0`, since the coordinates are 0-based closed-open.

src/main/java/htsjdk/tribble/readers/TabixReader.java
```diff
@@ -430,7 +430,7 @@ public String next() throws IOException {
     }
 
     /**
-     * Return
+     * Get an iterator for an interval specified by the sequence id and begin and end coordinates
      * @param tid Sequence id
      * @param beg beginning of interval, genomic coords
      * @param end end of interval, genomic coords
```
droazen: The javadoc for this method should document the convention being used for the genomic coordinates. Based on the callsites for this method, they appear to be BED-style (0-based, closed-open).

src/test/java/htsjdk/tribble/readers/TabixReaderTest.java
```diff
@@ -95,7 +98,15 @@ public void testIterators() throws IOException {
...
+        Assert.assertNotNull(iter);
+        Assert.assertNull(iter.next());
+
+        iter=tabixReader.query("1",Integer.MAX_VALUE,-1);
+        Assert.assertNotNull(iter);
+        Assert.assertNull(iter.next());
+
```
droazen: Can we add a test case in which end == 0?

droazen: Also add test cases for the other overloads of `query()`, since they have difference interval conventions (the String version is 1-based closed, the 3-arg version is 0-based closed-open!)

src/main/java/htsjdk/tribble/readers/TabixReader.java
droazen: Also applies to the other overloads of `query()`. It looks like some overloads use the 1-based closed convention, while others like this one use the 0-based closed-open convention.

droazen:
Review complete, back to @cmnbroad

src/main/java/htsjdk/tribble/readers/TabixReader.java
droazen: Also document the behavior with non-existent contigs, etc.

yfarjoun:
ping....

fleharty:
@yfarjoun I'm taking this over, but don't have the ability to assign since I don't have commit privileges.

yfarjoun:
@droazen ? @cmnbroad ? who's court is this in now?

fleharty:
@yfarjoun my court

yfarjoun:
looks good now.

PR #912, opened 2017-06-26
Use quality header and SAMTag.CO in FastqEncoder

magicDGS:
### Description

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/fastq/FastqEncoder.java
+++ b/src/main/java/htsjdk/samtools/fastq/FastqEncoder.java
@@ -26,6 +26,7 @@ package htsjdk.samtools.fastq;
 import htsjdk.samtools.SAMException;
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SAMRecord;
+import htsjdk.samtools.SAMTag;
 import htsjdk.samtools.util.SequenceUtil;
 
 import java.io.IOException;
@@ -91,7 +92,7 @@ public final class FastqEncoder {
         if(record.getReadPairedFlag() && (record.getFirstOfPairFlag() || record.getSecondOfPairFlag())) {
             readName += (record.getFirstOfPairFlag()) ? FastqConstants.FIRST_OF_PAIR : FastqConstants.SECOND_OF_PAIR;
         }
-        return new FastqRecord(readName, record.getReadString(), null, record.getBaseQualityString());
+        return new FastqRecord(readName, record.getReadString(), record.getStringAttribute(SAMTag.CO.name()), record.getBaseQualityString());
     }
 
     /**
@@ -106,6 +107,7 @@ public final class FastqEncoder {
         // set the basic information from the FastqRecord
         samRecord.setReadName(readName);
         samRecord.setReadBases(record.getReadBases());
+        samRecord.setAttribute(SAMTag.CO.name(), record.getBaseQualityHeader());
         samRecord.setBaseQualities(record.getBaseQualities());
         return samRecord;
     }
--- a/src/main/java/htsjdk/samtools/fastq/FastqEncoder.java
+++ b/src/main/java/htsjdk/samtools/fastq/FastqEncoder.java
@@ -30,6 +30,7 @@ import htsjdk.samtools.SAMTag;
 import htsjdk.samtools.util.SequenceUtil;
 
 import java.io.IOException;
+import java.util.function.BiConsumer;
 
 /**
  * Codec for encoding records into FASTQ format.
@@ -99,18 +100,20 @@ public final class FastqEncoder {
      * Converts a {@link FastqRecord} into a simple unmapped {@link SAMRecord}.
      */
     public static SAMRecord asSAMRecord(final FastqRecord record, final SAMFileHeader header) {
-        return asSAMRecord(record, header, false);
+        return asSAMRecord(record, header, (s, r) -> {});
     }
 
     /**
      * Converts a {@link FastqRecord} into a simple unmapped {@link SAMRecord}.
      *
-     * @param record                 object to encode.
-     * @param header                 header for the returned object.
-     * @param qualityHeaderToComment if {@code true}, encode the quality header into the {@link SAMTag#CO} tag;
-     *                               otherwise, do not encode the quality header.
+     * <p>This method allows to pass a {@link BiConsumer} to add the information from the record in
+     * a customizable manner.
+     *
+     * @param record    object to encode.
+     * @param header    header for the returned object.
+     * @param custom    function to customize encoding. Note that default information might be overriden.
      */
-    public static SAMRecord asSAMRecord(final FastqRecord record, final SAMFileHeader header, final boolean qualityHeaderToComment) {
+    public static SAMRecord asSAMRecord(final FastqRecord record, final SAMFileHeader header, final BiConsumer<FastqRecord, SAMRecord> custom) {
         // construct the SAMRecord and set the unmapped flag
         final SAMRecord samRecord = new SAMRecord(header);
         samRecord.setReadUnmappedFlag(true);
@@ -119,11 +122,16 @@ public final class FastqEncoder {
         // set the basic information from the FastqRecord
         samRecord.setReadName(readName);
         samRecord.setReadBases(record.getReadBases());
-        if (qualityHeaderToComment) {
-            samRecord.setAttribute(SAMTag.CO.name(), record.getBaseQualityHeader());
-        }
         samRecord.setBaseQualities(record.getBaseQualities());
+        custom.accept(record, samRecord);
         return samRecord;
     }
 
+    /**
+     * Encodes the quality header into the comment tag (use in {@link #asSAMRecord(FastqRecord, SAMFileHeader, BiConsumer)}.
+     *
+     * <p>Note that all tabs present in the quality header are replaced by spaces.
+     */
+    public static final BiConsumer<FastqRecord, SAMRecord> QUALITY_HEADER_TO_COMMENT_TAG = (record, samRecord) ->
+        samRecord.setAttribute(SAMTag.CO.name(), record.getBaseQualityHeader().replaceAll("\t", " "));
 }
--- a/src/main/java/htsjdk/samtools/fastq/FastqEncoder.java
+++ b/src/main/java/htsjdk/samtools/fastq/FastqEncoder.java
@@ -27,9 +27,11 @@ import htsjdk.samtools.SAMException;
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SAMRecord;
 import htsjdk.samtools.SAMTag;
+import htsjdk.samtools.TextTagCodec;
 import htsjdk.samtools.util.SequenceUtil;
 
 import java.io.IOException;
+import java.util.Map;
 import java.util.function.BiConsumer;
 
 /**
@@ -134,4 +136,14 @@ public final class FastqEncoder {
      */
     public static final BiConsumer<FastqRecord, SAMRecord> QUALITY_HEADER_TO_COMMENT_TAG = (record, samRecord) ->
         samRecord.setAttribute(SAMTag.CO.name(), record.getBaseQualityHeader().replaceAll("\t", " "));
+
+
+    public static final BiConsumer<FastqRecord, SAMRecord> QUALITY_HEADER_PARSE_SAM_TAGS = (record, samRecord) -> {
+        final String[] tokens = record.getBaseQualityHeader().split("\t");
+        final TextTagCodec codec = new TextTagCodec();
+        for (final String token: tokens) {
+            final Map.Entry<String, Object> tagValue = codec.decode(token);
+            samRecord.setAttribute(tagValue.getKey(), tagValue.getValue());
+        }
+    };
 }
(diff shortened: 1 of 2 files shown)


lbergelson:
@magicDGS Could you explain this one?  It makes sense to capture the comment lines from fastq somewhere, but is CO the standard place?  

@yfarjoun Any objections to this?  It could potentially increase bam size if we have comments in the fastq, but A) I assume we don't, B) I don't think we even go through fastq anymore?

magicDGS:
@lbergelson - specifications have the `CO` tag defined as a [free-text comment](https://samtools.github.io/hts-specs/SAMtags.pdf). Some FASTQ files that I found in the wild include in the quality header (starting with `+`) comments about the read. I thought that these comments can leave on the `CO` tag as it is defined as comments. Other information that can be stored in the `CO` tag are the comments in the read name, but that's more tricky as some of this information might be barcodes or pair-end information.

Another option is to design a new tag in the specs for the FASTQ quality-header (`FQ`?), and use the `CO` for other comments, such the ones in the read name (after the space). What do you think that is better, @lbergelson/@yfarjoun? I would like to have a lossless way of transforming FASTQ to SAM, and store this kind of information in tags is the best way...

yfarjoun:
To be more accurate, the header CO tag is specified as 
> One-line text comment. Unordered multiple @CO lines are allowed. UTF-8 encoding may be used.

the "free text" is from the alignment tag "CO"...

magicDGS:
The suggestion is for a per-alignment `CO` tag, containing the information from the "quality header" from the FASTQ. This could vary between reads.

yfarjoun:
Ah, I see. sorry, my bad.

yfarjoun:
I'd prefer if you made the conversion of quality header to comment optional (default false).

magicDGS:
Sounds reasonable - done in next commit.

src/main/java/htsjdk/samtools/fastq/FastqEncoder.java
```diff
@@ -91,13 +92,25 @@ public static FastqRecord asFastqRecord(final SAMRecord record) {
         if(record.getReadPairedFlag() && (record.getFirstOfPairFlag() || record.getSecondOfPairFlag())) {
             readName += (record.getFirstOfPairFlag()) ? FastqConstants.FIRST_OF_PAIR : FastqConstants.SECOND_OF_PAIR;
         }
-        return new FastqRecord(readName, record.getReadString(), null, record.getBaseQualityString());
+        return new FastqRecord(readName, record.getReadString(), record.getStringAttribute(SAMTag.CO.name()), record.getBaseQualityString());
```
yfarjoun: Are you concerned that a record that has a comment that it not related to the QualityHeader will be misunderstood and the comment will then become the quality header? 

The wikipedia page seems to imply that the sample_id needs to be the "first" part of this header, followed by arbitrary text, but on the other hand, the fastq format is poorly defined...so it's not even clear what the separator is.

src/main/java/htsjdk/samtools/fastq/FastqEncoder.java
```diff
@@ -106,6 +119,9 @@ public static SAMRecord asSAMRecord(final FastqRecord record, final SAMFileHeade
         // set the basic information from the FastqRecord
         samRecord.setReadName(readName);
         samRecord.setReadBases(record.getReadBases());
+        if (qualityHeaderToComment) {
+            samRecord.setAttribute(SAMTag.CO.name(), record.getBaseQualityHeader());
```
yfarjoun: make sure that tabs are not part of this string....as you'll destroy the bam otherwise. 
(please test this)

src/main/java/htsjdk/samtools/fastq/FastqEncoder.java
magicDGS: I just would like to have a lossless conversion from FASTQ to SAM with this encoder. Validation of the comment could be performed downstream.

src/main/java/htsjdk/samtools/fastq/FastqEncoder.java
magicDGS: Actually, this is a difficult issue: in bwa, for example, the quality-header could be use to assign tags to SAM/BAM. From http://bio-bwa.sourceforge.net/bwa.shtml:

```
-C | Append append FASTA/Q comment to SAM output. This option can be used to transfer read meta information (e.g. barcode) to the SAM output. Note that the FASTA/Q comment (the string after a space in the header line) must conform the SAM spec (e.g. BC:Z:CGTAC). Malformated comments lead to incorrect SAM output.
```

Maybe the way to go here is to parse the quality header for tag-formatted comments and if present add to the attributes; otherwise, add to the `SAMTag.CO` by converting any tab to spaces. What do you think?

yfarjoun: I think that if there's some standard, (e.g. the one in bwa) we should use it. I don't like the idea of "either/or"....can you just use SAM-style tags and not allow for free text?

magicDGS: I don't think that the `bwa` quality header is a standard, but an ad-hoc solution to allow including per-read information from FASTQ input reads into the output of `bwa`. I still would like to allow for free-text, because in the wild there are FASTQ files with that kind of header, and there are options when downloading from SRA/ENA servers to include free-text information in quality headers (e.g., sratoolkit's `fastq-dump` command allows custom `--defline-qual`, which won't complies to SAM-style tags).

yfarjoun: In that case, I would advise for a parameter switch for the different modes, rather than something that depends on "if present".....

magicDGS: I added a even better configuration mode: use a `BiConsumer` to add extra information from the `FastqRecord` to the `SAMRecord` is a custom way: I included an implementation for the `CO` tag (converting tabs into spaces) and we can also include the one for setting the tags if you feel that is necessary.

magicDGS: I did add also another custom encoder.

src/test/java/htsjdk/samtools/fastq/FastqEncoderTest.java
```diff
@@ -72,7 +72,18 @@ public void testAsSAMRecord() throws Exception {
         testConvertedSAMRecord(FastqEncoder.asSAMRecord(fastqRecord, samRecord.getHeader()), samRecord);
         // test with qualityHeaderToComment=true populates the CO tag
         samRecord.setAttribute("CO", fastqRecord.getBaseQualityHeader());
-        testConvertedSAMRecord(FastqEncoder.asSAMRecord(fastqRecord, samRecord.getHeader(), true), samRecord);
+        testConvertedSAMRecord(FastqEncoder.asSAMRecord(fastqRecord, samRecord.getHeader(), FastqEncoder.QUALITY_HEADER_TO_COMMENT_TAG), samRecord);
+    }
+
+    @Test
+    public void testAsSAMRecordParsedTags() throws Exception {
+        final String tags = "BC:Z:ACTG\tRG:Z:sample1";
```
yfarjoun: please test other types as well..

src/test/java/htsjdk/samtools/fastq/FastqEncoderTest.java
```diff
@@ -46,13 +46,15 @@ public void testAsFastqRecord() throws Exception {
         testRecord(record.getReadName() + FastqConstants.SECOND_OF_PAIR, FastqEncoder.asFastqRecord(record), record);
         record.setSecondOfPairFlag(false);
         testRecord(record.getReadName(), FastqEncoder.asFastqRecord(record), record);
+        record.setAttribute("CO", "Comment in SAM tag");
```
yfarjoun: use `SAMTag.CO.name()`

src/test/java/htsjdk/samtools/fastq/FastqEncoderTest.java
```diff
@@ -46,13 +46,15 @@ public void testAsFastqRecord() throws Exception {
...
+        testRecord(record.getReadName(), FastqEncoder.asFastqRecord(record), record);
     }
 
     private void testRecord(final String expectedReadName, final FastqRecord fastqRecord, final SAMRecord samRecord) {
         Assert.assertEquals(fastqRecord.getReadName(), expectedReadName);
         Assert.assertEquals(fastqRecord.getBaseQualities(), samRecord.getBaseQualities());
         Assert.assertEquals(fastqRecord.getReadBases(), samRecord.getReadBases());
-        Assert.assertNull(fastqRecord.getBaseQualityHeader());
+        Assert.assertEquals(fastqRecord.getBaseQualityHeader(), samRecord.getStringAttribute("CO"));
```
yfarjoun: `SAMTag.CO.name()`

src/test/java/htsjdk/samtools/fastq/FastqEncoderTest.java
```diff
@@ -65,12 +67,30 @@ public void testAsSAMRecord() throws Exception {
         testConvertedSAMRecord(FastqEncoder.asSAMRecord(fastqRecord, samRecord.getHeader()), samRecord);
         fastqRecord = new FastqRecord(samRecord.getReadName() + FastqConstants.SECOND_OF_PAIR, samRecord.getReadBases(), "", samRecord.getBaseQualities());
         testConvertedSAMRecord(FastqEncoder.asSAMRecord(fastqRecord, samRecord.getHeader()), samRecord);
+        fastqRecord = new FastqRecord(samRecord.getReadName() + FastqConstants.SECOND_OF_PAIR, samRecord.getReadBases(), "Quality header comment", samRecord.getBaseQualities());
+        // default method does not include the comment header
+        testConvertedSAMRecord(FastqEncoder.asSAMRecord(fastqRecord, samRecord.getHeader()), samRecord);
+        // test with qualityHeaderToComment=true populates the CO tag
+        samRecord.setAttribute("CO", fastqRecord.getBaseQualityHeader());
```
yfarjoun: `SAMTag.CO.name()`

src/test/java/htsjdk/samtools/fastq/FastqEncoderTest.java
```diff
@@ -65,12 +67,30 @@ public void testAsSAMRecord() throws Exception {
...
+        testConvertedSAMRecord(converted, samRecord);
+        Assert.assertEquals(converted.getAttribute("BC"), "ACTG");
+        Assert.assertEquals(converted.getAttribute("RG"), "sample1");
     }
 
     private void testConvertedSAMRecord(final SAMRecord converted, final SAMRecord original) {
         Assert.assertEquals(converted.getReadName(), original.getReadName());
         Assert.assertEquals(converted.getBaseQualities(), original.getBaseQualities());
         Assert.assertEquals(converted.getReadBases(), original.getReadBases());
+        Assert.assertEquals(converted.getStringAttribute("CO"), original.getStringAttribute("CO"));
```
yfarjoun: `SAMTag.CO.name()`

src/test/java/htsjdk/samtools/fastq/FastqEncoderTest.java
```diff
@@ -65,12 +67,30 @@ public void testAsSAMRecord() throws Exception {
...
+    }
+
+    @Test
+    public void testAsSAMRecordParsedTags() throws Exception {
+        final String tags = "BC:Z:ACTG\tRG:Z:sample1";
+        final SAMRecord samRecord = new SAMRecordSetBuilder().addFrag("test", 0, 1, false, false, "10M", null, 2);
+        final FastqRecord record = new FastqRecord(samRecord.getReadName(), samRecord.getReadBases(), tags, samRecord.getBaseQualities());
+        final SAMRecord converted = FastqEncoder.asSAMRecord(record, samRecord.getHeader(), FastqEncoder.QUALITY_HEADER_PARSE_SAM_TAGS);
+        testConvertedSAMRecord(converted, samRecord);
+        Assert.assertEquals(converted.getAttribute("BC"), "ACTG");
```
yfarjoun: use SAMTag variable here as well. if they do not exist, then please add them.

src/test/java/htsjdk/samtools/fastq/FastqEncoderTest.java
magicDGS: Done (concretely integer and character)

src/test/java/htsjdk/samtools/fastq/FastqEncoderTest.java
magicDGS: Done

src/test/java/htsjdk/samtools/fastq/FastqEncoderTest.java
magicDGS: Done

src/test/java/htsjdk/samtools/fastq/FastqEncoderTest.java
magicDGS: Done

src/test/java/htsjdk/samtools/fastq/FastqEncoderTest.java
magicDGS: Done

src/test/java/htsjdk/samtools/fastq/FastqEncoderTest.java
magicDGS: Done

magicDGS:
@yfarjoun - addressed comments and hopefully ready to go!
@lbergelson - now the default behaviour is the same, so no worries about increased BAM size

magicDGS:
Friendly ping here @yfarjoun

lbergelson:
@magicDGS Rerunning the tests since they failed with what looks like a transient issue.  I'll merge when they pass.  Thank you.

magicDGS:
Thanks!

PR #913, opened 2017-06-26
SAMFileWriterImpl improvements for subclasses

magicDGS:
### Description

This PR addresses two issues with `SAMFileWriterImpl`:

* When sub-classing, it is difficult to access to some important features unless the class lives in the same pacakge. Concretely, this are the maximum records in RAM and the temporary directory. This PR makes this methods and the getters protected.
* While adding a header to the writer the current implementation generates the text header always, even if it is not used. This is rather inefficient for writers where the header should not be written down for whatever reason. This PR deprecates the abstract method `writeHeader` for a `String`in favor of a new one with a `Supplier` (which defaults in the previous one).

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMFileWriterImpl.java
+++ b/src/main/java/htsjdk/samtools/SAMFileWriterImpl.java
@@ -105,14 +105,14 @@ public abstract class SAMFileWriterImpl implements SAMFileWriter
      * before spilling to disk.  Must be called before setHeader().
      * @param maxRecordsInRam
      */
-    void setMaxRecordsInRam(final int maxRecordsInRam) {
+    protected void setMaxRecordsInRam(final int maxRecordsInRam) {
         if (this.header != null) {
             throw new IllegalStateException("setMaxRecordsInRam must be called before setHeader()");
         }
         this.maxRecordsInRam = maxRecordsInRam;
     }
 
-    int getMaxRecordsInRam() {
+    protected int getMaxRecordsInRam() {
         return maxRecordsInRam;
     }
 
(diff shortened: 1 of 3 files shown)


magicDGS:
Could someone please review this one, maybe a maintainer (@jacarey, @tfenne, @lbergelson)? It is a very simple patch but useful to implement custom writers...

magicDGS:
Friendly ping to maintainers: @jacarey, @tfenne, @lbergelson? I need this for implementing a custom `SAMFileWriter`...

magicDGS:
I really need this - please, any maintainer (@jacarey - @tfenne  - @lbergelson) can take care of it? Thank you!

src/main/java/htsjdk/samtools/SAMFileWriterImpl.java
```diff
@@ -243,9 +245,22 @@ public final void close()
...
     abstract protected void writeHeader(String textHeader);
 
+    /**
+     * Write the header to disk. Header object is available via getHeader().
+     *
+     * <p>Note: default implementation uses {@link #writeHeader(String)}.
+     *
+     * @param textHeaderSupplier for convenience if the implementation needs it.
+     */
+    protected void writeHeader(Supplier<String> textHeaderSupplier) {
```
lbergelson: This seems like a bit of a confusing mechanism for allowing people to avoid converting the header to string.  Would it make more sense to add a `writeHeader(SAMFileHeader header) method that takes SAMFileHeader and constructs the String from it?  Then provide an implementation of it that does the conversion to string.  Your class can  use that override and not write anything. The `CRAMFileWriter` could make use of that method as well, since it currently takes the header string, converts it back to a header and then must convert that again.

magicDGS: Sounds good. Done and modify the rest of implementations to move away from the `String` method.

magicDGS:
Back to you @lbergelson!

lbergelson:
@magicDGS Thank you.  Looks good.

magicDGS:
Thanks!

Issue #924, opened 2017-06-30
Add option to use the maximum instead of the first overlap for LiftOver


ronlevine:
### Subject of the issue
LiftOver uses the first chain overlapping the an interval that satisfies the "minimum fraction of bases that must remap" criteria. This makes it sensitive to the the order of items in the chain file. It should have the option of choosing the remapping that has the highest number of bases that overlap.

### Your environment
* version of htsjdk -  - 2.10.1
* version of java - 1.8.0_66-b17
* which OS - MacOS 10.12.5

### Steps to reproduce
Change the logic in the `for (final Chain chain : chains.getOverlaps(interval))` loop in `LiftOver.liftOver()` to choose the `targetIntersection` with the maximum number of bases rather than the first `candidateIntersection.intersectionLength >= minMatchSize)`.

### Expected behaviour
For a given interval, the remapped interval that satisfies the minimum overlap criteria and whose bases the have the `maximum overlap` with the input assembly.

### Actual behaviour
For a given interval, the `first` remapped interval that satisfies the minimum overlap criteria and whose bases overlap the input assembly.

yfarjoun:
closing due to lack of interest.

PR #926, opened 2017-07-03
Add option to use the maximum overlap for LiftOver

ronlevine:
### Description

Implements #924. 
LiftOver uses the first chain overlapping the an interval that satisfies the "minimum fraction of bases that must remap" criteria. This makes it sensitive to the the order of items in the chain file. It should have the option of choosing the remapping that has the highest number of bases that overlap.

### Checklist

- [X] Code compiles correctly
- [X] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/liftover/LiftOver.java
+++ b/src/main/java/htsjdk/samtools/liftover/LiftOver.java
@@ -51,13 +51,13 @@ public class LiftOver {
     public static final double DEFAULT_LIFTOVER_MINMATCH = 0.95;
 
     private double liftOverMinMatch = DEFAULT_LIFTOVER_MINMATCH;
-    private final OverlapDetector<Chain> chains;
+    private OverlapDetector<Chain> chains;
     private final Map<String, Set<String>> contigMap = new HashMap<>();
 
     /**
      * Load UCSC chain file in order to lift over Intervals.
      */
-    public LiftOver(File chainFile) {
+    public LiftOver(final File chainFile) {
         IOUtil.assertFileIsReadable(chainFile);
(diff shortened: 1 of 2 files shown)


ronlevine:
@yfarjoun Please review.

ronlevine:
@yfarjoun Is this useful?

yfarjoun:
I don't know! There are other ways to choose the "best" liftover...for
example the one with the best agreement of Reference...or the one with the
same orientation...

(I haven't read the PR yet)

src/main/java/htsjdk/samtools/liftover/LiftOver.java
```diff
@@ -115,12 +126,21 @@ public Interval liftOver(final Interval interval, final double liftOverMinMatch)
...
 
         // Find the appropriate Chain, and the part of the chain corresponding to the interval to be lifted over.
+        int maxIntersectionSize = 0;
         for (final Chain chain : chains.getOverlaps(interval)) {
             final TargetIntersection candidateIntersection = targetIntersection(chain, interval);
             if (candidateIntersection != null && candidateIntersection.intersectionLength >= minMatchSize) {
                 if (chainHit != null) {
-                    // In basic liftOver, multiple hits are not allowed.
-                    return null;
+                    if (useMaxOverlap) {
```
yfarjoun: I don't think this works. in particular, while you are saving the `maxIntersectionSize`, you are not keeping the actual `candidateIntersection` around, which means that you will always end up using the last `candidateIntersection` that passes the `minMatchSize`. please add tests in which the maximal intersection is the middle of three valid (i.e. long enough) intersections.

yfarjoun:
closing due to lack of interest

Issue #917, opened 2017-06-26
Mixed licensing is unfortunate; can we remedy this?


tfenne:
So I just looked through the code and as best I can tell we have:

* The majority of the code MIT licensed
* Tribble under LGPL
* SRA under US gov public domain notice
* CRAM under Apache 2.0
* Some files without a license statement, but I'm assuming those are MIT 

It would be really nice to fix this.  The vast majority of tribble is under LGPL and copyright Broad Institute.  @lbergelson & @jacarey given the Broad's recent commitment to open source, do you think one of you could explore what it would take for the Broad to agree to a license change to MIT for the tribble code currently under LGPL?

That would leave CRAM.  I think that there we'd probably need agreement from _both_ EBI (@vadimzalunin ?) and Broad since the majority of the code came in from EBI, but there have been lots of edits and additions from Broad users.  @vadimzalunin is it institute policy at EBI that code must be apache licensed, or do you think it would be possible to get approval to switch to MIT to match the rest of the library?

It sound like the SRA code is largely unsupported at this point and we may just remove it?

vdauwera:
@tfenne It would probably be an easier sell if we could make it all BSD 3-clause since that's what the Data Sciences Platform has chosen as go-to license for public projects. If we advocated for that I can almost guarantee an automatic yes as far as DSP is concerned.

It might also be a closer fit for those other parts of the code that aren't MIT licensed.

tfenne:
@vdauwera I'm not personally opposed to switching to BSD 3 clause.  The only downside of that is that historically we've allowed other contributors to contribute source under the MIT license but copyrighted to the author who did the work.  Going to BSD 3 clause instead of MIT would presumably require contacting and getting approval from all of those authors too.  Perhaps we can investigate the possibility of both?

vdauwera:
@tfenne Sure, will work with our IP managers to see what we can do on the Broad end.

tfenne:
Thanks @vdauwera, much appreciated!

vdauwera:
@tfenne Our IP manager will go over Broad's options when he's back from vacation; I'll be out of office for a few weeks but hopefully we can get this sorted out over email. Feel free to ping me if we end up being a blocker.

MikeThomsen:
@vdauwera Did you get a resolution from your IP manager?

Issue #896, opened 2017-06-09
Proposal: separate HTSJDK in different gradle build artifacts


magicDGS:
This is related a bit with #520 (which I'm still looking forward to be approve by the maintainers), but it could still be done in the current API status. My idea is that while importing HTSJDK for a project that only requires SAM/BAM processing, it is also including a lot of unnecessary code for Tribble/VCF. I propose to have in gradle/maven three/four different build artifacts:

1. `htdjsk-common`: this should have all classes that are common to different packages, such as streams, utils, etc.
2. `htsjdk-samtools`: this corresponds to the code specific for SAM/BAM/CRAM and other sequencing sources (e.g. SRA if not removed).
3. `htsjdk-tribble`: with the code for genomic-feature encoder/decoders, tabix/tribble indexing, etc.
4. `htsjdk-vcf` (Optional): because the VCF code is focus on diploids and it makes other assumptions about the data, I think that it will be better to keep separate from tribble. Maybe it will be better to have interfaces here and call it `htsjdk-variant`, but this belongs to #520 (I guess).

@jacarey, @tfenne, @lbergelson - as maintainers, what do you think about this? And what about other people using the library? Do you think that this is a good step forward regarding the API?

tfenne:
@magicDGS I was thinking about something similar, but a slightly different organization.  I would like to get to a point where we have a `htsjdk-core` that contains common code, interface/api definitions for things and then the implementations of most things that don't include significant dependencies.  So I'd see this having SAM/BAM, VCF and tribble within it.  But then add on modules for extensions of plugins that bring in dependencies, e.g.:

1. `htsjdk-cram` since CRAM requires a lot of external libraries
2. `htsjdk-sra` ditto
3. `htsjdk-jexl` perhaps, so that it's not a core dependency for just reading/writing VCFs
4. etc.

I guess I see the totality of code that's in HTSJDK as quite small, but when combined with dependencies it grows considerably.

Issue #894, opened 2017-06-09
Remove SRA code from htsjdk 


yfarjoun:
I was wondering if we could remove the SRA code from htsjdk. To me it seems that SRA is the source of little more than headaches due to intermittently broken tests. 

Is anyone deriving enough utility from the SRA code that they are willing to make the code more robust so that we don't get intermittent test failures? This should still be done while still testing SRA, naturally... 

Discussion? Vote? or 🏏

lindenb:
on my side, I've never used the SRA code

magicDGS:
I've never used the SRA code either, although I was planning to explore it at some point. But if it is buggy and/or broken, I vote for removal, with plans to implement it in the future and make it more robust from scratch.

tfenne:
Similarly, I _never_ use the SRA code and would be happy to see the problems that it causes go away.  @yfarjoun my impression was that the GATK valued having this here ... is that no longer the case?

vdauwera:
This was proposed by ncbi and I'm aware of a few users who use it, but afaik it's not widely used. I'm reluctant to just remove functionality that some people might depend on, however. What kinhds of problems are we talking about? Can these be quarantined?

ronlevine:
IIRC, the problem had to do with a server timeout when accessing the SRA data. @lbergelson was dealing more with this.

yfarjoun:
@vdauwera perhaps ncbi should implement it as a tribble plugin that they maintain and test...(is that possible?) It's very frustrating to have tests fail due to server problems that we do not control, especially when we hardly know anyone that is using the functionality that is provided. alternatively, We could disable all the tests that use SRA and have bit rot.

yfarjoun:
@vdauwera The "functionality" has been broken for several months at this point...were there any complaints?

vdauwera:
If the break happened after we last cut a GATK release (nov 16 iirc) we wouldn't expect any complaints.

I thought we had asked the ncbi devs to address the test timeout issues. @ronlevine I think you were involved, do you remember anything about that?

ronlevine:
@vdauwera If anyone was in communication with them, it was @lbergelson and/or @droazen.

droazen:
We have not been in communication with them since the initial PR.

vdauwera:
@a-nikitiuk Would you be able to comment on the status of SRA code and functionality?

yfarjoun:
seems like no-one cares...we removed SRA from the testing framework....so no it will just rot away.

PR #928, opened 2017-07-06
Proposal to evolve to HTSJDK version 3 (DO NOT MERGE)

magicDGS:
### Description

Because the governance model is gonna change soon (#871), I would like to revive the discussion about several issues about the API and proposing to evolve to HTSJDK in a join effort by the community. The related issues are the following:

* Non-backwards compatible API (https://github.com/samtools/htsjdk/issues/520), called in the proposal HTSJDK3. This includes implementing new interfaces as a public API independent of the implementation (see https://github.com/samtools/htsjdk/pull/868 for a discussion about the read interface).
* Evolving API (https://github.com/samtools/htsjdk/issues/378). By moving to HTSJDK-3 we can remove deprecated code and start from scratch. In addition, annotations for beta features could be implemented for the HTSJDK-3 release, marking as beta features that may change.
* Sub-module structure to reduce dependencies in the core functionality (https://github.com/samtools/htsjdk/issues/896). This also includes the issue swith the SRA support (e.g., https://github.com/samtools/htsjdk/issues/414 and https://github.com/samtools/htsjdk/issues/894), which can be moved to a different package.
* Mixing license (https://github.com/samtools/htsjdk/issues/917). By moving to HTSJDK-3, we can ask for porting the code to the new implementation using a common license.

### Proposal

The main idea is to create a folder structure as the one in `htsjdk3-dev` in the root folder of the project and move the current version of htsjdk to a sub-module called htsjdk-2 (not done in the PR for reduce the number of changes). Tasks in gradle to build the current version should be modified to keep the same behavior, and a new task to build the developmental version 3 should be included too.

In addition, I drafted a new annotation for mark the status of the classes with respect to version 3 (`@Htsjdk3`). As an example, I ported the `Locatable` and `Feature` interfaces to the new package. This will allow to see the progress of the new version and keep in sync fixes between versions.

I know that this is a huge proposal, but I expect that most of the code will require just copy-paste and/or repackage (and maybe adding a `@Beta` annotation for some code that may be changed in the future). I think that the community will benefit from this change to create a more consistent API for NGS data processing. In addition, some [dev issues](https://github.com/samtools/htsjdk/issues?q=is%3Aissue+is%3Aopen+label%3Adev) could be solved on the port to HTSJDK-3.

Bringing the attention to the maintainers of the project: @tfenne, @lbergelson, @jacarey

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/htsjdk3-dev/htsjdk-core/README.md
+++ b/htsjdk3-dev/htsjdk-core/README.md
@@ -1,0 +1,7 @@
+# HTSJDK-CORE
+
+Contains common code and interface/api definitions.
+
+Includes SAM/BAM, VCF and tribble support.
+
+Should not include significant dependencies.
--- a/htsjdk3-dev/htsjdk-core/src/main/java/htsjdk/samtools/package-info.java
+++ b/htsjdk3-dev/htsjdk-core/src/main/java/htsjdk/samtools/package-info.java
@@ -1,0 +1,6 @@
+/**
+ * Contains SAM support.
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+package htsjdk.samtools;
--- a/htsjdk3-dev/htsjdk-core/src/main/java/htsjdk/tribble/Feature.java
+++ b/htsjdk3-dev/htsjdk-core/src/main/java/htsjdk/tribble/Feature.java
@@ -1,7 +1,7 @@
 /*
- * The MIT License
+ * The MIT License (MIT)
  *
- * Copyright (c) 2013 The Broad Institute
+ * Copyright (c) 2017 Daniel Gomez-Sanchez
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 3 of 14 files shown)


lindenb:
yes please !

> The main idea is to create a folder structure as the one in htsjdk3-dev

just a suggestion: project like apache-math have created a package-version since v3. 

* org.apache.commons.math3
* org.apache.commons.math4

so we could have

`import htsjdk.*`
or
`import htsjdk3.*`

but I don't know what is the best option

magicDGS:
@lindenb, I was thinking about the new package too. Nevertheless, I opted for proposing this as a temporary solution and then keep the same package for two reasons:

* If there is a new package `htsjdk3`, what will happen with the previous one? Will be removed from the repo at some point or they will continue living together? It means that `htsjdk3`will be implemented in a different repo and `htsjdk` stay here for historical reasons?
* If both `htsjdk` and `htsjdk3`artifacts exists, that could bring the problem of pulling down the library twice, which I personally don't like; I rather prefer a complete change of name in case that the library gets that different to need a distinction of packages. Although for development I think that it could be a good idea...

magicDGS:
Maybe the maintainers can chime in here: @tfenne, @lbergelson, @jacarey? Thank you in advance!

yfarjoun:
Given that there's a new repo for htsjdk-next-beta, can we close this PR?

PR #933, opened 2017-07-10
Write vcf to stdout

nh13:
Related to: https://github.com/broadinstitute/picard/issues/863

If we call `.setOutputFile` with `"/dev/stdout"`, then `this.outputStream` is `null`, this.outFile` is `File("/dev/stdout")`, and `this.outType` is `OutputType.VCF_STREAM`.  When calling `build()`, it will throw a NPE!

### DIFF
--- a/src/main/java/htsjdk/samtools/util/IOUtil.java
+++ b/src/main/java/htsjdk/samtools/util/IOUtil.java
@@ -366,7 +366,7 @@ public class IOUtil {
      * and if it is a file then not a directory and is readable.  If any
      * condition is false then a runtime exception is thrown.
      *
-     * @param files the list of files to check for readability
+     * @param inputs the list of files to check for readability
      */
     public static void assertInputsAreValid(final List<String> inputs) {
         for (final String input : inputs) assertInputIsValid(input);
@@ -465,7 +465,7 @@ public class IOUtil {
     public static void assertFilesEqual(final File f1, final File f2) {
         try {
             if (f1.length() != f2.length()) {
-                throw new SAMException("Files " + f1 + " and " + f2 + " are different lengths.");
+                throw new SAMException("File " + f1 + " is " + f1.length() + " bytes but file " + f2 + " is " + f2.length() + " bytes.");
             }
             final FileInputStream s1 = new FileInputStream(f1);
             final FileInputStream s2 = new FileInputStream(f2);
--- a/src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
@@ -406,8 +406,9 @@ public class VariantContextWriterBuilder {
                 typeToBuild = OutputType.BCF_STREAM;
         }
 
+        // If we are writing to a file, or a special file type (ex. pipe) where the stream is not yet open.
         OutputStream outStreamFromFile = this.outStream;
-        if (FILE_TYPES.contains(this.outType)) {
+        if (FILE_TYPES.contains(this.outType) || (STREAM_TYPES.contains(this.outType) && this.outStream == null)) {
             try {
                 outStreamFromFile = IOUtil.maybeBufferOutputStream(new FileOutputStream(outFile), bufferSize);
             } catch (final FileNotFoundException e) {
@@ -445,7 +446,7 @@ public class VariantContextWriterBuilder {
                 if (options.contains(Options.INDEX_ON_THE_FLY))
                     throw new IllegalArgumentException("VCF index creation not supported for stream output.");
 
-                writer = createVCFWriter(null, outStream);
+                writer = createVCFWriter(null, outStreamFromFile);
                 break;
             case BCF_STREAM:
                 if (options.contains(Options.INDEX_ON_THE_FLY))
--- a/src/test/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilderUnitTest.java
+++ b/src/test/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilderUnitTest.java
@@ -396,4 +396,12 @@ public class VariantContextWriterBuilderUnitTest extends VariantBaseTest {
             Assert.assertFalse(builder.isOptionSet(option)); // has been unset
         }
     }
+
+    @Test
+    public void testStdOut() {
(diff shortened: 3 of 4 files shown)


ronlevine:
Looks good. Please add a test.

nh13:
@ronlevine. Want to take this PR on and add a test?

ronlevine:
Will do.

ronlevine:
@nh13 I added a test. It showed that the NPE was really due to `stream == null`  in `IndexingVariantContextWriter.writerName`.

nh13:
You fixed it so it doesn't give a NPE but not the actual problem. I think the actual problem is that outFile is set but the outType is a VCF_STREAM when you use setOutputFile with /deck/stdout. Can you try again? It may be helpful to revert your commit in Picard and trace the original NPE. Note that we shouldn't be indexing of the output is a stream!!!

ronlevine:
I already tested my reverted commit and found that the NPE is in `IndexingVariantContextWriter.writerName`. I'll check in my "reverted" Picard branch soon with an updated test.

ronlevine:
My Picard  branch with the reverted changes and fixed test is `rhl_uvsd_fix_stdout`. The test's file descriptor magic for capturing stdout output was processed as a VCF instead of VCF_STREAM. A test for a gzipped file will be added.

ronlevine:
Here's the stack trace from rhl_uvsd_fix_stdout 
```
Gradle suite > Gradle test > picard.vcf.UpdateVcfSequenceDictionaryTest.testUpdateVcfSequenceDictionary[1](/dev/stdout) FAILED
        at htsjdk.variant.variantcontext.writer.IndexingVariantContextWriter.writerName(IndexingVariantContextWriter.java:181)
        at htsjdk.variant.variantcontext.writer.VCFWriter.<init>(VCFWriter.java:86)
        at htsjdk.variant.variantcontext.writer.VariantContextWriterBuilder.createVCFWriter(VariantContextWriterBuilder.java:513)
        at htsjdk.variant.variantcontext.writer.VariantContextWriterBuilder.build(VariantContextWriterBuilder.java:448)
        at picard.vcf.UpdateVcfSequenceDictionary.doWork(UpdateVcfSequenceDictionary.java:92)
        at picard.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:205)
        at picard.vcf.UpdateVcfSequenceDictionaryTest.testUpdateVcfSequenceDictionary(UpdateVcfSequenceDictionaryTest.java:73)
```

ronlevine:
Added a `vcf.gz` file test in the Picard branch, `rhl_uvsd_fix_stdout`.  The added test decompresses the `vcf.gz` file and verifies it's contents.

ronlevine:
@nh13 I think this is good to go. I successfully tested this fix as follows:
* Build htsjdk: ./gradlew install printVersion
* Build the [Picard branch](https://github.com/broadinstitute/picard/compare/rhl_uvsd_fix_stdout): /gradlew shadowJar -Dhtsjdk.version=2.10.1-21-g309914c-SNAPSHOT
* Run test in the Picard branch: . ./gradlew test -Dtest.single=UpdateVcfSequenceDictionaryTest -Dhtsjdk.version=2.10.1-21-g309914c-SNAPSHOT

ronlevine:
@yfarjoun Please review.

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -465,7 +465,7 @@ else if (!dir.canRead()) {
     public static void assertFilesEqual(final File f1, final File f2) {
         try {
             if (f1.length() != f2.length()) {
-                throw new SAMException("Files " + f1 + " and " + f2 + " are different lengths.");
+                throw new SAMException("Files " + f1 + " is " + f1.length() + " bytes but " + f2 + " is " + f2.length() + " bytes.");
```
nh13: 1. `Files` -> `File`
2. `...bytes but ` -> `...butes but file `.

nh13:
@ronlevine one minor change from my end, then @jacarey

src/main/java/htsjdk/samtools/util/IOUtil.java
ronlevine: Done.

ronlevine:
@jacarey Please take a look.

src/test/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilderUnitTest.java
```diff
@@ -396,4 +396,11 @@ public void testModifyOption() {
             Assert.assertFalse(builder.isOptionSet(option)); // has been unset
         }
     }
+
+    @Test
+    public void testStdOut() {
+        final VariantContextWriter writer = new VariantContextWriterBuilder().setOutputFile("/dev/stdout").clearOptions().build();
+        Assert.assertNotEquals(((VCFWriter) writer).getOutputStream(), null);
```
jacarey: Assert.notNull

src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
```diff
@@ -445,7 +446,8 @@ else if (STREAM_TYPES.contains(this.outType))
                 if (options.contains(Options.INDEX_ON_THE_FLY))
                     throw new IllegalArgumentException("VCF index creation not supported for stream output.");
 
-                writer = createVCFWriter(null, outStream);
+                final OutputStream writerStream = outStream != null ? outStream : outStreamFromFile;
```
jacarey: If I'm reading this correctly in the case where you have VCF_STREAM as the type and outString != null we would have set outStreamFromFile to outStream on line 410. In which case we don't need writerStream at all and can just use
createVCFWriter(null, outStreamFromFile) on like 450

src/test/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilderUnitTest.java
ronlevine: Done.

src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
ronlevine: Correct, good catch!

PR #936, opened 2017-07-19
Fix no running tests due to lack of HtsjdkTest marker

magicDGS:
### Description

Some tests were not running because they lack the HtsjdkTest marker. This PR adds the missing marker to these classes, which now are executed by the test task.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


magicDGS:
@tfenne or @lbergelson, can you please review this one? It was reducing the test coverage and including new tests without noticing...

lbergelson:
@magicDGS Good catch!  
@tfenne @jacarey We need to add a mechanism for making sure this doesn't continue to happen as new test classes are added or edited.

PR #937, opened 2017-07-19
Proposal: deprecate or remove SAMTools.java

magicDGS:
### Description

The `SAMTools` class does not add anything to the HTSJDK library except an example of how to iterate over a SAM/BAM file. In addition, it is not tested. I propose to either deprecate (first commit) or delete (second commit) this class.

I think that if HTSJDK is a library, we should clean up untested classes that does not add anything to the codebase, and are not really useful for an API user. 

### Checklist

- [x] Code compiles correctly
- [ ] All tests passing


lbergelson:
@magicDGS I agree with removing this.  I think we should deprecate it first though for at least a release or two on the off chance that someone is relying on it for something.

magicDGS:
Sure @lbergelson. I reverted the last commit.

lbergelson:
@magicDGS 👍

magicDGS:
Thanks @lbergelson!

PR #939, opened 2017-07-20
Increase test coverage for Cigar components

magicDGS:
### Description

Includes full coverage for `CigarElement` and `CigarOperator`, and new tests for `Cigar` (including explicit tests for already covered methods).

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/test/java/htsjdk/samtools/util/CigarElementUnitTest.java
+++ b/src/test/java/htsjdk/samtools/util/CigarElementUnitTest.java
@@ -4,6 +4,8 @@ package htsjdk.samtools.util;
 import htsjdk.HtsjdkTest;
 import htsjdk.samtools.CigarElement;
 import htsjdk.samtools.CigarOperator;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
 public class CigarElementUnitTest extends HtsjdkTest {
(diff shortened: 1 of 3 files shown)


src/test/java/htsjdk/samtools/util/CigarElementUnitTest.java
```diff
@@ -12,4 +14,28 @@
...
+                {mElement, new CigarElement(mElement.getLength(), CigarOperator.X), false},
+                // different class
+                {mElement, mElement.toString(), false}
+        };
+    }
+
+    @Test(dataProvider = "elementsForEquals")
+    public void testEqualsAndHashCode(final CigarElement element, final Object other, final boolean isEquals) {
+        Assert.assertEquals(element.equals(other), isEquals);
+        Assert.assertEquals(element.hashCode() == other.hashCode(), isEquals);
```
lbergelson: I think this test isn't technically correct.  It's possible 
for `A != B` but `A.hashCode() == B.hashCode()`.   I think the correct assertion for the the hashCode part should be.

```
if( element.equals(other)) {
 Assert.assertEquals(element.hashCode(), other.hashCode());
}
```

magicDGS: True! My bad! But in the condition I used the expected isEquals instead.

magicDGS:
Addressed the comment, @lbergelson!

magicDGS:
Thanks @lbergelson!

PR #952, opened 2017-07-27
Deprecate FastqRecord.toString to warn users for a future change in behaviour

magicDGS:
### Description

The current implementation of `FastqRecord.toString()` is using a FASTQ-formatted String. Usually, the `toString()` method should implement a description of the object that should be useful for the developer, but not for rely on it for usage on other parts of the code (such a writer). Because there is a TODO for changing the method for something more simple, this PR deprecates the current implementation to warn the users to use the `FastqRecord.toFastQString()` method instead, and change the behaviour in a future release.

In addition, this PR includes changes in the javadoc to point out where a null value could be returned and adds tests for a "null" record.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/fastq/FastqRecord.java
+++ b/src/main/java/htsjdk/samtools/fastq/FastqRecord.java
@@ -101,7 +101,7 @@ public class FastqRecord implements Serializable {
     /**
      * Get the read name
      *
-     * @return the read name
+     * @return the read name (may be {@code null}).
      */
     public String getReadName() {
         return readName;
@@ -110,7 +110,7 @@ public class FastqRecord implements Serializable {
     /**
      * Get the DNA sequence
      *
-     * @return read sequence as a string of ACGTN=.
+     * @return read sequence as a string of ACGTN= (may be {@code null}).
      */
     public String getReadString() {
         return readString;
@@ -128,7 +128,7 @@ public class FastqRecord implements Serializable {
     /**
      * Get the base qualities encoded as a FASTQ string
      *
-     * @return the quality string
+     * @return the quality string (may be {@code null}).
      */
     public String getBaseQualityString() {
         return baseQualityString;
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/fastq/FastqRecord.java
```diff
@@ -228,8 +228,13 @@ public String toFastQString() {
...
 
     /**
-     * Returns {@link #toFastQString()}
+     * Returns {@link #toFastQString()}.
+     *
+     * @deprecated since 07/2017. This method will be deprecated until it is changed in the future
+     * for a simpler representation of the object. For code relying on the formatting as a FASTQ
+     * String, please refactor your code to use {@link #toFastQString()}.
      */
+    @Deprecated
```
lbergelson: I like the idea of warning people, but in general but I'm not sure what the behavior of deprecating a method on Object is... It seems strange.

magicDGS: I believe that deprecating concrete implementation of `Object.toString` is a good way to warn in direct uses of the method for developers; if they are using the `toString` only from the object perspective, they shouldn't get a warning.

If you have any proposal on how to warn users about this to be able to change the behaviour in the next one/two releases, I can change the deprecation of the class in this PR.

lbergelson:
@magicDGS I think I still disagree with deprecating toString().  It's just a very strange and unexpected thing to do.  Would you want update this so we don't do the deprecation, but do get the comments and test?

magicDGS:
Done. But I would like to make sure that people start moving away from the `toString()` method - maybe adding some verbosity (logging? stderr?) when called to make clear that they should not use it? What do you think @lbergelson?

lbergelson:
@magicDGS Thank you very much.

lbergelson:
I think we'll probably do the switch over as part of the changes in htsjdk3, so people will be expecting changes there.

PR #953, opened 2017-07-27
Simplify BufferedLineReader by using LineNumberReader

magicDGS:
### Description

`BufferedLineReader` had a `TODO` to be replaced by the `LineNumberReader` instead. This commit simplify the `BufferedLineReader` by extending `LineNumberReader`.

If the original purpose is to remove completely the `BufferedLineReader` class, this commit can also deprecate it and replace the usages for the `LineNumberReader`. Nevertheless, I do not know if this is possible because I haven't check it.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/BufferedLineReader.java
+++ b/src/main/java/htsjdk/samtools/util/BufferedLineReader.java
@@ -29,29 +29,24 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.LineNumberReader;
 import java.nio.charset.Charset;
 
 /**
  * Implementation of LineReader that is a thin wrapper around BufferedReader.  On Linux, this is faster
  * than AsciiLineReaderImpl.  If you use AsciiLineReader rather than this class, it will detect the OS
  * and delegate to the preferred implementation.
- *
- * TODO: Replace this with {@link java.io.LineNumberReader}?
  * 
  * @author alecw@broadinstitute.org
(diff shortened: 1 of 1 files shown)


pshapiro4broad:
Can you also re-implement `StringLineReader` using `BufferedLineReader` with a `ByteArrayInputStream`?

magicDGS:
Hi @pshapiro4broad, I will do in a different PR to keep the two changes independent.

pshapiro4broad:
@magicDGS I may be wrong, but I believe `StringLineReader` can be replaced with something like:
```
public class StringLineReader extends BufferedLineReader {
    public StringLineReader(final String s) {
        super(new ByteArrayInputStream(s.getBytes()));
    }
}
```
It's not really part of what you're doing, but it seemed like it would be an easy fix if it worked.

magicDGS:
@pshapiro4broad: If that simplification works (that I'm sure it will), I think that I will open a different PR to make the simplification. In addition, I think that in that case the class is unnecessary and could be deprecated in favor of an static method in `BufferedLineReader` to construct from a `String`.

Anyway, could you merge this branch if it's ok or nominate someone with credentials to do it, @pshapiro4broad?

pshapiro4broad:
I agree with your suggestion to deprecate and add a static method, and use a separate PR.

I don't have write access to this repo. I think @lbergelson can help.

lbergelson:
@magicDGS 👍 Merging this.

magicDGS:
Thanks @lbergelson!

PR #956, opened 2017-07-28
Add support for SQ-AN in SAMSequenceDictionary

magicDGS:
### Description

Adding the support for the new `@SQ` AN tag for alternative sequences (https://github.com/samtools/hts-specs/pull/220).

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
+++ b/src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
@@ -86,6 +86,8 @@ public class SAMSequenceDictionary implements Serializable {
             if (mSequenceMap.put(record.getSequenceName(), record) != null) {
                 throw new IllegalArgumentException("Cannot add sequence that already exists in SAMSequenceDictionary: " +
                         record.getSequenceName());
+            } else {
+                record.getAlternativeSequeneNames().forEach(an -> addSequenceAlias(record.getSequenceName(), an));
             }
         }
     }
@@ -98,6 +100,7 @@ public class SAMSequenceDictionary implements Serializable {
         sequenceRecord.setSequenceIndex(mSequences.size());
         mSequences.add(sequenceRecord);
         mSequenceMap.put(sequenceRecord.getSequenceName(), sequenceRecord);
+        sequenceRecord.getAlternativeSequeneNames().forEach(an -> addSequenceAlias(sequenceRecord.getSequenceName(), an));
     }
 
     /**
@@ -220,6 +223,9 @@ public class SAMSequenceDictionary implements Serializable {
      * <code>1,chr1,chr01,01,CM000663,NC_000001.10</code> e.g:
      * <code>MT,chrM</code>
      *
+     * NOTE: this method does not add the alias to the alternative sequence name tag (AN) in the SAMSequenceRecord.
+     * If you would like to add it to the AN tag, use {@link #setAlternativeSequenceName(String, String)} instead.
+     *
      * @param originalName
      *            existing contig name
      * @param altName
@@ -244,6 +250,27 @@ public class SAMSequenceDictionary implements Serializable {
         }
         mSequenceMap.put(altName, originalSeqRecord);
         return originalSeqRecord;
+    }
+
+    /**
+     * Add an alternative sequence name (AN tag) to a SAMSequenceRecord, including it into the aliases
+     * to retrieve the contigs (as with {@link #addSequenceAlias(String, String)}.
+     *
+     * <p>This can be use to provide some alternate names fo a given contig. e.g:
+     * <code>1,chr1,chr01,01,CM000663,NC_000001.10</code> or
+     * <code>MT,chrM</code>.
+     *
+     * @param originalName
+     *            existing contig name
+     * @param altName
+     *            new contig name
+     * @return the contig associated to the 'originalName/altName', with the AN tag including the altName
+     */
+    public SAMSequenceRecord addAlternativeSequenceName(final String originalName,
+            final String altName) {
+        final SAMSequenceRecord record = addSequenceAlias(originalName, altName);
+        record.addAlternativeSequenceName(altName);
+        return record;
     }
 
     /**
--- a/src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
+++ b/src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
@@ -224,7 +224,7 @@ public class SAMSequenceDictionary implements Serializable {
      * <code>MT,chrM</code>
      *
      * NOTE: this method does not add the alias to the alternative sequence name tag (AN) in the SAMSequenceRecord.
-     * If you would like to add it to the AN tag, use {@link #setAlternativeSequenceName(String, String)} instead.
+     * If you would like to add it to the AN tag, use {@link #addAlternativeSequenceName(String, String)} instead.
      *
      * @param originalName
      *            existing contig name
--- a/src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
+++ b/src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
@@ -87,7 +87,7 @@ public class SAMSequenceDictionary implements Serializable {
                 throw new IllegalArgumentException("Cannot add sequence that already exists in SAMSequenceDictionary: " +
                         record.getSequenceName());
             } else {
-                record.getAlternativeSequeneNames().forEach(an -> addSequenceAlias(record.getSequenceName(), an));
+                record.getAlternativeSequenceNames().forEach(an -> addSequenceAlias(record.getSequenceName(), an));
             }
         }
     }
@@ -100,7 +100,7 @@ public class SAMSequenceDictionary implements Serializable {
         sequenceRecord.setSequenceIndex(mSequences.size());
         mSequences.add(sequenceRecord);
         mSequenceMap.put(sequenceRecord.getSequenceName(), sequenceRecord);
-        sequenceRecord.getAlternativeSequeneNames().forEach(an -> addSequenceAlias(sequenceRecord.getSequenceName(), an));
+        sequenceRecord.getAlternativeSequenceNames().forEach(an -> addSequenceAlias(sequenceRecord.getSequenceName(), an));
     }
 
     /**
@@ -206,7 +206,11 @@ public class SAMSequenceDictionary implements Serializable {
         return !thatSequences.hasNext();
     }
 
-    /** returns true if the two dictionaries are the same, aliases are NOT considered */
+    /**
+     * Returns {@code true} if the two dictionaries are the same.
+     *
+     * <p>NOTE: Aliases are NOT considered, but alternative sequence names (AN tag) names ARE.
+     */
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
@@ -223,13 +227,11 @@ public class SAMSequenceDictionary implements Serializable {
      * <code>1,chr1,chr01,01,CM000663,NC_000001.10</code> e.g:
      * <code>MT,chrM</code>
      *
-     * NOTE: this method does not add the alias to the alternative sequence name tag (AN) in the SAMSequenceRecord.
+     * <p>NOTE: this method does not add the alias to the alternative sequence name tag (AN) in the SAMSequenceRecord.
      * If you would like to add it to the AN tag, use {@link #addAlternativeSequenceName(String, String)} instead.
      *
-     * @param originalName
-     *            existing contig name
-     * @param altName
-     *            new contig name
+     * @param originalName  existing contig name
+     * @param altName       new contig name
      * @return the contig associated to the 'originalName/altName'
      */
     public SAMSequenceRecord addSequenceAlias(final String originalName,
@@ -257,13 +259,11 @@ public class SAMSequenceDictionary implements Serializable {
      * to retrieve the contigs (as with {@link #addSequenceAlias(String, String)}.
      *
      * <p>This can be use to provide some alternate names fo a given contig. e.g:
-     * <code>1,chr1,chr01,01,CM000663,NC_000001.10</code> or
+     * <code>1,chr1,chr01,01,CM000663</code> or
      * <code>MT,chrM</code>.
      *
-     * @param originalName
-     *            existing contig name
-     * @param altName
-     *            new contig name
+     * @param originalName  existing contig name
+     * @param altName       new contig name
      * @return the contig associated to the 'originalName/altName', with the AN tag including the altName
      */
     public SAMSequenceRecord addAlternativeSequenceName(final String originalName,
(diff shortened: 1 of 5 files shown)


src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -247,6 +253,27 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
+     *
+     * @param originalName
+     *            existing contig name
+     * @param altName
+     *            new contig name
+     * @return the contig associated to the 'originalName/altName', with the AN tag including the altName
+     */
+    public SAMSequenceRecord addAlternativeSequenceName(final String originalName,
+            final String altName) {
+        final SAMSequenceRecord record = addSequenceAlias(originalName, altName);
```
pshapiro4broad: According to codecov this new code isn't covered by a test.

magicDGS: Done.

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -139,6 +148,32 @@ private void setSequenceName(final String name) {
     // Private state used only by SAM implementation.
     public void setSequenceIndex(final int value) { mSequenceIndex = value; }
 
+    /** Returns unmodifiable set with alternative sequence names. */
+    @XmlTransient //we use the field instead of getter/setter
+    public Set<String> getAlternativeSequeneNames() {
```
lbergelson: typo: getAlternative**Sequene**Names()

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -159,6 +194,7 @@ public boolean isSameSequence(final SAMSequenceRecord that) {
             }
         }
         else {
+            // TODO: should this also check in the alternative sequences names?
```
lbergelson: It seems like this should match on alternative sequence names if present

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -184,6 +220,7 @@ public boolean equals(final Object o) {
         if (mSequenceLength != that.mSequenceLength) return false;
         if (!attributesEqual(that)) return false;
         if (mSequenceName != that.mSequenceName) return false; // Compare using == since we intern() the name
+        // TODO: should this also compare the alternative names?
```
lbergelson: it seems strange to leave alternative names out of equals especially given the weaker `isSameSequence()` also exists

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -247,6 +253,27 @@ public SAMSequenceRecord addSequenceAlias(final String originalName,
...
     /**
+     * Add an alternative sequence name (AN tag) to a SAMSequenceRecord, including it into the aliases
+     * to retrieve the contigs (as with {@link #addSequenceAlias(String, String)}.
+     *
+     * <p>This can be use to provide some alternate names fo a given contig. e.g:
+     * <code>1,chr1,chr01,01,CM000663,NC_000001.10</code> or
+     * <code>MT,chrM</code>.
+     *
+     * @param originalName
+     *            existing contig name
```
lbergelson: there are funny line breaks here for some reason

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -43,9 +49,12 @@
 {
     public static final long serialVersionUID = 1L; // AbstractSAMHeaderRecord implements Serializable
     private String mSequenceName = null; // Value must be interned() if it's ever set/modified
+    @XmlAttribute(name = "alternative_sequece_names")
```
lbergelson: typo: alternative_**sequece**_names

src/main/java/htsjdk/samtools/SAMTextHeaderCodec.java
```diff
@@ -498,12 +498,14 @@ private void writeSQLine(final SAMSequenceRecord sequenceRecord) {
...
     protected String getSQLine(final SAMSequenceRecord sequenceRecord) {
-        final int numAttributes = sequenceRecord.getAttributes() != null ? sequenceRecord.getAttributes().size() : 0;
+        int numAttributes = sequenceRecord.getAttributes() != null ? sequenceRecord.getAttributes().size() : 0;
+        if (sequenceRecord.hasAlternativeSequenceNames()) numAttributes++;
         final String[] fields = new String[3 + numAttributes];
         fields[0] = HEADER_LINE_START + HeaderRecordType.SQ;
         fields[1] = SAMSequenceRecord.SEQUENCE_NAME_TAG + TAG_KEY_VALUE_SEPARATOR + sequenceRecord.getSequenceName();
         fields[2] = SAMSequenceRecord.SEQUENCE_LENGTH_TAG + TAG_KEY_VALUE_SEPARATOR + Integer.toString(sequenceRecord.getSequenceLength());
         encodeTags(sequenceRecord, fields, 3);
+        if (sequenceRecord.hasAlternativeSequenceNames()) fields[fields.length - 1] = SAMSequenceRecord.ALTERNATIVE_SEQUENCE_NAME_TAG + TAG_KEY_VALUE_SEPARATOR + String.join(",", sequenceRecord.getAlternativeSequeneNames());
```
lbergelson: Why does this need special handling?  it seems like the alternate contig names should be added to the attributes map and handled the same way as the other attributes.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -220,6 +223,9 @@ public boolean equals(Object o) {
      * <code>1,chr1,chr01,01,CM000663,NC_000001.10</code> e.g:
```
lbergelson: I think the comment on the equals method above needs updating to say that alias's are not considered but Alternative Sequences are.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
lbergelson: oh, it matches the weird line breaks above, lets fix those too...

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -43,9 +49,12 @@
...
+    private Set<String> mAlternativeSequenceName = new LinkedHashSet<>();
```
lbergelson: This should be handled the same as the other attributes, stored in the attributes map.  It means doing annoying string processing when you add a single alias to an existing set of aliases, but I think making it consistent is worthwhile

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -139,6 +148,32 @@ private void setSequenceName(final String name) {
...
 
+    /** Returns unmodifiable set with alternative sequence names. */
+    @XmlTransient //we use the field instead of getter/setter
+    public Set<String> getAlternativeSequeneNames() {
+        return Collections.unmodifiableSet(mAlternativeSequenceName);
+    }
+
+    /** Adds an alternative sequence name if it is not the same as the sequence name or it is not present already. */
+    public void addAlternativeSequenceName(final String name) {
+        if (!mSequenceName.equals(name) && ! mAlternativeSequenceName.contains(name) ) {
```
lbergelson: These should change to store the names in the attributes map so that you don't need special handling for all the places that handle the tags.

lbergelson: this and the set method should validate that the alternative names are legal according to the regex provided in the sam spec.

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
magicDGS: Done.

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
magicDGS: Done by comparing the sets.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
magicDGS: Done.

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
magicDGS: Done.

magicDGS:
Friendly reminder here, @lbergelson!

lbergelson:
@magicDGS Hey, the changes you made here look good.  I've been debating with myself what the right solution is though.  I hadn't thought through what a large potential performance impact the changes I asked for would be.  I'm torn because it's much cleaner than the alternative, but introducing a gross performance regression for minor code cleanlyness seems wrong.

magicDGS:
Thanks for the reply @lbergelson - should I revert the last commit?

magicDGS:
Friendly ping here, @lbergelson!

jmarshall:
`ALTERNATIVE_SEQUENCE_NAME_REGEXP` is missing `_` and `|` compared to the definition of AN in the SAM spec. Is there a reason for this?

magicDGS:
I can't remember - maybe it was still under discussion if they should be allowed or I just forgot to add them. Changed in next commit matching the current definition.

jmarshall:
Apologies and FYI: After further contemplation of the issue (see https://github.com/samtools/hts-specs/issues/124#issuecomment-393777566), I suspect we may end up allowing colons in AN after all.

magicDGS:
@jmarshall - also in the `AN`? I think that not allowing it in the alternative-name is a good solution for unambiguously define contigs with colons: although the main contig name has colons, if the `AN` does not then it is safe to use to define intervals...

jmarshall:
Yes, also in AN. See the draft PR samtools/hts-specs#333.

I used to think as you do, but… For better or worse these things exist in the wild, so IMHO we have to give up on thinking AN might be a transition mechanism away from allowing them. Reread that comment, which contains pseudocode showing how tools should parse their interval notation even in the presence of stupid colons in names (and I plan to turn that into an appendix for the spec).

jmarshall:
FYI RNAME changes have now landed in the SAM spec: in particular the allowed characters (which include colons!) for SN and AN are now the same, so a separate `ALTERNATIVE_SEQUENCE_NAME_REGEXP` is unnecessary.

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -67,14 +76,18 @@
...
+            new HashSet<String>(Arrays.asList(SEQUENCE_NAME_TAG, SEQUENCE_LENGTH_TAG, ASSEMBLY_TAG, ALTERNATIVE_SEQUENCE_NAME_TAG, MD5_TAG, URI_TAG,
                                                 SPECIES_TAG));
 
     // Split on any whitespace
     private static Pattern SEQUENCE_NAME_SPLITTER = Pattern.compile("\\s");
     // These are the chars matched by \\s.
     private static char[] WHITESPACE_CHARS = {' ', '\t', '\n', '\013', '\f', '\r'}; // \013 is vertical tab
 
+    // alternative sequence name regexp
+    private static final Pattern ALTERNATIVE_SEQUENCE_NAME_REGEXP = Pattern.compile("[0-9A-Za-z][0-9A-Za-z*+.@_|-]*");
```
nh13: no longer needed, use `SAMSequenceRecord.LEGAL_RNAME_PATTERN`

yfarjoun:
let's resolve this and move it past the finish line....

nh13:
@yfarjoun agreed. I am already using Picard and this PR to store alternative coffee tog names, and I actually use them! https://github.com/fulcrumgenomics/fgbio/pull/467

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -139,6 +152,49 @@ private void setSequenceName(final String name) {
     // Private state used only by SAM implementation.
     public void setSequenceIndex(final int value) { mSequenceIndex = value; }
 
+    /** Returns a set with alternative sequence names. Modifications will not affect the record instance. */
+    @XmlAttribute(name = "alternative_sequence_names")
```
yfarjoun: remove this line

src/test/java/htsjdk/samtools/SAMSequenceRecordTest.java
```diff
@@ -83,4 +85,72 @@ public void testIsSameSequence(final SAMSequenceRecord rec1 , final SAMSequenceR
...
+
+    @DataProvider
+    public Object[][] invalidAlternativeSequences() {
+        return new Object[][] {
+                // invalid start
+                {"@chr1"}, {",chr1"},
+                // comma-separated
+                {"chr1,alt"},
+                // coordinate-like
+                {"chr1:1000"}, {"chr1:100-200"}
```
yfarjoun: add a testcase with "{" and "}"

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -139,6 +152,49 @@ private void setSequenceName(final String name) {
...
+    public Set<String> getAlternativeSequenceNames() {
+        final String anTag = getAttribute(ALTERNATIVE_SEQUENCE_NAME_TAG);
+        return (anTag == null) ? new LinkedHashSet<>() : new LinkedHashSet<>(Arrays.asList(anTag.split(ALTERNATIVE_SEQUENCE_NAME_SEPARATOR)));
+    }
+
+    /** Adds an alternative sequence name if it is not the same as the sequence name or it is not present already. */
+    public void addAlternativeSequenceName(final String name) {
+        validateAltRegExp(name);
+        final Set<String> altSequences = getAlternativeSequenceNames();
+        if (!mSequenceName.equals(name) && ! altSequences.contains(name) ) {
```
pshapiro4broad: inconsistent whitespace
```suggestion
        if (!mSequenceName.equals(name) && !altSequences.contains(name)) {
```

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -139,6 +152,49 @@ private void setSequenceName(final String name) {
     // Private state used only by SAM implementation.
     public void setSequenceIndex(final int value) { mSequenceIndex = value; }
 
+    /** Returns a set with alternative sequence names. Modifications will not affect the record instance. */
+    @XmlAttribute(name = "alternative_sequence_names")
+    public Set<String> getAlternativeSequenceNames() {
+        final String anTag = getAttribute(ALTERNATIVE_SEQUENCE_NAME_TAG);
+        return (anTag == null) ? new LinkedHashSet<>() : new LinkedHashSet<>(Arrays.asList(anTag.split(ALTERNATIVE_SEQUENCE_NAME_SEPARATOR)));
```
pshapiro4broad: This shouldn't return a modifiable set, should it? E.g. I would use `Collections.emptySet()` for the `null` case.

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
yfarjoun: fixed

yfarjoun:
I'm closing this PR as I don't know how to push to daniel's fork....I'll open it with a new branch...sorry for the inconvenience.

PR #959, opened 2017-07-31
Change contract for getAnyAttribute to return null for missing values

### LINKED ISSUE #957: Missing value in integer FORMAT field returns -1
Given this VCF file:

```
##fileformat=VCFv4.2
##FORMAT=<ID=DP,Number=1,Type=Integer,Description="test">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	sample1
chr1	10	.	C	T	.	.	.	DP	.
```

I would expect the DP field to return null when queried with `getAnyAttribute("DP")` instead it returns -1. Test:

```
@Test
public void test(){
	VCFFileReader reader = new VCFFileReader(new File("test.vcf.gz"));
	VariantContext ctx = reader.query("chr1", 1, 100).next();

	System.err.println(ctx.getGenotype(0).getAnyAttribute("DP")); // <- Prints -1
}
```

It seems related to this issue https://github.com/samtools/bcftools/issues/407

Am I missing something?  

This is using htsjdk 2.9.1

magicDGS:
### Description

Proposal to fix  #957, to have a consistent behaviour for missing values in `Genotype.getAnyAttribute()`.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/Genotype.java
+++ b/src/main/java/htsjdk/variant/variantcontext/Genotype.java
@@ -529,33 +529,37 @@ public abstract class Genotype implements Comparable<Genotype>, Serializable {
 
     /**
      * A totally generic getter, that allows you to get specific keys that correspond
-     * to even inline values (GQ, for example).  Can be very expensive.  Additionally,
-     * all <code>int[]</code> are converted inline into <code>List&lt;Integer&gt;</code> for convenience.
+     * to even inline values (GQ, for example). Can be very expensive.
+     *
+     * <p>Note: all {@code int[]} values are converted inline into {@code List<Integer>} for convenience.
+     *
+     * <p>Warning: values associated with missing fields returns always {@code null}. For example,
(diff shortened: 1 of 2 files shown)


yfarjoun:
This is a seriously breaking change that will affect many downstream tools...e.g. every tool that assumed that getGC() will return -1 when unavailable will break. 

I think it's up to the maintainers (@tfenne, @jacarey, @lbergelson) to make the call vis-a-vis both 

1. should we do this; and if so
2. how to do this in the least disruptive and well communicated manner.

lbergelson:
Yeah, I'm torn on this one.  I think it's an improvement in consistency, but I'm sure that it would cause all sorts of confusing errors downstream.

yfarjoun:
closing with the hope that [htsjdk-next-beta](github.com/samtools/htsjdk-next-beta) will solve this issue better.

PR #960, opened 2017-07-31
Broke out the static variables so that they can be used separately.

eitanbanks:
### Description

I want to use the individual constants outside of the list, so need them broken out.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/IOUtil.java
+++ b/src/main/java/htsjdk/samtools/util/IOUtil.java
@@ -86,8 +86,11 @@ public class IOUtil {
     public static final long TWO_GBS = 2 * ONE_GB;
     public static final long FIVE_GBS = 5 * ONE_GB;
 
+    public static final String VCF_FILE_EXTENSION = ".vcf";
+    public static final String BCF_FILE_EXTENSION = ".bcf";
+    public static final String COMPRESSED_VCF_FILE_EXTENSION = ".vcf.gz";
     /** Possible extensions for VCF files and related formats. */
-    public static final String[] VCF_EXTENSIONS = new String[] {".vcf", ".vcf.gz", ".bcf"};
+    public static final String[] VCF_EXTENSIONS = {VCF_FILE_EXTENSION, COMPRESSED_VCF_FILE_EXTENSION, BCF_FILE_EXTENSION};
 
     public static final String INTERVAL_LIST_FILE_EXTENSION = IntervalList.INTERVAL_LIST_FILE_EXTENSION;
 


lbergelson:
@eitanbanks 👍  to merge when tests pass, but can you update the commit name and comment when you merge so it's clear what constants are being split out.

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -87,7 +87,10 @@
     public static final long FIVE_GBS = 5 * ONE_GB;
 
     /** Possible extensions for VCF files and related formats. */
-    public static final String[] VCF_EXTENSIONS = new String[] {".vcf", ".vcf.gz", ".bcf"};
+    public static final String VCF_EXTENSION = ".vcf";
+    public static final String BCF_EXTENSION = ".bcf";
+    public static final String COMPRESSED_VCF_EXTENSION = ".vcf.gz";
+    public static final String[] VCF_EXTENSIONS = new String[] {VCF_EXTENSION, COMPRESSED_VCF_EXTENSION, BCF_EXTENSION};
```
pshapiro4broad: You don't need `new String[]` when you're using an array literal as an initializer. You can write it like this:
```
 public static final String[] VCF_EXTENSIONS = {VCF_EXTENSION, COMPRESSED_VCF_EXTENSION, BCF_EXTENSION};
```

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -87,7 +87,10 @@
     public static final long FIVE_GBS = 5 * ONE_GB;
 
     /** Possible extensions for VCF files and related formats. */
```
pshapiro4broad: This only applies to the field `VCF_EXTENSION`. Maybe it should be moved down to the line before `VCF_EXTENSIONS`. And you could add javadocs for the other new public fields.

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -87,7 +87,10 @@
...
-    public static final String[] VCF_EXTENSIONS = new String[] {".vcf", ".vcf.gz", ".bcf"};
+    public static final String VCF_EXTENSION = ".vcf";
```
pshapiro4broad: Below I see fields named `XXX_FILE_EXTENSION` so the names here aren't consistent with them. I'm not sure how important consistency is in this file.

src/main/java/htsjdk/samtools/util/IOUtil.java
eitanbanks: Bah, I was just copying what was already there...
Alright, fine.

PR #963, opened 2017-08-09
Add convenience method hasAttibute() to SAMRecord

SHuang-Broad:
### Description

As described in title.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMRecord.java
+++ b/src/main/java/htsjdk/samtools/SAMRecord.java
@@ -1359,22 +1359,16 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
     }
 
     /**
-     * @return SAMBinaryTagAndValue may be null if
-     *          1) {@link #mAttributes} is null or
-     *          2) this record does not have the requested attribute
-     */
-    private SAMBinaryTagAndValue getAttributeMayBeNull(final short tag) {
-        if (this.mAttributes == null) return null;
-        return this.mAttributes.find(tag);
-    }
-
-    /**
      * @see SAMRecord#getAttribute(java.lang.String)
      * @param tag Binary representation of a 2-char String tag as created by SAMTagUtil.
      */
     public Object getAttribute(final short tag) {
-        final SAMBinaryTagAndValue mayBeNull = getAttributeMayBeNull(tag);
-        return mayBeNull==null ? null : mayBeNull.value;
+        if (this.mAttributes == null) return null;
+        else {
+            final SAMBinaryTagAndValue tmp = this.mAttributes.find(tag);
+            if (tmp != null) return tmp.value;
+            else return null;
+        }
     }
 
     /**
--- a/src/main/java/htsjdk/samtools/SAMRecord.java
+++ b/src/main/java/htsjdk/samtools/SAMRecord.java
@@ -1116,8 +1116,11 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
         this.mValidationStringency = validationStringency;
     }
 
+    /**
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1352,16 +1356,22 @@ private void requireUnsigned(final String tag) {
     }
 
     /**
+     * @return SAMBinaryTagAndValue may be null if
+     *          1) {@link #mAttributes} is null or
+     *          2) this record does not have the requested attribute
+     */
+    private SAMBinaryTagAndValue hasAttribute(final short tag) {
```
lbergelson: This seems like it's wrong.  It should return a boolean.

lbergelson: Or it could be named something else.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1116,6 +1116,10 @@ public void setValidationStringency(final ValidationStringency validationStringe
         this.mValidationStringency = validationStringency;
     }
 
+    public boolean hasAttribute(final String tag) {
```
lbergelson: needs javadoc

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1116,6 +1116,10 @@ public void setValidationStringency(final ValidationStringency validationStringe
...
+        return hasAttribute(SAMTagUtil.getSingleton().makeBinaryTag(tag))!=null;
```
lbergelson: shouldn't this just be something like `getAttribute != null?`

src/main/java/htsjdk/samtools/SAMRecord.java
SHuang-Broad: done. Thought the name is very obvious...

src/main/java/htsjdk/samtools/SAMRecord.java
SHuang-Broad: done

src/main/java/htsjdk/samtools/SAMRecord.java
SHuang-Broad: renamed

SHuang-Broad:
@lbergelson updated, please take another look. Thanks!

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1352,16 +1359,22 @@ private void requireUnsigned(final String tag) {
     }
 
     /**
+     * @return SAMBinaryTagAndValue may be null if
+     *          1) {@link #mAttributes} is null or
+     *          2) this record does not have the requested attribute
+     */
+    private SAMBinaryTagAndValue getAttributeMayBeNull(final short tag) {
```
lbergelson: @SHuang-Broad Do we need this method at all?  I still don't really understand why we need to extract this out.

yfarjoun: is `getNullableAttribute` more consistent with Java naming schemes?

lbergelson: it's not a naming issue anymore, it's just that if there's only 1 use of this, I don't see the need to extract it at all.

SHuang-Broad:
@lbergelson removed the refactoring. Please take another look. Thanks!

__EDIT__
not sure about what happened to the "15 files have unexpected coverage changes..."

lbergelson:
@SHuang-Broad 👍

PR #967, opened 2017-08-14
adding new method IOUtil.toPath(File)

lbergelson:
### Description

this method converts a `File` to a `Path`, but passes a null through unchanged instead of crashing with an NPE
it is useful when adapting existing code that takes `File`s into code that takes `Path`s and should help avoid introducing new NPEs when converting code

updated existing uses of `file.toPath()` to `IOUtil.toPath(file)`

### Checklist

- [ x ] Code compiles correctly
- [ X ] New tests covering changes and new functionality
- [ x ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SamFiles.java
+++ b/src/main/java/htsjdk/samtools/SamFiles.java
@@ -3,6 +3,7 @@ package htsjdk.samtools;
 import htsjdk.samtools.cram.CRAIIndex;
 import htsjdk.samtools.cram.build.CramIO;
 
+import htsjdk.samtools.util.IOUtil;
 import htsjdk.samtools.util.Log;
 import java.io.File;
 import java.io.IOException;
@@ -24,7 +25,7 @@ public class SamFiles {
      * @return The index for the provided SAM, or null if one was not found.
      */
     public static File findIndex(final File samFile) {
-        Path path = findIndex(samFile.toPath());
+        final Path path = findIndex(IOUtil.toPath(samFile));
         return path == null ? null : path.toFile();
     }
 
--- a/src/main/java/htsjdk/samtools/SamInputResource.java
+++ b/src/main/java/htsjdk/samtools/SamInputResource.java
@@ -271,7 +271,7 @@ class FileInputResource extends InputResource {
 
     @Override
     public Path asPath() {
-        return fileResource.toPath();
+        return IOUtil.toPath(fileResource);
     }
 
     @Override
--- a/src/main/java/htsjdk/samtools/cram/ref/ReferenceSource.java
+++ b/src/main/java/htsjdk/samtools/cram/ref/ReferenceSource.java
@@ -20,12 +20,11 @@ package htsjdk.samtools.cram.ref;
 import htsjdk.samtools.Defaults;
 import htsjdk.samtools.SAMException;
 import htsjdk.samtools.SAMSequenceRecord;
-import htsjdk.samtools.SAMUtils;
-import htsjdk.samtools.cram.build.Utils;
 import htsjdk.samtools.cram.io.InputStreamUtils;
 import htsjdk.samtools.reference.ReferenceSequence;
 import htsjdk.samtools.reference.ReferenceSequenceFile;
 import htsjdk.samtools.reference.ReferenceSequenceFileFactory;
+import htsjdk.samtools.util.IOUtil;
 import htsjdk.samtools.util.Log;
 import htsjdk.samtools.util.SequenceUtil;
 import htsjdk.samtools.util.StringUtil;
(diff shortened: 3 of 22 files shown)


lbergelson:
@cmnbroad Could you take a look at this when you get a chance?  I noticed this as problematic in JP's branch and we had a bug from this another time so I figure we should just make it easy to pass the nulls through.

src/main/java/htsjdk/samtools/SamInputResource.java
```diff
@@ -255,7 +253,7 @@ public File asFile() {
 
     @Override
     public Path asPath() {
-        return fileResource.toPath();
+        return IOUtil.toPath(fileResource);
```
pshapiro4broad: This code makes unchecked references to `fileResource` elsewhere; why is it expected that it's OK for it to be null here?

src/main/java/htsjdk/samtools/SamFiles.java
```diff
@@ -24,7 +25,7 @@
      * @return The index for the provided SAM, or null if one was not found.
      */
     public static File findIndex(final File samFile) {
-        Path path = findIndex(samFile.toPath());
+        final Path path = findIndex(IOUtil.toPath(samFile));
```
pshapiro4broad: It doesn't look like this will change the program behavior. This will cause `null` to be passed to `findIndex(Path)`, which will call `lookForIndex(Path)`, which will generate a NPE on line 69 below when it calls `samPath.getFileName()`.

If you do want to support `null` here, it would be easier to check `if (samFile == null)` directly IMO.

src/main/java/htsjdk/samtools/cram/ref/ReferenceSource.java
```diff
@@ -63,7 +62,7 @@ private ReferenceSource() {
     }
 
     public ReferenceSource(final File file) {
-        this(file == null ? null : file.toPath());
+        this(IOUtil.toPath(file));
```
pshapiro4broad: IMO the original code was clearer, since `null` is a directly supported case in the constructor.

src/main/java/htsjdk/samtools/cram/ref/ReferenceSource.java
```diff
@@ -63,7 +62,7 @@ private ReferenceSource() {
...
     }
 
     public ReferenceSource(final Path path) {
```
pshapiro4broad: With a minor change to this constructor, you can make the field `rsFile` `final`, which would match its current usage in the code

src/main/java/htsjdk/samtools/reference/ReferenceSequenceFileFactory.java
```diff
@@ -81,7 +81,7 @@ public static ReferenceSequenceFile getReferenceSequenceFile(final File file, fi
      * @param preferIndexed if true attempt to return an indexed reader that supports non-linear traversal, else return the non-indexed reader
      */
     public static ReferenceSequenceFile getReferenceSequenceFile(final File file, final boolean truncateNamesAtWhitespace, final boolean preferIndexed) {
-        return getReferenceSequenceFile(file.toPath(), truncateNamesAtWhitespace, preferIndexed);
+        return getReferenceSequenceFile(IOUtil.toPath(file), truncateNamesAtWhitespace, preferIndexed);
```
pshapiro4broad: Like the changes above, the changes in this file won't change the code behavior as an NPE will still be generated, just later in the call chain.

lbergelson:
@pshapiro4broad So I don't think I explained my motivation here very well.  As part of our work to add support for `Path` we've taken a lot of methods and added new versions that take paths:

ex:
```
void doThing(File file){
   ...
}

```
goes to:
```
void doThing(File file){
  doThing(file == null ? null : file.toPath());
}

void doThing(Path path){
   ...
}
```

During this we've also introduced a lot of potential NPE's where people assumed file was non-null and ported it as:
```
void doThings(File file){
  doThing(file.toPath);
}
```

It's not always trivial to do the analysis to see if the old method could accept nulls, and our documentation is pretty shoddy in that regard, so I think it's safer if we port these methods in a way that will maintain the original behavior when given a null, rather than introduce new potential NPE's during a port.  That's why I think this method is useful.

Now, to be fair, I was pretty lazy here and just did a find replace on uses of file.toPath() since I figured it was harmless, so I may have introduced it in pointless places.

src/main/java/htsjdk/samtools/cram/ref/ReferenceSource.java
lbergelson: Good point, but I opened a simple pr here:  https://github.com/samtools/htsjdk/pull/970

yfarjoun:
bump

yfarjoun:
bump again!

Issue #934, opened 2017-07-14
Performance issues with VCF header parsing


SquidDev:
When parsing VCF files with a large number of `##contig` headers (32k in this case), a significant amount of time is spent processing them (about 60 seconds on my machine).

I've done a small amount of profiling with VisualVM, with the result ending up looking something like this:
![image](https://user-images.githubusercontent.com/4346137/28231798-45d3d46c-68e5-11e7-9ef8-1d43f8e806f8.png)

Running `PrintVariantsExample` on a such a file results in [this log](https://gist.githubusercontent.com/SquidDev/c1a17641322a97b11a1fd82a97aed633/raw/5001506f1f20579862d625bb691b0549a1cedc79/htsjdk.variant.example.PrintVariantsExample.log).

### Your environment
* version of htsjdk: 2.10.1
* version of java: Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
* which OS: Windows 10

### Steps to reproduce
This can be tested by running `htsjdk.variant.example.PrintVariantsExample` with [this](https://gist.githubusercontent.com/SquidDev/74521af2a5e32f3b48d00cf75c22b1a1/raw/8862e99d8da90196dd89bc0f7f0ea058e78ea1dc/Sample01.vcf) VCF file.

cmnbroad:
This issue will be fixed by https://github.com/samtools/htsjdk/pull/835. The current VCFHeader implementation checks for duplicates every time a new contig line is encountered by doing a string compare with every existing contig line, which scales quadratically. The new implementation replaces this with a hash probe. On my machine, running PrintVariantsExample on the attached .vcf with 32k header lines takes around 30 seconds; on the branch it takes .6 seconds.

cmnbroad:
Fixed via https://github.com/samtools/htsjdk/pull/968.

PR #968, opened 2017-08-16
Fix performance issue when creating a VCF header with a large number of contigs

cmnbroad:
### Description

Interim fix for https://github.com/samtools/htsjdk/issues/934. https://github.com/samtools/htsjdk/pull/835 has a more permanent fix but is a much bigger change.

Whenever a new contig header line is added to a VCFHeader (i.e. during file parsing), the implementation looks for duplicates by comparing the new id with every contig header line already in the header, which is very slow on large headers. Replaced the brute force search with a hash lookup.

PrintVariantsExample running time for the test case provided in https://github.com/samtools/htsjdk/issues/934 (which is a 32k line, header-only file) goes from 30 seconds to .5 seconds.

### Checklist

- [X] Code compiles correctly
- [X] All tests passing

### DIFF
--- a/src/main/java/htsjdk/variant/vcf/VCFHeader.java
+++ b/src/main/java/htsjdk/variant/vcf/VCFHeader.java
@@ -68,7 +68,7 @@ public class VCFHeader implements Serializable {
     private final Map<String, VCFFormatHeaderLine> mFormatMetaData = new LinkedHashMap<String, VCFFormatHeaderLine>();
     private final Map<String, VCFFilterHeaderLine> mFilterMetaData = new LinkedHashMap<String, VCFFilterHeaderLine>();
     private final Map<String, VCFHeaderLine> mOtherMetaData = new LinkedHashMap<String, VCFHeaderLine>();
-    private final List<VCFContigHeaderLine> contigMetaData = new ArrayList<VCFContigHeaderLine>();
+    private final Map<String, VCFContigHeaderLine> contigMetaData = new LinkedHashMap<>();
 
     // the list of auxillary tags
     private final List<String> mGenotypeSampleNames = new ArrayList<String>();
@@ -188,7 +188,8 @@ public class VCFHeader implements Serializable {
      * @return all of the VCF header lines of the ##contig form in order, or an empty list if none were present
      */
     public List<VCFContigHeaderLine> getContigLines() {
-        return Collections.unmodifiableList(contigMetaData);
+        // this must preserve input order
+        return Collections.unmodifiableList(new ArrayList<>(contigMetaData.values()));
     }
 
     /**
@@ -223,10 +224,8 @@ public class VCFHeader implements Serializable {
         }
         mMetaData.removeAll(toRemove);
         for (final SAMSequenceRecord record : dictionary.getSequences()) {
-            contigMetaData.add(new VCFContigHeaderLine(record, record.getAssembly()));
+            addMetaDataLine(new VCFContigHeaderLine(record, record.getAssembly()));
         }
-
-        this.mMetaData.addAll(contigMetaData);
     }
 
     public VariantContextComparator getVCFRecordComparator() {
@@ -321,18 +320,15 @@ public class VCFHeader implements Serializable {
      * @return true if line was added to the list of contig lines, otherwise false
      */
     private boolean addContigMetaDataLineLookupEntry(final VCFContigHeaderLine line) {
-        for (VCFContigHeaderLine vcfContigHeaderLine : contigMetaData) {
-            // if we are trying to add a contig for the same ID
-            if (vcfContigHeaderLine.getID().equals(line.getID())) {
-                if ( GeneralUtils.DEBUG_MODE_ENABLED ) {
-                    System.err.println("Found duplicate VCF contig header lines for " + line.getID() + "; keeping the first only" );
-                }
-                // do not add this contig if it exists
-                return false;
+        // if we are trying to add a contig for the same ID
+        if (contigMetaData.containsKey(line.getID())) {
+            if ( GeneralUtils.DEBUG_MODE_ENABLED ) {
+                System.err.println("Found duplicate VCF contig header lines for " + line.getID() + "; keeping the first only" );
             }
+            // do not add this contig if it exists
+            return false;
         }
-
-        contigMetaData.add(line);
+        contigMetaData.put(line.getID(), line);
         return true;
     }
 
--- a/src/test/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
+++ b/src/test/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
@@ -223,7 +223,8 @@ public class VCFHeaderUnitTest extends VariantBaseTest {
     @Test
     public void testVCFHeaderAddContigLine() {
         final VCFHeader header = getHiSeqVCFHeader();
-        final VCFContigHeaderLine contigLine = new VCFContigHeaderLine("<ID=chr1,length=1234567890,assembly=FAKE,md5=f126cdf8a6e0c7f379d618ff66beb2da,species=\"Homo sapiens\">", VCFHeaderVersion.VCF4_0, "chr1", 0);
+        final VCFContigHeaderLine contigLine = new VCFContigHeaderLine(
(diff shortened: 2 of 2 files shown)


src/main/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -188,7 +188,10 @@ public void addMetaDataLine(final VCFHeaderLine headerLine) {
      * @return all of the VCF header lines of the ##contig form in order, or an empty list if none were present
      */
     public List<VCFContigHeaderLine> getContigLines() {
-        return Collections.unmodifiableList(contigMetaData);
+        List<VCFContigHeaderLine> contigLines = new ArrayList<>(contigMetaData.size());
+        // this must preserve input order
+        contigLines.addAll(contigMetaData.values());
+        return Collections.unmodifiableList(contigLines);
```
pshapiro4broad: If you make this method return `Collection<>` you could avoid the copying and return `Collections.unmodifiableCollection(contigMetadata.values())`.

If not, you can avoid the two-step `new ...` `addAll()` by passing `contigMetaData.values()` to the `ArrayList` constructor.

In general, when writing APIs it's better to return `Collection` when possible so the API isn't tied to an implementation.

src/main/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -68,7 +68,7 @@
     private final Map<String, VCFFormatHeaderLine> mFormatMetaData = new LinkedHashMap<String, VCFFormatHeaderLine>();
     private final Map<String, VCFFilterHeaderLine> mFilterMetaData = new LinkedHashMap<String, VCFFilterHeaderLine>();
     private final Map<String, VCFHeaderLine> mOtherMetaData = new LinkedHashMap<String, VCFHeaderLine>();
-    private final List<VCFContigHeaderLine> contigMetaData = new ArrayList<VCFContigHeaderLine>();
+    private final Map<String, VCFContigHeaderLine> contigMetaData = new LinkedHashMap<>();
```
lbergelson: We should probably the type here to a LinkedHashMap since we definitely do care about the ordering.  Honestly, these should all probably be declared as LinkedHashMaps. It's gross that there isn't an interface type for this sort of sorted Map, but I think it's better to put the type as what we actually require than pretending we can substitute any map.

src/main/java/htsjdk/variant/vcf/VCFHeader.java
lbergelson: I think we need to provide an ordered collection here, so list makes sense.  An ordered set would probably be better, but that's awkard in java and the contract is already that it returns a list.

src/main/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -321,18 +322,15 @@ private boolean addMetadataLineLookupEntry(final VCFHeaderLine line) {
...
-            if (vcfContigHeaderLine.getID().equals(line.getID())) {
-                if ( GeneralUtils.DEBUG_MODE_ENABLED ) {
-                    System.err.println("Found duplicate VCF contig header lines for " + line.getID() + "; keeping the first only" );
-                }
-                // do not add this contig if it exists
-                return false;
+        // if we are trying to add a contig for the same ID
+        if (contigMetaData.containsKey(line.getID())) {
+            if ( GeneralUtils.DEBUG_MODE_ENABLED ) {
+                System.err.println("Found duplicate VCF contig header lines for " + line.getID() + "; keeping the first only" );
```
lbergelson: lets switch to using some sort of logger while we're here.

src/main/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -223,10 +226,8 @@ public void setSequenceDictionary(final SAMSequenceDictionary dictionary) {
         }
         mMetaData.removeAll(toRemove);
```
lbergelson: @cmnbroad This method also iterates over all the headerlines, as do some others, do you think we should clean them up now or wait for your big change?

src/test/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
```diff
@@ -236,6 +237,36 @@ public void testVCFHeaderAddContigLine() {
...
+            final List<VCFContigHeaderLine> orderedList = new ArrayList<>();
+            final int splitInTheMiddle = originalHeaderList.size() / 2;
+            orderedList.addAll(originalHeaderList.subList(0, splitInTheMiddle));
+            final VCFContigHeaderLine outrageousContigLine = new VCFContigHeaderLine(
+                    "<ID=outrageousID,length=1234567890,assembly=FAKE,md5=f126cdf8a6e0c7f379d618ff66beb2da,species=\"Homo sapiens\">",
+                    VCFHeaderVersion.VCF4_2,
+                    VCFHeader.CONTIG_KEY,
+                    0);
+            orderedList.add(outrageousContigLine);
+            // make sure the extrs contig line is outrageous enough to not collide with a real key
```
lbergelson: typo: extr**s**

src/main/java/htsjdk/variant/vcf/VCFHeader.java
lbergelson: Good idea to use the constructor that takes a collection here like @pshapiro4broad suggests though I think.

pshapiro4broad: Changing the return type has no effect on the underlying storage, so in my mind making the declared type more general is perfectly OK. The javadoc for the method is where I'd put the information about the method result being ordered, since that is a caller concern (whereas the underlying type, `List`, is not). But either way would work.

src/main/java/htsjdk/variant/vcf/VCFHeader.java
pshapiro4broad: I would also disagree about this, for the same reason; the ordering is a data storage requirement but not an API restriction. If in the future we switched to a different implementation of an ordered map then we'd need to update our code unnecessarily.

src/main/java/htsjdk/variant/vcf/VCFHeader.java
lbergelson: I don't agree with you here.  The return type is a List because we guarantee the return value is ordered.  In general we want to make outputs narrowly typed and inputs broadly typed since that makes our functions more useful.  There's a competing goal of hiding the implementation details so we're free to change them.  In general we probably want to return the narrowest abstract type that provides the guarantees we need for our method to function.  Enforcing conventions in javadoc is a much weaker and more brittle way to do things.

src/main/java/htsjdk/variant/vcf/VCFHeader.java
cmnbroad: There are several of these in this file and in related files - I'd definitely prefer to minimize the changes here and leave these to other PR.

src/main/java/htsjdk/variant/vcf/VCFHeader.java
cmnbroad: Definitely prefer to wait; these are all replaced in the other PR.

src/test/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
cmnbroad: Done.

src/main/java/htsjdk/variant/vcf/VCFHeader.java
lbergelson: In general I agree with you about this, but using a LinkedHashMap here is actually an API restriction.  If we change to using a different Map that doesn't provide the ordering properties of LinkedHashMap then this class will no longer work.  It would be ideal if java provided an InsertionOrderedMap interface, but since it doesn't it's probably best to specify the LinkedHashMap here even though it's bad practice in general to declare a specific type.

cmnbroad:
@lbergelson Made a couple of the requested changes, but I'd definitely prefer to wait for these other changes in the other PR (especially since most of them are already done there), and keep this PR minimal. Feel free to push back.

lbergelson:
@cmnbroad  👍 merge when ready

PR #970, opened 2017-08-18
cleaning up constructors in ReferenceSource

lbergelson:
making field `rsFile` final and removing the private no-arg constructor
making the other constructors chain

- [ ] Code compiles correctly
- [ ] All tests passing

@pshapiro4broad  Please review since you pointed this change out :)


lbergelson:
@cmnbroad Yeah, this is a very minor fix, but Phil pointed it out during a different code review and I figured I'd just make the change since it would only take a minute.

PR #974, opened 2017-08-25
Simplify and deprecate string line StringLineReader

magicDGS:
### Description

* First commit from #953
* Simplify `StringLineReader` to use `BufferedLineReader` as described in #953 
* Deprecate `StringLineReader`  in favor of a `BufferedLineReader.fromString()` method.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


magicDGS:
Here it is the simplification of `StringLineReader` as promised, @pshapiro4broad

src/main/java/htsjdk/samtools/util/StringLineReader.java
```diff
@@ -23,81 +23,18 @@
...
 
-import java.util.regex.Pattern;
+import java.io.ByteArrayInputStream;
 
 /**
  * Implementation of LineReader that gets its input from a String.  No charset conversion
  * is necessary because the String is in unicode.  Handles CR, LF or CRLF line termination,
  * but if asked to return the line terminator, it always comes back as LF.
+ *
+ * @deprecated use
```
pshapiro4broad: need to add `{@link}`

src/test/java/htsjdk/samtools/util/StringLineReaderTest.java
```diff
@@ -27,6 +27,8 @@
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
+/** @deprecated testing a deprecated class */
```
pshapiro4broad: I would delete this test instead of deprecating it, as there is no interesting code left to test.

src/main/java/htsjdk/samtools/util/StringLineReader.java
magicDGS: Opps...Fixed.

src/test/java/htsjdk/samtools/util/StringLineReaderTest.java
magicDGS: True. Removed.

src/main/java/htsjdk/samtools/util/BufferedLineReader.java
```diff
@@ -26,32 +26,36 @@
...
     public BufferedLineReader(final InputStream is, final int bufferSize) {
-        reader = new BufferedReader(new InputStreamReader(is, Charset.forName("UTF-8")), bufferSize);
+        super(new InputStreamReader(is, Charset.forName("UTF-8")), bufferSize);
+    }
+
+    /**
+     * Returns a {@link BufferedLineReader} that gets its input from a String. No charset conversion
+     * is necessary because the String is in unicode.
+     */
+    public static BufferedLineReader fromString(final String s) {
```
lbergelson: Why not make this another constructor instead of a static method?  It seems like it would fit more naturally with the class if you did that.

magicDGS: For my point of view, a Reader is usually constructed from a file or stream, so it looks wierd to have a String constructor, which may be confused by a method like this:

```java
public BufferedLineReader(final String pathString) throws IOException {
    this(Files.newInputStream(Paths.get(pathString)));
}
```

If it is not a good reason, let me know to change it ASAP. I will rebase anyway to remove the redundant commit.

magicDGS:
Let me know if the refactor of the static method to a constructor is required - back to you, @lbergelson

src/main/java/htsjdk/samtools/util/BufferedLineReader.java
lbergelson: That's a good point about the path string.  I hadn't thought of that.

PR #977, opened 2017-09-06
Deprecate 'not-primary' in favor of secondary along HTSJDK

magicDGS:
### Description

The term "not-primary alignment" used in HTSJDK is confusing, because it refers only to "secondary alignment" but not "supplementary alignments". This commit deprecate classes/methdos using "not-primary" in favor of the "secondary" option (implemented if not already included).

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMRecord.java
+++ b/src/main/java/htsjdk/samtools/SAMRecord.java
@@ -967,20 +967,19 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
 
     /**
      * the alignment is not primary (a read having split hits may have multiple primary alignment records).
-     * @deprecated use {@link #getSecondaryAlignmentFlag()} instead.
+     * @deprecated use {@link #isSecondaryAlignment()} instead.
      */
     @Deprecated
     public boolean getNotPrimaryAlignmentFlag() {
-        return getSecondaryAlignmentFlag();
+        return isSecondaryAlignment();
     }
 
     /**
      * the alignment is secondary (a read having slipt hits have multiple alignment records).
      */
-    public boolean getSecondaryAlignmentFlag() {
+    public boolean isSecondaryAlignment() {
         return (mFlags & SAMFlag.SECONDARY_ALIGNMENT.flag) != 0;
     }
-
 
     /**
      * the alignment is supplementary (TODO: further explanation?).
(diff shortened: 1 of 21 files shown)


src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -967,11 +967,21 @@ private boolean getSecondOfPairFlagUnchecked() {
...
+    @Deprecated
     public boolean getNotPrimaryAlignmentFlag() {
-        return (mFlags & SAMFlag.NOT_PRIMARY_ALIGNMENT.flag) != 0;
+        return getSecondaryAlignmentFlag();
+    }
+
+    /**
+     * the alignment is secondary (a read having slipt hits have multiple alignment records).
+     */
+    public boolean getSecondaryAlignmentFlag() {
```
lindenb: time to switch to 

`isSecondaryAlignmentFlag() `

?

magicDGS: I would say yes, but there is another PR for discussing that: https://github.com/samtools/htsjdk/pull/868

tfenne: Or even `isSecondaryAlignment()`

src/main/java/htsjdk/samtools/SecondarySkippingIterator.java
```diff
@@ -0,0 +1,62 @@
...
+
+import htsjdk.samtools.util.CloseableIterator;
+import htsjdk.samtools.util.PeekIterator;
+
+/**
+ * Wrapper around SAMRecord iterator that skips over secondary elements.
+ * This iterator conflates a filtering iterator and a peekable iterator.  It would be cleaner to
+ * handle those concerns separately.
+ */
+public class SecondarySkippingIterator {
```
tfenne: I think this should be `SecondaryAlignmentSkippingIterator` to make it a little clearer.

magicDGS:
Oops, I always confound while writing "supplementary" and "secondary". I am sorry. I will change the commit message right now.

src/main/java/htsjdk/samtools/SAMRecord.java
magicDGS: I agree, but wouldn't be discordant with the rest of the class? I guess that in the PR that I mentioned it was decided to change everything together... but I added a commit changing that.

src/main/java/htsjdk/samtools/SecondarySkippingIterator.java
magicDGS: True, I just used the same form as `NotPrimarySkippingIterator`. Changed.

magicDGS:
Friendly ping here, @tfenne!

magicDGS:
Thanks @tfenne!

Issue #971, opened 2017-08-18
optimizeIntervals fails to merge abutting intervals


jean-philippe-martin:
### Subject of the issue

`QueryInterval.optimizeIntervals` does not merge adjacent intervals as I would have expected.

Here is test code to demonstrate the issue:

```
  @Test
  public void testOptimizeIntervals() throws Exception {
    QueryInterval[] intervals = new QueryInterval[] {
        new QueryInterval(0, 1519, 1520),
        new QueryInterval(0, 1521, 1525)
    };
    QueryInterval[] expected = new QueryInterval[] {
        new QueryInterval(0, 1519, 1525),
    };
    final QueryInterval[] got = QueryInterval.optimizeIntervals(intervals);
    Assert.assertEquals(expected, got);
  }
```

It fails because `got` still contains the two initial intervals, even though they touch.

### Your environment

I tried using htsjdk 2.10.1, java 8 and Linux.

### Steps to reproduce

Compile and run the test code above. Observe it fail.

### Expected behaviour

I expect abutting intervals to be merged.

### Actual behaviour

Overlapping intervals are merged, but abutting ones are not.

yfarjoun:
fixed by #979

PR #979, opened 2017-09-11
Optimize intervals falls to merge abutting intervals

Kirvolque:
### Description
Optimizing intervals in `QueryInterval` class works incorrectly: intervals are considered abutting only if they overlap each other by a single base.

### Prerequisites
Need bug fix for current implementation of abuts(): `QueryInterval[]` array still contains two initial intervals after invoking `optimizeIntervals()` method, even though they touch. [Github issue](https://github.com/samtools/htsjdk/issues/971).

### Solution
* `abuts()` method has been fixed. It used to return `true` only in case if one of them starts exactly where other ends. Now it returns `true` if one of the given intervals ends just before another one starts;
* `equals()` and `hashcode()` methods were overridden in `QueryInterval` class for using `assert` in TestNg;
* `testOptimizeIntervals()` was added to the `CRAMFileBAIIndexTest` class to test 3 options: abutting, overlapping and non-overlapping intervals.

### Checklist
- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/QueryInterval.java
+++ b/src/main/java/htsjdk/samtools/QueryInterval.java
@@ -44,7 +44,7 @@ public class QueryInterval implements Comparable<QueryInterval> {
     /**
      * @return true if both are on same reference, and other starts exactly before this ends.
      */
-    public boolean abuts(final QueryInterval other) {
+    public boolean endsAtStartOf(final QueryInterval other) {
         return this.referenceIndex == other.referenceIndex && this.end + 1 == other.start;
     }
 
@@ -81,7 +81,7 @@ public class QueryInterval implements Comparable<QueryInterval> {
 
         for (int i = 1; i < inputIntervals.length; ++i) {
             final QueryInterval next = inputIntervals[i];
-            if (previous.abuts(next) || previous.overlaps(next)) {
+            if (previous.endsAtStartOf(next) || previous.overlaps(next)) {
                 final int newEnd = ((previous.end == 0 || next.end == 0) ? 0 : Math.max(previous.end, next.end));
                 previous = new QueryInterval(previous.referenceIndex, previous.start, newEnd);
             } else {
@@ -102,9 +102,7 @@ public class QueryInterval implements Comparable<QueryInterval> {
 
         QueryInterval that = (QueryInterval) o;
 
-        if (referenceIndex != that.referenceIndex) return false;
-        if (start != that.start) return false;
-        return end == that.end;
+        return this.compareTo(that) == 0;
     }
 
     @Override
--- a/src/main/java/htsjdk/samtools/QueryInterval.java
+++ b/src/main/java/htsjdk/samtools/QueryInterval.java
@@ -42,10 +42,10 @@ public class QueryInterval implements Comparable<QueryInterval> {
     }
 
     /**
-     * @return true if both are on same reference, and other starts exactly where this ends.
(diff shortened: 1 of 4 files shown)


src/main/java/htsjdk/samtools/QueryInterval.java
```diff
@@ -42,10 +42,10 @@ public int compareTo(final QueryInterval other) {
     }
 
     /**
-     * @return true if both are on same reference, and other starts exactly where this ends.
+     * @return true if both are on same reference, and other starts exactly before this ends.
```
yfarjoun: I don't think we should be changing the definition of "abuts". since we use closed intervals the implementation was correct before the change.

yfarjoun:
👎  from me. 

I can understand why you would want Query Interval to work slightly differently (in particular this bit us in a recent GATK PR), but this is not the right way to go about it.

Kirvolque:
@yfarjoun, thank you for quick response. If offered changes are unnecessary, may we close this pull request?

src/main/java/htsjdk/samtools/QueryInterval.java
tfenne: @yfarjoun If `QueryInterval` really uses closed intervals, then I think @Kirvolque is right and the existing implementation was wrong.  E.g. two closed intervals `1-9` and `10-20` abut one another,  but the coordinates are indeed off by one.

That said, I think this method should be private since it doesn't test bi-directionally.  That's fine given how it's used in `optimizeOverlaps`, but it would be confusing if called by other code.

src/main/java/htsjdk/samtools/QueryInterval.java
```diff
@@ -94,4 +94,24 @@ public String toString() {
 
         return unique.toArray(EMPTY_QUERY_INTERVAL_ARRAY);
     }
+
+    @Override
+    public boolean equals(Object o) {
```
tfenne: In classes that implement `Comparable` it's preferable to implement `equals()` as `this.compareTo(that) == 0` after the initial type checks.

src/main/java/htsjdk/samtools/QueryInterval.java
yfarjoun: Ahh, I see. Sorry to judge so quick. The method is called from BAMFileReader, so it can't be changed to private. perhaps we just change it's name to something like `endAtStartOf` so that the code would read `prev.endsAtStartOf(nextInterval)`.

yfarjoun:
Please note that there are a few failures in the tests...testQueryOverlappingMultipleIntervals fails in  3 cases (probably because they were written using the wrong definition of "abutting" :-)

yfarjoun:
This looks better to me. @tfenne ?

src/main/java/htsjdk/samtools/QueryInterval.java
Kirvolque: @tfenne, done. Thank you for your note.

src/main/java/htsjdk/samtools/QueryInterval.java
Kirvolque: @yfarjoun, thank you. The method has been renamed.

tfenne:
LGTM.  I think the failed check is due to a random error contacting ENA in one of the tests.  I've restarted it.  Once it passes I'll merge.

PR #980, opened 2017-09-11
Add support for long reference sequences

Kirvolque:
### Description

BAI-index has length limit for reference sequence — 512Mbp.

BAM file has bin field (UShort — 2 bytes) that stores bin number for BAI-index. If reference sequence length exceeds 512Mbp, SAMValidationError occurs during validation of bin-field in ```SAMRecord::isValid```.

The similar situation occurs during writing SAMRecord in ```BAMRecordCodec::encode``` — UShort overflowing happens.

*This pull request is necessary for CSI-index implementation.*

### Implementation

- Validation of bin-field is ignored when reference sequence length is larger than 512Mbp for ```SAMRecord::isValid``` and ```BAMRecordCodec::encode```;
- Nevertheless, warning *about incorrect bin-field* is displayed in that case;
- In case reference sequence length is larger than 512Mbp ```BAMRecordCodec::encode``` writes only 16 last bits of calculated bin-index *(the same as SamTools does)*;
- Tests for reading and writing BAM files were refactored and updated for cases of long reference sequences.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


src/main/java/htsjdk/samtools/BAMRecordCodec.java
```diff
@@ -174,6 +184,15 @@ public void encode(final SAMRecord alignment) {
         }
     }
 
+    private int getUshort(int value) {
+        return value & (int) BinaryCodec.MAX_USHORT;
+    }
+
+    private boolean isReferenceTooLargeForBAI() {
+        return header.getSequenceDictionary().getReferenceLength() > GenomicIndexUtil.BIN_GENOMIC_SPAN
+                && !isReferenceSizeWarningShowed;
```
pshapiro4broad: This works but it would be clearer to test `isReferenceSizeWarningShowed` at the same `if` block that sets it, instead of inside here. Also you should test the boolean first, if it's `true` then it doesn't matter if the reference length is too long or not.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2058,13 +2061,18 @@ public int hashCode() {
...
-                    "bin field of BAM record does not equal value computed based on alignment start and end, and length of sequence to which read is aligned",
-                    getReadName()));
-            if (firstOnly) return ret;
+        if (this.getAlignmentStart() != NO_ALIGNMENT_START) {
+            if (getHeader() != null
+                    && isReferenceTooLargeForBAI()
+                    && getValidationStringency() != ValidationStringency.SILENT) {
+                LOG.warn(String.format("Reference length is too large for BAM bin field for %s record. Values in the bin field could be incorrect.", getReadName()));
+            } else if (isIndexingBinNotEqualsComputedBin()) {
+                if (ret == null) ret = new ArrayList<SAMValidationError>();
```
pshapiro4broad: According to codecov this code is not covered by a test, it should be.

Kirvolque:
@pshapiro4broad, Thank you for your notes. We added test for SAMRecord bin field validation.
Also we found that ```isReferenceLengthTooLargeForBAI()``` worked incorrectly: it didn't compare a reference length but a sequence dictionary length. We have fixed it.

Kirvolque:
@pshapiro4broad, Thank you for your notes. We added test for SAMRecord bin field validation.
Also we found that ```isReferenceLengthTooLargeForBAI()``` worked incorrectly: it didn't compare a reference length but a sequence dictionary length. We have fixed it.

src/main/java/htsjdk/samtools/BAMRecordCodec.java
Kirvolque: Done.

src/main/java/htsjdk/samtools/SAMRecord.java
Kirvolque: Done.

Kirvolque:
@pshapiro4broad so do you have any thoughts about this pull request? Coverage seems to be good now.

pshapiro4broad:
The code looks ok to me but it will need a repo maintainer to approve for merging.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2058,13 +2061,19 @@ public int hashCode() {
...
-            if (ret == null) ret = new ArrayList<>();
-            ret.add(new SAMValidationError(SAMValidationError.Type.INVALID_INDEXING_BIN,
-                    "bin field of BAM record does not equal value computed based on alignment start and end, and length of sequence to which read is aligned",
-                    getReadName()));
-            if (firstOnly) return ret;
+        if (this.getAlignmentStart() != NO_ALIGNMENT_START) {
+            final SAMSequenceRecord sequence = getHeader() != null ? getHeader().getSequence(getReferenceName()) : null;
+            if (sequence != null && isReferenceTooLargeForBAI(sequence)) {
+                if (getValidationStringency() != ValidationStringency.SILENT) {
+                    LOG.warn(String.format("Reference length is too large for BAM bin field for %s record. Values in the bin field could be incorrect.", getReadName()));
```
cmnbroad: "Values in the bin field could be incorrect" is pretty vague. If we're declaring that the right thing to do for this case is to use a narrowed value (the one the BAMRecordCodec now emits), then it seems like we should check for that value here, and use a different log message for the case where the value is the "expected" truncated value, vs. a completely incorrect value. Also I think it would be useful to include the reference contig name, and possibly even the bin values in the case where its completely incorrect.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2099,6 +2108,14 @@ public int hashCode() {
...
     }
 
+    private boolean isIndexingBinNotEqualsComputedBin() {
+        return this.getIndexingBin() != null && this.computeIndexingBin() != this.getIndexingBin();
+    }
+
+    private boolean isReferenceTooLargeForBAI(final SAMSequenceRecord sequence) {
+        return sequence.getSequenceLength() > GenomicIndexUtil.BIN_GENOMIC_SPAN;
+    }
+
```
cmnbroad: This method is duplicated here and in BAMRecordCodec. I think it would be useful to move it so it can be shared, maybe to GenomicUtils.

src/main/java/htsjdk/samtools/BAMRecordCodec.java
```diff
@@ -113,11 +118,13 @@ public void encode(final SAMRecord alignment) {
...
-        if (alignment.getReferenceIndex() >= 0) {
+        if (alignment.getAlignmentStart() != SAMRecord.NO_ALIGNMENT_START) {
+            warnIfReferenceIsTooLargeForBinField(alignment);
+
             if (alignment.getIndexingBin() != null) {
                 indexBin = alignment.getIndexingBin();
             } else {
-                indexBin = alignment.computeIndexingBin();
+                indexBin = getUshort(alignment.computeIndexingBin());
             }
```
cmnbroad: I think it would be better to move this whole if/else stmt (lines 124-128), and it's companion the getUshort method, into SAMRecord as a method that is colocated with computeIndexingBin.

src/main/java/htsjdk/samtools/BAMRecordCodec.java
```diff
@@ -174,6 +181,25 @@ public void encode(final SAMRecord alignment) {
...
+                && isReferenceTooLargeForBAI(sequence)
+                && rec.getValidationStringency() != ValidationStringency.SILENT) {
+            LOG.warn("Reference length is too large for BAM bin field. Values in the bin field could be incorrect.");
+            isReferenceSizeWarningShowed = true;
+        }
+    }
+
+    private boolean isReferenceTooLargeForBAI(final SAMSequenceRecord sequence) {
+        return sequence.getSequenceLength() > GenomicIndexUtil.BIN_GENOMIC_SPAN;
+    }
```
cmnbroad: This method appears in two places and should probably be consolidated. See comment below.

Kirvolque:
@cmnbroad, thank you for your notes.
The only thing has been changed is that ```INVALID_INDEXING_BIN``` exception isn't thrown in case of a long reference. These changes are essential for implementing CSI index in HTSJDK or any other long reference index.

src/main/java/htsjdk/samtools/SAMRecord.java
Kirvolque: Done.

src/main/java/htsjdk/samtools/SAMRecord.java
Kirvolque: We should have emphasize that in case reference length is larger than BAI index can handle, bin values are always incorrect. So there are no different cases.
We have clarified warning message, now it shows that all bin field values are incorrect.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -104,6 +105,8 @@
  * @author mishali.naik@intel.com
  */
 public class SAMRecord implements Cloneable, Locatable, Serializable {
+    private final static Log LOG = Log.getInstance(BAMRecordCodec.class);
```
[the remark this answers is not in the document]
Kirvolque: Done.

src/main/java/htsjdk/samtools/BAMRecordCodec.java
Kirvolque: Done.

src/main/java/htsjdk/samtools/BAMRecordCodec.java
Kirvolque: Done.

WimSpee:
Is there any news on this issue?

yfarjoun:
:party:

PR #982, opened 2017-09-12
deprecating SortingVariantContextWriter

lbergelson:
### Description

* deprecating the classes `SortingVariantContextWriter` and `SortingVariantContextWriterBase`
* these classes are completely untested and unused in htsjdk, picard, and gatk, we should either deprecate them or add tests and support them, but unless anyone has any use case I'd recommend deprecating

### Checklist

- [ x ] Is not backward compatible (breaks binary or source compatibility)


lbergelson:
@yfarjoun You originally brought up the unusedness of these.  
@tfenne @jacarey Any objections?  They may be good classes but they're totally untested and I didn't find any uses of them anywhere I looked.

tfenne:
image

jacarey:
no objections

lbergelson:
That was easy!

PR #983, opened 2017-09-12
minor changes to snappy

lbergelson:
### Description
There were a few changes I wanted to make to the previous snappy pull request but it seemed better to get the main part of that merged in and then deal with them rather than sit in limbo forever.

There are a 2 commits, hopefully the first is uncontroversial:

 1st Commit: 
  * move the system properties to Defaults to be consistent with the rest of the htsjdk system properties
 * produce a more specific error message when failing to wrap a stream with snappy, and refactor to avoid duplicating the error message logic, the refactoring here might be overkill, but it seemed slightly cleaner to me than just duplicating the code, happy to roll that bit back though if it's too java8 for people's taste
* adding a few finals and try-with-resources to taste

 2nd commit:
 * removes the newly added verbosity argument and replaces it by making the log levels better tuned.  This seems to me like a reasonable tradeoff but @tfenne you were the one who wanted better control of the logging messages, so I'm not sure if it does what you need.  I think your initial logging changes may have gotten clobbered by the changes, so i'm not sure it's doing what you want anyway right now. 

### Checklist
- [X] Is not backward compatible (breaks binary or source compatibility)  this is not strictly compatible with the previous snappy pr, but that's never been released, so assuming we can get this in before a release it's not a breaking change

### DIFF
--- a/src/main/java/htsjdk/samtools/util/SnappyLoader.java
+++ b/src/main/java/htsjdk/samtools/util/SnappyLoader.java
@@ -44,7 +44,11 @@ public class SnappyLoader {
     private final boolean snappyAvailable;
 
     public SnappyLoader() {
-        if (Defaults.DISABLE_SNAPPY_COMPRESSOR) {
+        this(Defaults.DISABLE_SNAPPY_COMPRESSOR);
+    }
+
+    public SnappyLoader(boolean disableSnappyCompressor) {
+        if (disableSnappyCompressor) {
             logger.debug("Snappy is disabled via system property.");
             snappyAvailable = false;
         }
--- a/src/main/java/htsjdk/samtools/util/SnappyLoader.java
+++ b/src/main/java/htsjdk/samtools/util/SnappyLoader.java
@@ -23,6 +23,7 @@
  */
 package htsjdk.samtools.util;
 
+import htsjdk.samtools.Defaults;
 import htsjdk.samtools.SAMException;
 import org.xerial.snappy.SnappyError;
 import org.xerial.snappy.SnappyInputStream;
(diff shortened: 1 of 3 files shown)


lbergelson:
@tfenne Would you mind taking a look at this when you get a chance?

src/main/java/htsjdk/samtools/util/SnappyLoader.java
```diff
@@ -38,73 +39,71 @@
...
-    /** Wrap an InputStream in a SnappyInputStream. If Snappy is not available will throw an exception. */
+    /**
+     * Wrap an OutputStream in a SnappyOutputStream.
+     * @throws SAMException if Snappy is not available
+     */
     public OutputStream wrapOutputStream(final OutputStream outputStream) {
+        return wrapWithSnappyOrThrow(outputStream, (stream) -> new SnappyOutputStream(stream, SNAPPY_BLOCK_SIZE));
+    }
+
+    private interface IOFunction<T,R> {
```
tfenne: I'm not really up on the Java 8 lambda stuff, but is this interface needed?  Can't you just use `java.util.Function<T,R>' in it's place?

src/main/java/htsjdk/samtools/util/SnappyLoader.java
```diff
@@ -38,73 +39,71 @@
...
+    }
+
+    private <T,R> R wrapWithSnappyOrThrow(T stream, IOFunction<T, R> wrapper){
         if (isSnappyAvailable()) {
             try {
-                return new SnappyOutputStream(outputStream, SNAPPY_BLOCK_SIZE);
+                return wrapper.apply(stream);
             } catch (Exception e) {
-                throw new SAMException("Error instantiating SnappyOutputStream", e);
+                throw new SAMException("Error instantiating SnappyInputStream", e);
```
tfenne: This should probably just read `Error wrapping stream with snappy` or something similarly generic.

src/main/java/htsjdk/samtools/util/SnappyLoader.java
lbergelson: sadly you can't, the default types don't throw checked exceptions, so you can't capture something that throws IOException as a standard `Function` there doesn't yet seem to a be standard set of "throwing interfaces" anywhere yet either.

src/main/java/htsjdk/samtools/util/SnappyLoader.java
lbergelson: oh, good point, will change.

lbergelson:
@tfenne Thanks for the quick review.  I added a few simple tests too while I was at it.

PR #985, opened 2017-09-15
Discussion: implement Read interface

magicDGS:
### Description
Create common Read interface to move to a interface-based API in htsjdk.
The new Read interface is packed in  `htsjdk.samtools.read` to add more
structure to the API. The interface is based in the current SAMRecord
and in the GATK's read interface, keeping some ideas from both.

In addition to the new interface, some more classes were added:

* ReadConstants: include constants for the Read interface. Correspond to some constants in the SAMRecord
* ReadAttribute: another API interface to allow different representations of the read attributes.
* ReadUtils: utility class for working with the read interface in a implementation-independent way. Include few examples of no-API methods for getting un-clipped start/end.

Once the interface is stable, I propose to implement a `SAMRecord`class in the `htsjdk.samtools.read` to substitute the `htsjdk.samtools.SAMRecord`, which would be a deprecated extension of that one. This proposal may help to keep the classes better organized in the htsjdk library and move towards the new interface-based API in the library deprecating the old code.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/read/Read.java
+++ b/src/main/java/htsjdk/samtools/read/Read.java
@@ -233,7 +233,7 @@ public interface Read extends Locatable {
     String getContig();
 
     /**
-     * Gets the 1-based inclusive leftmost position of the sequence remaining after clipping.
+     * Gets the 1-based leftmost mapping POSition of the first CIGAR operation that “consumes” a reference base.
      *
      * <p>Equivalent to <b>POS</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
      *
@@ -243,7 +243,7 @@ public interface Read extends Locatable {
     int getStart();
 
     /**
-     * Gets the 1-based inclusive rightmost position of the sequence remaining after clipping.
+     * Gets the 1-based inclusive rightmost mapping POSition of the first CIGAR operation that “consumes” a reference base.
      *
      * @return 1-based closed-ended position; {@link ReadConstants#NO_ALIGNMENT_START} if there is no position (e.g. for unmapped read).
      */
@@ -285,7 +285,7 @@ public interface Read extends Locatable {
      *
      * <p>Note: this is not necessarily the same as the number of reference bases the read is aligned to.
      *
-     * @return The number of bases in the read
+     * @return The number of bases in the read-sequence.
      */
     int getLength();
 
@@ -301,7 +301,7 @@ public interface Read extends Locatable {
     }
 
     /**
-     * Gets the PHERD scaled mapping quality.
+     * Gets the phred scaled mapping quality.
      *
      * <p>The {@link ReadConstants#UNKNOWN_MAPPING_QUALITY} implies valid mapping, but hard to compute quality.
      *
@@ -312,7 +312,7 @@ public interface Read extends Locatable {
     int getMappingQuality();
 
     /**
-     * Sets the PHRED scaled mapping quality.
+     * Sets the phred scaled mapping quality.
      *
      * <p>Equivalent to <b>MAPQ</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
      *
@@ -323,6 +323,8 @@ public interface Read extends Locatable {
     /**
      * Gets the {@link Cigar} object describing how the read aligns to the reference.
      *
+     * <p>Equivalent to <b>CIGAR</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
      * <p>This method should make a defensive copy of the {@link Cigar} before returning it to allow
      * modification of the returned {@link Cigar} without effects on the {@link Read}.
      *
@@ -332,6 +334,8 @@ public interface Read extends Locatable {
 
     /**
      * Gets the number of cigar elements (number + operator) in the cigar string.
+     *
+     * <p>Equivalent to <b>CIGAR</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
      *
      * <p>Default implementation returns {@code getCigar().numCigarElements()}.
      * Subclasses may override to provide more efficient implementations.
@@ -344,6 +348,8 @@ public interface Read extends Locatable {
 
     /**
      * Sets the {@link Cigar} object describing how the read aligns to the reference.
+     *
+     * <p>Equivalent to <b>CIGAR</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
      *
      * @param cigar Cigar object for the read; empty Cigar if there is none.
      *
@@ -397,27 +403,33 @@ public interface Read extends Locatable {
     }
 
     /**
-     * Gets the insert size (difference between the 5' end of the read and the 5' end of the mate).
-     * The insert size is negative if the mate maps to lower position than the read.
+     * Gets the signed observed template length. If  all  segments  are  mapped  to  the  same  reference,
+     * the unsigned observed template length equals the number of bases from the leftmost mapped base to the
+     * rightmost mapped base.  The leftmost segment has a plus sign and the rightmost has a minus sign.
+     * The sign of segments in the middle is undefined. It is set as 0 for single-segment template or when the
+     * information is unavailable.
      *
      * <p>Equivalent to <b>TLEN</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
      *
-     * @return insert size if possible; {@code 0} otherwise.
+     * @return insert size if possible to compute; {@code 0} otherwise.
      */
     int getInferredInsertSize();
 
     /**
-     * Sets the insert size (difference between the 5' end of the read and the 5' end of the mate).
-     * The insert size is negative if the mate maps to lower position than the read.
+     * Sets the signed observed template length. If  all  segments  are  mapped  to  the  same  reference,
+     * the unsigned observed template length equals the number of bases from the leftmost mapped base to the
+     * rightmost mapped base.  The leftmost segment has a plus sign and the rightmost has a minus sign.
+     * The sign of segments in the middle is undefined. It is set as 0 for single-segment template or when the
+     * information is unavailable.
      *
      * <p>Equivalent to <b>TLEN</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
      *
-     * @param insertSize insert size (negative if mate mapts to lower position than the read) if possible; {@code 0} otherwise.
+     * @param insertSize insert size if possible to compute; {@code 0} otherwise.
      */
     void setInsertSize(final int insertSize);
 
     /**
-     * Gets the read sequence as ASCII bytes ACGTN=
+     * Gets the read sequence as ASCII bytes.
      *
      * <p>Equivalent to <b>SEQ</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
      *
@@ -436,7 +448,7 @@ public interface Read extends Locatable {
      * <p>Default implementation returns {@code getBases()[i]}.
      * Subclasses may override to provide a more efficient implementation.
      *
-     * @return base at index i.
+     * @return base at index i (0-based).
      *
      * @throws IndexOutOfBoundsException if i is negative or of i is not smaller than the number
      * of bases (as reported by {@link #getLength()}. In particular, if no sequence is present.
@@ -446,7 +458,7 @@ public interface Read extends Locatable {
     }
 
     /**
-     * Gets all the bases in the read as a String of ACGTN=
+     * Gets all the bases in the read as an ASCII String.
      *
      * <p>Equivalent to <b>SEQ</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
      *
@@ -464,14 +476,14 @@ public interface Read extends Locatable {
      *
      * <p>Equivalent to <b>SEQ</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
      *
-     * @param bases read sequence as ASCII bytes ACGTN=; {@link ReadConstants#NULL_SEQUENCE} if no sequence is present.
+     * @param bases read sequence as ASCII bytes; {@link ReadConstants#NULL_SEQUENCE} if no sequence is present.
      *
      * @throws IllegalArgumentException if the bases are null.
      */
     void setBases(final byte[] bases);
 
     /**
-     * Gets the base qualities, as binary PHRED scores (not ASCII).
+     * Gets the base qualities, as binary phred scores (not ASCII).
      *
      * <p>Equivalent to <b>QUAL</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
      *
@@ -513,7 +525,7 @@ public interface Read extends Locatable {
      * <p>Default implementation returns {@code getBaseQualities()[i]}.
      * Subclasses may override to provide a more efficient implementation.
      *
-     * @return The base quality at index i.
+     * @return The base quality at index i (0-based).
      *
      * @throws IndexOutOfBoundsException if i is negative or of i is not smaller than the number
      * of base qualities (as reported by {@link #getBaseQualityLength()}.
@@ -549,7 +561,7 @@ public interface Read extends Locatable {
     ///////////////////////////////////////////////////////////
 
     /**
-     * Gets the list of optional fields for the read. This are encoded as tag/value pairs, where
+     * Gets the list of optional fields for the read. These are encoded as tag/value pairs, where
      * the tag is a 2-character String and the value is from a concrete type.
      *
      * @return immutable list of attributes.
@@ -563,7 +575,7 @@ public interface Read extends Locatable {
      * and returns the first one matching the tag.
      * Subclasses may override to provide more efficient implementations.
      *
-     * @param tag attribute tag.
+     * @param tag attribute tag (two character string).
      *
      * @return optional value associated with the tag; may be empty if the attribute is not present.
      *
@@ -583,7 +595,7 @@ public interface Read extends Locatable {
      *
      * <p>Default implementation returns {@code getAttribute(tag).isPresent()}.
      *
-     * @param tag attribute tag.
+     * @param tag attribute tag (two character string).
      * @return {@code true} if the attribute was set; {@code false} otherwise.
      *
      * @throws SAMException if the tag is invalid following the contract of {@link ReadAttribute#isValidTag(String)}.
@@ -597,7 +609,7 @@ public interface Read extends Locatable {
      *
      * <p>Note: it is preferable to use setters for typed objects.
      *
-     * @param tag attribute tag.
+     * @param tag attribute tag (two character string).
      * @param value object value.
      *
      * @throws SAMException if the tag is invalid following the contract of {@link ReadAttribute#isValidTag(String)}.
@@ -613,7 +625,7 @@ public interface Read extends Locatable {
     /**
      * Clear an individual attribute on the read.
      *
-     * @param tag the attribute tag.
+     * @param tag attribute tag (two character string).
      *
      * @throws SAMException if the tag is invalid following the contract of {@link ReadAttribute#isValidTag(String)}.
      */
@@ -691,6 +703,7 @@ public interface Read extends Locatable {
      *
      * @return {@code true} if this read is unmapped; {@code false} otherwise.
      */
+    // TODO (before merging) - add method for get/set the reverse
     default boolean isUnmapped() {
         return isSet(SAMFlag.READ_UNMAPPED);
     }
@@ -704,6 +717,7 @@ public interface Read extends Locatable {
      *
      * @param unmapped {@code true} if this read is unmapped; {@code false} otherwise.
      */
+    // TODO (before merging) - add method for get/set the reverse
     default void setUnmapped(final boolean unmapped) {
         addFlag(SAMFlag.READ_UNMAPPED, unmapped);
     }
@@ -717,6 +731,7 @@ public interface Read extends Locatable {
      *
      * @return {@code true} if this read's mate is unmapped; {@code false} otherwise.
      */
+    // TODO (before merging) - add method for get/set the reverse
     default boolean isMateUnmapped() {
         return isSet(SAMFlag.MATE_UNMAPPED);
     }
@@ -730,6 +745,7 @@ public interface Read extends Locatable {
      *
      * @param mateUnmapped {@code true} if this read's mate is unmapped; {@code false} otherwise.
      */
+    // TODO (before merging) - add method for get/set the reverse
     default void setMateUnmapped(final boolean mateUnmapped) {
         addFlag(SAMFlag.MATE_UNMAPPED, mateUnmapped);
     }
@@ -743,6 +759,7 @@ public interface Read extends Locatable {
      *
      * @return {@code true} if the read is in the reverse strand; {@code false} otherwise.
      */
+    // TODO (before merging) - add method for get/set the reverse
     default boolean isReverseStrand() {
         return isSet(SAMFlag.READ_REVERSE_STRAND);
     }
@@ -756,6 +773,7 @@ public interface Read extends Locatable {
      *
      * @param reverseStrand {@code true} if the read is in the reverse strand; {@code false} otherwise.
      */
+    // TODO (before merging) - add method for get/set the reverse
     default void setReverseStrand(final boolean reverseStrand) {
         addFlag(SAMFlag.READ_REVERSE_STRAND, reverseStrand);
     }
@@ -769,6 +787,7 @@ public interface Read extends Locatable {
      *
      * @return {@code true} if the read's mate is in the reverse strand; {@code false} otherwise.
      */
+    // TODO (before merging) - add method for get/set the reverse
     default boolean isMateReverseStrand() {
         return isSet(SAMFlag.MATE_REVERSE_STRAND);
     }
@@ -782,6 +801,7 @@ public interface Read extends Locatable {
      *
      * @param mateReverseStrand {@code true} if the read's mate is in the reverse strand; {@code false} otherwise.
      */
+    // TODO (before merging) - add method for get/set the reverse
     default void setMateReverseStrand(final boolean mateReverseStrand) {
         addFlag(SAMFlag.MATE_REVERSE_STRAND, mateReverseStrand);
     }
@@ -873,6 +893,7 @@ public interface Read extends Locatable {
      *
      * @return {@code true} if the read fails the quality vendor check; {@code false} otherwise.
      */
+    // TODO (before merging) - add method for get/set the reverse
     default boolean failsQualityVendorCheck() {
         return isSet(SAMFlag.READ_FAILS_VENDOR_QUALITY_CHECK);
     }
@@ -886,6 +907,7 @@ public interface Read extends Locatable {
      *
      * @param failsQualityVendorCheck {@code true} if the read fails the quality vendor check; {@code false} otherwise.
      */
+    // TODO (before merging) - add method for get/set the reverse
     default void setFailsQualityVendorCheck(final boolean failsQualityVendorCheck) {
         addFlag(SAMFlag.READ_FAILS_VENDOR_QUALITY_CHECK, failsQualityVendorCheck);
     }
@@ -938,10 +960,23 @@ public interface Read extends Locatable {
      *
      * @param supplementaryAlignment {@code true} if the read is a supplementary alignment; {@code false} otherwise.
      */
+    // TODO (before merging) - add method for get the reverse
     default void setSupplementaryAlignment(final boolean supplementaryAlignment) {
         addFlag(SAMFlag.SUPPLEMENTARY_ALIGNMENT, supplementaryAlignment);
     }
 
+    /**
+     * Checks if the read is a secondary or supplementary alignment.
+     *
+     * <p>Equivalent to the 0x100 or 0x800 <b>FLAG</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation calls {@code isSecondaryAlignment() || isSupplementaryAlignment()}.
+     *
+     * @return {@code true} if the read is secondary or supplementary; {@code false} otherwise.
+     */
+    default boolean isSecondaryOrSupplementary() {
+        return isSecondaryAlignment() || isSupplementaryAlignment();
+    }
 
     ///////////////////////////////////////////////////////////
     // OPTIONAL FIELDS - typed getters/setters
@@ -954,7 +989,7 @@ public interface Read extends Locatable {
      *
      * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a {@link Character}.
      *
-     * @param tag the attribute tag.
+     * @param tag attribute tag (two character string).
      *
      * @return the character value.
      *
@@ -975,18 +1010,18 @@ public interface Read extends Locatable {
     }
 
     /**
-     * Gets the singed integer value associated with the tag.
+     * Gets the signed Integer value associated with the tag.
      *
-     * <p>Equivalent to an attribute with <b>TYPE i</b> or any <b>TYPE B</b> integer (signed or unsigned) in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     * <p>Equivalent to an attribute with <b>TYPE i</b> (signed or unsigned) in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
      *
      * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a numeric value within the integer range.
      *
-     * @param tag the attribute tag.
+     * @param tag attribute tag (two character string).
      *
-     * @return the integer value.
+     * @return the Integer value.
      *
      * @throws SAMException if the tag is invalid following the contract of {@link ReadAttribute#isValidTag(String)}.
-     * @throws SAMException if the attribute value is not an integer.
+     * @throws SAMException if the attribute value is not an Integer.
      */
     default Optional<Integer> getIntegerAttribute(final String tag) {
         final Optional<Object> attr = getAttribute(tag);
@@ -1006,18 +1041,18 @@ public interface Read extends Locatable {
     }
 
     /**
-     * Gets the float value associated with the tag.
+     * Gets the Float value associated with the tag.
      *
-     * <p>Equivalent to an attribute with <b>TYPE f</b> or <b>TYPE B</b> float in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     * <p>Equivalent to an attribute with <b>TYPE f</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
      *
      * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a {@link Float}.
      *
-     * @param tag the attribute tag.
+     * @param tag attribute tag (two character string).
      *
-     * @return the float value.
+     * @return the Float value.
      *
      * @throws SAMException if the tag is invalid following the contract of {@link ReadAttribute#isValidTag(String)}.
-     * @throws SAMException if the attribute is not a float.
+     * @throws SAMException if the attribute is not a Float.
      */
     default Optional<Float> getFloatAttribute(final String tag) {
         final Optional<Object> attr = getAttribute(tag);
@@ -1032,18 +1067,18 @@ public interface Read extends Locatable {
     }
 
     /**
-     * Gets the string associated with the tag.
+     * Gets the String associated with the tag.
      *
      * <p>Equivalent to an attribute with <b>TYPE Z</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
      *
      * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a {@link String}.
      *
-     * @param tag the attribute tag.
+     * @param tag attribute tag (two character string).
      *
-     * @return the string value.
+     * @return the String value.
      *
      * @throws SAMException if the tag is invalid following the contract of {@link ReadAttribute#isValidTag(String)}.
-     * @throws SAMException if the attribute is not a string.
+     * @throws SAMException if the attribute is not a String.
      */
     default Optional<String> getStringAttribute(final String tag) {
         final Optional<Object> attr = getAttribute(tag);
@@ -1058,18 +1093,18 @@ public interface Read extends Locatable {
     }
 
     /**
-     * Gets the short associated with the tag.
+     * Gets the Byte associated with the tag.
      *
      * <p>Equivalent to an attribute with <b>TYPE B</b> byte (signed or unsigned) in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
      *
-     * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a numeric within the byte range.
+     * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a numeric within the Byte range.
      *
-     * @param tag the attribute tag.
+     * @param tag attribute tag (two character string).
      *
-     * @return the byte array value.
+     * @return the Byte value.
      *
      * @throws SAMException if the tag is invalid following the contract of {@link ReadAttribute#isValidTag(String)}.
-     * @throws SAMException if the attribute is not a float.
+     * @throws SAMException if the attribute is not a Byte.
      */
     default Optional<Byte> getByteAttribute(final String tag) {
         final Optional<Object> attr = getAttribute(tag);
@@ -1091,18 +1126,18 @@ public interface Read extends Locatable {
     }
 
     /**
-     * Gets the short associated with the tag.
+     * Gets the Short associated with the tag.
      *
      * <p>Equivalent to an attribute with <b>TYPE B</b> short (signed or unsigned) in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
      *
-     * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a numeric within the short range.
+     * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a numeric within the Short range.
      *
-     * @param tag the attribute tag.
+     * @param tag attribute tag (two character string).
      *
-     * @return the byte array value.
+     * @return the Short value.
      *
      * @throws SAMException if the tag is invalid following the contract of {@link ReadAttribute#isValidTag(String)}.
-     * @throws SAMException if the attribute is not a float.
+     * @throws SAMException if the attribute is not a Short.
      */
     default Optional<Short> getShortAttribute(final String tag) {
         final Optional<Object> attr = getAttribute(tag);
@@ -1130,7 +1165,7 @@ public interface Read extends Locatable {
      *
      * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a {@code int[]}.
      *
-     * @param tag the attribute tag.
+     * @param tag attribute tag (two character string).
      *
      * @return the int array value.
      *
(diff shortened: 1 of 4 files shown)


src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    /**
+     * Gets the {@link SAMFlag} that are set.
+     *
+     * <p>Equivalent to <b>FLAG</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation calls {@code SAMFlag.getFlags(getFlag())}.
+     *
+     * @return a set of the flags that are set for the read.
+     */
+    default Set<SAMFlag> getSetFlags() {
```
yfarjoun: this is a rather unfortunate name..."set" as an adjective....rename perhaps?

magicDGS: Thanks! I know that there are going to be a lot of discussion about naming, and that's why I submitted the PR. Any suggestion for a new name?

yfarjoun: actually, it's not clear if you meant "set" as a noun perhaps.....Argh. I don't have a better suggestion though.

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    void setFlag(final int flag);
+
+    /**
+     * Clears the bitwise flag, un-setting all the bits.
+     *
+     * <p>Equivalent to <b>FLAG</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation calls {@code setFlag(0)}.
+     */
+    default void clearFlag() {
```
yfarjoun: plural? `clearFlags`

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     * <p>Equivalent to <b>FLAG</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Deafault implementation modifies the {@link SAMFlag} bit of the {@link #getFlag()}, setting it afterwards with {@link #setFlag(int)} .
+     *
+     * @param flag flag to be set or unset.
+     * @param set {@code true} if the flag should be set; {@code false} otherwise.
+     *
+     * @throws IllegalArgumentException if the flag is {@code null}.
+     */
+    default void addFlag(final SAMFlag flag, final boolean set) {
```
yfarjoun: I think that this should still be called "setFlag", it can be set to True or set to False.

src/main/java/htsjdk/samtools/read/Read.java
yfarjoun: how about `getTrueFlags`?

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     *
+     * <p>Equivalent to <b>FLAG</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Note: it is preferable to use the methods for set/unset specific flags are set (using {@link #setFlag(SAMFlag)} or {@link #unsetFlag(SAMFlag)}),
+     * clear the flags using {@link #clearFlag()}, set/unset the specific flags (using {@link #setFlags(Set) and {@link #unsetFlags(Set)}},
+     * or use the shortcuts.
+     *
+     * @param flag bitwise flag to assign to the read.
+     */
+    void setFlag(final int flag);
```
yfarjoun: there should also be an unsetFlag, no?

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     * @param flag flag to be set or unset.
+     * @param set {@code true} if the flag should be set; {@code false} otherwise.
+     *
+     * @throws IllegalArgumentException if the flag is {@code null}.
+     */
+    default void addFlag(final SAMFlag flag, final boolean set) {
+        if (flag == null) {
+            throw new IllegalArgumentException("null flag");
+        }
+        int bitwiseFlag = getFlag();
```
yfarjoun: final

src/main/java/htsjdk/samtools/read/Read.java
yfarjoun: oh, I see this is for setting all the bits...

I think we should have a nomenclature that distinguishes between a single bit and the entires set of bits. I would advocate that "flag" (singular) refers to a bit, and "flags" (plural) refers to the entire integer that is all the bits.

yfarjoun: and thus, would call this set Flags.

src/main/java/htsjdk/samtools/read/Read.java
yfarjoun: nm

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     *
+     * <p>Equivalent to <b>RNAME</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * @return name of the contig this read is mapped to; {@link ReadConstants#NO_ALIGNMENT_REFERENCE_NAME} if unset (e.g. for unmapped reads).
+     */
+    @Override
+    String getContig();
+
+    /**
+     * Gets the 1-based inclusive leftmost position of the sequence remaining after clipping.
```
yfarjoun: I would stick to the language from the spec: 

>1-based leftmost mapping POSition of the first CIGAR operation that “consumes” a reference base

in particular this doesn't include insertions, N operations and P operators (in addition to padding)

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: I used this nomenclature to do not confound with the `setFlag(SAMFlag)` and `unsetFlag(SAMFlag)` for single flags. Does this make sense?

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     *
+     * <p>Equivalent to <b>POS</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * @return 1-based start position; {@link ReadConstants#NO_ALIGNMENT_START} if there is no position (e.g. for unmapped read).
+     */
+    @Override
+    int getStart();
+
+    /**
+     * Gets the 1-based inclusive rightmost position of the sequence remaining after clipping.
```
yfarjoun: again, we should be clear about the INP operators.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Good point! That's a much better description...

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+        }
+        setAlignmentPosition(position.getContig(), position.getStart());
+    }
+
+    /**
+     * Gets the read length.
+     *
+     * <p>Note: this is not necessarily the same as the number of reference bases the read is aligned to.
+     *
+     * @return The number of bases in the read
```
yfarjoun: do you mean the number of bases in the read-**sequence**?

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     * <p>Default implementation returns {@code getLength() == 0}.
+     *
+     * @return {@code true} if the read has no bases; {@code false} otherwise.
+     */
+    default boolean isEmpty() {
+        return getLength() == 0;
+    }
+
+    /**
+     * Gets the PHERD scaled mapping quality.
```
yfarjoun: PHERD->PHRED

yfarjoun: also, it's not capitalized normally: `phred`

yfarjoun: Link to MAPQ in the sam spec

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: I think that after you pointed out the language problem with the flags, we should really think carefully about the naming of all the flag methods:

* Setters: we should distinguish between single/all bits, as you pointed out below. `Flag` vs `Flags` is a good suggestion, but maybe we can be more clear/explicit. Because using the int flag is kind of discourage, what's about calling the ones setting all the bits `setRawFlag()` and `getRawFlag()`. Or maybe `Bitwise` instead of `Raw`.
* Getters: for single values, we have the `isSet` and `isUnset`. The problem comes with methods like the one pointed here: get all the flags that are set or unset. If we change the nomenclature to `getTrueFlags()` and `getFalseFlags()`, they would be out-of-sync with other getters/setters.

This name is kind of difficult to solve. Any proposal for all the methods will be welcomed!

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: True.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Opps.

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    /**
+     * Gets the short associated with the tag.
+     *
+     * <p>Equivalent to an attribute with <b>TYPE B</b> byte (signed or unsigned) in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a numeric within the byte range.
+     *
+     * @param tag the attribute tag.
+     *
+     * @return the byte array value.
```
yfarjoun: Byte ~array~

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    /**
+     * Gets the string associated with the tag.
+     *
+     * <p>Equivalent to an attribute with <b>TYPE Z</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a {@link String}.
+     *
+     * @param tag the attribute tag.
+     *
+     * @return the string value.
```
yfarjoun: capital S String

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+        }
+        final Object val = attr.get();
+        if (val instanceof Float) {
+            return Optional.of((Float) val);
+        }
+        throw new SAMException("Value for tag " + tag + " is not a Float: " + val.getClass());
+    }
+
+    /**
+     * Gets the string associated with the tag.
```
yfarjoun: String

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+        }
+        final long longVal = ((Number)val).longValue();
+        if (longVal < Integer.MIN_VALUE || longVal > Integer.MAX_VALUE) {
+            throw new SAMException("Value for tag " + tag + " is not in Integer range: " + longVal);
+        }
+        return Optional.of((int) longVal);
+    }
+
+    /**
+     * Gets the float value associated with the tag.
```
yfarjoun: Float

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     *
+     * <p>Equivalent to an attribute with <b>TYPE f</b> or <b>TYPE B</b> float in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a {@link Float}.
+     *
+     * @param tag the attribute tag.
+     *
+     * @return the float value.
```
yfarjoun: Float

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    }
+
+    /**
+     * Gets the float value associated with the tag.
+     *
+     * <p>Equivalent to an attribute with <b>TYPE f</b> or <b>TYPE B</b> float in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a {@link Float}.
+     *
+     * @param tag the attribute tag.
```
yfarjoun: (two character string)

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+        }
+        final Object val = attr.get();
+        if (val instanceof String) {
+            return Optional.of((String) val);
+        }
+        throw new SAMException("Value for tag " + tag + " is not a String: " + val.getClass());
+    }
+
+    /**
+     * Gets the short associated with the tag.
```
yfarjoun: Byte

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+        final Object val = attr.get();
+
+        if (val instanceof Character) {
+            return Optional.of((Character)val);
+        }
+        throw new SAMException("Value for tag " + tag + " is not a Character: " + val.getClass());
+    }
+
+    /**
+     * Gets the singed integer value associated with the tag.
```
yfarjoun: signed Integer

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     *
+     * <p>Equivalent to an attribute with <b>TYPE i</b> or any <b>TYPE B</b> integer (signed or unsigned) in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a numeric value within the integer range.
+     *
+     * @param tag the attribute tag.
+     *
+     * @return the integer value.
```
yfarjoun: Integer

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+        }
+        final long longVal = ((Number)val).longValue();
+        if (longVal < Byte.MIN_VALUE || longVal > Byte.MAX_VALUE) {
+            throw new SAMException("Value for tag " + tag + " is not in Short range: " + longVal);
+        }
+        return Optional.of((byte)longVal);
+    }
+
+    /**
+     * Gets the short associated with the tag.
```
yfarjoun: Short

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     * <p>Equivalent to an attribute with <b>TYPE B</b> short (signed or unsigned) in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a numeric within the short range.
+     *
+     * @param tag the attribute tag.
+     *
+     * @return the byte array value.
+     *
+     * @throws SAMException if the tag is invalid following the contract of {@link ReadAttribute#isValidTag(String)}.
+     * @throws SAMException if the attribute is not a float.
```
yfarjoun: ~float~ Short

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     * <p>Equivalent to an attribute with <b>TYPE B</b> byte (signed or unsigned) in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a numeric within the byte range.
+     *
+     * @param tag the attribute tag.
+     *
+     * @return the byte array value.
+     *
+     * @throws SAMException if the tag is invalid following the contract of {@link ReadAttribute#isValidTag(String)}.
+     * @throws SAMException if the attribute is not a float.
```
yfarjoun: byte

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    default Optional<float[]> getFloatArrayAttribute(final String tag) {
+        final Optional<Object> attr = getAttribute(tag);
+        if (!attr.isPresent()) {
+            return Optional.empty();
+        }
+        final Object val = attr.get();
+        if (val instanceof float[]) {
+            return Optional.of((float[])val);
+        }
+        throw new SAMException("Value for tag " + tag + " is not a byte[]: " + val.getClass());
```
yfarjoun: ~byte~ float

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+        if (val instanceof int[]) {
+            return Optional.of((int[])val);
+        }
+        throw new SAMException("Value for tag " + tag + " is not a int[]: " + val.getClass());
+    }
+
+    /**
+     * Gets the float array associated with the tag.
+     *
+     * <p>Equivalent to an attribute with <b>TYPE B</b> float array (signed or unsigned) in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
```
yfarjoun: there's only signed floats.

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+        if (val instanceof Character) {
+            return Optional.of((Character)val);
+        }
+        throw new SAMException("Value for tag " + tag + " is not a Character: " + val.getClass());
+    }
+
+    /**
+     * Gets the singed integer value associated with the tag.
+     *
+     * <p>Equivalent to an attribute with <b>TYPE i</b> or any <b>TYPE B</b> integer (signed or unsigned) in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
```
yfarjoun: what do you mean "or any type B integer"? thats an array...how will you put an array into an Integer?

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+        if (longVal < Integer.MIN_VALUE || longVal > Integer.MAX_VALUE) {
+            throw new SAMException("Value for tag " + tag + " is not in Integer range: " + longVal);
+        }
+        return Optional.of((int) longVal);
+    }
+
+    /**
+     * Gets the float value associated with the tag.
+     *
+     * <p>Equivalent to an attribute with <b>TYPE f</b> or <b>TYPE B</b> float in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
```
yfarjoun: same here about "or TYPE B"

src/main/java/htsjdk/samtools/read/ReadAttribute.java
```diff
@@ -0,0 +1,97 @@
...
+     *
+     * @return the tag.
+     */
+    String getTag();
+
+    /**
+     * Gets the read attribute value.
+     *
+     * <p>Subclasses should return a value that returns {@code true} by {@link
+     * #isAllowedAttributeValue(Object)}.
```
yfarjoun:
```
Subclasses should return a value on which {@link #isAllowedAttributeValue(Object)} returns {@code true}
```

src/main/java/htsjdk/samtools/read/Read.java
yfarjoun: no..add means something very different....

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+
+    /**
+     * Gets the {@link Cigar} object describing how the read aligns to the reference.
+     *
+     * <p>This method should make a defensive copy of the {@link Cigar} before returning it to allow
+     * modification of the returned {@link Cigar} without effects on the {@link Read}.
+     *
+     * @return Cigar object for the read; empty cigar if there is none.
+     */
+    Cigar getCigar();
```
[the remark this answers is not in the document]
magicDGS: I think that enforcing that in the interface is the best way to ensure that `Read` implementations don't behave unexpectedly when returned values are modified. An idea to be able to retrieve "unsafe" fields: `getUnsafe*` methods leaving to the implementation how to handle them, default getters for returning defensive copies (using the `getUnsafe* ` implementation), and javadoc pointing that it is recommended to use the methods returning the defensive copy unless it is required.

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     * <p>Default implementation returns {@code getBases()[i]}.
+     * Subclasses may override to provide a more efficient implementation.
+     *
+     * @return base at index i.
+     *
+     * @throws IndexOutOfBoundsException if i is negative or of i is not smaller than the number
+     * of bases (as reported by {@link #getLength()}. In particular, if no sequence is present.
+     */
+    default byte getBase(final int i){
+        return getBases()[i];
```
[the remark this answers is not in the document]
magicDGS: That's the default implementation...sub-classes should override for a more efficient way of getting the base (as stated in the javadoc). In principle this method is good for iterate over each base without getting a defensive copy, if the sub-class cares about efficiency. Another option is to leave the method without a default implementation....

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+
+    /**
+     * Checks if the read has qualities.
+     *
+     * <p>Default implementation returns {@code getBaseQualityLength() != 0}.
+     *
+     * @return {@code true} if the read has qualities; {@code false} otherwise.
+     */
+    default boolean hasQualities() {
+        return getBaseQualityLength() != 0;
```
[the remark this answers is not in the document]
magicDGS: If implementation details are maintained, this one will be more efficient if the sub-class access directly the length; otherwise, this will do a defensive copy...

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    default String getBaseQualityString() {
+        return hasQualities() ? SAMUtils.phredToFastq(getBaseQualities()) : ReadConstants.NULL_QUALS_STRING;
+    }
+
+    /**
+     * Set the base qualities.
+     *
+     * <p>Equivalent to <b>QUAL</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * @param baseQualities base qualities as binary phred scores (not ASCII); {@link ReadConstants#NULL_QUALS} if no base qualities are present.
```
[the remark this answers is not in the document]
magicDGS: This documentation comes from the `SAMRecord`one. It refers to the `SAMUtils.fastqToPhred(char)` return value, which is not the char byte (ASCII) but the actual integer value (char - 33, in Sanger format and the one assumed in `SAMUtils`).

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: True, I don't know what I did here....

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     * <p>The {@link ReadConstants#UNKNOWN_MAPPING_QUALITY} implies valid mapping, but hard to compute quality.
+     *
+     * <p>Equivalent to <b>MAPQ</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * @return mapping quality.
+     */
+    int getMappingQuality();
+
+    /**
+     * Sets the PHRED scaled mapping quality.
```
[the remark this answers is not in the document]
magicDGS: Done

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     *
+     * <p>Equivalent to <b>MAPQ</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * @param mapq mapping quality.
+     */
+    void setMappingQuality(final int mapq);
+
+    /**
+     * Gets the {@link Cigar} object describing how the read aligns to the reference.
+     *
```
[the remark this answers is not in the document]
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     * modification of the returned {@link Cigar} without effects on the {@link Read}.
+     *
+     * @return Cigar object for the read; empty cigar if there is none.
+     */
+    Cigar getCigar();
+
+    /**
+     * Gets the number of cigar elements (number + operator) in the cigar string.
+     *
+     * <p>Default implementation returns {@code getCigar().numCigarElements()}.
```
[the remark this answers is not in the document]
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     *
+     * @return number of cigar elements (number + operator) in the cigar string.
+     */
+    default int getCigarLength() {
+        return getCigar().numCigarElements();
+    }
+
+    /**
+     * Sets the {@link Cigar} object describing how the read aligns to the reference.
+     *
```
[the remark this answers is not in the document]
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    void setCigar(final Cigar cigar);
+
+    /**
+     * Gets the contig name where the read's mate is mapped to. May return null if there is no unique mapping.
+     *
+     * <p>Equivalent to <b>RNEXT</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * @return name of the contig the read's mate is mapped to; {@link ReadConstants#NO_ALIGNMENT_REFERENCE_NAME} if unset (e.g. for unmapped reads).
+     */
+    String getMateContig();
```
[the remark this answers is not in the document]
magicDGS: Using `next` to allow multiple segments would be nice. Let's see what the rest of the community things. Maybe we should point an example for pair-end reads too, because is the most common next-template, I guess...

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     * @param position location the read's mate is mapped to.
+     *
+     * @throws IllegalArgumentException if the locatable is {@code null}.
+     */
+    default void setMateAlignmentPosition(final Locatable position) {
+        setMateAlignmentPosition(position.getContig(), position.getStart());
+    }
+
+    /**
+     * Gets the insert size (difference between the 5' end of the read and the 5' end of the mate).
```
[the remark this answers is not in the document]
magicDGS: I used the same documentation as `SAMRecord`. I changed it to the specs documentation, but maybe it would require a change of the name if so. New text:

```java
    /**
     * Gets the signed observed template length. If  all  segments  are  mapped  to  the  same  reference,
     * the unsigned observed template length equals the number of bases from the leftmost mapped base to the
     * rightmost mapped base.  The leftmost segment has a plus sign and the rightmost has a minus sign.
     * The sign of segments in the middle is undefined. It is set as 0 for single-segment template or when the
     * information is unavailable.
     *
     * <p>Equivalent to <b>TLEN</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
     *
     * @return insert size if possible to compute; {@code 0} otherwise.
     */
    int getInferredInsertSize();
```

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     * The insert size is negative if the mate maps to lower position than the read.
+     *
+     * <p>Equivalent to <b>TLEN</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * @return insert size if possible; {@code 0} otherwise.
+     */
+    int getInferredInsertSize();
+
+    /**
+     * Sets the insert size (difference between the 5' end of the read and the 5' end of the mate).
```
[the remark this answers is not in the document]
magicDGS: Same as above. Done.

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     * The insert size is negative if the mate maps to lower position than the read.
+     *
+     * <p>Equivalent to <b>TLEN</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * @param insertSize insert size (negative if mate mapts to lower position than the read) if possible; {@code 0} otherwise.
```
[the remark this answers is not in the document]
magicDGS: This line was rephrased.

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+
+    /**
+     * Gets the base at position i.
+     *
+     * <p>Equivalent to <b>SEQ</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation returns {@code getBases()[i]}.
+     * Subclasses may override to provide a more efficient implementation.
+     *
+     * @return base at index i.
```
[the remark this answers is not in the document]
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     *
+     * @throws IndexOutOfBoundsException if i is negative or of i is not smaller than the number
+     * of bases (as reported by {@link #getLength()}. In particular, if no sequence is present.
+     */
+    default byte getBase(final int i){
+        return getBases()[i];
+    }
+
+    /**
+     * Gets all the bases in the read as a String of ACGTN=
```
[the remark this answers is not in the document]
magicDGS: Again, copied from the `SAMRecord` javadoc. Re-written. By the way, the code in the `SAMRecord` sometimes normalize the bases (if they are set with a String), but not in other cases... That should be solved by using the current interface, that does not allow to set the bases with a String

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    default String getBasesString() {
+        return isEmpty() ? ReadConstants.NULL_SEQUENCE_STRING : StringUtil.bytesToString(getBases());
+    }
+
+    /**
+     * Sets the sequence bases.
+     *
+     * <p>Equivalent to <b>SEQ</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * @param bases read sequence as ASCII bytes ACGTN=; {@link ReadConstants#NULL_SEQUENCE} if no sequence is present.
```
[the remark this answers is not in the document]
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     *
+     * @throws IllegalArgumentException if the bases are null.
+     */
+    void setBases(final byte[] bases);
+
+    /**
+     * Gets the base qualities, as binary PHRED scores (not ASCII).
```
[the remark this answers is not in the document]
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+     * @throws IllegalArgumentException if the base qualities are null or an invalid (negative) base quality is provided.
+     */
+    void setBaseQualities(final byte[] baseQualities);
+
+    ///////////////////////////////////////////////////////////
+    // OPTIONAL FIELDS - untyped getters/setters
+    ///////////////////////////////////////////////////////////
+
+    /**
+     * Gets the list of optional fields for the read. This are encoded as tag/value pairs, where
```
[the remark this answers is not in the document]
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    /**
+     * Checks if the read is unmapped.
+     *
+     * <p>Equivalent to the 0x4 <b>FLAG</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation calls {@link #isSet(SAMFlag)} for {@link SAMFlag#READ_UNMAPPED}.
+     *
+     * @return {@code true} if this read is unmapped; {@code false} otherwise.
+     */
+    default boolean isUnmapped() {
```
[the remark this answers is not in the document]
magicDGS: I haven't added the reverse methods for any flag yet, waiting for comments about the naming. I will add them in a later commit once we agree on what is going to be the convention.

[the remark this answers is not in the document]
magicDGS: By now, I added a TODO line for being aware of that before merging.

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    /**
+     * Mark the read as unmapped or not.
+     *
+     * <p>Equivalent to the 0x4 <b>FLAG</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation calls the {@code addFlag(SAMFlag.READ_UNMAPPED, unmapped)}.
+     *
+     * @param unmapped {@code true} if this read is unmapped; {@code false} otherwise.
+     */
+    default void setUnmapped(final boolean unmapped) {
```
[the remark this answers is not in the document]
magicDGS: Added TODO

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    /**
+     * Checks if the read's mate is unmapped.
+     *
+     * <p>Equivalent to the 0x8 <b>FLAG</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation calls {@link #isSet(SAMFlag)} for {@link SAMFlag#MATE_UNMAPPED}.
+     *
+     * @return {@code true} if this read's mate is unmapped; {@code false} otherwise.
+     */
+    default boolean isMateUnmapped() {
```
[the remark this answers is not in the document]
magicDGS: Added TODO

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    /**
+     * Mark the read's mate as unmapped or not.
+     *
+     * <p>Equivalent to the 0x8 <b>FLAG</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation calls the {@code addFlag(SAMFlag.MATE_UNMAPPED, mateUnmapped)}.
+     *
+     * @param mateUnmapped {@code true} if this read's mate is unmapped; {@code false} otherwise.
+     */
+    default void setMateUnmapped(final boolean mateUnmapped) {
```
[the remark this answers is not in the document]
magicDGS: Added TODO

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    /**
+     * Checks if the read is in the reverse strand.
+     *
+     * <p>Equivalent to the 0x10 <b>FLAG</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation calls {@link #isSet(SAMFlag)} for {@link SAMFlag#READ_REVERSE_STRAND}.
+     *
+     * @return {@code true} if the read is in the reverse strand; {@code false} otherwise.
+     */
+    default boolean isReverseStrand() {
```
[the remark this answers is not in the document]
magicDGS: Added TODO

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    /**
+     * Marks the read as being in the reverse strand or not.
+     *
+     * <p>Equivalent to the 0x10 <b>FLAG</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation calls the {@code addFlag(SAMFlag.READ_REVERSE_STRAND, reverseStrand)}.
+     *
+     * @param reverseStrand {@code true} if the read is in the reverse strand; {@code false} otherwise.
+     */
+    default void setReverseStrand(final boolean reverseStrand) {
```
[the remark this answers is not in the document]
magicDGS: Added TODO

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    /**
+     * Checks if the read's mate is in the reverse strand.
+     *
+     * <p>Equivalent to the 0x20 <b>FLAG</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation calls {@link #isSet(SAMFlag)} for {@link SAMFlag#MATE_REVERSE_STRAND}.
+     *
+     * @return {@code true} if the read's mate is in the reverse strand; {@code false} otherwise.
+     */
+    default boolean isMateReverseStrand() {
```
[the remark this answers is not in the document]
magicDGS: Added TODO

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    /**
+     * Marks the read's mate as being in the reverse strand or not.
+     *
+     * <p>Equivalent to the 0x20 <b>FLAG</b> in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation calls the {@code addFlag(SAMFlag.MATE_REVERSE_STRAND, mateReverseStrand)}.
+     *
+     * @param mateReverseStrand {@code true} if the read's mate is in the reverse strand; {@code false} otherwise.
+     */
+    default void setMateReverseStrand(final boolean mateReverseStrand) {
```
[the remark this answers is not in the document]
magicDGS: Added TODO

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Removed.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Removed

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Changed everywhere (but already specified in the `SAMException` documentation).

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
```diff
@@ -0,0 +1,1438 @@
...
+    /**
+     * Gets the short associated with the tag.
+     *
+     * <p>Equivalent to an attribute with <b>TYPE B</b> short (signed or unsigned) in the <a href="http://samtools.github.io/hts-specs/SAMv1.pdf">SAM specifications</a>.
+     *
+     * <p>Default implementation gets the attribute with {@link #getAttribute(String)}, checking if it is a numeric within the short range.
+     *
+     * @param tag the attribute tag.
+     *
+     * @return the byte array value.
```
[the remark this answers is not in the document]
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Done.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Removed.

src/main/java/htsjdk/samtools/read/Read.java
magicDGS: Done.

src/main/java/htsjdk/samtools/read/ReadAttribute.java
magicDGS: Done.

magicDGS:
After addressing some minor comments (mostly documentation), there are things to be discussed:

* Naming of getters/setters for bitwise flags. Clunky names like `getSetFlags()` are included at the moment. Inconsistency between the int value and the `SAMFlag` getters/setters should be solved. Any proposal?
* Rename some methods for being more general: `*Mate*` -> `*Next*` (more than one template, not only PE), `*InsertSize` -> `*TemplateLength*,  `failsQualityVendorCheck` -> `failsQualityFilters()`. I think that this changes are fair enough...should they be included in the next commit?
* Big discussion: force defensive copy vs. implementation-dependent. I'm in favor of minimize the code that could be modified without noticing due to implementation decisions. Nevertheless, I see the case where the developer wants to access directly the fields to modify in-place; in that cases, I suggest to add to the interface "unsafe" methods that are discourage in favor of the safe methods, unless the API user knows what they are doing. I think that this is a fair compromise between safety and accessibility...

No action was done for more getters/setters for flags that are the reverse, until a consensus for the names is reached (that will make easier to add them at one without including too many changes). They will be included after most of the changes are already accepted according naming of the methods.

Thanks for your first pass, @yfarjoun - looking forward to @lbergelson review!

src/main/java/htsjdk/samtools/read/Read.java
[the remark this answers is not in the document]
magicDGS: I realized that this will be important also for the flag methods (isPaired, isMateUnmapped). Maybe we can have also sugar-syntax for pair-end apart of multiple segments...

magicDGS:
Most of the renaming issues were solved in independent commits (failsQualityVendorCheck, InsertSize, and flags). I am still waiting for the feedback about the Mate/Next nomenclature and the defensive copy behaviour.

magicDGS:
Friendly ping here, @yfarjoun!

yfarjoun:
it's not in my court. waiting for @lbergelson or one of the maintainers.

magicDGS:
Thank @yfarjoun - I though that you had some suggestion still...

yfarjoun:
closing due to imminent inclusion in [github.com/samtools/htsjdk-next-beta](htsjdk-next-beta)

PR #986, opened 2017-09-15
Some refactoring of the Strand enum adding encoding (string and char)…

vruano:
… functionality.

### Description

Some refactoring of the Strand enum adding encoding (string and char) functionality.

Strand is lacking encoding methods it only had a decoding one (toStrand(String)) and an attempt to use toString or name would return POSITIVE, NEGATIVE... when in practice one wants '+', '-'... 

It also deprecates 'toStrand(String)' for 'decode(String)' for name consistency sake (encode/decode) and 'Strand' in toStrand is kinda redundant... ideally one would like to name it valueOf but is impossible (nor desirable) to override valueOf(String). 

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/tribble/annotation/Strand.java
+++ b/src/main/java/htsjdk/tribble/annotation/Strand.java
@@ -24,29 +24,113 @@
 package htsjdk.tribble.annotation;
 
 /**
- * Enum for strand, which can be encoded as string
+ * Enum for strand, which can be encoded as a string
  */
 public enum Strand {
-    POSITIVE("+"), NEGATIVE("-"), NONE("!");  // not really sure what we should do for the NONE Enum
 
     /**
-     * How we represent the strand information as text
+     * Represents the positive or forward strand.
      */
-    private String encoding;
-    Strand(String str) {
-        encoding = str;
+    POSITIVE('+'),
+
+    /**
+     * Represents the negative or reverse strand.
+     */
+    NEGATIVE('-'),
+
+    /**
+     * Denotes that a strand designation is not applicable
+     * or is unknown.
+     */
+    NONE('.');  // not really sure what we should do for the NONE Enum
+
+    /**
+     * Common alias for the {@link #POSITIVE} strand.
+     */
+    public static final Strand FORWARD = POSITIVE;
+
+    /**
+     * Common alias for the {@link #NEGATIVE} strand.
+     */
+    public static final Strand REVERSE = NEGATIVE;
+
+    /**
+     * Cached array of instances.
+     */
+    private final static Strand[] VALUES = values();
+
+    /**
+     * How we represent the strand as a single {@code char}.
+     */
+    private final char charEncoding;
+
+    /**
+     * How we represent the strand as a {@link String}.
+     */
+    private final String stringEncoding;
+
+    Strand(final char ch) {
+        charEncoding = ch;
+        stringEncoding = String.valueOf(charEncoding);
     }
 
     /**
      * provide a way to take an encoding string, and produce a Strand
      * @param encoding the encoding string
      * @return a Strand object, if an appropriate one cannot be located an IllegalArg exception
+     * @deprecated please use {@link #decode} instead.
      */
-    public static Strand toStrand(String encoding) {
-        for (Strand st : Strand.values())
-            if (st.encoding.equals(encoding))
-                return st;
-        throw new IllegalArgumentException("Unable to match encoding to Strand enum for encoding string " + encoding);
+    @Deprecated
+    public static Strand toStrand(final String encoding) {
+        return decode(encoding);
     }
 
+    /**
+     * Returns the {@link Strand} that a {@code char} value represents.
+     * @param ch the char encoding for a Strand.
+     * @return never {@code null}, a value so that {@code decode(c).encodeAsChar() == c} or {@link #NONE} if
+     *   the encoding char is not recognized.
+     */
+    public static Strand decode(final char ch) {
+        for(final Strand value : VALUES) {
+            if (value.charEncoding == ch) {
+                return value;
+            }
+        }
+        return NONE;
+    }
+
+    /**
+     * Returns the {@link Strand} that a {@link String} encodes for.
+     * @param encoding the strand string representation.
+     * @return never {@code null}, a value so that {@code decode(s).encode().equals(s)}, or {@link #NONE} if
+     * the encoding string is not recognized.
+     */
+    public static Strand decode(final String encoding) {
+        if (encoding != null && encoding.length() == 1) {
+            return decode(encoding.charAt(0));
+        }
+        return NONE;
+    }
+
+    /**
+     * Returns a string representation of this {@link Strand}
+     * @return never {@code null}, a value so that {@code decode(encode(X)) == X}.
+     */
+    public String encode() {
+        return stringEncoding;
+    }
+
+    /**
+     * Returns a single char encoding of this {@link Strand}.
+     * @return a value so that {@code decode(encodeAsChar(X)) == X}.
+     */
+    public char encodeAsChar() {
+        return charEncoding;
+    }
+
+    @Override
+    public String toString() {
+        return stringEncoding;
+    }
 }
--- a/src/test/java/htsjdk/tribble/annotation/StrandTest.java
+++ b/src/test/java/htsjdk/tribble/annotation/StrandTest.java
@@ -1,0 +1,124 @@
+package htsjdk.tribble.annotation;
+
+import htsjdk.HtsjdkTest;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Unit tests for {@link Strand}.
+ * <p>
+ *     Makes sure that {@link Strand} will behave according to contract in run-time.
+ * </p>
+ */
+public class StrandTest extends HtsjdkTest {
+
+    @Test
+    public void testEncodeReturnNoNulls() {
+        for (final Strand subject : Strand.values()) {
+            Assert.assertNotNull(subject.encode());
+        }
+    }
+
+    @Test
+    public void testEncodeDecode() {
+        for (final Strand subject : Strand.values()) {
+            final String encode = subject.encode();
+            final Strand back = Strand.decode(encode);
+            Assert.assertSame(back, subject);
+        }
+    }
+
+    @Test
+    public void testEncodeDecodeAsChar() {
+        for (final Strand subject : Strand.values()) {
+            final char encode = subject.encodeAsChar();
+            final Strand back = Strand.decode(encode);
+            Assert.assertSame(back, subject);
+        }
+    }
+
+    @Test
+    public void testEncodeUniqueness() {
+        Assert.assertEquals(Stream.of(Strand.values())
+                .map(Strand::encode)
+                .collect(Collectors.toSet()).size(), Strand.values().length);
+    }
+
+    @Test
+    public void testEncodeAsCharUniqueness() {
+        Assert.assertEquals(Stream.of(Strand.values())
+                .map(Strand::encodeAsChar)
+                .collect(Collectors.toSet()).size(), Strand.values().length);
+    }
+
+    @Test
+    public void testEncodeAsStringAndCharAreEquivalent() {
+        for (final Strand subject : Strand.values()) {
+            final String str = subject.encode();
+            final char ch = subject.encodeAsChar();
+            Assert.assertEquals(str, String.valueOf(ch));
+        }
+    }
+
+    @Test
+    public void testAliases() {
+        Assert.assertSame(Strand.FORWARD, Strand.POSITIVE);
+        Assert.assertSame(Strand.REVERSE, Strand.NEGATIVE);
+    }
+
+    @Test
+    public void testToStrandMatchesDecode() {
+        for (final Strand value : Strand.values()) {
+            Assert.assertSame(Strand.toStrand(value.encode()), Strand.decode(value.encode()));
+        }
+    }
+
+    @Test
+    public void testToStringMatchesEncode() {
+        for (final Strand value : Strand.values()) {
+            Assert.assertEquals(value.toString(), value.encode());
+        }
+    }
+
+    @Test
+    public void testCurrentEncodings() {
+        Assert.assertSame(Strand.decode('+'), Strand.POSITIVE);
+        Assert.assertSame(Strand.decode('-'), Strand.NEGATIVE);
+        Assert.assertSame(Strand.decode('.'), Strand.NONE);
+
+        Assert.assertSame(Strand.decode("+"), Strand.POSITIVE);
+        Assert.assertSame(Strand.decode("-"), Strand.NEGATIVE);
+        Assert.assertSame(Strand.decode("."), Strand.NONE);
+    }
+
+    @Test(dataProvider = "invalidEncodingStringsData")
+    public void testDecodeOfInvalidEncodingStrings(final String encoding) {
+        Assert.assertSame(Strand.decode(encoding), Strand.NONE);
+    }
+
+    @Test(dataProvider = "invalidEncodingCharsData")
+    public void testDecodeOfInvalidEncodingChars(final char encoding) {
+        Assert.assertSame(Strand.decode(encoding), Strand.NONE);
+    }
+
+    @DataProvider(name = "invalidEncodingStringsData")
+    public Object[][] invalidEncodingStringsData() {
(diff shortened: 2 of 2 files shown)


src/main/java/htsjdk/tribble/annotation/Strand.java
```diff
@@ -24,29 +24,102 @@
...
+    /**
+     * Returns the {@link Strand} that a {@code char} value represents.
+     * @param ch the char encoding for a Strand.
+     * @throws IllegalArgumentException if input char does not
+     *   represent a {@link #Strand} instance.
+     * @return never {@code null}, a value so that {@code decode(c).encodeAsChar() == c}.
+     */
+    public static Strand decode(final char ch) {
+        switch (ch) {
+            case '+': return POSITIVE;
```
pshapiro4broad: This duplicates the relationship already established in the enum definition. Instead you could write
```
for (Strand strand : values()) {
    if (strand.charEncoding == ch) {
        return strand;
    }
}
throw new ...
```

src/test/java/htsjdk/tribble/annotation/StrandTest.java
```diff
@@ -0,0 +1,94 @@
...
+                "FORWARD", "REVERSE",
+                " +", "- ", "p", "m", "x")
+                .map(s -> new Object[] { s}).toArray(Object[][]::new);
+    }
+
+    @DataProvider(name = "invalidEncodingCharsData")
+    public Object[][] invalidEncodingCharsData() {
+        final char[] testCases = new char[] {'c', '\t', '\n', (char) -1, 'p', 'n', 'x' };
+        return IntStream.range(0, testCases.length)
+                .mapToObj(i -> testCases[i])
```
pshapiro4broad: Do you need `IntStream` here? It looks like you could use `Stream.of()` as do you for the string data.

vruano: Yeah I guess i should just Use a ```Character[]``` array rather than a ```char``` array, perhaps implicitly boxed by Stream.of(Object ... o)

src/main/java/htsjdk/tribble/annotation/Strand.java
vruano: yeah... I guess it could be kinda of faster with the switch and I am relaying on the unit-test to verify consistency at build-time. But at best is premature optimization so happy to revert to that.

src/main/java/htsjdk/tribble/annotation/Strand.java
```diff
@@ -24,29 +24,102 @@
...
     /**
-     * How we represent the strand information as text
+     * Represent the positive or forward strand.
      */
-    private String encoding;
-    Strand(String str) {
-        encoding = str;
+    POSITIVE('+'),
+    /**
+     * Represent the negative or reverse strand.
```
vruano: Represents.

src/main/java/htsjdk/tribble/annotation/Strand.java
```diff
@@ -24,29 +24,102 @@
...
+     */
+    NEGATIVE('-'),
+
+    /**
+     * Denotes that an strand designation is not applicable
```
vruano: an -> a.

src/main/java/htsjdk/tribble/annotation/Strand.java
pshapiro4broad: If it turns out to be a performance issue, you can build a static `Map` using the enum values and also avoid the data duplication. Given that there are only three enum members it seems unlikely that this will be a bottleneck.

vruano: In fact the map is likely to carry too much overhead.

src/main/java/htsjdk/tribble/annotation/Strand.java
```diff
@@ -24,29 +24,102 @@
...
+     * or is unknown.
+     */
+    NONE('!');  // not really sure what we should do for the NONE Enum
```
magicDGS: Now that this is refactored, maybe is a good time to add two instances for no strand to distinguish between non-available (not parsed, column not present in the file) and no strand (`.` in the BED format).

src/test/java/htsjdk/tribble/annotation/StrandTest.java
```diff
@@ -0,0 +1,94 @@
+package htsjdk.tribble.annotation;
```
yfarjoun: these tests aren't running for some reason...there is no tests coverage in Strand.java

src/test/java/htsjdk/tribble/annotation/StrandTest.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+/**
+ * Unit tests for {@link Strand}.
+ * <p>
+ *     Makes sure that {@link Strand} will behave according to contract in run-time.
+ * </p>
+ */
+public class StrandTest {
```
yfarjoun: oh, you need to derive this class from HtsTestJava or something like that. take a look at one of the other files.

src/main/java/htsjdk/tribble/annotation/Strand.java
vruano: I'm happy to change '!' for a '.' ... in what situation/format a '!' would mean no-strand? 

However I'm not sure about adding more NONEs... I would like to see a situation in where one would need to distinguish between more than one NONE.

In any case one could use ```Optional<Strand>``` to add an extra NONE... or even drop the current one!

src/main/java/htsjdk/tribble/annotation/Strand.java
pshapiro4broad: I was curious so I wrote a short test. The test showed that switch is only slightly faster than the Map. Using a for loop is an order of magnitude slower than both the switch and map. So if you want to avoid duplication, Map seems to be the best solution.

https://gist.github.com/pshapiro4broad/1acabaa3fe74ac03007f15b989fe78a8

vruano: Wow! Never thought of it... perhaps the culprit is values()... if one is allowed to change its contents then it makes sense that each invocation would instantiate a new array.

vruano: If you are still curious and since you have the test code ready for action then you could try to statically cache the result of values().

pshapiro4broad: The code is in the gist link in my comment, I'll update it when I have a chance, but you can update it as well.

I would expect that `Enum.values()` is a static reference since all enum instances are known at the time the .class file is loaded. It would be easy to try out.

lbergelson: Arrays are always mutable, so I'm pretty sure it will allocate a fresh one on every call to values.

vruano: I have done the test and that is the cause of the bad performance. Caching it into an static final bring time to the par with other solutions.

pshapiro4broad: Good to know, thanks for checking this out.

src/test/java/htsjdk/tribble/annotation/StrandTest.java
vruano: Just submitted a commit extending HtsjdkTest. Hopefully that fixes it.

vruano:
Just submitted a commit addressing comments thus far. For decode I have opted for an updated for-each loop using a cached values arrays. 

For now I have not changed the NONE char representation from '!' to '.'. Please comment of whether you think that this change would be wise since it brakes back-comapatibily slightly. I have never seen '!' to be used to denote a No strand and '.' is commonly use to denote undefined values in common formats so it seems like a logical change.

Coverage test seems to do well now except for an unexpected change in coverage in an unchanged file... don't know why is that but it seems to be for the better so I guess it should not be a deterrent for a merge. 

There is a couple of uncovered lines in Strand (deprecated toStrand and toString) Will fix that before merging.

src/test/java/htsjdk/tribble/annotation/StrandTest.java
vruano: That did the trick, thanks!

vruano:
Also I have added a couple alias/constants FORWARD = POSITIVE, REVERSE = NEGATIVE.

src/main/java/htsjdk/tribble/annotation/Strand.java
```diff
@@ -24,29 +24,116 @@
...
+    /**
+     * Returns a single char encoding of this {@link Strand}.
+     * @return a value so that {@code decode(encodeAsChar(X)) == X}.
+     */
+    public char encodeAsChar() {
+        return charEncoding;
+    }
+
+    @Override
+    public String toString() {
```
magicDGS: Overriding this method makes conversion of  `String` <-> `Strand` failed in an unexpected way (unless documented). For example, `Strand.valueOf(Strand.POSITIVE.toString())` won't return `Strand.POSITIVE` anymore. Because `valueOf` cannot be overriden, I think that the `toString()` method shouldn't return the encoding....

src/main/java/htsjdk/tribble/annotation/Strand.java
```diff
@@ -24,29 +24,116 @@
...
+    /**
+     * Represents the negative or reverse strand.
+     */
+    NEGATIVE('-'),
+
+    /**
+     * Denotes that a strand designation is not applicable
+     * or is unknown.
+     */
+    NONE('!');  // not really sure what we should do for the NONE Enum
```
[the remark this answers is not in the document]
vruano: I prefer the ```Optional<X>``` to handle that for those that really need to make such a distinction. The fact that there is as many non-proper enum constants (?, .) than proper one (+, -) is already smelly IMO. Anyone else can chime in? Can be de tie breaker?

src/main/java/htsjdk/tribble/annotation/Strand.java
```diff
@@ -24,29 +24,116 @@
...
+     *   represent a {@link #Strand} instance.
+     * @return never {@code null}, a value so that {@code decode(c).encodeAsChar() == c}.
+     */
+    public static Strand decode(final char ch) {
+        for(final Strand value : VALUES) {
+            if (value.charEncoding == ch) {
+                return value;
+            }
+        }
+        throw new IllegalArgumentException("Unable to match encoding to Strand enum for encoding char: " + ch);
```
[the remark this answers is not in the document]
vruano: .... so that is even suggesting yet another constant INVALID? this is a dangerous slippery slope. I think this even strengths the point that no more non-proper enum constant should be added... tomorrow someone may thing of something else like "NOT_APPLICABLE"... "BOTH".... In this particular case I think is best to simply fail early if an invalid string is to be converted to a Strand

src/main/java/htsjdk/tribble/annotation/Strand.java
vruano: mmm... looking at the standard Java doc for ```name()```  it seems as the overriding of toString() is actually encouraged to provide a more user friendly string representation of the enum constant. 

Also valueOf javadoc says explicitly is mean to recover constant by its constant name so to be used like so 
```
String x = X.name();
...
// millions lines later in another class far far a way:
...
EnumX X = EnumX.valueOf(y) // where somehow y came from x.
```
So assuming that ```valueOf(X.toString()) == X``` is a understandable misinterpretation  of what ```valueOf``` is supposed to do. I guess that that is why it makes sense for it to fail with an Exception rather than return a null as go are getting wrong the hard-wired programatic name for a constant.

That said, it is true that this change break back-copatability a bit and I don't feel very strong about it ...to show "strand = POSITIVE" rather than "strand = +" in error messages doesn't really make much of a difference.... an further enforces the use of ```encode``` and ```decode``` for reading it in or out of files... so I will revert that. 

Any other opinions.

src/main/java/htsjdk/tribble/annotation/Strand.java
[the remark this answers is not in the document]
magicDGS: That's a change in the codecs, not in the enum,but I do not block the PR due to this discussion (I can propose a change in a following PR, either for the codec or for the enum). I am fine if only one "none" concept exists as far as it is encoded with the default for BED, because it is the codec which uses this currently (and also it is common in other formats).

For maintain backwards compatibility I guess that the best approach will be to deprecate `NONE` (to keep the `'!'` econding) in favor of a `MISSING`or `UNKNOWN` encoded with `'.'`. Once the `NONE` enum is removed, an alias can be added (similarly to the `POSITIVE`/`FORWARD` alias).

src/main/java/htsjdk/tribble/annotation/Strand.java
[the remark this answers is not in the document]
magicDGS: I am fine with the exception too, it was just an idea...

src/main/java/htsjdk/tribble/annotation/Strand.java
magicDGS: You're perfectly right, sorry. It is the value of `X.name()` the one that should be kept the same. I'm OK with the change, even if it is not backwards compatible -  in principle I am in favor of `toString()` methods only implemented for error/debugging messages and not rely on them for other purposes...

src/main/java/htsjdk/tribble/annotation/Strand.java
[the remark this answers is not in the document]
yfarjoun: I would go with "."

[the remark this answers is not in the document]
yfarjoun: where did the "!" suggestion come from?

src/test/java/htsjdk/tribble/annotation/StrandTest.java
```diff
@@ -0,0 +1,113 @@
...
+            Assert.assertSame(Strand.toStrand(value.encode()), Strand.decode(value.encode()));
+        }
+    }
+
+    @Test
+    public void testToStringMatchesEncode() {
+        for (final Strand value : Strand.values()) {
+            Assert.assertEquals(value.toString(), value.encode());
+        }
+    }
```
yfarjoun: it would be god to have an explicit test that shows that you can decode valid chars and strings into the three difference `Strand`s.

src/main/java/htsjdk/tribble/annotation/Strand.java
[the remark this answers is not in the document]
magicDGS: The `!` was the previous implementation... I guess that it comes from times when undefined wasn't `.` yet.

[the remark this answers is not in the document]
yfarjoun: oh, OK. leave as is then!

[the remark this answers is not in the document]
magicDGS: I would like to have `.`, because it is actually the representation for the BED file (at the moment, the only Strand usage within the library).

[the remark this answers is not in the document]
vruano: Unless someone have show that '!' is actually a used in practice I'm inclined to change it to '.'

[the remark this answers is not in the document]
vruano: I'm not going to add yet another "NONE" constant in this pull request.  @magicDGS if you fill strong about this I suggest that you work on another pull request with that change. Hopefully you can show an example how that that change improve some other piece of code in htsjdk.

yfarjoun:
an online search shows that BED format doesn't have a specified value for unspecified strand...however GFF (which in some-sense an extension of BED) does specify '.' as "unspecified".

I guess that's a 👍 from me.

please make sure that when *reading* a file you convert non-confoming values to "NONE" so that the odd BED file that has "!" will be read correctly.

src/main/java/htsjdk/tribble/annotation/Strand.java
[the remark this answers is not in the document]
magicDGS: I was referring in the past comment to do exactly the change from '!' to '.', as you've already did. Thanks!

vruano:
Now seems that is all in order. 

Now unrecognized encodings (including the '!') are decoded into NONE. 

Changes are mostly back-compatibly except for Strand#toString (uses the encoding rather than the constant name) and NONE canonic encoding is now '.' instead of '!'.

An unrelated issue with a URL change in ENA service was causing a test to fail, presumably master is also currently broken.

yfarjoun:
Looks good, though I'd ask that you remove the ENA modification and rebase once I merge that commit.

@lbergelson do you need to 👍 this?

vruano:
@lbergelson any updates?

lbergelson:
@vruano This looks good to me.  👍

Issue #969, opened 2017-08-18
unhelpful exception message when failing to load index


lbergelson:
A user on the gatk forums [reported](https://gatkforums.broadinstitute.org/gatk/discussion/comment/40871#Comment_40871) a gatk crash that seems to be caused by a corrupt index being read by htsjdk.  An underlying `IOException` is causing a `InvocationTargetException` which is then rewrapped as a `RuntimeException`.  The result is a confusing exception that gives you no information about what went wrong without digging into the stack trace.  
```
INFO  11:13:55,502 HelpFormatter - -------------------------------------------------------------------------------- 
INFO  11:13:55,615 HelpFormatter - The Genome Analysis Toolkit (GATK) v3.7-0-gcfedb67, Compiled 2016/12/12 11:21:18 
INFO  11:13:55,615 HelpFormatter - Copyright (c) 2010-2016 The Broad Institute 
INFO  11:13:55,616 HelpFormatter - For support and documentation go to https://software.broadinstitute.org/gatk 
INFO  11:13:55,616 HelpFormatter - [Wed Jul 26 11:13:55 PDT 2017] Executing on Linux 4.4.0-47-generic amd64 
INFO  11:13:55,616 HelpFormatter - Java HotSpot(TM) 64-Bit Server VM 1.8.0_72-b15 
INFO  11:13:55,621 HelpFormatter - Program Args: --analysis_type RealignerTargetCreator --reference_sequence /share/carvajal-archive/REFERENCE_DATA/genomes/GRCh38_decoy_LCCpanel/Homo_sapiens_assembly38_LCCpanel.fasta --intervals BED/3025671_Covered_hg38_decoyLCCpnl.pad200.bed --input_file DATA/CH-59/N2/CH-59N2.dedup.bam --out DATA/CH-59/N2/CH-59N2.realign.interval_list --validation_strictness STRICT -known /share/carvajal-archive/REFERENCE_DATA/GATK_Bundle/hg38/Mills_and_1000G_gold_standard.indels.hg38.vcf 
INFO  11:13:55,626 HelpFormatter - Executing as twtoal@carcinos on Linux 4.4.0-47-generic amd64; Java HotSpot(TM) 64-Bit Server VM 1.8.0_72-b15. 
INFO  11:13:55,626 HelpFormatter - Date/Time: 2017/07/26 11:13:55 
INFO  11:13:55,627 HelpFormatter - -------------------------------------------------------------------------------- 
[...]

PR #987, opened 2017-09-18
Fix unhelpful exception message in VCF index

PolinaBevad:
### Description
Creating constructor of class Index and getting newInstance of it (with Reflection API) causes **EOFException** if VCF index file corrupts which wrapped by **InvocationTargetException.** Description of this exception is not very useful. 
As said in [issue](https://github.com/samtools/htsjdk/issues/969) description of the error including the path to VCF file will be more helpful.
### Prerequisites
Actions with VCF index file through few editors (Idea Editor, Windows Notepad, maybe more) change some symbols in bytes of vcf index file in unrecognizable while decode-encode charset. After that changes index file is corrupted and cannot be read with **loadIndex(),** because it causes exception **InvocationTargetException.** Also corrupts of VCF index can be caused by errors while it was created with utils.
### Solution
* Create new exception: **TribbleException.CorruptedIndexFile.** 
* Now method **loadIndex()** in **IndexFactory** class catch **InvocationTargetException.** 
* If its cause is instance of **EOFException,** then method throws **TribbleException.CorruptedIndexFile** with understandable text of the error and the name of VCF file for which the error is appeared.
* If it is not instance of **EOFException** method throws **RuntimeException.** 
### Checklist
- [x] Code compiles correctly
- [x] All tests passing


yfarjoun:
this looks nice, @lbergelson can you test if this looks better in GATK when reading a vcf with a corrupt index?

Issue #978, opened 2017-09-07
VCFFileReader.fromVCF throws an NPE when input contains no CONTIG lines


cmnbroad:
You can see this using Picard VcfToIntervalList, which just calls fromVCF, on this vcf:

##fileformat=VCFv4.2
#CHROM  POS     ID      REF     ALT     QUAL    FILTER  INFO
1       100     a       G       A       232.46  PASS    .

More generally, SamFileHeader.getSequence, which is where the NPE originates, should probably handle a null sequence dictionary better and throw something more specific.

Stack (from the GATK version of VcfToIntervalList):
  java.lang.Thread.State: RUNNABLE
	  at htsjdk.samtools.SAMFileHeader.getSequence(SAMFileHeader.java:153)
	  at htsjdk.samtools.util.IntervalList.add(IntervalList.java:93)
	  at htsjdk.variant.vcf.VCFFileReader.fromVcf(VCFFileReader.java:119)
	  at htsjdk.variant.vcf.VCFFileReader.fromVcf(VCFFileReader.java:88)
	  at org.broadinstitute.hellbender.tools.picard.vcf.VcfToIntervalList.doWork(VcfToIntervalList.java:51)

cmnbroad:
Fixed via #988.

PR #988, opened 2017-09-18
VCFFileReader.fromVCF throws an NPE when input contains no CONTIG lines

EkaterinaBel:
### Description
fix for [issue](https://github.com/samtools/htsjdk/issues/978) 978: VCFFileReader.fromVCF throws an NPE when vcf input file contains no CONTIG lines. More specific message is necessary.
### Implementation
- NullPointerException was replaced to TribbleException.InvalidHeader;
- Log warning with info about absent contigs was added;
- Test for case of absent contigs was added.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


lbergelson:
@cmnbroad Could you review this since you originally opened the issue?

src/main/java/htsjdk/samtools/SAMFileHeader.java
```diff
@@ -150,6 +151,7 @@ public SAMSequenceDictionary getSequenceDictionary() {
      * Look up sequence record by name.
      */
     public SAMSequenceRecord getSequence(final String name) {
+        if (mSequenceDictionary == null) throw new TribbleException.InvalidHeader("sequence dictionary is empty");
```
cmnbroad: Having looked at this a bit more, I don't think we want to throw here. As far as I can tell its not required that you have a sequence dictionary, and the current implementation allows you to set it to null on the header, so throwing InvalidHeader is probably too strong (and we generally don't use the Tribble exceptions from within the samtools package anyway). Since it looks like there are already cases where this method is called and the return value is checked for null, I think we should just make this method return null if the sequence dictionary is null, and update the javadoc to make that explicit. Then IntervalList will throw an IllegalArgumentException instead of getting an NPE.

Also, for new code we generally like to put the if statement body on a separate line, with braces, even if its a single statement. To just return null, a ternary might be even better.

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -222,6 +228,7 @@ protected VCFHeader parseHeaderFromLines( final List<String> headerStrings, fina
             }
         }
 
+        if (!contigsExist) LOG.warn("There are no contig lines in vcf file header");
```
cmnbroad: I don't think we need to warn here; as with SAM/BAM, as far as I can tell contig lines are not required in a VCF. So I think we can revert these changes.

src/test/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
```diff
@@ -362,6 +362,13 @@ public void testVCFHeaderSerialization() throws Exception {
         Assert.assertEquals(deserializedHeader.toString(), originalHeader.toString(), "String representation of header not the same before/after serialization");
     }
 
+    @Test(expectedExceptions = TribbleException.class)
+    public void testContigsAbsenceInHeader() {
+        String vcf = "src/test/resources/htsjdk/variant/HiSeq.10000.vcf";
```
cmnbroad: I think it would be a good idea to create a separate, small test file for this case, named something like "noContigLines.vcf", that is otherwise valid, rather than relying on the fact that an existing test file happens to satisfy the criteria.

EkaterinaBel:
@cmnbroad, Thank you for your notes. Now IntervalList throws an IllegalArgumentException instead of getting an NPE. Also test file for case of contigs absence was added.

EkaterinaBel:
@cmnbroad, 
According to your requests, several changes were made: 
- javadoc was updated for getSequence method;
- tests for getSequence were added into SAMFileHeaderTest;
- test contigAbsentInHeader was renamed and moved to IntervalListTest;
- test file "IntervalListFromVCFNoContigLines.vcf" was moved to src/test/resources/htsjdk/samtools/intervallist/.

cmnbroad:
@lbergelson any objections on this ? I've reviewed it; its a pretty simple change and should be ready to merge. The only actual code change is in SamFileHeader to handle the case where getSequence is called and there is no sequence dictionary (return null instead of NPE). The rest is tests.

cmnbroad:
Thanks @EkaterinaBel !

PR #989, opened 2017-09-18
renaming VariantContextWriter.setVCFHeader -> setHeader

lbergelson:
this seems cleaner, since we've never released this change it should be
ok to change the name without a deprecation process

### Description

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


lbergelson:
@amilamanoj I'm renaming this method from your PR to be slightly less wordy.

PR #990, opened 2017-09-18
rename and fix IoUtilTest

lbergelson:
### Description

It's been bothering me that the class is IOUtil and the test is IoUtil.  Renaming it to IOUtilUnitTest.
In a second commit I fixed a bunch of issues intellij was complaining about in the test class.
### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/test/java/htsjdk/samtools/util/IOUtilTest.java
+++ b/src/test/java/htsjdk/samtools/util/IOUtilTest.java
@@ -54,7 +54,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
-public class IoUtilTest extends HtsjdkTest {
+public class IOUtilTest extends HtsjdkTest {
 
 
     private static final Path TEST_DATA_DIR = Paths.get ("src/test/resources/htsjdk/samtools/io/");
@@ -66,11 +66,11 @@ public class IoUtilTest extends HtsjdkTest {
     private static final String TEST_FILE_PREFIX = "htsjdk-IOUtilTest";
     private static final String TEST_FILE_EXTENSIONS[] = {".txt", ".txt.gz"};
     private static final String TEST_STRING = "bar!";
+
     private File existingTempFile;
     private String systemUser;
     private String systemTempDir;
-
-    private FileSystem inMemoryfileSystem;
+    private FileSystem inMemoryFileSystem;
 
     @BeforeClass
     public void setUp() throws IOException {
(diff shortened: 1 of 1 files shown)


src/test/java/htsjdk/samtools/util/IOUtilUnitTest.java
```diff
@@ -116,12 +108,12 @@ public void testGetCanonicalPath() throws IOException {
             Assert.assertEquals(IOUtil.getFullCanonicalPath(f), actual.getCanonicalPath());
         }
 
-        actual.delete();
-        symlink.delete();
-        lnToActual.delete();
-        lnToSymlink.delete();
-        lnDir.delete();
-        tmpDir.delete();
+        actual.deleteOnExit();
```
vruano: Are you trying to silence the "Not check of returned boolean value" Intellij warning by chaning from delete to deleteOnExit? I would say that if it is not a warning as compiled by gradle one should not care about that and other warnings... it shouldn't influence the code (you should silence it in Intellij if it bother you). 

In any case it seem more logic to call deleteOnExit right after you create the File as a failure in the middle would prevent the .delete() to be reached... I think you can call deleteOnExit on a not yet existent file.

vruano: If you want to silence those delete() warning you could make them part of the test by calling ```Assert.assertTrue(x.delete())``` if they are supposed exist after running the test code. If the contrary you could call ```Assert.assertFalse(x.delete())``` just the same.

src/test/java/htsjdk/samtools/util/IOUtilUnitTest.java
```diff
@@ -100,7 +92,7 @@ public void testGetCanonicalPath() throws IOException {
         }
 
         File tmpDir = new File(tmpPath, userName);
```
vruano: Since you are changing this code, can you verify that this is necessay? why we need to use the actual user name for such a test a supposed to random subdir name.

src/test/java/htsjdk/samtools/util/IOUtilUnitTest.java
lbergelson: Yeah, those yellows in intellij bother me.   That's a good point about moving them to be next to creation.  I think I'm just going to revert the change and leave them with delete though.

src/test/java/htsjdk/samtools/util/IOUtilUnitTest.java
```diff
@@ -1,397 +1,391 @@
...
+    @Test(groups = {"unix"})
+    public void testGetCanonicalPath() throws IOException {
+        String tmpPath = System.getProperty("java.io.tmpdir");
+        String userName = System.getProperty("user.name");
+
+        if (tmpPath.endsWith(userName)) {
+            tmpPath = tmpPath.substring(0, tmpPath.length() - userName.length());
+        }
+
+        File tmpDir = new File(tmpPath, userName);
```
droazen: You forgot to delete `tmpDir` on exit.

src/test/java/htsjdk/samtools/util/IOUtilUnitTest.java
```diff
@@ -1,397 +1,391 @@
...
+        tmpDir.mkdir();
+        File actual = new File(tmpDir, "actual.txt");
+        actual.deleteOnExit();
```
droazen: You should really consider introducing a `createTempFile()` method in htsjdk that auto-marks things for deletion, so that test code doesn't have to worry about it.

PR #994, opened 2017-09-21
Sam record pair identification depends on required SAM fields

DenisVerkhoturov:
### Description
Associated pull-request, please, follow [the link](https://github.com/broadinstitute/picard/pull/932) for details. 

### Checklist

- [x] Code compiles correctly
- [ ] All tests passing


src/test/java/htsjdk/samtools/SAMRecordTest.java
```diff
@@ -0,0 +1,54 @@
+package htsjdk.samtools;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class SAMRecordTest {
```
pshapiro4broad: Unfortunately it looks like `SAMRecordUnitTest` is the class where tests for `SAMRecord` are stored.

DenisVerkhoturov: Thank you, done [here](https://github.com/samtools/htsjdk/pull/994/commits/7843453cf45892d135c824eeac24264bb0db79fa).

nh13:
I am a bit time constrained at the moment. It strongly object to this PR on my first read. Can you wait until early next week for me or a maintainer to take a look?

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -334,6 +334,19 @@ private boolean hasReferenceName() {
     }
 
     /**
+     * @return {@code true} if records belong to the same pairwise alignment
+     */
+    public boolean isPair(final SAMRecord record) {
```
yfarjoun: I'm confused about this function, a mate is another record with the same query name...why are you checking everything else, and not the QNAME? I'm sure there's a reason, but it should be explained in the javadoc as it is not obvious at all!

yfarjoun:
could you explain in the description what problem this is solving? what is the purpose of this new function?

src/main/java/htsjdk/samtools/SAMRecord.java
DenisVerkhoturov: Thank you for the remark, I think it does matter in `SamRecord` and I fixed it, but if we implement it in `HitsForInsert` it doesn't matter because it deals records with the same `QNAME` only. [Here](https://github.com/broadinstitute/picard/blob/master/src/main/java/picard/sam/MultiHitAlignedReadIterator.java#L115) is the logic that produces HitsForInsert.

yfarjoun:
yes, but a public function with the name `isPair` is going to be found by users who will then make the wrong assumption. either keep is inside MergeSamFiles or make it clear from the name what is the purpose of this function.

DenisVerkhoturov:
@yfarjoun Opps, it was an accident! :sweat_smile:

DenisVerkhoturov:
@yfarjoun If we leave it in `HitsForInsert` it is a `private` [function](https://github.com/broadinstitute/picard/pull/932/files#diff-fb83641c8b09326f5028047ee001cda6R288). There is no need to make it `public` it will be used only in `HitsForInsert`, and it can be a method-local function because it is used only in `HitsForInsert#coordinateMyMate` method scope.

src/main/java/htsjdk/samtools/SAMRecord.java
magicDGS: I would suggest to move this to `SamPairUtil`, because there are already a lot of methods in `SAMRecord` that may be better to be used within an utility class. Eventually, reads would be accessed by an interface (e.g., #985) and thus this methods may live in an utility class that is implementation-independent.

DenisVerkhoturov:
Thank everybody for comments!=)
So, to recap, should I just close this pull-request and leave `arePair (SAMRecord, SAMRecord) -> boolean` function in [this request](https://github.com/broadinstitute/picard/pull/932)?

lbergelson:
@DenisVerkhoturov I think that's the right way to proceed.  Thank you for your PR and interest in the project!  We welcome future contributions.

PR #995, opened 2017-09-22
Ability to opt-out of logging all failed intervals for LiftOver

bbimber:
### Description

Liftover currently logs each time an interval fails liftOverMinMatch.  This can result in a lot of logging, and callers cannot current control this behavior.  This adds a setLogFailedIntervals(), so callers can optionally turn off this logging.  It's opt-in and will not change current behavior for any callers.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/liftover/LiftOver.java
+++ b/src/main/java/htsjdk/samtools/liftover/LiftOver.java
@@ -69,7 +69,7 @@ public class LiftOver {
     /**
      * Resets the internal counter that tracks intervals that failed liftover due to insufficient intersection length
      */
-    public void resetFailedIntervalCounter(){
+    public void resetFailedIntervalCounter() {
         this.totalFailedIntervalsBelowThreshold = 0L;
     }
 
@@ -77,7 +77,7 @@ public class LiftOver {
      *
      * @return The total number of intervals that have failed liftover due to insufficient intersection length
      */
-    public long getFailedIntervalCounter(){
+    public long getFailedIntervalCounter() {
         return totalFailedIntervalsBelowThreshold;
     }
 
@@ -154,7 +154,7 @@ public class LiftOver {
                 targetIntersection = candidateIntersection;
             } else if (candidateIntersection != null) {
                 hasOverlapBelowThreshold = true;
-                if (logFailedIntervals){
+                if (logFailedIntervals) {
                     LOG.info("Interval " + interval.getName() + " failed to match chain " + chain.id +
                             " because intersection length " + candidateIntersection.intersectionLength + " < minMatchSize "
                             + minMatchSize +
@@ -163,7 +163,7 @@ public class LiftOver {
             }
         }
         if (chainHit == null) {
-            if (hasOverlapBelowThreshold){
+            if (hasOverlapBelowThreshold) {
                 totalFailedIntervalsBelowThreshold++;
             }
 
--- a/src/main/java/htsjdk/samtools/liftover/LiftOver.java
+++ b/src/main/java/htsjdk/samtools/liftover/LiftOver.java
@@ -62,14 +62,14 @@ public class LiftOver {
      * will be logged.  Set this to false to prevent logging.
      * @param logFailedIntervals
      */
-    public void setLogFailedIntervals(boolean logFailedIntervals) {
+    public void setShouldLogFailedIntervalsBelowThreshold(boolean logFailedIntervals) {
         this.logFailedIntervals = logFailedIntervals;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/liftover/LiftOver.java
```diff
@@ -125,10 +136,12 @@ public Interval liftOver(final Interval interval, final double liftOverMinMatch)
                 chainHit = chain;
                 targetIntersection = candidateIntersection;
             } else if (candidateIntersection != null) {
-                LOG.info("Interval " + interval.getName() + " failed to match chain " + chain.id +
-                " because intersection length " + candidateIntersection.intersectionLength + " < minMatchSize "
-                + minMatchSize +
-                " (" + (candidateIntersection.intersectionLength/(float)interval.length()) + " < " + liftOverMinMatch + ")");
+                if (logFailedIntervals){
```
yfarjoun: I think that a single (warn) message should still be emitted. something like:

"there are intervals that are failing to liftOver, but logFailedIntervals is false. To see what intervals have failed, re-run with logFailedIntervals=true."

yfarjoun: although this exposes the private member logFailedIntervals....so perhaps there's a better mechanism, but I would hate for there to be no indicatino at all that there are fialing intervals. The "correct" way to avoid information like this is to run at log_level=WARNING....

bbimber:
ok, that's reasonable.  It is unfortunate that liftover doesnt really have a concept of a batch.  In the spirit of what you suggest, I think we'd want one warning per sample, or maybe even a 'XX intervals were skipped due to intersection length' type of message when the file finishes.  that would seem to be the most informative useful, and then tools like LiftoverVcf could have a command-line option to adjust logging.

I added one commit with something (granted not super elegant), to do this, and testing.  As before, this doesnt change anything about current behavior, but LiftoverVcf could be extended to take advantage of this.

yfarjoun:
the tests are failing...

bbimber:
Yes, sorry - doing too many thing at once.  This is now passing.

src/main/java/htsjdk/samtools/liftover/LiftOver.java
```diff
@@ -54,6 +54,33 @@
...
 
+    private boolean logFailedIntervals = true;
+    private long totalFailedIntervalsBelowThreshold = 0L;
+
+    /**
+     * By default any lifted interval that falls below liftOverMinMatch
+     * will be logged.  Set this to false to prevent logging.
+     * @param logFailedIntervals
+     */
+    public void setLogFailedIntervals(boolean logFailedIntervals) {
```
yfarjoun: what is this method for? 

could you replace it with "increment..." and use that function instead of ++ below?

src/main/java/htsjdk/samtools/liftover/LiftOver.java
```diff
@@ -54,6 +54,33 @@
...
+        this.logFailedIntervals = logFailedIntervals;
+    }
+
+    /**
+     * Resets the internal counter that tracks intervals that failed liftover due to insufficient intersection length
+     */
+    public void resetFailedIntervalCounter(){
```
yfarjoun: here and below, please put a space between the `()` and the `{`. 

also after the condition in an `if` statement:
```
if (condition)␣{
     expression
}
```

src/main/java/htsjdk/samtools/liftover/LiftOver.java
bbimber: I'm not sure I follow this.  The intent of this PR is to let callers of LiftOver adjust the logging behavior.  This isnt a setter for the counter itself (though I can see the confusing name choice - more of setDoLogFailedIntervals().  A different method name would be fine).  This method allows callers to turn off logging every single time an interval is rejected.  

It is not intended to be some kind of method for addFailedInterval(), which I think your second comment might be aimed toward?

src/main/java/htsjdk/samtools/liftover/LiftOver.java
bbimber: OK.  4 instances have been replaced.

src/main/java/htsjdk/samtools/liftover/LiftOver.java
yfarjoun: yes. sorry, I got confused.

yfarjoun: perhaps setShouldLogFailedIntervals ?

yfarjoun:
LGTM @lbergelson care to give it an official approval?

lbergelson:
Thank you @bbimber!

Issue #864, opened 2017-04-25
Support for bgzipped fasta indexes


magicDGS:
### Motivation
After #842, there is the possibility to create .fai indexes for a FASTA file, either bgzipped or not. Nevertheless, bgzipped files are not fully supported:

- [x] Index creation for bgzipped FASTA does not creates a .gzi index, which is a difference with respect to samtools. This index cannot be created for other bgzipped files, and will be interesting to be able to do it for any kind of file too.
- [ ] Indexed bgzipped FASTA cannot be used within HTSJDK. This is supported in samtools, but this limits the usage of compressed FASTA in toolkits such as Picard or GATK.

It will be great if this support is implemented.

yfarjoun:
this seems like a nice project....any takers?

ffinfo:
:+1:

magicDGS:
@ffinfo, there is a PR (#997) for the .gzi indexes if you wanna have a look!

magicDGS:
Update: new PR for fully supporting bgzip FASTA references (#1014)

magicDGS:
@ffinfo - the PR is in! Next release HTSJDK will have compressed FASTA support!

PR #996, opened 2017-09-24
Test for Silent bug when query a compressed indexed reference file

ffinfo:
### Description

This code is only to show the issue. Initial the tests will fail on this PR.

When executing <IndexedFastaSequenceFile>.getSubsequenceAt the method gives wrong results but not a exception that this is not yet supported, see also https://github.com/samtools/htsjdk/issues/864

I have added here a test that is currently failing.

Here there are 2 possible solutions:
- Implement https://github.com/samtools/htsjdk/issues/864 (preferred)
- Raise a exception when the method is called (workaround)

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


lbergelson:
This is really nasty.  We should probably make this throw until we can properly support fasta.gz.

cmnbroad:
This should be addressed by https://github.com/samtools/htsjdk/pull/1014.

magicDGS:
Now this is not longer an issue - in the master branch there is support for bgzip FASTA files!

lbergelson:
@ffinfo I'm going to close this because the issue is fixed and your test has been incorporated into the branch that fixed.  Thank you for bringing this to our attention.  It's always helpful to have a failing test to show off a problem.

PR #997, opened 2017-09-25
Add BlockCompressedIndex

magicDGS:
### Description

This is the first part to https://github.com/samtools/htsjdk/issues/864, by implementing the index abstraction for block-compressed files.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


magicDGS:
@yfarjoun, are you interested on looking at this?

lbergelson:
@magicDGS How is this going to be different than the existing tabix indexes which are block compressed indexes?

magicDGS:
@lbergelson - this is a general feature in the bgzip library, which it is used in samtools for parse FASTA compressed files. The general index indicates the offset (both compressed and uncompressed) where every block starts. In contrast, tabix indexes are for querying coordinate-sorted data.  I think that supporting the .gzi index is the first step to support compressed FASTA in HTSJDK.

yfarjoun:
if there's no coordinate information in gzi, so why use it for fasta?
wouldn't it make more sense to use something like tabix given that the
reference...is reference sorted?

magicDGS:
FASTA sequences in HTSJDK works with .fai indexes, which are based in bytes on file-offsets in combination with total length of the sequence and per-line bases/bytes. The .gzi index is a translation of virtual-offsets to file-offsets, and also pointing to blocks. Thus, I guess that using .gzi indexes was done for being faster while querying a sequence: query in .fai -> offset to block/file offset in bgzip -> reading the data without decompress the rest of the file.

magicDGS:
Do you mind to chime in, @ffinfo?

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
```diff
@@ -0,0 +1,248 @@
...
+            for (int i = 0; i < numberOfEntries; i++) {
+                entries[i] = new IndexEntry(buffer.getLong(), (int) buffer.getLong());
+            }
+
+            return new BlockCompressedIndex(entries);
+        }
+    }
+
+    /**
+     * Creates an index for a block-compressed file.
```
lbergelson: could this comment be expanded to make it clear that it's creating an index in memory only

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
```diff
@@ -0,0 +1,248 @@
...
+     *
+     * @param bgzipFile the bqzip file.
```
lbergelson: typo: b**q**zip

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
```diff
@@ -0,0 +1,248 @@
...
+            throws IOException {
+        if (bgzipFile == null) {
+            throw new IllegalArgumentException("null input path");
+        }
+        // open the file for reading as a block-compressed file
+        final BlockCompressedInputStream bgzipStream =
+                new BlockCompressedInputStream(Files.newInputStream(bgzipFile));
+
+        // store the entries as a list
+        final List<IndexEntry> entries = new ArrayList<>();
```
lbergelson: Should we be initializing the entries with an in memory-only 0,0 entry like they do in samtools?  (according to the issue you linked too at least)

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
```diff
@@ -0,0 +1,248 @@
...
+            if (bgzipStream.endOfBlock()) {
+                // gets the block address (compressed offset)
+                // requires to parse with the file pointer utils
+                final long compressed = BlockCompressedFilePointerUtil.getBlockAddress(bgzipStream.getFilePointer());
+                // add a new IndexEntry
+                entries.add(new IndexEntry(compressed, currentOffset));
+            }
+        }
+        // construct by converting into an array
+        return new BlockCompressedIndex(entries.toArray(new IndexEntry[entries.size()]));
```
lbergelson: should this just be stored as a `List` instead of converting back and forth from array?

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
```diff
@@ -0,0 +1,248 @@
...
+        }
+
+        final ByteBuffer buffer = allocateBuffer(entries.length, true);
+
+        // put the number of entries
+        buffer.putLong(Integer.toUnsignedLong(entries.length));
+
+        for (final IndexEntry entry : entries) {
+            // TODO - check if it is unsigned?
+            buffer.putLong(entry.getCompressedOffset());
```
lbergelson: I think we should assert that they're not negative somewhere.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
```diff
@@ -0,0 +1,248 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Represents an index of a block-compressed file (.gzi)
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
```
lbergelson: I'm a confused about the gzi index (I never use them, since htsjdk hasn't supported them...), it seems like you also need a .fai to use this index with a fasta since this index doesn't include any information about where contigs start.  Is that true?

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
```diff
@@ -0,0 +1,248 @@
...
+            // allocate array with the entries
+            final IndexEntry[] entries = new IndexEntry[numberOfEntries];
+
+            // create a new buffer with the correct size and read into it
+            buffer = allocateBuffer(numberOfEntries, false);
+            channel.read(buffer);
+            buffer.flip();
+
+            for (int i = 0; i < numberOfEntries; i++) {
+                entries[i] = new IndexEntry(buffer.getLong(), (int) buffer.getLong());
```
lbergelson: Could you add checks here that the index entries values are increasing? It would be an inexpensive check that would catch a lot of potential index corruption.  You might want to check that they're positive as well.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
```diff
@@ -0,0 +1,248 @@
...
+     *
+     * @throws IOException if an I/O error occurs.
+     */
+    public static final BlockCompressedIndex createIndex(final Path bgzipFile)
+            throws IOException {
+        if (bgzipFile == null) {
+            throw new IllegalArgumentException("null input path");
+        }
+        // open the file for reading as a block-compressed file
+        final BlockCompressedInputStream bgzipStream =
```
lbergelson: this should be in a try-with resources block so the stream gets closed automatically

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
```diff
@@ -0,0 +1,248 @@
...
+                new BlockCompressedInputStream(Files.newInputStream(bgzipFile));
+
+        // store the entries as a list
+        final List<IndexEntry> entries = new ArrayList<>();
+
+        // we need to track how many bytes we read to compute the offset
+        // because the at the end of the block, the offset is set to 0
```
lbergelson: This comment is  confusing to me.  It sounds like it's saying we need to reset the offset back to 0 at the end of each block.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
magicDGS: For the case of a block-gzipped FASTA file, the .gzi index should be used in combination with the .fai index as you say. That will be another PR to include a .gzi index in the indexed FASTA file abstraction.

Nevertheless, the .gzi index may be used in other block-compressed files for different purposes (e.g., checking block integrity). That's why I made it independent than the FASTA usage.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
magicDGS: Good point. I'll do in next commit.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
```diff
@@ -0,0 +1,248 @@
...
+     * Load the index from the provided file.
+     *
+     * @param indexPath the path for the index to load.
+     *
+     * @return loaded index.
+     *
+     * @throws IOException if an I/O error occurs.
+     */
+    public static final BlockCompressedIndex loadIndex(final Path indexPath) throws IOException {
+        // allocate a buffer for re-use for read each byte
```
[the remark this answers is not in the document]
magicDGS: Done.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
```diff
@@ -0,0 +1,248 @@
...
+        ByteBuffer buffer = ByteBuffer.allocate(Long.BYTES);
+        buffer.order(ByteOrder.LITTLE_ENDIAN);
+
+        try (final ByteChannel channel = Files.newByteChannel(indexPath)) {
+            if (Long.BYTES != channel.read(buffer)) {
+                throw new IOException("corrupted index file");
```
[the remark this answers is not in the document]
magicDGS: True, I forgot to add it. Done.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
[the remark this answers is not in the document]
magicDGS: Remaining from a previous implementation. Removed casting.

magicDGS: That's a very good indea. I'll add in next commit.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
magicDGS: Changed.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
magicDGS: True. I forgot to do it when I changed the method signature. Done now.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
magicDGS: True, I'll do in next commit. Requires also some changes in the write method.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
magicDGS: This comment is referring to a previous implementation (actually is was a self-note). The `BlockCompressedInputStream.getFilePointer()` returns a virtual file pointer which includes the block-address and block-offset, and my first thought was that the block-offset is the file-offset; nevertheless, this is a pointer to the offset within the block, so for the end of the block the returned value for `BlockCompressedInputStream.getFilePointer()` is 0 for the block-offset (beginning of the next block). My first implementation was using `BlockCompressedFilePointerUtil `.

I removed the comment for something more informative now.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
magicDGS: I though that it will be more efficient to have an array, but that's is an early optimization. I changed it for a list everywhere.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
```diff
@@ -0,0 +1,248 @@
...
+                entries.add(new IndexEntry(compressed, currentOffset));
+            }
+        }
+        // construct by converting into an array
+        return new BlockCompressedIndex(entries.toArray(new IndexEntry[entries.size()]));
+    }
+
+    // helper method for allocate a buffer for read/write
+    private static final ByteBuffer allocateBuffer(final int numberOfEntries,
+            final boolean includeNumerOfEntries) {
```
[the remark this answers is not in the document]
magicDGS: Changed.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
magicDGS: Re-factored to be named `buildIndex`, added javadoc information, and a new `createIndex` method that writes down the index file.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
```diff
@@ -0,0 +1,248 @@
...
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Represents an index of a block-compressed file (.gzi)
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+// as defined in https://github.com/samtools/htslib/issues/473
+public final class BlockCompressedIndex {
```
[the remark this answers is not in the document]
magicDGS: Renamed and added javadoc information.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
magicDGS: Done.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
magicDGS: Done.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
magicDGS: Done.

src/main/java/htsjdk/samtools/util/BlockCompressedIndex.java
magicDGS: Done.

magicDGS:
Thanks for the first pass @lbergelson. I addressed:

* 1st commit: minor comments (including refactor array to be a list).
* 2nd commit: another file for testing several blocks (FASTA)
* 3rd commit: comments about checking of offset (positive and increasing) and adding first entry only in-memory (I am not sure about this).
* 4th commit: createIndex method for in-memory and on-disk generation (syntactic -sugar).
* 5th commit: method for get a virtual offset for an uncompressed offset using the gzi-index, for re-use in the `BlockCompressedInputStream.seek` directly. I think that this is what you are referring when you talk about a "query" method. 

Looking forward to the second round of comments.

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -0,0 +1,365 @@
...
+     * block starts.
+     *
+     * @see <a href="http://github.com/samtools/htslib/issues/473">https://github.com/samtools/htslib/issues/473</a>
+     */
+    public static final class IndexEntry {
+        private final long compressedOffset;
+        private final long uncompressedOffset;
+
+        private IndexEntry(final long compressedOffset, final long uncompressedOffset) {
+            if (Long.signum(compressedOffset) == -1) {
```
lbergelson: this might be clearer as just checks for `compressedOffset < 0`

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -0,0 +1,365 @@
...
+        }
+
+        @Override
+        public int hashCode() {
+            return 31 * Long.hashCode(compressedOffset) + Long.hashCode(uncompressedOffset);
+        }
+    }
+
+    // first entry in the mapping (do not write down)
+    private static final IndexEntry FIRST_MAPPING = new IndexEntry(0, 0);
```
lbergelson: I wasn't sure if you needed this or not, I just saw it mentioned in the samtools issue.  If we don't need it for the query algorithm then we should just skip it.

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -0,0 +1,365 @@
...
+     */
+    public long getVirtualOffsetForSeek(final long uncompressedOffset) {
+        if (uncompressedOffset == 0) {
+            return BlockCompressedFilePointerUtil.makeFilePointer(0, 0);
+        }
+
+        // list of entries at the uncompressed-offset (or afterwards)
+        final IndexEntry indexEntry = entries.stream()
+                // filter for only the ones where the uncompressed offset is larger than the requested
+                .filter(entry -> entry.getUncompressedOffset() >= uncompressedOffset)
```
lbergelson: I'm confused by this.  It seems like this is finding the first block after the offset we're looking for.  Don't we want to find the last block before the offset we're looking for?   Or am I confused?  

We might want to implement this as a binary search since there could potentially be many entries in a large gzipped file?

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -0,0 +1,365 @@
...
+     *
+     * @throws IOException if an I/O error occurs.
+     */
+    public void writeIndex(final Path output) throws IOException {
+        if (output == null) {
+            throw new IllegalArgumentException("null output path");
+        }
+
+        // the first entry is never written - 0, 0
+        final int numberOfBlocksToWrite = getNumberOfBlocks() - 1;
```
lbergelson: yeah, this complicated things, maybe we should just drop it since we probably don't need it for any thing

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -0,0 +1,365 @@
...
+            if (Long.BYTES != channel.read(buffer)) {
+                throw getCorruptedIndexException(indexPath, "less than " + Long.BYTES+ "bytes");
+            }
+            buffer.flip();
+
+            final int numberOfEntries;
+            try {
+                numberOfEntries = Math.toIntExact(buffer.getLong());
+            } catch (ArithmeticException e) {
+                buffer.flip();
```
lbergelson: why bother flipping the buffer here?

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -0,0 +1,365 @@
...
+            buffer = allocateBuffer(numberOfEntries, false);
+            channel.read(buffer);
+            buffer.flip();
+
+            for (int i = 1; i <= numberOfEntries; i++) {
+                final IndexEntry entry;
+                try {
+                    entry = new IndexEntry(buffer.getLong(), buffer.getLong());
+                } catch (IllegalArgumentException e) {
+                    throw new IOException(String.format("Corrupted index file: %s (%s)",
```
lbergelson: I think it's always a good idea to pass the underlying exception in as the exception cause, even if your'e including the message in the top level exception.

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -0,0 +1,365 @@
...
+        return new IOException(String.format("Corrupted index file: %s (%s)",
+                msg,
+                indexPath == null ? "unknown" : indexPath.toUri()));
+    }
+
+    /**
+     * Builds a {@link GZIIndex} on the fly from a BGZIP file.
+     *
+     * <p>Note that this method does not write the index on disk. Use {@link #writeIndex(Path)} on
+     * the returned object to safe the index.
```
lbergelson: typo: safe -> save

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -0,0 +1,365 @@
...
+            throw new IOException("Index file " + indexFile + " already exists for " + bgzipFile);
+        }
+        // build the index, write and return
+        final GZIIndex index = buildIndex(bgzipFile);
+        index.writeIndex(indexFile);
+        return index;
+    }
+
+    /** Gets the default index path for the bgzip file. */
+    public static Path resolveIndexNameForBgzipFile(final Path bgzipFile) {
```
lbergelson: 👍 for adding this

src/test/java/htsjdk/samtools/util/GZIIndexTest.java
```diff
@@ -72,16 +72,18 @@ public void testWriteIndex(final File indexFile, final int exprectedBlocks) thro
...
                 {new File("src/test/resources/htsjdk/samtools/block_compressed.sam.gz"),
-                        new File("src/test/resources/htsjdk/samtools/block_compressed.sam.gz.gzi")}
+                        new File("src/test/resources/htsjdk/samtools/block_compressed.sam.gz.gzi")},
+                {new File("src/test/resources/htsjdk/samtools/reference/Homo_sapiens_assembly18.trimmed.fasta.gz"),
+                        new File("src/test/resources/htsjdk/samtools/reference/Homo_sapiens_assembly18.trimmed.fasta.gz.gzi")}
         };
     }
 
     @Test(dataProvider = "filesWithIndex")
     public void testCreateIndex(final File fileToIndex, final File expectedIndex) throws Exception {
```
lbergelson: this test is now misnamed, and create index isn't tested

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -0,0 +1,365 @@
...
+    /**
+     * Creates a {@link GZIIndex} from a BGZIP file and store it in memory and disk.
+     *
+     * @param bgzipFile the bgzip file.
+     * @param overwrite if the .fai index already exists override it if {@code true}; otherwise, throws a {@link IOException}.
+     *
+     * @return the in-memory representation for the created index.
+     * @throws IOException  if an IO error occurs.
+     */
+    public static GZIIndex createIndex(final Path bgzipFile, final boolean overwrite) throws IOException {
```
lbergelson: I'm not sure this method is necessary, although it could be convenient for someone.  It's not tested right now.

src/test/java/htsjdk/samtools/util/GZIIndexTest.java
```diff
@@ -72,16 +72,18 @@ public void testWriteIndex(final File indexFile, final int exprectedBlocks) thro
         return new Object[][] {
                 // TODO - requires more test files
```
lbergelson: this seems like a good set of tests files since the two cases of 1 block and multiple blocks are both tested

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -0,0 +1,365 @@
...
+                currentOffset++;
+                // if we are at the end of the block
+                if (bgzipStream.endOfBlock()) {
+                    // gets the block address (compressed offset) - requires to parse with the file pointer utils
+                    final long compressed = BlockCompressedFilePointerUtil.getBlockAddress(bgzipStream.getFilePointer());
+                    // add a new IndexEntry
+                    entries.add(new IndexEntry(compressed, currentOffset));
+                }
+            }
+            // construct by converting into an array
```
lbergelson: comment is out of date now

src/main/java/htsjdk/samtools/util/GZIIndex.java
lbergelson: This needs a test.

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -0,0 +1,365 @@
...
+     *
+     * @return index entries.
+     */
+    public List<IndexEntry> getIndexEntries() {
+        return Collections.unmodifiableList(entries);
+    }
+
+    /**
+     * Gets the virtual offset for seek with {@link BlockCompressedInputStream#seek(long)}.
+     *
```
lbergelson: Could you give this a bit more detailed explanation in the comment?

src/main/java/htsjdk/samtools/util/GZIIndex.java
magicDGS: We do not really require this in the list; if there is no index entry where the uncompressed offset is larger than the requested one, then the first block is the one where the offset is (represented by this `FIRST_MAPPING`).

src/main/java/htsjdk/samtools/util/GZIIndex.java
magicDGS: This was the first implementation without optimization (I was planning to add the binary search once I implemented tests and after your review). I will add tests and implement the binary search in the next commit.

src/main/java/htsjdk/samtools/util/GZIIndex.java
magicDGS: To show in the error message the value of the number of entries field in the index file, because the maximum that java will accept is Integer.MAX_VALUE. I am re-using the buffer here instead of storing the long value, but I can change that....

src/main/java/htsjdk/samtools/util/GZIIndex.java
magicDGS: True, and actually this should use the static method for default formatting. Changed.

src/main/java/htsjdk/samtools/util/GZIIndex.java
magicDGS: Done.

src/main/java/htsjdk/samtools/util/GZIIndex.java
magicDGS: Removed.

src/main/java/htsjdk/samtools/util/GZIIndex.java
magicDGS: I did it in a similar way that the createIndex in the FASTA code; I think that it is nice for creating the default index and do not bother about extensions.

I will add a test afterwards, when I add tests for `java.nio.Path` with Jimfs, because if not it will override the current ones.

src/test/java/htsjdk/samtools/util/GZIIndexTest.java
magicDGS: This is out of date, because at the beginning I haven't add the FASTA file. I removed the comment.

src/test/java/htsjdk/samtools/util/GZIIndexTest.java
magicDGS: Renamed test.

src/main/java/htsjdk/samtools/util/GZIIndex.java
magicDGS: Changed!

src/main/java/htsjdk/samtools/util/GZIIndex.java
magicDGS: I added two paragraphs with more details.

src/main/java/htsjdk/samtools/util/GZIIndex.java
magicDGS: And that's correct, the implementation was completely wrong for several reasons. With the binary search and the tests I debug the code and it should work now.

src/main/java/htsjdk/samtools/util/GZIIndex.java
magicDGS: I'll do in the next commit, to be sure that it does not break anything.

magicDGS:
Back to you @lbergelson - the failures are caused by the FTP and ENA tests, so they are unrelated with the PR...

magicDGS:
Rebased to solve the failing ENA tests: back to you @lbergelson!

PR #1000, opened 2017-09-28
Test that all the @DataProviders are successful 

yfarjoun:
(otherwise the dependent tests will be silently skipped)

### Description
Adds a test that makes sure that the @DataProviders are not failing. This is important since A failing data-provider causes the dependent tests to fail **silently** which means that an api change could cause parts of the code to be untested even though we have tests for them...

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/utils/ClassFinder.java
+++ b/src/main/java/htsjdk/utils/ClassFinder.java
@@ -28,13 +28,16 @@ import htsjdk.samtools.util.Log;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.Modifier;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.Enumeration;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Set;
+import java.util.stream.Collectors;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 
@@ -74,7 +77,8 @@ public class ClassFinder {
     /** Convert a filename to a class name by removing '.class' and converting '/'s to '.'s. */
     public String toClassName(final String filename) {
         return filename.substring(0, filename.lastIndexOf(".class"))
-                .replace('/', '.').replace('\\', '.');
+                .replace('/', '.')
+                .replace('\\', '.');
     }
 
     /**
@@ -188,5 +192,27 @@ public class ClassFinder {
     /** Fetches the set of classes discovered so far. */
     public Set<Class<?>> getClasses() {
         return this.classes;
+    }
+
+
+    /**
+     * Fetches the set of classes discovered so far, subsetted down to concrete (non-abstract/interface) classes only
+     *
+     * @return subset of classes discovered so far including only concrete (non-abstract/interface) classes
+     */
+    public Set<Class<?>> getConcreteClasses() {
+        return getClasses().stream()
+                .filter(ClassFinder::isConcrete)
+                .collect(Collectors.toSet());
+    }
+
+    /**
+     * Determines whether or not the specified class is concrete (ie., non-abstract and non-interface)
+     *
+     * @param clazz class to check
+     * @return true if the class is neither abstract nor an interface, otherwise false
+     */
+    public static boolean isConcrete( final Class<?> clazz ) {
+        return ! Modifier.isAbstract(clazz.getModifiers()) && ! Modifier.isInterface(clazz.getModifiers());
     }
 }
--- a/src/main/java/htsjdk/utils/ClassFinder.java
+++ b/src/main/java/htsjdk/utils/ClassFinder.java
@@ -1,0 +1,192 @@
+package htsjdk.utils;
+
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2010 The Broad Institute
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+import htsjdk.samtools.util.Log;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.util.Enumeration;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+
+
+/**
+ * Utility class that can scan for classes in the classpath and find all the ones
+ * annotated with a particular annotation.
+ *
+ * @author Tim Fennell
+ */
+public class ClassFinder {
+    private final Set<Class<?>> classes = new HashSet<Class<?>>();
+    private final ClassLoader loader;
+    private Class<?> parentType;
+    // If not null, only look for classes in this jar
+    private String jarPath = null;
+
+    private static final Log log = Log.getInstance(ClassFinder.class);
+
+    public ClassFinder() {
+        loader = Thread.currentThread().getContextClassLoader();
+    }
+
+    public ClassFinder(final ClassLoader loader) {
+        this.loader = loader;
+    }
+
+    public ClassFinder(final File jarFile) throws IOException {
+        // The class loader must have the context in order to load dependent classes when loading a class,
+        // but the jarPath is remembered so that the iteration over the classpath skips anything other than
+        // the jarPath.
+        jarPath = jarFile.getCanonicalPath();
+        final URL[] urls = {new File(jarPath).toURI().toURL()};
+        loader = new URLClassLoader(urls, Thread.currentThread().getContextClassLoader());
+    }
+
+    /** Convert a filename to a class name by removing '.class' and converting '/'s to '.'s. */
+    public String toClassName(final String filename) {
+        return filename.substring(0, filename.lastIndexOf(".class"))
+                .replace('/', '.').replace('\\', '.');
+    }
+
+    /**
+     * Scans the classpath for classes within the specified package and sub-packages that
+     * extend the parentType. This method can be called repeatedly
+     * with different packages. Classes are accumulated internally and
+     * can be accessed by calling {@link #getClasses()}.
+     */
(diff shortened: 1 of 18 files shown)


src/main/java/htsjdk/utils/ClassFinder.java
```diff
@@ -0,0 +1,192 @@
...
+                if (parentType.isAssignableFrom(type)) {
+                    this.classes.add(type);
+                }
+            }
+            catch (Throwable t) {
+                log.debug("could not load class: " + classname, t);
+            }
+        }
+    }
+
```
lbergelson: Can you add the following additional methods to this?  Then we can delete our copy in gatk.

```
    /**
     * Fetches the set of classes discovered so far, subsetted down to concrete (non-abstract/interface) classes only
     *
     * @return subset of classes discovered so far including only concrete (non-abstract/interface) classes
     */
    public Set<Class<?>> getConcreteClasses() {
        final Set<Class<?>> concreteClassSet = new LinkedHashSet<>();

        for ( final Class<?> clazz : classes ) {
            if ( isConcrete(clazz) ) {
                concreteClassSet.add(clazz);
            }
        }

        return concreteClassSet;
    }

    /**
     * Determines whether or not the specified class is concrete (ie., non-abstract and non-interface)
     *
     * @param clazz class to check
     * @return true if the class is neither abstract nor an interface, otherwise false
     */
    public static boolean isConcrete( final Class<?> clazz ) {
        return ! Modifier.isAbstract(clazz.getModifiers()) && ! Modifier.isInterface(clazz.getModifiers());
    }
```

src/main/java/htsjdk/utils/ClassFinder.java
```diff
@@ -0,0 +1,192 @@
...
+                final File file = new File(urlPath);
+                if ( file.isDirectory() ) {
+                    scanDir(file, packageName);
+                }
+                else {
+                    scanJar(file, packageName);
+                }
+            }
+            catch (IOException ioe) {
+                log.warn("could not read entries", ioe);
```
lbergelson: Would we be better off removing these catch and log statements?  It seems like they could potentially hide critical errors.  We have them in gatk as well, but I'm not sure they're the right thing to do.

src/test/java/htsjdk/TestDataProviders.java
```diff
@@ -0,0 +1,75 @@
...
+ * It is needed because in the case that a data-provider fails (for some reason, perhaps a change in some other code
+ * causes it to throw an exception) the tests that rely on it will be silently skipped.
+ * The only mention of this will be in test logs but since we normally avoid reading these logs and rely on the
+ * exit code, it will look like all the tests have passed.
+ *
+ * @author Yossi Farjoun
+ */
+public class TestDataProviders {
+
+
```
lbergelson: extra new line here

src/test/java/htsjdk/TestDataProviders.java
```diff
@@ -0,0 +1,75 @@
...
+    @Test
+    public void IndependentTestOfDataProviderTest() throws IllegalAccessException, InvocationTargetException, InstantiationException {
```
lbergelson: this should start with a lowercase letter

src/test/java/htsjdk/TestDataProviders.java
```diff
@@ -0,0 +1,75 @@
...
+        testAllDataProvidersdata();
+    }
+
+    @DataProvider(name = "DataprovidersThatDontTestThemselves")
```
lbergelson: 👍 good naming

src/test/java/htsjdk/TestDataProviders.java
```diff
@@ -0,0 +1,75 @@
...
+    public Iterator<Object[]> testAllDataProvidersdata() throws IllegalAccessException, InstantiationException, InvocationTargetException {
```
lbergelson: The capitalization here looks weird, should we have 2 capital Datas?   
image

src/test/java/htsjdk/TestDataProviders.java
```diff
@@ -0,0 +1,75 @@
...
+
+        List<Object[]> data = new ArrayList<>();
+        final ClassFinder classFinder = new ClassFinder();
+        classFinder.find("htsjdk", Object.class);
```
lbergelson: should we restrict this to subclasses of HtsjdkTest?

lbergelson: Now that we have this in htjsjdk, maybe we should also test that all `@Test` annotations are in a subclass of `HtsjdkTest` since they will fail to run otherwise.  `MetaTestTest.java`

src/test/java/htsjdk/TestDataProviders.java
```diff
@@ -0,0 +1,75 @@
...
+        // make sure that this @DataProvider is in the list
+        Assert.assertEquals(data.stream().filter(c -> ((Method) c[0]).getName().equals("testAllDataProvidersdata")).count(), 1);
+
+        return data.iterator();
+    }
+
+    // @NoInjection annotations required according to this test:
+    // https://github.com/cbeust/testng/blob/master/src/test/java/test/inject/NoInjectionTest.java
+    @Test(dataProvider = "DataprovidersThatDontTestThemselves")
+    public void testDataProviderswithDP(@NoInjection final Method method, final Class clazz) throws IllegalAccessException, InstantiationException, InvocationTargetException {
```
lbergelson: I see you figured out that weird issue you were having...

src/test/java/htsjdk/TestDataProviders.java
```diff
@@ -0,0 +1,75 @@
...
+
+        System.err.println("Method: " + method + " Class: " + clazz.getName());
```
lbergelson: do we really want to spam the names of all of our test classes here?

src/test/java/htsjdk/TestDataProviders.java
```diff
@@ -0,0 +1,75 @@
...
+
+        Object instance = clazz.newInstance();
```
lbergelson: Are all of our test classes guaranteed a no-arg constructor?

src/test/java/htsjdk/TestDataProviders.java
```diff
@@ -0,0 +1,75 @@
...
+
+        // Some tests assume that the @BeforeSuite methods will be called before the @DataProviders
+        for (final Method otherMethod : clazz.getMethods()) {
```
lbergelson: Don't we have the same problem with `BeforeTest`, `BeforeClass`, etc?

src/test/java/htsjdk/TestDataProviders.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @DataProvider(name = "DataprovidersThatDontTestThemselves")
+    public Iterator<Object[]> testAllDataProvidersdata() throws IllegalAccessException, InstantiationException, InvocationTargetException {
+
+        List<Object[]> data = new ArrayList<>();
+        final ClassFinder classFinder = new ClassFinder();
+        classFinder.find("htsjdk", Object.class);
+
+        for (final Class<?> testClass : classFinder.getClasses()) {
```
lbergelson: you could use `getConcreteClasses` here if we add it I think

src/test/java/htsjdk/TestDataProviders.java
```diff
@@ -0,0 +1,75 @@
...
+            if (Modifier.isAbstract(testClass.getModifiers())) continue;
+            for (final Method method : testClass.getMethods()) {
```
lbergelson: this will only work with public dataproviders, but maybe that's not an issue, I'm not sure if TestNG works with private dataproviders...

src/test/java/htsjdk/TestDataProviders.java
```diff
@@ -0,0 +1,75 @@
...
+        Object instance = clazz.newInstance();
+
+        // Some tests assume that the @BeforeSuite methods will be called before the @DataProviders
+        for (final Method otherMethod : clazz.getMethods()) {
+            if (otherMethod.isAnnotationPresent(BeforeSuite.class)) {
+                otherMethod.invoke(instance);
+            }
+        }
+
+        method.invoke(instance);
```
lbergelson: I think we should make sure that if this throws it has a very clear error message which explains what this test is, what exact DataProvider it was calling, and what the exception thrown was.  Otherwise people are going to be mystified by it.

src/main/java/htsjdk/utils/ClassFinder.java
yfarjoun: I copied it verbatim from Picard (which is probably how it got into GATK...) so I would argue that it's out of scope for this PR....do you agree?

src/test/java/htsjdk/TestDataProviders.java
yfarjoun: that's a great idea! I just implemented and found many classes in violation! hurray!

yfarjoun: (including the one I just added!!!)

yfarjoun: Also added test for non public methods (found a few...)

src/test/java/htsjdk/TestDataProviders.java
yfarjoun: yup. good idea.

src/test/java/htsjdk/TestDataProviders.java
yfarjoun: can check...prolly a good idea.

src/test/java/htsjdk/TestDataProviders.java
yfarjoun: removed.

src/test/java/htsjdk/TestDataProviders.java
yfarjoun: no, just the data-providers.

src/test/java/htsjdk/TestDataProviders.java
yfarjoun: we are running the providers, not the tests...so yes with before class, but not before method or before test....I think

src/test/java/htsjdk/TestDataProviders.java
yfarjoun: ok. I'll see what I can do.

src/test/java/htsjdk/TestDataProviders.java
yfarjoun: hmmm. according to http://testng.org/doc/documentation-main.html I just need to call @Before Suite. Everything else will be called before the tests, not the DataProvider.

yfarjoun:
@lbergelson, one of the SRA tests failed...it seems. But could you take a look at the changes I made and see if they are to your liking?

src/main/java/htsjdk/utils/ClassFinder.java
yfarjoun: done.

yfarjoun:
I'm mystified as to why tests are failing here...

yfarjoun:
The tests are failing since the ENA is no longer accepting http requests, only https requests...I submitted a PR to resolve this. #1004

yfarjoun:
@lbergelson I modified the code to include private functions, as they are seen by TestNG. of the two instances, the failing one is the PUSH, while the PR is passing, reflecting the fact that the original master was failing when I created the branch. I could rebase but don't really see the point...

more comments?

src/main/java/htsjdk/utils/ClassFinder.java
lbergelson: Agreed, more of a thought for the future.

src/test/java/htsjdk/TestClassDependenceTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.*;
+
+/**
+ * Created by farjoun on 10/2/17.
+ */
+public class TestClassDependenceTest extends HtsjdkTest {
```
lbergelson: @yfarjoun Sorry to keep nitpicking this branch, but this class really needs a comment explaining what it's testing.

src/test/java/htsjdk/utils/TestNGUtils.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Small class implementing some utility functions that are useful for test and interfacing with the TestNG framework.
+ */
+public class TestNGUtils {
+
+    /** A Method that returns all the Methods that are annotated with @DataProvider
+     * in a given package.
+     *
+     * @param pakkage the package under hwich to look for classes and methods
```
lbergelson: typo: **hw**ich

PR #1002, opened 2017-10-01
Fix BufferedLineReader peek() at end of line with Windows line endings.

Lenbok:
### Description

LineNumberReader.read() compresses multiple-character line terminators into a single \n character. Thus, a single-character mark is insufficient in the peek() method, resulting in "Mark invalid" exceptions being thrown when a Windows line ending is encountered. The existing unit tests have been extended to check that peek() does not throw, and these fail when using the original 1 character mark.

### Checklist

- [X] Code compiles correctly
- [X] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/BufferedLineReader.java
+++ b/src/main/java/htsjdk/samtools/util/BufferedLineReader.java
@@ -25,7 +25,6 @@ package htsjdk.samtools.util;
 
 import htsjdk.samtools.Defaults;
 
-import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -80,7 +79,7 @@ public class BufferedLineReader extends LineNumberReader implements LineReader {
     @Override
     public int peek() {
         try {
-            mark(1);
+            mark(2); // Two characters required here as the next read will collapse \r\n to a single \n
             final int ret = read();
             reset();
             return ret;
--- a/src/test/java/htsjdk/samtools/util/BufferedLineReaderTest.java
+++ b/src/test/java/htsjdk/samtools/util/BufferedLineReaderTest.java
@@ -71,10 +71,12 @@ public class BufferedLineReaderTest {
             input = "";
         }
         final BufferedLineReader blr = BufferedLineReader.fromString(input);
+        blr.peek();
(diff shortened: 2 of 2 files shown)


Lenbok:
@lbergelson Status check? This should be pretty straight forward, right?

src/main/java/htsjdk/samtools/util/BufferedLineReader.java
```diff
@@ -80,7 +79,7 @@ public String readLine() {
     @Override
     public int peek() {
         try {
-            mark(1);
```
lbergelson: Could you add a comment here explaining this, it will be confusing for future generations otherwise.

src/test/java/htsjdk/samtools/util/BufferedLineReaderTest.java
```diff
@@ -71,10 +71,12 @@ private void fromStringEmptyTestHelper(final String terminator, final boolean la
             input = "";
         }
         final BufferedLineReader blr = BufferedLineReader.fromString(input);
+        blr.peek();
```
lbergelson: Instead of adding these peek calls in the existing tests, would it be possible to add a new test that includes a file with windows line endings?  Ideally a test would fail on all machines without your change, instead of just on windows ones.

src/main/java/htsjdk/samtools/util/BufferedLineReader.java
Lenbok: Good idea.

src/test/java/htsjdk/samtools/util/BufferedLineReaderTest.java
Lenbok: I was pleasantly surprised to see that the test was already explicitly testing different terminators (see TERMINATORS), it was just not exercising peek(). That is, the test does fail on non-windows machines when the mark size is 1 (I dev on linux). So an external file seems redundant, right?

lbergelson: Hmn, that's a good point.  I don't think I was paying that much attention to what was being tested exactly.

lbergelson:
@Lenbok Looks good, thanks for this.  I'll merge when tests pass.

PR #1004, opened 2017-10-04
Fixed ENA url to use https (instead of http) 

yfarjoun:
### Description

- It seems that the ENA has moved to https which is why tests are failing. 
- This probably also affected cram access to ena resources. fixed.
- Reformated tests to use a data provider

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/cram/ref/EnaRefService.java
+++ b/src/main/java/htsjdk/samtools/cram/ref/EnaRefService.java
@@ -16,6 +16,8 @@ public class EnaRefService {
     private static final int HTTP_NOT_FOUND = 404;
     private static final int HTTP_INTERNAL_SEVER_PROBLEM = 500;
     private static final int HTTP_CONNECTION_TIMEOUT = 522;
+    private static final int HTTP_MOVED_PERMANENTLY = 301;
+
 
     byte[] getSequence(final String md5) throws GaveUpException {
         final int restBetweenTries_ms = 0;
@@ -43,7 +45,7 @@ public class EnaRefService {
         if (!md5.matches("[a-z0-9]{32}"))
             throw new RuntimeException("Does not look like an md5 checksum: " + md5);
 
-        final String httpEbiString = "http://www.ebi.ac.uk/ena/cram/md5/%s";
+        final String httpEbiString = "https://www.ebi.ac.uk/ena/cram/md5/%s";
         final String urlString = String.format(httpEbiString, md5);
         final URL url;
         try {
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/cram/ref/EnaRefService.java
```diff
@@ -75,6 +77,12 @@
                     case HTTP_CONNECTION_TIMEOUT:
                     case HTTP_INTERNAL_SEVER_PROBLEM:
                         break;
+                    case HTTP_MOVED_PERMANENTLY:
+                        log.error("It seems that the base URL for the ENA service has changed permanently. Got error:" + code +
```
vruano: It seems that this failure is recoverable as the response header/body contains the new URL... Perhaps to much work considering that these changes should not happen that often.

src/main/java/htsjdk/samtools/cram/ref/EnaRefService.java
```diff
@@ -16,6 +16,8 @@
     private static final int HTTP_NOT_FOUND = 404;
     private static final int HTTP_INTERNAL_SEVER_PROBLEM = 500;
     private static final int HTTP_CONNECTION_TIMEOUT = 522;
+    private static final int HTTP_MOVED_PERMANENTLY = 301;
```
vruano: It seems that ```HttpURLConnection``` defined most if not all this constants since you are changing this bit perhaps you could go ahead and use those instead of defining our own.

src/main/java/htsjdk/samtools/cram/ref/EnaRefService.java
yfarjoun: I don't want to automatically follow a random URL that was given....sounds like not very secure behaviour

yfarjoun:
failures are coverage issues due to non-determinism.....

src/main/java/htsjdk/samtools/cram/ref/EnaRefService.java
```diff
@@ -11,12 +11,13 @@
 
 public class EnaRefService {
     private static final Log log = Log.getInstance(EnaRefService.class);
-    private static final int HTTP_OK = 200;
+    private static final int HTTP_OK = HttpURLConnection.HTTP_OK;
     private static final int HTTP_FOUND = 302;
```
vruano: Please remove HTTP_FOUND which is never used and beside it seems to correspond to MOVED_TEMP; I don't know what the "FOUND" actually means here.

src/main/java/htsjdk/samtools/cram/ref/EnaRefService.java
```diff
@@ -11,12 +11,13 @@
...
-    private static final int HTTP_NOT_FOUND = 404;
-    private static final int HTTP_INTERNAL_SEVER_PROBLEM = 500;
+    private static final int HTTP_NOT_FOUND = HttpURLConnection.HTTP_NOT_FOUND;
+    private static final int HTTP_INTERNAL_SEVER_PROBLEM = HttpURLConnection.HTTP_INTERNAL_ERROR;
+    // this is a non-standard code and I'm not sure why it's here
     private static final int HTTP_CONNECTION_TIMEOUT = 522;
```
vruano: This one is interesting (522), it is not a standard one... looking at the way is used with HTTP_INTERNAL_SVER_PROBLEM I wonder whether that case means to handle all the 5xx codes.

Is this one the reason why you prefer to redefined this constants in this class rather than use HttpURLConnection's directly?

vruano:
Ah it was merged already... whatever 👍

yfarjoun:
Sorry, @vruano I thought you had already approved......

vruano:
Yeah sorry, my bad.... I thought it was still open otherwise I wouldn't bother is not important.

PR #1005, opened 2017-10-05
Add SAMFileWriterFactory methods for java.nio.Path referenceFile

magicDGS:
### Description

More `java.nio.Path` support for `SAMFileWriterFactory` methods, in this case for supporting referenceFile paths for CRAM.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMFileWriterFactory.java
+++ b/src/main/java/htsjdk/samtools/SAMFileWriterFactory.java
@@ -473,21 +473,6 @@ public class SAMFileWriterFactory implements Cloneable {
 
     /**
      *
-     * Create a SAM, BAM or CRAM writer based on examination of the outputFile extension.
-     *
-     * @param header header. Sort order is determined by the sortOrder property of this arg.
-     * @param presorted if true, SAMRecords must be added to the SAMFileWriter in order that agrees with header.sortOrder.
-     * @param outputFile where to write the output.  Must end with .sam, .bam or .cram.
-     * @param referenceFasta reference sequence file
-     * @return SAMFileWriter appropriate for the file type specified in outputFile
-     *
-     */
-    public SAMFileWriter makeWriter(final SAMFileHeader header, final boolean presorted, final File outputFile, final Path referenceFasta) {
-        return makeWriter(header, presorted, null == outputFile ? null : outputFile.toPath(), referenceFasta);
-    }
-
-    /**
-     *
      * Create a SAM, BAM or CRAM writer based on examination of the outputPath extension.
      *
      * @param header header. Sort order is determined by the sortOrder property of this arg.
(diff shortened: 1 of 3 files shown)


src/main/java/htsjdk/samtools/SAMFileWriterFactory.java
```diff
@@ -468,6 +468,36 @@ public SAMFileWriter makeWriter(final SAMFileHeader header, final boolean presor
...
+     * Create a SAM, BAM or CRAM writer based on examination of the outputFile extension.
+     *
+     * @param header header. Sort order is determined by the sortOrder property of this arg.
+     * @param presorted if true, SAMRecords must be added to the SAMFileWriter in order that agrees with header.sortOrder.
+     * @param outputFile where to write the output.  Must end with .sam, .bam or .cram.
+     * @param referenceFasta reference sequence file
+     * @return SAMFileWriter appropriate for the file type specified in outputFile
+     *
+     */
+    public SAMFileWriter makeWriter(final SAMFileHeader header, final boolean presorted, final File outputFile, final Path referenceFasta) {
```
lbergelson: Do we really want to add another mixed file/path one?  It seems unnecessary unless I'm missing a use case.

src/main/java/htsjdk/samtools/SAMFileWriterFactory.java
```diff
@@ -556,6 +601,71 @@ public CRAMFileWriter makeCRAMWriter(final SAMFileHeader header, final boolean p
...
+     *
+     * Note: does not honor factory settings for USE_ASYNC_IO.
+     *
+     * @param header entire header. Sort order is determined by the sortOrder property of this arg.
+     * @param outputFile where to write the output.  Must end with .sam, .bam or .cram.
+     * @param referenceFasta reference sequence file
+     * @return CRAMFileWriter
+     *
+     */
+    public CRAMFileWriter makeCRAMWriter(final SAMFileHeader header, final File outputFile, final Path referenceFasta) {
```
lbergelson: same here, I don't think I understand the use case for this one

src/main/java/htsjdk/samtools/SAMFileWriterFactory.java
```diff
@@ -556,6 +601,71 @@ public CRAMFileWriter makeCRAMWriter(final SAMFileHeader header, final boolean p
...
+     * Note: does not honor factory setting for USE_ASYNC_IO.
+     *
+     * @param header entire header. Sort order is determined by the sortOrder property of this arg.
+     * @param presorted  if true, SAMRecords must be added to the SAMFileWriter in order that agrees with header.sortOrder.
+     * @param outputFile where to write the output.  Must end with .sam, .bam or .cram.
+     * @param referenceFasta reference sequence file
+     * @return CRAMFileWriter
+     *
+     */
+    public CRAMFileWriter makeCRAMWriter(final SAMFileHeader header, final boolean presorted, final File outputFile, final Path referenceFasta) {
```
lbergelson: and here

src/main/java/htsjdk/samtools/SAMFileWriterFactory.java
magicDGS: Removed.

src/main/java/htsjdk/samtools/SAMFileWriterFactory.java
magicDGS: Removed.

src/main/java/htsjdk/samtools/SAMFileWriterFactory.java
magicDGS: Removed.

magicDGS:
@lbergelson - I added that methods for symmetry, but it is true that there is no usage case (from my point of view).

Back to you!

lbergelson:
@magicDGS Ack, there's a compilation failure because there are now ambiguous calls where a null used to unambiguously be a File, but now it can be either a file or a path. Just have to cast that to File and then good to merge 👍 Thank you!

cmnbroad:
I'd like to get this merged in, but I'd lobby that we not add new public overloads of makeCRAMWriter, since they're redundant with makeWriter (this PR brings the number of makeCRAMWriter overloads to 8). makeWriter should (in my opinion) be the canonical way forward.

magicDGS:
@cmnbroad - I agree with you that the canonical should be the `makeWriter()` and to create an specific SAM/BAM/CRAM (I think that this is useful from the programmatically point of view to create files in a concrete format independently of the extension).

Maybe we can reduce the overloads of the SAM/BAM/CRAM `java.nio.Path` to the full ones, to make sure that they are used with all params or use the `makeWriter` versions.

magicDGS:
Any chance to have a look to this soon, @cmnbroad  and/or @lbergelson? Thanks!

magicDGS:
Friendly ping here - @lbergelson or @cmnbroad? Thanks!

magicDGS:
I added a simple test based on another one - back to you @lbergelson!

lbergelson:
@magicDGS Thank you!

magicDGS:
Thank you for review!

PR #1007, opened 2017-10-06
Defining <NON_REF> allele in htsjdk, where it belongs.

yfarjoun:
### Description

We are currently defining the <NON_REF> allele in GATK (3 & 4) and Picard....it seems more reasonable to do it here in htsjdk.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/Allele.java
+++ b/src/main/java/htsjdk/variant/variantcontext/Allele.java
@@ -132,6 +132,9 @@ public class Allele implements Comparable<Allele>, Serializable {
     /** A generic static SPAN_DEL allele for use */
     public final static String SPAN_DEL_STRING = "*";
 
+    /** A generic static NON_REF allele for use */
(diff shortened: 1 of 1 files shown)


droazen:
(apart from the travis failure, that is :)

yfarjoun:
the travis failure is due to the change in ENA url. if you review that PR #1004 we unblock htsjdk which will have master fail tests now.

Lenbok:
Is <NON_REF> part of the VCF specification now? If it's something that only GATK uses, perhaps it doesn't really belong here. Or on the other hand, perhaps formalize it's meaning by introducing it to the VCF spec.

yfarjoun:
At this point, I've added it since I want picard to be able to parse
g.vcf's...I agree that adding it to the vcf spec is a good idea, but note
that alleles do not have to be formally added to the vcf spec in order to
be used.

PR #1009, opened 2017-10-09
Add out/ to .gitignore

magicDGS:
### Description

Building with Intellij 2017.2 generates an /out directory instead of
using the build/ folder from gradle. This commit remove this directory
from the tracked files.

This folder cannot be changed, even if the compiler output is changed (see CrazyCoder comment in https://stackoverflow.com/questions/45174989/building-with-intellij-2017-2-out-directory-duplicates-files-in-build-director).

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ src/htsjdk.iml
 *.iml
 *.ipr
 *.iws
+out/
 
 
 


lbergelson:
@magicDGS 👍 Thank you.

PR #1010, opened 2017-10-10
VariantContextFilter extends Predicate<VariantContext>

lindenb:
### Description

small simplication : `VariantContextFilter` is just a  `Predicate<VariantContext>`.

We can now use this in a `stream().filter(ctxFilter)`

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/filter/VariantContextFilter.java
+++ b/src/main/java/htsjdk/variant/variantcontext/filter/VariantContextFilter.java
@@ -24,20 +24,14 @@
 package htsjdk.variant.variantcontext.filter;
 
 import htsjdk.variant.variantcontext.VariantContext;
+import java.util.function.Predicate;
 
 /**
  *
- * API for filtering VariantContexts
+ * API for filtering VariantContexts.
  *
  * @author Yossi Farjoun
  *
  */
-public interface VariantContextFilter {
-    /**
-     * Determines whether a VariantContext matches this filter
-     *
-     * @param record the VariantContext to evaluate
-     * @return true if the VariantContext matches the filter, otherwise false
-     */
-    boolean test(VariantContext record);
+public interface VariantContextFilter extends Predicate<VariantContext> {
 }


magicDGS:
:+1 to functional interfaces!

yfarjoun:
LGTM.

👍 

I'll give it some time for others to comment before merging.

src/main/java/htsjdk/variant/variantcontext/filter/VariantContextFilter.java
```diff
@@ -24,20 +24,14 @@
...
  */
-public interface VariantContextFilter {
-    /**
-     * Determines whether a VariantContext matches this filter
-     *
-     * @param record the VariantContext to evaluate
-     * @return true if the VariantContext matches the filter, otherwise false
-     */
-    boolean test(VariantContext record);
+public interface VariantContextFilter extends Predicate<VariantContext> {
```
pshapiro4broad: This is OK but note that it's not required as you can use the original as `stream.filter(f::test)`. But this form lets you write `stream.filter(f)` which is a bit shorter.

lbergelson:
@lindenb This is fine, but I think it has issues when using the functional combinators, ex: `or`,`and`,`negate` I.e. the functional combination operations don't produce a result of type `VariantContextFilter`.  This is really a problem with java's type system, but it's kind of gross.  In some ways because of that issue it might just be better to use `Predicate<VariantContext>` and deprecate this class. 

See gatk's [VariantFilter](https://github.com/broadinstitute/gatk/blob/master/src/main/java/org/broadinstitute/hellbender/engine/filters/VariantFilter.java) which includes specializations of some of the functions that produce values of the correct type.  (although I just noticed a bunch of copy and paste typos in there from when we duplicated it from ReadFilter...)

lbergelson:
Merging this, it's an improvement even if it doesn't fix all the flaws with java's type system..

lbergelson:
Thanks @lindenb

PR #1015, opened 2017-10-19
Add mark support to SeekableStream with a new class

magicDGS:
### Description

Currently, every `SeekableStream` does not have mark support. This PR adds it by a wrapper class, that is also used in factory methods to allow marking the stream and reset by using seek.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/seekablestream/MarkSeekableStream.java
+++ b/src/main/java/htsjdk/samtools/seekablestream/MarkSeekableStream.java
@@ -1,0 +1,125 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Daniel Gomez-Sanchez
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package htsjdk.samtools.seekablestream;
+
+import java.io.IOException;
+import java.util.Optional;
+import java.util.OptionalLong;
+
+/**
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class MarkSeekableStream extends SeekableStream {
+
+    private final SeekableStream underlyingStream;
+    private OptionalLong mark = OptionalLong.empty();
+
+    public MarkSeekableStream(final SeekableStream underlyingStream) {
+        if (underlyingStream == null) {
+            throw new IllegalArgumentException("null stream");
+        }
+        this.underlyingStream = underlyingStream;
+    }
+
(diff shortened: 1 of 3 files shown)


src/main/java/htsjdk/samtools/seekablestream/MarkSeekableStream.java
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * Mark the current position of the stream.
+     *
+     * <p>Note: there is not limit for reading.
+     *
+     * @param readlimit ignored.
+     */
+    @Override
+    public synchronized void mark(int readlimit) {
```
yfarjoun: needs testing....

src/main/java/htsjdk/samtools/seekablestream/MarkSeekableStream.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Override
+    public String getSource() {
+        return underlyingStream.getSource();
+    }
+
+    /**
+     * Mark the current position of the stream.
+     *
+     * <p>Note: there is not limit for reading.
```
yfarjoun: not->no

src/main/java/htsjdk/samtools/seekablestream/MarkSeekableStream.java
```diff
@@ -0,0 +1,125 @@
...
+package htsjdk.samtools.seekablestream;
+
+import java.io.IOException;
+import java.util.Optional;
+import java.util.OptionalLong;
+
+/**
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class MarkSeekableStream extends SeekableStream {
```
yfarjoun: One of us is confused about the use of `this` (implicitly in 115 and 117) and `underlyingStream` (everywhere else).

It seems simpler (and more correct) to remove the underlyingStream alltogether, not override most of the methods, and keep the last three methods....

Also, I think that some tests are needed...

magicDGS: For wrappers, I tend to use the wrap object for methods that are expected to do not change, and `this` for the methods that change some behaviour: like that, it is always using the current class method even if it changes in the future (e.g., bug fix) and if it is not what is supposed to do, it would pop up while testing another change.

I do not understand what you mean by remove the override of the rest of the methods: the `SeekableStream` is an interface and as a wrapper should implement the layaout. If you are referring to make `MarkSeekableStream` and abstract class and make all the rest of the implementations extend this, it was also one of my ideas; nevertheless, I didn't want to be intrusive in the code and add the mark object to every instance. If mark is not necessary, it is not required to keep.

Regarding the tests, I definitely need to implement some. I just wanted to have some feedback before it, to do not waste my time in refactoring the tests if something like my previous proposal (all `SeekableStreams` supporting mark) is required...

src/main/java/htsjdk/samtools/seekablestream/MarkSeekableStream.java
magicDGS: Done.

src/main/java/htsjdk/samtools/seekablestream/MarkSeekableStream.java
magicDGS: Agree. But I would like to know first how to proceed with the implementation: make all `SeekableStream`s have mark support, or keep the current implementation and test a simple wrapper around a byte array...

src/main/java/htsjdk/samtools/seekablestream/MarkSeekableStream.java
yfarjoun: Like I said "one of us is confused"...it was me. sorry.

yfarjoun: Seems better to have this wrapper than to convert all SeekableStreams to MarkSeekableStreams..

magicDGS:
@yfarjoun - back to you with some simple tests...

src/main/java/htsjdk/samtools/seekablestream/MarkSeekableStream.java
cmnbroad: I'm not sure I see the advantage of the wrapper - it would be much less complicated to just implement mark in SeekableStream. The overhead of the storing the mark position itself seems pretty minimal.

magicDGS: @cmnbroad - if you think that it is ok to do so, that was my first idea. Nevertheless, I opted for this instead to do not change the behaviour of every seekable stream in case it is important to do not have this support...

magicDGS: I'll submit a mutually exclusive PR with the idea...

magicDGS: See #1016 for a clean implementation without a wrapper.

lbergelson:
@magicDGS I'm closing this one because I believe it's redundant with the #1016 which we just merged.  Please reopen if I'm confused about that.

magicDGS:
No worries, it is redundant. The other was the better option, but I open first this one as a proof-of-concept without breaking changes...

PR #1016, opened 2017-10-23
Add mark support for SeekableStream

magicDGS:
### Description

`SeekableStream` is a random access implementation of `InputStream`, which conceptually is able to support marking in any case. This PR adds mark support to the abstract class `SeekableStream`, to allow every implementation to support marks.

In addition, it improves the seekable stream documentation.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/seekablestream/SeekableStream.java
+++ b/src/main/java/htsjdk/samtools/seekablestream/SeekableStream.java
@@ -26,8 +26,15 @@ package htsjdk.samtools.seekablestream;
 import java.io.EOFException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.OptionalLong;
 
 public abstract class SeekableStream extends InputStream {
+
+    /**
+     * If the stream is marked with {@link #mark(int)} this represents the {@link #position()}
+     * where the stream was; otherwise, this is empty.
+     */
+    protected OptionalLong mark = OptionalLong.empty();
 
     public abstract long length();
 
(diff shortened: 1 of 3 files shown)


magicDGS:
@cmnbroad and @yfarjoun - here is a substitution of #1015

src/main/java/htsjdk/samtools/seekablestream/SeekableStream.java
```diff
@@ -68,4 +86,38 @@ public void readFully(byte b[]) throws IOException {
...
+     * Mark the current position of the stream.
+     *
+     * <p>Note: there is no limit for reading.
+     *
+     * @param readlimit ignored.
+     */
+    @Override
+    public final synchronized void mark(int readlimit) {
+        try {
+            mark = OptionalLong.of(position());
```
yfarjoun: in reviewing this we were concerned about the scenario where position legitimately throws an exception causing mark to remain un-updated. In this cas, due to the catch, it will lead to a silent error.

There are several ways around this:
1. throw an unchecked exception instead
2. look at the specific exceptions thrown by current implementations and work with these
3. modify contract for position() to not throw, and simultaneously modify known implementations.

if you have other solutions, please discuss.

magicDGS: I opted for a combination of 1 and 2: if the exception is a `ClosedChannelException`, there's no effect at all (as the contract suggests); otherwise, it throws an `RuntimeIOException` for converting into an unchecked exception.

magicDGS:
@yfarjoun - back to you after a while without taking a look to this...

yfarjoun:
@lbergelson could you take a look please?

lbergelson:
@magicDGS Thanks for the contribution and for you patience!

PR #1018, opened 2017-10-26
Enable VcfFileReader and other classes to use Path

yfarjoun:
### Description

- Enables the use of Path, and therefore NIO implementations such as google-cloud-nio when opening VCF readers. (though the google-cloud-nio library is not added) 

- Cleans up ununsed test files and some white-space/java8 issues

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


yfarjoun:
hmmm. all the tests of actually reading from GCS are failing...this is despite the buckets I've setup having allUsers ObjectReader roles....is there something we can do to enable a blank machine to read from an open bucket?

jacarey:
An open bucket seems dangerous to me. That means anyone can download htsjdk and read from the bucket and potentially rack up costs for the Broad... whether by accident or maliciously.

yfarjoun:
it's actually "worse", as you do not even need to download htsjdk to download these files...the bucket is totally open.

I asked Eric about it and he isn't concerned....but right now it isn't working even for the tests (which is why they are failing in Travis, but succeeded on my computer...

yfarjoun:
BTW, the files are tiny, and our egress charges are quite small.

build.gradle
```diff
@@ -43,6 +43,8 @@ dependencies {
     testRuntime 'org.pegdown:pegdown:1.4.2' // Necessary for generating HTML reports with ScalaTest
     testCompile "org.testng:testng:6.9.9"
     testCompile "com.google.jimfs:jimfs:1.1"
+    // https://mvnrepository.com/artifact/com.google.cloud/google-cloud-nio
+    testCompile group: 'com.google.cloud', name: 'google-cloud-nio', version: '0.5.0'
```
magicDGS: I would use `jimfs` instead of google-cloud for testing a different filesystem. Like that, you can avoid the issues with your GCS costs...

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -889,6 +901,45 @@ public static String readFully(final InputStream in) {
...
 
+
+    /**
+     * Returns an iterator over the lines in a text file. The underlying resources are automatically
+     * closed when the iterator hits the end of the input, or manually by calling close().
+     *
+     * @param p a {@link Path} that is to be read in as text
+     * @return an iterator over the lines in the file
+     */
+    public static IterableOnceIterator<String> readLines(final Path p) {
```
magicDGS: Could be `Files.lines(Path)` used instead of creating a new utility method?

src/test/java/htsjdk/variant/vcf/VCFPathReaderTest.java
```diff
@@ -0,0 +1,61 @@
...
+                // various ways to refer to a local file
+                {"src/test/resources/htsjdk/variant/VCF4HeaderTest.vcf", false, true},
+
+                {Paths.get("").toAbsolutePath().toString() + "/src/test/resources/htsjdk/variant/VCF4HeaderTest.vcf", false, true},
+                {"file://" + Paths.get("").toAbsolutePath().toString() + "/src/test/resources/htsjdk/variant/VCF4HeaderTest.vcf", false, true},
+
+                //testing GCS files:
+
+                // this is almost a vcf, but not quite it's missing the #CHROM line and it has no content...
+                {"gs://htsjdk-testdata/htsjdk/variant/Homo_sapiens_assembly38.tile_db_header.vcf", false, false},
```
magicDGS: Tests for other FS could use Jimfs: copy the local files into Jimfs, and use the filesystem for the test to create the strings...

yfarjoun:
had to rebase to resolve a conflit. Kept all the original commits

src/test/java/htsjdk/variant/vcf/VCFPathReaderTest.java
yfarjoun: done

build.gradle
yfarjoun: done

src/main/java/htsjdk/samtools/util/IOUtil.java
yfarjoun: done

yfarjoun:
@magicDGS @jacarey  I fixed things as per your comments...care to take another glance?

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -1064,6 +1076,60 @@ public static String slurp(final InputStream is, final Charset charSet) {
...
+                }
+            }
+
+            // If the file didn't match a given extension, treat it as a list of files
+            if (!matched) {
+                IOUtil.assertFileIsReadable(p);
+
+                try {
+                    Files.lines(p).forEach(s -> {
+                        if (!s.trim().isEmpty()) {
```
pshapiro4broad: instead of calling `String.trim()` multiple times here use `map()`, e.g.
```
Files.lines(p).map(String::trim).forEach(...)
```

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -1064,6 +1076,60 @@ public static String slurp(final InputStream is, final Charset charSet) {
...
+    public static List<Path> unrollPaths(final Collection<Path> inputs, final String... extensions) {
+        if (extensions.length < 1) throw new IllegalArgumentException("Must provide at least one extension.");
+
+        final Stack<Path> stack = new Stack<>();
+        final List<Path> output = new ArrayList<>();
+        stack.addAll(inputs);
+
+        while (!stack.empty()) {
+            final Path p = stack.pop();
+            final String name = p.toUri().toString();
```
pshapiro4broad: Why call `.toUri().toString()` here? Seems like `p.toString()` would produce the same result.

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -1064,6 +1076,60 @@ public static String slurp(final InputStream is, final Charset charSet) {
     }
 
     /**
+     * Go through the files provided and if they have one of the provided file extensions pass the file into the output
+     * otherwise assume that file is a list of filenames and unfold it into the output.
```
pshapiro4broad: This method also recursively expands files of filenames, if one of the filenames in a file of filenames points to another file of filenames. If this is the intended behavior it should be documented and tested.

src/main/java/htsjdk/variant/vcf/VCFPathReader.java
```diff
@@ -0,0 +1,183 @@
...
+ */
+public class VCFPathReader implements Closeable, Iterable<VariantContext> {
+
+    private final FeatureReader<VariantContext> reader;
+
+    /**
+     * Returns true if the given path appears to be a BCF file.
+     */
+    public static boolean isBCF(final Path path) {
+        return path.toUri().getRawPath().endsWith(".bcf");
```
pshapiro4broad: Why use `.toUri()` here and below? In this case you could write `path.toString().endsWIth(".bcf")`

src/main/java/htsjdk/variant/vcf/VCFPathReader.java
```diff
@@ -0,0 +1,183 @@
...
+        final SAMFileHeader samFileHeader = new SAMFileHeader();
+        samFileHeader.setSequenceDictionary(dict);
+        final IntervalList list = new IntervalList(samFileHeader);
+
+        int intervals = 0;
+        for (final VariantContext vc : vcf) {
+            if (includeFiltered || !vc.isFiltered()) {
+                String name = vc.getID();
+                final Integer intervalEnd = vc.getCommonInfo().getAttributeAsInt("END", vc.getEnd());
+                if (".".equals(name) || name == null)
```
pshapiro4broad: Need `{` `}` around `if` body. Also IMO it's more natural to write `if (name == null || name.equals("."))`

src/main/java/htsjdk/variant/vcf/VCFPathReader.java
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * Returns an iterator over all records in this VCF/BCF file.
+     */
+    @Override
+    public CloseableIterator<VariantContext> iterator() {
+        try {
+            return reader.iterator();
+        } catch (final IOException ioe) {
+            throw new TribbleException("Could not create an iterator from a feature reader.", ioe);
```
pshapiro4broad: Why is this code using `TribbleException` if it's not part of `htsjdk.tribble`? If you want an unchecked IO Exception you can use `UncheckedIOException`.

src/test/java/htsjdk/samtools/util/IoUtilTest.java
```diff
@@ -46,17 +48,18 @@
...
 import java.util.List;
 
 public class IoUtilTest extends HtsjdkTest {
 
-    private static final File SLURP_TEST_FILE = new File("src/test/resources/htsjdk/samtools/io/slurptest.txt");
-    private static final File EMPTY_FILE = new File("src/test/resources/htsjdk/samtools/io/empty.txt");
-    private static final File FIVE_SPACES_THEN_A_NEWLINE_THEN_FIVE_SPACES_FILE = new File("src/test/resources/htsjdk/samtools/io/5newline5.txt");
+
+    private static final File TEST_DATA_DIR = new File ("src/test/resources/htsjdk/samtools/io/");
+    private static final File SLURP_TEST_FILE = new File(TEST_DATA_DIR,"slurptest.txt");
```
pshapiro4broad: missing space after `,`, this occurs many places in this checkin

src/test/java/htsjdk/samtools/util/IoUtilTest.java
```diff
@@ -46,17 +48,18 @@
...
 import java.util.Collections;
 import java.util.List;
 
 public class IoUtilTest extends HtsjdkTest {
 
-    private static final File SLURP_TEST_FILE = new File("src/test/resources/htsjdk/samtools/io/slurptest.txt");
-    private static final File EMPTY_FILE = new File("src/test/resources/htsjdk/samtools/io/empty.txt");
-    private static final File FIVE_SPACES_THEN_A_NEWLINE_THEN_FIVE_SPACES_FILE = new File("src/test/resources/htsjdk/samtools/io/5newline5.txt");
+
+    private static final File TEST_DATA_DIR = new File ("src/test/resources/htsjdk/samtools/io/");
```
pshapiro4broad: Would be better to use `Path` here if possible.

src/test/java/htsjdk/samtools/util/IoUtilTest.java
```diff
@@ -394,4 +397,27 @@ public void testAssertDirectoryIsWritableFile(final File file, final boolean wri
             }
         }
     }
+    @DataProvider
+    public Object[][] fofnData() throws IOException {
+        Path fofnPath1 = inMemoryfileSystem.getPath("Level1.fofn");
+        Files.copy(new File(TEST_DATA_DIR.getAbsolutePath(),"Level1.fofn").toPath(), fofnPath1);
```
pshapiro4broad: This should be `TEST_DATA_DIR.getPath().resolve("Level1.fofn")`

src/main/java/htsjdk/samtools/util/IOUtil.java
yfarjoun: I don't think it does. it drops the schema.

src/main/java/htsjdk/samtools/util/IOUtil.java
yfarjoun: is tested, will update message.

src/main/java/htsjdk/variant/vcf/VCFPathReader.java
yfarjoun: because I wanted to remove the possible junk after a ? in a URL.

src/main/java/htsjdk/variant/vcf/VCFPathReader.java
yfarjoun: this is a copy of the VCFFileReader...I made very few changes....I think unless we want a serious rewrite of htsjdk (which is out of scope for this PR) I'd vote for keeping the current (slightly annoying) behavior.

src/test/java/htsjdk/samtools/util/IoUtilTest.java
yfarjoun: OK...though it makes things a bit more cumbersome

src/test/java/htsjdk/samtools/util/IoUtilTest.java
yfarjoun: done.

src/test/java/htsjdk/samtools/util/IoUtilTest.java
yfarjoun: TIL. thanks.

src/test/java/htsjdk/samtools/util/IoUtilTest.java
yfarjoun: Oh. nice. thanks.

src/main/java/htsjdk/tribble/util/ParsingUtils.java
```diff
@@ -100,16 +101,16 @@ public static InputStream openInputStream(String path)
...
      */
-    public static InputStream openInputStream(String path, Function<SeekableByteChannel, SeekableByteChannel> wrapper)
+    public static InputStream openInputStream(final String path, final Function<SeekableByteChannel, SeekableByteChannel> wrapper)
             throws IOException {
 
         final InputStream inputStream;
         if (path.startsWith("http:") || path.startsWith("https:") || path.startsWith("ftp:")) {
             inputStream = getURLHelper(new URL(path)).openInputStream();
-        } else if (IOUtil.hasScheme(path)) {
+        } else if (IOUtil.hasScheme(path) && !IOUtil.getScheme(path).equals("file")) {
```
magicDGS: Maybe get the scheme before, and use it for checking http, ftp, etc...

src/main/java/htsjdk/tribble/util/ParsingUtils.java
```diff
@@ -100,16 +101,16 @@ public static InputStream openInputStream(String path)
...
             inputStream = new SeekablePathStream(IOUtil.getPath(path), wrapper);
         } else {
-            File file = new File(path);
+            File file = IOUtil.getScheme(path) != null && IOUtil.getScheme(path).equals("file") ? new File(path.substring("file:".length())) : new File(path);
```
magicDGS: This branch can be used with the SeekablePathStream too, no?

src/main/java/htsjdk/variant/vcf/VCFFileReader.java
```diff
@@ -16,74 +16,84 @@
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Path;
 
 /**
  * Simplified interface for reading from VCF/BCF files.
  */
 public class VCFFileReader implements Closeable, Iterable<VariantContext> {
```
magicDGS: deprecate in favor of the Path version?

yfarjoun: sure.

src/main/java/htsjdk/tribble/util/ParsingUtils.java
yfarjoun: the problem, I think, is that a file can represent a named pipe, which is not seekable.

src/main/java/htsjdk/samtools/util/IOUtil.java
pshapiro4broad: There is no schema for a `Path`. Anyway the code is testing the end of the path, so it doesn't matter what the schema is.

src/main/java/htsjdk/tribble/util/ParsingUtils.java
```diff
@@ -100,16 +101,16 @@ public static InputStream openInputStream(String path)
      * @return
      * @throws IOException
      */
-    public static InputStream openInputStream(String path, Function<SeekableByteChannel, SeekableByteChannel> wrapper)
+    public static InputStream openInputStream(final String path, final Function<SeekableByteChannel, SeekableByteChannel> wrapper)
```
pshapiro4broad: Argument `path` should be named `uri` since it's not a path but a URI. Making the caller create a URI would be helpful too.

The javadoc for this could be improved too, it should have full sentences for its description text. And omitting the text for a `@return` or `@throws` is not helpful; either remove the `@` form from the javadoc or add the missing documentation.

src/main/java/htsjdk/variant/vcf/VCFPathReader.java
pshapiro4broad: In what case does a `Path` have a query string? A `Path` is not a URI and a query string would be invalid, as far as I know.

src/main/java/htsjdk/variant/vcf/VCFPathReader.java
```diff
@@ -0,0 +1,183 @@
...
+    public static boolean isBCF(final Path path) {
+        return path.toUri().getRawPath().endsWith(".bcf");
+    }
+
+    /**
+     * Returns the SAMSequenceDictionary from the provided VCF file.
+     */
+    public static SAMSequenceDictionary getSequenceDictionary(final Path path) {
+        final SAMSequenceDictionary dict = new VCFPathReader(path, false).getFileHeader().getSequenceDictionary();
+        CloserUtil.close(path);
```
pshapiro4broad: What is this trying to do? `Path` is not closeable and has no `close()` method.

src/main/java/htsjdk/variant/vcf/VCFPathReader.java
```diff
@@ -0,0 +1,183 @@
...
+    public VCFPathReader(final Path path, final Path indexPath) {
+        this(path, indexPath, true);
+    }
+
+    /**
+     * Allows construction of a VCFFileReader that will or will not assert the presence of an index as desired.
+     */
+    public VCFPathReader(final Path path, final boolean requireIndex) {
+        // Note how we deal with type safety here, just casting to (FeatureCodec)
+        // in the call to getFeatureReader is not enough for Java 8.
```
pshapiro4broad: This is because the compiler can't infer the common type that both branches of the ternary op produces and the function accepts. If you cast using the parameterized type you can inline this expression, e.g. `(FeatureCodec<VariantContext, ?>) (isBCF(path) ? new BCF2Codec() : new VCFCodec())` works.

pshapiro4broad: It looks like `isBCF()` is only called in two places and in both places it's used to determine which codec to return. If so you can avoid the code duplication and and replace with `isBCF()` with `getCodecForPath(Path)`. Or keep `isBCF()` and write:
```
    private static FeatureCodec<VariantContext, ?> getCodecForPath(Path path) {
        return isBCF(path) ? new BCF2Codec() : new VCFCodec();
    }
```
This also eliminates the need for a cast or introducing a local variable.

src/main/java/htsjdk/variant/vcf/VCFPathReader.java
```diff
@@ -0,0 +1,183 @@
...
+     * @return
+     */
+    public static IntervalList fromVcf(final Path path) {
+        return fromVcf(path, false);
+    }
+
+    public static IntervalList fromVcf(final Path path, final boolean includeFiltered) {
+        final VCFPathReader vcfReader = new VCFPathReader(path, false);
+        final IntervalList intervalList = fromVcf(vcfReader, includeFiltered);
+        vcfReader.close();
```
pshapiro4broad: Using try-with-resources would eliminate the need to call `close()` here

src/main/java/htsjdk/variant/vcf/VCFPathReader.java
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * Converts a vcf to an IntervalList. The name field of the IntervalList is taken from the ID field of the variant, if it exists. If not,
+     * creates a name of the format interval-n where n is a running number that increments only on un-named intervals
+     * Will use a "END" tag in the info field as the end of the interval (if exists).
+     *
+     * @param vcf the vcfReader to be used for the conversion
+     * @return an IntervalList constructed from input vcf
+     */
+
+    public static IntervalList fromVcf(final VCFPathReader vcf) {
```
pshapiro4broad: In general a static method on object X that also accepts an argument of type X should be converted to a non-static method. E.g. this should be written as:
```
public IntervalList fromVcf() {
    return vcf.fromVcf(false);
}
```
Although it's unclear to me if the caller would know that this is going to produce an interval list. Normally you see static `fromX()` methods to convert object type X to another type. So a better name here might be `toIntervalList()`.

src/main/java/htsjdk/variant/vcf/VCFPathReader.java
pshapiro4broad: OK.

src/main/java/htsjdk/variant/vcf/VCFPathReader.java
yfarjoun: I think that path could have a query string...but I could be wrong.

src/main/java/htsjdk/tribble/util/ParsingUtils.java
yfarjoun: hmmm. I don't think that changing the API is justified here....I'll develop the docs though

src/main/java/htsjdk/variant/vcf/VCFPathReader.java
yfarjoun: not sure what I was thinking there....thanks.

src/main/java/htsjdk/variant/vcf/VCFPathReader.java
yfarjoun: done.

src/main/java/htsjdk/variant/vcf/VCFPathReader.java
yfarjoun: good idea

yfarjoun:
back to you @pshapiro4broad

src/main/java/htsjdk/tribble/util/ParsingUtils.java
```diff
@@ -95,22 +88,27 @@ public static InputStream openInputStream(String path)
...
-            inputStream = new SeekablePathStream(IOUtil.getPath(path), wrapper);
-        } else {
-            File file = new File(path);
+        final Set<String> URL_SCHEMES = new HashSet<>(Arrays.asList("http","ftp","https"));
+
+        if (URL_SCHEMES.contains(scheme)) {
+            inputStream = getURLHelper(new URL(uri)).openInputStream();
+        } else if (scheme.isEmpty()) {
+            File file = new File(uri);
             inputStream = new FileInputStream(file);
```
lindenb: when possible use `Files.newInputStream` instead of `new FileInputStream(` ( see https://www.cloudbees.com/blog/fileinputstream-fileoutputstream-considered-harmful )

yfarjoun: ok.

src/main/java/htsjdk/tribble/util/ParsingUtils.java
```diff
@@ -95,22 +88,27 @@ public static InputStream openInputStream(String path)
...
+        final String scheme = Optional.ofNullable(IOUtil.getScheme(uri)).orElse("");
+
         final InputStream inputStream;
-        if (path.startsWith("http:") || path.startsWith("https:") || path.startsWith("ftp:")) {
-            inputStream = getURLHelper(new URL(path)).openInputStream();
-        } else if (IOUtil.hasScheme(path)) {
-            inputStream = new SeekablePathStream(IOUtil.getPath(path), wrapper);
-        } else {
-            File file = new File(path);
+        final Set<String> URL_SCHEMES = new HashSet<>(Arrays.asList("http","ftp","https"));
```
pshapiro4broad: Should make this a private static to avoid creating every time the method is called.

src/main/java/htsjdk/variant/utils/SAMSequenceDictionaryExtractor.java
```diff
@@ -111,7 +144,10 @@ SAMSequenceDictionary extractDictionary(final File intervalList) {
             applicableExtensions = extensions;
         }
 
+        /** @deprecated in favor of extractDictionary(final Path file) */
```
pshapiro4broad: Better to use `@link` here so it shows up as a hyperlink, e.g. 
```
/**
  * @deprecated use  {@link #extractDictionary(Path)} instead
  */
```

src/main/java/htsjdk/variant/utils/SAMSequenceDictionaryExtractor.java
```diff
@@ -111,7 +144,10 @@ SAMSequenceDictionary extractDictionary(final File intervalList) {
...
+        @Deprecated
         abstract SAMSequenceDictionary extractDictionary(final File file);
```
pshapiro4broad: Wouldn't it be simpler to make this non-abstract and implement it as
```
return extractDictionary(file.toPath());
```
?

src/main/java/htsjdk/variant/utils/SAMSequenceDictionaryExtractor.java
```diff
@@ -75,30 +85,53 @@ SAMSequenceDictionary extractDictionary(final File dictionary) {
...
-                VCFFileReader vcfFileReader = null;
+                VCFPathReader vcfReader = null;
                 try {
-                    vcfFileReader = new VCFFileReader(vcf, false);
-                    return vcfFileReader.getFileHeader().getSequenceDictionary();
+                    vcfReader = new VCFPathReader(vcf.toPath(), false);
+                    return vcfReader.getFileHeader().getSequenceDictionary();
                 } finally {
-                    CloserUtil.close(vcfFileReader);
+                    CloserUtil.close(vcfReader);
```
pshapiro4broad: Can you replace this `close()` with try-with-resources?

src/main/java/htsjdk/variant/utils/SAMSequenceDictionaryExtractor.java
```diff
@@ -75,30 +85,53 @@ SAMSequenceDictionary extractDictionary(final File dictionary) {
...
+            @Override
+            SAMSequenceDictionary extractDictionary(Path sam) {
+                return SamReaderFactory.makeDefault().getFileHeader(sam).getSequenceDictionary();
+            }
         },
         VCF(IOUtil.VCF_EXTENSIONS) {
             @Override
             SAMSequenceDictionary extractDictionary(final File vcf) {
-                VCFFileReader vcfFileReader = null;
+                VCFPathReader vcfReader = null;
```
pshapiro4broad: Personally I don't think renaming the class is necessary. Even though the API now accepts/requires `Path` it is still reading a VCF File. The implementation detail of how a file is represented has changed but the functionality hasn't.

src/main/java/htsjdk/variant/utils/SAMSequenceDictionaryExtractor.java
```diff
@@ -124,6 +160,17 @@ static TYPE forFile(final File dictionaryExtractable) {
             throw new SAMException("Cannot figure out type of file " + dictionaryExtractable.getAbsolutePath() + " from extension. Current implementation understands the following types: " + Arrays.toString(TYPE.values()));
         }
 
+        static TYPE forPath(final Path dictionaryExtractable) {
```
pshapiro4broad: The enum should really be called `Type` as uppercase is typically reserved for static final objects.

src/main/java/htsjdk/variant/utils/SAMSequenceDictionaryExtractor.java
yfarjoun: done.

src/main/java/htsjdk/variant/utils/SAMSequenceDictionaryExtractor.java
yfarjoun: sorry...that ship has sailed....I can file it under "things to do in htsjdk3"...

src/main/java/htsjdk/variant/utils/SAMSequenceDictionaryExtractor.java
yfarjoun: done

src/main/java/htsjdk/variant/utils/SAMSequenceDictionaryExtractor.java
yfarjoun: done.

src/main/java/htsjdk/variant/utils/SAMSequenceDictionaryExtractor.java
yfarjoun: done.

src/main/java/htsjdk/samtools/util/IOUtil.java
yfarjoun: sure.

yfarjoun:
Tests now pass... @jacarey ? @magicDGS ? @pshapiro4broad ? any more comments?

src/main/java/htsjdk/tribble/util/ParsingUtils.java
```diff
@@ -90,35 +82,36 @@ public static InputStream openInputStream(String path)
...
-        if (path.startsWith("http:") || path.startsWith("https:") || path.startsWith("ftp:")) {
-            inputStream = getURLHelper(new URL(path)).openInputStream();
-        } else if (IOUtil.hasScheme(path)) {
-            inputStream = new SeekablePathStream(IOUtil.getPath(path), wrapper);
+
+        if (URL_SCHEMES.stream().anyMatch(uri::startsWith)) {
+            inputStream = getURLHelper(new URL(uri)).openInputStream();
+        } else if (scheme.isEmpty()) {
+            File file = new File(uri);
+            inputStream = Files.newInputStream(file.toPath());
```
magicDGS: Will this handle the case of named-pipes, as @yfarjoun commented in my previous comment? In addition, is there any way to differentiate the pipes from files? If the scheme is empty and it is a `File`, it feels wrong to me that the method does not return a `SeekablePathStream` and it does if the scheme is "file:".

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -142,6 +142,9 @@ public SamReader open(final Path path,
     /** Utility method to open the file get the header and close the file */
     abstract public SAMFileHeader getFileHeader(File samFile);
 
+    /** Utility method to open the file get the header and close the file */
+    abstract public SAMFileHeader getFileHeader(Path samFile);
```
magicDGS: Maybe HTSJDK should start deprecating the `File` versions of the methods in favor of the `Path` versions. Anyway, for users using `File` the change is as simple as using `File.toPath` if there are no differences in the methods...

src/main/java/htsjdk/variant/vcf/VCFFileReader.java
```diff
@@ -16,105 +16,275 @@
...
-                      requireIndex);
-	}
+    /**
+     * Allows construction of a VCFFileReader that will or will not assert the presence of an index as desired.
+     *
+     * @deprecated in favor of {@link VCFFileReader##VCFFileReader(Path, boolean)}
+     */
+    @Deprecated
+    public VCFFileReader(final File file, final boolean requireIndex) {
+        // Note how we deal with type safety here, just casting to (FeatureCodec)
```
magicDGS: Maybe it is better to call `this(file.toPath(), requireIndex)` instead of doing the magic in place (it should result in the same reader).

src/main/java/htsjdk/variant/vcf/VCFFileReader.java
```diff
@@ -16,105 +16,275 @@
...
-      // in the call to getFeatureReader is not enough for Java 8.
-      FeatureCodec<VariantContext, ?> codec = isBCF(file) ? new BCF2Codec() : new VCFCodec();
-      this.reader = AbstractFeatureReader.getFeatureReader(
-                      file.getAbsolutePath(),
-                      indexFile.getAbsolutePath(),
-                      codec,
-                      requireIndex);
+        // Note how we deal with type safety here, just casting to (FeatureCodec)
+        // in the call to getFeatureReader is not enough for Java 8.
+        FeatureCodec<VariantContext, ?> codec = isBCF(file) ? new BCF2Codec() : new VCFCodec();
```
magicDGS: Same as before.

src/test/java/htsjdk/samtools/sra/SRALazyRecordTest.java
```diff
@@ -32,9 +32,9 @@ public void testSerialization(final SRAAccession accession) throws Exception {
         Assert.assertEquals(deserializedSAMRecord, initialSAMRecord, "Deserialized SAMRecord not equal to original SAMRecord");
     }
 
-    @Test
-    public void testCloneAndEquals() throws Exception {
-        final SRAFileReader reader = new SRAFileReader(DEFAULT_ACCESSION);
+    @Test(dataProvider = "serializationTestData")
```
magicDGS: Does this correspond to this PR?

src/test/java/htsjdk/samtools/util/IoUtilTest.java
```diff
@@ -394,4 +398,27 @@ public void testAssertDirectoryIsWritableFile(final File file, final boolean wri
             }
         }
     }
+    @DataProvider
+    public Object[][] fofnData() throws IOException {
+        Path fofnPath1 = inMemoryfileSystem.getPath("Level1.fofn");
+        Files.copy(TEST_DATA_DIR.resolve("Level1.fofn"), fofnPath1);
+
+        Path fofnPath2 = inMemoryfileSystem.getPath("Level2.fofn");
+        Files.copy(TEST_DATA_DIR.resolve("Level2.fofn"), fofnPath2);
```
magicDGS: Strings "Level1.fofn" and "Level2.fofn" can be factor out (static or within the test) to re-use here and in the returned array.

src/test/java/htsjdk/tribble/TestUtils.java
```diff
@@ -18,11 +18,30 @@
...
 
 /**
  * User: jacob
  * Date: 2012-Dec-13
  */
 public class TestUtils {
     public static String DATA_DIR = "src/test/resources/htsjdk/tribble/";
+
+
+    public static Path getTribbleFileInJimfs(String vcf, String index, FileSystem fileSystem) throws IOException, URISyntaxException {
```
magicDGS: Add a javadoc explaining that the `FileSystem` should be a `JimfsFileSystem`. It took me a while to figure out what was it.

Another option is to provide a destination `JimfsPath` instead to get the `FileSystem` from it.

src/test/java/htsjdk/tribble/TestUtils.java
```diff
@@ -18,11 +18,30 @@
...
+        final FileSystem fs = fileSystem;
+        final Path root = fs.getPath("/");
+        final Path vcfPath = Paths.get(vcf);
+        if (index != null) {
+            final Path idxPath = Paths.get(index);
+            final Path idxDestination = Paths.get(AbstractFeatureReader.isTabix(vcf, index) ? Tribble.tabixIndexFile(vcf) : Tribble.indexFile(vcf));
+            Files.copy(idxPath, root.resolve(idxDestination.getFileName().toString()));
```
magicDGS: You can resolve with the root the __vcfPath__ for the `JimfsFileSystem` before, and use the Tribble methods for get the idxDestination name without the need to resolve it again against the root.

src/test/java/htsjdk/variant/utils/SAMSequenceDictionaryExtractorTest.java
```diff
@@ -48,7 +48,6 @@
                 new Object[]{"Homo_sapiens_assembly18.trimmed.fasta", "Homo_sapiens_assembly18.trimmed.dict"},
                 new Object[]{"test2_comp.interval_list", "Homo_sapiens_assembly18.trimmed.dict"},
                 new Object[]{"ScreenSamReads.100.input.sam", "test3_comp.interval_list"},
-                new Object[]{"ScreenSamReads.100.input.sam", "test4_comp.interval_list"},
```
magicDGS: Why is this test case removed?

src/test/java/htsjdk/variant/vcf/VCFPathReaderTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.nio.file.FileSystem;
+import java.nio.file.Path;
+
+/**
+ * Created by farjoun on 10/12/17.
+ */
+public class VCFPathReaderTest extends HtsjdkTest {
```
magicDGS: Now that you unified the classes, this should be named `VCFFileReaderTest`

src/test/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
```diff
@@ -48,6 +49,7 @@
 import java.io.PrintWriter;
 import java.io.StringReader;
 import java.math.BigInteger;
+import java.nio.file.Path;
```
magicDGS: Changes in this file are not longer need it.

src/main/java/htsjdk/samtools/SamReaderFactory.java
yfarjoun: I think we should probably only start doing that when we have full support of Path....

src/main/java/htsjdk/tribble/util/ParsingUtils.java
yfarjoun: unclear.... @lbergelson ?

src/main/java/htsjdk/variant/vcf/VCFFileReader.java
yfarjoun: done.

src/main/java/htsjdk/variant/vcf/VCFFileReader.java
yfarjoun: done.

src/test/java/htsjdk/samtools/sra/SRALazyRecordTest.java
yfarjoun: no....but I noticed it, so I fixed it...it requires a provider but wasn't given one....

src/test/java/htsjdk/samtools/util/IoUtilTest.java
yfarjoun: done.

src/test/java/htsjdk/tribble/TestUtils.java
yfarjoun: Thanks. that's a bit simpler.

src/test/java/htsjdk/variant/utils/SAMSequenceDictionaryExtractorTest.java
yfarjoun: not sure...putting it back.

src/test/java/htsjdk/variant/vcf/VCFPathReaderTest.java
yfarjoun: thanks.

yfarjoun:
@magicDGS responded to your comments.

@jacarey @tfenne @lbergelson, this needs a directive from the maintainers I think...

magicDGS:
👍 for my point of view.

yfarjoun:
@lbergelson ?

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -1063,6 +1075,61 @@ public static String slurp(final InputStream is, final Charset charSet) {
         return output;
     }
 
+    /**
+     * Go through the files provided and if they have one of the provided file extensions pass the file to the output
+     * otherwise assume that file is a list of filenames and unfold it into the output (recursively).
+     */
+    public static List<Path> unrollPaths(final Collection<Path> inputs, final String... extensions) {
```
lbergelson: @yfarjoun Is there a reason we can't replace the file version with a call to the new path code?

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -1101,6 +1181,20 @@ public static Path getPath(String uriString) throws IOException {
         }
     }
 
+    public static List<Path> getPaths(List<String> uriStrings) throws RuntimeException {
```
lbergelson: this is untested

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -1101,6 +1181,20 @@ public static Path getPath(String uriString) throws IOException {
...
+        return uriStrings.stream().map(s -> {
+            try {
+                return IOUtil.getPath(s);
+            } catch (IOException e) {
+                throw new RuntimeException(e);
```
lbergelson: RuntimeIOException maybe?

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -1101,6 +1181,20 @@ public static Path getPath(String uriString) throws IOException {
...
+            }
+        }).collect(Collectors.toList());
+    }
+
+    public static List<Path> promoteFiles(List<File> files){
```
lbergelson: This needs a better name.  maybe `filesToPaths`?   It's also untested and undocumented.

src/main/java/htsjdk/tribble/util/ParsingUtils.java
```diff
@@ -90,35 +82,36 @@ public static InputStream openInputStream(String path)
...
      * open an input stream from the given path and wrap the raw byte stream with a wrapper if given
      *
      * the wrapper will only be applied to paths that are not http, https, ftp, or file, i.e. any {@link java.nio.file.Path}
      * using a custom filesystem plugin
-     * @param path a uri like string
+     * @param uri a uri like string
      * @param wrapper to wrap the input stream in, may be used to implement caching or prefetching, etc
-     * @return
-     * @throws IOException
+     * @return An inputStream appropriately created from uri and conditionally wrapped with wrapper (only in certain cases
```
lbergelson: 👍  for updating the doc.  missing a closing `)`

src/main/java/htsjdk/tribble/util/ParsingUtils.java
lbergelson: I assume named pipes can be referred to as either a file or a path, which means that we'd need a different mechanism to check for them.  We don't ever do a check for `isRegularFile` which is I think what we'd need to do.

src/main/java/htsjdk/tribble/util/ParsingUtils.java
```diff
@@ -90,35 +82,36 @@ public static InputStream openInputStream(String path)
...
+     * @throws IOException when stream cannot be opened against uri
      */
-    public static InputStream openInputStream(String path, Function<SeekableByteChannel, SeekableByteChannel> wrapper)
+    public static InputStream openInputStream(final String uri, final Function<SeekableByteChannel, SeekableByteChannel> wrapper)
             throws IOException {
 
+        // will be empty string if there's no scheme
+        final String scheme = Optional.ofNullable(IOUtil.getScheme(uri)).orElse("");
```
lbergelson: `getScheme` doesn't return a null, there's no need for this complex dance.  It's also definitely preferable to just have a ternary expression that checks for null instead of creating a new nullable just to check if it's empty.

src/main/java/htsjdk/tribble/util/ParsingUtils.java
lbergelson: I'm not sure how much we care if we return a SeekablePathStream for a file, vs a simple input stream.  They should have similar performance I hope, and this function only guarantees an input stream.   It is weird though to return different types depending on how you write the path though and it probably has some weird knock on effects. That said, it's not a change from how it was set up before is it?  

Why did you change from the previous order and using hasScheme? It seems like you added a new local variable and some additional logic and then only use it to do what the simpler check accomplished?

src/main/java/htsjdk/variant/vcf/VCFFileReader.java
```diff
@@ -16,105 +16,266 @@
...
-
-    /** Constructs a VCFFileReader with a specified index. */
+    private final FeatureReader<VariantContext> reader;
+
+    /**
+     * Returns true if the given file appears to be a BCF file.
+     *
+     * @deprecated in favor of {@link VCFFileReader#isBCF(Path)}
+     */
+    @Deprecated
```
lbergelson: We've typically been avoiding deprecating the file methods so far, but it would be nice to move away from them.  Probably fine to do so here.

lbergelson: That said, if we deprecate them, we should remove uses of them except for direct tests.  Lots of test code is using the `File` constructors.

src/main/java/htsjdk/samtools/util/IOUtil.java
yfarjoun: well, it returns a List<Path> instead of a List<File>...

src/main/java/htsjdk/samtools/util/IOUtil.java
yfarjoun: added

src/main/java/htsjdk/samtools/util/IOUtil.java
yfarjoun: ok

src/main/java/htsjdk/samtools/util/IOUtil.java
yfarjoun: done.

src/main/java/htsjdk/tribble/util/ParsingUtils.java
yfarjoun: look at getScheme:

```    *
     * @return  The scheme component of this URI,
     *          or {@code null} if the scheme is undefined
     */
```

yfarjoun: if I use a ternary, I need to either add a line of code (to save the result of getSceme) or call it twice....I'll do it if you insist, but I thought that this is a better solution.

src/main/java/htsjdk/variant/vcf/VCFFileReader.java
yfarjoun: That seems like a larger effort than this PR is meant to do...reversing the deprecations...

yfarjoun: Perhaps this should be on the roadmap for htsjdk3?

yfarjoun:
not sure what you're asking me to do about the `isRegularFile`...can you elaborate?

yfarjoun:
another look @lbergelson ?

src/main/java/htsjdk/samtools/util/IOUtil.java
lbergelson: You could do the conversion on the way in and out.  It's slightly less efficient, but probably worth it to avoid a bunch of lines of code.

src/main/java/htsjdk/tribble/util/ParsingUtils.java
lbergelson: You're not calling uri.get scheme, you're calling your own version that you added, IOUtil.getScheme() which returns true/false.

lbergelson: Ternary or not doesn't really matter, it's just kind of ugly to create a new wrapper object just to hide the null check.  I see your point about needing to call it twice or store it in a temp variable.  In any case, it feels like something is weird here.  You never actually use the scheme you get, you just check if it's empty.  Isn't calling `IOUtil.hasScheme`, which is what was originally done here correct?

lbergelson:
@yfarjoun A fwe new comments.  What I meant with regular file is that we don't ever check for fifos so we're handling them weirdly to begin with, your pr doesn't make it worse as far as I can tell.  We should do it correctly, but I think that's a different pr that should hit a lot more places in the codebase.

src/main/java/htsjdk/samtools/util/IOUtil.java
yfarjoun: fair enough.

yfarjoun:
@lbergelson ?

yfarjoun:
:party:

src/main/java/htsjdk/variant/vcf/VCFFileReader.java
magicDGS: Is there any plan appart of #520 to implement htsjdk3?

PR #910, opened 2017-06-23
More support for java.nio.Path: DiskBackedQueue and SortingCollection 

magicDGS:
### Description

- New methods in IOUtil to support Paths
- Change DiskBackedQueue to use Paths
- Change SortingCollection to use Paths
- Change SortingLongCollection to use Paths

### Checklist

- [ ] Code compiles correctly
- [ ] New tests covering changes and new functionality
- [ ] All tests passing
- [ ] Extended the README / documentation, if necessary
- [ ] Is not backward compatible (breaks binary or source compatibility)

PR #1019, opened 2017-10-30
- added a recursiveDeleteOnExit Utility method for the testing framework

yfarjoun:
- I have no idea how to test this...I need to shut-down the jvm so that the files will be deleted...any ideas?

### Description

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/TestUtil.java
+++ b/src/main/java/htsjdk/samtools/util/TestUtil.java
@@ -73,6 +73,20 @@ public class TestUtil {
     }
 
     /**
+     * Schedule a file or directory to be (recursively) deleted on jvm exit.
+     *
+     * This will silently delete the directory as well as it's contents.
+     * It improves upon {@link TestUtil#recursiveDelete(File)} by deleting directories and files that did not exist at call time.
+     *
+     * It cannot be undone, dir will be deleted unless program is killed.
+     *
+     * @param dir to be deleted
+     */
+    public static void deleteRecursivelyOnExit(final File dir){
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> recursiveDelete(dir)));
+    }
+
+    /**
      * Serialize and Deserialize an object
      * Useful for testing if serialization is correctly handled for a class.
      * @param input an object to serialize and then deserialize


lbergelson:
in gatk we have a test that just leaves an obviously named file in your test directory if it fails...

```
    @Test
    public void testDeleteDirOnExit() throws IOException {
        //This just tests that the code runs without crashing.
        //It runs at jvm shutdown so there isn't a good way to test it properly.
        //If you see a directory in the hellbender main folder called

        final File dir = new File(BaseTest.publicTestDir + "I_SHOULD_HAVE_BEEN_DELETED");
        IOUtils.deleteRecursivelyOnExit(dir);

        FileUtils.mkdir(dir, true);
        final File subdir = new File(dir, "subdir");
        FileUtils.mkdir(subdir, true);
        File someFile = new File(dir, "someFile");
        someFile.createNewFile();
        File anotherFile = new File(subdir, "anotherFile");
        anotherFile.createNewFile();
    }
```

src/main/java/htsjdk/samtools/util/TestUtil.java
```diff
@@ -73,6 +73,20 @@ public static void recursiveDelete(final File directory) {
...
+     *
+     * This will silently delete the directory as well as it's contents.
+     * It improves upon {@link TestUtil#recursiveDelete(File)} by deleting directories and files that did not exist at call time.
+     *
+     * It cannot be undone, dir will be deleted unless program is killed.
+     *
+     * @param dir to be deleted
+     */
+    public static void deleteRecursivelyOnExit(final File dir){
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> recursiveDelete(dir)));
```
lbergelson: I think the implementation of recursive delete may not be ideal here, since it can throw.  I'm not sure what happens if a shutdown hook throws, but I imagine it's not greatest thing to do.  You probably want to use a silent delete that eats any exceptions since I think it's better to leave a file undeleted than to have the test suite die inexplicably.  It's sort of a matter of taste I think.

lbergelson: It's also possible that throwing during shutdown isnt' a problem as well..

src/main/java/htsjdk/samtools/util/TestUtil.java
```diff
@@ -73,6 +73,20 @@ public static void recursiveDelete(final File directory) {
...
+     * Schedule a file or directory to be (recursively) deleted on jvm exit.
+     *
+     * This will silently delete the directory as well as it's contents.
+     * It improves upon {@link TestUtil#recursiveDelete(File)} by deleting directories and files that did not exist at call time.
+     *
+     * It cannot be undone, dir will be deleted unless program is killed.
+     *
+     * @param dir to be deleted
+     */
+    public static void deleteRecursivelyOnExit(final File dir){
```
magicDGS: In https://github.com/samtools/htsjdk/pull/910, I implemented a [`deleteOnExit(Path)`](https://github.com/samtools/htsjdk/pull/910/files#diff-828bef0ca8f35f36a26a63b0efac01d1R351) using a shutdown hook too but with  a thread implementation that can be tested.

Maybe you can have a look there, and add a version to `IOUtils` that can be used also outside the testing framework...

Issue #838, opened 2017-04-02
Very slow iteration over a single read


dariober:
### Subject of the issue

I don't know if this is a bug or me doing something wrong. 

I iterate over a small region of 100 bp in a bam file containing a single read using `SamLocuIterator`. If this single read contains a long skipped region, such as an intron, the iteration goes very slowly, something like ~30 seconds to go through 100 bp. This can't be right!

### Your environment
* version of htsjdk: 2.9.1
* version of java: 1.8
* which OS: MacOS 10.10.5

### Steps to reproduce

```
@Test
public void testIteratorWithIntron() {

	// Prepare a sam header
	String sqHeader = "@HD\tSO:coordinate\tVN:1.0\n"
                        + "@SQ\tSN:chr1\tAS:HG18\tLN:10000000\n";

	// Prepare one read with a 500,000 bases skipped
	String cigar= "18M500000N18M";
	String s1 = "read1\t0\tchr1\t1\t255\t" + cigar + "\t*\t0\t0\tACCTACGTTCAATATTACAGGCGAACATACTTACTA\t*\n";
	
	
	// Prepare sam input and samReader
	String exampleSam = sqHeader + s1 + s1;
	ByteArrayInputStream inputStream = new ByteArrayInputStream(exampleSam.getBytes());

	SamReader samReader= SamReaderFactory.makeDefault().open(SamInputResource.of(inputStream));

	// A small interval to iterate over:
	IntervalList il= new IntervalList(samReader.getFileHeader());
	il.add(new Interval("chr1", 1, 100));
	
	SamLocusIterator sli= new SamLocusIterator(samReader, il, true);
	
	// Iterate and time it
	long t0= System.currentTimeMillis();
	int n= 0;
	for (SamLocusIterator.LocusInfo li : sli) {
		n++;
	}
	long t1= System.currentTimeMillis();
	System.err.println("Time to iterate " + n + " loci: " + (t1-t0) / 1000.0 + " sec");
	sli.close();
}

// Result
Time to iterate 100 loci: 26.967 sec
```

### Expected and actual behaviour

The test method above should be much faster than ~27 seconds. (Note that a skipped region of 500 kb is not unusual with RNA-Seq).

PR #1020, opened 2017-11-02
Fix very slow iteration over single read

PolinaBevad:
### Description
Optimization of the iteration algorithm of the `AbstractLocusIterator`. 

When we use `SamLocusIterator` iteration over read that contains a long skipped region is very slow (about 27 seconds to skip 500 kbp - measurement from [issue](https://github.com/samtools/htsjdk/issues/838)).

Method `populateCompleteQueue()` used to remove empty regions from the `accumulator` by multiple invocations of the 'ArrayList' method  `remove()` (which uses System.arrayCopy() on each invocation). 

## Implementation
Empty regions now removed from accumulator with `ArrayList` method `removeIf()`.

## Performance
Test from issue took 1,6 sec. in our environment (in old realization it took 120 sec under the same conditions).

Other performance measurements:

| Cigar | Old method (ops/ms) | New method (ops/ms) |
| ------ | ------ | ------ |
| 2M2N2M2N2M2N2M2N2M2N2M2N2M2N2M2N2M2N18M | 97.144 ± 7.383 | 91.954 ± 8.647 |
| 18M2N18M | 96.661 ± 7.635 | 97.909 ± 13.210 |
| 18M5N18M | 103.623 ± 9.811 | 115.499 ± 10.209 |
| 18M50N18M | 103.476 ± 3.624 | 86.691 ± 12.006 |
| 18M500N18M | 27.233 ± 1.057 | 29.789 ± 2.166 |
| 18M5000N18M | 0.986 ± 0.203 | 3.403 ± 0.183 |
| 18M50000N18M | 0.013 ± 0.001 | 0.297 ± 0.016 |
| 18M500000N18M | ~ 10⁻⁴ | 60 * 10⁻⁴ ± 10 * 10⁻⁴ |

### Checklist

- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
+++ b/src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
@@ -480,9 +480,7 @@ public abstract class AbstractLocusIterator<T extends AbstractRecordAndOffset, K
             i++;
         }
         if (i > 0){
-            List<K> temporaryList = new ArrayList<>(accumulator.subList(i, accumulator.size()));
-            accumulator.clear();
-            accumulator.addAll(temporaryList);
+            accumulator.subList(0, i).clear();
         }
     }
 
--- a/src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
+++ b/src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
@@ -438,10 +438,8 @@ public abstract class AbstractLocusIterator<T extends AbstractRecordAndOffset, K
     private void populateCompleteQueue(final Locus stopBeforeLocus) {
         // Because of gapped alignments, it is possible to create LocusInfo's with no reads associated with them.
         // Skip over these if not including indels
-        while (!accumulator.isEmpty() && accumulator.get(0).isEmpty() &&
-                locusComparator.compare(accumulator.get(0), stopBeforeLocus) < 0) {
-            accumulator.remove(0);
-        }
+        accumulator.removeIf(x -> x.isEmpty() && locusComparator.compare(x, stopBeforeLocus) < 0);
+
         if (accumulator.isEmpty()) {
             return;
         }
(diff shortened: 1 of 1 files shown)


lbergelson:
@PolinaBevad Actually, on further thought I think this might introduce a performance regression for non-gapped reads.  Previously, if the read wasn't gapped it would short circuit as soon as it reached a point where x > locusToStop.

Using a slight modification of your test code, I measured the following:

Cigar = `36000M`

| original code | removeIf |
| --- | --- |
| 0.206 sec |  11.741 sec |

I think we may need a solution that does a fast remove and short circuits.  I don't see a neat way to do that in a one liner.

PolinaBevad:
@lbergelson Thank you for review!
Yes, I have tested this case and now see the problem. Code from new commit solve it: measurement of test for old and new implementations with cigar 36000M without gaps:

original, sec| new, sec
-----|------
0.941 | 0.92

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
```diff
@@ -475,6 +473,19 @@ private void populateCompleteQueue(final Locus stopBeforeLocus) {
...
     }
 
+    private void removeSkippedRegion(Locus stopBeforeLocus) {
+        int i = 0;
+        while (i < accumulator.size() && accumulator.get(i).isEmpty() &&
+                locusComparator.compare(accumulator.get(i), stopBeforeLocus) < 0) {
+            i++;
+        }
+        if (i > 0){
+            List<K> temporaryList = new ArrayList<>(accumulator.subList(i, accumulator.size()));
```
lbergelson: This seems to do a bunch of unnecessary allocations. 

 I wasn't sure the best way to do this, but according to the java doc for subList

```
  * <p>This method eliminates the need for explicit range operations (of
     * the sort that commonly exist for arrays).  Any operation that expects
     * a list can be used as a range operation by passing a subList view
     * instead of a whole list.  For example, the following idiom
     * removes a range of elements from a list:
     * <pre>
     *      list.subList(from, to).clear();
```

Which will use the efficient internal method `removeRange`.  So, I learned something new.  Would you update to use `subList.clear` instead of new list allocations?

Good to go after that.

PolinaBevad: @lbergelson Thanks for information about this method! I corrected code according to your comment.

lbergelson:
@PolinaBevad Thank you for this fix!

PR #1022, opened 2017-11-05
Fix for issue #1021

jrobinso:
This is a patch for issue #1021.  It is admittedly crude, you might want to handle it in some other way, in which case consider this PR illustrative.

It looks like someone has assumed that uri.getPath() will return the path minus query parameters, which in fact it does not.

### DIFF
--- a/src/main/java/htsjdk/tribble/AbstractFeatureReader.java
+++ b/src/main/java/htsjdk/tribble/AbstractFeatureReader.java
@@ -189,7 +189,13 @@ public abstract class AbstractFeatureReader<T extends Feature, SOURCE> implement
      * @return
      */
     public static boolean hasBlockCompressedExtension (final URI uri) {
-        return hasBlockCompressedExtension(uri.getPath());
+
+        String path = uri.getPath();
+        int qIdx = path.indexOf('?');
+        if(qIdx > 0) {
+            path = path.substring(0, qIdx);
+        }
+        return hasBlockCompressedExtension(path);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


lbergelson:
Interesting.  The URI version of `hasBlockCompressedExtension` seems to work correctly with uri's with queries.  We even have tests for it.  The problem seems to be that we're not using the URI version in the codepath codepath you're hitting, and the `String` version doesn't handle it.  

I haven't tracked down whatever underlying change made the difference yet.

```
    @DataProvider(name = "blockCompressedExtensionExtensionURIStrings")
    public Object[][] createBlockCompressedExtensionURIs() {
        return new Object[][]{

                 ...

                {"https://www.googleapis.com/download/storage/v1/b/deflaux-public-test/o/NA12877.vcf.gz?alt=media", true},
                {"https://www.googleapis.com/download/storage/v1/b/deflaux-public-test/o/NA12877.vcf.gzip?alt=media", true},
                {"https://www.googleapis.com/download/storage/v1/b/deflaux-public-test/o/NA12877.vcf.bgz?alt=media", true},
                {"https://www.googleapis.com/download/storage/v1/b/deflaux-public-test/o/NA12877.vcf.bgzf?alt=media", true},
                {"https://www.googleapis.com/download/storage/v1/b/deflaux-public-test/o/NA12877.vcf.bzip2?alt=media", false},

                 ...
        };
    }

    @Test(dataProvider = "blockCompressedExtensionExtensionURIStrings")
    public void testBlockCompressionExtension(final String testURIString, final boolean expected) throws URISyntaxException {
        URI testURI = URI.create(testURIString);
        Assert.assertEquals(AbstractFeatureReader.hasBlockCompressedExtension(testURI), expected);
    }
```

lbergelson:
@jrobinso Thanks for this patch, I made a few changes and pushed a version at #1033, so I'm going to close this one. Hopefully that will work to solve your problem.

PR #1025, opened 2017-11-08
Adds option to validator to skip mate validation

eitanbanks:
### Description

Added an option to the SAM validator to skip mate validation.

Mate validation is an extremely expensive operation memory-wise for certain classes of data.  Specifically, inputs that either have duplicate sets with many millions of reads or that have high rates of chimerism will trigger terrible edge case behavior requiring the tool to maintain many (many, many) reads in memory.  This just provides a simple way for one to generally validate those files without requiring the massive memory footprint (or ripping the guts out and rewriting it all). 
    
### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SamFileValidator.java
+++ b/src/main/java/htsjdk/samtools/SamFileValidator.java
@@ -147,6 +147,10 @@ public class SamFileValidator {
         this.skipMateValidation = skipMateValidation;
     }
 
+    public boolean getSkipMateValidation() {
+        return skipMateValidation;
+    }
+
     /**
      * Outputs validation summary report to out.
      *
--- a/src/main/java/htsjdk/samtools/SamFileValidator.java
+++ b/src/main/java/htsjdk/samtools/SamFileValidator.java
@@ -140,8 +140,8 @@ public class SamFileValidator {
     }
 
     /**
-     * Sets whether or not we should go ahead and run mate validation.  For extreme cases, mate validation
-     * can require a ton of memory and we want to provide the option of skipping it.
+     * Sets whether or not we should go ahead and run mate validation beyond the mate cigar check.  For extreme cases,
+     * mate validation can require a ton of memory and we want to provide the option of skipping it.
      */
     public void setSkipMateValidation(final boolean skipMateValidation) {
         this.skipMateValidation = skipMateValidation;
@@ -311,10 +311,7 @@ public class SamFileValidator {
                     }
                 }
 
-                if (!skipMateValidation) {
-                    validateMateFields(record, recordNumber);
-                }
-
+                validateMateFields(record, recordNumber);
                 final boolean hasValidSortOrder = validateSortOrder(record, recordNumber);
                 validateReadGroup(record, header);
                 final boolean cigarIsValid = validateCigar(record, recordNumber);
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/SamFileValidator.java
```diff
@@ -301,7 +311,10 @@ private void validateSamRecordsAndQualityFormat(final Iterable<SAMRecord> samRec
                     }
                 }
 
-                validateMateFields(record, recordNumber);
+                if (!skipMateValidation) {
+                    validateMateFields(record, recordNumber);
```
yfarjoun: this skips the mate-cigar validation as well, would it be enough to move the if statement into validateMateFields and avoid the expensive storing of the reads, but not the validation of the mate-cigars?

eitanbanks:
Certainly, that works.  I've gone ahead and made that change.  Look ok?

jacarey:
Looks good to me as well.

eitanbanks:
@lbergelson and/or @tfenne any objections here?  There will be a corresponding ~2 line change in Picard to enable this option in ValidateSamFile once this is merged.

src/main/java/htsjdk/samtools/SamFileValidator.java
```diff
@@ -138,6 +140,14 @@ public void setIgnoreWarnings(final boolean ignoreWarnings) {
     }
 
     /**
+     * Sets whether or not we should go ahead and run mate validation beyond the mate cigar check.  For extreme cases,
+     * mate validation can require a ton of memory and we want to provide the option of skipping it.
+     */
+    public void setSkipMateValidation(final boolean skipMateValidation) {
```
lbergelson: You should probably add a matching getter, someone always ends up wanting one for some reason or another.

src/main/java/htsjdk/samtools/SamFileValidator.java
```diff
@@ -508,6 +518,10 @@ private void validateMateFields(final SAMRecord record, final long recordNumber)
         }
         validateMateCigar(record, recordNumber);
 
+        if (skipMateValidation) {
+            return;
```
lbergelson: You might want a test for this.  Check that it validates the mate cigar but not missing mates.   Otherwise the monkeys tampering with the code in the future could change the semantics unexpectedly.

src/main/java/htsjdk/samtools/SamFileValidator.java
nh13: Can you reword the comment?  It reads like an email, not javadoc.

src/main/java/htsjdk/samtools/SamFileValidator.java
```diff
@@ -138,6 +140,18 @@ public void setIgnoreWarnings(final boolean ignoreWarnings) {
...
 
     /**
+     * Sets whether or not we should go ahead and run mate validation beyond the mate cigar check.  For extreme cases,
+     * mate validation can require a ton of memory and we want to provide the option of skipping it.
+     */
+    public void setSkipMateValidation(final boolean skipMateValidation) {
+        this.skipMateValidation = skipMateValidation;
+    }
+
+    public boolean getSkipMateValidation() {
```
nh13: javadoc

eitanbanks:
Okay,
@lbergelson I added a test and a getter.
@nh13 I fixed the docs.

nh13:
👍

PR #1027, opened 2017-11-15
Add method to log the last record recorded.

nh13:

### DIFF
--- a/src/main/java/htsjdk/samtools/util/AbstractProgressLogger.java
+++ b/src/main/java/htsjdk/samtools/util/AbstractProgressLogger.java
@@ -46,7 +46,7 @@ abstract public class AbstractProgressLogger implements ProgressLoggerInterface 
      */
     abstract protected void log(String ... message);
 
-    private synchronized void log(final String chrom, final int pos) {
+    private synchronized void record() {
         final long now = System.currentTimeMillis();
         final long lastPeriodSeconds = (now - this.lastStartTime) / 1000;
         this.lastStartTime = now;
@@ -57,20 +57,22 @@ abstract public class AbstractProgressLogger implements ProgressLoggerInterface 
         final String processed = pad(fmt.format(this.processed), 13);
 
         final String readInfo;
-        if (chrom == null) readInfo = "*/*";
-        else readInfo = chrom + ":" + fmt.format(pos);
+        if (this.lastChrom == null) readInfo = "*/*";
+        else readInfo = this.lastChrom + ":" + fmt.format(this.lastPos);
 
-        log(this.verb, " ", processed, " " + noun + ".  Elapsed time: ", elapsed, "s.  Time for last ", fmt.format(this.n),
+        final long n = (this.processed % this.n == 0) ? this.n : this.processed % this.n;
+
+        log(this.verb, " ", processed, " " + noun + ".  Elapsed time: ", elapsed, "s.  Time for last ", fmt.format(n),
                 ": ", period, "s.  Last read position: ", readInfo);
     }
 
     /**
-     * Logs the last last record if it wasn't previoiusly logged.
+     * Logs the last last record if it wasn't previously logged.
      * @return boolean true if logging was triggered, false otherwise
      */
-    public synchronized boolean recordLast() {
+    public synchronized boolean log() {
         if (this.processed % this.n != 0) {
-            log(this.lastChrom, this.lastPos);
+            record();
             return true;
         }
         else {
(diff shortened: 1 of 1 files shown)


src/main/java/htsjdk/samtools/util/AbstractProgressLogger.java
```diff
@@ -44,35 +46,51 @@ protected AbstractProgressLogger(final String noun, final String verb, final int
      */
     abstract protected void log(String ... message);
 
+    private synchronized void log(final String chrom, final int pos) {
```
tfenne: Why have this take parameters and not just use `this.lastChrom` and `this.lastPos`?

src/main/java/htsjdk/samtools/util/AbstractProgressLogger.java
```diff
@@ -44,35 +46,51 @@ protected AbstractProgressLogger(final String noun, final String verb, final int
...
-            if (chrom == null) readInfo = "*/*";
-            else readInfo = chrom + ":" + fmt.format(pos);
+        log(this.verb, " ", processed, " " + noun + ".  Elapsed time: ", elapsed, "s.  Time for last ", fmt.format(this.n),
+                ": ", period, "s.  Last read position: ", readInfo);
+    }
 
-            log(this.verb, " ", processed, " " + noun + ".  Elapsed time: ", elapsed, "s.  Time for last ", fmt.format(this.n),
-                    ": ", period, "s.  Last read position: ", readInfo);
+    /**
+     * Logs the last last record if it wasn't previoiusly logged.
```
tfenne: previoiusly -> previously

src/main/java/htsjdk/samtools/util/AbstractProgressLogger.java
```diff
@@ -44,35 +46,51 @@ protected AbstractProgressLogger(final String noun, final String verb, final int
...
 
-            final long seconds = (System.currentTimeMillis() - startTime) / 1000;
-            final String elapsed   = formatElapseTime(seconds);
-            final String period    = pad(fmt.format(lastPeriodSeconds), 4);
-            final String processed = pad(fmt.format(this.processed), 13);
-
-            final String readInfo;
-            if (chrom == null) readInfo = "*/*";
-            else readInfo = chrom + ":" + fmt.format(pos);
+        log(this.verb, " ", processed, " " + noun + ".  Elapsed time: ", elapsed, "s.  Time for last ", fmt.format(this.n),
```
tfenne: I think that `this.n` needs to be replaced with something else here, as it'll be incorrect when invoked from `recordLast()`.  If think you want something like:

```java
final int n = (this.processed % this.n == 0) ? this.n : this.processed % this.n;
```

src/main/java/htsjdk/samtools/util/AbstractProgressLogger.java
```diff
@@ -44,35 +46,51 @@ protected AbstractProgressLogger(final String noun, final String verb, final int
...
+                ": ", period, "s.  Last read position: ", readInfo);
+    }
 
-            log(this.verb, " ", processed, " " + noun + ".  Elapsed time: ", elapsed, "s.  Time for last ", fmt.format(this.n),
-                    ": ", period, "s.  Last read position: ", readInfo);
+    /**
+     * Logs the last last record if it wasn't previoiusly logged.
+     * @return boolean true if logging was triggered, false otherwise
+     */
+    public synchronized boolean recordLast() {
```
tfenne: I think this should probably just be called `log()` since it's not recording anything and technically could be used any time you want to print progress, not just at the last record.

src/main/java/htsjdk/samtools/util/AbstractProgressLogger.java
```diff
@@ -44,35 +46,51 @@ protected AbstractProgressLogger(final String noun, final String verb, final int
...
             return false;
         }
     }
 
-
+    @Override
+    public synchronized boolean record(final String chrom, final int pos) {
+        this.lastChrom = chrom;
+        this.lastPos = pos;
+        if (this.lastStartTime == -1) this.lastStartTime = System.currentTimeMillis();
```
pshapiro4broad: Odd indent on line 86 and lack of braces makes this code confusing. Better to use a newline and `{` `}` around the `if` body to make the intent clear. Fixing the indent will also help.

nh13: Just keeping the style the same ;)

nh13:
@tfenne fixed based on your comments. @pshapiro4broad thanks for the review as well!

Issue #1030, opened 2017-11-16
PR #910 adds a `user.name` property component to the default `tmpdir` for `SortingCollection`.


tbl3rd:
PR #910 adds a `user.name` property component to the default `tmpdir` for `SortingCollection`.

This change to line 347 of `SortingCollection`.
```
-        final File tmpDir = new File(System.getProperty("java.io.tmpdir"));
+        final Path tmpDir = IOUtil.getDefaultTmpDirPath();
```
adds the value of the `user.name` property as an extra component to the default `tmpdir` for `SortingCollection` if it is not already present.

Furthermore, `SortingCollection.newTempFile()` will now throw unless the `java.io.tmpdir/user.name` temporary directory exists.

```
Caused by: java.nio.file.NoSuchFileException:
  java.io.tmpdir/user.name/sortingcollection.<some-number>.tmp
	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
	at sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:214)
	at java.nio.file.Files.newByteChannel(Files.java:361)
	at java.nio.file.Files.createFile(Files.java:632)
	at java.nio.file.TempFileHelper.create(TempFileHelper.java:138)
	at java.nio.file.TempFileHelper.createTempFile(TempFileHelper.java:161)
	at java.nio.file.Files.createTempFile(Files.java:852)
	at htsjdk.samtools.util.IOUtil.newTempPath(IOUtil.java:316)
	at htsjdk.samtools.util.SortingCollection.newTempFile(SortingCollection.java:255)
	at htsjdk.samtools.util.SortingCollection.spillToDisk(SortingCollection.java:220)
	... 8 more
```

The following change would restore the prior behavior while retaining the `java.nio.file.Path` changes.
```
-        final Path tmpDir = IOUtil.getDefaultTmpDirPath();
+        final Path tmpDir = Paths.get(System.getProperty("java.io.tmpdir"));
```

Saw this while running htsjdk 2.13.0 over HotSpot 1.8.0_121-b13 on Red Hat 2.6.32.

tbl3rd:
... for the PR link which doesn't work in the issue title.

Issue #1029, opened 2017-11-16
PR 910 introduces regression whereby sorting collection throws exceptions at JVM exit


tfenne:
@magicDGS It looks to me like your PR #910 changes the behavior of temporary file creation and cleanup in a way that breaks lots of downstream usage.

The problem appears to be in `DeletePathThread` where it attempts to delete a Path and rethrows any exceptions that are encountered.  The problem is that one of those exceptions that can be thrown is a `NoSuchFileException` because the file has already been cleaned up.

Running SortSam using the latest version of Picard I get:

```
$  java -Xmx4g -jar picard.jar SortSam I=in.bam O=out.bam SO=queryname
...
INFO	2017-11-16 07:50:23	SortSam	Finished reading inputs, merging and writing to output now.
[Thu Nov 16 07:50:27 EST 2017] picard.sam.SortSam done. Elapsed time: 0.18 minutes.
Runtime.totalMemory()=1586495488
Exception in thread "Thread-2" Exception in thread "Thread-1" htsjdk.samtools.util.RuntimeIOException: java.nio.file.NoSuchFileException: /var/folders/mz/34h8j89n1jj2mg6frd0lmqxh0000gn/T/tfenne/sortingcollection.3170510762910173658.tmp
	at htsjdk.samtools.util.IOUtil$DeletePathThread.run(IOUtil.java:374)
Caused by: java.nio.file.NoSuchFileException: /var/folders/mz/34h8j89n1jj2mg6frd0lmqxh0000gn/T/tfenne/sortingcollection.3170510762910173658.tmp
	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
	at sun.nio.fs.UnixFileSystemProvider.implDelete(UnixFileSystemProvider.java:244)
	at sun.nio.fs.AbstractFileSystemProvider.delete(AbstractFileSystemProvider.java:103)
	at java.nio.file.Files.delete(Files.java:1126)
	at htsjdk.samtools.util.IOUtil$DeletePathThread.run(IOUtil.java:372)
htsjdk.samtools.util.RuntimeIOException: java.nio.file.NoSuchFileException: /var/folders/mz/34h8j89n1jj2mg6frd0lmqxh0000gn/T/tfenne/sortingcollection.5241745018661624118.tmp
[...]

PR #1031, opened 2017-11-17
Issue #1030: Restore default tmpdir for SortingCollection before PR #910.

tbl3rd:
Fix Issue #1030 by restoring `SortingCollection.newInstance()` to its behavior before PR #910.

- [x] Code compiles correctly
It compiles.
Wasn't tested before. Isn't tested now.
- [x] All tests passing
They pass, but the JVM throws on shutdown as before. See Issue #1029.

### DIFF
--- a/src/main/java/htsjdk/samtools/util/SortingCollection.java
+++ b/src/main/java/htsjdk/samtools/util/SortingCollection.java
@@ -34,6 +34,7 @@ import java.io.Serializable;
 import java.lang.reflect.Array;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/util/SortingCollection.java
```diff
@@ -344,7 +345,7 @@ public void cleanup() {
                                                        final Comparator<T> comparator,
                                                        final int maxRecordsInRAM) {
 
-        final Path tmpDir = IOUtil.getDefaultTmpDirPath();
+        final Path tmpDir = Paths.get(System.getProperty("java.io.tmpdir"));
```
lbergelson: should we worry about the case where this is null?

pshapiro4broad: There's no need, this is the same property name that java uses internally when you call APIs like `Files.createTempFile()`. If it fails then java is not configured correctly.

It would be nice if there was a non-magic string way to do it, but Java doesn't have a constant for property names like this.

PR #1032, opened 2017-11-17
fixing crash on shutdown hook #1029

### LINKED ISSUE #1029: PR 910 introduces regression whereby sorting collection throws exceptions at JVM exit

lbergelson:
fixing the crash on shutdown introduced in #910 by changin the command
that gets run on deleteOnExit
Files.delete() -> Files.deleteIfExists()
this will prevent the jvm from crashing on shutdown if the temporary
files get cleaned up before it closes

This still leaves open the possibilty of crash on shutdown if the
temporary file is unable to be removed for some other reason.  That's
subtly different than the File version of deleteOnExit, but it may be
useful to get these error messages in the unusual case of failing to
delete a file that exists.

fixes #1029

### DIFF
--- a/src/main/java/htsjdk/samtools/util/IOUtil.java
+++ b/src/main/java/htsjdk/samtools/util/IOUtil.java
@@ -360,16 +360,16 @@ public class IOUtil {
      *
      * @see #deleteOnExit(Path)
      */
-    protected static final class DeletePathThread extends Thread {
+    static final class DeletePathThread extends Thread {
 
         private final Path path;
 
-        protected DeletePathThread(Path path) {this.path = path;}
+        DeletePathThread(Path path) {this.path = path;}
 
         @Override
         public void run() {
             try {
-                Files.delete(path);
+                Files.deleteIfExists(path);
             } catch (IOException e) {
                 throw new RuntimeIOException(e);
             }


lbergelson:
@tfenne Does this fix the problem for you?  We could also completely suppress deletions during shutdown, which would match the behavior of File.deleteOnExit, but would potentially mask legitimate issues.

tfenne:
I think so, thanks @lbergelson.  I think I would lean towards not rethrowing exceptions but instead logging them.  Since this is happening during a JVM shutdown I'm not sure it's useful to have programs fail if temporary files cannot be deleted.  What do you think?

lbergelson:
I was leaning towards warn on failure.  I was discussing it with @yfarjoun and he felt that failure was preferred.  I can see an argument either way.  Happy to change it.

tfenne:
I'm ok either way.  If it becomes annoying in practice we can change it again.

yfarjoun:
My thought was, if a future change causes these temporary files to be read-only, we want the tests to fail since we will be creating temporary files and not deleting them....

yfarjoun:
(though now that I think of it, since the JVM will not exit, the failure will not happen... :-/ )

lbergelson:
I think it won't be a big deal in practice, since it will only fail under rare circumstances.  I'm going to merge this and we can revisit if it continues to be an issue.

magicDGS:
Maybe we can add a similar approach as the [`java.io.DeleteOnExitHook`](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/io/DeleteOnExitHook.java) for Paths.

I didn't look into that possibility when I added the support in `IOUtil`...

PR #1036, opened 2017-11-20
fix a typo in SAMRecord

lbergelson:
fix a typo in SAMRecord that was bothering me.


tfenne:
@lbergelson The change is totally reasonable.  Except that the message there is confusing in the first place.  It implies that secondary alignments are for "split hits", but they are not.  Secondary alignments (not primary) are for "alternative alignments" while supplementary are for "split alignments" or alignments that cannot be represented co-linearly with the reference.

PR #1037, opened 2017-11-21
Add DeleteOnExitPathHook to register java.nio.Path for deletion

magicDGS:
### Description

After https://github.com/samtools/htsjdk/pull/910, I introduced an crash while deleting on exit a `java.nio.Path`. This was fixed by https://github.com/samtools/htsjdk/pull/1032, but I explore a bit the java base code to check what the `File` version is handling the deletion on exit.

It turns out that it is also using a shutdown hook, but a common one for all the registered files (see [`java.io.DeleteOnExitHook`](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/io/DeleteOnExitHook.java)). Here, I mimic that class to register for deletion the `java.nio.Path` in `IOUtils` and run a single shutdown hook for all of them, ignoring any `IOException` or `RuntimeException`. I think that this will be more consistent with the behaviour expected by `File.deleteOnExit()`

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/IOUtil.java
+++ b/src/main/java/htsjdk/samtools/util/IOUtil.java
@@ -31,6 +31,7 @@ import htsjdk.samtools.seekablestream.SeekableFileStream;
 import htsjdk.samtools.seekablestream.SeekableHTTPStream;
 import htsjdk.samtools.seekablestream.SeekableStream;
 import htsjdk.tribble.Tribble;
+import htsjdk.samtools.util.nio.DeleteOnExitPathHook;
 
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
@@ -353,34 +354,12 @@ public class IOUtil {
     }
 
     /**
-     * Deletes on exit a path by creating a shutdown hook.
+     * Register a {@link Path} for deletion on JVM exit.
+     *
+     * @see DeleteOnExitPathHook
      */
     public static void deleteOnExit(final Path path) {
-        // add a shutdown hook to remove the path on exit
-        Runtime.getRuntime().addShutdownHook(new DeletePathThread(path));
-    }
-
-    /**
-     * WARNING: visible for testing. Do not use.
-     *
-     * Class for delete a path, used in a shutdown hook for delete on exit.
-     *
-     * @see #deleteOnExit(Path)
-     */
-    static final class DeletePathThread extends Thread {
-
-        private final Path path;
-
-        DeletePathThread(Path path) {this.path = path;}
-
-        @Override
-        public void run() {
-            try {
-                Files.deleteIfExists(path);
-            } catch (IOException e) {
-                throw new RuntimeIOException(e);
-            }
-        }
+        DeleteOnExitPathHook.add(path);
     }
 
     /** Returns the name of the file minus the extension (i.e. text after the last "." in the filename). */
--- a/src/main/java/htsjdk/samtools/util/nio/DeleteOnExitPathHook.java
+++ b/src/main/java/htsjdk/samtools/util/nio/DeleteOnExitPathHook.java
@@ -1,0 +1,63 @@
+package htsjdk.samtools.util.nio;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+
+/**
+ * Class to hold a set of {@link Path} to be delete on the JVM exit through a shutdown hook.
+ *
+ * <p>This class is a modification of {@link java.io.DeleteOnExitHook} to handle {@link Path}
+ * instead of {@link java.io.File}.
+ *
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class DeleteOnExitPathHook {
+    private static LinkedHashSet<Path> paths = new LinkedHashSet<>();
+    static {
+        Runtime.getRuntime().addShutdownHook(new Thread(DeleteOnExitPathHook::runHooks));
+    }
+
+    private DeleteOnExitPathHook() {}
+
+    /**
+     * Adds a {@link Path} for deletion on JVM exit.
+     *
+     * @param path path to be deleted.
+     *
+     * @throws IllegalStateException if the shutdown hook is in progress.
+     */
+    public static synchronized void add(Path path) {
+        if(paths == null) {
+            // DeleteOnExitHook is running. Too late to add a file
+            throw new IllegalStateException("Shutdown in progress");
+        }
+
+        paths.add(path);
+    }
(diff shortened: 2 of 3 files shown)


src/main/java/htsjdk/samtools/util/nio/DeleteOnExitPathHook.java
```diff
@@ -0,0 +1,63 @@
...
+        ArrayList<Path> toBeDeleted = new ArrayList<>(thePaths);
+
+        // reverse the list to maintain previous jdk deletion order.
+        // Last in first deleted.
+        Collections.reverse(toBeDeleted);
+        for (Path path : toBeDeleted) {
+            try {
+                Files.delete(path);
+            } catch (IOException | RuntimeException e) {
+                // do nothing if cannot be deleted, because it is a shutdown hook
```
pshapiro4broad: It seems wrong to throw away an exception here. Can it safely be logged?

magicDGS: This is what the File version of the hook is doing (because the result of the `File.delete` is ignored there). In addition, because it is a shutdown-hook, the JVM will crash on exit if we throw them (see an example in https://github.com/samtools/htsjdk/issues/1029).

I think that we can re-write this block as:

```java
...
try {
    Files.delete(path);
} catch (IOException | RuntimeException e) {
    System.err.println(String.format("Error deleting on exit %s: %s",
                path.toUri().toString(),
                e.getMessage())
    );
    e.printStackTrace();
}
...
```

But I think that this is too verbose, and actually it will be a difference with the `File` version.

pshapiro4broad: This isn't what the `File` version is doing; your code catches `RuntimeException` and the `File` code does not. This is a pretty big difference in behavior.

Maybe logging a stack trace isn't necessary here, though. Also we don't plan to delete GCS bucket entries on exit, correct? If so you can format the output using `String.format(".... %s ...", path);` and it will look fine.

magicDGS: I am catching `RuntimeException` because of the `SecurityException` thrown by `Files.delete`.

Next commit I will catch only that one, without the logging. I think that it is too verbose for just saying that a temp file was not removed if we add them (e.g., if it does not exists or the directory isn't empty).

pshapiro4broad: > I am catching RuntimeException because of the SecurityException thrown by Files.delete.

OK, so this is a required difference between our delete-on-exit hook and the one in `java.io`?

> I think that it is too verbose for just saying that a temp file was not removed if we add them (e.g., if it does not exists or the directory isn't empty).

That is not how `File.delete()` works. It will not throw an exception if the file doesn't exist or if the directory isn't empty, it will only return `false` to indicate failure.

magicDGS: But `File.deleteOnExit` uses a hook that ignores the value of `File.delete`. This hook should be use for mimic deleteOnExit for Path, that's why is ignoring the result of delete...

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -346,34 +347,12 @@ public static Path getDefaultTmpDirPath() {
...
-        @Override
-        public void run() {
-            try {
-                Files.deleteIfExists(path);
-            } catch (IOException e) {
-                throw new RuntimeIOException(e);
-            }
-        }
+    public static void deleteOnExit(final Path path) {
+        DeleteOnExitPathHook.add(path);
```
pshapiro4broad: Why not call `path.toFile().deleteOnExit()` here?

pshapiro4broad: So on HDFS this will throw? Can we special case HDFS by examining `Path.getFileSystem()`?

Do we have cases where we'll be creating temporary files in HDFS or is this a theoretical case?

lbergelson: @pshapiro4broad The reason we're adding support for Path is to support file systems that can't be represented as files, i.e. hdfs://, gs://, s3://, etc.   We definitely want to be able to make temp files on hdfs, gcs, etc.  Not so much for production work, it wouldn't make much sense to store temp files on gcs, but it's incredibly use for testing.

pshapiro4broad: OK. I think we'd be better off with a more test specific way to delete test data, and avoid using shutdown hooks complete, but that seems unlikely to change at this point.

magicDGS: I think that it is a common practice to use `File.deleteOnExit` for some temp files (not only for test data, but also for temporary files written on disk) because it is better to use the shutdown hook as it will run always and thus not keeping the tmp files if there is a crash that does not finalize.

For `Path`, there is not such a method. This commit here is just using the same approach as the `File` version, to be in-line with what it is expected if we change HTSJDK to use only `Path`. For example, this is also used in the sorting collection and using it with a, for example, HDFS/GCS `Path` as a temporary directory requires that this method remove the files on shutdown of the JVM.

jacarey:
resolve conflicts and squash commits and we can merge it.

src/main/java/htsjdk/samtools/util/IOUtil.java
pshapiro4broad: It's a common practice in this codebase but it's not IMO a good practice, and is not common everywhere. A better system is to make the code define the scope of all resources, so that when a resource goes out of scope, it can be deleted. There is no need to use a finalizer, using a `finally` clause (or its equivalent) is sufficient.

lbergelson: @pshapiro4broad The problem with that is that it's a pain and people don't do it when they write tests so they just leave files everywhere.  This is the best way I know that doesn't impose a lot of boilerplate on the test code.

gbggrant:
Hi @magicDGS could you go ahead and resolve the conflicts and squash your commits?  We would like to get this PR merged as it fixes a problem in our pipelines.

magicDGS:
@gbggrant - Sorry for the delay, I wasn't sure that this was affecting someone.

lbergelson:
Heh, it's hard to for us to expect contributors to be super speedy when I still have 50 unreviewed PR's to get too...

lbergelson:
@magicDGS Thank you!

gbggrant:
Yes - thanks @magicDGS

magicDGS:
Thank you too for consider my contributions!

PR #1024, opened 2017-11-07
Bugfix for BAM index bins

valeriuo:
1. Fixes a bug: `ArrayIndexOutOfBoundsException` would be thrown if `levelNumber` is too big.
2. Fixes a bug: the level size for the last level was miscalculated. It should be **32768** (2^15).
3. Removes the unused method `regionToBins`, whose code was already implemented in the class `GenomicIndexUtil`.

### Description

Please explain the changes you made here.
Explain the **motivation** for making this change. What existing problem does the pull request solve?

### Checklist

- [x] Code compiles correctly
- [x] New tests covering changes and new functionality
- [x] All tests passing
- [ ] Extended the README / documentation, if necessary
- [ ] Is not backward compatible (breaks binary or source compatibility)

PR #1040, opened 2017-11-22
CSI index support for BAM files

valeriuo:
### Description

1. Adds the ability to read and use [CSI indexes](https://github.com/samtools/hts-specs/blob/master/CSIv1.pdf) for BAM files (no write support, yet). BAI indexes still have priority. Thus, if both .bai and .csi files are present alongside a BAM file, the BAI index will be used.
2. Includes #1024.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


lbergelson:
@valeriuo I'm sorry we haven't gotten to this yet.  I've been very busy and this is obviously a complicated PR.  It's a really valuable PR though, lots of people want CSI support, so we will definitely get around to reviewing it when someone has bandwidth.

valeriuo:
@lbergelson I realize there is a lot of code in here. Some of it is duplicated from `AbstractBAMFileIndex` class, which I couldn't extend, although that might have seemed like the natural approach. 
We could arrange a call, if you need me to present the solution and talk about potential improvements.

yfarjoun:
@lbergelson could this be prioritized, there's a picard PR waiting on it. https://github.com/broadinstitute/picard/pull/998

lbergelson:
@yfarjoun Yes, I was planning on prioritizing this as the next thing after going through a bunch of small stuff that has been in flight for a while.

lbergelson:
@valeriuo I've begun to take a look at this.  Sorry for the extremely long delay.  Are you still around and interested in helping get this in?

valeriuo:
@lbergelson, definitely.

src/main/java/htsjdk/samtools/BAMIndexMetaData.java
```diff
@@ -268,4 +279,29 @@ static public void printIndexStats(final File inputBamFile) {
 
         return result;
     }
+
+    /**
+     * Prints meta-data statistics from BAM index (.csi) file
+     * Statistics include count of aligned and unaligned reads for each reference sequence
+     * and a count of all records with no start coordinate
+     */
+    static public BAMIndexMetaData[] getCSIIndexStats(final BAMFileReader bam) {
```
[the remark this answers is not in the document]
lbergelson: The index hierarchy is such a mess.  I think we should consider restructuring it to some degree to make this method unnecessary.

src/main/java/htsjdk/samtools/BAMCSIFileIndex.java
```diff
@@ -0,0 +1,764 @@
...
+
+        sequenceIndexes = new int[getNumberOfReferences() + 1];
+        Arrays.fill(sequenceIndexes, -1);
+    }
+
+    public BAMCSIFileIndex(final File file, final SAMSequenceDictionary dictionary) {
+        this(file, dictionary, true);
+    }
+
+    public BAMCSIFileIndex(final File file, final SAMSequenceDictionary dictionary, int minShift, int binDepth) {
```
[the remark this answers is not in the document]
valeriuo: True. I used it for some test. I'll remove it.

src/main/java/htsjdk/samtools/BAMCSIFileIndex.java
```diff
@@ -0,0 +1,764 @@
...
+
+    /**
+     * Getters and setters
+     */
+
+    public int getBinDepth() {
+        return binDepth;
+    }
+
+    public void setBinDepth(int binDepth) {
```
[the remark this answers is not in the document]
valeriuo: Good point.

src/main/java/htsjdk/samtools/BAMIndex.java
```diff
@@ -34,6 +34,7 @@
 public interface BAMIndex extends Closeable {
 
     public static final String BAMIndexSuffix = ".bai";
+    public static final String BAMIndexSuffix2 = ".csi";
```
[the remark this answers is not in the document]
valeriuo: `BAI_INDEX_SUFFIX` would probably be more appropriate.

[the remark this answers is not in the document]
lbergelson: You're right, that's better.

src/main/java/htsjdk/samtools/BAMCSIFileIndex.java
```diff
@@ -0,0 +1,764 @@
...
+    private final IndexFileBuffer mIndexBuffer;
+    private SAMSequenceDictionary mBamDictionary = null;
+
+    private final int [] sequenceIndexes;
+
+    /**
+     * Constructors
+     */
+
+    private BAMCSIFileIndex(final SeekableStream stream, final SAMSequenceDictionary dictionary)
```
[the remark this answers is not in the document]
valeriuo: Just to clarify, `Path` would be memory mapped and `File` would be random accessed. Is this what you had in mind?

valeriuo:
@lbergelson I've addressed all the topics mentioned in your review. I didn't add any unit tests, but I wanted to push the latest code version, because I am going on a short holiday until next week and I thought you could have another look over it in the meantime. 

I've managed to refactor the BAM index class hierarchy by opening slightly the `AbstractBAMFileIndex` class, without exposing anything to the public interface or braking its contract, in order to allow `CSIIndex` to extend it, thus removing some duplicate code.

I've tested the CSI index functionality by running [Picard](https://github.com/broadinstitute/picard/pull/998) `BamIndexStats` command against some indexes generated with **samtools**. The results were OK, but the inconvenient is that samtools generates compressed CSI files, by default, and I had to decompress them manually so that Picard could read them. For that matter, BAI files are uncompressed. I don't know if Picard can read compressed index files, but we'll have to address this limitation somehow, as I can see some usability issue.

lbergelson:
@valeriuo Oh, that's awesome!  I will take a look.  That's interesting about the compression.  It doesn't seem to mention it anywhere in the spec.  I assume they're just gzip?  Or are they bgzip?  We should be able to add detection of the compression and automatically decompress.  There was just recently some code added to make that sort of detection easier as well.

jkbonfield:
Htslib writes CSI using bgzf (but I doubt it matters whether it is this or pure gzip as we have no random access capability to our indices anyway).  You are correct in that it is not documented.  I note that Tabix does document this, as bgzf, so I assume this is just an omission in the CSI documentation.

Note that what CSI does to BAI - allowing for longer chromosomes - needs to be applied to tabix TBI also.  There is no point having long chromosome support in BAM if we then cannot analyse it and handle bgzipped VCF.  Bcftools does this via an intermediate CSI/TBI hybrid, which is sadly also undocumented!  It uses a tabix header as the meta-data for a CSI file, which means it doesn't have to have every #contig line present in the index and reference numbering is based on presence in the file rather than presence in the VCF header.   I believe this is an attempt to handle the case where the VCF header does not exist (#contig lines are optional), but I can't really be sure.  Maybe it's simply an accident of sharing code paths!

Note, I've just added CSI and BAI support for bgzipped SAM in Htslib, mainly for completeness but also as a safe workaround for when BAM just doesn't cut it (eg a chromosome more than 2Gb long).  Does htsjdk support this? I haven't got TBI support for it, but arguably it's not helpful anyway given it also has the same size limit.

jrobinso:
Ping,  @lbergelson et al I wonder what the status of CSI support is?    Just asking so I know how to respond to IGV tickets,  not requesting anything.

valeriuo:
I don't have anything else to add to the PR, at this point. I did some basic tests with [Picard](https://github.com/broadinstitute/picard/pull/998) and they looked OK.

lbergelson:
I would like to get this into the next release, but I've been saying that to myself for every release...

src/main/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -43,49 +43,73 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
         // look at all chunk offsets in a linear index to make sure they are valid
 
         if (reader.indexing().hasBrowseableIndex()) {
+            if (SamIndexes.BAI.fileNameSuffix.endsWith(reader.type().indexExtension())) {
```
lbergelson: There's a lot of duplicated code here.  It seems like it could be deduplicated it by using AbstractBamReader and then adding an instanceof check around the linear index checking.  That's kind of gross too though...  this is probably just as good.

src/main/java/htsjdk/samtools/BamIndexValidator.java
```diff
@@ -100,50 +124,79 @@ public static int exhaustivelyTestIndex(final SamReader reader) { // throws Exce
...
+                                iter.close();
+                            }
+                        } catch (final Exception e) {
+                            throw new SAMException("Exception in BamIndexValidator. Linear index access failure " + l + " indexCount=" + indexCount, e);
+                        }
                     }
                 }
+                return chunkCount;
+            } else if (SamIndexes.CSI.fileNameSuffix.endsWith(reader.type().indexExtension())) {
+
```
lbergelson: This branch isn't run at all, because there's no explicit test for lessExhaustively check and we don't run SamValidator tests on anything with  a CSI index.  It would be good to add a test for this case since it's  fairly complicated.

src/main/java/htsjdk/samtools/BinWithOffset.java
```diff
@@ -0,0 +1,21 @@
+package htsjdk.samtools;
+
+import htsjdk.samtools.Bin;
+
+public class BinWithOffset extends Bin {
+
+    private long lOffset;
```
lbergelson: If there isn't any need for changing it I'd make this final and remove the setter.  Always simpler that way.

src/main/java/htsjdk/samtools/BinWithOffset.java
```diff
@@ -0,0 +1,21 @@
+package htsjdk.samtools;
+
+import htsjdk.samtools.Bin;
+
+public class BinWithOffset extends Bin {
```
lbergelson: Could you add javadoc to this class explaining it's use case.

src/main/java/htsjdk/samtools/CSIIndex.java
```diff
@@ -0,0 +1,466 @@
...
+
+import htsjdk.samtools.seekablestream.SeekableStream;
+import htsjdk.samtools.util.RuntimeIOException;
+
+import java.io.File;
+import java.nio.file.Path;
+import java.util.*;
+
+
+public class CSIIndex extends AbstractBAMFileIndex implements BrowseableBAMIndex {
```
lbergelson: Could you add class javadoc that explains what the CSI index is and gives some intuition about how it's different from the Bai index.

src/main/java/htsjdk/samtools/CSIIndex.java
```diff
@@ -0,0 +1,466 @@
...
+
+    private int binDepth;
```
lbergelson: I'd love to make these final but I also want the methods that set them to not just be inlined into the constructor... I hate that java makes me choose.

src/main/java/htsjdk/samtools/CSIIndex.java
```diff
@@ -0,0 +1,466 @@
...
+    /**
+     * Constructors
+     */
+
+    public CSIIndex(final SeekableStream stream, final SAMSequenceDictionary dictionary) {
+        this(new IndexStreamBuffer(stream), stream.getSource(), dictionary);
+    }
+
+    public CSIIndex(final Path path, boolean enableMemoryMapping, final SAMSequenceDictionary dictionary) {
+        this(IndexFileBufferFactory.getBuffer(path.toFile(), enableMemoryMapping), path.toString(), dictionary);
```
lbergelson: It's not safe to do path.toFile because many paths will not be representable as files.  This constructor should create a SeekableStream from the path instead.  And it probably shouldn't have the memory mapping parameter after the change since I don't believe SeekableStream respects it.

src/main/java/htsjdk/samtools/CSIIndex.java
```diff
@@ -0,0 +1,466 @@
...
+    }
+
+    public CSIIndex(final File file, boolean enableMemoryMapping, final SAMSequenceDictionary dictionary) {
+        this(IndexFileBufferFactory.getBuffer(file, enableMemoryMapping), file.getName(), dictionary);
+    }
+
+    private CSIIndex(final IndexFileBuffer indexFileBuffer, final String source, final SAMSequenceDictionary dictionary) {
+        super(indexFileBuffer, dictionary);
+
+        verifyCSIMagicNumber(source);
```
lbergelson: I think we should probably make this a protected method on AbstractBamFileIndex and implement it there for bai and here for CSI.  Or we maybe make it abstract and pull out a common AbstractBaiFileIndex that is shared in common between the 2 bai implementations.  It would let us clean up the constructor chain in AbstractBamFileIndex so there isn't any constructor you can call that won't validate the magic number.

src/main/java/htsjdk/samtools/CompressedIndexFileBuffer.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Override
+    public int position() {
+        return (int)mCompressedStream.getPosition();
+    }
+
+    @Override
+    public void close() {
+        try {
+            mCompressedStream.close();
```
lbergelson: this should check for null incase the constructor failed

src/main/java/htsjdk/samtools/CompressedIndexFileBuffer.java
```diff
@@ -0,0 +1,84 @@
...
+        try {
+            mCompressedStream.seek(position);
+        } catch (IOException ioe) {
+            throw(new RuntimeIOException("Seek error in CSI compressed stream: " + ioe));
+        }
+    }
+
+    @Override
+    public int position() {
+        return (int)mCompressedStream.getPosition();
```
lbergelson: This should probably check for > maxInt and throw in that case.  Unlikely with an index file... but you never know...

src/main/java/htsjdk/samtools/CompressedIndexFileBuffer.java
```diff
@@ -0,0 +1,84 @@
...
+            mCompressedStream = new BlockCompressedInputStream(file);
+        } catch (IOException ioe) {
+            throw(new RuntimeIOException("Construction error of CSI compressed stream: " + ioe));
+        }
+    }
+
+    @Override
+    public void readBytes(final byte[] bytes) {
+        try {
+            mCompressedStream.read(bytes);
```
lbergelson: I think there's a subtle bug here.   `mCompressedStream.read(bytes)` makes no guarantee that it will read as many bytes as you ask for.  You have to check the return value of read and continue reading until you either have read as many bytes as you require, or it returns -1 in which case you should probably throw.

src/main/java/htsjdk/samtools/CompressedIndexFileBuffer.java
```diff
@@ -0,0 +1,84 @@
+package htsjdk.samtools;
+
+import htsjdk.samtools.util.BlockCompressedInputStream;
+import htsjdk.samtools.util.RuntimeIOException;
+
+import java.io.File;
+import java.io.IOException;
+
+
+class CompressedIndexFileBuffer implements IndexFileBuffer {
```
lbergelson: needs class level javadoc

src/main/java/htsjdk/samtools/CompressedIndexFileBuffer.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private BlockCompressedInputStream mCompressedStream;
```
lbergelson: make this final

src/main/java/htsjdk/samtools/IndexFileBuffer.java
```diff
@@ -0,0 +1,13 @@
+package htsjdk.samtools;
+
+import java.io.Closeable;
+
+interface IndexFileBuffer extends Closeable {
```
lbergelson: This should probably get some javadoc.

src/main/java/htsjdk/samtools/CSIIndex.java
```diff
@@ -0,0 +1,466 @@
...
+                for (int ci = 0; ci < nChunks; ci++) {
+                    final long chunkBegin = readLong(); // chunk_beg
+                    final long chunkEnd = readLong(); // chunk_end
+                    lastChunk = new Chunk(chunkBegin, chunkEnd);
+                    chunks.add(lastChunk);
+                }
+            } else if (indexBin == getMaxBins() + 1) {
+                // meta data - build the bin so that the count of bins is correct;
+                // but don't attach meta chunks to the bin, or normal queries will be off
+                for (int ci = 0; ci < nChunks; ci++) {
```
lbergelson: Lets pull out a `readChunks` in the superclass and use it in all 4 places this block of code is repeated.

src/main/java/htsjdk/samtools/IndexStreamBuffer.java
```diff
@@ -0,0 +1,84 @@
...
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+class IndexStreamBuffer implements IndexFileBuffer {
+    private final SeekableStream in;
+    private final ByteBuffer tmpBuf;
+
+    /** Continually reads from the provided {@link SeekableStream} into the buffer until the specified number of bytes are read, or
+     * until the stream is exhausted, throwing a {@link RuntimeIOException}. */
+    private static void readFully(final SeekableStream in, final byte[] buffer, final int offset, final int length) {
```
lbergelson: It might make sense to use BinaryCodec instead of reimplementing readFully.

src/main/java/htsjdk/samtools/CompressedIndexFileBuffer.java
lbergelson: You need something like readFully or readBytes from BinaryCodec here I think.

src/main/java/htsjdk/samtools/CompressedIndexFileBuffer.java
```diff
@@ -0,0 +1,84 @@
...
+        try {
+            mCompressedStream.read(bytes);
+        } catch (IOException ioe) {
+            throw(new RuntimeIOException("Read error in CSI compressed stream: " + ioe));
+        }
+    }
+
+    @Override
+    public int readInteger() {
+        final byte[] intbuff = new byte[4];
```
lbergelson: Allocating a new byte[] for each integer read is going to be super inefficient.   It would make sense to reuse the buffer.  This method suffers from the same problem with read not guaranteeing a definite number of bytes.

src/main/java/htsjdk/samtools/CSIIndex.java
```diff
@@ -0,0 +1,466 @@
...
+        }
+        int level = getLevelForBin(bin);
+        int firstBinOnLevel = getFirstBinOnLevel(level);
+        int levelSize = getLevelSize(level);
+
+        return (bin.getBinNumber() - firstBinOnLevel)*(getMaxSpan()/levelSize) + 1;
+    }
+
+    @Override
+    public int getLastLocusInBin(Bin bin) {
```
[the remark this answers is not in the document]
valeriuo: Unfortunately, that was the constraint.

src/main/java/htsjdk/samtools/CompressedIndexFileBuffer.java
valeriuo: Good call on this. BGZF compressed streams use virtual offsets, which are 64 bit by construction and usually exceed `Integer.MAX_VALUE`. I will change `position` return type and `seek` argument type to `long`.

src/main/java/htsjdk/samtools/BamIndexValidator.java
valeriuo: The discrepancy between the two BAI indexes appears to be caused by the fact that htslib uses a bin reduction [method](https://github.com/samtools/htslib/blob/bcf9bff178f81c9c1cf3a052aeb6cbe32fe5fdcc/hts.c#L1436), not implemented in htsjdk. However, both index files are functional and porting this piece of code should not be part of this PR. Thus, I kept the original BAI file.

valeriuo:
I've addressed most of the change requests and also added a new test class `BAMFileReaderTest`, that makes use of the query methods in the `BAMFileReader`, for comparing the results of BAI and CSI indexes.
The test harness currently fails due to some connection issues with the Broad ftp server.

PR #1041, opened 2017-11-26
Add hasIndex to FeatureReader interface and expose it in VcfFileReader

yfarjoun:
- add hasIndex to FeatureReader interface
- expose it as isQueryable in VcfFileReader
- add constants for various vcf suffixes 
- replace many of them in the code
- added tests covering isQueriable

### Description
The only way right now to find out if a VcfReader is queryable it to try and catch on failure. This is not a good way of doing things. Since the underlaying Tribble Reader knows if it has an index or not, and further more all the (internal) implementations of FeatureReader have a method called hasIndex, adding this method to the interface (with default false) should not change the API but it will enable the wrapping readers to know if they can query with an index or not.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


src/main/java/htsjdk/tribble/FeatureReader.java
```diff
@@ -27,16 +27,48 @@
...
+    /**
+     * Provides the list of sequenceNames if known. Otherwise will return an empty list.
+     * @return  the list of sequenceNames if known. Otherwise will return an empty list.
+     */
+    List<String> getSequenceNames();
 
-    public Object getHeader();
+    /**
+     * Provide access to the header of the reader
+     * @return the header of the reader
```
magicDGS: It can return `null` if there is no header, no? If so, add to the javadoc...

src/main/java/htsjdk/tribble/example/CountRecords.java
```diff
@@ -188,7 +188,7 @@ public static Index createAndWriteNewIndex(File featureFile, File indexFile, Fea
      */
     public static FeatureCodec getFeatureCodec(File featureFile) {
         // quickly determine the codec type
-        //if (featureFile.getName().endsWith(".vcf") || featureFile.getName().endsWith(".VCF") )
+        //if (featureFile.getName().endsWith(IOUtil.VCF_FILE_EXTENSION) || featureFile.getName().endsWith(".VCF") )
```
magicDGS: This change can be removed. By the way, this class can be mark as `@Deprecated` and removed in a following release, no? I think that no one look at example code in HTSJDK, or if so a new repository or a clear package for it might be better (although the Wiki can be used for such kind of examples....

src/main/java/htsjdk/variant/vcf/VCFFileReader.java
```diff
@@ -159,4 +156,8 @@ public void close() {
 			throw new TribbleException("Could not close a variant context feature reader.", ioe);
 		}
 	}
+
+    public boolean isQueriable() {
```
magicDGS: javadoc

magicDGS: And maybe can be used in the `query` method, to throw an informative exception if it cannot be queried..

src/main/java/htsjdk/variant/vcf/VCFUtils.java
```diff
@@ -157,21 +159,77 @@ public static VCFHeader withUpdatedContigs(final VCFHeader oldHeader, final File
...
     }
 
+    /**
+     * This method creates a temporary VCF file and it's appropriately named index file, and will delete them on exit.
+     * @param prefix - The prefix string to be used in generating the file's name; must be at least three characters long
+     * @param suffix - The suffix string to be used in generating the file's name; may be null, in which case the suffix ".tmp" will be used
+     * @return A File object referencing the newly created temporary VCF file
+     * @throws IOException - if a file could not be created.
+     */
+    public static File createTemporaryIndexedVcfFile(final String prefix, final String suffix) throws IOException {
```
magicDGS: I'll opt to return a `java.nio.Path` to continue moving forward to `Path` representation of files insted of `File`...

src/main/java/htsjdk/variant/vcf/VCFUtils.java
```diff
@@ -157,21 +159,77 @@ public static VCFHeader withUpdatedContigs(final VCFHeader oldHeader, final File
...
+    }
+
+    /**
+     * This method makes a copy of the input VCF and creates an index file for it in the same location.
+     * This is done so that we don't need to store the index file in the same repo
+     * The copy of the input is done so that it and its index are in the same directory which is typically required.
+     * @param vcfFile the vcf file to index
+     * @return File a vcf file (index file is created in same path).
+     */
+    public static File createTemporaryIndexedVcfFromInput(final File vcfFile, final String tempFilePrefix) throws IOException {
```
magicDGS: Same as before...

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -87,8 +88,13 @@
     public static final long FIVE_GBS = 5 * ONE_GB;
 
     public static final String VCF_FILE_EXTENSION = ".vcf";
+    public static final String VCF_INDEX_EXTENSION = Tribble.STANDARD_INDEX_EXTENSION;
```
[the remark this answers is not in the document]
yfarjoun: I think it would be good to have some traction on the hts-specs issue first...

src/main/java/htsjdk/tribble/FeatureReader.java
```diff
@@ -27,16 +27,48 @@
...
+    /**
+     * Query the reader for a particular interval corresponding to a contig and a 1-based closed
+     *
+     * @param chr the contig to be queried
+     * @param start the start of the interval (1-based) to be queried
+     * @param end the last base in the interval to be queried
+     * @return an iterator containing the features that at in the interval.
+     * @throws IOException If there's a problem reading or if the reader is not queryable, e.g. if it doesn't have an index.
+     */
+    CloseableTribbleIterator<T> query(final String chr, final int start, final int end) throws IOException;
```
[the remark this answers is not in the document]
yfarjoun: ok.

src/main/java/htsjdk/tribble/FeatureReader.java
yfarjoun: ok.

src/main/java/htsjdk/tribble/example/CountRecords.java
yfarjoun: not in this PR...

src/main/java/htsjdk/variant/vcf/VCFFileReader.java
yfarjoun: not sure what you mean in the second comment.

src/main/java/htsjdk/variant/vcf/VCFUtils.java
yfarjoun: enough parts still do not support path...so this is still non-trivial...I'd prefer not to mix up the different purposes.

src/main/java/htsjdk/variant/vcf/VCFUtils.java
yfarjoun: same here.

src/main/java/htsjdk/variant/vcf/VCFFileReader.java
magicDGS: That the query method catches IOE to throw a TribbleException. Now it can use this method to throw a TribbleException with a more informative error message (without changing the contract)

src/main/java/htsjdk/variant/vcf/VCFUtils.java
magicDGS: If the code moves to Path for parts using this method, the File version should be replicated introducing unexpected problems (as unfortunately I did with the sorting collections). Thus, it's simpler to implement directly methods with Path and use the Path.toFile/File.toPath methods if required... to be sure that it's less painful to evolve to a pure Path htsjdk (something that I'm currently  kind of working on)...

src/main/java/htsjdk/tribble/FeatureReader.java
```diff
@@ -27,16 +29,60 @@
...
+    /**
+     * Provide access to the header of the reader
+     * @return the header of the reader. May be null.
+     */
+    Object getHeader();
 
+    /**
+     * @return true if the reader has an index, which means that it can be queried.
+     */
+    default boolean hasIndex() {
```
lbergelson: I think we might want to separate out the concepts of `hasIndex` and `isQueriable`.  In gatk we have a datasource that has no index but supports random access.   I also think `supportsRandomAccess` might be clearer name.  I'm not sure how granular we want to go though, you could potentially imagine a stream with an index that supports fast forward skipping but not rewinding, which wouldn't be easily described here either.

yfarjoun: hmmm. thats interesting...the ReferenceFile implementation is basically like that...you can move forwards but not backwards in a reference.

src/main/java/htsjdk/variant/vcf/VCFUtils.java
yfarjoun: The problem is VariantContextWriterBuilder.setOutputFile(File).... 

I don't want this PR to become a "get htsjdk to write in Path" as well....

src/main/java/htsjdk/tribble/FeatureReader.java
yfarjoun: are you referring to the comment (which ties the two concepts together) or the default implementation (that does the same)?

I think that we should support implementations that exist...and possibly ones that we can easily support...but not implementations that one "could potentially imagine"....

in GATK, is that datasource implemented as a FeatureReader? or on top of one of the standard readers?

src/main/java/htsjdk/variant/vcf/VCFUtils.java
magicDGS: ok...

yfarjoun:
@lbergelson is this better now, or do you think that a more detailed queriable return type would be preferable?

src/main/java/htsjdk/variant/vcf/VCFFileReader.java
```diff
@@ -308,4 +302,14 @@ public void close() {
...
         }
     }
+
+    /**
+     * a method to check if the reader is queriable, i.e. if a call to {@link VCFFileReader#query(String, int, int)}
+     * can be successful
+     *
+     * @return true if the reader can be quried, i.e. if the underlying Tribble reader has an index.
+     */
+    public boolean isQueriable() {
```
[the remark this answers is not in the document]
yfarjoun: just because more people are searching for queryable doesn't make it more correct...but OK. will change.

src/main/java/htsjdk/tribble/FeatureReader.java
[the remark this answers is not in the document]
yfarjoun: This is in the API...do you really want me to change it now?

src/main/java/htsjdk/variant/vcf/VCFFileReader.java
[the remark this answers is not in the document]
lbergelson: I'm fine either way. I'll definitely  get it confused everytime though.

src/main/java/htsjdk/tribble/FeatureReader.java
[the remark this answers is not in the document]
lbergelson: Didn't you just add this to the api? It's part of the abstract feature reader, but I don't see it as a huge problem if that declares two methods..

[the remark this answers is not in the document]
yfarjoun: I guess I'm a little confused since there's already a has index method in many implementations...

lbergelson:
@yfarjoun Looks good to me 👍

PR #1043, opened 2017-11-29
Add checks on Histogram::getPercentile to protect against total sum…

yfarjoun:
… equalling zero and negative values.

### Description
a recent (short-lived) picard issue showed that it is possible to get to an unreasonable part of the code of getPercentile if the histogram contains values that do not lend themselves to calculating percentiles...this PR adds protective code to error out more gracefully in that case.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/Histogram.java
+++ b/src/main/java/htsjdk/samtools/util/Histogram.java
@@ -324,17 +324,17 @@ public final class Histogram<K extends Comparable> implements Serializable {
                 .filter(b -> b.getValue() < 0)
                 .findFirst()
                 .ifPresent(b -> {
-            throw new IllegalArgumentException("Cannot calculate Percentile when negative counts are present " +
+            throw new IllegalStateException("Cannot calculate Percentile when negative counts are present " +
                     "in histogram. Bin " + b.getId() + "=" + b.getValue());
         });
 
         final double total = getCount();
-        if (total == 0) throw new IllegalArgumentException("Cannot calculate percentiles when total is zero.");
+        if (total == 0) throw new IllegalStateException("Cannot calculate percentiles when total is zero.");
 
-        double sofar = 0;
+        double soFar = 0;
         for (Bin<K> bin : values()) {
-            sofar += bin.getValue();
-            if (sofar / total >= percentile) return bin.getIdValue();
+            soFar += bin.getValue();
+            if (soFar / total >= percentile) return bin.getIdValue();
         }
 
         throw new IllegalStateException("UNPOSSIBLE! Could not find percentile: " + percentile);
--- a/src/main/java/htsjdk/samtools/util/Histogram.java
+++ b/src/main/java/htsjdk/samtools/util/Histogram.java
@@ -24,6 +24,8 @@
 
 package htsjdk.samtools.util;
 
+import htsjdk.samtools.SamReaderFactory;
+
 import java.io.Serializable;
 import java.util.*;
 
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/util/Histogram.java
```diff
@@ -317,14 +320,24 @@ public double getPercentile(final double percentile) {
...
-        double total = getCount();
+        values().stream()
+                .filter(b -> b.getValue() < 0)
+                .findFirst()
+                .ifPresent(b -> {
+            throw new IllegalArgumentException("Cannot calculate Percentile when negative counts are present " +
+                    "in histogram. Bin " + b.getId() + "=" + b.getValue());
+        });
+
+        final double total = getCount();
```
magicDGS: `getCount()` iterates also every value, the same as the check for the negative values. I will recommend for efficiency (this PR adds a third iteration over all the values) to do it in one go:

```java
double total = 0;
for (final Bin<K> b : values()) {
    if (b.getValue() < 0) {
        throw new IllegalArgumentException("Cannot calculate Percentile when negative counts are present " +
                "in histogram. Bin " + b.getId() + "=" + b.getValue());
    }
    total += bin.getValue();
}
```

This will safe an extra iteration...

yfarjoun: no. since there's an early return that might happen before the iteration concludes....

magicDGS: For a call on a histogram with correct values, the check and getCount perform two complete iterations, and it's when it is supposed to be used. The last iteration might return early, but before the values were iterated twice....

lbergelson: @yfarjoun I think @magicDGS is suggesting to replace the getCount() call with a for-loop that also checks nothing is negative, not to add the check to the existing for loop that finds the bin, which would potentially terminate to soon.

It seems like a reasonable optimization.  I don't think it's particularly necessary because I don't think there's any software that has a bottleneck on `getPercentile()`, so it may make more sense to leave the checks independent for clarity.

src/main/java/htsjdk/samtools/util/Histogram.java
```diff
@@ -317,14 +320,24 @@ public double getPercentile(final double percentile) {
         if (percentile <= 0) throw new IllegalArgumentException("Cannot query percentiles of 0 or below");
         if (percentile >= 1) throw new IllegalArgumentException("Cannot query percentiles of 1 or above");
 
-        double total = getCount();
+        values().stream()
+                .filter(b -> b.getValue() < 0)
+                .findFirst()
+                .ifPresent(b -> {
+            throw new IllegalArgumentException("Cannot calculate Percentile when negative counts are present " +
```
lbergelson: I think these should be `IllegalStateException`s because they depend on the state of the `Histogram` object and not the parameters the function is being called with.

src/main/java/htsjdk/samtools/util/Histogram.java
yfarjoun: the issue is that the Histogram class might be used with negative values...and in those cases getCount should work. Its only in the case that you need to convert a Histogram to a density that you need the values to be non-negative and that the count will be positive...

src/main/java/htsjdk/samtools/util/Histogram.java
yfarjoun: reasonable.

yfarjoun:
back to you @lbergelson

PR #1046, opened 2017-12-04
Do not accumulate entire header in StringBuilder.

### LINKED ISSUE #1045: SAMTextHeaderCodec should not accumulate entire header in StringBuilder
From a thread on our gitter channel

> Also, I should've mention that my BAM is kind of special in the way that there are about 39M contigs (i.e. reference sequences) in the header, it's a result of an alignment against a very fragmented assembly.

To reproduce, I added 1,000,000 @SG records to the header for a 1G BAM file from 1000G
```
17/11/30 13:01:21 ERROR SparkUncaughtExceptionHandler: [Container in shutdown] Uncaught exception in thread Thread[Executor task launch worker for task 2,5,main]
java.lang.OutOfMemoryError: Java heap space
    at java.util.Arrays.copyOf(Arrays.java:3332)
    at java.lang.AbstractStringBuilder.expandCapacity(AbstractStringBuilder.java:137)
    at java.lang.AbstractStringBuilder.ensureCapacityInternal(AbstractStringBuilder.java:121)
    at java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:421)
    at java.lang.StringBuilder.append(StringBuilder.java:136)
    at htsjdk.samtools.SAMTextHeaderCodec.advanceLine(SAMTextHeaderCodec.java:139)
    at htsjdk.samtools.SAMTextHeaderCodec.decode(SAMTextHeaderCodec.java:94)
    at htsjdk.samtools.BAMFileReader.readHeader(BAMFileReader.java:667)
    at htsjdk.samtools.BAMFileReader.<init>(BAMFileReader.java:328)
    at htsjdk.samtools.BAMFileReader.<init>(BAMFileReader.java:268)
    at htsjdk.samtools.SamReaderFactory$SamReaderFactoryImpl.open(SamReaderFactory.java:317)
    at org.seqdoop.hadoop_bam.BAMRecordReader.createSamReader(BAMRecordReader.java:198)
    at org.seqdoop.hadoop_bam.BAMRecordReader.initialize(BAMRecordReader.java:147)
    at org.apache.spark.rdd.NewHadoopRDD$$anon$1.liftedTree1$1(NewHadoopRDD.scala:182)
    at org.apache.spark.rdd.NewHadoopRDD$$anon$1.<init>(NewHadoopRDD.scala:179)
    at org.apache.spark.rdd.NewHadoopRDD.compute(NewHadoopRDD.scala:134)
    at org.apache.spark.rdd.NewHadoopRDD.compute(NewHadoopRDD.scala:69)
    at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:323)
    at org.apache.spark.rdd.RDD.iterator(RDD.scala:287)
    at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:38)
    at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:323)
    at org.apache.spark.rdd.RDD.iterator(RDD.scala:287)
    at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:38)
    at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:323)
    at org.apache.spark.rdd.RDD.iterator(RDD.scala:287)
    at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:87)
    at org.apache.spark.scheduler.Task.run(Task.scala:108)
    at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:335)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
    at java.lang.Thread.run(Thread.java:745)
```

Relevant code at
https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/samtools/SAMTextHeaderCodec.java#L56
https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/samtools/SAMTextHeaderCodec.java#L139

heuermh:
Fixes #1045.

I haven't a clue why the entire header was accumulated in a `StringBuilder`, but removing the line that did so still passes all unit tests.

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMTextHeaderCodec.java
+++ b/src/main/java/htsjdk/samtools/SAMTextHeaderCodec.java
@@ -136,7 +136,6 @@ public class SAMTextHeaderCodec {
             return null;
         }
         mCurrentLine = mReader.readLine();
-        textHeader.append(mCurrentLine).append('\n');
         return mCurrentLine;
     }
 


src/main/java/htsjdk/samtools/SAMTextHeaderCodec.java
```diff
@@ -136,7 +136,6 @@ private String advanceLine() {
             return null;
         }
         mCurrentLine = mReader.readLine();
-        textHeader.append(mCurrentLine).append('\n');
```
nh13: It's set in SAMFileHeader.java if there's an error parsing the header.  👎

heuermh: Point me to the unit test that covers that situation.

nh13: That kind of attitude is not going to earn any internet points here.  Not everything is covered by a unit test (fortunately or unfortunately). Taking 10 seconds to search the source your changing would have revealed it.  I'll let @tfenne @jacarey and @lbergelson to take over from here.

heuermh: > That kind of attitude is not going to earn any internet points here. 

Based on past experience with you and others in this repository I felt the only way to fight snark was with more snark.  I apologize.

Let me try again, how about instead of collecting the entire header in RAM on the off chance that it might need to be reported in an error condition, track only the line number, and then include that with the offending header line when reporting the error?

nh13: @heuermh I have a bolder proposal:
1. Deprecate `getTextHeader()` in `SAMFileHeader` and change it to `public String getTextHeader() { return null; }`.  I see only Picard's `ViewSam` tool use this method, and it does the right thing if `null` is returned.
2. Remove any reference to `textHeader` in `SAMTextHeaderCodec`.  We don't need it to report errors, since we store line numbers and the error message using `reportErrorParsingLine`.

lbergelson: @heuermh Thank you for identifying and reporting the problem.  It definitely looks like this is a vestigial method we don't need.  I checked gatk as well and didn't find any uses.  

I like @nh13's proposal to fix the problem.  Much better to deprecate and remove the unnecessary stuff than just to disconnect it and leave it hanging around in the code.  @heuermh would you be able to revise the PR?

heuermh: It appears that we call `SAMFileHeader.getTextHeader` in Hadoop-BAM at [SAMRecordReader.java#L196](https://github.com/HadoopGenomics/Hadoop-BAM/blob/f2e9595d35ea2118e3c56ebeedc29cd432ed4b9e/src/main/java/org/seqdoop/hadoop_bam/SAMRecordReader.java#L196), in a class full of workarounds for various issues with SAMFileReader.  I'll check if these changes cause trouble there.

What is htsjdk's policy on deprecation, versioning, and binary compatibility?  Typically I would annotate as @deprecated in a minor or point release without code change and then remove on next major release.

I'll update this PR as necessary, squash and rebase with force pushes ok?

tfenne: Maybe rather than returning `null` the deprecated method could `return getSAMString()` instead, which will reconstruct the header as a String from the internal state?  Otherwise I agree with @nh13's approach.

@heuermh: we tend to prefer not squashing until the final merge if you don't mind please.

lbergelson: @heuermh We don't have a really well defined deprecation policy.  In general we try to deprecate without breaking the method and let that hang out for a few minor versions before removing any functionality.  We avoid any incompatible changes between patch releases, but minor releases change the api in non-backward's compatible ways.  

Since this method already has a lot of caveats about when it returns a value and when it returns null, callers should already be able to handle the null safely without loss of functionality.  In any case though, @tfenne's suggestion is better than returning null.

yfarjoun: @heuermh are you going to get back to this? it seems like the way forward was resolved a few months back.

Thanks!

yfarjoun:
A full solution would be able to report the line number in the case of an error in the header.

Closing due to lack of activity. Relevant issue is #1045

PR #1051, opened 2017-12-08
EBI_REFERENCE_SERVICE_URL_MASK to use https.

raskoleinonen:
The EBI CRAM reference registry has been recently changed to use https instead of http.

The EBI_REFERENCE_SERVICE_URL_MASK should be changed to reflect this.

Best Regards,
Rasko Leinonen
European Nucleotide Archive (ENA)

### Description

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/Defaults.java
+++ b/src/main/java/htsjdk/samtools/Defaults.java
@@ -110,7 +110,7 @@ public class Defaults {
         }
         REFERENCE_FASTA = getFileProperty("reference_fasta", null);
         USE_CRAM_REF_DOWNLOAD = getBooleanProperty("use_cram_ref_download", false);
-        EBI_REFERENCE_SERVICE_URL_MASK = "http://www.ebi.ac.uk/ena/cram/md5/%s";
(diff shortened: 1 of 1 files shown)


yfarjoun:
is there a formal announcement that this is indeed the case? Could you please link to it here?

raskoleinonen:
I am writing in the capacity of the technical coordinator of the EBI CRAM reference registry service.

There has been no CRAM reference registry specific announcement of this change. This was an oversight.

The modification was done as part of replacing all www.ebi.ac.uk HTTP access points with HTTPS equivalents.

If you prefer I can send a formal announcement e.g. to ena-announce mailing list right away. I can send an announcement in any channel you would find acceptable.

Please note that the CRAM reference registry official document has been updated to refer to HTTPS instead of HTTP:

{code}
https://www.ebi.ac.uk/ena/cram/md5/<hashvalue>
https://www.ebi.ac.uk/ena/cram/sha1/<hashvalue>
{code}

We have advised our users on a case-by-case basis to switch to the HTTPS protocol. These discussions have been prompted by users complaining that the HTTP protocol no longer worked.

This pull request has been created as a result of an investigation into such problem.

PR #1052, opened 2017-12-08
Modify FilterTag to Pass or Filter reads

jsotobroad:
### Description

Some people would like to select reads based on tag values and not just filter.  This allows this filter to both pass and filter reads.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/filter/TagFilter.java
+++ b/src/main/java/htsjdk/samtools/filter/TagFilter.java
@@ -37,6 +37,7 @@ public class TagFilter implements SamRecordFilter {
 
     private final String tag;           // The key of the tag to match
     private final List<Object> values;  // The list of matching values
+    private boolean includeReads = false;
 
     /**
      * Constructor for a single value
@@ -44,9 +45,10 @@ public class TagFilter implements SamRecordFilter {
      * @param tag       the key of the tag to match
      * @param value     the value to match
      */
-    public TagFilter(String tag, Object value) {
+    public TagFilter(String tag, Object value, final boolean includeReads) {
         this.tag = tag;
         this.values = Arrays.asList(value);
+        this.includeReads = includeReads;
     }
 
     /**
@@ -55,9 +57,10 @@ public class TagFilter implements SamRecordFilter {
      * @param tag       the key of the tag to match
      * @param values    the matching values
      */
-    public TagFilter(String tag, List<Object> values) {
+    public TagFilter(String tag, List<Object> values, final boolean includeReads) {
         this.tag = tag;
         this.values = values;
+        this.includeReads = includeReads;
     }
 
     /**
@@ -68,7 +71,16 @@ public class TagFilter implements SamRecordFilter {
      */
     @Override
     public boolean filterOut(SAMRecord record) {
-        return values.contains(record.getAttribute(tag));
+        if (includeReads) {
+            if (values.contains(record.getAttribute(tag))) {
+                return false;
+            }
+        } else {
+            if (!values.contains(record.getAttribute(tag))) {
+                return false;
+            }
+        }
+        return true;
     }
 
     /**
@@ -82,6 +94,15 @@ public class TagFilter implements SamRecordFilter {
     @Override
     public boolean filterOut(final SAMRecord first, final SAMRecord second) {
         // both first and second must have the tag in order for it to be filtered out
-         return values.contains(first.getAttribute(tag)) && values.contains(second.getAttribute(tag));
+        if (includeReads) {
+            if (values.contains(first.getAttribute(tag)) && values.contains(second.getAttribute(tag))) {
+                return false;
+            }
+        } else {
+            if (!values.contains(first.getAttribute(tag)) && values.contains(second.getAttribute(tag))) {
+                return false;
+            }
+        }
+        return true;
     }
 }
--- a/src/main/java/htsjdk/samtools/filter/TagFilter.java
+++ b/src/main/java/htsjdk/samtools/filter/TagFilter.java
@@ -104,7 +104,8 @@ public class TagFilter implements SamRecordFilter {
      */
     @Override
     public boolean filterOut(final SAMRecord first, final SAMRecord second) {
-
+        // With includeReads==true, allow any pairs through that contain the tag value
+        // With includeReads==false, exclude pairs where both reads contain the tag value
         if (includeReads) {
             return !(values.contains(first.getAttribute(tag)) || values.contains(second.getAttribute(tag)));
         } else {
--- a/src/main/java/htsjdk/samtools/filter/TagFilter.java
+++ b/src/main/java/htsjdk/samtools/filter/TagFilter.java
@@ -37,7 +37,7 @@ public class TagFilter implements SamRecordFilter {
 
     private final String tag;           // The key of the tag to match
     private final List<Object> values;  // The list of matching values
-    private boolean includeReads = false;
+    private boolean includeReads;
 
     /**
      * Constructor for a single value
--- a/src/main/java/htsjdk/samtools/filter/TagFilter.java
+++ b/src/main/java/htsjdk/samtools/filter/TagFilter.java
@@ -25,7 +25,7 @@ package htsjdk.samtools.filter;
 
 import htsjdk.samtools.SAMRecord;
 
-import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 
 /**
@@ -37,7 +37,7 @@ public class TagFilter implements SamRecordFilter {
 
     private final String tag;           // The key of the tag to match
     private final List<Object> values;  // The list of matching values
-    private boolean includeReads = false;
+    private Boolean includeReads;
 
     /**
      * Constructor for a single value
@@ -46,8 +46,7 @@ public class TagFilter implements SamRecordFilter {
      * @param value     the value to match
      */
     public TagFilter(String tag, Object value) {
-        this.tag = tag;
-        this.values = Arrays.asList(value);
+        this(tag, Collections.singletonList(value), null);
     }
 
     /**
@@ -57,8 +56,7 @@ public class TagFilter implements SamRecordFilter {
      * @param values    the matching values
      */
     public TagFilter(String tag, List<Object> values) {
-        this.tag = tag;
-        this.values = values;
+        this(tag, values, null);
     }
 
     /**
@@ -68,10 +66,8 @@ public class TagFilter implements SamRecordFilter {
      * @param value         the value to match
      * @param includeReads  whether to include or not include reads that match filter
      */
-    public TagFilter(String tag, Object value, final boolean includeReads) {
-        this.tag = tag;
-        this.values = Arrays.asList(value);
-        this.includeReads = includeReads;
+    public TagFilter(String tag, Object value, final Boolean includeReads) {
+        this(tag, Collections.singletonList(value), includeReads);
     }
 
     /**
@@ -81,17 +77,17 @@ public class TagFilter implements SamRecordFilter {
      * @param values        the matching values
      * @param includeReads  whether to include or not include reads that match filter
      */
-    public TagFilter(String tag, List<Object> values, final boolean includeReads) {
+    public TagFilter(String tag, List<Object> values, final Boolean includeReads) {
         this.tag = tag;
         this.values = values;
-        this.includeReads = includeReads;
+        this.includeReads = includeReads == null ? false : includeReads;
     }
 
     /**
      * Determines whether a SAMRecord matches this filter
      *
      * @param record    the SAMRecord to evaluate
-     * @return  true if the SAMRecord matches the filter, otherwise false
+     * @return  the XOR of SAMRecord matches the filter and includeReads.
      */
     @Override
     public boolean filterOut(SAMRecord record) {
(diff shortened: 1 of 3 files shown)


src/main/java/htsjdk/samtools/filter/TagFilter.java
```diff
@@ -37,16 +37,18 @@
...
     private final List<Object> values;  // The list of matching values
+    private boolean includeReads;
 
     /**
      * Constructor for a single value
      *
      * @param tag       the key of the tag to match
      * @param value     the value to match
      */
-    public TagFilter(String tag, Object value) {
```
yfarjoun: keep the default constructor around to avoid breaking the API.

src/main/java/htsjdk/samtools/filter/TagFilter.java
```diff
@@ -68,7 +71,16 @@ public TagFilter(String tag, List<Object> values) {
      */
     @Override
     public boolean filterOut(SAMRecord record) {
-        return values.contains(record.getAttribute(tag));
+        if (includeReads) {
```
yfarjoun: wouldn't this also work with 
```java
return values.contains(record.getAttribute(tag))!=includeReads;
```
?
seems much simpler to follow

src/main/java/htsjdk/samtools/filter/TagFilter.java
```diff
@@ -82,6 +94,15 @@ public boolean filterOut(SAMRecord record) {
     @Override
     public boolean filterOut(final SAMRecord first, final SAMRecord second) {
         // both first and second must have the tag in order for it to be filtered out
-         return values.contains(first.getAttribute(tag)) && values.contains(second.getAttribute(tag));
```
yfarjoun: same here:
```java
return ( ... ) != includeReads
```

src/test/java/htsjdk/samtools/filter/TagFilterTest.java
```diff
@@ -69,10 +69,10 @@ public void testTagFilter(final String testName, final String tag, final List<Ob
...
     {
         return new Object[][]{
-            {"Basic positive test", ReservedTagConstants.XN, Arrays.asList(1), 1, true},
-            {"Multi-value positive test", ReservedTagConstants.XN, Arrays.asList(1,2,3), 1, true},
-            {"Incorrect value negative test", ReservedTagConstants.XN, Arrays.asList(1), 2, false},
-            {"Null value negative test", ReservedTagConstants.XN, Arrays.asList(1), null, false} 
+                {"Basic positive test", ReservedTagConstants.XN, Arrays.asList(1), 1, true, false},
+                {"Multi-value positive test", ReservedTagConstants.XN, Arrays.asList(1,2,3), 1, false, true},
+                {"Incorrect value negative test", ReservedTagConstants.XN, Arrays.asList(1), 2, false, false},
+                {"Null value negative test", ReservedTagConstants.XN, Arrays.asList(1), null, true, true}
```
yfarjoun: please add positive tests with includeReads==true

jsotobroad: isnt the 2nd test a positive test with includeReads==true?

yfarjoun: could be, but there were 4 tests before your changes...so it woudl be good if you left those unchanges, and added some to test the new functionality...I was reacting to the fact that the total number of tests didn't increase.

jsotobroad: I mean i will add more tests but I dont think it is always the the case that new functionality requires "more" tests if you can change the current tests to test both new and old functionality

src/main/java/htsjdk/samtools/filter/TagFilter.java
```diff
@@ -60,6 +61,32 @@ public TagFilter(String tag, List<Object> values) {
...
+    /**
+     * Constructor for a single value
+     *
+     * @param tag           the key of the tag to match
+     * @param value         the value to match
+     * @param includeReads  whether to include or not include reads that match filter
+     */
+    public TagFilter(String tag, Object value, final boolean includeReads) {
+        this.tag = tag;
+        this.values = Arrays.asList(value);
```
yfarjoun: might as well use a singleton list here.

src/main/java/htsjdk/samtools/filter/TagFilter.java
```diff
@@ -60,6 +61,32 @@ public TagFilter(String tag, List<Object> values) {
...
     }
 
+    /**
+     * Constructor for a single value
+     *
+     * @param tag           the key of the tag to match
+     * @param value         the value to match
+     * @param includeReads  whether to include or not include reads that match filter
+     */
+    public TagFilter(String tag, Object value, final boolean includeReads) {
```
yfarjoun: could you modify the old constructors to delegate to the new constructors instead of duplicating code?

infact, I'd only keep around the actual code for TagFilter(String tag, List<Object> values, final boolean includeReads) and everything else should delegate via `this(...,..,...)`

src/main/java/htsjdk/samtools/filter/TagFilter.java
```diff
@@ -81,7 +108,6 @@ public boolean filterOut(SAMRecord record) {
      */
     @Override
     public boolean filterOut(final SAMRecord first, final SAMRecord second) {
-        // both first and second must have the tag in order for it to be filtered out
```
yfarjoun: please correct the comment rather than remove it..

src/test/java/htsjdk/samtools/filter/TagFilterTest.java
```diff
@@ -49,8 +48,34 @@
      * @param testValue         The value to test for in the record
      * @param expectedResult    The expected result (true is the sequence should match the filter, otherwise false)
      */
-    @Test(dataProvider="data")
-    public void testTagFilter(final String testName, final String tag, final List<Object> validValues,
+    @Test(dataProvider="dataInclude")
+    public void testIncludeTagFilter(final String testName, final String tag, final List<Object> validValues,
+                              final Object testValue, final boolean expectedResult, final boolean includeReads) {
```
yfarjoun: all the data provided by the dataprovider has `includeReads==true`. it would be good to copy over the (implicit) false test cases from below and have them here. Also, you could use a Boolean (rather than a boolean) and have a third value (null) which could imply using the `TagFilter(tag, validValues) constructor, thus refactoring even further the tests (since that the only difference between the two tests)

src/main/java/htsjdk/samtools/filter/TagFilter.java
```diff
@@ -81,7 +108,6 @@ public boolean filterOut(SAMRecord record) {
      */
     @Override
     public boolean filterOut(final SAMRecord first, final SAMRecord second) {
-        // both first and second must have the tag in order for it to be filtered out
-         return values.contains(first.getAttribute(tag)) && values.contains(second.getAttribute(tag));
+        return (values.contains(first.getAttribute(tag)) && values.contains(second.getAttribute(tag))) != includeReads;
```
yfarjoun: hmmm. I'm actually not sure that this does the right thing...

if includeReads==false, and one has the tag/value and the other doesn't...should the pair be filtered or not?

at any rate...this should be tested. I know it wasn't already tested...but it should. could you please?

src/main/java/htsjdk/samtools/filter/TagFilter.java
jsotobroad: oh wow nice catch, didn't mean to do that, will remedy

src/main/java/htsjdk/samtools/filter/TagFilter.java
jsotobroad: 👍

jsotobroad: I think for now I will assume the tag value has to be on both (or vice versa not on both) in order to be filtered/allowed because thats how it originally was implemented.  Is there a case for this not to be the logic?

jsotobroad: actually can we talk about this when you have some time

jsotobroad:
@yfarjoun can you double check my logic for the Paired filter.  If it is what we discussed I would like to point out that it is different than the logic used in ReadNameFilter.java - https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/samtools/filter/ReadNameFilter.java#L106

src/test/java/htsjdk/samtools/filter/TagFilterTest.java
```diff
@@ -61,18 +121,109 @@ public void testTagFilter(final String testName, final String tag, final List<Ob
...
+        if (testValue != null) {
+            record1.setAttribute(tag, testValue);
+        }
+        final SAMRecord record2 = iterator.next();
+        if (matchPairs && testValue != null) {
+            record2.setAttribute(tag, testValue);
+        } else if (!matchPairs){
+            record2.setAttribute(tag, 0);
+        }
+        Assert.assertEquals(filter.filterOut(record1, record2), expectedResult, testName);
```
yfarjoun: shouldn't you also test `filterOut(SamRecord)` on it's own?

src/main/java/htsjdk/samtools/filter/TagFilter.java
```diff
@@ -77,11 +100,17 @@ public boolean filterOut(SAMRecord record) {
      * @param first  the first SAMRecord to evaluate
      * @param second the second SAMRecord to evaluate
      *
-     * @return true if the SAMRecords matches the filter, otherwise false
+     * @return the XOR of paired SAMRecord matches the filter and includeReads
```
yfarjoun: could you clarify the `@return` doc?

src/main/java/htsjdk/samtools/filter/TagFilter.java
```diff
@@ -77,11 +100,17 @@ public boolean filterOut(SAMRecord record) {
...
-        // both first and second must have the tag in order for it to be filtered out
-         return values.contains(first.getAttribute(tag)) && values.contains(second.getAttribute(tag));
+        // With includeReads==true, allow any pairs through that contain the tag value
+        // With includeReads==false, exclude pairs where both reads contain the tag value
+        if (includeReads) {
+            return !(values.contains(first.getAttribute(tag)) || values.contains(second.getAttribute(tag)));
+        } else {
+            return values.contains(first.getAttribute(tag)) && values.contains(second.getAttribute(tag));
+        }
+
```
yfarjoun: extra nl

yfarjoun:
sorry for the delay @jsotobroad back to you.

src/test/java/htsjdk/samtools/filter/TagFilterTest.java
jsotobroad: Ah yes I can just refactor all of my tests and do both at once 👍

jsotobroad:
@yfarjoun comments corrected and a bit more refactoring done for the tests if you could give another review

src/test/java/htsjdk/samtools/filter/TagFilterTest.java
```diff
@@ -38,41 +39,163 @@
...
+        if (matchPairs && testValue != null) {
+            record2.setAttribute(tag, testValue);
+        } else if (!matchPairs){
+            record2.setAttribute(tag, 0);
+        }
+        // Test paired reads
+        Assert.assertEquals(filter.filterOut(record1, record2), pairedExpectedResult, testName);
+    }
+
+    @Test(dataProvider="dataDefaultMatchingPairedFilter")
```
yfarjoun: These are all the same test! why not join all the data-providers and put comments describing what each "block" does?

jsotobroad: if i combine them its gonna be super annoying to figure out which test failed if one does

yfarjoun: hmmm. we have large data-providers all the time...it's not a problem...but if you like, you can include a String "name" in the input data and include it in the assertion. repeated tests that only differ by their names are worse (IMHO)

jsotobroad: I mean a test that "only differ by their names" mean the values woudl be the same...in which this case they are not.  Could you point me to one of these large data providers.

jsotobroad: i just don't think having to sift through 24 tests to figure out where the ` true != false` test failed seems like the most fun thing ever but if its done everywhere then i can change it.

yfarjoun: take a look at the tests for LiftOverVcf in picard for example.

yfarjoun: no sifting required...in the data object, add a string that decribes the data "DefaultMatchingPairedFilter" or whatever. and if the test fails include that string in the exception....

jsotobroad: ok

yfarjoun:
Thanks. I think that this is a good change. it's small enough that it can be added without further input. 

👍

jsotobroad:
thanks for the review and merging!

PR #1053, opened 2017-12-11
- replaced ">=" with "more than" in the description

yfarjoun:
replaced ">=" with "more than" in the description of Chained downsampling strategy so that it's possible to render it using the javadoc ''@value' pragma.

### Description

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/DownsamplingIteratorFactory.java
+++ b/src/main/java/htsjdk/samtools/DownsamplingIteratorFactory.java
@@ -53,7 +53,7 @@ public class DownsamplingIteratorFactory {
             "strategy to finish. Works in a single pass, and will provide accuracy close to (but often not as good as) HighAccuracy while requiring " +
             "memory proportional to the set of reads emitted from the ConstantMemory strategy to the HighAccuracy strategy. Works well when downsampling " +
             "large inputs to small proportions (e.g. downsampling hundreds of millions of reads and retaining only 2%. Should be accurate 99.9% of the time " +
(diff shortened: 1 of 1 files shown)


src/main/java/htsjdk/samtools/DownsamplingIteratorFactory.java
```diff
@@ -53,7 +53,7 @@
             "strategy to finish. Works in a single pass, and will provide accuracy close to (but often not as good as) HighAccuracy while requiring " +
             "memory proportional to the set of reads emitted from the ConstantMemory strategy to the HighAccuracy strategy. Works well when downsampling " +
             "large inputs to small proportions (e.g. downsampling hundreds of millions of reads and retaining only 2%. Should be accurate 99.9% of the time " +
-            "when the input contains >= 50,000 templates (read names). For smaller inputs, HighAccuracy is recommended instead.";
```
lbergelson: more than is slightly less precise, but I doubt anyone will be tripped up by the exactly equal case...

magicDGS: It is already late because it is accepted, but what's about using HTML code here? For example, `&ge;` will be converted to &ge; for HTML, markdown, etc. In that case, we don't have to worry about the exactly equal case...

yfarjoun:
I was hemming and hawing about that option and opted to go with the human readable version.

lbergelson:
I just hate having html like that in javadoc.  I wish javadoc would move to markdown or something similar.

Issue #1055, opened 2017-12-12
MergingIterator throws ConcurrentModificationException when closing mid-iteration


cwhelan:
`MergingIterator.close` throws a `ConcurrentModificationException` if it is called mid-iteration and it is merging over at least three sub-iterators.

While working on a tool based on GATK's `MultiVariantWalker` (building on htsjdk 2.13.1) that walks over multiple VCF files, I noticed that if my tool threw an exception while processing, instead of printing the application exception, it instead failed with a `java.util.ConcurrentModificationException` and swallowed the application exception. A typical displayed stacktrace is:

```
java.util.ConcurrentModificationException
	at java.util.PriorityQueue$Itr.next(PriorityQueue.java:535)
	at htsjdk.samtools.util.MergingIterator.close(MergingIterator.java:135)
	at org.broadinstitute.hellbender.engine.MultiVariantDataSource.closeOpenIterationIfNecessary(MultiVariantDataSource.java:368)
	at org.broadinstitute.hellbender.engine.MultiVariantDataSource.close(MultiVariantDataSource.java:223)
	at org.broadinstitute.hellbender.engine.MultiVariantWalker.onShutdown(MultiVariantWalker.java:103)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.runTool(CommandLineProgram.java:131)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMainPostParseArgs(CommandLineProgram.java:185)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:204)
	at org.broadinstitute.hellbender.engine.MultiVariantWalkerIntegrationTest.testWalkerThrowsException(MultiVariantWalkerIntegrationTest.java:272)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:108)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:661)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:869)
[...]

PR #1056, opened 2017-12-12
Fix MergingIterator.close() to avoid ConcurrentModificationException

cwhelan:
### Description

`MergingIterator.close()` could throw `ConcurrentModificationException` when called mid-iteration with at least three sub-iterators as described in https://github.com/samtools/htsjdk/issues/1055.

This PR changes the `close()` method to remove iterators from `MergingIterator`'s internal queue via an iterator rather than from the collection directly.

### Checklist

- [X] Code compiles correctly
- [X] All tests passing


cwhelan:
Thanks for the quick review, @tfenne . I don't have commit access to the master repo, could you do the merge for me?

PR #1058, opened 2017-12-20
Class for writing fasta format file.

alecw:
### Description

Class for writing fasta format file moved from private repo to public.  Enables use of this code in other public repositories.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/reference/FastaSequenceFileWriter.java
+++ b/src/main/java/htsjdk/samtools/reference/FastaSequenceFileWriter.java
@@ -1,0 +1,101 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009 The Broad Institute
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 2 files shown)


lindenb:
wouldn't it be better to use an interface + a factory pattern ? something like:

`new FastaWriterFactory().setLineLength(50).open(thefile).add("name","atgagatc").close();`

alecw:
My opinion is that making this change would add a bunch of additional code that ultimately won't provide much value because I don't see this class accumulating a ton of option like, e.g. SAMReader.  That said, if the consensus is that I should do it as Pierre suggests, I will do so.

> @lindenb: 
> wouldn't it be better to use an interface + a factory pattern ? something like:
> new FastaWriterFactory().setLineLength(50).open(thefile).add("name","atgagatc").close();

yfarjoun:
There are a few options that I see the value in being able to add/modify:

- output format (fasta / gzipped fasta) but that should go by filename
- create dictionary on the fly
- create index on the fly

I'm not saying we need to add these now, but it would be good to think about how one would add them in the future without breaking the API. 

A factory does allow for that. 
As do non-factory setters. 

@lbergelson @tfenne @jacarey ?

magicDGS:
A nice implementation of what @yfarjoun mentioned: https://github.com/broadinstitute/gatk/blob/master/src/main/java/org/broadinstitute/hellbender/utils/reference/FastaReferenceWriter.java

Maybe worth to port a that from GATK4?

lbergelson:
I think we should port and update the GATK version instead since it already handles a lot of the things that we want, like creating the index and dictionary. While we're at it we should make it support bgz fasta and create the GZI index for it.

yfarjoun:
@alecw, sorry for the time-lag. We would like to eventually use the GATK-ported version which will be a api change. we could merge this version as a "BETA" version and then change the API, or wait till someone ports the GATK version. Do you have a preference?

yfarjoun:
replaced by #1172

PR #1060, opened 2017-12-20
Update AbstractFastaSequenceFile.java

jamesnemesh:
Add the same verbose message to the UnsupportedOperationException when you fetch an entire contig's sequence as you receive when you fetch a portion of a contig.

### Description

When fetching an entire contig of an unindexed fasta, the operation fails with an UnsupportedOperationException that has no message.  This is difficult for users to determine what the problem is and how to fix it.  Since the exception is a RuntimeException, it's not easy to catch and warn the user.


lbergelson:
@jamesnemesh Thanks for this contribution!

PR #1061, opened 2017-12-22
SortingCollection: freeUpMemory util method

nkarulin:
In parallel implementation of `MarkDuplicates` (https://github.com/broadinstitute/picard/pull/1036)  there is need to free up RAM held by `SortingCollection` with saving the ability to add records later.

There is such a functionality in `SortingCollection`: `spillToDisk()` method. But the method is encapsulated as private. To keep this encapsulation (we sure there could be no usage of the method out of the package) yet to provide public interface to this fully functional behavior, we introduce a new util class in the same package.

### DIFF
--- a/src/main/java/htsjdk/samtools/util/SortingCollection.java
+++ b/src/main/java/htsjdk/samtools/util/SortingCollection.java
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2009 The Broad Institute
+ * Copyright (c) 2018 The Broad Institute
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


lbergelson:
@nkarulin Sorry for the long wait on this.  I appreciate the desire to keep things encapsulated, I'm not sure there's really a benefit of doing it this way though, since it's essentially just exposing it with a different name.  I think it would be better to just make the method public if necessary rather than adding the round-about access to it.  Thanks for adding a test.  If you drop the new class and just expose the method than this should be good to go I think.

nkarulin:
@lbergelson, thanks for the review!
Don't mind to make it simple :+1: 

(What is the reason for the random coverage change?)

lbergelson:
The tests are somewhat non-deterministic, which results in some files randomly fluctuating in coverage.  It's really annoying and we should fix the coverage reporting so that it doesn't always fail because of it, but it's been super low priority so I haven't gotten to it.

lbergelson:
thanks @nkarulin

Issue #1042, opened 2017-11-29
TabixFeatureReader.readHeader uses a different mechanism to open featureFile path URL than other methods.


delocalizer:
### Use case
I'd like to use `TabixFeatureReader` on some paths that have s3:// protocol (AWS object storage). 
Almost everywhere in `TabixFeatureReader`, both index and featureFile paths eventually get opened by `SeekableStreamFactory.getStreamFor()`, which is good because I can implement my own `ISeekableStreamFactory` that knows how to deal with s3:// protocol, and call `SeekableStreamFactory.setInstance(new S3AwareSeekableStreamFactory())`. 
### Issue
Like I said, _almost_ everywhere... except in `TabixFeatureReader.readHeader()`, which uses `ParsingUtils.openInputStream()` to open the featureFile path. That method currently has a hard-coded list of protocols (ftp, http, https) that it checks against, before defaulting to SeekableFileStream. 
### Suggestion
1. `readHeader` should just use `SeekableStreamFactory.getStreamFor()` — this would fix my specific issue.
2. from a design perspective, I don't think `ParsingUtils.openInputStream` should check a list of protocols at all; it should just leave the choice of how to handle any given protocol to the `URLHelper` returned by `getURLHelper`. The explicit list of ftp, http, https largely renders redundant the ability to register a custom `URLHelper` using `ParsingUtils.registerHelperClass` because it will never be called when the protocol is not in that list.

I'm happy to put in a PR for your review if you agree that this is an issue, and that my suggestions aren't too far off the mark.

PR #1064, opened 2018-01-08
fix for #1042

delocalizer:
### Description

In `readHeader()` and `iterator()` methods, use `SeekableStreamFactory` instance to read the feature file instead of `ParsingUtils`. This brings them into line with how it's accessed via `TabixReader` everywhere else in this class. The reason it's not just foolish consistency is explained in #1042; i.e. if you implement your own `ISeekableStreamFactory` to handle some special protocol then that will be used everywhere that the URL is opened in `TabixFeatureReader`, not just in some places.

### Checklist

- [x] Code compiles correctly
- [x] No new tests required (no new functionality)
- [x] All tests passing
- [x] Documentation update not required
- [x] IS backward compatible

### DIFF
--- a/src/main/java/htsjdk/tribble/TabixFeatureReader.java
+++ b/src/main/java/htsjdk/tribble/TabixFeatureReader.java
@@ -23,10 +23,10 @@
  */
 package htsjdk.tribble;
 
+import htsjdk.samtools.seekablestream.SeekableStreamFactory;
 import htsjdk.samtools.util.BlockCompressedInputStream;
 import htsjdk.samtools.util.RuntimeIOException;
 import htsjdk.tribble.readers.*;
-import htsjdk.tribble.util.ParsingUtils;
 
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


lbergelson:
@delocalizer Looks good to me.  Thanks for contributing this, and for the detailed explanation.  I didn't realize that anyone was providing their own factories like you're describing.

lbergelson:
It seems like this fix will only fix the specific case you have though.  Do you need a more general fix that replaces all use of ParsingUtils.openInputStream?  Or a refactoring of that method to use the default SeekableStreamFactory?

lbergelson:
@delocalizer Thank you for your contribution!  Going forward we're considering some changes to the way that new data sources are registered.  

We currently support adding new types of data source using java.nio.file.spi.FileSystemProviders.  The mechanism you're using is one that I didn't even realize existed, it's a legacy from the olden days.  The support for Paths is pretty good.  We've tested with the GCS version extensively in GATK, so anywhere that accepts a Path should work.  A similar plugin exists for S3 which might be worth giving a shot.  (Actually, I noticed you have a [fork](https://github.com/nextflow-io/nextflow-s3fs) of it.. have you tried using it with htsjdk? )

delocalizer:
Hi @lbergelson;
Thanks for the detailed feedback. I did see the work on Paths in htsjdk and even tried out the `nextflow` S3 library you mentioned in that context but [their implementation of SeekableByteChannel](https://github.com/nextflow-io/nextflow-s3fs/blob/a315d27a371a167b408d3c7c8bfbaccbb0d31024/src/main/java/com/upplication/s3fs/S3FileSystemProvider.java#L362) simply downloads the entire remote object to a local temp file. For my purposes (random seek of small sections of a large remote file) this is super-slow and wasteful. I could write my own smarter S3FileSystemProvider I guess, but I did something much easier: use the HTTPS endpoint of S3 buckets so that s3:// protocol URLs are opened just like HTTPS URLs, and random seek is accomplished under the hood with a byte-range request.

delocalizer:
FWIW I think handling all datasources in htsjdk the same way would be a great idea.

magicDGS:
@lbergelson - I am working, as @delocalizer point out, on a small library for supporting HTTP/S protocols for NIO.2

It is a project that I started with the aim of using it in combination with HTSJDK and GATK4, so feel free to contribute!

PR #1066, opened 2018-01-10
made a protected method public

davidbenjamin:
See broadinstitute/gatk#2128.

@cmnbroad I was going through code in `GATKVariantContextUtils` and saw this issue.  Is this the htsjdk change you had in mind?

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
@@ -468,7 +468,7 @@ public class VariantContextWriterBuilder {
      * attempt to resolve any symlinks and try again.  If that fails, and the output file exists
      * but is neither a file or directory then VCF_STREAM is returned.
      */
-    protected static OutputType determineOutputTypeFromFile(final File f) {
+    public static OutputType determineOutputTypeFromFile(final File f) {
         if (isBCF(f)) {
             return OutputType.BCF;
         } else if (isCompressedVCF(f)) {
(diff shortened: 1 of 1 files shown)


cmnbroad:
@davidbenjamin Yes, this is the code that I had in mind when I made the ticket. @lberegelson - What do you think about this making this public ? Should we wait for a path-aware version ?

davidbenjamin:
@lbergelson

lbergelson:
@davidbenjamin Would you mind adding some tests to this thing before we make it public?  The only tricky one is the pipe one,  If you want to make a version with paths that would be even better but it might be kind of a pain.

src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
```diff
@@ -468,7 +468,7 @@ else if (STREAM_TYPES.contains(this.outType))
      * attempt to resolve any symlinks and try again.  If that fails, and the output file exists
      * but is neither a file or directory then VCF_STREAM is returned.
      */
-    protected static OutputType determineOutputTypeFromFile(final File f) {
+    public static OutputType determineOutputTypeFromFile(final File f) {
```
droazen: Can you add a test + javadoc for this newly-public method?

cmnbroad: Looks like there is a test in VariantContextWriterBuilderUnitTest.

davidbenjamin:
@droazen I added javadoc.  @lbergelson as @cmnbroad mentioned there are already tests, but they don't include one for a named pipe, which I'm at a loss for.  I tried something like
```
Assert.assertEquals(OutputType.VCF, VariantContextWriterBuilder.determineOutputTypeFromFile(new File(new FileOutputStream(vcf).___getAbsolutePath()___)));
```
but that `___getAbsolutePath()___` method doesn't exist.

lbergelson:
@davidbenjamin I think we can merge this without a test for a pipe.  Those are tricky to write and until we set up infrastructure to make it easy I don't think we can require everyone to test on one.

PR #1069, opened 2018-01-15
MergeSamFiles accept SO:UNKNOWN

a-filipp0v:
### Description

Request is associated with Picard's issue https://github.com/broadinstitute/picard/issues/796

As said in that issue MergeSamFiles can't accept SAM-file with SO:UNKNOWN. That behavior is expected, because tag's value is in improper case(in accordance with SAM-format specification, SO value must be in lowercase). We've found that wrong tag case also provides an exception.

The idea of implementation is that htsjdk must check case of header's tags and values. If tag/value itself is valid, but in improper case(e.g. lowercase instead of uppercase) - it should be converted to proper case before sending the result to output(without changing the source file). Also, there must be a message if that conversion was made.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMFileHeader.java
+++ b/src/main/java/htsjdk/samtools/SAMFileHeader.java
@@ -259,20 +259,20 @@ public class SAMFileHeader extends AbstractSAMHeaderRecord
             if (so == null) {
                 sortOrder = SortOrder.unsorted;
             } else {
-                try {
-                    String modified = so;
-                    if (!SortOrder.valueOf(modified).toString().matches("[a-z]+")) {
-                        log.warn("Warning! Your sort order value has improper case("
-                                         + SortOrder.valueOf(so) + "), "
-                                         + "reformatted input data to lowercase. "
-                                         + "NOTE: Your input file WASN'T modified.");
-                        modified = modified.toLowerCase();
-                    }
-                    return SortOrder.valueOf(modified);
-                } catch (IllegalArgumentException e) {
+//                try {
+//                    return SortOrder.valueOf(so);
+//                } catch (IllegalArgumentException e) {
+//                    log.warn("Found non conforming header SO tag: " + so + ". Treating as 'unknown'.");
+//                    sortOrder = SortOrder.unknown;
+//                }
+//**********************
+                String modified = so;
+                if(Arrays.stream(SortOrder.values())
+                         .noneMatch((t)->t.name().equals(so))) {
+                    modified = "unknown";
                     log.warn("Found non conforming header SO tag: " + so + ". Treating as 'unknown'.");
-                    sortOrder = SortOrder.unknown;
                 }
+                return SortOrder.valueOf(modified);
             }
         }
         return sortOrder;
--- a/src/main/java/htsjdk/samtools/SAMFileHeader.java
+++ b/src/main/java/htsjdk/samtools/SAMFileHeader.java
@@ -260,12 +260,15 @@ public class SAMFileHeader extends AbstractSAMHeaderRecord
                 sortOrder = SortOrder.unsorted;
             } else {
                 try {
-                    if (!SortOrder.valueOf(so).toString().matches("[a-z]+")) {
-                        System.out.printf("Warning! Your sort order value has improper case(%s), "
-                                                  + "reformatted input data to lowercase. "
-                                                  + "NOTE: Your input file WASN'T modified.\n", SortOrder.valueOf(so));
-                        return SortOrder.valueOf(so.toLowerCase());
-                    } else return SortOrder.valueOf(so);
+                    String modified = so;
+                    if (!SortOrder.valueOf(modified).toString().matches("[a-z]+")) {
+                        log.warn("Warning! Your sort order value has improper case("
+                                         + SortOrder.valueOf(so) + "), "
+                                         + "reformatted input data to lowercase. "
+                                         + "NOTE: Your input file WASN'T modified.");
+                        modified = modified.toLowerCase();
+                    }
+                    return SortOrder.valueOf(modified);
                 } catch (IllegalArgumentException e) {
                     log.warn("Found non conforming header SO tag: " + so + ". Treating as 'unknown'.");
                     sortOrder = SortOrder.unknown;
--- a/src/main/java/htsjdk/samtools/SAMFileHeader.java
+++ b/src/main/java/htsjdk/samtools/SAMFileHeader.java
@@ -259,20 +259,12 @@ public class SAMFileHeader extends AbstractSAMHeaderRecord
             if (so == null) {
                 sortOrder = SortOrder.unsorted;
             } else {
-//                try {
-//                    return SortOrder.valueOf(so);
-//                } catch (IllegalArgumentException e) {
-//                    log.warn("Found non conforming header SO tag: " + so + ". Treating as 'unknown'.");
-//                    sortOrder = SortOrder.unknown;
-//                }
-//**********************
-                String modified = so;
-                if(Arrays.stream(SortOrder.values())
-                         .noneMatch((t)->t.name().equals(so))) {
-                    modified = "unknown";
-                    log.warn("Found non conforming header SO tag: " + so + ". Treating as 'unknown'.");
+                try {
+                    return SortOrder.valueOf(so);
+                } catch (IllegalArgumentException e) {
+                    log.warn("Found non-conforming header SO tag: " + so + ". Treating as 'unknown'.");
+                    sortOrder = SortOrder.unknown;
                 }
-                return SortOrder.valueOf(modified);
             }
         }
         return sortOrder;
@@ -332,12 +324,18 @@ public class SAMFileHeader extends AbstractSAMHeaderRecord
      */
     @Override
     public void setAttribute(final String key, final String value) {
+        String tempVal = value;
         if (key.equals(SORT_ORDER_TAG)) {
-            this.sortOrder = null;
+            try {
+                this.sortOrder = null;
+                tempVal = SortOrder.valueOf(value).toString();
+            } catch (IllegalArgumentException e) {
+                tempVal = "unknown";
+            }
         } else if (key.equals(GROUP_ORDER_TAG)) {
             this.groupOrder = null;
         }
-        super.setAttribute(key, value);
+        super.setAttribute(key, tempVal);
     }
 
     /**
--- a/src/main/java/htsjdk/samtools/SAMFileHeader.java
+++ b/src/main/java/htsjdk/samtools/SAMFileHeader.java
@@ -326,11 +326,11 @@ public class SAMFileHeader extends AbstractSAMHeaderRecord
     public void setAttribute(final String key, final String value) {
         String tempVal = value;
         if (key.equals(SORT_ORDER_TAG)) {
+            this.sortOrder = null;
             try {
-                this.sortOrder = null;
                 tempVal = SortOrder.valueOf(value).toString();
             } catch (IllegalArgumentException e) {
-                tempVal = "unknown";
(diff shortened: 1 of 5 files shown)


yfarjoun:
Personally I'm opposed to this change. 

I do not see why htsjdk needs to be made more complicated in order to accept non-SAM-spec compliant input. Its hard enough as it is even with SAM-compliant input and this will make the code more complicated. What if someone has a typo in the header and they spelled the sort order "unkwnon"? Or perhaps they use a different "locale" and the used "SO: desconocido" ? should htsjdk also accept that? 

That off my chest, I'm not a maintainer...so I'd love to hear from them before even looking at the details of the implementation: @lbergelson @tfenne  @jacarey

yfarjoun:
I think that the correct solution is to fix the offending software so that it produces compliant BAM files and to write a simple script that fixes "broken" headers.

yfarjoun:
Alternatively a better  (in my view) solution would be an ability to ignore (when validation=silent or leniant) an unknown sort order and use "unknown" when that happens (instead of lowercasing)

magicDGS:
I also prefer that HTSJDK is SAM-compliant and do not allow non-supported `SO`...

jacarey:
I agree with @yfarjoun

yfarjoun:
@a-filipp0v could you modify the code as follows:
1. remove the lower-casing and the logs related to that
2. if SortOrder.valueOf(unmodified) throws an error, AND VALIDATION_STRINGENCY!=STRICT, use SortOrder.UNKNOWN and either emit a warning or not depending on VALIDATION_STRINGENCY.

I think that this will both solve the issue at hand and not introduce un-needed complexity.

As a bonus, I think that your tests should still be useful!

a-filipp0v:
Thank you for your comments. Hope we've understood you correctly in case of non-conforming SO tag values.
Our implementation has been reworked according to your requests. Also, we've reworked previous tests in SamFileHeaderTest, added few more and removed one test case from MergeSamFilesTest because it's redundant after our implementation.

src/main/java/htsjdk/samtools/SAMFileHeader.java
```diff
@@ -324,12 +324,18 @@ public void setAttribute(final String key, final Object value) {
...
     @Override
     public void setAttribute(final String key, final String value) {
+        String tempVal = value;
         if (key.equals(SORT_ORDER_TAG)) {
-            this.sortOrder = null;
+            try {
+                this.sortOrder = null;
+                tempVal = SortOrder.valueOf(value).toString();
+            } catch (IllegalArgumentException e) {
+                tempVal = "unknown";
```
yfarjoun: use `SortOrder.unknown.toString()` here.

src/main/java/htsjdk/samtools/SAMTextHeaderCodec.java
```diff
@@ -323,11 +326,30 @@ private void reportErrorParsingLine(String reason, final SAMValidationError.Type
...
                     continue;
                 }
+                validateSortOrderValue(keyAndValue);
                 mKeyValuePairs.put(keyAndValue[0], keyAndValue[1]);
             }
             lineValid = true;
         }
 
+        private void validateSortOrderValue(String[] value) {
+            if ("SO".equals(value[0])) {
```
yfarjoun: use SAMFileHeader.SORT_ORDER_TAG

src/main/java/htsjdk/samtools/SAMTextHeaderCodec.java
```diff
@@ -323,11 +326,30 @@ private void reportErrorParsingLine(String reason, final SAMValidationError.Type
...
+                } catch (IllegalArgumentException e) {
+                    if (validationStringency == ValidationStringency.STRICT) {
+                        throw new SAMFormatException("Found non conforming header SO tag: "
+                                                     + value[1]
+                                                     + ", exiting because VALIDATION_STRINGENCY=STRICT");
+                    } else if (validationStringency == ValidationStringency.LENIENT) {
+                        log.warn("Found non conforming header SO tag: "
+                                 + value[1] + ". Treating as 'unknown'.");
+                    }
+                    value[1] = "unknown";
```
yfarjoun: use SortOrder.unknown.toString()

src/main/java/htsjdk/samtools/SAMFileHeader.java
```diff
@@ -324,12 +324,18 @@ public void setAttribute(final String key, final Object value) {
      */
     @Override
     public void setAttribute(final String key, final String value) {
+        String tempVal = value;
         if (key.equals(SORT_ORDER_TAG)) {
-            this.sortOrder = null;
+            try {
+                this.sortOrder = null;
```
yfarjoun: Why move this line into the try clause?

src/main/java/htsjdk/samtools/SAMTextHeaderCodec.java
```diff
@@ -69,6 +70,8 @@
     private static final Pattern FIELD_SEPARATOR_RE = Pattern.compile(FIELD_SEPARATOR);
 
     public static final String COMMENT_PREFIX = HEADER_LINE_START + HeaderRecordType.CO.name() + FIELD_SEPARATOR;
+    private static final Log log = Log.getInstance(SAMTextHeaderCodec.class);
+
```
yfarjoun: remove extra newline

src/main/java/htsjdk/samtools/SAMTextHeaderCodec.java
```diff
@@ -233,8 +236,8 @@ private void parseHDLine(final ParsedHeaderLine parsedHeaderLine) {
         try {
             if (soString != null) SAMFileHeader.SortOrder.valueOf(soString);
         } catch (IllegalArgumentException e) {
-            reportErrorParsingLine(HEADER_LINE_START + parsedHeaderLine.getHeaderRecordType() +
-                            " line has non-conforming SO tag value: "+ soString + ".",
+                reportErrorParsingLine(HEADER_LINE_START + parsedHeaderLine.getHeaderRecordType() +
```
yfarjoun: revert extra tab

src/main/java/htsjdk/samtools/SAMTextHeaderCodec.java
```diff
@@ -323,11 +326,30 @@ private void reportErrorParsingLine(String reason, final SAMValidationError.Type
                             SAMValidationError.Type.HEADER_TAG_MULTIPLY_DEFINED, null);
                     continue;
                 }
+                validateSortOrderValue(keyAndValue);
```
yfarjoun: why is this needed here? could you put it instead in SamFileValidator? perhaps in the ValidateHeader method?

note that the validation methods create a tally of validation failures... and this should probably do the same.

src/test/java/htsjdk/samtools/ValidateSamFileTest.java
```diff
@@ -552,12 +552,6 @@ public void duplicateReadsOutOfOrder() throws Exception {
                         "@RG\tID:0\tSM:Hi,Mom!\n" +
                         "E\t147\tchr1\t15\t255\t10M\t=\t2\t-30\tCAACAGAAGC\t)'.*.+2,))\tU2:Z:CAA";
 
-        final String SOTagCorrectlyProcessTestData =
```
yfarjoun: why is this removed?

a-filipp0v:
Hello, @yfarjoun , thanks for your reply.

According to your previous comments, we've tried to move an appropriate check to SamFileValidator, but found that it still might make sense to leave our checks in SAMTextHeaderCodec and SAMFileHeader.

First, the check in SamFileValidator will be called only when the ValidateSamFile is used, which does not prevent an error when using any tools separately (for example, MergeSamFile as in this issue).

Secondly, SamFileValidator does not contain separate data about Validation Stringency, because this parameter is parsed from Picard during start.

We suggest leaving checks in SAMTextHeaderCodec and SAMFileHeader AND adding a similar one in SamFileValidator. What do you think about it?

lbergelson:
@yfarjoun Since you did the initial review, could you take a look at this again when you get a chance?

yfarjoun:
@a-filipp0v thanks for the explanation. I've removed my comment relating to the validation.

yfarjoun:
@a-filipp0v sorry for the delay. please respond to the review comments.

yfarjoun:
@a-filipp0v just so we know, are you planning to get back to this PR?

src/test/java/htsjdk/samtools/ValidateSamFileTest.java
IanaKabakova: thank you for your reply. We've removed this test case since as @a-filipp0v said, it's redundant after our implementation.

IanaKabakova:
@yfarjoun sorry for the delay and thank you for your reply. I made some fixes according to your notes. Could you take a look at PR again?

src/test/java/htsjdk/samtools/ValidateSamFileTest.java
yfarjoun: not sure I understand...if it's just "Redundant" but not exactly equals, I'd ask that it be left in.

yfarjoun:
Thanks for the changes. Please leave tests as is, unless they no longer pass.

src/test/java/htsjdk/samtools/ValidateSamFileTest.java
IanaKabakova: This case does not pass now, since we no longer treat SO:NOTKNOWN as error as test expects.

yfarjoun: got it

yfarjoun:
@jacarey @tfenne @lbergelson any opinions about this small API change?

src/main/java/htsjdk/samtools/SAMFileHeader.java
```diff
@@ -324,12 +324,18 @@ public void setAttribute(final String key, final Object value) {
      */
     @Override
     public void setAttribute(final String key, final String value) {
+        String tempVal = value;
         if (key.equals(SORT_ORDER_TAG)) {
             this.sortOrder = null;
+            try {
+                tempVal = SortOrder.valueOf(value).toString();
```
pshapiro4broad: In `setSortOrder()` the code uses `.name()` instead of `.toString()` which I think is clearer; either way the code should be consistent.

Also, it seems odd that `setSortOrder()` sets `sortOrder` to non-null, if the argument is a valid `SortOrder`, but this method sets it to `null`. Maybe this method should just call `setSortOrder()` when `key` is `SORT_ORDER_TAG`?

src/main/java/htsjdk/samtools/SAMTextHeaderCodec.java
```diff
@@ -323,11 +326,30 @@ private void reportErrorParsingLine(String reason, final SAMValidationError.Type
...
                 mKeyValuePairs.put(keyAndValue[0], keyAndValue[1]);
             }
             lineValid = true;
         }
 
+        private void validateSortOrderValue(String[] value) {
```
pshapiro4broad: I think it would be slightly clearer if the caller extracted `value[0]` and `value[1]` so they could be declared as `String key` and `String value` here.

src/main/java/htsjdk/samtools/SAMFileHeader.java
yfarjoun: @pshapiro4broad that's how the old implementation also does it...it let's the super-class deal with it...take a look at htsjdk.samtools.SAMFileHeader#getSortOrder, I think that might clarify things up (and note that this.sortOrder is private)

yfarjoun:
No response from the maintainers for a while now, I suspect that no-one objects....if I'm wrong, please revert and accept my apologies.

PR #1072, opened 2018-01-19
Fix #1049

### LINKED ISSUE #1049: FastaSequenceIndex not usable outside its package
The class FastaSequenceIndex includes a _FastaSequenceIndexEntry getIndexEntry(contigName)_ method in its API, but that class is package protected. All the useful methods in FastaSequenceIndex are based in FastaSequenceIndexEntry, thus, if you are outside that package (I'm using HTSJDK as a library), the FastaSequenceIndex class does not have any real use.

FastaSequenceIndexEntry is inside FastaSequenceIndex file. Java only allows a public class per file, is that the reason that FastaSequenceIndexEntry is package protected? Maybe it was an inner class in the past... Would you consider making this class public (moving it outside FastaSequenceIndex.java) so it can be used by external modules?

Initially, I was trying to read the sequence dictionary for a fasta file, but I found that the SAMSequenceDictionary is empty if I create the fasta file like this:  `ReferenceSequenceFileFactory.getReferenceSequenceFile(fastaFile, true);`
I then tried to use the Fasta index, but I found that dead end with the package protected class. Is there any other turnaround to do this?

Thanks and Regards

danking:
### Description

Fixes #1049, return value of `FastaSequenceIndexEntry.getIndexEntry` is unusable.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


danxmoran:
Hi @danking, you'll need to move `FastaSequenceIndexEntry` to its own file to get this to compile.

pshapiro4broad:
> Hi @danking, you'll need to move FastaSequenceIndexEntry to its own file to get this to compile.

Given the way this class is used, making it a static inner class might be a better option. If the inner class was called `Entry` then you'd refer to it as `FastaSequenceIndex.Entry`.

danking:
@danxmoran hah sorry, too much time in Scala land.

I'll move it since that seems to be the lowest impact change.

danking:
@danxmoran I think this will compile.

jacarey:
Pinging @lbergelson @tfenne for objections to merge.

lbergelson:
👍

PR #1075, opened 2018-01-23
Fix for incorrect MISMATCH_CIGAR_SEQ_LENGTH check

d-cameron:
### Description

Removed MISMATCH_CIGAR_SEQ_LENGTH validation check when no read sequence is defined.

When bwa reports split read alignments it only includes read sequence for the primary alignment record. Omitting the read sequence at the supplementary alignment reduces the file size without loss of information as the read sequence for the supplementary alignment records can be reconstructed from the primary alignment record and the supplementary alignment CIGAR.

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMRecord.java
+++ b/src/main/java/htsjdk/samtools/SAMRecord.java
@@ -2101,7 +2101,7 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
             if (firstOnly) return ret;
         }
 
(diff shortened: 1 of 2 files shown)


d-cameron:
Has this PR exposed poor coverage somewhere else? It's not immediately obvious why Md5CalculatingOutputStream coverage has changed at all.

danxmoran:
@d-cameron I think the coverage diff is unrelated, I'm seeing the same thing happen in #1072.

Issue #898, opened 2017-06-09
SeekableStream subclasses should implement InputStream.available()


lbergelson:
The default implementation of `InputStream.available()` just returns 0.  SeekableStream implementations should make a best effort implementation of this instead of using the lousy default.

lbergelson:
This is more important than we thought due to a bug in `GZIPInputStream` where it relies on the result of `available` in order to determine EOF.

droazen:
See full description of the interaction of the `GZIPInputStream` bug with htsjdk `SeekableStream` classes here: https://github.com/broadinstitute/gatk/issues/4224#issuecomment-360282461

magicDGS:
Maybe a different way to go is to use [`GzipCompressorInputStream`](https://commons.apache.org/proper/commons-compress/apidocs/org/apache/commons/compress/compressors/gzip/GzipCompressorInputStream.html) from commons-compress instead, which is already a dependency of HTSJDK (similar proposal as #799)

PR #1077, opened 2018-01-25
Implement SeekableStream.available

magicDGS:
### Description

Addresses #898 by using a common `SeekableStream` available implementation. This is necessary to fix problems while reading (b)gzip compressed files with a wrapped `SeekableStream` in a `GZIPInputStream`.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/seekablestream/SeekableStream.java
+++ b/src/main/java/htsjdk/samtools/seekablestream/SeekableStream.java
@@ -90,15 +90,15 @@ public abstract class SeekableStream extends InputStream {
     }
 
     /**
-     * Computes how many bytes are available in the stream.
+     * The return value of this method is unusable for any purpose, and we are only implementing it
+     * because certain Java classes like {@link java.util.zip.GZIPInputStream} incorrectly rely on
+     * it to detect EOF
      *
      * <p>If {@code eof() == true}, 0 bytes are available. Otherwise, available bytes are the
      * difference between the length of the stream ({@link #length()}) and the current position
      * ({@link #position()}.
      *
-     * <p>Note: implementations might override to provide more reliable results.
-     *
-     * @return {@code 0} if the end of the file has been reached or the length cannot be determine;
+     * @return {@code 0} if the end of the file has been reached or the length cannot be determined;
      * number of bytes remaining in the stream otherwise.
      */
     @Override
--- a/src/main/java/htsjdk/samtools/seekablestream/SeekableStream.java
+++ b/src/main/java/htsjdk/samtools/seekablestream/SeekableStream.java
@@ -90,6 +90,28 @@ public abstract class SeekableStream extends InputStream {
     }
 
     /**
+     * Computes how many bytes are available in the stream.
+     *
+     * <p>If {@code eof() == true}, 0 bytes are available. Otherwise, available bytes are the
+     * difference between the length of the stream ({@link #length()}) and the current position
+     * ({@link #position()}.
+     *
+     * <p>Note: implementations might override to provide more reliable results.
+     *
+     * @return {@code 0} if the end of the file has been reached or the length cannot be determine;
+     * number of bytes remaining in the stream otherwise.
+     */
+    @Override
+    public int available() throws IOException {
+        if (eof()) {
+            return 0;
+        }
+        final long remaining = length() - position();
+        // the remaining might be negative if the length is not available (0)
+        return (remaining < 0) ? 0 : (int) remaining;
+    }
+
+    /**
      * Mark the current position of the stream.
      *
      * <p>Note: there is no limit for reading.
--- a/src/test/java/htsjdk/samtools/seekablestream/SeekableStreamGZIPinputStreamIntegrationTest.java
+++ b/src/test/java/htsjdk/samtools/seekablestream/SeekableStreamGZIPinputStreamIntegrationTest.java
@@ -1,0 +1,120 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2018 Daniel Gomez-Sanchez
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package htsjdk.samtools.seekablestream;
+
+import htsjdk.HtsjdkTest;
+import htsjdk.samtools.util.BufferedLineReader;
+import htsjdk.samtools.util.IOUtil;
+import htsjdk.samtools.util.LineReader;
+import htsjdk.samtools.util.TestUtil;
+import htsjdk.variant.variantcontext.Allele;
+import htsjdk.variant.variantcontext.VariantContext;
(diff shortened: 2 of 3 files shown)


magicDGS:
Please @droazen or @cmnbroad, can you review this one? I feel that it is quite important...

src/main/java/htsjdk/samtools/seekablestream/SeekableStream.java
```diff
@@ -89,6 +89,28 @@ public void readFully(byte b[]) throws IOException {
         }
     }
 
+    /**
+     * Computes how many bytes are available in the stream.
```
pshapiro4broad: Javadoc for an overridden method isn't necessary as documentation is inherited. It can even make things worse, since any updates to the base class documentation wouldn't be reflected when viewing the javadoc for the overridden method.

droazen:
@jamesemery please review

src/main/java/htsjdk/samtools/seekablestream/SeekableStream.java
magicDGS: So do you suggest to add a `{@inheritDoc}` at the top? I just added the javadoc to show the concrete implementation...

pshapiro4broad: No javadoc is necessary, the base class doc will be used if no javadoc comment is provided. `@inheritdoc` is useful if you want to quote the entire base class doc and add more docs to it. In the case where you just want the base class doc, it's better to omit the javadoc completely.

magicDGS: This is a different javadoc, that's why I am asking...

pshapiro4broad: Sure, but why is it different? Does it change the API as far as the caller is concerned?

I think the right way to think about javadoc is that it's documentation for the client of the API (the caller). Implementation details should by nature be not a client concern, and don't need to go in a javadoc. They would be more suited for an inline comment, e.g.
```
public int available() throws IOException {
    // If we are at the eof, 0 bytes are available.
    if (eof()) {
        return 0;
    }
    // Otherwise, available bytes are the difference between the length of the
    // stream and the current position.
    final long remaining = length() - position();
    // remaining might be negative if the length is not available, so return 0.
    return Math.max(0, (int) remaining);
}
```

magicDGS: I understand your point here, but as an interface/abstract class it should point that implementations should be aware of this behavior... otherwise, it can have unexpected failures (I guess).

Anyway, I can just remove it if it's necessary for accepting this (there's not `@implNote` considered in HTSJDK yet, that would be ideal for this case)

pshapiro4broad: Looking at this class more closely I see that there are already many comments that replace the base class documentation, so I guess there's no harm in continuing the practice here.

droazen: The "API" for `available()` is a complete disaster: "an estimate of the number of bytes that can be read without blocking", "returns 0 at end of stream, but also possibly at other times as well".

We should just override the javadoc to say "the return value of this method is unusable for any purpose, and we are only implementing it because certain Java classes like GZIPInputStream incorrectly rely on it to detect EOF"  :)

src/main/java/htsjdk/samtools/seekablestream/SeekableStream.java
```diff
@@ -89,6 +89,28 @@ public void readFully(byte b[]) throws IOException {
...
+     *
+     * <p>If {@code eof() == true}, 0 bytes are available. Otherwise, available bytes are the
+     * difference between the length of the stream ({@link #length()}) and the current position
+     * ({@link #position()}.
+     *
+     * <p>Note: implementations might override to provide more reliable results.
+     *
+     * @return {@code 0} if the end of the file has been reached or the length cannot be determine;
```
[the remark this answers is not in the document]
magicDGS: Modified.

magicDGS:
@jamesemery - I agree that there is a requirement of a test for the `GZIPinputStream`, but I am not sure how to reproduce the issue. As you told me in the PR for disabling BGZIP input without index in the GATK, it looks complicated to reproduce the issue.

If you have any concrete implementation suggestion for the test, I can implement it. I investigate a bit the code of `GZIPinputStream` and looks complicated to exercise the code that fails.

lbergelson:
A good test would be one that had a file that previously failed using the gzip reader and doesn't with the change.  My theory is that it fails when a read randomly starts very close to a block boundary, specifically one that starts within 26 bytes from the edge of the block.  One way we might be able to reproduce this condition in a small file is to craft a file that very small bgcf blocks so boundaries are more common.

yfarjoun:
one could generate a series of vcf files, where the first record is of
variable size (say with a format field that has a string of random
characters and undetermined length in it) and lots of small records
afterwards. A data-provider will generate many vcfs with ever growing first
record and if @lbergelson is right, some of them will fail to be read back
properly.

magicDGS:
@lbergelson and @yfarjoun - I generated a test as you said, and it was failing without the implementation of `available()` for all test cases implemented.

Back to you!

src/test/java/htsjdk/samtools/seekablestream/SeekableStreamGZIPinputStreamIntegrationTest.java
```diff
@@ -0,0 +1,120 @@
...
+        final int nSmallRecords = 1000000;
+        for (int firstRecordLength = 1000; firstRecordLength <= 10000; firstRecordLength+=1000) {
+            data.add(new Object[]{createBgzipVcfsWithVariableSize(firstRecordLength, nSmallRecords), nSmallRecords+1});
+        }
+        return data.iterator();
+    }
+
+    @Test(dataProvider = "compressedVcfsToTest")
+    public void testWrappedSeekableStreamInGZIPinputStream(final File input, final long nLines) throws Exception {
+        try (final LineReader reader = new BufferedLineReader(new GZIPInputStream(new SeekableFileStream(input)))) {
```
jamesemery: This test is very good, it appears to definitely fail when run without seekableStream.available(). Maybe you should assert that the file was actually written properly though, perhaps by asserting the lenght is consistent when you open a blockCompressedInputStream and compare the two?

src/test/java/htsjdk/samtools/seekablestream/SeekableStreamGZIPinputStreamIntegrationTest.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Random;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * Tests for the integration between SeekableStream and GZIPInputStream.
```
jamesemery: You need to add a more explicit explanation of what this test is asserting to the description, as well as a link to the htsjdk issue related to this. As of now this test is a little confusing and it might not be entirely clear what it is testing if you don't already know.

src/test/java/htsjdk/samtools/seekablestream/SeekableStreamGZIPinputStreamIntegrationTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+
+    private static File createBgzipVcfsWithVariableSize(final int firstRecordAttributeLength, final int nSmallRecords) throws Exception {
+        final VariantContext longRecord = new VariantContextBuilder("long", TEST_CHR, 1, 1, alleles)
+                .attribute(RANDOM_ATTRIBUTE, generateRandomString(firstRecordAttributeLength))
+                .make();
+        final File tempFile = Files.createTempFile("test" + firstRecordAttributeLength + "_" + nSmallRecords, ".vcf.gz").toFile();
+        try (final VariantContextWriter writer = new VariantContextWriterBuilder()
+                .setOptions(VariantContextWriterBuilder.NO_OPTIONS)
+                .setOutputFile(tempFile).build()) {
```
jamesemery: add `.setOutputFileType(VariantContextWriterBuilder.OutputType.BLOCK_COMPRESSED_VCF)` to this builder so we can be sure the test is writing a block compressed file even if the extension logic later changes.

src/test/java/htsjdk/samtools/seekablestream/SeekableStreamGZIPinputStreamIntegrationTest.java
magicDGS: Added a more detailed javadoc.

src/test/java/htsjdk/samtools/seekablestream/SeekableStreamGZIPinputStreamIntegrationTest.java
magicDGS: Done.

src/test/java/htsjdk/samtools/seekablestream/SeekableStreamGZIPinputStreamIntegrationTest.java
magicDGS: I think that it such a test is over-killing, because we already know how many lines should be. Anyway, I added a different test for check the consistency with a `BlockCompressedInputStream`, checking every byte returned by the stream.

magicDGS:
@jamesemery - back to you!

magicDGS:
Thanks @lbergelson!

lbergelson:
@magicDGS Thanks for fixing this, and for the nice test!

Issue #1080, opened 2018-01-31
ENA Reference Service tests failing


lbergelson:
The ENA reference service tests are failing and causing all pr's to fail.  It may be a transient (but it's been happening for hours) failure, or we may need to make changes somewhere.

```
EnaRefServiceTest:
- testEnaRefServiceData(57151e6196306db5d9f33133572a5482) *** FAILED ***
  [secure].samtools.cram.ref.EnaRefService$GaveUpException:
  at [secure].samtools.cram.ref.EnaRefService.getSequence(EnaRefService.java:112)
  at [secure].samtools.cram.ref.EnaRefService.getSequence(EnaRefService.java:26)
  at [secure].samtools.cram.ref.EnaRefServiceTest.testEnaRefServiceData(EnaRefServiceTest.java:20)
  at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
  at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  at java.lang.reflect.Method.invoke(Method.java:498)
  at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:85)
  at org.testng.internal.Invoker.invokeMethod(Invoker.java:639)
  at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:816)
  ...
- testEnaRefServiceData(0000088cbcebe818eb431d58c908c698) *** FAILED ***
  [secure].samtools.cram.ref.EnaRefService$GaveUpException:
  at [secure].samtools.cram.ref.EnaRefService.getSequence(EnaRefService.java:112)
  at [secure].samtools.cram.ref.EnaRefService.getSequence(EnaRefService.java:26)
  at [secure].samtools.cram.ref.EnaRefServiceTest.testEnaRefServiceData(EnaRefServiceTest.java:20)
  at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
  at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #1081, opened 2018-01-31
skipping ENA tests if they throw GaveUpException

lbergelson:
this will let us work around #1080 until we figure out what's causing the failures on travi


yfarjoun:
is this intended to be the long-term solution, or are you planning to swap it out with a different solution eventually? if the latter, please add a dated comment to that effect.

lbergelson:
this seems to have resolved itself

Issue #1084, opened 2018-02-08
input streams opened from Paths fail when the underlying file is a pipe


lbergelson:

lbergelson:
We should also file a bug with java.

nh13:
@lbergelson I just hit this issue with when upgrading from 2.13.0 to 2.14.3.  Not supporting `/dev/stdin` as a path is a serious regression causing production issues on my end.  I have confirmed it works with 2.13.0 and now fails (unexpectedly) in 2.14.3: https://github.com/fulcrumgenomics/fgbio/issues/404

nh13:
@tfenne

nh13:
I can also confirm that reading from stdin works in 2.14.1 and not in 2.14.2, with the latter including #1077 and on.

lbergelson:
@nh13 Did reading from Stdin as a path work before?  I was under the impression that it had never worked.  

There's a bug in the java implementation of `Files.newInputStream()` which makes it annoying to work around.
The following 
```
    @Test
    public void testStdin() throws IOException {
        final InputStream inputStream = Files.newInputStream(Paths.get("/dev/stdin/"));
        inputStream.available();
    }
```

```
java.io.IOException: Illegal seek

	at sun.nio.ch.FileChannelImpl.position0(Native Method)
	at sun.nio.ch.FileChannelImpl.position(FileChannelImpl.java:264)
	at sun.nio.ch.ChannelInputStream.available(ChannelInputStream.java:116)
	at htsjdk.samtools.SamReaderFactoryTest.testStdin(SamReaderFactoryTest.java:539)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:85)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:639)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:816)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1124)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:125)
[...]

PR #1085, opened 2018-02-08
re-enable reading bam files from pipes

### LINKED ISSUE #1083: opening a SamReader on a named pipe is broken in 2.14.2
#1077 caused a regression that prevents opening names pipes with SamReaderFactory.

The issue is the `eof()` makes a call to `RandomAccessFile.length()` which causes a failure in native code if the input file isn't actually a randomly accessible file...  For some reason we're opening named pipes as if they were seekable, which now explodes on the first read because we are calling `length()` as part of available.

The fix should be to not open named pipes as seekable inputs.

lbergelson:
fixing an issue that prevented reading a bam file from an unseekable file, ex: a unix pipe
fixes #1083 which was introduced by #1077

this only fixes the case where the bam is being opened as file, a similar issue exists for Paths, #1084
support for reading pipes as paths has never worked
### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


src/main/java/htsjdk/samtools/SamInputResource.java
```diff
@@ -335,7 +347,7 @@ public URL asUrl() {
 
     @Override
     public SeekableStream asUnbufferedSeekableStream() {
-        return lazySeekableStream.get();
+            return lazySeekableStream.get();
```
jacarey: extra whitespace

src/main/java/htsjdk/samtools/SamInputResource.java
```diff
@@ -269,12 +266,27 @@ public URL asUrl() {
...
+        } else {
+            return null;
+        }
     }
 
     @Override
     public InputStream asUnbufferedInputStream() {
-        return asUnbufferedSeekableStream();
+        final SeekableStream seekableStream = asUnbufferedSeekableStream();
+        if( seekableStream != null) {
```
jacarey: space after if and not before seekableStream

src/main/java/htsjdk/samtools/SamInputResource.java
```diff
@@ -269,12 +266,27 @@ public URL asUrl() {
 
     @Override
     public SeekableStream asUnbufferedSeekableStream() {
-        return lazySeekableStream.get();
+        //if the file doesn't exist, the try to open the stream anyway because users might be expecting the exception
+        //if it not a regular file than we won't be able to seek on it, so return null
+        if(!fileResource.exists() || fileResource.isFile()) {
```
jacarey: space after if

PR #1086, opened 2018-02-11
Implementing > 64K cigar operators for BAM

yfarjoun:
### Description
This is based on @lh3 's PR #1003  but implements it using htsjdk constructs rather than the low-level implementations that were made there.

This PR implements the long-cigar solution that was agreed to in hts-spec. In the presence of CIGAR with >65535 operators, we move the real CIGAR to the CG tag and put a fake full clipping CIGAR (`<readLength>S<referenceLength>N`) in its place.

Also, many tests and checks were implemented.
### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


yfarjoun:
@lh3 can you check my work?

@lbergelson you want to take a look here?

src/main/java/htsjdk/samtools/BAMRecord.java
```diff
@@ -256,6 +283,55 @@ public Cigar getCigar() {
         return super.getCigar();
     }
 
+    private static boolean isSentinelCigar(final Cigar cigar, final int readLength) {
```
yfarjoun: are SEQ allowed to be '*' when the cigar is populated? if so, this will not work in that case as it looks at the read-length for guidance. If folks think that this is too restrictive, I can just look for S and N operators and remove the readlength check.

lh3: Perhaps change the condition to:
```java
return cigar.getCigarElement(0).getOperator() == CigarOperator.S &&
       (cigar.getCigarElement(0).getLength() == readLength || readLength == 0) &&
       cigar.getCigarElement(1).getOperator() == CigarOperator.N;
```

yfarjoun: ah. I ignored this caveat. I thought that the sentinel is always xSyN

src/main/java/htsjdk/samtools/BAMRecordCodec.java
```diff
@@ -174,6 +186,18 @@ public void encode(final SAMRecord alignment) {
...
+     * Create a "Sentinel" cigar that will be placed in BAM file when the actual cigar has more than 0xffff operator,
+     * which are not supported by the bam format. The actual cigar will be encoded and placed in the CG attribute.
+     * @param cigar actual cigar to create sentinel cigar for
+     * @return sentinel cigar xSyN with readLength (x) and referenceLength (y) matching the input cigar.
+     */
+    public static Cigar makeSentinelCigar(final Cigar cigar) {
+        return new Cigar(Arrays.asList(
+                new CigarElement(cigar.getReadLength(), CigarOperator.S),
+                new CigarElement(cigar.getReferenceLength(), CigarOperator.N)));
+    }
```
lh3: When writing to BAM, a CIGAR length is at most `(1<<28)-1`. Probably this has been taken care of somewhere else in the code. Just to make sure.

src/main/java/htsjdk/samtools/BAMRecord.java
lh3: Actually I am not familiar with this part. If SEQ='*', is readLength equal to 0? If so, your old initial code is ok.

src/main/java/htsjdk/samtools/BAMRecordCodec.java
yfarjoun: where is this restriction coming from?

lh3: In BAM, the lower 4 bits for the operator and the higher 28 bits for the length, so if either query or the reference length in the alignment longer than `(1<<28)-1`, we will not be able to encode this in BAM. Encoding this in SAM is ok. I guess htsjdk probably has this check already somewhere.

yfarjoun:
Got it. Did we say that in the spec pr?

src/main/java/htsjdk/samtools/BAMRecord.java
lh3: From @yfarjoun:

> Got it. Did we say that in the spec pr?

I don't know. This limitation is implicit in BAM (like chromosome lengths can't be longer than 2Gbp), but I agree it would be good to make it explicit anyway.

src/main/java/htsjdk/samtools/BAMRecordCodec.java
lh3: From @yfarjoun:

Got it. Did we say that in the spec pr?

I don't know. This limitation is implicit in BAM (like chromosome lengths can't be longer than 2Gbp), but I agree it would be good to make it explicit anyway.

yfarjoun: This is an extra restriction for the case of a large cigar, not in general..since we will need to put the readlength into the cigar element and there's less room there. The read=length itself is encoded in a 32-bit int. I wonder if we can try to get around this restriction (e.g. if readlength>28-bit use (1<<28-)-1 in the cigar...)

lh3: In principle, yes. We may do that if Sanger people are happy with it.

yfarjoun: shouldn't the same restriction be placed on the reference length?

yfarjoun:
@lh3 I made the fixes you asked for. a "*" read should have readlength 0, but even if not...it's not for this PR.

yfarjoun:
@tfenne @jacarey @lbergelson: @lh3 approved this PR, could one of you check that it's OK with you as well?

src/test/java/htsjdk/samtools/BAMFileWriterTest.java
```diff
@@ -60,50 +59,68 @@ private SAMRecordSetBuilder getRecordSetBuilder(final boolean sortForMe, final S
...
+        }
         if (presorted) { // If SAM text input was presorted, then we can compare SAM object to BAM object
             verifyBAMFile(samRecordSetBuilder, bamFile);
         }
+
+        File tempMetrics = File.createTempFile("CGTagTest", ".validation_metrics");
+        try (final SamReader samReader = SamReaderFactory.makeDefault().open(bamFile)) {
+
+            new SamFileValidator(new PrintWriter(new FileOutputStream(tempMetrics)),
+                    100).validateSamFileSummary(samReader, null);
```
lbergelson: strange place for a linebreak here

src/test/java/htsjdk/samtools/BAMFileWriterTest.java
```diff
@@ -60,50 +59,68 @@ private SAMRecordSetBuilder getRecordSetBuilder(final boolean sortForMe, final S
...
+            }
+        }
         if (presorted) { // If SAM text input was presorted, then we can compare SAM object to BAM object
             verifyBAMFile(samRecordSetBuilder, bamFile);
         }
+
+        File tempMetrics = File.createTempFile("CGTagTest", ".validation_metrics");
+        try (final SamReader samReader = SamReaderFactory.makeDefault().open(bamFile)) {
+
+            new SamFileValidator(new PrintWriter(new FileOutputStream(tempMetrics)),
```
lbergelson: The `PrintWriter` could be created in the resource block for auto closing

src/test/java/htsjdk/samtools/BAMFileWriterTest.java
```diff
@@ -229,4 +243,179 @@ public void testBAMReadBases() throws IOException {
...
+                {10_000},
+                {BAMRecord.MAX_CIGAR_OPERATORS - 1},
+                {BAMRecord.MAX_CIGAR_OPERATORS},
+                {BAMRecord.MAX_CIGAR_OPERATORS + 1},
+                {100_000},
+                {1_000_000}
+        };
+    }
+
+    final static CigarOperator operatorsToUse[] = new CigarOperator[]{CigarOperator.M, CigarOperator.D, CigarOperator.M, CigarOperator.I};
```
lbergelson: private?  it's weird to put the `[]` on the name instead of the type

src/test/java/htsjdk/samtools/BAMFileWriterTest.java
```diff
@@ -229,4 +243,179 @@ public void testBAMReadBases() throws IOException {
...
+
+    @Test(dataProvider = "longCigarsData")
+    public void testLongCigarsZerolengthRead(final int numOps) throws Exception {
+        final SAMRecordSetBuilder builder = new SAMRecordSetBuilder(true, SAMFileHeader.SortOrder.coordinate);
+        final Cigar cigar = Cigar.fromCigarOperators(getCigarOperatorsForTest(numOps));
+
+        final SAMRecord sam = builder.addFrag("frag1", 0, 1, false, false, cigar.toString(), null, 30);
+        sam.setReadBases(new byte[]{});
+        sam.setBaseQualityString("");
+        // in htsjdk only secondary alignments are allowed to have read-length zero (doens't validate otherwise)
```
lbergelson: or supplementary?

src/test/java/htsjdk/samtools/BAMFileWriterTest.java
```diff
@@ -229,4 +243,179 @@ public void testBAMReadBases() throws IOException {
...
+                bamWriter.addAlignment(record);
+        }
+
+        try (final SamReader reader = SamReaderFactory.makeDefault().validationStringency(ValidationStringency.SILENT).open(bamFile)) {
+            reader.iterator().forEachRemaining(SAMRecord::getCigar);
+        }
+    }
+
+    @Test(dataProvider = "longCigarsData")
+    public void testNoCGTagCigars(final int numOps) throws Exception {
```
lbergelson: I think we should add tests to show that we don't have an incorrectly empty cigar if clearAttributes is called, and also test that we don't end up with an CG tag in the case where setCigar or setCigarString is called.

src/main/java/htsjdk/samtools/BAMRecord.java
```diff
@@ -256,6 +284,56 @@ public Cigar getCigar() {
...
+        }
+
+        if (decodedCigar.getReadLength() != sentinelCigar.getReadLength()) {
+            throw new IllegalStateException(String.format(
+                    "Sentinel cigar and CG cigar should have the same read length. Found %d and %d.",
+                    sentinelCigar.getReadLength(),
+                    decodedCigar.getReadLength()));
+        }
+
+        setCigar(decodedCigar);
```
lbergelson: Shouldn't this be a call to `initializeCigar` instead of `setCigar`? This will reset the indexing bin to null.

src/test/java/htsjdk/samtools/BAMFileWriterTest.java
```diff
@@ -60,50 +59,68 @@ private SAMRecordSetBuilder getRecordSetBuilder(final boolean sortForMe, final S
...
+        try (final SamReader samReader = SamReaderFactory.makeDefault().open(bamFile)) {
+
+            new SamFileValidator(new PrintWriter(new FileOutputStream(tempMetrics)),
+                    100).validateSamFileSummary(samReader, null);
+        }
+
+        final MetricsFile<SamFileValidator.ValidationMetrics, String> validationMetrics = new MetricsFile<>();
+        validationMetrics.read(new FileReader(tempMetrics));
+
+        Assert.assertTrue(validationMetrics.getHistogram().get("ERROR:CG_TAG_FOUND_IN_ATTRIBUTES") == null);
```
lbergelson: this can be `.assertNull`

src/test/java/htsjdk/samtools/BAMFileWriterTest.java
```diff
@@ -60,50 +59,68 @@ private SAMRecordSetBuilder getRecordSetBuilder(final boolean sortForMe, final S
...
+                final SAMRecord samRecord1 = samIt1.next();
+                final SAMRecord samRecord2 = samIt2.next();
 
-            // Force reference index attributes to be populated
-            samRecord.getReferenceIndex();
-            bamRecord.getReferenceIndex();
-            samRecord.getMateReferenceIndex();
-            bamRecord.getMateReferenceIndex();
+                // SAMRecords don't have this set, so stuff it in there
+                samRecord2.setIndexingBin(samRecord1.getIndexingBin());
```
lbergelson: I think this line is masking a bug that's resetting the indexing bin.  I'm not sure what to do about that in testing though.  Maybe we should have a test specifically for the indexing bin.

src/main/java/htsjdk/samtools/BAMRecord.java
yfarjoun: yes.....though now I'm annoyed that none of the tests caught this....what is bin used for ?

src/main/java/htsjdk/samtools/BAMRecord.java
```diff
@@ -308,11 +386,17 @@ private void decodeAttributes() {
...
         }
+
         mAttributesDecoded = true;
         final int tagsOffset = readNameSize() + cigarSize() + basesSize() + qualsSize();
         final int tagsSize = mRestOfBinaryData.length - tagsOffset;
         final SAMBinaryTagAndValue attributes = BinaryTagCodec.readTags(mRestOfBinaryData, tagsOffset, tagsSize, getValidationStringency());
         setAttributes(attributes);
+
+        // if there's a CG tag, we should getCigar() so that the CG tag has a chance of turning into the CIGAR
+        if (hasAttribute(CG.name())) {
```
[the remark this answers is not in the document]
yfarjoun: hmmm not sure I understand this....

src/main/java/htsjdk/samtools/BAMRecord.java
lbergelson: I think the tests explicitly don't test this.

lbergelson: I'm honestly not certain how the bin is used.  I assume it's used in indexing?

src/main/java/htsjdk/samtools/BAMRecord.java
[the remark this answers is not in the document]
lbergelson: The potential edge cases or the short from SAMTagUtils?

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2048,7 +2048,7 @@ public int hashCode() {
             if (firstOnly) return ret;
         }
         // TODO(mccowan): Is this asking "is this the primary alignment"?
```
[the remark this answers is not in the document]
yfarjoun: not sure...reverting this, as I'm not sure it's right anymore...

src/test/java/htsjdk/samtools/BAMFileWriterTest.java
yfarjoun: nope.

src/test/java/htsjdk/samtools/BAMFileWriterTest.java
yfarjoun: lets discuss these cases in person.

yfarjoun:
@lbergelson thanks for the review. the tests for the non-null bin catch the bug you identified regarding setCigar vs. initializeCigar...thanks!

yfarjoun:
had to rebase to get rid of the FTP errors...sorry.

yfarjoun:
oops. didn't mean to close this.

yfarjoun:
I couldn't find the typo....

src/main/java/htsjdk/samtools/BAMRecord.java
```diff
@@ -284,15 +284,27 @@ public Cigar getCigar() {
...
+                (cigar.getCigarElement(0).getLength() == readLength || readLength == 0) ;
     }
 
-    private void extractCigarFromCGAttribute(final Cigar sentinelCigar) {
-        final int[] cigarFromCG = (int[]) getAttribute(CG.name());
+
+    /**
+     * Long cigars (with more than 64K operators) cannot be encoded into BAM. Instead a sentinel cigar is
+     * placed as a placeholder, and the actual cigar is placed in the CG tag. This method
+     * extracts the CIGAR from the CG tag and places is into the (in memory) cigar.
```
lbergelson: here:  is -> it

I thought I'd commented here but somehow I didn't post it.

lbergelson: Sorry, I didn't mean it to be a game of wordhunt

yfarjoun:
:party: long cigars are supported in htsjdk!!!

Issue #1076, opened 2018-01-24
CRAM implementation reports validation errors at container granularity


cmnbroad:
The CRAM implementation validates all records in a container when the container is first opened, which:

- is different than the BAM implementation, which validates each record as its retrieved
- means that when validation stringency is not SILENT, reading any record in the container throws if any other record in the container fails validation
- has at least some performance implications; the default size of a container/slice is 10000 records

Opinions welcome on whether this should be changed.

cmnbroad:
Fixed via https://github.com/samtools/htsjdk/pull/1091. Thanks @AntonMazur!

PR #1091, opened 2018-02-20
Cram implementation reports validation errors at container granularity

AntonMazur:
### Description
Associated issue: https://github.com/samtools/htsjdk/issues/1076

#### Motivation
- CRAM implementation is different then BAM implementation and validates all records in container when it first opened, but BAM implementation validated each record as its retrieved
- CRAM iterator contained useless memory allocation for 10000 elements of `ArrayList<SAMRecord> records;`

#### Solution 
The first problem was solved by deletion of validation from cramIterator.nextContainer() method, and addition it in cramIterator.next(), for validation records when they retrieves.
And the second problem was solved by deletion of objects reservation.

### Checklist

- [X] Code compiles correctly
- [X] All tests passing


src/main/java/htsjdk/samtools/CRAMIterator.java
```diff
@@ -106,7 +106,7 @@ public CRAMIterator(final SeekableStream seekableStream, final CRAMReferenceSour
         this.containerIterator = containerIterator;
 
         firstContainerOffset = containerIterator.getFirstContainerOffset();
-        records = new ArrayList<SAMRecord>(10000);
+        records = new ArrayList<SAMRecord>();
```
magicDGS: How many records are expected in each container? This could have some performance effect, but I understand that there is no explanation for the magic number. Should it be extracted as an default constant at the class level or maybe to an argument to the iterator?

vadimzalunin:
10k is/used to be container default size but you are right of course.

src/main/java/htsjdk/samtools/CRAMIterator.java
AntonMazur: @magicDGS hello, thanks for your review, I've extracted it as constant at the class level

src/test/java/htsjdk/samtools/CRAMIteratorTest.java
```diff
@@ -0,0 +1,50 @@
...
+import htsjdk.samtools.cram.ref.ReferenceSource;
+import htsjdk.samtools.seekablestream.SeekableStream;
+import org.junit.Assert;
+import org.testng.annotations.Test;
+
+import java.io.File;
+
+
+/**
+ * @author Anton_Mazur@epam.com, EPAM Systems, Inc.
```
cmnbroad: This is a test file, so it doesn't matter, but block comments like "author" need to go after the main javadoc or else the entire javadoc text becomes part of the author block.

src/test/java/htsjdk/samtools/CRAMIteratorTest.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * This test serve for verifying CRAMIterator records validation using strict validation strategy
+ **/
+
+public class CRAMIteratorTest extends HtsjdkTest {
+    private static final File refFile = new File("src/test/resources/htsjdk/samtools/cram/ce.fa");
+    private static final File cramFile = new File("src/test/resources/htsjdk/samtools/cram/ce#supp.3.0.cram");
```
cmnbroad: Presumably this file has some records that fail validation. It would be better to make a copy of it with a name that indicates that, and reinforce it by using a variable name here that reflects it.

src/test/java/htsjdk/samtools/CRAMIteratorTest.java
```diff
@@ -0,0 +1,50 @@
...
+    ReferenceSource source = new ReferenceSource(refFile);
```
cmnbroad: We don't publish a style guide, but for new code we try to use "final" everywhere possible (paramaters, locals, etc.).

src/test/java/htsjdk/samtools/CRAMIteratorTest.java
```diff
@@ -0,0 +1,50 @@
...
+        Assert.assertTrue(cramIter.hasNext());
+    }
+
+    @Test(expectedExceptions = SAMException.class)
+    public void shouldThrowExceptionIfCRAMFileContainsInvalidRecods() {
+        SAMRecordIterator cramIter =
+                getCramFileIterator(cramFile, source, ValidationStringency.STRICT);
+
+        while (cramIter.hasNext())
+        {
```
cmnbroad: We don't publish a style guide, but we generally try to put opening braces on the line with the previous statement.

src/test/java/htsjdk/samtools/CRAMIteratorTest.java
cmnbroad: Also, although many of the test are written this same way, it would be preferable for each test to do its own setup rather than relying on class-level state.

src/test/java/htsjdk/samtools/CRAMIteratorTest.java
```diff
@@ -0,0 +1,50 @@
...
+            cramIter.next();
+        }
+    }
+
+    private SAMRecordIterator getCramFileIterator(File cramFile,
+                                                  ReferenceSource source,
+                                                  ValidationStringency valStrigency) {
```
cmnbroad: valStrigency -> valStringency

src/test/java/htsjdk/samtools/CRAMIteratorTest.java
```diff
@@ -0,0 +1,50 @@
...
+    @Test
+    public void shouldNotThrowsExceptionIfCRAMfileContainsInvalidRecords() {
+        SAMRecordIterator cramIter =
+                getCramFileIterator(cramFile, source, ValidationStringency.STRICT);
+
+        Assert.assertTrue(cramIter.hasNext());
+    }
+
+    @Test(expectedExceptions = SAMException.class)
+    public void shouldThrowExceptionIfCRAMFileContainsInvalidRecods() {
```
cmnbroad: maybe rename to `throwOnRecordValidationFailure`.

src/test/java/htsjdk/samtools/CRAMIteratorTest.java
```diff
@@ -0,0 +1,50 @@
...
+    private static final File cramFile = new File("src/test/resources/htsjdk/samtools/cram/ce#supp.3.0.cram");
+    ReferenceSource source = new ReferenceSource(refFile);
+
+    @Test
+    public void shouldNotThrowsExceptionIfCRAMfileContainsInvalidRecords() {
+        SAMRecordIterator cramIter =
+                getCramFileIterator(cramFile, source, ValidationStringency.STRICT);
+
+        Assert.assertTrue(cramIter.hasNext());
+    }
```
cmnbroad: This test needs a comment explaining what its testing (specifically, that validation is deferred until records are retrieved, rather than when cracking open the container), and a name that reflects that. Also, it would be helpful to reinforce that in the code by naming the iterator something like "cramIteratorOverInvalidRecords".

src/test/java/htsjdk/samtools/CRAMIteratorTest.java
cmnbroad: One other comment I forgot to make; since this test is expected to throw an exception, it would be a good idea to use a try/finally block to close the reader.

src/main/java/htsjdk/samtools/CRAMIterator.java
```diff
@@ -43,6 +43,7 @@
 
 public class CRAMIterator implements SAMRecordIterator {
     private static final Log log = Log.getInstance(CRAMIterator.class);
+    private static final int DEFAULT_CONTAINER_CAPACITY = 10000;
```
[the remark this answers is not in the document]
AntonMazur: @cmnbroad hello, thanks for your review, all requested changes are done

src/test/java/htsjdk/samtools/CRAMIteratorTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Test(expectedExceptions = SAMException.class)
+    public void throwOnRecordValidationFailure() {
+        final File refFile = new File("src/test/resources/htsjdk/samtools/cram/ce.fa");
+        final File cramFileWithInvalidRecs = new File("src/test/resources/htsjdk/samtools/cram/ce#containsInvalidRecords.3.0.cram");
+        final ReferenceSource source = new ReferenceSource(refFile);
+        final SAMRecordIterator cramIteratorOverInvalidRecords =
+                getCramFileIterator(cramFileWithInvalidRecs, source, ValidationStringency.STRICT);
+        try{
```
cmnbroad: try-with-resources

src/test/java/htsjdk/samtools/CRAMIteratorTest.java
```diff
@@ -0,0 +1,56 @@
...
+        cramIteratorOverInvalidRecords.close();
+    }
+
+    @Test(expectedExceptions = SAMException.class)
+    public void throwOnRecordValidationFailure() {
+        final File refFile = new File("src/test/resources/htsjdk/samtools/cram/ce.fa");
+        final File cramFileWithInvalidRecs = new File("src/test/resources/htsjdk/samtools/cram/ce#containsInvalidRecords.3.0.cram");
+        final ReferenceSource source = new ReferenceSource(refFile);
+        final SAMRecordIterator cramIteratorOverInvalidRecords =
+                getCramFileIterator(cramFileWithInvalidRecs, source, ValidationStringency.STRICT);
```
cmnbroad: Most of the code up to this point is duplicated in both test methods. Can you factor that out (maybe into `getCramFileIterator`).

src/main/java/htsjdk/samtools/CRAMIterator.java
```diff
@@ -271,7 +263,12 @@ public boolean hasNext() {
 
     @Override
     public SAMRecord next() {
-        return iterator.next();
+        SAMRecord samRecord = iterator.next();
+        if (validationStringency != ValidationStringency.SILENT) {
+            SAMUtils.processValidationErrors(samRecord.isValid(), samRecordIndex++, validationStringency);
```
cmnbroad: Since `samRecordIndex` is a class not a local variable, can you add a comment to the declaration saying that it is only used when Validation Stringency is not SILENT, and otherwise it isn't valid. (this was always true even before this PR, but it would be good to make that clear).

src/test/java/htsjdk/samtools/CRAMIteratorTest.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * This test serve for verifying CRAMIterator records validation using strict validation strategy
+ *
+ * @author Anton_Mazur@epam.com, EPAM Systems, Inc.
+ **/
+
+public class CRAMIteratorTest extends HtsjdkTest {
+
+    @Test(description = "This test checks that records validation is deferred until they are retrieved")
+    public void notThrowOnOpeningContainerWithInvalidRecords() {
```
cmnbroad: Maybe rename this to "noValidationFailureOnContainerOpen".

src/test/java/htsjdk/samtools/CRAMIteratorTest.java
```diff
@@ -0,0 +1,56 @@
...
+        final File refFile = new File("src/test/resources/htsjdk/samtools/cram/ce.fa");
+        final File cramFileWithInvalidRecs = new File("src/test/resources/htsjdk/samtools/cram/ce#containsInvalidRecords.3.0.cram");
+        final ReferenceSource source = new ReferenceSource(refFile);
+        final SAMRecordIterator cramIteratorOverInvalidRecords =
+                getCramFileIterator(cramFileWithInvalidRecs, source, ValidationStringency.STRICT);
+
+        Assert.assertTrue(cramIteratorOverInvalidRecords.hasNext());
+        cramIteratorOverInvalidRecords.close();
```
cmnbroad: SAMRecordIterator is closable, so you should be able to use try-with-resources here.

AntonMazur: @cmnbroad Hello! Thank you for review, requested changes are done

cmnbroad:
@AntonMazur Just pinging you one more time - on this. I'll plan to update and rebase this branch if I don't hear back from you soon, since I thinks pretty close, but has conflicts that need to be resolved.

src/main/java/htsjdk/samtools/CRAMIterator.java
```diff
@@ -71,6 +70,10 @@ public void setValidationStringency(
         this.validationStringency = validationStringency;
     }
 
+    /**
+     * `samRecordIndex` only used when validation is not `SILENT`
+     * (for tracking numbers of invalid records(if it's present) and to use it in exception's message)
+     */
```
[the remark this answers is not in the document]
AntonMazur: Hello! Sorry for such delay, I've corrected this.

PR #1093, opened 2018-02-23
Adding check on validator on VCFHeadLineCount

ffinfo:
### Description

A header with a incorrect count number can be created but after that is not readable anymore with htsjdk itself. Adding a exception to prevent this from happening.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
+++ b/src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
@@ -202,12 +202,12 @@ public abstract class VCFCompoundHeaderLine extends VCFHeaderLine implements VCF
     }
 
     private void validate() {
-        if (countType == VCFHeaderLineCount.INTEGER && count <= 0)
+        if (type != VCFHeaderLineType.Flag && countType == VCFHeaderLineCount.INTEGER && count <= 0)
             throw new IllegalArgumentException(String.format("Invalid count number, with fixed count the number should be 1 or higher: key=%s name=%s type=%s desc=%s lineType=%s count=%s",
-                    super.getKey(), name, type, description, lineType, count));
+                    getKey(), name, type, description, lineType, count));
         if (name == null || type == null || description == null || lineType == null)
             throw new IllegalArgumentException(String.format("Invalid VCFCompoundHeaderLine: key=%s name=%s type=%s desc=%s lineType=%s",
-                    super.getKey(), name, type, description, lineType));
+                    getKey(), name, type, description, lineType));
         if (name.contains("<") || name.contains(">"))
             throw new IllegalArgumentException("VCFHeaderLine: ID cannot contain angle brackets");
         if (name.contains("="))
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
```diff
@@ -202,6 +202,9 @@ protected VCFCompoundHeaderLine(String line, VCFHeaderVersion version, Supported
     }
 
     private void validate() {
+        if (countType == VCFHeaderLineCount.INTEGER && count <= 0)
+            throw new IllegalArgumentException(String.format("Invalid count number, with fixed count the number should be 1 or higher: key=%s name=%s type=%s desc=%s lineType=%s count=%s",
+                    super.getKey(), name, type, description, lineType, count));
```
pshapiro4broad: `super.` is unnecessary here as `getKey()` is not overridden

src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
```diff
@@ -202,6 +202,9 @@ protected VCFCompoundHeaderLine(String line, VCFHeaderVersion version, Supported
     }
 
     private void validate() {
+        if (countType == VCFHeaderLineCount.INTEGER && count <= 0)
```
pshapiro4broad: The code style in this file is inconsistent, but it would be nice to migrate it to use `{` `}` for `if` blocks. Can you migrate the ones in this method as part of this PR?

src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
ffinfo: can remove it, just did the same as 1 line down ;)

src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
pshapiro4broad: Does this check conflict with the one on line 213 below that sets `count` to `0`? Maybe this validation should be done after that check?

ffinfo:
Will look into the testing, it seems CountType.Integer is also used for type=FLAG

ffinfo:
@pshapiro4broad: I did add some testing for this now.
Must say testing in java is very verbose compared to scala ;)

magicDGS:
@ffinfo - you can also implement scala tests in HTSJDK: see https://github.com/samtools/htsjdk/wiki/Writing-tests-in-HTSJDK if you are interested!

ffinfo:
@magicDGS: Good to know, will take a look at it.
For now this works ( also because it's already there now ;) ), this is still small enough. For larger PR's this is really useful

yfarjoun:
LGTM as well. I think that this is small enough to not need the maintainers input.

yfarjoun:
This is now failing after being merged into master... the failures are all FTP related...so I restarted the build..I hope it passes.

yfarjoun:
whew! now it passes on master..

PR #1094, opened 2018-02-26
Cache result of OverlapDetector.getAll for speed

alecw:
### Description

The result of getAll() method is cached, with cache invalidated if an object-modifying method is called.  For my test program, this reduces run time by about 1/3.  Note that the cache is not thread-safe, but this class is already not thread-safe.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/OverlapDetector.java
+++ b/src/main/java/htsjdk/samtools/util/OverlapDetector.java
@@ -103,7 +103,6 @@ public class OverlapDetector<T> {
      * and the corresponding objects.
      */
     public void addAll(final List<T> objects, final List<? extends Locatable> intervals) {
-        invalidateAllObjectsCache();
         if (objects == null) {
             throw new IllegalArgumentException("null objects");
         }
--- a/src/main/java/htsjdk/samtools/util/OverlapDetector.java
+++ b/src/main/java/htsjdk/samtools/util/OverlapDetector.java
@@ -23,6 +23,7 @@
  */
 package htsjdk.samtools.util;
 
+import java.lang.ref.SoftReference;
 import java.util.*;
 
 /**
@@ -42,7 +43,7 @@ public class OverlapDetector<T> {
     private final int lhsBuffer;
     private final int rhsBuffer;
 
-    private Set<T> allObjectsCache = null;
+    private transient SoftReference<Set<T>> allObjectsCache = new SoftReference<>(null);
 
     /**
      * Constructs an overlap detector.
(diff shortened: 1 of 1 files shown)


lbergelson:
@alecw What's the motivation for this?  It seems like the requester could just hang onto the result.  Why does it need to be cached within the object?

src/main/java/htsjdk/samtools/util/OverlapDetector.java
```diff
@@ -42,6 +42,8 @@
     private final int lhsBuffer;
     private final int rhsBuffer;
 
+    private Set<T> allObjectsCache = null;
```
pshapiro4broad: `= null` is unnecessary here since that's the value the field already has

src/main/java/htsjdk/samtools/util/OverlapDetector.java
```diff
@@ -100,6 +103,7 @@ public void addLhs(final T object, final Locatable interval) {
      * and the corresponding objects.
      */
     public void addAll(final List<T> objects, final List<? extends Locatable> intervals) {
+        invalidateAllObjectsCache();
```
pshapiro4broad: This is unnecessary as this method calls `addLhs()` which also invalidates the cache.

alecw:
Hi @lbergelson , good question.  The motivation is that the invocation is in code that I don't control (Picard), and they suggested that putting a cache in htsjdk would make more sense.  See https://github.com/broadinstitute/picard/pull/1121.  

I thought this was a worthwhile addition that others might appreciate, but if not, I think I can write a wrapper for OverlapDetector and put caching there.

src/main/java/htsjdk/samtools/util/OverlapDetector.java
alecw: I'm inclined to leave the assignment for clarity, unless you really object.

alecw:
Should I merge this based on @pshapiro4broad 's approval, or do I wait until @lbergelson weighs in again?

src/main/java/htsjdk/samtools/util/OverlapDetector.java
lbergelson: If we want to include this I think it would be a good idea to make this be 
- 1 a `transient`field (even thought it's not serializable, I think it's a good idea to mark it so it's clear)
- 2 possibly make it into a `SoftReference` so that it won't blow up the memory requirements.  The overlap detector can potentially be storing a rather large set of things, and unintentionally doubling it could be problematic.

lbergelson:
I think this may be adding unnecessary complexity when it could really just be left up to the client.  I'm not sure that repeatedly calling `getAll` is very common behavior with OverlapsDetector and adding caching forces people to think much harder when they modify the class in the future.  

It's not uncommon to have very large OverlapDetectors, so accidentally doubling the memory required is not a nice behavior.  I'd vote for doing the caching in a layer in picard, but I could be convinced otherwise if people feel strongly.

alecw:
OK.  Unless @pshapiro4broad disagrees, I'll retract this request.

alecw:
Thanks, @lbergelson and @pshapiro4broad , I'm going to close this and address my issue a different way.

PR #1098, opened 2018-03-07
Fixed 1097 bug where GL field overrides PL field. 

yfarjoun:
When GL and PL are inconsistent, and GL is "after" PL, the resulting genotype has the wrong PL value.

- fixed bug
- added test which failed prior to the change.


yfarjoun:
to the reviewer: use `?w=1` as many lines are just whitespace changes (tabs->spaces)

src/test/java/htsjdk/variant/vcf/AbstractVCFCodecTest.java
```diff
@@ -12,49 +12,63 @@
...
+        Assert.assertEquals(AbstractVCFCodec.canDecodeFile(potentialInput, VCFCodec.VCF4_MAGIC_HEADER), canDecode);
+    }
+
+    @Test
+    public void testGetTabixFormat() {
+        Assert.assertEquals(new VCFCodec().getTabixFormat(), TabixFormat.VCF);
+        Assert.assertEquals(new VCF3Codec().getTabixFormat(), TabixFormat.VCF);
+    }
+
+    private void assertIntArraysAreEqual(int[] v1, int[] v2) {
```
magicDGS: I think that TestNG has already this method (https://github.com/cbeust/testng/blob/master/src/main/java/org/testng/Assert.java#L238)

yfarjoun: I couldn't find that in the version that we have in htsjdk....I could rev TestNG I guess.

magicDGS: Fair enough. I just didn't know in which version it was introduced...

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -719,7 +719,10 @@ private static boolean isVCFStream(final InputStream stream, final String MAGIC_
                         } else if (gtKey.equals(VCFConstants.GENOTYPE_PL_KEY)) {
                             gb.PL(decodeInts(genotypeValues.get(i)));
                         } else if (gtKey.equals(VCFConstants.GENOTYPE_LIKELIHOODS_KEY)) {
-                            gb.PL(GenotypeLikelihoods.fromGLField(genotypeValues.get(i)).getAsPLs());
+                            // Do not overwrite PL with data from GL
+                            if (!gb.make().hasPL()) {
+                                gb.PL(GenotypeLikelihoods.fromGLField(genotypeValues.get(i)).getAsPLs());
```
lbergelson: @yfarjoun Thanks for fixing this.  I don't love the call to make() in this if statement.  It would be nice to avoid any extra allocations.  It might be worth adding a getPL method to the builder in order to avoid having to do that.

src/test/java/htsjdk/variant/vcf/AbstractVCFCodecTest.java
```diff
@@ -12,49 +12,56 @@
...
+    public void testGetTabixFormat() {
+        Assert.assertEquals(new VCFCodec().getTabixFormat(), TabixFormat.VCF);
+        Assert.assertEquals(new VCF3Codec().getTabixFormat(), TabixFormat.VCF);
+    }
+
+    @Test
+    public void testGLnotOverridePL() {
+        VCFFileReader reader = new VCFFileReader(new File("src/test/resources/htsjdk/variant/test_withGLandPL.vcf"), false);
+        VariantContext variant = reader.iterator().next();
+        reader.close();
```
lbergelson: always better to do this in a try-with-resources instead of the manual call to close

yfarjoun:
I don't want to be detecting PL/GL disagreements, but I think that overwriting the PL with a transformed GL is wrong...this is what I'm protecting against.

Issue #958, opened 2017-07-29
getAnyAttribute returns String instead of list


dariober:
Hello again,

Given this vcf file:

```
##fileformat=VCFv4.2
##FORMAT=<ID=XP,Number=2,Type=Integer,Description="test">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	sample1
chr1	10	.	C	T	.	.	.	XP	10,20
```

I would expect `getAnyAttribute("XP")` to return a `List<Integer>` since the data type of XP is Integer and the count of elements is 2. However, `getAnyAttribute("XP")` returns a String.

Test:

```
@Test
public void test(){
	VCFFileReader reader = new VCFFileReader(new File("test2.vcf.gz"));
	VariantContext ctx = reader.query("chr1", 1, 100).next();


	System.err.println(reader.getFileHeader().getFormatHeaderLine("XP").getCount()); // <- 2 OK
	System.err.println(ctx.getGenotype(0).getAnyAttribute("XP")); // <- prints: 10,20

        // Returns "class java.lang.String". Expected: List<Integer> or similar
	System.err.println(ctx.getGenotype(0).getAnyAttribute("XP").getClass()); 
}
```

Is this behavior expected?

This is with htsjdk 2.10.1 and Java 1.8

Thank you for any feedback!

magicDGS:
This is one of the flaws in the HTSJDK code parsing VCF (as far as I know): HTSJDK does not take into account the definitions in the header and thus any attribute parsed by the framework will return a `String` unless it is an inline-handled attribute (such as DP or GT).

I think that a big refactoring of the `VariantContext`/`Genotype` is required for fixing this kind of things, supporting even more the idea that a new API is required and be based in interfaces with properly defined contracts.

In addition, it will be nice if the codec for parsing VCF uses the header to parse all attributes. Or if  header lines are required to retrieve specific attributes (allowing to "guess" the type of the attribute and see for discordances).

dariober:
Thanks for taking care of this magicDGS. 

For the record, if I'm not mistaken the FORMAT and INFO values can contain comma only if the comma is used as item separator in lists. (See http://samtools.github.io/hts-specs/VCFv4.3.pdf section 1.2, although in previous versions is not obvious to me if this restriction is enforced). If this is correct, it should be safe to split-to-list the value string at commas and cast to values to Integer or null if '.'.

PR #1100, opened 2018-03-19
getAnyAttribute returns String instead of List<Integer>

AntonMazur:
### Description
Associated issue: #958 

#### Motivation
Genotype.getAnyAttribute(...) always returns String independently of attribute's type specified in format lines of *.vcf file's header

#### Solution 
For resolving this issue there were made some changes in `AbstractVCFCodec`:
1) Added method `genotypeValsToTargetType()` which defines how many values the genotype attribute has: if it's single, this method will return result parsed to target type (that specified in related FORMAT vcf header line). If genotype attribute's key is associated with multiple attributes, current method will call another new method `mapToFormatLineType()`.
2) Added method `mapToFormatLineType()`  which iterates across genotype attribute's values and put it after parsing in result List. 
3) Added method `getMapperFunction()` which gets `VCFHeaderLineType` and returns reference on corresponding to the argument parsing function.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -750,6 +750,56 @@ private static boolean isVCFStream(final InputStream stream, final String MAGIC_
...
     }
 
+    private Object genotypeValsToTargetType(final String key, final String vals) {
+        if (!header.hasFormatLine(key)) {
+            return vals;
+        }
+
+        final VCFFormatHeaderLine formLine = header.getFormatHeaderLine(key);
+        final VCFHeaderLineType valType = formLine.getType();
+        final boolean isSingleVal = formLine.isFixedCount()
```
yfarjoun: this doesn't seem to be a reasonable way to find the count.....I think you should consider the count type (A, R, G, etc) instead.

yfarjoun: please add a test with a A-type FORMAT field, and a string that contains commas as the value.

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -750,6 +750,56 @@ private static boolean isVCFStream(final InputStream stream, final String MAGIC_
...
+            if (!genVal.equals(VCFConstants.MISSING_VALUE_v4)) {
+                genRealTypeVals.add(mapperFunc.apply(genVal));
+            }
+        });
+
+        return genRealTypeVals;
+    }
+
+    private Function<String, Object> getMapperFunction(final VCFHeaderLineType genValType) {
+        if (genValType == VCFHeaderLineType.Integer) {
```
lbergelson: this should be a switch that handles all known types and has a default that explodes

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -750,6 +750,56 @@ private static boolean isVCFStream(final InputStream stream, final String MAGIC_
...
+            return genVals;
+        }
+
+        final List<Object> genRealTypeVals = new ArrayList<>(genVals.size());
+        final Function<String, Object> mapperFunc = getMapperFunction(genValType);
+
+        genVals.forEach(genVal -> {
+            if (!genVal.equals(VCFConstants.MISSING_VALUE_v4)) {
+                genRealTypeVals.add(mapperFunc.apply(genVal));
+            }
```
yfarjoun: something else should happen if there's a missing value in the middle of a list of strings....so "0,1,.,2" should NOT convert to [0,1,2] but rather something like [0, 1, null, 2] or [0, 1, MISSING_INT_VALUE, 2]

src/test/java/htsjdk/variant/variantcontext/GenotypeUnitTest.java
```diff
@@ -31,10 +31,15 @@
...
 import htsjdk.variant.vcf.VCFConstants;
+import htsjdk.variant.vcf.VCFFileReader;
 import org.testng.Assert;
 import org.testng.annotations.BeforeSuite;
 import org.testng.annotations.Test;
 
+import java.io.File;
+import java.util.Arrays;
+import java.util.List;
+
```
yfarjoun: extra nl

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
AntonMazur: @yfarjoun, @lbergelson Hello! Thanks for review! 
All requested changes are done

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -750,6 +750,45 @@ private static boolean isVCFStream(final InputStream stream, final String MAGIC_
...
     }
 
+    private Object genotypeValsToTargetType(final String key, final String vals) {
+        if (!header.hasFormatLine(key)) {
+            return vals;
+        }
+
+        final VCFFormatHeaderLine formLine = header.getFormatHeaderLine(key);
+        final VCFHeaderLineType valType = formLine.getType();
+        final boolean isSingleVal = !vals.contains(VCFConstants.GENOTYPE_ATTRIBUTE_VALUE_SEPARATOR);
```
yfarjoun: check count or if unspecified, then resort to counting commas

src/test/java/htsjdk/variant/variantcontext/GenotypeUnitTest.java
```diff
@@ -71,6 +74,24 @@ public void testFilters() {
...
+        Assert.assertEquals(gt.getAnyAttribute("XP"), Arrays.asList(1, 2));
+        Assert.assertEquals(gt.getAnyAttribute("XE"), Arrays.asList(1.1, 1.2));
+        Assert.assertEquals(gt.getAnyAttribute("XR"), Arrays.asList("a", "b"));
+        Assert.assertEquals(gt.getAnyAttribute("XT"), Arrays.<Object>asList('a', 'b'));
+        Assert.assertEquals(gt.getAnyAttribute("XU"), Arrays.<Object>asList(1, 2, null));
+        Assert.assertEquals(gt.getAnyAttribute("XY"), Arrays.asList(null, null));
+        Assert.assertEquals(gt.getAnyAttribute("XJ"), Arrays.asList("", "", "", ""));
+        Assert.assertEquals(gt.getAnyAttribute("XI"), 1);
+        Assert.assertEquals(gt.getAnyAttribute("XO"), 1.1);
+        Assert.assertEquals(gt.getAnyAttribute("XQ"), 'a');
```
yfarjoun: Please test flag tags as well.

src/test/resources/htsjdk/variant/analyzeHeaderFormatLines.vcf
```diff
@@ -0,0 +1,13 @@
...
+##FORMAT=<ID=XR,Number=R,Type=String,Description="test">
+##FORMAT=<ID=XT,Number=2,Type=Character,Description="test">
+##FORMAT=<ID=XY,Number=2,Type=Integer,Description="test">
+##FORMAT=<ID=XU,Number=3,Type=Integer,Description="test">
+##FORMAT=<ID=XI,Number=1,Type=Integer,Description="test">
+##FORMAT=<ID=XO,Number=1,Type=Float,Description="test">
+##FORMAT=<ID=XQ,Number=1,Type=Character,Description="test">
+##FORMAT=<ID=XJ,Number=2,Type=String,Description="test">
+#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	sample1
+chr1	10	.	C	T	.	.	.	XP:XE:XR:XT:XY:XU:XI:XO:XQ:XJ	1,2:1.1,1.2:a,b:a,b:.,.:1,2,.:1:1.1:a:,,,
```
yfarjoun: Can you convert the vcf.gz and generate the index on the fly in the test rather than including it as a pair of opaque files? (I'll create a PR shortly that grabs the code that does this from picard)

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -750,6 +750,45 @@ private static boolean isVCFStream(final InputStream stream, final String MAGIC_
...
+                : mapToFormatLineType(
+                        ParsingUtils.split(vals, VCFConstants.GENOTYPE_ATTRIBUTE_VALUE_SEPARATOR_CHAR),
+                        valType);
+    }
+
+    private List<?> mapToFormatLineType(final List<String> genVals, final VCFHeaderLineType genValType) {
+        final List<Object> genRealTypeVals = new ArrayList<>(genVals.size());
+        final Function<String, Object> mapperFunc = getMapperFunction(genValType);
+
+        genVals.forEach(genVal -> genRealTypeVals.add(
```
lbergelson: these should catch parsing exceptions and convert them to TribbleException explaining what happened

PR #1109, opened 2018-05-01
Have SortingCollection check prior to final sort to adjust file buffer size given the amount of memory we have left.

jacarey:
Added debugging messages that estimate the size per record for further tweaking if needed.
Use java 8 parallel sort.

These changes address issues seen in https://broadinstitute.atlassian.net/browse/DSDEGP-2232

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/SortingCollection.java
+++ b/src/main/java/htsjdk/samtools/util/SortingCollection.java
@@ -66,15 +66,11 @@ public class SortingCollection<T> implements Iterable<T> {
     public interface Codec<T> extends Cloneable {
         /**
          * Where to write encoded output
-         *
-         * @param os
          */
         void setOutputStream(OutputStream os);
 
         /**
          * Where to read encoded input from
-         *
-         * @param is
          */
         void setInputStream(InputStream is);
 
@@ -133,7 +129,7 @@ public class SortingCollection<T> implements Iterable<T> {
 
     private boolean destructiveIteration = true;
 
-    private TempStreamFactory tempStreamFactory = new TempStreamFactory();
+    private final TempStreamFactory tempStreamFactory = new TempStreamFactory();
 
     /**
      * Prepare to accumulate records to be sorted
@@ -324,7 +320,7 @@ public class SortingCollection<T> implements Iterable<T> {
                                                        final Comparator<T> comparator,
                                                        final int maxRecordsInRAM,
                                                        final File... tmpDir) {
-        return new SortingCollection<T>(componentType, codec, comparator, maxRecordsInRAM, Arrays.stream(tmpDir).map(File::toPath).toArray(Path[]::new));
+        return new SortingCollection<>(componentType, codec, comparator, maxRecordsInRAM, Arrays.stream(tmpDir).map(File::toPath).toArray(Path[]::new));
 
     }
 
@@ -344,7 +340,7 @@ public class SortingCollection<T> implements Iterable<T> {
                                                        final Comparator<T> comparator,
                                                        final int maxRecordsInRAM,
                                                        final Collection<File> tmpDirs) {
-        return new SortingCollection<T>(componentType,
+        return new SortingCollection<>(componentType,
                 codec,
                 comparator,
                 maxRecordsInRAM,
@@ -367,7 +363,7 @@ public class SortingCollection<T> implements Iterable<T> {
                                                        final int maxRecordsInRAM) {
 
         final Path tmpDir = Paths.get(System.getProperty("java.io.tmpdir"));
-        return new SortingCollection<T>(componentType, codec, comparator, maxRecordsInRAM, tmpDir);
+        return new SortingCollection<>(componentType, codec, comparator, maxRecordsInRAM, tmpDir);
     }
 
     /**
@@ -384,7 +380,7 @@ public class SortingCollection<T> implements Iterable<T> {
                                                        final Comparator<T> comparator,
                                                        final int maxRecordsInRAM,
                                                        final Path... tmpDir) {
-        return new SortingCollection<T>(componentType, codec, comparator, maxRecordsInRAM, tmpDir);
+        return new SortingCollection<>(componentType, codec, comparator, maxRecordsInRAM, tmpDir);
 
     }
 
@@ -402,7 +398,7 @@ public class SortingCollection<T> implements Iterable<T> {
                                                                 final Comparator<T> comparator,
                                                                 final int maxRecordsInRAM,
                                                                 final Collection<Path> tmpDirs) {
-        return new SortingCollection<T>(componentType,
+        return new SortingCollection<>(componentType,
                 codec,
                 comparator,
                 maxRecordsInRAM,
@@ -472,7 +468,7 @@ public class SortingCollection<T> implements Iterable<T> {
             this.queue = new TreeSet<>(new PeekFileRecordIteratorComparator());
             int n = 0;
             log.info(String.format("Creating merging iterator from %d files", files.size()));
-            int suggestedBufferSize = checkMemoryAndAdjustBuffer(files.size(), Defaults.BUFFER_SIZE);
+            int suggestedBufferSize = checkMemoryAndAdjustBuffer(files.size());
             for (final Path f : files) {
                 final FileRecordIterator it = new FileRecordIterator(f, suggestedBufferSize);
                 if (it.hasNext()) {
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/util/SortingCollection.java
```diff
@@ -153,6 +150,7 @@ private SortingCollection(final Class<T> componentType, final SortingCollection.
         this.codec = codec;
         this.comparator = comparator;
         this.maxRecordsInRam = maxRecordsInRam;
+
```
als364: intentional newline?

src/main/java/htsjdk/samtools/util/SortingCollection.java
```diff
@@ -448,19 +464,43 @@ public void remove() {
...
-                else {
+                } else {
                     it.close();
                 }
             }
         }
 
+        // Since we need to open and buffer all temp files in the sorting collection at once it is important
+        // to have enough memory left to do this. This method checks to make sure that, given the number of files and
+        // the size of the buffer, we can reasonable open all files. If we can't it will return a buffer size that
```
als364: "reasonable" => "reasonably"

src/main/java/htsjdk/samtools/util/StringUtil.java
```diff
@@ -606,4 +606,18 @@ public static boolean isWithinHammingDistance(final String s1, final String s2,
...
+     * @param bytes The number of bytes to create a human readable string for.
+     * @return A human readable string of the number of bytes given.
+     */
+    public static String humanReadableByteCount(long bytes) {
+        if (bytes < 1024) {
+            return bytes + " B";
+        }
+
+        int exp = (int) (Math.log(bytes) / Math.log(1024));
+        return String.format("%.1f %sB", bytes / Math.pow(1024, exp), "kMGTPE".charAt(exp - 1));
```
als364: you actually think we'll need exabyte-scale memory? 😛

jacarey: haha probably not.. but we had one cloud attached drive that has exabytes... so.. :man_shrugging:

src/main/java/htsjdk/samtools/util/SortingCollection.java
```diff
@@ -153,6 +150,7 @@ private SortingCollection(final Class<T> componentType, final SortingCollection.
         this.codec = codec;
         this.comparator = comparator;
         this.maxRecordsInRam = maxRecordsInRam;
+
         this.ramRecords = (T[])Array.newInstance(componentType, maxRecordsInRam);
```
pshapiro4broad: It's not clear to me why this code isn't using `ArrayList` here. There is only a small overhead for using it over a bare array, and it wouldn't require the bogus `componentType` argument to `SortingCollection`.

If the ship has sailed on this API, it would at least be better to use a narrower scope for the `SuppressWarnings` annotation. You can do this by adding an intermediate variable, e.g.
```
        T[] ramRecords = (T[]) Array.newInstance(componentType, maxRecordsInRam);
        this.ramRecords = ramRecords
```

src/main/java/htsjdk/samtools/util/SortingCollection.java
```diff
@@ -139,6 +135,7 @@
      * @param maxRecordsInRam how many records to accumulate before spilling to disk
      * @param tmpDir Where to write files of records that will not fit in RAM
      */
+    @SuppressWarnings("unchecked")
```
pshapiro4broad: It is unclear to me why this code doesn't use `ArrayList` for `ramRecords`, as there is very little overhead if you preallocate at construction time. That would the remove the need for this annotation and the bogus `componentType` constructor argument.

If the API can't be changed at this point it would at least be better to narrow the scope of the `SuppressWarnings` annotation. This can be done by introducing a local variable, e.g.
```
        T[] ramRecords = (T[]) Array.newInstance(componentType, maxRecordsInRam);
        this.ramRecords = ramRecords;
```

src/main/java/htsjdk/samtools/util/SortingCollection.java
```diff
@@ -215,45 +233,42 @@ public void setDestructiveIteration(boolean destructiveIteration) {
...
         try {
-            Arrays.sort(this.ramRecords, 0, this.numRecordsInRam, this.comparator);
+            Arrays.parallelSort(this.ramRecords, 0, this.numRecordsInRam, this.comparator);
+
             final Path f = newTempFile();
-            OutputStream os = null;
-            try {
-                os = tempStreamFactory.wrapTempOutputStream(Files.newOutputStream(f), Defaults.BUFFER_SIZE);
+            try (OutputStream os
+                         = tempStreamFactory.wrapTempOutputStream(Files.newOutputStream(f), Defaults.BUFFER_SIZE)){
```
pshapiro4broad: need space before `{`

src/main/java/htsjdk/samtools/util/SortingCollection.java
```diff
@@ -215,45 +233,42 @@ public void setDestructiveIteration(boolean destructiveIteration) {
...
-                if (os != null) {
-                    os.close();
-                }
             }
 
             this.numRecordsInRam = 0;
             this.files.add(f);
-
         }
         catch (IOException e) {
```
pshapiro4broad: `catch` should be pulled up to previous line

src/main/java/htsjdk/samtools/util/SortingCollection.java
```diff
@@ -448,19 +463,43 @@ public void remove() {
...
+            log.info(String.format("Creating merging iterator from %d files", files.size()));
+            int suggestedBufferSize = checkMemoryAndAdjustBuffer(files.size(), Defaults.BUFFER_SIZE);
+            for (final Path f : files) {
+                final FileRecordIterator it = new FileRecordIterator(f, suggestedBufferSize);
                 if (it.hasNext()) {
                     this.queue.add(new PeekFileRecordIterator(it, n++));
-                }
-                else {
+                } else {
                     it.close();
```
pshapiro4broad: This can be skipped if the previous lines throw, using try-with-resources would fix that

src/main/java/htsjdk/samtools/util/StringUtil.java
pshapiro4broad: It's a shame that we have to reinvent this wheel. I see that apache commons io has a method for this but we don't import that library.

src/main/java/htsjdk/samtools/util/SortingCollection.java
jacarey: I suspect that at the time of creation this was not the case. I think it is ok to change the API over time as long as we have sufficient time with deprecated constructors. Do you think that the benefits of using an ArrayList are worth that change?

src/main/java/htsjdk/samtools/util/StringUtil.java
jacarey: yeah htsjdk tries to be dependency light.. :man_shrugging:

src/main/java/htsjdk/samtools/util/SortingCollection.java
```diff
@@ -66,23 +66,28 @@
     public interface Codec<T> extends Cloneable {
         /**
          * Where to write encoded output
+         *
          * @param os
```
pshapiro4broad: IMO it's just clutter to have blank `@param` entries. Either we should have some text here, or we should remove `@param`. Either way the javadoc will contain the parameter name.

jacarey:
I am currently testing with the 80x bam that spawned this ticket originally. I could certainly do a memory restricted test as well.

src/main/java/htsjdk/samtools/util/SortingCollection.java
```diff
@@ -164,7 +167,28 @@ public void add(final T rec) {
             throw new IllegalStateException("Cannot add after calling iterator()");
         }
         if (numRecordsInRam == maxRecordsInRam) {
+            // sample every 100 files written.
+            long startMem = 0;
+            boolean printRecordSizeSampling = files.size() % 100 == 0 && Log.getGlobalLogLevel() == Log.LogLevel.DEBUG;
```
magicDGS: Can all this profiling code be activated with a boolean provided on construction?  The default value should be true if debug level is on, and false otherwise. This code performs GC and computes memory usage even if the log is not printed (not debug), and I prefer to being able to switch off and to do not run unless the debug option is specified or I explicitly set it to true...

jacarey: It is set only if your log level is set to DEBUG which, if it is, I assume you are doing some debugging. I could add a flag too but in most cases you'd have to bubble that up in order for it to be useful at all since sorting collection is used under the hood in most places.

PR #1111, opened 2018-05-02
Profiled approximate file overhead instead of a flat % when calculating memory needed per file.

jacarey:
Also moved sampling boolean to constructor.
This should help in low memory situations to allow use to be a little bit more conservative as to when we turn off buffering.
### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/SortingCollection.java
+++ b/src/main/java/htsjdk/samtools/util/SortingCollection.java
@@ -131,6 +131,8 @@ public class SortingCollection<T> implements Iterable<T> {
 
     private final TempStreamFactory tempStreamFactory = new TempStreamFactory();
 
+    private final boolean sampleRecordSize;
+
     /**
      * Prepare to accumulate records to be sorted
      *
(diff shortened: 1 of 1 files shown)


src/main/java/htsjdk/samtools/util/SortingCollection.java
```diff
@@ -141,7 +143,8 @@
      * @param tmpDir          Where to write files of records that will not fit in RAM
      */
     private SortingCollection(final Class<T> componentType, final SortingCollection.Codec<T> codec,
-                              final Comparator<T> comparator, final int maxRecordsInRam, final Path... tmpDir) {
+                              final Comparator<T> comparator, final int maxRecordsInRam,
+                              final boolean sampleRecordSize, final Path... tmpDir) {
```
danxmoran: Can you add javadoc for `sampleRecordSize`?

src/main/java/htsjdk/samtools/util/SortingCollection.java
```diff
@@ -167,18 +171,17 @@ public void add(final T rec) {
             throw new IllegalStateException("Cannot add after calling iterator()");
         }
         if (numRecordsInRam == maxRecordsInRam) {
-            // sample every 100 files written.
+
             long startMem = 0;
-            boolean printRecordSizeSampling = files.size() % 100 == 0 && Log.getGlobalLogLevel() == Log.LogLevel.DEBUG;
```
danxmoran: Bike-shedding: why not keep the original name? The new one reads more like an `int` to me.

src/main/java/htsjdk/samtools/util/SortingCollection.java
jacarey: Sure. I added it to the public methods.. but I can add it here too no problem.

src/main/java/htsjdk/samtools/util/SortingCollection.java
jacarey: Agreed.

PR #1118, opened 2018-05-18
Extended SamReaderFactoryTest.testWriteAndReadFromPipe to be more com…

tfenne:
…prehensive.

### Description

Highlight that reading from a named pipe is broken with using `Path` instead of `File`.

### Checklist

- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
+++ b/src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
@@ -48,9 +48,10 @@ public class SamReaderFactoryTest extends HtsjdkTest {
         final String inputFile = "compressed.bam";
         final int[] inflateCalls = {0}; //Note: using an array is a HACK to fool the compiler
         class MyInflater extends Inflater {
-            MyInflater(boolean gzipCompatible){
+            MyInflater(boolean gzipCompatible) {
                 super(gzipCompatible);
             }
+
             @Override
             public int inflate(byte[] b, int off, int len) throws java.util.zip.DataFormatException {
                 inflateCalls[0]++;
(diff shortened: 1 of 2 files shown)


src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -538,55 +536,48 @@ public void testSamReaderFromMalformedSeekableStream() throws IOException {
...
+                    final SAMFileWriter writer = new SAMFileWriterFactory()
+                            .setCreateIndex(false).setCreateMd5File(false).makeBAMWriter(builder.getHeader(), true, fifo);
+
+                    int written = 0;
+                    for (SAMRecord read : builder) {
                         writer.addAlignment(read);
-                        written++;
+                        written += 1;
                     }
+                    writer.close();
```
pshapiro4broad: use try-with-resources to avoid the need for explicit close

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -538,55 +536,48 @@ public void testSamReaderFromMalformedSeekableStream() throws IOException {
 
     @Test(singleThreaded = true, groups="unix")
     public void testWriteAndReadFromPipe() throws IOException, InterruptedException, ExecutionException, TimeoutException {
+        final SAMRecordSetBuilder builder = new SAMRecordSetBuilder(false, SortOrder.unsorted);
+        for (int i=1; i<= 5000; ++i) {
```
pshapiro4broad: format code to have spaces around operators, e.g. `int i = 1; i < = 5000`

or use `IntStream.range(1, 5000).forEach(i -> builder.addPair("q" + i, 0, i, i));`

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -538,55 +536,48 @@ public void testSamReaderFromMalformedSeekableStream() throws IOException {
...
+
+                final SamReader in = SamReaderFactory.make().open(res);
+                int count = 0;
+                for (SAMRecord rec : in) {
+                    Assert.assertEquals(rec.getReadName(), "q" + rec.getAlignmentStart());
+                    count += 1;
                 }
-            } catch (final Exception e) {
-                Assert.fail("Failed during writing to pipe", e);
+                in.close();
```
pshapiro4broad: try-with-resources here as well

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -538,55 +536,48 @@ public void testSamReaderFromMalformedSeekableStream() throws IOException {
...
+
+                Assert.assertEquals(count, builder.size());
+                Assert.assertEquals(count, future.get().intValue());
+            } finally {
+                if (executor != null) executor.shutdownNow();
```
pshapiro4broad: `if` body should have `{` `}` around it and be formatted across multiple lines

yfarjoun:
We had a long discussion about this issue in the last htsjdk review Party. 

The proposal we came up with is to modify htsjdk.samtools.SamInputResource#of(java.nio.file.Path) so that if the input is a pipe (!Files.isRegularFile(path)) we will wrap it in a FileInputStream and call htsjdk.samtools.SamInputResource#of(java.io.InputStream).

While this doesn't work in the test as written, it is because of a race condition. 

@tfenne can I push to this branch?

yfarjoun:
superseded by #1124

Issue #1112, opened 2018-05-03
Improvements to support Hadoop-BAM/Squark


tomwhite:
This describes some of the improvements, changes and additions to htsjdk to better support Hadoop-BAM and its proposed new incarnation codenamed [Squark](https://github.com/tomwhite/squark).

Individual issues or PRs can be created later to implement the changes as needed.

Small changes
- Code to write a BAM header to a stream. Currently resides in htsjdk's BAMFileWriter#writeHeader, which is protected (and the class is package private). Needed by [BamSink](https://github.com/tomwhite/squark/blob/htsjdk-comments/src/main/java/com/tom_e_white/squark/impl/formats/bam/BamSink.java#L74) in Squark.
- More block compressed file pointer utils. BlockCompressedFilePointerUtil#makeFilePointer should be public. It would also be useful to have an overloaded version that doesn't have an offset (i.e. is 0).  Squark has [BgzfVirtualFilePointerUtil](https://github.com/tomwhite/squark/blob/htsjdk-comments/src/main/java/com/tom_e_white/squark/impl/formats/bgzf/BgzfVirtualFilePointerUtil.java)
- Expose CRAMIntervalIterator. Currently private, it's needed in [Squark](https://github.com/tomwhite/squark/blob/htsjdk-comments/src/main/java/com/tom_e_white/squark/impl/formats/cram/CramSource.java#L158) to get reads overlapping intervals, much like BAMFileReader#createIndexIterator.

Improvements
- Seeks within SeekableBufferedStream's buffer should not create a new buffer. See Squark's [ExtSeekableBufferedStream](https://github.com/tomwhite/squark/blob/htsjdk-comments/src/main/java/htsjdk/samtools/ExtSeekableBufferedStream.java) for the changes. 
- An optimized version of CramContainerIterator that only reads the header for each container. See Squark's [CramContainerHeaderIterator](https://github.com/tomwhite/squark/blob/htsjdk-comments/src/main/java/htsjdk/samtools/CramContainerHeaderIterator.java)
- A way to read a VCFHeader from a stream without knowing if the file is VCF or BCF, or compressed or not. Implementation in Hadoop-BAM's [VCFHeaderReader](https://github.com/HadoopGenomics/Hadoop-BAM/blob/master/src/main/java/org/seqdoop/hadoop_bam/util/VCFHeaderReader.java).
[...]

PR #1120, opened 2018-05-22
Public method for CRAMIntervalIterator for overlapping intervals

tomwhite:
### Description

Expose a method to create a CRAMIntervalIterator to get reads overlapping intervals, much like BAMFileReader#createIndexIterator. See #1112 

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/CRAMFileReader.java
+++ b/src/main/java/htsjdk/samtools/CRAMFileReader.java
@@ -466,6 +466,32 @@ public class CRAMFileReader extends SamReader.ReaderImplementation implements Sa
             iterator.setFileSource(enabled ? reader : null);
     }
 
+    /**
+     * Prepare to iterate through SAMRecords that match the intersection of the given intervals and chunk boundaries.
+     * @param intervals the intervals to restrict reads to
+     * @param contained if <code>true</code>, return records that are strictly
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -466,6 +466,32 @@ void enableFileSource(final SamReader reader, final boolean enabled) {
             iterator.setFileSource(enabled ? reader : null);
     }
 
+    /**
+     * Prepare to iterate through SAMRecords that match the given intervals.
```
cmnbroad: I think we should slightly narrow the description of what this method does. Since there is no validation that the provided chunks actually match (fully span) the territory covered by the intervals, it possible that the iterator doesn't return the full set of records matching the query. The records are constrained first by the chunks, then further constrained by the intervals.

tomwhite: Good point. I've revised the wording to say "Prepare to iterate through SAMRecords that match the intersection of the given intervals and chunk boundaries."

tomwhite:
Added a new test.

lbergelson:
@tomwhite Thank you.

PR #1121, opened 2018-05-22
Seeks within SeekableBufferedStream's buffer do not need to create a new buffer

tomwhite:
### Description

This is an optimization to avoid unnecessary seeks (which can be expensive when reading from cloud stores, or Hadoop). See #1112

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/seekablestream/SeekableBufferedStream.java
+++ b/src/main/java/htsjdk/samtools/seekablestream/SeekableBufferedStream.java
@@ -47,14 +47,18 @@ public class SeekableBufferedStream extends SeekableStream {
             else return this.buf.length - this.pos;
         }
 
-        /** Returns the position in the buffer. */
-        int getPositionInBuffer() {
-            return pos;
+        /** Return true if the position can be changed by the given delta and remain in the buffer. */
+        boolean canChangePos(long delta) {
+            return this.pos + delta >= 0 && delta < getBytesInBufferAvailable();
         }
 
         /** Changes the position in the buffer by a given delta. */
         void changePos(int delta) {
-            this.pos += delta;
+            int newPos = this.pos + delta;
+            if (newPos < 0 || newPos > this.buf.length) {
+                throw new IllegalArgumentException("New position not in buffer pos=" + this.pos + ", delta=" + delta);
+            }
+            this.pos = newPos;
         }
     }
 
@@ -90,7 +94,7 @@ public class SeekableBufferedStream extends SeekableStream {
             return retval;
         } else {
             final long position = this.position + skipLength;
-            seek(position);
+            seekInternal(position);
             return skipLength;
         }
     }
@@ -100,13 +104,20 @@ public class SeekableBufferedStream extends SeekableStream {
         if (this.position == position) {
             return;
         }
-        if ((position > this.position && (position - this.position < this.bufferedStream.getBytesInBufferAvailable()))
-                || (position < this.position && (this.position - position < this.bufferedStream.getPositionInBuffer()))) {
-            this.bufferedStream.changePos((int) (position - this.position));
+        // check if the seek is within the buffer
+        long delta = position - this.position;
+        if (this.bufferedStream.canChangePos(delta)) {
+            // casting to an int is safe since the buffer is less than the size of an int
+            this.bufferedStream.changePos((int) delta);
+            this.position = position;
         } else {
-            wrappedStream.seek(position);
-            bufferedStream = new ExtBufferedInputStream(wrappedStream, bufferSize);
+            seekInternal(position);
         }
+    }
+
+    private void seekInternal(final long position) throws IOException {
+        wrappedStream.seek(position);
+        bufferedStream = new ExtBufferedInputStream(wrappedStream, bufferSize);
         this.position = position;
     }
 
--- a/src/main/java/htsjdk/samtools/seekablestream/SeekableBufferedStream.java
+++ b/src/main/java/htsjdk/samtools/seekablestream/SeekableBufferedStream.java
@@ -46,6 +46,16 @@ public class SeekableBufferedStream extends SeekableStream {
             if (this.count == this.pos) return 0; // documented test for "is buffer empty"
             else return this.buf.length - this.pos;
         }
+
+        /** Returns the position in the buffer. */
+        int getPositionInBuffer() {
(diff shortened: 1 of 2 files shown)


src/test/java/htsjdk/samtools/seekablestream/SeekableBufferedStreamTest.java
```diff
@@ -123,6 +124,43 @@ public void testSkip() throws IOException {
...
+        int bufferSize = 20000;
+        int startPosition = 250000;
+        int length = 5000;
+
+        final int[] RELATIVE_SEEK_OFFSET = new int[]{-bufferSize*2, -bufferSize, -bufferSize/2, -length, -length/2, 0,
+                length/2, length, bufferSize/2, bufferSize, bufferSize*2};
+
+        for (final int seekOffset : RELATIVE_SEEK_OFFSET) {
+            byte[] buffer1 = new byte[length];
+            SeekableStream unBufferedStream = new SeekableFileStream(BAM_FILE);
```
lbergelson: You should pull the stream creation into a try-with-resources block so they get closed.

src/test/java/htsjdk/samtools/seekablestream/SeekableBufferedStreamTest.java
```diff
@@ -123,6 +124,43 @@ public void testSkip() throws IOException {
         }
     }
 
+    @Test
+    public void testSeek() throws IOException {
+        int bufferSize = 20000;
```
lbergelson: I'd make these final.

src/test/java/htsjdk/samtools/seekablestream/SeekableBufferedStreamTest.java
```diff
@@ -123,6 +124,43 @@ public void testSkip() throws IOException {
...
+        int startPosition = 250000;
+        int length = 5000;
+
+        final int[] RELATIVE_SEEK_OFFSET = new int[]{-bufferSize*2, -bufferSize, -bufferSize/2, -length, -length/2, 0,
```
lbergelson: it's weird that this is all caps generally style is to only use caps for static constants, but it does match the rest of the tests, so I guess keep it for consistency?

src/test/java/htsjdk/samtools/seekablestream/SeekableBufferedStreamTest.java
```diff
@@ -123,6 +124,43 @@ public void testSkip() throws IOException {
...
+                length/2, length, bufferSize/2, bufferSize, bufferSize*2};
+
+        for (final int seekOffset : RELATIVE_SEEK_OFFSET) {
+            byte[] buffer1 = new byte[length];
+            SeekableStream unBufferedStream = new SeekableFileStream(BAM_FILE);
+            unBufferedStream.seek(startPosition);
+            int bytesRead = unBufferedStream.read(buffer1, 0, length);
+            Assert.assertEquals(length, bytesRead);
```
lbergelson: It would be nice if we were explicitly checking that we were doing the cheap jumps when possible instead of expensive seeks.  It seems like it could be done by instrumenting the SeekableStream before we wrap it. Maybe not worth it though.

src/main/java/htsjdk/samtools/seekablestream/SeekableBufferedStream.java
```diff
@@ -87,9 +97,17 @@ public long skip(final long skipLength) throws IOException {
 
     @Override
     public void seek(final long position) throws IOException {
+        if (this.position == position) {
+            return;
+        }
+        if ((position > this.position && (position - this.position < this.bufferedStream.getBytesInBufferAvailable()))
```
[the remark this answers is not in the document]
tomwhite: Good suggestions. Changed as necessary.

src/main/java/htsjdk/samtools/seekablestream/SeekableBufferedStream.java
```diff
@@ -46,6 +46,16 @@ int getBytesInBufferAvailable() {
...
             else return this.buf.length - this.pos;
         }
+
+        /** Returns the position in the buffer. */
+        int getPositionInBuffer() {
+            return pos;
+        }
+
+        /** Changes the position in the buffer by a given delta. */
+        void changePos(int delta) {
```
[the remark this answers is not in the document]
tomwhite: Done.

src/test/java/htsjdk/samtools/seekablestream/SeekableBufferedStreamTest.java
tomwhite: Done.

src/test/java/htsjdk/samtools/seekablestream/SeekableBufferedStreamTest.java
tomwhite: Done.

src/test/java/htsjdk/samtools/seekablestream/SeekableBufferedStreamTest.java
tomwhite: Yes, I did this for consistency.

src/test/java/htsjdk/samtools/seekablestream/SeekableBufferedStreamTest.java
tomwhite: This turned out to be easy to implement by checking if the internal buffer's object identity changed.

tomwhite:
@lbergelson thanks for the review. I also made `skip()` call a `skipInternal()` method that has the old logic (with no buffer awareness).

tomwhite:
The code coverage changes look unrelated.

src/main/java/htsjdk/samtools/seekablestream/SeekableBufferedStream.java
```diff
@@ -80,16 +94,31 @@ public long skip(final long skipLength) throws IOException {
...
 
     @Override
     public void seek(final long position) throws IOException {
-        this.position = position;
+        if (this.position == position) {
+            return;
+        }
+        // check if the seek is within the buffer
+        long delta = position - this.position;
+        if (this.bufferedStream.canChangePos(delta)) {
```
lbergelson: This is so much cleaner than before.

PR #1122, opened 2018-05-22
Add VCFHeaderReader

tomwhite:
### Description

Add a utility class (VCFHeaderReader) to read a VCFHeader from a stream without knowing if the file is VCF or BCF, or compressed or not. This is used a lot in Hadoop-BAM and GATK, but has general utility (it's not tied to Hadoop or Spark, for example), so htsjdk would be an appropriate home for it. See #1112

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/variant/utils/VCFHeaderReader.java
+++ b/src/main/java/htsjdk/variant/utils/VCFHeaderReader.java
@@ -1,0 +1,56 @@
+package htsjdk.variant.utils;
+
+import htsjdk.samtools.SamStreams;
+import htsjdk.samtools.seekablestream.SeekableStream;
+import htsjdk.samtools.util.BlockCompressedInputStream;
+import htsjdk.tribble.FeatureCodecHeader;
+import htsjdk.tribble.TribbleException;
+import htsjdk.tribble.readers.AsciiLineReader;
+import htsjdk.tribble.readers.AsciiLineReaderIterator;
+import htsjdk.tribble.readers.PositionalBufferedStream;
+import htsjdk.variant.bcf2.BCF2Codec;
+import htsjdk.variant.vcf.VCFCodec;
+import htsjdk.variant.vcf.VCFHeader;
+
+import java.io.BufferedInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * Utility class to read a VCF header without being told beforehand whether the input is VCF or BCF.
+ */
+public final class VCFHeaderReader {
+    /**
+     * Read a VCF header from a stream assuming first that it is a VCF file (possibly gzip or block compressed),
+     * and falling back to trying BCF if reading VCF fails. After successfully reading a header the stream is
+     * positioned immediately after the header, otherwise, if an exception is thrown, the state of the stream is
+     * undefined.
+     *
+     * @param in the stream to read the header from
+     * @return the VCF header read from the stream
+     * @throws IOException if no VCF header is found
+     */
+    public static VCFHeader readHeaderFrom(final SeekableStream in) throws IOException {
+        FeatureCodecHeader headerCodec;
+        final long initialPos = in.position();
+        try {
+            BufferedInputStream bis = new BufferedInputStream(in);
+            // despite the name, SamStreams.isGzippedSAMFile looks for any gzipped stream (including block compressed)
+            InputStream is = SamStreams.isGzippedSAMFile(bis) ? new GZIPInputStream(bis) : bis;
+            headerCodec = new VCFCodec().readHeader(new AsciiLineReaderIterator(AsciiLineReader.from(is)));
+        } catch (TribbleException e) {
+            in.seek(initialPos);
+            InputStream bin = new BufferedInputStream(in);
+            if (BlockCompressedInputStream.isValidFile(bin)) {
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/variant/utils/VCFHeaderReader.java
```diff
@@ -0,0 +1,54 @@
...
+     * @return the VCF header read from the stream
+     * @throws IOException if no VCF header is found
+     */
+    public static VCFHeader readHeaderFrom(final SeekableStream in) throws IOException {
+        FeatureCodecHeader headerCodec;
+        final long initialPos = in.position();
+        try {
+            BufferedInputStream bis = new BufferedInputStream(in);
+            // despite the name, SamStreams.isGzippedSAMFile looks for any gzipped stream
+            InputStream is = SamStreams.isGzippedSAMFile(bis) ? new GZIPInputStream(bis) : bis;
```
lbergelson: Does isGzippedSAMFile return true for a block compressed file?  It must or vcf.bgz wouldn't work here?  I think you should add a comment about that case here.

src/main/java/htsjdk/variant/utils/VCFHeaderReader.java
```diff
@@ -0,0 +1,54 @@
...
+import java.io.InputStream;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * Utility class to read a VCF header without being told beforehand whether the input is VCF or BCF.
+ */
+public final class VCFHeaderReader {
+    /**
+     * Read a VCF header from a stream assuming first that it is a VCF file (possibly gzip or block compressed),
+     * and falling back to trying BCF if reading VCF fails.
```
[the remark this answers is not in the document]
tomwhite: Done

src/main/java/htsjdk/variant/utils/VCFHeaderReader.java
tomwhite: Yes it does. Updated comment.

src/main/java/htsjdk/variant/utils/VCFHeaderReader.java
```diff
@@ -0,0 +1,56 @@
...
+        } catch (TribbleException e) {
+            in.seek(initialPos);
+            InputStream bin = new BufferedInputStream(in);
+            if (BlockCompressedInputStream.isValidFile(bin)) {
+                bin = new BlockCompressedInputStream(bin);
+            }
+            try {
+                headerCodec = new BCF2Codec().readHeader(new PositionalBufferedStream(bin));
+            } catch (TribbleException e2) {
+                throw new IOException("No VCF header found", e2);
```
cmnbroad: One suggestion - the codecs throw TribbleException for any header parsing issue, so in some cases, the first exception (on line 42) represents the real issue, but the second one will be what the user sees. We could deterministically figure out the stream type and delegate based on that, otherwise this code will report misleading errors, i.e. something like "No VCF header found - BCF magic header info not found" when the real issue is the a malformed header in a VCF stream. At one time there was a PR that had stream sniffing code in it, but I can't find it. There is code in the BCF codec in readHeader to determine if a stream is a BCF stream, though it would have to be exposed.

tomwhite: @cmnbroad thanks for the suggestion. I've reworked the method so that it detects VCF vs BCF from the magic number, then tries to parse the header.

src/test/java/htsjdk/variant/utils/VCFHeaderReaderTest.java
```diff
@@ -0,0 +1,47 @@
...
+    public void testReadHeaderFromThrowsIOExceptionForInvalidFile() throws IOException {
+        try {
+            VCFHeaderReader.readHeaderFrom(new SeekableFileStream(new File("src/test/resources/htsjdk/samtools/empty.bam")));
+        } catch (FileNotFoundException e) {
+            Assert.fail("Should fail with IOException, not FileNotFoundException");
+        }
+    }
+
+    @Test(expectedExceptions = FileNotFoundException.class)
+    public void testReadHeaderFromThrowsFileNotFoundExceptionForNonExistentFile() throws IOException {
```
lbergelson: This test passes even if you remove the call to readHeaderFrom, it's failing during stream construction

src/main/java/htsjdk/variant/utils/VCFHeaderReader.java
```diff
@@ -0,0 +1,56 @@
...
+    public static VCFHeader readHeaderFrom(final SeekableStream in) throws IOException {
+        final long initialPos = in.position();
+        byte[] magicBytes = InputStreamUtils.readFully(bufferAndDecompressIfNecessary(in), BCFVersion.MAGIC_HEADER_START.length);
+        in.seek(initialPos);
+        if (magicBytes[0] == '#') { // VCF
+            return readHeaderFrom(in, new VCFCodec());
+        } else if (Arrays.equals(magicBytes, BCFVersion.MAGIC_HEADER_START)) {
+            return readHeaderFrom(in, new BCF2Codec());
+        }
+        throw new IOException("No VCF header found");
```
lbergelson: change this to throw TribbleException.InvalidHeader so that it matches what will happens if the readHeadFrom method fails internally.

lbergelson:
@tomwhite Merged this here #1148

PR #1123, opened 2018-05-22
Use ReferenceSequenceFileFactory with streams

tomwhite:
### Description

Add a method to open a reference sequence by passing an input stream to a FASTA (and to its index). This would allow reading from Hadoop filesystems without having to use the file NIO library. See #1112

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
+++ b/src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
@@ -32,6 +32,7 @@ import htsjdk.samtools.util.BufferedLineReader;
 import htsjdk.samtools.util.IOUtil;
 
 import java.io.File;
+import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 
@@ -41,6 +42,7 @@ import java.nio.file.Path;
  */
 abstract class AbstractFastaSequenceFile implements ReferenceSequenceFile {
     private final Path path;
+    private final String source;
     protected SAMSequenceDictionary sequenceDictionary;
 
     /**
@@ -57,25 +59,30 @@ abstract class AbstractFastaSequenceFile implements ReferenceSequenceFile {
      */
     AbstractFastaSequenceFile(final Path path) {
         this.path = path;
+        this.source = path == null ? "unknown" : path.toAbsolutePath().toString();
         final Path dictionary = findSequenceDictionary(path);
 
         if (dictionary != null) {
             IOUtil.assertFileIsReadable(dictionary);
-
-            try {
-                final SAMTextHeaderCodec codec = new SAMTextHeaderCodec();
-                final BufferedLineReader reader = new BufferedLineReader(Files.newInputStream(dictionary));
-                final SAMFileHeader header = codec.decode(reader,
-                        dictionary.toString());
-                if (header.getSequenceDictionary() != null && !header.getSequenceDictionary().isEmpty()) {
-                    this.sequenceDictionary = header.getSequenceDictionary();
-                }
-                reader.close();
+            try (InputStream dictionaryIn = Files.newInputStream(dictionary)) {
+                this.sequenceDictionary = ReferenceSequenceFileFactory.loadDictionary(dictionaryIn);
             }
             catch (Exception e) {
                 throw new SAMException("Could not open sequence dictionary file: " + dictionary, e);
             }
         }
+    }
+
+    /**
+     * Constructs an {@link AbstractFastaSequenceFile} with an optional sequence dictionary.
+     * @param path Fasta file to read.  Also acts as a prefix for supporting files.
+     * @param source Named source used for error messages.
+     * @param sequenceDictionary The sequence dictionary, or null if there isn't one.
+     */
+    AbstractFastaSequenceFile(final Path path, final String source, final SAMSequenceDictionary sequenceDictionary) {
+        this.path = path;
+        this.source = source;
+        this.sequenceDictionary = sequenceDictionary;
     }
 
     protected static File findSequenceDictionary(final File file) {
@@ -111,6 +118,11 @@ abstract class AbstractFastaSequenceFile implements ReferenceSequenceFile {
         return path;
     }
 
+    /** Returns the named source of the reference file. */
+    protected String getSource() {
+        return source;
+    }
+
     /**
      * Returns the list of sequence records associated with the reference sequence if found
      * otherwise null.
@@ -122,12 +134,15 @@ abstract class AbstractFastaSequenceFile implements ReferenceSequenceFile {
 
     /** Returns the full path to the reference file. */
     protected String getAbsolutePath() {
+        if (path == null) {
+            return null;
+        }
         return path.toAbsolutePath().toString();
     }
 
-    /** Returns the full path to the reference file. */
+    /** Returns the full path to the reference file, or the source if no path was specified. */
     public String toString() {
-        return getAbsolutePath();
+        return source;
     }
 
     /** default implementation -- override if index is supported */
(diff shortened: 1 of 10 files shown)


src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
```diff
@@ -127,7 +127,7 @@ protected String getAbsolutePath() {
 
     /** Returns the full path to the reference file. */
     public String toString() {
-        return getAbsolutePath();
+        return path == null ? super.toString() : getAbsolutePath();
```
lbergelson: I think adding a field that can be specified at construction time that gives this a name would be useful.  Usually the people creating the input streams will have some named source, it's just not easy to convert it to a `Path`, but the name is still useful for error messages.

src/main/java/htsjdk/samtools/reference/FastaSequenceFile.java
```diff
@@ -59,6 +60,16 @@ public FastaSequenceFile(final Path path, final boolean truncateNamesAtWhitespac
         this.in = new FastLineReader(IOUtil.openFileForReading(path));
     }
 
+    /**
+     * Constructs a FastaSequenceFile that reads from the specified input stream (which must not be compressed, i.e.
+     * the caller is responsible for decompressing the stream).
+     */
+    public FastaSequenceFile(final InputStream inputStream, final boolean truncateNamesAtWhitespace) {
+        super((Path) null); // no path
```
lbergelson: We need some way of loading the dict file when the path is null, lots of things in gatk want to have a sequence dictionary from the reference.  I think the new constructors should either take _another_  input stream where the user can pass an open stream to the .dict file, or accept a `SAMSequenceDictionary` directly and rely on them loading it separately. (Is there an exposed way to read a dict file anywhere?)

src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
```diff
@@ -127,7 +127,7 @@ protected String getAbsolutePath() {
```
lbergelson: Get absolute path needs to change because it will crash with NPE if there's no path.  It should crash with a more useful exception instead.

src/main/java/htsjdk/samtools/reference/FastaSequenceFile.java
```diff
@@ -59,6 +60,16 @@ public FastaSequenceFile(final Path path, final boolean truncateNamesAtWhitespac
         this.in = new FastLineReader(IOUtil.openFileForReading(path));
     }
 
+    /**
+     * Constructs a FastaSequenceFile that reads from the specified input stream (which must not be compressed, i.e.
+     * the caller is responsible for decompressing the stream).
+     */
+    public FastaSequenceFile(final InputStream inputStream, final boolean truncateNamesAtWhitespace) {
```
lbergelson: `reset()` is going to fail in a confusing way.  Since this isn't seekable it's not possible to modify it to work with this.  We should either add a better message when calling reset on a stream created this way, or require this to be seekable and modify reset.

src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
lbergelson: There are a lot of places that use `getAbsolutePath` as part of error messages.  Those should be changed to use the new name field that I mention below, or have explicit checks for the availability of path first.  It's not good to crash when generating an error message...

src/main/java/htsjdk/samtools/reference/IndexedFastaSequenceFile.java
```diff
@@ -111,6 +114,18 @@ public IndexedFastaSequenceFile(final Path path) throws FileNotFoundException {
         this(path, new FastaSequenceIndex((findRequiredFastaIndexFile(path))));
     }
 
+    /**
+     * Initialise the given indexed fasta sequence file stream.
+     * @param in The input stream to read the fasta file from.
+     * @param indexIn The input stream to read the fasta file index from.
+     */
+    public IndexedFastaSequenceFile(final SeekableStream in, final InputStream indexIn) {
```
lbergelson: Would it make more sense to have this take SeekableStream and a preconstructed `FastSequenceIndex` instead of two streams?

src/main/java/htsjdk/samtools/reference/FastaSequenceFile.java
tomwhite: I've added a dictionary argument, and added a `loadDictionary(InputStream)` convenience method to `ReferenceSequenceFileFactory`.

src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
tomwhite: Fixed to use source, which defaults to the absolute path if a path is supplied.

src/main/java/htsjdk/samtools/reference/FastaSequenceFile.java
tomwhite: Changed to use `SeekableStream` so `reset()` can be implemented properly.

src/main/java/htsjdk/samtools/reference/IndexedFastaSequenceFile.java
tomwhite: Yes. Done.

tomwhite:
@lbergelson hopefully this is better.

src/test/java/htsjdk/samtools/reference/IndexedFastaSequenceFileTest.java
```diff
@@ -41,6 +43,7 @@
 public class IndexedFastaSequenceFileTest extends HtsjdkTest {
     private static File TEST_DATA_DIR = new File("src/test/resources/htsjdk/samtools/reference");
```
lbergelson: can you make these statics final while you're here?

magicDGS:
#1014 is already merged @tomwhite - rebase should unblock this PR, which might be amazing too for the new "squark" codebase!

src/test/java/htsjdk/samtools/reference/IndexedFastaSequenceFileTest.java
tomwhite: Done

tomwhite:
@lbergelson @magicDGS I've rebased on #1014. I haven't added stream support for block compressed FASTA files - that could be added later.

magicDGS:
@tomwhite - maybe open a ticket for the bgz FASTA with stream to track it. I can take the ticket whenever I have time, because it looks like an easy implementation.

tomwhite:
@magicDGS - I opened #1133. Thanks!

src/main/java/htsjdk/samtools/reference/FastaSequenceFile.java
```diff
@@ -56,9 +60,21 @@ public FastaSequenceFile(final File file, final boolean truncateNamesAtWhitespac
...
         this.truncateNamesAtWhitespace = truncateNamesAtWhitespace;
+        this.seekableStream = null;
         this.in = new FastLineReader(IOUtil.openFileForReading(path));
     }
 
+    /**
+     * Constructs a FastaSequenceFile that reads from the specified stream (which must not be compressed, i.e.
+     * the caller is responsible for decompressing the stream).
+     */
+    public FastaSequenceFile(String source, final SeekableStream seekableStream, SAMSequenceDictionary dictionary, final boolean truncateNamesAtWhitespace) {
```
lbergelson: @tomwhite This constructor and the following reset method are untested and have non-trivial interactions with the rest of the class.  Could you add tests that exercise them?

src/main/java/htsjdk/samtools/reference/FastaSequenceFile.java
```diff
@@ -56,9 +60,21 @@ public FastaSequenceFile(final File file, final boolean truncateNamesAtWhitespac
     public FastaSequenceFile(final Path path, final boolean truncateNamesAtWhitespace) {
         super(path);
         this.truncateNamesAtWhitespace = truncateNamesAtWhitespace;
+        this.seekableStream = null;
```
magicDGS: Maybe a way to clean up this code is to open the `java.nio.Path` as a `SeekableStream` here and initialize the field.

src/main/java/htsjdk/samtools/reference/FastaSequenceFile.java
```diff
@@ -89,8 +105,16 @@ public ReferenceSequence nextSequence() {
     public void reset() {
         this.sequenceIndex = -1;
         this.in.close();
-        this.in = new FastLineReader(IOUtil.openFileForReading(getPath()));
-
+        if (getPath() != null) {
+            this.in = new FastLineReader(IOUtil.openFileForReading(getPath()));
```
magicDGS: This if clause can be removed if the `Path` in open on construction as a `SeekableStream`

src/main/java/htsjdk/samtools/reference/FastaSequenceFile.java
tomwhite: Done

src/main/java/htsjdk/samtools/reference/FastaSequenceFile.java
tomwhite: This would probably work, but I was being cautious and leaving the old implementation unchanged.

lbergelson:
@tomwhite Thank you Tom.

PR #1124, opened 2018-05-22
Fix regression regarding pipeing

yfarjoun:
Based on @tfenne 's branch #1118 

includes:
- Test that piping works for both File and Path (using mkfifo and a second thread)
- Fixed .of(Path) to call .of(File) when the input is not a regular file.
- Lots of whitespace changes (sorry)
- Closed readers (in the tests) using try-with-resources

### DIFF
--- a/src/main/java/htsjdk/samtools/SamInputResource.java
+++ b/src/main/java/htsjdk/samtools/SamInputResource.java
@@ -96,13 +96,7 @@ public class SamInputResource {
         if (Files.isRegularFile(path) &&  Files.exists(path)) {
             return new SamInputResource(new PathInputResource(path));
         } else {
-            try {
-                return of(new FileInputStream(path.toFile()));
-            } catch (FileNotFoundException e) {
-                // we just checked for existence...so I'm not sure how this could happen!
-                e.printStackTrace();
-                return null;
-            }
+                return of(path.toFile());
         }
     }
 
--- a/src/main/java/htsjdk/samtools/SamInputResource.java
+++ b/src/main/java/htsjdk/samtools/SamInputResource.java
@@ -38,6 +38,7 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.channels.SeekableByteChannel;
 import java.nio.file.FileSystemNotFoundException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.function.Function;
 import java.util.function.Supplier;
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/SamInputResource.java
```diff
@@ -91,7 +92,12 @@ public static SamInputResource of(final File file) {
 
     /** Creates a {@link SamInputResource} reading from the provided resource, with no index. */
     public static SamInputResource of(final Path path) {
-        return new SamInputResource(new PathInputResource(path));
+
+        if (Files.isRegularFile(path) &&  Files.exists(path)) {
+            return new SamInputResource(new PathInputResource(path));
+        } else {
+                return of(path.toFile());
```
tfenne: Indentation is funny

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -512,81 +518,81 @@ public void testSamReaderFromSeekableStream() throws IOException {
...
-        countRecords(reader);
+        try (final SamReader reader = factory.open(
+                SamInputResource.of(new SeekableFileStream(samFile)))) {
+            countRecords(reader);
+        }
     }
 
-    @Test(singleThreaded = true, groups="unix")
-    public void testWriteAndReadFromPipe() throws IOException, InterruptedException, ExecutionException, TimeoutException {
+    @DataProvider()
```
tfenne: I think turning this into a data provider is _less_ clear that just having `for x in Arrays.asList(true, false)` in the test.

src/main/java/htsjdk/samtools/SamInputResource.java
```diff
@@ -91,7 +92,12 @@ public static SamInputResource of(final File file) {
 
     /** Creates a {@link SamInputResource} reading from the provided resource, with no index. */
     public static SamInputResource of(final Path path) {
-        return new SamInputResource(new PathInputResource(path));
+
+        if (Files.isRegularFile(path) &&  Files.exists(path)) {
+            return new SamInputResource(new PathInputResource(path));
+        } else {
```
lbergelson: This needs a comment linking back to the discussion that spawned it.  Otherwise it will cause much confusion in the future.

src/main/java/htsjdk/samtools/SamInputResource.java
yfarjoun: I had originally implemented it as 
```
return of(new FileInputStream(path))
```
But that requires a try-catch block to guard against non-existant paths.
perhaps in the presence of the `Files.exists(path)` clause this is a non-issue and I can swallow the exception and return a null (which should never ever happen)?

happy to change it or not...as people prefer.

yfarjoun:
I added text, and fixed the indents. merging.

src/main/java/htsjdk/samtools/SamInputResource.java
```diff
@@ -91,7 +92,12 @@ public static SamInputResource of(final File file) {
 
     /** Creates a {@link SamInputResource} reading from the provided resource, with no index. */
     public static SamInputResource of(final Path path) {
-        return new SamInputResource(new PathInputResource(path));
+
+        if (Files.isRegularFile(path) &&  Files.exists(path)) {
```
droazen: @yfarjoun Could you open a separate PR to remove the `Files.exists()` check here? This operation is expensive over certain NIO filesystems (eg., takes on the order of seconds over GCS), so forcing it at reader creation time could potentially add hours to the wall-clock time of a task that needed to access many bams over GCS. 

It's also not particularly logical to go to the else clause here if the file doesn't exist -- if it's a non-existent GCS file, `PathInputResource` is better equipped to produce a more relevant error message.

PR #1125, opened 2018-05-22
Fix reading a CRAM with an index from a non-File Path.

cmnbroad:
Fix for GATK issue https://github.com/broadinstitute/gatk/issues/3669. Testing the change requires opening a CRAM through a Path that is backed by something that isn't a File. The newly added test, which simulates this condition, fails without this change. Is there a better way to force execution of this code path without adding any additional dependencies ?

### DIFF
--- a/src/main/java/htsjdk/samtools/SamReaderFactory.java
+++ b/src/main/java/htsjdk/samtools/SamReaderFactory.java
@@ -406,7 +406,17 @@ public abstract class SamReaderFactory {
                             referenceSource = ReferenceSource.getDefaultCRAMReferenceSource();
                         }
                         if (sourceFile == null || !sourceFile.isFile()) {
-                            primitiveSamReader = new CRAMFileReader(bufferedStream, indexFile, referenceSource, validationStringency);
+                            final SeekableStream indexSeekableStream =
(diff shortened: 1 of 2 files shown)


lbergelson:
@cmnbroad We typically write a test using jimfs in the case were we want a `Path` that is definitely not a File.  It's already a test dependency.  There are a bunch of example tests already in `CRAMComplianceTest` `IOUtilTest` and `SAMFileWriterTest`.  It's an in memory file system, you basically just create a file system, get a Path into the newly created filesystem, and then copy in files and use them like normal.  Just be sure to close the filesystem when the test finishes or it will eat memory until the process ends.

lbergelson:
@cmnbroad I'm terrified of editing this method.  It's such a mess.  I'm going to run the gatk tests against this change and see if anything falls over because I don't trust my ability think through all the cases on this thing.

cmnbroad:
@lbergelson Yeah, I tried to make the smallest possible change I could, but its definitely hard to modify this method.

lbergelson:
@cmnbroad Thank you.

lbergelson:
this will fix broadinstitute/gatk#3669 when updating to htsjdk 2.16.0

PR #1129, opened 2018-05-24
Add an optimised version of CramContainerIterator

tomwhite:
### Description

An optimized version of CramContainerIterator that only reads the header for each container. The motivation for this change is to make it possible to find CRAM container offsets without having to decompress the blocks in the container. See #1112

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


lbergelson:
@cmnbroad Could you review this one when you get a chance?
@tomwhite The tests are failing because `CramContainerHeaderIteratorTest` doesn't extend `HtsJdkTest`.   If it doesn't extend that then it may not be picked up by the test runner, so there is a test that checks for test files that don't extend it...

tomwhite:
Made the test extend `HtsJdkTest`.

src/test/java/htsjdk/samtools/CramContainerHeaderIteratorTest.java
```diff
@@ -0,0 +1,56 @@
...
+            Assert.assertEquals(headerOnlyContainer.containerByteSize, fullContainer.containerByteSize);
+            Assert.assertEquals(headerOnlyContainer.sequenceId, fullContainer.sequenceId);
+            Assert.assertEquals(headerOnlyContainer.alignmentStart, fullContainer.alignmentStart);
+            Assert.assertEquals(headerOnlyContainer.alignmentSpan, fullContainer.alignmentSpan);
+            Assert.assertEquals(headerOnlyContainer.nofRecords, fullContainer.nofRecords);
+            Assert.assertEquals(headerOnlyContainer.globalRecordCounter, fullContainer.globalRecordCounter);
+            Assert.assertEquals(headerOnlyContainer.bases, fullContainer.bases);
+            Assert.assertEquals(headerOnlyContainer.blockCount, fullContainer.blockCount);
+            Assert.assertEquals(headerOnlyContainer.landmarks, fullContainer.landmarks);
+            Assert.assertEquals(headerOnlyContainer.checksum, fullContainer.checksum);
```
cmnbroad: Can you add an equality assert here for container.offset.

src/main/java/htsjdk/samtools/cram/build/CramContainerHeaderIterator.java
```diff
@@ -0,0 +1,85 @@
...
+public class CramContainerHeaderIterator implements Iterator<Container> {
+  private CramHeader cramHeader;
+  private SeekableStream inputStream;
+  private Container nextContainer;
+  private boolean eof = false;
+  private long offset = 0;
+
+  public CramContainerHeaderIterator(final SeekableStream inputStream) throws IOException {
+    cramHeader = CramIO.readCramHeader(inputStream);
+    offset = inputStream.position();
```
cmnbroad: In looking at this I noticed that the existing `CramContainerIterator` doesn't correctly set the offset as you're ding here (it fails to account for the size of the Cram header, so the first real container has it's offset set to 0 - all of them are off after that). If we fix that, then this class can be a subclass of that one, and we won't have to duplicate the rest of the implementation. Plus we won't have to artificially constrain the stream to be a `SeekableStream`.

cmnbroad: I wanted to explore this a bit since I'm a little surprised the existing `CramContainerIterator` doesn't cause problems due to the bad offsets. Anyway, here is a [commit](https://github.com/cmnbroad/htsjdk/commit/d128c26f7ed6fae425380a515da3ce56655e2d24) with my proposed fix - see what you think.

src/main/java/htsjdk/samtools/cram/io/InputStreamUtils.java
```diff
@@ -77,4 +77,29 @@ public static void readFully(final InputStream inputStream, final byte[] b, fina
...
     }
+
+    /**
+     * Skip the specified number of bytes from the {@link InputStream}.
+     * @param in the input stream to skip bytes from
+     * @param length the number of bytes to skip
+     * @throws IOException  as per java IO contract
+     * @throws EOFException if there is less than length bytes in the stream
+     */
+    public static void skipFully(final InputStream in, final long length) throws IOException {
```
cmnbroad: Can we add a positive test, plus one that fails to consume the entire length.

src/main/java/htsjdk/samtools/cram/build/CramContainerHeaderIterator.java
tomwhite: Your fix is a great improvement - it simplifies the new code and fixes a previous issue with the code. Thank you!

tomwhite: BTW I had a quick look at why this wasn't detected before. I changed the code in Squark that uses CramContainerHeaderIterator to use CramContainerIterator - and for small CRAM files it still works, most likely because a Chunk that starts at the start of the file, rather than the first offset, works in the same way. For large CRAM files the offsets don't work however and the code throws an exception.

So it may be that testing on small CRAM files has masked this bug. CRAM indexes written with the old code may be invalid though.

lbergelson:
@tomwhite Thank you.

PR #1135, opened 2018-05-30
removing new finals in AbstractIndexedFastaSequenceFile

lbergelson:
* a number of methods that were previously non final when they were
defined in IndexFastaSequenceFile were made final when they moved to
AbstractIndexedFastaSequenceFile
* this was an overlooked API change that causes problems for existing
subclasses of IndexFastaSequenceFile
* removing final from all newly final methods except for isIndexed()


magicDGS:
Oops, sorry for that. I always forget that almost every public class is public API in htsjdk because they aren't final. Looks good to me!

lbergelson:
@magicDGS It's tricky to tell these things by eye.  No harm done.

PR #1136, opened 2018-05-31
Add wrapper class AutoClose to allow classes that aren't AutoCloseable to be used in a try-with-resources block

pshapiro4broad:
### Description

- add wrapper class `AutoClose` to allow classes that aren't `AutoCloseable` to be used in a try-with-resources block, by using CloserUtil.

- minor cleanups to `CloserUtil`

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/AutoClose.java
+++ b/src/main/java/htsjdk/samtools/util/AutoClose.java
@@ -1,0 +1,26 @@
+package htsjdk.samtools.util;
+
+/**
+ * Use this class to enable using try-with-resources on objects that do not implement AutoCloseable but
+ * do have a close() method.
+ *
+ * <pre>{@code
+ * try (AutoClose<Connection> auto = new AutoClose<>(makeConnection()) {
+ *     Connection conection = auto.object;
+ *     ...
+ * }}
+ *</pre>
+ */
+public class AutoClose<T> implements AutoCloseable {
+
+    public final T object;
+
+    public AutoClose(T object) {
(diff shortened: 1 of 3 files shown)


lbergelson:
@pshapiro4broad Why don't we just implement Closeable on the objects we want to close?  Are there specific classes you're thinking of?  This, and CloserUtil itself, seem like a bizarre hack for a situation we could probably fix in less unexpected way.   Maybe I just don't understand the use case though.

pshapiro4broad:
I needed this when adding liquibase migrations to picard-private because liquibase's database connections object `liquibase.database.DatabaseConnection` is not `AutoCloseable`. When I went to write this wrapper, it seemed like it should go next to `CloserUtil` since it was a thin wrapper around it.

But I agree that for any type we define, there's no reason not to make properly auto closeable. At one point I tried to get rid of `CloserUtil` and found that most cases are using it to (1) bypass a null check and (2) avoiding the checked exception on `close()`. It's probably worth taking another look at removing CloserUtil.

lbergelson:
@pshapiro4broad I see, that makes more sense.  I'd advocate for us getting rid of CloserUtil rather than expanding support for it.  I think it might make sense for this to live in picard-private unless there's a strong desire for it here.

lbergelson:
@pshapiro4broad Going to close this one because of my general philosophical objects to expanding support for CloserUtil.  I want to make it easier to deprecate it rather than harder.

PR #1142, opened 2018-06-06
restoring SRA support by updating the backing library

lbergelson:
* the underlying ngs-java library that supplies SRA support required updating in order to continue connecting to the SRA servers
* fixes SRA tests and restores support
* update ngs-java 1.2.4 -> 2.9.0

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,7 +41,7 @@ dependencies {
     compile "org.xerial.snappy:snappy-java:1.1.4"
     compile "org.apache.commons:commons-compress:1.4.1"
     compile "org.tukaani:xz:1.5"
-    compile "gov.nih.nlm.ncbi:ngs-java:1.2.4"
+    compile "gov.nih.nlm.ncbi:ngs-java:2.9.0"
 
     testCompile "org.scala-lang:scala-library:2.12.1"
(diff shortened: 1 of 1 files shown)


lbergelson:
@cmnbroad I was recently recontacted about SRA support.  They published a new version of the library which seems to restore support for SRA ( at least according to the tests).

lbergelson:
SRA tests now pass with this branch.

```
SRATest:
- testCounts(SRR2096940,10591,498)
- testCounts(SRR000123,0,4583)

... etc

===============================================
Command line suite
Total tests run: 35, Failures: 0, Skips: 0
===============================================
```

cmnbroad:
@lbergelson The change looks simple enough. Should we keep the SkipException tests and the exclusion for SRA when running tests locally ? Just checking to make sure there isn't anything else that needs to be be re-enabled.

lbergelson:
@cmnbroad I'm not sure what to do with the skip.. it's kind of a flakey test since it's over the internet, so the skip keeps it from randomly failing, but we also don't get alerted to it...  Maybe we could move it to a separate test matrix entry?

We should keep them as separate tests I think, some people don't like the auto-download of code that happens when you run them.

cmnbroad:
Sounds like we should keep them as i for now at least.

PR #1145, opened 2018-06-11
Load tribble index from a stream.

tomwhite:
Make it possible to load a tribble index from a stream, not just a file. This is needed for #1112.

The change is covered by existing tests.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/tribble/index/IndexFactory.java
+++ b/src/main/java/htsjdk/tribble/index/IndexFactory.java
@@ -162,7 +162,7 @@ public class IndexFactory {
      * @param indexFile from which to load the index
      */
     public static Index loadIndex(final String indexFile) {
-        return loadIndex(indexFile, null);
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/tribble/index/IndexFactory.java
```diff
@@ -174,18 +174,30 @@ public static Index loadIndex(final String indexFile) {
...
+        }
+    }
+
+    /**
+     * Load in index from the specified stream. Note that the caller is responsible for decompressing the stream if necessary.
+     *
+     * @param source the stream source (typically the file name)
+     * @param inputStream the raw byte stream of the index
+     */
+    public static Index loadIndex(final String source, final InputStream inputStream) {
```
lbergelson: @tomwhite Could you add an explicit test for this new method?  It can reuse the existing test code for the previous methods if you want.

tomwhite: Added a simple test.

PR #1146, opened 2018-06-12
making public static fields in SamReader.Type final

lbergelson:
* there were a number of fields that should be constants but were not
### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SamReader.java
+++ b/src/main/java/htsjdk/samtools/SamReader.java
@@ -81,10 +81,10 @@ public interface SamReader extends Iterable<SAMRecord>, Closeable {
             }
         }
 
-        public static Type SRA_TYPE = new TypeImpl("SRA", "sra", null);
-        public static Type CRAM_TYPE = new TypeImpl("CRAM", "cram", "crai");
-        public static Type BAM_TYPE = new TypeImpl("BAM", "bam", "bai");
-        public static Type SAM_TYPE = new TypeImpl("SAM", "sam", null);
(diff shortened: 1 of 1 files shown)


lbergelson:
@droazen easy one here for you

src/main/java/htsjdk/samtools/SamReader.java
```diff
@@ -81,10 +81,10 @@ public String toString() {
...
         }
 
-        public static Type SRA_TYPE = new TypeImpl("SRA", "sra", null);
-        public static Type CRAM_TYPE = new TypeImpl("CRAM", "cram", "crai");
-        public static Type BAM_TYPE = new TypeImpl("BAM", "bam", "bai");
-        public static Type SAM_TYPE = new TypeImpl("SAM", "sam", null);
+        public static final Type SRA_TYPE = new TypeImpl("SRA", "sra", null);
+        public static final Type CRAM_TYPE = new TypeImpl("CRAM", "cram", "crai");
+        public static final Type BAM_TYPE = new TypeImpl("BAM", "bam", "bai");
+        public static final Type SAM_TYPE = new TypeImpl("SAM", "sam", null);
```
pshapiro4broad: It looks like this could be replaced with a Java enum (assuming we don't need to allow others to subclass `SamReader.Type`).

lbergelson:
@pshapiro4broad It definitely could, but it came up in the context of someone wanting to extend this class, so I didn't want to do that.  We could expose an interface an an enum that implements it, but that seemed like overkill.

PR #1149, opened 2018-06-15
Additional unit tests for IOUtil

kachulis:
### Description

Additional unit test to IOUtilTest to increase coverage of IOUtil.

### Checklist

- [x ] Code compiles correctly
- [x ] All tests passing
- [NA] Extended the README / documentation, if necessary


src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +527,253 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
+
+
+        };
+
+    }
+    @Test(dataProvider = "filesToCompress")
+    public void testCompressionLevel(Path file,String extension,int lastDifference) throws IOException{
+        long origSize=file.toFile().length();
+        long previousSize=origSize;
+        String tmpPath = System.getProperty("java.io.tmpdir");
```
yfarjoun: use File.createTempFile for creating temporary files. (here and below)

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +527,253 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
+        String userName=System.getProperty("user.name");
+        File tmpDir = new File(tmpPath, userName);
+        tmpDir.mkdir();
+        tmpDir.deleteOnExit();
+        for(int cl=1;cl<=9;cl++)
```
yfarjoun: use meaningful variable names

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +527,253 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
+        {
+            Path outFile=Paths.get(tmpDir.getAbsolutePath(),"tmp"+extension);
+            IOUtil.setCompressionLevel(cl);
+            Assert.assertEquals(IOUtil.getCompressionLevel(),cl);
+            InputStream inStream=IOUtil.openFileForReading(file);
+            OutputStream outStream=IOUtil.openFileForWriting(outFile.toFile());
+            IOUtil.transferByStream(inStream,outStream,origSize);
+            outStream.close();
+            outFile.toFile().deleteOnExit();
```
yfarjoun: put deleteOnExit() call right after file "creation"

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +527,253 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
+        tmpDir.deleteOnExit();
+        for(int cl=1;cl<=9;cl++)
+        {
+            Path outFile=Paths.get(tmpDir.getAbsolutePath(),"tmp"+extension);
+            IOUtil.setCompressionLevel(cl);
+            Assert.assertEquals(IOUtil.getCompressionLevel(),cl);
+            InputStream inStream=IOUtil.openFileForReading(file);
+            OutputStream outStream=IOUtil.openFileForWriting(outFile.toFile());
+            IOUtil.transferByStream(inStream,outStream,origSize);
+            outStream.close();
```
yfarjoun: for resources that need to be closed, use try-with-resources

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +527,253 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
+            outFile.toFile().deleteOnExit();
+            long newSize=outFile.toFile().length();
```
yfarjoun: use Files.size(Path)

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +527,253 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
+    @DataProvider
+    public static Object [][] badCompressionLevels(){
+        return new Object[][]{
+                {-1},
+                {10}
+        };
+
+    }
+
+    @Test(dataProvider = "badCompressionLevels")
```
yfarjoun: use "expectedException" on the test here.

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +527,253 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
+
+    @Test(dataProvider="filesToCopy")
+    public void testCopyFileException(Path file){
+        String tmpPath = System.getProperty("java.io.tmpdir");
+        String userName=System.getProperty("user.name");
+        File tmpDir = new File(tmpPath, userName);
+        tmpDir.mkdir();
+        tmpDir.deleteOnExit();
+        Path outFile=Paths.get(tmpPath,userName,"tmp"+file.getFileName());
+        ProcessExecutor.execute(new String[]{"touch", outFile.toAbsolutePath().toString()});
```
yfarjoun: I think it would be better to open and close the file than to spawn an external command here.

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +527,253 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
+
+    @Test()
+    public void testReadLines() throws IOException {
+        String tmpPath = System.getProperty("java.io.tmpdir");
+        String userName = System.getProperty("user.name");
+        File tmpDir = new File(tmpPath, userName);
+        tmpDir.mkdir();
+        tmpDir.deleteOnExit();
+        Path file = Paths.get(tmpDir.getAbsolutePath(), "tmp.txt");
+        Random rand = new Random(System.currentTimeMillis());
```
yfarjoun: Better to have no true randomness in code...use a fixed seed here.

kachulis:
Thanks Yossi.  I have updated to address your requests.

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +527,240 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
     }
+
+    @DataProvider
+    public static Object[][] filesToCompress() {
+        return new Object[][]{
+                {TEST_DATA_DIR.resolve("words_longs.txt"), ".gz", 8},
+                {TEST_DATA_DIR.resolve("words_longs.txt"), ".bfq", 8},
+                {TEST_VARIANT_DIR.resolve("test1.vcf"), ".gz", 7},
+                {TEST_VARIANT_DIR.resolve("test1.vcf"), ".bfq", 7}
+
```
yfarjoun: still too many newlines here and other places.

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +527,240 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
+    }
+
+    @Test(dataProvider = "filesToCopy")
+    public void testCopyFile(final Path file) throws IOException {
+        final Path outFile = Files.createTempFile("tmp", ".tmp");
+        outFile.toFile().deleteOnExit();
+        IOUtil.copyFile(file.toFile(), outFile.toFile());
+
+
+        final Stream<String> stream = Files.lines(file);
```
yfarjoun: given that we do not need to save on memory here, 
you can replace the remaining lines of code here with a
```java
Assert.assertEquals(File.lines(file),Files.lines(outFile))
```

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +527,240 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
+
+    @Test(dataProvider = "filesToCopy", expectedExceptions = {SAMException.class})
+    public void testCopyFileReadException(final Path file) throws IOException {
+        final Path outFile = Files.createTempFile("tmp", ".tmp");
+        outFile.toFile().deleteOnExit();
+        file.toFile().setReadable(false);
+        try {
+            IOUtil.copyFile(file.toFile(), outFile.toFile());
+        } catch (SAMException e) {
+            file.toFile().setReadable(true);
```
yfarjoun: why change the readability status here?

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +527,240 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
+        regExpDir.toFile().deleteOnExit();
+        final List<String> listExpected = Arrays.asList(expected);
+        final List<File> expectedFiles = new ArrayList<File>();
+        for (String name : allNames) {
+            final Path file = regExpDir.resolve(name);
+            file.toFile().createNewFile();
+            if (listExpected.contains(name)) {
+                expectedFiles.add(file.toFile());
+            }
+            file.toFile().deleteOnExit();
```
yfarjoun: move this up to be right after the `create` call.

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +527,240 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
+        final String[] lines = new String[nLines];
+        try (final BufferedWriter writer = Files.newBufferedWriter(file)) {
+            for (int i = 0; i < nLines; i++) {
+                final String line = TEST_STRING + Integer.toString(rand.nextInt(100000000));
+                lines[i] = line;
+                writer.write(line);
+                writer.newLine();
+            }
+        }
+        final IterableOnceIterator<String> retLines = IOUtil.readLines(file.toFile());
```
yfarjoun: again, this whole block can be replaced with a single assert, I think.

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +527,240 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
+
+    @Test
+    public void testCopyDirectoryTree() throws IOException {
+        final Path copyToDir = Files.createTempDirectory("copyToDir");
+        copyToDir.toFile().deleteOnExit();
+        IOUtil.copyDirectoryTree(TEST_VARIANT_DIR.toFile(), copyToDir.toFile());
+        final Stream<Path> fileStream = Files.walk(TEST_VARIANT_DIR);
+        final Stream<Path> copyFileStream = Files.walk(copyToDir);
+
+        final Iterator<Path> itFileStream = fileStream.iterator();
```
yfarjoun: inline stream and iterator

yfarjoun: you can simplify this by actually using the stream...
```java
final List<Path> collect = Files.walk(TEST_VARIANT_DIR).filter(f -> !f.equals(TEST_VARIANT_DIR)).collect(Collectors.toList());
```

and then compare with Assert.assertEquals()

src/test/java/htsjdk/samtools/util/IOUtilTest.java
yfarjoun: if this is needed then put a comment.

yfarjoun: you want a finally here not a catch.

kachulis: yup, you're right should be a finally.  For reference, need to change back to readable so that other tests which need to read this file later will not fail (along with just being best to leave things as they were before the test).  adding explanation as comment in code.

kachulis:
Sure.  I went the dictionary route, so the 2mb random word file is replaced with a ~50kb dictionary which is used to create a temporary few mb random word file (which gets deleted on exit) instead.

lbergelson:
@kachulis Thanks. I appreciate the extra effort.  One 2mb isn't that big of a deal, but if we add a lot of them we end up with a huge unwieldy repo so it's always good to keep things slim.

kachulis:
@lbergelson no problem, that makes sense

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -91,6 +93,21 @@ public void setUp() throws IOException {
         if (!tmpDir.isDirectory())
             throw new RuntimeException("java.io.tmpdir (" + systemTempDir + ") is not a directory");
         systemUser = System.getProperty("user.name");
+        //build long file of random words for compression testing
+        WORDS_LONG = Files.createTempFile("words_long", ".txt");
+        WORDS_LONG.toFile().deleteOnExit();
+        List<String> wordsList = new ArrayList<String>();
```
yfarjoun: can be final

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +542,179 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
+        }
+    }
+
+    @DataProvider
+    public static Object[][] badCompressionLevels() {
+        return new Object[][]{
+                {-1},
+                {10}
+        };
+
```
yfarjoun: straggler unneeded newline.

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +542,179 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
+    @Test()
+    public void testReadLines() throws IOException {
+        final Path file = Files.createTempFile("tmp", ".txt");
+        file.toFile().deleteOnExit();
+        final int seed = 12394738;
+        final Random rand = new Random(seed);
+        final int nLines = 5;
+        final List<String> lines = new ArrayList<String>();
+        try (final BufferedWriter writer = Files.newBufferedWriter(file)) {
```
yfarjoun: Not sure it's nicer, but you can wrap the bufferedWriter with a PrintWriter and then call println() instead of write() + newLine().

src/test/java/htsjdk/samtools/util/IOUtilTest.java
```diff
@@ -520,4 +542,179 @@ public void testIsBlockCompressedOnJimfs(Path file, boolean checkExtension, bool
...
+        final String ret = IOUtil.fileSuffix(file.toFile());
+        Assert.assertEquals(ret, expected);
+    }
+
+    @Test
+    public void testCopyDirectoryTree() throws IOException {
+        final Path copyToDir = Files.createTempDirectory("copyToDir");
+        copyToDir.toFile().deleteOnExit();
+        IOUtil.copyDirectoryTree(TEST_VARIANT_DIR.toFile(), copyToDir.toFile());
+        final List<String> collect = Files.walk(TEST_VARIANT_DIR).filter(f -> !f.equals(TEST_VARIANT_DIR)).map(p -> p.getFileName().toString()).collect(Collectors.toList());
```
yfarjoun: is toString needed after getFileName?

kachulis: No, you're right, can make this a List<Path> instead of a List<String> and still works.

yfarjoun: I updated your comment since github sanitizes things that look like html....

kachulis: ha, oops, thanks

PR #1150, opened 2018-06-18
deprecating overloads makeCRAMWriter with mixed Path/File

lbergelson:
* deprecating overloads of SAMFileWriterFactory.makeCRAMWriter that took a mix of Path and File as inputs
   these were needed when we supported Path for the CRAM but not the reference

* removing one overload that seems redundant and hasn't been in any htsjdk release yet

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMFileWriterFactory.java
+++ b/src/main/java/htsjdk/samtools/SAMFileWriterFactory.java
@@ -452,7 +452,7 @@ public class SAMFileWriterFactory implements Cloneable {
      *
      */
     public SAMFileWriter makeWriter(final SAMFileHeader header, final boolean presorted, final File outputFile, final File referenceFasta) {
-        return makeWriter(header, presorted, null == outputFile ? null : outputFile.toPath(), referenceFasta);
+        return makeWriter(header, presorted, null == outputFile ? null : outputFile.toPath(), null == outputFile ? null : referenceFasta.toPath());
     }
 
     /**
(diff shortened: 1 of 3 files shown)


lbergelson:
@cmnbroad Since you brought up the issue I'm going to let you review this one.
@magicDGS any objections to this change?

lbergelson:
@cmnbroad What was the comment?

src/main/java/htsjdk/samtools/SAMFileWriterFactory.java
```diff
@@ -550,9 +550,11 @@ public CRAMFileWriter makeCRAMWriter(final SAMFileHeader header, final File outp
      * @param referenceFasta reference sequence file
      * @return CRAMFileWriter
      *
+     * @deprecated since 6/18, use {@link #makeCRAMWriter(SAMFileHeader, boolean, Path, Path)} instead
```
cmnbroad: @lbergelson Hm I dropped my connection. Anyway, should we suggest migration to `makeWriter(SAMFileHeader, boolean, Path, Path)` as the canonical all-knowing method. And someday deprecate all the rest ?

lbergelson: Oh, you're right.  I'll change.

lbergelson:
@cmnbroad Is this better?

magicDGS:
No objections to this @lbergelson - I did not want to remove or not include the overrides to do not break backwards compatibility. Deprecation sounds like a good solution!

PR #1153, opened 2018-06-19
don't run FTP tests since they will always fail

lbergelson:
* travis permanently changed their network configuration in a way that breaks ftp
* don't waste time running the tests that are guaranteed to fail

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,13 +11,6 @@ cache:
 jdk:
   - oraclejdk8
   - openjdk8
-matrix:
-  fast_finish: true
-  allow_failures: 
-     - env: TEST_TYPE=FTP
-  include:
-     - jdk: oraclejdk8
-       env: TEST_TYPE=FTP
 script:
   - if [[ $TEST_TYPE == "FTP" ]]; then
         ./gradlew testFTP jacocoTestReport;


yfarjoun:
there was a comment in one of the issues about using commons FTP client instead as it works....did you look into that? perhaps our client is dong it wrong?

yfarjoun:
I think that the tests should be moved back to the main group and be dependent on a '@BeforeTest' like we do with SRA. that way when developers test on their own machines, the tests will go through FTP as well.

.travis.yml
```diff
@@ -11,13 +11,6 @@ cache:
...
   - openjdk8
-matrix:
-  fast_finish: true
-  allow_failures: 
-     - env: TEST_TYPE=FTP
-  include:
-     - jdk: oraclejdk8
-       env: TEST_TYPE=FTP
 script:
   - if [[ $TEST_TYPE == "FTP" ]]; then
```
cmnbroad: Should we remove the reference here to "FTP" since its no longer set in the environment in the matrix block ?

yfarjoun:
the PR review clearly took  "forever" since now the FTP tests are passing. Closing.

PR #1154, opened 2018-06-19
update to gradle 4.8

lbergelson:
* Upgrading our version of gradle and our test dependencies.
  This is a move towards supporting building under java 9/10 although
  some issues remain.

  upgrading from gradle 3.2.1 -> 4.8
  upgrading shadow plugin  1.2.3 -> 2.0.4
  updating scalatest plugin 0.15 -> 0.22

  testng 6.13.1 -> 6.14.3
  scala-library 2.12.1 -> 2.12.6
  scalatest 3.0.1 -> 3.0.5
  pegdown 1.4.2 -> 1.6.0

  updating the gradle wrapper task definition because the way we were doing it has been deprecated
  removing jacoco version specification so we stay up to date automatically


build.gradle
```diff
@@ -11,19 +11,14 @@ plugins {
     id 'signing'
     id 'jacoco'
     id 'com.palantir.git-version' version '0.5.1'
```
magicDGS: Cannot you update also the git-version?

lbergelson: That's a good point.  I had avoided it because I didn't want any version name surprises, but it's worthing looking at.

lbergelson: It seems fine, I've updated it.

lbergelson:
@cmnbroad Yeah, they seem to systematically rewrite their wrapper script with every release.  It's not clear why since they all seem to work exactly the same.

lbergelson:
They released a 4.8.1 since I pushed this, so I'm bumping to that.  Merging when tests are green.

PR #1156, opened 2018-06-21
Fixes for CachingBAMFileIndex

### LINKED ISSUE #1127: CachingBAMFileIndex issue
### Verify
Can you see anything in the logs? 
No.

The code below is entered when SamReaderFactory.Option.CACHE_FILE_BASED_INDEXES is enabled.

### Subject of the issue
I'm getting some erroneous caching behaviour from the CachingBAMFileIndex.java getQueryResults(final int referenceIndex) method when loading indexed bam files, whereby the following statement:

// If this query is for the same reference index as the last query, return it.
if(mLastReferenceRetrieved!=null && mLastReferenceRetrieved == referenceIndex)
            return mQueriesByReference.get(referenceIndex);

returns null and the code does not attempt to go to disk to load the missing data. It seems that if mLastReferenceRetrieved is for example 2 and there is an entry with a key of 2 in mQueriesByReference (i.e. in the cache), but that key is garbage collected [removed] from the weak hash map (due to there being no current references to it) then if the method is called later with a referenceIndex input arg of 2, null gets returned instead of reloading the cache [i.e. falling through to the next code section]. Hence the data for that reference section becomes unfetchable and cannot be loaded, unless another reference sequence is selected before trying again. 

The problem seems to be that mLastReferenceRetrieved and mQueriesByReference get out of sync.

CachingBAMFileIndex is used by default for CRAM file loading (from CRAMFileReader), but I haven't been able to replicate the behaviour for that. However, this was using significantly smaller files so it may still be an issue with bigger ones.

This bug seems to show up most readily under memory intensive conditions with bigger bam/fasta files presumably where garbage collection is happening more often.

### Your environment
OSX, Ubuntu

lbergelson:
* fixing several related bugs in CachingBAMFileIndex
  * fixed an issue where cache hits in contigs with indexes higher than 127 could result in null values being returned incorrectly
  * fixed a memory leak where we never released contigs with indexes 0 -> 127 from the cache
* fixes #1127

The underlying issue was confusion over how WeakHashMap interacts with Integer.

### DIFF
--- a/src/main/java/htsjdk/samtools/CachingBAMFileIndex.java
+++ b/src/main/java/htsjdk/samtools/CachingBAMFileIndex.java
@@ -29,7 +29,6 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.BitSet;
 import java.util.List;
-import java.util.WeakHashMap;
 
 /**
  * Class for reading BAM file indices, caching each contig as it's loaded and
@@ -37,8 +36,10 @@ import java.util.WeakHashMap;
  */
 class CachingBAMFileIndex extends AbstractBAMFileIndex implements BrowseableBAMIndex
 {
-    private Integer mLastReferenceRetrieved = null;
-    private final WeakHashMap<Integer,BAMIndexContent> mQueriesByReference = new WeakHashMap<Integer,BAMIndexContent>();
+    private Integer lastReferenceIndex = null;
+    private BAMIndexContent lastReference = null;
+    private long cacheHits = 0;
+    private long cacheMisses = 0;
 
     public CachingBAMFileIndex(final File file, final SAMSequenceDictionary dictionary) {
         super(file, dictionary);
@@ -110,7 +111,7 @@ class CachingBAMFileIndex extends AbstractBAMFileIndex implements BrowseableBAMI
         final int firstLocusInBin = getFirstLocusInBin(bin);
 
         // Add the specified bin to the tree if it exists.
-        final List<Bin> binTree = new ArrayList<Bin>();
+        final List<Bin> binTree = new ArrayList<>();
         if(indexQuery.containsBin(bin))
             binTree.add(indexQuery.getBins().getBin(bin.getBinNumber()));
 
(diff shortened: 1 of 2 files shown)


lbergelson:
@droazen Could you take a look at this one when you get a chance?

src/main/java/htsjdk/samtools/CachingBAMFileIndex.java
```diff
@@ -143,31 +144,35 @@ public BAMFileSpan getSpanOverlapping(final Bin bin) {
...
-        if(mLastReferenceRetrieved!=null && mLastReferenceRetrieved == referenceIndex)
-            return mQueriesByReference.get(referenceIndex);
-
-        // If not, check to see whether it's available in the cache.
-        BAMIndexContent queryResults = mQueriesByReference.get(referenceIndex);
-        if(queryResults != null) {
-            mLastReferenceRetrieved = referenceIndex;
-            mQueriesByReference.put(referenceIndex,queryResults);
-            return queryResults;
+        if(lastReferenceIndex!=null && lastReferenceIndex == referenceIndex){
```
droazen: Can you add a comment about the unboxing going on here, just so this doesn't raise eyebrows in the future?

src/main/java/htsjdk/samtools/CachingBAMFileIndex.java
```diff
@@ -143,31 +144,35 @@ public BAMFileSpan getSpanOverlapping(final Bin bin) {
...
+                return lastReference;
+            } else {
+                throw new RuntimeException("Expected to find a cached reference but we found null.  This is a bug.");
+            }
         }
 
-        // If not in the cache, attempt to load it from disk.
-        queryResults = query(referenceIndex,1,-1);
+        // If not attempt to load it from disk.
+        BAMIndexContent queryResults = query(referenceIndex,1,-1);
```
droazen: When `query()` returns null, do you want to clear out the cache?

src/main/java/htsjdk/samtools/CachingBAMFileIndex.java
```diff
@@ -143,31 +144,35 @@ public BAMFileSpan getSpanOverlapping(final Bin bin) {
...
+            lastReferenceIndex = referenceIndex;
+            lastReference = queryResults;
+            return lastReference;
         }
 
         // Not even available on disk.
         return null;
     }
+
+    public long getCacheHits() {
```
droazen: Add javadoc for these new public methods.

src/test/java/htsjdk/samtools/CachingBAMFileIndexTest.java
```diff
@@ -0,0 +1,93 @@
...
+            assertCacheStats(index, 1, 2);
+
+            index.getQueryResults(150);
+            assertCacheStats(index,2,2);
+
+            index.getQueryResults(150);
+            assertCacheStats(index,3,2);
+
+            index.getQueryResults(1);
+            assertCacheStats(index, 3, 3);
```
droazen: Add one additional query on chromosome 1 to make sure it gets put back in the cache.

src/main/java/htsjdk/samtools/CachingBAMFileIndex.java
droazen: If you do make this change, be sure to add a test, and document what it means to cache an "empty" result.

PR #1157, opened 2018-06-26
fixing spelling error Indeces -> Indices

lbergelson:
There are two valid plurals of index, Indeces isn't one of them.

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
+++ b/src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
@@ -567,13 +567,26 @@ public class GenotypeLikelihoods {
     }
 
     /**
-     * get the PL indexes (AA, AB, BB) for the given allele pair; assumes allele1Index &lt;= allele2Index.
+     * get the PL indices (AA, AB, BB) for the given allele pair; assumes allele1Index &lt;= allele2Index.
+     *
+     * @param allele1Index    the index in VariantContext.getAllele() of the first allele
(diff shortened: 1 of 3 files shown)


src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -567,13 +567,26 @@ public static GenotypeLikelihoodsAllelePair getAllelePairUsingDeprecatedOrdering
...
+    }
+
+    /**
+     * get the PL indices (AA, AB, BB) for the given allele pair; assumes allele1Index &lt;= allele2Index.
+     *
+     * @param allele1Index    the index in VariantContext.getAllele() of the first allele
+     * @param allele2Index    the index in VariantContext.getAllele() of the second allele
+     * @return the PL indexes
+     */
+    public static int[] getPLIndicesOfAlleles(final int allele1Index, final int allele2Index) {
```
cmnbroad: Just a thought - should we consider having new/replacement methods prefer Lists over arrays ? or maybe thats overkill when the type is int.

lbergelson: It's a good question... I think it might be one of those cases where the extra cost isn't worth the convenience.  Adding any extra allocation overhead to the Variant classes is always kind of questionable. That said, I don't have any clue if this is a hotspot anywhere... so... I'm not sure.  I think I'm going to keep it as is, since I don't want to force people to change their downstream code here beyond just fixing the annoying spelling.  Definitely something to think about for htsjdk3 though.

PR #1158, opened 2018-07-02
adding position to SAMRecord.toString()

lbergelson:
* previously it only said if it was mapped or not, now it includes the position if it's mapped


yfarjoun:
👍

Issue #1065, opened 2018-01-09
queryMate method throws exception for CRAM records with the same read name


kpepper:
### Subject of the issue
Occasionally I am getting the following exception when calling the SamReader queryMate method for a CRAM file…..

htsjdk.samtools.SAMFormatException: Multiple SAMRecord with read name IL29_4505:7:93:5878:1091#2 for first end.
	at htsjdk.samtools.SamReader$PrimitiveSamReaderToSamReaderAdapter.queryMate(SamReader.java:451)

I don’t see the same thing for the equivalent BAM file.

### Your environment
* version of htsjdk = 2.12.0
* version of java = 1.8.151
* which OS = Mac Sierra

### Steps to reproduce
e.g. the CRAM file has the following reads, that seem to be causing the issue:

IL29_4505:7:93:5878:1091#2	163	Mito	267	60	108M	=	631	399	TTAATTAATTAATAATAAAAATATAATTATAAATAATATAAATATTATTCTTTATTAATAAATATATATTTATATATTATAAAAGTATCTTAATTAATAAAAATAAAC	BBBBBBB@BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB@BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB>BBBBBBBBBBB=BBBBBBBBB@?9A=A:/8	AS:i:108	XS:i:0	MD:Z:108	NM:i:0	RG:Z:lane1
IL29_4505:7:93:5878:1091#2	83	Mito	631	60	73S35M	=	267	-399	AATGAAAATTTTATAAAAAAAGTAATGAATATTCCTTTTTAAAAATACCTTCTTAATTAAATAAAATAAAATAAAAAGGGGACTTTATATTTATAAAGTAATTATATT	<=8=;BCBBABBBBBBBBBC?BDDDBBBBB0DDBBBBBCBBBBBBAAABB=CCCCCCBBBBBBBBBBBBBBBBBBDBBBDDCBBBBBBBBBBBBBDDDBDBBBBBDCD	AS:i:35	XS:i:0	SA:Z:Mito,44163,+,37S39M32S,35,1;	MD:Z:35	NM:i:0	RG:Z:lane1
[...]

PR #1164, opened 2018-07-12
CRAM queryAlignmentStart/queryMate fix.

cmnbroad:
Fixes an issue discovered as part of debugging https://github.com/samtools/htsjdk/issues/1065. The CRAM implementation of queryAlignmentStart (which is in turn used by queryMate) currently returns all reads in a container that start *after* the requested alignment start. This PR pulls out the `BAMStartingAtFilterIterator` into a top-level class so it can be reused by the CRAM implementation to filter, and then stop iterating after, all reads with the desired start have been returned.


src/test/java/htsjdk/samtools/CRAMIndexQueryTest.java
```diff
@@ -569,46 +607,95 @@ public void testQueryUnmapped(
...
     public void testQueryMate(
         final File cramFileName,
         final File referenceFileName,
-        final String expectedName) throws IOException
+        final String queryContig,
+        final int alignmentStart,
+        final String expectedReadName) throws IOException
     {
-        SamReaderFactory factory = SamReaderFactory.makeDefault();
+        SamReaderFactory factory = SamReaderFactory.makeDefault().validationStringency(ValidationStringency.LENIENT);
```
yfarjoun: why lenient?

src/test/java/htsjdk/samtools/CRAMIndexQueryTest.java
```diff
@@ -618,7 +705,7 @@ private void doQueryTest(
         final File referenceFileName,
         final String[] expectedNames) throws IOException
     {
-        SamReaderFactory factory = SamReaderFactory.makeDefault();
+        SamReaderFactory factory = SamReaderFactory.makeDefault().validationStringency(ValidationStringency.LENIENT);
```
yfarjoun: why lenient?

src/main/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -492,21 +494,30 @@ void enableFileSource(final SamReader reader, final boolean enabled) {
...
             this(queries, contained, coordinatesFromQueryIntervals(getIndex(), queries));
         }
 
-        public CRAMIntervalIterator(final QueryInterval[] queries, final boolean contained, final long[] coordinates) {
+        public CRAMIntervalIteratorBase(final QueryInterval[] queries, final boolean contained, final long[] coordinates) {
             super(queries, contained);
+        }
 
+        /**
+         * Subclasses must call this method in their constructors AFTER construction of this class is complete.
```
yfarjoun: could you not have a "abstract callInitialize()" method in the base class and then call that method from the base constructor? that would force an subclass to implement this method and read this comment...otherwise I'm not sure it will be noticed....

src/main/java/htsjdk/samtools/BAMStartingAtIteratorFilter.java
```diff
@@ -0,0 +1,43 @@
...
+        if (alignmentStart > mRegionStart) {
+            // If scanned beyond target region, end iteration
+            return FilteringIteratorState.STOP_ITERATION;
+        } else  if (alignmentStart == mRegionStart) {
+            return FilteringIteratorState.MATCHES_FILTER;
+        } else {
+            return FilteringIteratorState.CONTINUE_ITERATION;
+        }
+    }
+
```
yfarjoun: extra nl (from previous code)

src/main/java/htsjdk/samtools/BAMStartingAtIteratorFilter.java
```diff
@@ -0,0 +1,43 @@
+package htsjdk.samtools;
```
yfarjoun: license

yfarjoun: why not htsjdk.samtools.filter ?

src/main/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -492,21 +494,30 @@ void enableFileSource(final SamReader reader, final boolean enabled) {
         return BAMFileSpan.merge(spanArray).toCoordinateArray();
     }
 
-    private class CRAMIntervalIterator extends BAMQueryMultipleIntervalsIteratorFilter
+    private abstract class CRAMIntervalIteratorBase extends BAMQueryMultipleIntervalsIteratorFilter
```
yfarjoun: it would be good to note here the reason for the abstractness, not only in line 513....or better, actually have an abstract method whose implementation will resolve the issue.

src/main/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -542,19 +553,20 @@ public SAMRecord next() {
...
-                switch(compareToFilter(nextRecord)) {
+                switch (compareToFilter(nextRecord)) {
                     case MATCHES_FILTER:
                         nextRec = nextRecord;
                         break;
                     case CONTINUE_ITERATION:
                         continue;
                     case STOP_ITERATION:
-                        break;
+                        nextRec = null;
```
pshapiro4broad: Under what circumstances can you reach this line and have `nextRec` not be null? It's not modified outside of this method. Making it a `private` field would make this a little clearer.

src/main/java/htsjdk/samtools/CRAMFileReader.java
pshapiro4broad: I agree that this is error prone but at least it's a private class, so it's likely a future subclasser will read this comment before writing code.

I don't see why this class still has `CRAMIntervalIteratorBase(final QueryInterval[] queries, final boolean contained, final long[] coordinates)` as `coordinates` is ignored. Removing that argument will also be a way to tell callers that they must call this method, which you could rename to something like `setCoordinates()`.

src/test/java/htsjdk/samtools/CRAMFileBAIIndexTest.java
```diff
@@ -51,43 +51,43 @@
     @Test
     public void testConstructors () throws IOException {
         CRAMFileReader reader = new CRAMFileReader(cramFile, indexFile, source, ValidationStringency.SILENT);
-        CloseableIterator<SAMRecord> iterator = reader.queryAlignmentStart("chrM", 1500);
+        CloseableIterator<SAMRecord> iterator = reader.queryAlignmentStart("chrM", 1519);
```
pshapiro4broad: Using a named constant instead of a magic number would avoid the need to modify it in many places the next time this changes. And also make it more apparent that the same value is used.

src/test/java/htsjdk/samtools/CRAMFileBAIIndexTest.java
```diff
@@ -51,43 +51,43 @@
...
         Assert.assertTrue(iterator.hasNext());
         SAMRecord record = iterator.next();
 
         Assert.assertEquals(record.getReferenceName(), "chrM");
-        Assert.assertTrue(record.getAlignmentStart() >= 1500);
+        Assert.assertEquals(record.getAlignmentStart(),1519);
```
pshapiro4broad: missing space after `,` on many lines here

src/test/java/htsjdk/samtools/CRAMIndexQueryTest.java
```diff
@@ -569,46 +607,95 @@ public void testQueryUnmapped(
...
+    }
+
+    @Test(dataProvider="alignmentStartQueries")
+    public void testQueryAlignmentStart(
+            final File cramFileName,
+            final File referenceFileName,
+            final String queryContig,
+            final int alignmentStart,
+            final int expectedReadCount) throws IOException
+    {
```
pshapiro4broad: open brace should be at end of previous line, not on its own line

src/test/java/htsjdk/samtools/CRAMIndexQueryTest.java
cmnbroad: The test file ("mitoAlignmentStartTest.cram"), which was provided by a user who reported originally reported the queryMate bug, contains reads that have a mateReferenceName = "*", but with  mateAlignmentStart != 0; so we have to use ValidationStringency.LENIENT to suppress validation from throwing when it sees these.

src/main/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: Yes, that constructor appears to be redundant. Removed. Not sure I understand how that helps with the requirement to call initialize though.

src/main/java/htsjdk/samtools/BAMStartingAtIteratorFilter.java
cmnbroad: To me, this seems to have more affinity with `BAMIteratorFilter' and it's other subclass `BAMQueryMultipleIntervalsIteratorFilter`, which are all in this package. Happy to move it if you still think its better there though.

src/main/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: Comment added. Not sure I see understand how the suggestion to add another abstract method resolves the initialize issue though. Can you give a more specific example ?

src/main/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: Yes, the assignment seems to be redundant. fixed.

src/test/java/htsjdk/samtools/CRAMFileBAIIndexTest.java
cmnbroad: Done, including using a constant for the sequence name.

src/test/java/htsjdk/samtools/CRAMFileBAIIndexTest.java
cmnbroad: Done.

PR #1165, opened 2018-07-13
CW only open and read a FileInputStream once in AbstractVCFCodec.canDecodeFile

cwhelan:
### Description

Currently classes implementing `AbstractVCFCodec` (`VCFCodec` and `VCF3Codec`) open and read from an input stream three times when `canDecodeFile` is called on them, to check whether the file is uncompressed text, gzipped, or block gzipped, and contains the VCF header line. This is an expensive operation in some cases, such as when reading files from cloud storage. 

This change makes each codec open a file input stream only once per file, buffering the stream so that it can be checked multiple times. When reading files from google cloud storage through NIO, this results in a speedup of .5 - 1.5 seconds per file per codec, or about 2-3 seconds for both `VCFCodec` and `VCF3Codec` to answer `canDecodeFile`. This can result in a significant overall speedup for tools that are trying to open many files from cloud storage. 

Added a couple of minor tests to try to catch edge conditions in stream reading logic.

### Checklist

- [ X ] Code compiles correctly
- [ X ] New tests covering changes and new functionality
- [ * ] All tests passing

* EnaRefServiceTest is failing for me, but that does not appear to be related to my change as far as I can tell.


src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -621,31 +618,50 @@ private static void parseSingleAltAllele(List<Allele> alleles, String alt, int l
...
+        InputStream fileInputStream = null;
         try {
-            Path path = IOUtil.getPath(potentialInput);
-            //isVCFStream closes the stream that's passed in
-            return isVCFStream(Files.newInputStream(path), MAGIC_HEADER_LINE) ||
-                    isVCFStream(new GZIPInputStream(Files.newInputStream(path)), MAGIC_HEADER_LINE) ||
-                    isVCFStream(new BlockCompressedInputStream(Files.newInputStream(path)), MAGIC_HEADER_LINE);
+            final Path path = IOUtil.getPath(potentialInput);
+            fileInputStream = Files.newInputStream(path);
+            BufferedInputStream bufferedInputStream = new BufferedInputStream(fileInputStream);
```
magicDGS: Add a buffer size argument to the constructor to ensure that it can reset safely (`MAGIC_HEADER_LINE.length()`)

cwhelan: I wouldn't want such a small buffer size because reading from the`GZIPInputStream` will read more than `MAGIC_HEADER_LINE.length()` bytes, since it will also read the gzip header. Changing this causes `Resetting to invalid mark` IO exceptions (which are then swallowed) when these guys try to reset after reading with the `GZIPInputStream`. Without this argument I was using the default `BufferedInputStream` buffer size, which is 8192. Is there a better buffer size to use?

magicDGS: What about `Defaults.NON_ZERO_BUFFER_SIZE` and/or ensure enough buffer size for the magic string `Math.max(Defaults.NON_ZERO_BUFFER_SIXE, MAGIC_HEADER_LINE.length())`?

For the `GZIPInputStream` there is also a constructor with buffer size; doesn't it work in that case?

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -621,31 +618,50 @@ private static void parseSingleAltAllele(List<Allele> alleles, String alt, int l
...
+            bufferedInputStream.mark(0);
```
yfarjoun: Could you only pass in one inputstream into `isVCFStream` and reset here rather than within isVCFStream?

the code here would then look like:
```
bufferedInputStream.mark(0);
if isVCFStream(bufferedInputStream, MAGIC_HEADER_LINE) return true;
bufferedInputStream.reset(0);
if isVCFStream(new GZIPInputStream(bufferedInputStream), MAGIC_HEADER_LINE) return true;
bufferedInputStream.reset(0);
if isVCFStream(new BlockCompressedInputStream(bufferedInputStream), MAGIC_HEADER_LINE) return true;
return false;

File: src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java [hunk 3]
```diff
@@ -621,31 +618,50 @@ private static void parseSingleAltAllele(List<Allele> alleles, String alt, int l
...
     public static boolean canDecodeFile(final String potentialInput, final String MAGIC_HEADER_LINE) {
+        InputStream fileInputStream = null;
         try {
-            Path path = IOUtil.getPath(potentialInput);
-            //isVCFStream closes the stream that's passed in
-            return isVCFStream(Files.newInputStream(path), MAGIC_HEADER_LINE) ||
-                    isVCFStream(new GZIPInputStream(Files.newInputStream(path)), MAGIC_HEADER_LINE) ||
-                    isVCFStream(new BlockCompressedInputStream(Files.newInputStream(path)), MAGIC_HEADER_LINE);
+            final Path path = IOUtil.getPath(potentialInput);
+            fileInputStream = Files.newInputStream(path);
```
yfarjoun: convert this to try-with-resources

yfarjoun: (for both fileInputStream, and bufferedInputStream)

lbergelson:
@cwhelan Did you want to respond to these comments or should someone take this over?

yfarjoun:
@cwhelan could you elaborate on the non-safe issues, or close this PR if it is no longer alive.

cwhelan:
Closing since I haven't had a chance to work on this. I will open a new PR if I can resolve the issues I was having.

PR #1172, opened 2018-08-08
Ported GATK's FastaReferenceWriter

yfarjoun:
- Implemented some basic validation functionality rather than stripping it out of the implementation of FastaReferenceWriter)
- Implemented a simple version of a RandomDNA creator (and modified SAMRecordSetBuilder to use it)
- Ported the tests (for FastaReferenceWriter) from GATK.
- Replaces https://github.com/samtools/htsjdk/pull/1058
### Description

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


yfarjoun:
@alecw will this work as well for you?

alecw:
@yfarjoun I'll try this when I get a chance.

src/main/java/htsjdk/ValidationUtils.java
```diff
@@ -0,0 +1,150 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package htsjdk;
```
magicDGS: This can better live in `htsjdk.utils`

src/main/java/htsjdk/ValidationUtils.java
```diff
@@ -0,0 +1,150 @@
...
+
+import java.util.Collection;
+import java.util.function.Supplier;
+
+/**
+ * Simple functions that streamline the checking of values.
+ */
+public class ValidationUtils {
```
magicDGS: Finally a class like this in HTSJDK!

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,758 @@
...
+ * The two main operations that one can invoke on a opened writer is {@link #startSequence} and {@link #appendBases}.
+ * The former indicates that we are going to append a new sequence to the output and is invoked once per sequence.
+ * The latter adds bases to the current sequence and can be called as many times as is needed.
+ * </p>
+ * <p>
+ * The writer will make sure that the output adheres to the FASTA reference sequence file format restrictions:
+ * <ul>
+ * <li>Sequence names are valid (non-empty, without space/blank, control characters),</li>
+ * <li>sequence description are valid (without control characters),</li>
+ * <li>bases are valid nucleotides ore IUPAC redundancy codes and X [ACGTNX...] (lower or uppercase are accepted),</li>
```
magicDGS: typo: `nucleotides ore IUPAC` -> `nucleotides or IUPAC`

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,758 @@
...
+     * respectively.
+     * </p>
+     *
+     * @param fastaFile      the output fasta file path.
+     * @param makeFaiOutput  whether an index must be generated.
+     * @param makeDictOutput whether a dictionary must be generated.
+     * @throws IllegalArgumentException if {@code fastaFile} is {@code null}.
+     * @throws IOException              if such exception is thrown when accessing the output path resources.
+     */
+    public FastaReferenceWriter(final Path fastaFile, final boolean makeFaiOutput, final boolean makeDictOutput)
```
magicDGS: Maybe it is better to have a factory instead of too many constructors...

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,758 @@
...
+     * @throws IllegalArgumentException if {@code fastaFile} is {@code null} or {@code basesPerLine} is 0 or negative.
+     * @throws IOException              if such exception is thrown when accessing the output path resources.
+     */
+    public FastaReferenceWriter(final Path fastaFile, final int basesPerLine, final Path indexFile, final Path dictFile)
+            throws IOException {
+        // This code is a slight repeat of {@link #FastaReferenceWriter(OutputStream,int,OutputStream,OutputStream)
+        // for the sake of avoiding creating output if basesPerLine is invalid.
+        this.defaultBasePerLine = checkBasesPerLine(basesPerLine);
+
+        this.fastaStream = new CountingOutputStream(Files.newOutputStream(fastaFile));
```
magicDGS: Support for compressed FASTA files (with gzi/fai index) can be also added.

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,758 @@
...
+            writeIndexEntry();
+            writeDictEntry();
+            fastaStream.write(LINE_SEPARATOR);
+            currentBasesCount = 0;
+            currentLineBasesCount = 0;
+            currentSequenceName = null;
+        }
+    }
+
+    private void writeIndexEntry()
```
magicDGS: Isn't this functionality somewhere already? Otherwise, it will be nice to have an index-codec for decoding/encoding

magicDGS: Or a `FastaIndexing` class to index on the fligh streams

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,758 @@
...
+     *
+     * @param whereTo     the path to. must not be null.
+     * @param makeIndex   whether the index file should be written at its standard location.
+     * @param makeDict    whether the dictionary file should be written at it standard location.
+     * @param name        the sequence name, cannot contain white space, or control chracter or the header start character.
+     * @param description the sequence description, "" if no description.
+     * @param bases       the sequence bases, cannot be {@code null}.
+     * @throws IOException if such exception is thrown when writing in the output resources.
+     */
+    public static void writeSingleSequenceReference(final Path whereTo, final boolean makeIndex,
```
magicDGS: If there is a factory/builder for the writer, the static methods can be moved there

src/main/java/htsjdk/utils/NullWriter.java
```diff
@@ -0,0 +1,32 @@
+package htsjdk.utils;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.Writer;
+
+/**
+ * Implementation of a writer that does nothing
+ */
+public class NullWriter extends Writer {
```
magicDGS: Should this be in the main or the test sources?

src/test/java/htsjdk/samtools/reference/FastaReferenceWriterTest.java
```diff
@@ -0,0 +1,543 @@
...
+import java.io.*;
+import java.net.URISyntaxException;
+import java.nio.file.Path;
+import java.security.GeneralSecurityException;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+public class FastaReferenceWriterTest extends HtsjdkTest {
```
magicDGS: Will be nice to have a test that reads an input FASTA, writes it down and checks if the files are the same (either MD5 or re-reading line by line).

src/main/java/htsjdk/ValidationUtils.java
yfarjoun: :-)

src/main/java/htsjdk/utils/NullWriter.java
yfarjoun: main....its used in the production code

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
yfarjoun: later perhaps...I wanted to get something in that works for the simple case...feel free to modify it afterwards...

src/main/java/htsjdk/ValidationUtils.java
```diff
@@ -0,0 +1,150 @@
...
+     * Checks that a {@link Collection} is not {@code null} and that it is not empty.
+     * If it's non-null and non-empty it returns the input, otherwise it throws an {@link IllegalArgumentException}
+     * @param collection any Collection
+     * @param message a message to include in the output
+     * @return the original collection
+     * @throws IllegalArgumentException if collection is null or empty
+     */
+    public static <I, T extends Collection<I>> T nonEmpty(T collection, String message){
+        nonNull(collection, "The collection is null: " + message);
+        if(collection.isEmpty()){
```
nh13: `){` -> `) {` here and throughout.  Generally, the spacing in `if/else` and so on is not consistent here and throughout.

nh13: `final`s can be used here and throughout.

src/main/java/htsjdk/ValidationUtils.java
```diff
@@ -0,0 +1,150 @@
...
+    /**
+     * Checks that an {@link Object} is not {@code null} and returns the same object or throws an {@link IllegalArgumentException}
+     * @param object any Object
+     * @param message the text message that would be passed to the exception thrown when {@code o == null}.
+     * @return the same object
+     * @throws IllegalArgumentException if a {@code o == null}
+     */
+    public static <T> T nonNull(final T object, final Supplier<String> message) {
+        if (object == null) {
+            throw new IllegalArgumentException(message.get());
```
nh13: this and below is ripe for a method like 

```
public static T require(final T actual, final T expected, final Supplier<String> message) {
    if (actual == expected) return actual;
    else throw new IllegalArgumentException(message.get());
}
```

I'd just use `require` in scala :P

src/main/java/htsjdk/ValidationUtils.java
```diff
@@ -0,0 +1,150 @@
...
+     * If it's non-null and non-empty it returns the input, otherwise it throws an {@link IllegalArgumentException}
+     * @param collection any Collection
+     * @return the original collection
+     * @throws IllegalArgumentException if collection is null or empty
+     */
+    public static  <I, T extends Collection<I>> T nonEmpty(T collection){
+        return nonEmpty(collection, "collection must not be null or empty.");
+    }
+
+    public static void validateArg(final boolean condition, final String msg){
```
nh13: docs here and blow

src/main/java/htsjdk/samtools/SAMRecordSetBuilder.java
```diff
@@ -360,11 +360,7 @@ private int getReadLengthFromCigar(final SAMRecord rec) {
      */
     private void fillInBases(final SAMRecord rec) {
         final int length = getReadLengthFromCigar(rec);
-        final byte[] bases = new byte[length];
-
-        for (int i = 0; i < length; ++i) {
-            bases[i] = BASES[this.random.nextInt(BASES.length)];
-        }
+        final byte[] bases = SequenceUtil.getRandomBases(random, length);
```
nh13: I think you can get rid of `BASES` now.

src/main/java/htsjdk/samtools/util/SequenceUtil.java
```diff
@@ -1062,4 +1063,14 @@ public static String getSamReadNameFromFastqHeader(final String fastqHeader) {
...
+
+    static public byte[] getRandomBases(Random random, final int length) {
+        final byte[] bases = new byte[length];
+
+        for (int i = 0; i < length; ++i) {
+            bases[i] = VALID_BASES_UPPER[random.nextInt(VALID_BASES_UPPER.length)];
+        }
+
+        return bases;
+    }
```
nh13: Just one test please

src/main/java/htsjdk/utils/NullWriter.java
```diff
@@ -0,0 +1,32 @@
+package htsjdk.utils;
```
nh13: license

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
yfarjoun: Currently FastaSequenceIndex and FastaSequenceIndexCreator are geared towards making an index from a fasta-file, not for generating the index on the fly....so without some serious refactoring that's not going to be easy.

src/main/java/htsjdk/utils/NullWriter.java
nh13: I am not sure it belongs in htsjdk and not in your production code.

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,758 @@
...
+import java.io.*;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Writes a FASTA formatted reference file.
+ * <p>
```
pshapiro4broad: Does this javadoc look correct to you when it's generated? I would change it in a few ways:

- omit `</p>` as it will result in (unnecessary) extra vertical space
- for a code block, wrap with `<pre>..</pre>` with proper indenting

also, `</li>` is unnecessary although it does no harm to include it

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,758 @@
...
+    }
+
+    private static Path defaultDictFile(final boolean makeDictFile, final Path fastaFile) {
+        return makeDictFile ? ReferenceSequenceFileFactory.getDefaultDictionaryForReferenceSequence(fastaFile) : null;
+    }
+
+    // checks that a sequence name is valid.
+    private static void checkSequenceName(final String name) {
+        ValidationUtils.nonNull(name ,"Sequence name should not be null");
+        ValidationUtils.nonEmpty(name,"Sequence name should not be empty");
```
pshapiro4broad: it seems unnecessary to force a caller to use both `nonNull()` and `nonEmpty()` in this case. As a caller, we'd want to assume that `nonEmpty()` will also check for null.

If you really want the user to get different messages for null vs empty, the usual way to solve this problem (validation of inputs with customized messages) is to pass in the name of the thing you're validating, so  the validation code can have more than one failure case.

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,758 @@
...
+            if (!SequenceUtil.isIUPAC(b)) {
+                throw new IllegalArgumentException("the input sequence contains invalid base calls like: " + (char) b);
+            }
+        }
+    }
+
+    private static String checkDescription(final String description) {
+        if (description == null || description.isEmpty()) {
+            return "";
+        } else {
```
pshapiro4broad: unnecessary `else` after `return`

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,758 @@
...
+        checkBasesPerLine(basesPerLine);
+        closeSequence();
+        if (sequenceNamesAndSizes.containsKey(sequenceName)) {
+            throw new IllegalStateException("the input sequence name '" + sequenceName + "' has already been added");
+        }
+        currentSequenceName = sequenceName;
+        currentBasesPerLine = basesPerLine;
+        final StringBuilder builder = new StringBuilder(sequenceName.length() + nonNullDescription.length() + 10);
+        builder.append(HEADER_START_CHAR).append(sequenceName);
+        if (!nonNullDescription.isEmpty()) {
```
pshapiro4broad: This check is unnecessary, if description is empty then using `append()` will have no effect

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,758 @@
...
+    }
+
+    private static String checkDescription(final String description) {
+        if (description == null || description.isEmpty()) {
+            return "";
+        } else {
+            for (int i = 0; i < description.length(); i++) {
+                final char c = description.charAt(i);
+                if (Character.isISOControl(c) && c != '\t') { // tab is the only valid control char in the description.
+                    throw new IllegalArgumentException("the input name contains non-tab control characters: '" +
```
pshapiro4broad: The error message says "the input name" but the code refers to this as "description". Which term will be more helpful for the user? It seems like we'd want the user and code names to agree, if possible.

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,758 @@
...
+    }
+
+    /**
+     * Starts the input of the bases of a new sequence.
+     * <p>
+     * This operation automatically closes the previous sequence base input if any.
+     * </p>
+     * <p>
+     * The sequence name cannot contain any blank characters (as determined by {@link Character#isWhitespace(char)}),
+     * control characters (as determined by {@link Character#isISOControl(char)}) or the the FASTA header star character
```
pshapiro4broad: typo `star character` => `start character`

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,758 @@
...
+    public void close() throws IOException {
+        if (!closed) {
+            closeSequence();
+            if (sequenceNamesAndSizes.isEmpty()) {
+                throw new IllegalStateException("no sequences where added to the reference");
+            }
+            fastaStream.close();
+            indexWriter.close();
+            dictWriter.close();
+            closed = true;
```
pshapiro4broad: Maybe put this in a `finally` in case other `close()` calls above throw? Once we've tried to close ourselves and failed we don't want to try again.

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
pshapiro4broad: Personally I would expect a static method like this to be in the class itself, not a factory method.

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,758 @@
...
+    }
+
+    /**
+     * Convenient method to write a FASTA file with a single sequence.
+     *
+     * @param whereTo     the path to. must not be null.
+     * @param makeIndex   whether the index file should be written at its standard location.
+     * @param makeDict    whether the dictionary file should be written at it standard location.
+     * @param name        the sequence name, cannot contain white space, or control chracter or the header start character.
+     * @param description the sequence description, "" if no description.
```
pshapiro4broad: can be null or "", here and below

src/main/java/htsjdk/samtools/util/SequenceUtil.java
```diff
@@ -1062,4 +1063,14 @@ public static String getSamReadNameFromFastqHeader(final String fastqHeader) {
 
         return readName;
     }
+
+    static public byte[] getRandomBases(Random random, final int length) {
```
pshapiro4broad: javadoc

src/main/java/htsjdk/utils/NullWriter.java
```diff
@@ -0,0 +1,32 @@
...
+    public void flush() throws IOException {
+        // no op
+    }
+
+    @Override
+    public void close() throws IOException {
+        // no op
+    }
+
+    private NullWriter(){super();}
```
pshapiro4broad: need spaces, unnecessary call to `super()`, e.g.
```
private NullWriter() {
}
```

src/main/java/htsjdk/utils/NullWriter.java
```diff
@@ -0,0 +1,32 @@
...
+
+    // the only singlton instance of this class (no need for more!)
```
pshapiro4broad: replace with javadoc

src/main/java/htsjdk/utils/NullWriter.java
pshapiro4broad: Until you have more than one use case in htsjdk, it could live inside the writer class as a static private inner class.

src/main/java/htsjdk/utils/ValidationUtils.java
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * Checks that an {@link Object} is not {@code null} and returns the same object or throws an {@link IllegalArgumentException}
+     * @param object any Object
+     * @param message the text message that would be passed to the exception thrown when {@code o == null}.
+     * @return the same object
+     * @throws IllegalArgumentException if a {@code o == null}
+     */
+    public static <T> T nonNull(final T object, final String message) {
+        if (object == null) {
```
pshapiro4broad: To avoid code duplication, implement as chained call, e.g.
```
return nonNull(object, () -> message);
```

src/main/java/htsjdk/utils/ValidationUtils.java
```diff
@@ -0,0 +1,150 @@
...
+     * Checks that a {@link Collection} is not {@code null} and that it is not empty.
+     * If it's non-null and non-empty it returns the input, otherwise it throws an {@link IllegalArgumentException}
+     * @param collection any Collection
+     * @param message a message to include in the output
+     * @return the original collection
+     * @throws IllegalArgumentException if collection is null or empty
+     */
+    public static <I, T extends Collection<I>> T nonEmpty(T collection, String message){
+        nonNull(collection, "The collection is null: " + message);
+        if(collection.isEmpty()){
```
pshapiro4broad: add spaces before/after `()`

src/main/java/htsjdk/utils/ValidationUtils.java
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * Checks that a {@link Collection} is not {@code null} and that it is not empty.
+     * If it's non-null and non-empty it returns the input, otherwise it throws an {@link IllegalArgumentException}
+     * @param collection any Collection
+     * @param message a message to include in the output
+     * @return the original collection
+     * @throws IllegalArgumentException if collection is null or empty
+     */
+    public static <I, T extends Collection<I>> T nonEmpty(T collection, String message){
```
pshapiro4broad: add space before `{`. many missing spaces in this file, would be easier to re-format in the IDE to fix them all

pshapiro4broad: Is it necessary to parameterize on `I` here? Is `<T extends Collection<?>>` not sufficient?

src/main/java/htsjdk/utils/ValidationUtils.java
```diff
@@ -0,0 +1,150 @@
...
+        nonNull(collection, "The collection is null: " + message);
+        if(collection.isEmpty()){
+            throw new IllegalArgumentException("The collection is empty: " + message);
+        } else {
```
pshapiro4broad: unnecessary `else` after `throw`

src/main/java/htsjdk/utils/ValidationUtils.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+
+    /**
+     * Checks that a {@link Collection} is not {@code null} and that it is not empty.
+     * If it's non-null and non-empty it returns the true
+     * @param collection any Collection
+     * @return true if the collection exists and has elements
+     */
+    public static boolean isNonEmpty(Collection<?> collection){
```
pshapiro4broad: Seems confusing to me to have two different styles of validation here that do pretty much the same thing.

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
yfarjoun: what about the separator?

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
yfarjoun: it's the name...it's the only part of the description that is user supplied...so that's where weird characters could slip in...

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,758 @@
...
+     */
+    public FastaReferenceWriter(final Path fastaFile, final int basesPerLine, final Path indexFile, final Path dictFile)
+            throws IOException {
+        // This code is a slight repeat of {@link #FastaReferenceWriter(OutputStream,int,OutputStream,OutputStream)
+        // for the sake of avoiding creating output if basesPerLine is invalid.
+        this.defaultBasePerLine = checkBasesPerLine(basesPerLine);
+
+        this.fastaStream = new CountingOutputStream(Files.newOutputStream(fastaFile));
+        this.indexWriter = indexFile == null ? NullWriter.NULL_WRITER : new OutputStreamWriter(Files.newOutputStream(indexFile), CHARSET);
+        final BufferedWriter dictWriter = new BufferedWriter(dictFile == null ? NullWriter.NULL_WRITER : new OutputStreamWriter(Files.newOutputStream(dictFile), CHARSET));
```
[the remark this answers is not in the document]
yfarjoun: the issue is that we want to check the basesPerLine before the OutputFile is created...and as the comment (in the other constructor says) this is hard to do without code duplication....

Everything else has to be in the constructor since it's final.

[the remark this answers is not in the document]
yfarjoun: nm. changed the signature and I think it works now.

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
magicDGS: Can you please open an issue and mention it here (or in the factory when the stream is open)? It is easier to track that way

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,758 @@
...
+     * @throws IllegalArgumentException if {@code fastaFile} is {@code null} or {@code basesPerLine} is 0 or negative.
+     */
+    public FastaReferenceWriter(final OutputStream fastaOutput,
+                                final int basesPerLine,
+                                final OutputStream indexOutput,
+                                final OutputStream dictOutput) {
+        this.defaultBasePerLine = checkBasesPerLine(basesPerLine);
+        this.fastaStream = new CountingOutputStream(fastaOutput);
+        this.indexWriter = indexOutput == null ? NullWriter.NULL_WRITER : new OutputStreamWriter(indexOutput, CHARSET);
+        final BufferedWriter dictWriter = new BufferedWriter(dictOutput == null ? NullWriter.NULL_WRITER : new OutputStreamWriter(dictOutput, CHARSET));
```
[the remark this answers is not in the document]
pshapiro4broad: Ah, I didn't notice that

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
pshapiro4broad: You're right, I missed the other append.

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
pshapiro4broad: Looking only at the file here, there is `sequenceName` which is a parameter to `startSequence()` and `name` which is a parameter to `appendSequence()`. Neither of these are used to create `description`, as far as I can tell. But maybe there's some code further upstream that is accepting a `name` argument which it uses to build `description`?

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,758 @@
...
+        checkSequenceName(sequenceName);
+        final String nonNullDescription = checkDescription(description);
+        checkBasesPerLine(basesPerLine);
+        closeSequence();
+        if (sequenceNamesAndSizes.containsKey(sequenceName)) {
+            throw new IllegalStateException("the input sequence name '" + sequenceName + "' has already been added");
+        }
+        currentSequenceName = sequenceName;
+        currentBasesPerLine = basesPerLine;
+        final StringBuilder builder = new StringBuilder(sequenceName.length() + nonNullDescription.length() + 10);
```
pshapiro4broad: It looks like what you really want is `sequenceName.length() + nonNullDescription.length() + 2`. In practice does it make much difference (regarding code performance) to preallocate here?

yfarjoun: I don't know about the performance benefit. I think that the idea is to preallocate if can.. but you are correct about the +2 (I think)

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,674 @@
...
+     *
+     * @param fastaOutput the output fasta file path.
+     * @param indexOutput the output stream to the index file, if requested, {@code null} if none should be generated.
+     * @param dictOutput  the output stream to the dictFile, if requested, {@code null} if none should be generated.
+     * @throws IllegalArgumentException if {@code fastaFile} is {@code null} or {@code basesPerLine} is 0 or negative.
+     */
+    protected FastaReferenceWriter(final int basesPerLine,
+                                final OutputStream fastaOutput,
+                                final OutputStream indexOutput,
+                                final OutputStream dictOutput) {
```
nh13: ident is wrong

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,674 @@
...
+     * <p>
+     * You can specify a specific output stream to each file: the main fasta output, its index and its dictionary.
+     * </p>
+     *
+     * @param fastaOutput the output fasta file path.
+     * @param indexOutput the output stream to the index file, if requested, {@code null} if none should be generated.
+     * @param dictOutput  the output stream to the dictFile, if requested, {@code null} if none should be generated.
+     * @throws IllegalArgumentException if {@code fastaFile} is {@code null} or {@code basesPerLine} is 0 or negative.
+     */
+    protected FastaReferenceWriter(final int basesPerLine,
```
nh13: Cant this be defaulted to `DEFAULT_BASES_PER_LINE`?

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,674 @@
...
+        ValidationUtils.nonEmpty(name, "Sequence name");
+
+        for (int i = 0; i < name.length(); i++) {
+            final char ch = name.charAt(i);
+            if (Character.isWhitespace(ch)) {
+                throw new IllegalArgumentException("the input name contains blank characters: '" + name + "'");
+            } else if (Character.isISOControl(ch)) {
+                throw new IllegalArgumentException("the input name contains control characters: '" + name + "'");
+            }
+        }
```
nh13: If you stored `SAMSequenceRecord`s this would take care of itself!

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,674 @@
...
+    /**
+     * Records the sequences that have been already fully appended to this writer.
+     * <p>
+     * The key is the sequence name.
+     * </p>
+     * <p>
+     * The value is the sequence length in bases.
+     * </p>
+     */
+    private final Map<String, Long> sequenceNamesAndSizes = new LinkedHashMap<>();
```
nh13: Can't you use a `Map<String, SAMSequenceRecord>`?

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
nh13: It'd be great if index/dict were optional.

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,674 @@
...
+     * Adds bases to current sequence from a {@code byte} array.
+     *
+     * @param bases array containing the bases to be added.
+     * @return this instance.
+     * @throws IllegalArgumentException if {@bases} is {@code null} or
+     *                                  the input array contains invalid bases (as assessed by: {@link SequenceUtil#isIUPAC(byte)}).
+     * @throws IllegalStateException    if no sequence was started or the writer is already closed.
+     * @throws IOException              if such exception is throw when writing in any of the outputs.
+     */
+    public FastaReferenceWriter appendBases(final byte[] bases)
```
nh13: It'd be nice to accept a string too

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,674 @@
...
+            throws IOException {
+        return appendBases(bases, 0, bases.length);
+    }
+
+    /**
```
nh13: I've stopped reviewing, because I just don't understand why this class doesn't leverage the various htsjdk classes for it's API.  For example, `ReferenceSequence` and `SAMSequenceRecord`.  I get wanting to feed sub-parts of the reference FASTA at a time, but can't that be handled by giving it an iterator over sub-sequences that's evaluated lazily?

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,674 @@
...
+     */
+    private boolean closed;
+
+    /**
+     * Creates a reference FASTA file writer (private...use the builder: {@link FastaReferenceWriterBuilder}.
+     * <p>
+     * You can specify a specific output stream to each file: the main fasta output, its index and its dictionary.
+     * </p>
+     *
+     * @param fastaOutput the output fasta file path.
```
magicDGS: Can this stream be passed as a bgzip compressed one, and in that case will it work properly? That should be specify here to help developers to refactor to include bgzip-support.

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,674 @@
...
+     * @param indexOutput the output stream to the index file, if requested, {@code null} if none should be generated.
+     * @param dictOutput  the output stream to the dictFile, if requested, {@code null} if none should be generated.
```
magicDGS: Specify that both index/dict streams should not be compressed (the same in the builder).

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
yfarjoun: in the factory it does that.

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
yfarjoun: they are (in builder)

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
yfarjoun: ok.

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
yfarjoun: That sounds nice...but not a simple port. I was looking for something so that folks could write FASTA files with htsjdk.....however, I've realized that the length isn't used...so this is a set of names...I'll make that change.

alecw:
Hi @yfarjoun,
I did a quick test.  Works for me.
Thanks!

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
yfarjoun: yeah, but I can't use that and still get things like description (which SAMSequenceRecord doesn't support) and also incremental buildup of a sequence..

yfarjoun:
@lbergelson back to you.

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,677 @@
...
+     * Adds bases to current sequence from a {@code byte} array.
+     *
+     * @param basesString String containing the bases to be added.
+     * @return this instance.
+     * @throws IllegalArgumentException if {@bases} is {@code null} or
+     *                                  the input array contains invalid bases (as assessed by: {@link SequenceUtil#isIUPAC(byte)}).
+     * @throws IllegalStateException    if no sequence was started or the writer is already closed.
+     * @throws IOException              if such exception is throw when writing in any of the outputs.
+     */
+    public FastaReferenceWriter appendString(final String basesString)
```
yfarjoun: rename appendBases()

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,677 @@
...
+            throws IOException {
+        return appendBases(basesString.getBytes());
```
yfarjoun: specify encoding UTF-8

src/main/java/htsjdk/samtools/reference/FastaReferenceWriter.java
```diff
@@ -0,0 +1,677 @@
...
+    }
+
+    private void writeDictEntry() {
+        dictCodec.encodeSequenceRecord(new SAMSequenceRecord(currentSequenceName, (int) currentBasesCount));
+    }
+
+    /**
+     * Adds bases to current sequence from a {@code byte} array.
+     *
+     * @param basesString String containing the bases to be added.
```
yfarjoun: add: string will be interpreted using ascii and will throw if any character is > 128.

yfarjoun:
merging due to thumb from louis.

PR #1175, opened 2018-08-30
Invalid buffer access in SeekableBufferedStream

tomwhite:
Fix a bug where invalid bytes in BufferedInputStream's buffer were being accessed.

Only bytes up to BufferedInputStream's `count` field in its internal buffer
are valid; this change ensures accesses are limited to this range.

The new test ensures that the internal buffer is not completely filled, and without the fix it fails.

The problem has been present for a long time, but possibly has only been exposed since #1121 in some circumstances. I encountered the bug while running the GATK test suite with Disq (which uses SeekableBufferedStream extensively) - see https://github.com/broadinstitute/gatk/pull/5138.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/seekablestream/SeekableBufferedStream.java
+++ b/src/main/java/htsjdk/samtools/seekablestream/SeekableBufferedStream.java
@@ -43,19 +43,19 @@ public class SeekableBufferedStream extends SeekableStream {
 
         /** Returns the number of bytes that can be read from the buffer without reading more into the buffer. */
         int getBytesInBufferAvailable() {
-            if (this.count == this.pos) return 0; // documented test for "is buffer empty"
-            else return this.buf.length - this.pos;
+            return this.count - this.pos;
         }
 
         /** Return true if the position can be changed by the given delta and remain in the buffer. */
         boolean canChangePos(long delta) {
-            return this.pos + delta >= 0 && delta < getBytesInBufferAvailable();
+            long newPos = this.pos + delta;
+            return newPos >= 0 && newPos < this.count;
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/seekablestream/SeekableBufferedStream.java
```diff
@@ -43,19 +43,19 @@ private ExtBufferedInputStream(final InputStream inputStream, final int i) {
...
-            return this.pos + delta >= 0 && delta < getBytesInBufferAvailable();
+            long newPos = this.pos + delta;
+            return newPos >= 0 && newPos < this.count;
         }
 
         /** Changes the position in the buffer by a given delta. */
         void changePos(int delta) {
             int newPos = this.pos + delta;
-            if (newPos < 0 || newPos > this.buf.length) {
+            if (newPos < 0 || newPos >= this.count) {
```
cmnbroad: This is the essential part of the fix right ? The other changes are just code simplification ?

src/main/java/htsjdk/samtools/seekablestream/SeekableBufferedStream.java
```diff
@@ -43,19 +43,19 @@ private ExtBufferedInputStream(final InputStream inputStream, final int i) {
 
         /** Returns the number of bytes that can be read from the buffer without reading more into the buffer. */
         int getBytesInBufferAvailable() {
-            if (this.count == this.pos) return 0; // documented test for "is buffer empty"
-            else return this.buf.length - this.pos;
+            return this.count - this.pos;
```
tomwhite: @cmnbroad this is the essential part.

PR #1176, opened 2018-08-30
Expose VCFCodec version and header fields.

tomwhite:
### Description

This change adds getters for the version and header fields in AbstractVCFCodec. Access to these fields is needed by Disq, see https://github.com/broadinstitute/gatk/pull/5138.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
+++ b/src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
@@ -228,7 +228,22 @@ public abstract class AbstractVCFCodec extends AsciiFeatureCodec<VariantContext>
         return this.header;
     }
 
-	/**
+    /**
+     * @return the header that was either explicitly set on this codec, or read from the file. May be null.
+     * The returned value should not be modified.
+     */
+    public VCFHeader getHeader() {
+        return header;
+    }
+
+    /**
+     * @return the version number that was either explicitly set on this codec, or read from the file. May be null.
+     */
+    public VCFHeaderVersion getVersion() {
+        return version;
+    }
+
+    /**
 	 * Explicitly set the VCFHeader on this codec. This will overwrite the header read from the file
 	 * and the version state stored in this instance; conversely, reading the header from a file will
 	 * overwrite whatever is set here. The returned header may not be identical to the header argument


tomwhite:
@lbergelson could you take a look please?

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -228,7 +228,21 @@ protected VCFHeader parseHeaderFromLines( final List<String> headerStrings, fina
         return this.header;
     }
 
-	/**
+    /**
+     * @return the header that was either explicitly set on this codec, or read from the file.
+     */
+    public VCFHeader getHeader() {
```
magicDGS: What will happen if the header is modified out of the codec? Maybe it is better to return a copy of the header...

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -228,7 +228,21 @@ protected VCFHeader parseHeaderFromLines( final List<String> headerStrings, fina
...
+        return header;
+    }
+
+    /**
+     * @return the version number that was either explicitly set on this codec, or read from the file.
+     */
+    public VCFHeaderVersion getVersion() {
```
magicDGS: Same as before.

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
pshapiro4broad: I'm not sure if such caution is warrented as there's an api `setVCFHeader(VCFHeader, VCFHeaderVersion)` below which can overwrite these fields completely.

magicDGS: I think that there is a difference in your use case: calling `setHeader` does not modify objects returned previously by the method; but if the header returned by `getHeader` is modified downstream, it affects the codec itself and objects retrieved previously. If I understand correctly the purpose of this PR, which is use in combination with disq, concurrent modification is possible. I think that either return a copy of the header (or document that the method is unsafe is needed, although I prefer the other option).

tomwhite: Do the codecs mutate the header? I tried out the change here and it solved the problem: https://github.com/tomwhite/disq/commit/33622668f576f72aa50f4464734b4ccf68a617d9

cmnbroad: I see - thx. I think the codecs currently only do so to repair version inconsistencies (during parsing the header itself, and in setHeader) if `doOnTheFlyModifications` is on. So I don't see any problematic code path in your disq commit as it stands.

lbergelson: @tomwhite Can you add a comment saying `Don't modify the returned value` and also that it can return null.

tomwhite: @lbergelson Done. Thanks.

PR #1177, opened 2018-09-05
restore no-args VariantContextWriterBuilder.build()

lbergelson:
* VariantContextWriterBuilder.build() was replaced by a varargs version of itself in a previous commit
  While this is a source compatible change it's not binary compatible and since it's a commonly used function it breaks a lot of libaries
* add a parameterless build() which just calls through to the varargs version with an empty array


lbergelson:
@cmnbroad What are your thoughts on this?

lbergelson:
@cmnbroad good catch

PR #1182, opened 2018-09-25
fixing bug in SeekableHttpStream.read

### LINKED ISSUE #1181: Indefinite loop caused by condition in SeekableHTTPStream
### Subject of the issue
Hi! I believe there's a bug in `SeekableHTTPStream`, but I'm not completely sure - at least there's some strange and unexpected (for me) behavior. 

When I try to read `SeekableHTTPStream` with Apache Commons IO (`IOUtils#copyLarge`), it hangs with an indefinite loop... I debugged what is happening under the hood and it appeared that there's a condition in `read` method of `SeekableHTTPStream`:
```java
        if (len == 0 || position == contentLength) {
            return 0;
        }
```
I'm convinced that this condition actually violates the contract of the `read` method of an `InputStream`: when we have no more bytes to read we should return `-1` as a result, signalizing the reader of the `EOF` event. But if we return `0`, like above, the reader will never know that we reached the end of the stream and will indefinitely try to read further. At least this is my understanding of the problem: please, correct me if I'm wrong.

### Your environment
* htsjdk version: 2.16.1
* Java version: 1.8
* OS: MacOS

### Steps to reproduce
```java
SeekableStream inputStream = new SeekableHTTPStream(new URL("someFileURL));
OutputStream outputStream = ... //some output stream
IOUtils.copyLarge(inputStream, outputStream);
```

### Expected behaviour
The input stream is copied to the output stream and the execution finishes successfully.

### Actual behaviour
The input stream is copied to the output stream, but in the end, execution ends up with an endless loop.

lbergelson:
* a bug in `SeekableHttpStream` could cause clients to go into an infinite loop because read would return 0 when trying to read one byte past the end of the file
* fixes #1181

### DIFF
--- a/src/test/java/htsjdk/samtools/seekablestream/SeekableBufferedStreamTest.java
+++ b/src/test/java/htsjdk/samtools/seekablestream/SeekableBufferedStreamTest.java
@@ -31,6 +31,7 @@ import org.testng.annotations.Test;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.net.MalformedURLException;
 import java.net.URL;
 
 import static org.testng.Assert.assertEquals;
@@ -68,6 +69,20 @@ public class SeekableBufferedStreamTest extends HtsjdkTest {
 
         assertEquals(buffer1, buffer2);
     }
+
+    @Test
+    public void testReadExactlyOneByteAtEndOfFile() throws IOException {
+        try( SeekableStream stream = new SeekableHTTPStream(new URL(BAM_URL_STRING))){
+            byte[] buff = new byte[1];
+            long length = stream.length();
+            stream.seek(length - 1);
+            Assert.assertFalse(stream.eof());
+            Assert.assertEquals(stream.read(buff), 1);
+            Assert.assertTrue(stream.eof());
+            Assert.assertEquals(stream.read(buff), -1);
+        }
+    }
+
 
     /**
      * Test an attempt to read past the end of the file.  The test file is 594,149 bytes in length.  The test
--- a/src/test/java/htsjdk/samtools/seekablestream/SeekableBufferedStreamTest.java
+++ b/src/test/java/htsjdk/samtools/seekablestream/SeekableBufferedStreamTest.java
@@ -31,7 +31,6 @@ import org.testng.annotations.Test;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


lbergelson:
@dtitov I think this should fix the problem you reported.

dtitov:
Thanks for the quick fix! I tested it and it works 👍

src/test/java/htsjdk/samtools/seekablestream/SeekableBufferedStreamTest.java
```diff
@@ -31,6 +31,7 @@
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.net.MalformedURLException;
```
pshapiro4broad: unnecessary import?

src/test/java/htsjdk/samtools/seekablestream/SeekableBufferedStreamTest.java
```diff
@@ -69,6 +70,20 @@ public void testRandomRead() throws IOException {
         assertEquals(buffer1, buffer2);
     }
 
+    @Test
+    public void testReadExactlyOneByteAtEndOfFile() throws IOException {
+        try( SeekableStream stream = new SeekableHTTPStream(new URL(BAM_URL_STRING))){
```
pshapiro4broad: odd whitespace here, I would expect to see `try (SeekableStream stream = new SeekableHTTPStream(new URL(BAM_URL_STRING))) {`

lbergelson: updated

src/test/java/htsjdk/samtools/seekablestream/SeekableBufferedStreamTest.java
```diff
@@ -69,6 +69,20 @@ public void testRandomRead() throws IOException {
         assertEquals(buffer1, buffer2);
     }
 
+    @Test
+    public void testReadExactlyOneByteAtEndOfFile() throws IOException {
+        try (final SeekableStream stream = new SeekableHTTPStream(new URL(BAM_URL_STRING))){
```
pshapiro4broad: almost, still missing the space between the `){`

lbergelson: whoops, didn't see your comment until after I merged

PR #1185, opened 2018-10-04
Add 4 CRAM compliance tests from htslib

jmthibault79:
### Description

Looking at htslib CRAM round trip, I noticed 6 that were missing from CramComplianceTest.  These 4 pass with Partial Verification but need further investigation.  Two others `c2#pad` and `md#1` are not included here because they do not pass Partial Verification.

CRAM files generated using `samtools view`

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


jmthibault79:
`c2#pad` passes fully with regenerated CRAM files, so I added it here.

Added links to issues to the 5 new failing tests, including `md#1`

PR #1188, opened 2018-10-05
BinaryCodec.readBytesOrFewer fails if no data is requested AND no date is available in the underlying input stream

nh13:
I ran into this bug when writing an empty string as the last element into a byte buffer (ex `binaryCodec.writeString("", true, false)`) and then trying to read it back in (ex. `binaryCodec.readLengthAndString(false)`).  The main issue is that `InputStream.read` will return `-1` even if the maximum number of bytes to read is zero (i.e. `InputStream.read(buffer, 0, 0)` will return `-1`).

I made a commit with a test to show the bug and a commit to fix it.

### DIFF
--- a/src/main/java/htsjdk/samtools/util/BinaryCodec.java
+++ b/src/main/java/htsjdk/samtools/util/BinaryCodec.java
@@ -415,7 +415,8 @@ public class BinaryCodec implements Closeable {
             throw new IllegalStateException("Calling read method on BinaryCodec open for write.");
         }
         try {
-            return inputStream.read(buffer, offset, length);
+            if (length == 0) return 0;
+            else return inputStream.read(buffer, offset, length);
         } catch (IOException e) {
             throw new RuntimeIOException(constructErrorMessage("Read error"), e);
         }
--- a/src/main/java/htsjdk/samtools/util/BinaryCodec.java
+++ b/src/main/java/htsjdk/samtools/util/BinaryCodec.java
@@ -415,8 +415,10 @@ public class BinaryCodec implements Closeable {
             throw new IllegalStateException("Calling read method on BinaryCodec open for write.");
         }
         try {
-            if (length == 0) return 0;
-            else return inputStream.read(buffer, offset, length);
+            // Some implementations of InputStream do not behave well when the buffer is empty and length is zero, for
+            // example ByteArrayInputStream, so we must check for length equal to zero.
+            // See: https://bugs.java.com/view_bug.do?bug_id=6766844
+            return (length == 0) ? 0 : inputStream.read(buffer, offset, length);
         } catch (IOException e) {
             throw new RuntimeIOException(constructErrorMessage("Read error"), e);
         }
--- a/src/test/java/htsjdk/samtools/util/BinaryCodecTest.java
+++ b/src/test/java/htsjdk/samtools/util/BinaryCodecTest.java
@@ -27,12 +27,7 @@ import htsjdk.HtsjdkTest;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
-import java.io.IOException;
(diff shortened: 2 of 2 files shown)


lbergelson:
Strange, input stream documentation says it should return 0 in that case.  


>    If <code>len</code> is zero, then no bytes are read and
>       <code>0</code> is returned; otherwise, there is an attempt to read at
>       least one byte. If no byte is available because the stream is at end of
>      file, the value <code>-1</code> is returned; otherwise, at least one
>       byte is read and stored into <code>b</code>.

lbergelson:
Oh, I didn't read carefully enough, it's the combination of the end of stream AND requesting to read nothing that is the issue?

nh13:
@lbergelson 

From [`ByteArrayInputStream`](https://docs.oracle.com/javase/8/docs/api/?java/io/ByteArrayInputStream.html)

> Reads up to len bytes of data into an array of bytes from this input stream. If pos equals count, then -1 is returned to indicate end of file.

which doesn't agree with [`InputStream](https://docs.oracle.com/javase/8/docs/api/?java/io/InputStream.html).

src/main/java/htsjdk/samtools/util/BinaryCodec.java
```diff
@@ -415,7 +415,8 @@ public int readBytesOrFewer(final byte[] buffer, final int offset, final int len
             throw new IllegalStateException("Calling read method on BinaryCodec open for write.");
         }
         try {
-            return inputStream.read(buffer, offset, length);
+            if (length == 0) return 0;
```
pshapiro4broad: use `{` `}` around `if` body

src/main/java/htsjdk/samtools/util/BinaryCodec.java
```diff
@@ -415,7 +415,8 @@ public int readBytesOrFewer(final byte[] buffer, final int offset, final int len
             throw new IllegalStateException("Calling read method on BinaryCodec open for write.");
         }
         try {
-            return inputStream.read(buffer, offset, length);
+            if (length == 0) return 0;
+            else return inputStream.read(buffer, offset, length);
```
pshapiro4broad: `else` unnecessary after `return`

nh13:
My interpretation is that `InpuStream.read` should return zero, and that `ByteArrayInputStream.read` says otherwise for its implementation.

https://bugs.java.com/view_bug.do?bug_id=7073237

src/main/java/htsjdk/samtools/util/BinaryCodec.java
nh13: How about `return (length == 0) ? 0 : inputStream.read(buffer, offset, length);`

pshapiro4broad:
Thanks for the note, I was looking at the source for `InputStream.read()` and was pretty confused by this. Given this behavior of `ByteArrayInputStream` it makes me wonder how many other cases like this there are, if we want to rely on correct results from `read()`.

src/main/java/htsjdk/samtools/util/BinaryCodec.java
```diff
@@ -415,7 +415,10 @@ public int readBytesOrFewer(final byte[] buffer, final int offset, final int len
             throw new IllegalStateException("Calling read method on BinaryCodec open for write.");
         }
         try {
-            return inputStream.read(buffer, offset, length);
+            // Some implementations of InputStream do not behave well when the buffer is empty and length is zero, for
+            // example ByteArrayInputStream, so we must check for length equal to zero.
+            // See: https://bugs.java.com/view_bug.do?bug_id=6766844
```
lbergelson: good find, thanks for adding this note.

lbergelson:
I love the open java bugs from 2008 that no one has bothered responding too.

pshapiro4broad:
On one of the linked bugs there's a comment

> The inputstream.read(byte[], int, 0) is not always expected return 0 in the case of the bug description. We should not try to read zero bytes from inputstream.

PR #1191, opened 2018-10-08
HttpUtils: set Method to HEAD for HttpUtils.getHeaders

lindenb:
### Description

Minor change: in httpUtils `getHeaderField` I set the Http-method to '**HEAD**'  instead of '**GET**' in order to only fetch the header fields.

I also added a test file for HttpUtils.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/HttpUtils.java
+++ b/src/main/java/htsjdk/samtools/util/HttpUtils.java
@@ -29,9 +29,14 @@ public class HttpUtils {
         try {
             // Create a URLConnection object for a URL
             conn = openConnection(url);
+            if(conn instanceof HttpURLConnection) {
+                // The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response.
+                final HttpURLConnection httpConn = HttpURLConnection.class.cast(conn);
+                httpConn.setRequestMethod("HEAD");
+            }
             return conn.getHeaderField(name);
 
-        } catch (IOException e) {
+        } catch (final IOException e) {
             e.printStackTrace();
             return null;
         }
--- a/src/main/java/htsjdk/samtools/util/HttpUtils.java
+++ b/src/main/java/htsjdk/samtools/util/HttpUtils.java
@@ -17,7 +17,7 @@ public class HttpUtils {
     }
 
     private static URLConnection openConnection(final URL url) throws IOException{
-        URLConnection conn = url.openConnection();
+        final URLConnection conn = url.openConnection();
         conn.setReadTimeout(3000);
         conn.setDefaultUseCaches(false);
         conn.setUseCaches(false);
@@ -29,10 +29,9 @@ public class HttpUtils {
         try {
             // Create a URLConnection object for a URL
             conn = openConnection(url);
-            if(conn instanceof HttpURLConnection) {
+            if (conn instanceof HttpURLConnection) {
                 // The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response.
-                final HttpURLConnection httpConn = HttpURLConnection.class.cast(conn);
-                httpConn.setRequestMethod("HEAD");
+                ((HttpURLConnection) conn).setRequestMethod("HEAD");
             }
             return conn.getHeaderField(name);
 
--- a/src/test/java/htsjdk/samtools/util/HttpUtilsTest.java
+++ b/src/test/java/htsjdk/samtools/util/HttpUtilsTest.java
@@ -1,0 +1,33 @@
+package htsjdk.samtools.util;
+import java.io.IOException;
+import java.net.URL;
+
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class HttpUtilsTest {
+    @DataProvider(name = "existing_urls")
+    public Object[][] testExistingURLsData() {
+        return new Object[][]{
+                {"http://broadinstitute.github.io/picard/testdata/index_test.bam"}
+        };
+    }
+    
+    @Test(dataProvider="existing_urls")
+    public void testGetHeaderField(final String url) throws IOException {
+    final String field = HttpUtils.getHeaderField(
+            new URL(url),
+            "Content-Length"
+            );
+    Assert.assertNotNull(field);
+    final long length = Long.parseLong(field);
+    Assert.assertTrue(length>0L);
+    }
+    
+    @Test(dataProvider="existing_urls")
+    public void testGetETag(final String url) throws IOException {
+    final String field = HttpUtils.getETag(new URL(url));
+    Assert.assertNotNull(field);
+    }
+}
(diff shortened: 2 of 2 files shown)


src/main/java/htsjdk/samtools/util/HttpUtils.java
```diff
@@ -29,9 +29,14 @@ public static String getHeaderField(final URL url, final String name) {
         try {
             // Create a URLConnection object for a URL
             conn = openConnection(url);
+            if(conn instanceof HttpURLConnection) {
```
pshapiro4broad: need space after `if`

src/test/java/htsjdk/samtools/util/HttpUtilsTest.java
```diff
@@ -0,0 +1,35 @@
...
+import java.net.URL;
+
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import htsjdk.HtsjdkTest;
+
+public class HttpUtilsTest extends HtsjdkTest {
+    @DataProvider(name = "existing_urls")
```
pshapiro4broad: If you only have a single test case, it would be easier to create a static field of type `URL` than use a data provider.

src/test/java/htsjdk/samtools/util/HttpUtilsTest.java
```diff
@@ -0,0 +1,35 @@
...
+    public Object[][] testExistingURLsData() {
+        return new Object[][]{
+                {"http://broadinstitute.github.io/picard/testdata/index_test.bam"}
+        };
+    }
+    
+    @Test(dataProvider="existing_urls")
+    public void testGetHeaderField(final String url) throws IOException {
+    final String field = HttpUtils.getHeaderField(
```
pshapiro4broad: Indent is wrong here, it would be best to reformat the entire file since there are many lines with incorrect indent.

src/test/java/htsjdk/samtools/util/HttpUtilsTest.java
```diff
@@ -0,0 +1,35 @@
...
+            "Content-Length"
+            );
+    Assert.assertNotNull(field);
+    final long length = Long.parseLong(field);
+    Assert.assertTrue(length>0L);
+    }
+    
+    @Test(dataProvider="existing_urls")
+    public void testGetETag(final String url) throws IOException {
+    final String field = HttpUtils.getETag(new URL(url));
```
pshapiro4broad: Is this a useful API to keep? It's not used in picard or htsjdk or gatk. It would be better to remove/deprecate `getEtag()` and `resourceAvailable` than add tests for them.

lbergelson: You never know who's using something downstream.  I think we're hoping to deprecate the entire http code package in the future once we've tested out an http path replacement, but I'm not sure if it's worth deprecating bits now.

src/main/java/htsjdk/samtools/util/HttpUtils.java
```diff
@@ -29,9 +29,14 @@ public static String getHeaderField(final URL url, final String name) {
...
+                // The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response.
+                final HttpURLConnection httpConn = HttpURLConnection.class.cast(conn);
+                httpConn.setRequestMethod("HEAD");
+            }
             return conn.getHeaderField(name);
 
-        } catch (IOException e) {
+        } catch (final IOException e) {
             e.printStackTrace();
```
lbergelson: We should probably be crashing here instead, but lets leave that for another pr.

src/main/java/htsjdk/samtools/util/HttpUtils.java
```diff
@@ -29,9 +29,14 @@ public static String getHeaderField(final URL url, final String name) {
         try {
             // Create a URLConnection object for a URL
             conn = openConnection(url);
+            if(conn instanceof HttpURLConnection) {
+                // The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response.
+                final HttpURLConnection httpConn = HttpURLConnection.class.cast(conn);
```
lbergelson: Why not just `(HttpUrlConnection) conn`?

lindenb:
@lbergelson 

> Do most major web servers support HEAD as well as they support GET?

citing  John Marshall on twitter: https://twitter.com/jomarnz/status/1049762552304074752

> RFC 7231 is definitive on this and says “All general-purpose servers MUST support the methods GET and HEAD”. Me, I test stuff with curl -I (which selects HEAD) all the time.

lindenb:
@lbergelson 
 
>  is it actually a noticeable improvement? 

a 'stupid test':

```

$ time ( for i in {1..500}; do  curl -s  -I -D - "http://ftp.1000genomes.ebi.ac.uk/vol1/ftp/current.tree"   > /dev/null ; done)

real	0m59.979s
user	0m3.380s
sys	0m2.084s

$ time ( for i in {1..500}; do  curl -s  -D - "http://ftp.1000genomes.ebi.ac.uk/vol1/ftp/current.tree" | grep "ftp" -m1  > /dev/null ; done)

real	1m16.346s
user	0m4.040s
sys	0m3.724s

```

lindenb:
@pshapiro4broad back to you. I kept 'getETag' for another PR

src/test/java/htsjdk/samtools/util/HttpUtilsTest.java
```diff
@@ -0,0 +1,33 @@
...
+    public Object[][] testExistingURLsData() {
+        return new Object[][]{
+            {"http://broadinstitute.github.io/picard/testdata/index_test.bam"},
+            {"http://ftp.1000genomes.ebi.ac.uk/vol1/ftp/current.tree"}
+        };
+    }
+
+    @Test(dataProvider="existing_urls")
+    public void testGetHeaderField(final String url) throws IOException {
+        final String field = HttpUtils.getHeaderField(new URL(url), "Content-Length");
```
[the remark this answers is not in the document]
yfarjoun: I don't think that it will test the right thing if we use HTTPHelper....

src/test/java/htsjdk/samtools/util/HttpUtilsTest.java
```diff
@@ -0,0 +1,33 @@
...
+        Assert.assertNotNull(field);
+        final long length = Long.parseLong(field);
+        Assert.assertTrue(length > 0L);
```
yfarjoun: can we assert the actual length or is that not stable?

src/test/java/htsjdk/samtools/util/HttpUtilsTest.java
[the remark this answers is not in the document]
lbergelson: Since this test is testing the call to getHeaderField it seems like it would be problematic to replace it with a different method.  You could certainly make a case for merging the two methods but I'm not sure that's necessarily in the scope of this.

PR #1193, opened 2018-10-10
fix CRAMIterator when next() is called without hasNext()

lbergelson:
* fixing a bug in CRAMIterator which previously threw if next() was called without first calling hasNext()
* adding CRAMTestUtils and extracting a method to it from CramEdgeCasesTest

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/CRAMIterator.java
+++ b/src/main/java/htsjdk/samtools/CRAMIterator.java
@@ -266,7 +266,7 @@ public class CRAMIterator implements SAMRecordIterator {
 
     @Override
     public SAMRecord next() {
-        if(hasNext()) {
+        if (hasNext()) {
             return iterator.next();
         } else {
             throw new NoSuchElementException();
--- a/src/main/java/htsjdk/samtools/CRAMIterator.java
+++ b/src/main/java/htsjdk/samtools/CRAMIterator.java
@@ -34,10 +34,7 @@ import htsjdk.samtools.util.Log;
 import java.io.IOException;
 import java.io.InputStream;
 import java.math.BigInteger;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
+import java.util.*;
 
 import htsjdk.samtools.cram.CRAMException;
 
@@ -259,9 +256,7 @@ public class CRAMIterator implements SAMRecordIterator {
         if (!iterator.hasNext()) {
             try {
                 nextContainer();
-            } catch (IOException e) {
-                throw new SAMException(e);
-            } catch (IllegalAccessException e) {
+            } catch (IOException | IllegalAccessException e) {
                 throw new SAMException(e);
             }
         }
@@ -271,7 +266,11 @@ public class CRAMIterator implements SAMRecordIterator {
 
     @Override
     public SAMRecord next() {
-        return iterator.next();
+        if(hasNext()) {
+            return iterator.next();
+        } else {
+            throw new NoSuchElementException();
+        }
     }
 
     @Override
--- a/src/test/java/htsjdk/samtools/CRAMFileReaderTest.java
+++ b/src/test/java/htsjdk/samtools/CRAMFileReaderTest.java
@@ -37,6 +37,7 @@ import java.io.InputStream;
 import java.io.File;
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.NoSuchElementException;
 
 /**
  * Additional tests for CRAMFileReader are in CRAMFileIndexTest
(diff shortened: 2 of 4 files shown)


src/main/java/htsjdk/samtools/CRAMIterator.java
```diff
@@ -271,7 +266,11 @@ public boolean hasNext() {
 
     @Override
     public SAMRecord next() {
-        return iterator.next();
+        if(hasNext()) {
```
pshapiro4broad: missing space after `if`

src/test/java/htsjdk/samtools/CRAMTestUtils.java
```diff
@@ -0,0 +1,64 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Iterator;
+
+public final class CRAMTestUtils {
+
+    private CRAMTestUtils(){};
```
pshapiro4broad: I understand that there's no point in instantiating this class, but why bother enforcing it with a private constructor?

src/test/java/htsjdk/samtools/CRAMTestUtils.java
```diff
@@ -0,0 +1,64 @@
...
+     * @param records a SAMRecordSetBuilder which has been initialized with records
+     * @return a CRAMFileReader reading from an in memory buffer that has had the records written into it, uses a fake reference with all A's
+     */
+    public static CRAMFileReader writeAndReadFromInMemoryCram(SAMRecordSetBuilder records) throws IOException {
+        return writeAndReadFromInMemoryCram(records.getRecords(), getFakeReferenceSource(), records.getHeader());
+    }
+
+    private static CRAMFileReader writeAndReadFromInMemoryCram(Collection<SAMRecord> records, CRAMReferenceSource source, SAMFileHeader header) throws IOException {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        CRAMFileWriter cramFileWriter = new CRAMFileWriter(baos, source, header, "whatever");
```
pshapiro4broad: Use try-with-resources to avoid `close()` below. Since `baos` is closable too you could put that in the `try` as well.

src/test/java/htsjdk/samtools/CRAMTestUtils.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+        cramFileWriter.close();
+
+        return new CRAMFileReader(new ByteArrayInputStream(baos.toByteArray()), (SeekableStream) null, source, ValidationStringency.SILENT);
+    }
+
+    /**
+     * return a CRAMReferenceSource that returns all A's for any sequence queried
+     */
+    public static CRAMReferenceSource getFakeReferenceSource(){
```
pshapiro4broad: missing space between `){`

src/test/java/htsjdk/samtools/CRAMFileReaderTest.java
```diff
@@ -229,4 +230,14 @@ public void testCRAMReader7_WithoutCRAMIndex() throws IOException {
...
     }
+
+    @Test
+    public void testCramIteratorWithoutCallingHasNextFirst() throws IOException {
+        final SAMRecordSetBuilder builder = new SAMRecordSetBuilder(false, SAMFileHeader.SortOrder.unsorted);
+        builder.addFrag("1", 0, 2, false);
+        final CRAMFileReader reader = CRAMTestUtils.writeAndReadFromInMemoryCram(builder);
+        final SAMRecordIterator iterator = reader.getIterator();
+        Assert.assertNotNull(iterator.next());
+        Assert.assertThrows(NoSuchElementException.class, iterator::next);
```
cmnbroad: That two tests, but, ok...

src/test/java/htsjdk/samtools/CRAMTestUtils.java
cmnbroad: It does make it clear that there is no point in instantiating the class, which I like.

src/test/java/htsjdk/samtools/CRAMTestUtils.java
lbergelson: I stuck it in a try with resources so that it will always close on exit, but I left the close to force a flush before reading from the buffer, the alternative was a gross nested try statement.

src/test/java/htsjdk/samtools/CRAMTestUtils.java
pshapiro4broad: Unlike a singleton, there's no harm if a user decides to instantiate it. They can't break anything or change behavior that we'd like to control. So I'd still say there is no need to bother enforcing it for this class.

PR #1197, opened 2018-10-11
Google java code style is now used on compile

nh13:


pshapiro4broad:
I didn't realize google java style is 2 space indent instead of 4. That's a shame, as it doesn't appear to be configurable either.

lbergelson:
We realized it's possible to use the AOSP style with this plugin instead of google which uses 4 spaces instead of 2.  It makes the diffs much smaller because it doesn't change every single line of htsjdk.  Unfortunately the things it does change generally seem to be for the worst.

It makes some very weird decisions about how to chop down long lines and ends up with strange things like these:

```
-    protected static void writeHeader(final BinaryCodec outputBinaryCodec, final SAMFileHeader samFileHeader) {
-        // Do not use SAMFileHeader.getTextHeader() as it is not updated when changes to the underlying object are made
+    protected static void writeHeader(
+            final BinaryCodec outputBinaryCodec, final SAMFileHeader samFileHeader) {
+        // Do not use SAMFileHeader.getTextHeader() as it is not updated when changes to the
+        // underlying
+        // object are made
```
and

```
-        // Initially set each window to -1 so we can distinguish between windows that have no overlapping
-        // features, and those whose lowest offset is 0, which is a valid (virtual file) offset for feature
+        // Initially set each window to -1 so we can distinguish between windows that have no
+        // overlapping
+        // features, and those whose lowest offset is 0, which is a valid (virtual file) offset for
+        // feature
```

and
```
-            if (read != length) throw new RuntimeIOException("Expected to read " + length + " bytes, but expired stream after " + read + ".");
+            if (read != length)
+                throw new RuntimeIOException(
+                        "Expected to read "
+                                + length
+                                + " bytes, but expired stream after "
+                                + read
+                                + ".");
```
It breaks formatting on some specially formatted things like this:
```
-        for (k =    1 + (start>>26); k <=    1 + (end>>26); ++k) bitSet.set(k);
-        for (k =    9 + (start>>23); k <=    9 + (end>>23); ++k) bitSet.set(k);
-        for (k =   73 + (start>>20); k <=   73 + (end>>20); ++k) bitSet.set(k);
-        for (k =  585 + (start>>17); k <=  585 + (end>>17); ++k) bitSet.set(k);
-        for (k = 4681 + (start>>14); k <= 4681 + (end>>14); ++k) bitSet.set(k);
+        for (k = 1 + (start >> 26); k <= 1 + (end >> 26); ++k) bitSet.set(k);
+        for (k = 9 + (start >> 23); k <= 9 + (end >> 23); ++k) bitSet.set(k);
+        for (k = 73 + (start >> 20); k <= 73 + (end >> 20); ++k) bitSet.set(k);
+        for (k = 585 + (start >> 17); k <= 585 + (end >> 17); ++k) bitSet.set(k);
+        for (k = 4681 + (start >> 14); k <= 4681 + (end >> 14); ++k) bitSet.set(k);
```

I like the idea of having auto linting of some type.  I'm not convinced that this particular linter is an improvement to the codebase though...

pshapiro4broad:
I noticed the wrapping for comments and string concats, which also seemed odd to me. I thought the google convention was for 100 character lines.

I think checkstyle might be a better solution; although it would force users to manually fix things, it also allows for a lot more configurability. Not sure about the issue with preserving whitespace in the `for` example, but it can be configured to avoid wrapping comments and string concat operations. With checkstyle the build will fail, which means that a PR would be blocked by the build failure. Users could also set up git commit hooks to run checkstyle if they wanted to.

lbergelson:
@nh13 This was a good idea, but I think until we find a formatter that does less weird things I think we can't merge it.

PR #1198, opened 2018-10-16
Unit Tests and fixes for a few classes in samtools.cram.io

jmthibault79:
### Description

Increasing test coverage for the samtools.cram.io package, and fixes to bugs found by these tests

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/cram/io/CramInt.java
+++ b/src/main/java/htsjdk/samtools/cram/io/CramInt.java
@@ -40,7 +40,10 @@ public class CramInt {
      * @return an integer value read from the buffer
      */
     public static int int32(final ByteBuffer buffer) {
-        return buffer.get() | buffer.get() << 8 | buffer.get() << 16 | buffer.get() << 24;
+        return (0xFF & (int) buffer.get()) |
+                (0xFF & (int) buffer.get()) << 8 |
+                (0xFF & (int) buffer.get()) << 16 |
+                (0xFF & (int) buffer.get()) << 24;
     }
 
     /**
@@ -61,7 +64,7 @@ public class CramInt {
     }
 
     /**
-     * Write int value to {@link OutputStream} encoded as CRAM int data type.
+     * Write int value to an array of bytes encoded as CRAM int data type.
      *
      * @param value value to be written out
      * @return the byte array holding the value encoded as CRAM int data type
@@ -74,5 +77,4 @@ public class CramInt {
         data[3] = (byte) (value >> 24 & 0xFF);
         return data;
     }
-
 }
--- a/src/main/java/htsjdk/samtools/cram/io/CramIntArray.java
+++ b/src/main/java/htsjdk/samtools/cram/io/CramIntArray.java
@@ -7,7 +7,7 @@ import java.io.OutputStream;
 /**
  * Methods to read and write CRAM array of integers data type.
  */
-public class CramArray {
+public class CramIntArray {
     /**
      * Read CRAM int array from a {@link InputStream}.
      *
--- a/src/main/java/htsjdk/samtools/cram/io/CramLong.java
+++ b/src/main/java/htsjdk/samtools/cram/io/CramLong.java
@@ -1,0 +1,107 @@
+package htsjdk.samtools.cram.io;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+
+/**
+ * Methods to read and write CRAM long values as given in the file format specification.
+ */
+
+public class CramLong {
+    /**
+     * Read unsigned little-endian 8 byte integer from an {@link InputStream}.
+     *
+     * @param inputStream input stream to read from
+     * @return a long value read
+     * @throws IOException as per java IO contract
+     */
+    public static long int64(final InputStream inputStream) throws IOException {
+        return (long) inputStream.read() |
+                (long) inputStream.read() << 8 |
+                (long) inputStream.read() << 16 |
+                (long) inputStream.read() << 24 |
+                (long) inputStream.read() << 32 |
+                (long) inputStream.read() << 40 |
+                (long) inputStream.read() << 48 |
+                (long) inputStream.read() << 56;
+    }
+
+    /**
+     * Read unsigned little-endian 8 byte integer from an array of bytes.
+     *
+     * @param data input stream to read from
+     * @return a long value read
+     */
+    public static long int64(final byte[] data) {
+        if (data.length != 8)
+            throw new IllegalArgumentException("Expecting an 8-byte integer. ");
+        return (long) (0xFF & data[0]) |
+                ((long) (0xFF & data[1]) << 8) |
+                ((long) (0xFF & data[2]) << 16) |
+                ((long) (0xFF & data[3]) << 24) |
+                ((long) (0xFF & data[4]) << 32) |
+                ((long) (0xFF & data[5]) << 40) |
+                ((long) (0xFF & data[6]) << 48) |
+                ((long) (0xFF & data[7]) << 56);
+    }
+
+    /**
+     * Read unsigned little-endian 8 byte integer from an {@link ByteBuffer}.
+     *
+     * @param buffer {@link ByteBuffer} to read from
+     * @return a long value read from the buffer
+     */
+    public static long int64(final ByteBuffer buffer) {
+        return (0xFF & (long) buffer.get()) |
+                (0xFF & (long) buffer.get()) << 8 |
+                (0xFF & (long) buffer.get()) << 16 |
+                (0xFF & (long) buffer.get()) << 24 |
+                (0xFF & (long) buffer.get()) << 32 |
+                (0xFF & (long) buffer.get()) << 40 |
+                (0xFF & (long) buffer.get()) << 48 |
+                (0xFF & (long) buffer.get()) << 56;
+    }
+
+    /**
+     * Write int value to {@link OutputStream} encoded as CRAM int64 data type.
+     *
+     * @param value value to be written out
+     * @param outputStream    the output stream
+     * @return the number of bits written out
+     * @throws IOException as per java IO contract
+     */
+    @SuppressWarnings("SameReturnValue")
+    public static long writeInt64(final long value, final OutputStream outputStream) throws IOException {
+        outputStream.write((byte) value);
+        outputStream.write((byte) (value >> 8));
+        outputStream.write((byte) (value >> 16));
+        outputStream.write((byte) (value >> 24));
+        outputStream.write((byte) (value >> 32));
+        outputStream.write((byte) (value >> 40));
+        outputStream.write((byte) (value >> 48));
+        outputStream.write((byte) (value >> 56));
+        return 8 * 8;
+    }
+
+    /**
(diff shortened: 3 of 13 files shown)


src/main/java/htsjdk/samtools/cram/io/CramInt.java
```diff
@@ -40,7 +40,10 @@ public static int int32(final byte[] data) {
      * @return an integer value read from the buffer
      */
     public static int int32(final ByteBuffer buffer) {
-        return buffer.get() | buffer.get() << 8 | buffer.get() << 16 | buffer.get() << 24;
```
jmthibault79: These shifts were being attempted in a `byte` context

src/main/java/htsjdk/samtools/cram/io/CramIntArray.java
```diff
@@ -7,7 +7,7 @@
 /**
  * Methods to read and write CRAM array of integers data type.
  */
-public class CramArray {
+public class CramIntArray {
```
jmthibault79: Renamed because this gave me the impression of a generic array type, though it's limited to integer.  I can revert this if you prefer.

src/main/java/htsjdk/samtools/cram/io/CramLong.java
```diff
@@ -0,0 +1,107 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+
+/**
+ * Methods to read and write CRAM long values as given in the file format specification.
+ */
+
+public class CramLong {
```
jmthibault79: All methods correspond to those in `CramInt`

src/test/java/htsjdk/cram/io/ExternalCompressionTest.java
```diff
@@ -1,29 +0,0 @@
-package htsjdk.cram.io;
```
jmthibault79: moved from `htsjdk.cram.io` to `htsjdk.samtools.cram.io` like the class it's testing

src/test/java/htsjdk/samtools/cram/io/CramIntTest.java
```diff
@@ -0,0 +1,55 @@
...
+            CramInt.writeInt32(value, baos);
+        }
+
+        byte[] bytes = baos.toByteArray();
+
+        ByteArrayInputStream bais = new ByteArrayInputStream(bytes);
+        ByteBuffer bb = ByteBuffer.wrap(bytes);
+
+        // a manually-advanced byte buffer
+        final int manualBufferSize = 4;
```
jmthibault79: To exercise the read from `byte[]` method

src/test/java/htsjdk/samtools/cram/io/IOTestCases.java
```diff
@@ -0,0 +1,116 @@
...
+public class IOTestCases extends HtsjdkTest {
+
+    private static <T> Object[][] asDataProvider(List<T> list) {
+        Object[][] params = new Object[list.size()][] ;
+        for (int i=0; i<params.length; i++)
+            params[i] = new Object[]{list.get(i)} ;
+        return params;
+    }
+
+    static List<Integer> int32Tests() {
```
jmthibault79: moved from `ITF8Test` for reuse

src/test/java/htsjdk/samtools/cram/io/IOTestCases.java
```diff
@@ -0,0 +1,116 @@
...
+        List<Integer> shuffled = new ArrayList<>(int32Tests);
+        Collections.shuffle(shuffled);
+
+        return new Object[][]{
+                {int32Tests},
+                {shuffled}
+        };
+    }
+
+    static List<Long> int64Tests() {
```
jmthibault79: moved from `LTF8Test` for reuse

src/test/java/htsjdk/samtools/cram/io/IOTestCases.java
```diff
@@ -0,0 +1,116 @@
...
+        for (int i = 0; i <= 64; i++) {
+            list.add((1L << i) - 2);
+            list.add((1L << i) - 1);
+            list.add(1L << i);
+            list.add((1L << i) + 1);
+            list.add((1L << i) + 1);
+        }
+
+        // special cases:
+        list.add(Long.MAX_VALUE);
```
jmthibault79: these two were not in the original test cases

src/main/java/htsjdk/samtools/cram/io/CramInt.java
```diff
@@ -40,7 +40,10 @@ public static int int32(final byte[] data) {
...
+        return (0xFF & (int) buffer.get()) |
+                (0xFF & (int) buffer.get()) << 8 |
+                (0xFF & (int) buffer.get()) << 16 |
+                (0xFF & (int) buffer.get()) << 24;
```
cmnbroad: While we're in this class I'd be in favor of renaming the `int32` overloads in  to `readInt32` so they're symmetric with the `writeInt32` ones. WDYT ?

src/main/java/htsjdk/samtools/cram/io/CramLong.java
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Read unsigned little-endian 8 byte integer from an {@link InputStream}.
+     *
+     * @param inputStream input stream to read from
+     * @return a long value read
+     * @throws IOException as per java IO contract
+     */
+    public static long int64(final InputStream inputStream) throws IOException {
```
cmnbroad: Same naming comment here if we keep these.

src/test/java/htsjdk/samtools/cram/io/CramIntTest.java
```diff
@@ -0,0 +1,55 @@
...
+            Assert.assertEquals(fromManualBuffer, value, "Value did not match");
+            manualBufferOffset += manualBufferSize;
+
+            manualWriteBuffer = CramInt.writeInt32(value);
+            int fromManualWriteBuffer = CramInt.int32(manualWriteBuffer);
+            Assert.assertEquals(fromManualWriteBuffer, value, "Value did not match");
+        }
+
+        baos.close();
+        bais.close();
```
cmnbroad: We should probably add something to verify that these are really respecting little endianness. 

I'd prefer to see this factored so there is one test method per method (or one test method per read/write overload/pair), rather than mixing them all up, with common code factored out as appropriate.

src/test/java/htsjdk/samtools/cram/io/IOTestCases.java
cmnbroad: Nice consolidation!

src/main/java/htsjdk/samtools/cram/io/CramLong.java
jmthibault79: Looking through `cram_io.h` in htslib, they only seem to use ITF8, LFT8, and int32.

CramInt in htsjdk is only used for Block checksums and Container Header sizes and checksums.  Most other such values are ITF8/LTF8.

So you're right that this does seem unused.  I can pull this change out of this PR into a different branch we can bring back if we find a later use for it.

src/main/java/htsjdk/samtools/cram/io/CramInt.java
jmthibault79: I like that

src/test/java/htsjdk/samtools/cram/io/CramIntTest.java
jmthibault79: ok

jmthibault79:
Comments addressed

src/main/java/htsjdk/samtools/cram/io/CramInt.java
```diff
@@ -39,8 +39,11 @@ public static int int32(final byte[] data) {
      * @param buffer {@link ByteBuffer} to read from
      * @return an integer value read from the buffer
      */
-    public static int int32(final ByteBuffer buffer) {
-        return buffer.get() | buffer.get() << 8 | buffer.get() << 16 | buffer.get() << 24;
+    public static int readInt32(final ByteBuffer buffer) {
+        return (0xFF & (int) buffer.get()) |
+                (0xFF & (int) buffer.get()) << 8 |
```
pshapiro4broad: casting to `int` should be unnecessary here

src/test/java/htsjdk/samtools/cram/io/IOTestCases.java
```diff
@@ -0,0 +1,135 @@
...
+                {65536, new byte[]{0, 0, 1, 0}},            // 0x010000
+                {16777216, new byte[]{0, 0, 0, 1}},         // 0x01000000
+                {2147483647, new byte[]{-1, -1, -1, 127}},  // 0x7FFFFFFF
+                {-2147483648, new byte[]{0, 0, 0, -128}},   // 0x80000000
+                {-1, new byte[]{-1, -1, -1, -1}}            // 0xFFFFFFFF
+        };
+    }
+
+    private static <T> Object[][] asDataProvider(List<T> list) {
+        Object[][] params = new Object[list.size()][] ;
```
pshapiro4broad: extra space
```suggestion
        Object[][] params = new Object[list.size()][];
```

src/test/java/htsjdk/samtools/cram/io/IOTestCases.java
```diff
@@ -0,0 +1,135 @@
...
+        for (int i=0; i<params.length; i++)
```
pshapiro4broad: missing space
```suggestion
        for (int i = 0; i < params.length; i++)
```

src/test/java/htsjdk/samtools/cram/io/ExternalCompressionTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+
+public class ExternalCompressionTest extends HtsjdkTest {
+    public static final File BZIP2_FILE = new File("src/test/resources/htsjdk/samtools/cram/io/bzip2-test.bz2");
+    public static final byte [] TEST_BYTES = "This is a simple string to test compression".getBytes();
```
pshapiro4broad: odd spacing, I would expect to not have a space before `[]`
```suggestion
    public static final byte[] TEST_BYTES = "This is a simple string to test compression".getBytes();
```

src/main/java/htsjdk/samtools/cram/io/CramInt.java
pshapiro4broad: Do you have a test case that fails without this change?

src/main/java/htsjdk/samtools/cram/io/CramInt.java
jmthibault79: I'll remove the int casts - good catch.

`0xFF &` is still necessary because int-promotion extends the negative sign into higher bytes.

pshapiro4broad: OK I now see why this is necessary, I forgot that byte isn't an unsigned type. Why, Java? Why?

Issue #1202, opened 2018-10-24
Htsjdk mangles unicode when writing sam files (and possibly bam/cram)


lbergelson:
The newest version of sam specifies which fields are allowed to be utf-8 and which must be standard 7bit ascii.  I tested it and it turns out we do not support utf-8 in `sam` files at all.  We mangle them to the ascii characters in all cases.

`AsciiWriter`  uses the very simple `StringUtil.charsToBytes` which just downcasts the input `char` to a `byte`.  This is incorrect.
We don't detect this case and instead silently corrupt the output.

We may have similar problems with cram/bam.

PR #1211, opened 2018-10-26
Changing htsjdk to produce sam version 1.6

lbergelson:
* Htsjdk has techinically been producing Sam version 1.6 since support for long cigars was added.
* Updating that list of acceptable versions to include 1.6 and setting the header version of new bams to 1.6

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMFileHeader.java
+++ b/src/main/java/htsjdk/samtools/SAMFileHeader.java
@@ -41,7 +41,7 @@ public class SAMFileHeader extends AbstractSAMHeaderRecord
     public static final String SORT_ORDER_TAG = "SO";
     public static final String GROUP_ORDER_TAG = "GO";
     public static final String CURRENT_VERSION = "1.6";
-    public static final Set<String> ACCEPTABLE_VERSIONS = CollectionUtil.makeSet("1.0", "1.3", "1.4", "1.5", "1.6");
+    public static final Set<String> ACCEPTABLE_VERSIONS = CollectionUtil.makeSet("1.0", "1.3", "1.4", "1.5", CURRENT_VERSION );
 
     private SortOrder sortOrder = null;
     private GroupOrder groupOrder = null;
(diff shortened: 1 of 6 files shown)


src/main/java/htsjdk/samtools/SAMFileHeader.java
```diff
@@ -40,8 +40,8 @@
     public static final String VERSION_TAG = "VN";
     public static final String SORT_ORDER_TAG = "SO";
     public static final String GROUP_ORDER_TAG = "GO";
-    public static final String CURRENT_VERSION = "1.5";
-    public static final Set<String> ACCEPTABLE_VERSIONS = CollectionUtil.makeSet("1.0", "1.3", "1.4", "1.5");
+    public static final String CURRENT_VERSION = "1.6";
+    public static final Set<String> ACCEPTABLE_VERSIONS = CollectionUtil.makeSet("1.0", "1.3", "1.4", "1.5", "1.6");
```
magicDGS: I would use instead the following to always keep the current version without hard-coding:

```suggestion
    public static final Set<String> ACCEPTABLE_VERSIONS = CollectionUtil.makeSet("1.0", "1.3", "1.4", "1.5", CURRENT_VERSION );
```

src/test/java/htsjdk/samtools/ValidateSamFileTest.java
```diff
@@ -528,7 +528,7 @@ public void testHeaderVersions() throws Exception {
         }
 
         // Test an unacceptable version
-        testHeaderVersion("1.6", false);
+        testHeaderVersion("1.7", false);
```
pshapiro4broad: If you used `1.1` instead then you'd no longer need to update it when the version was changed.
```suggestion
        testHeaderVersion("1.1", false);
```

cmnbroad:
👍 to taking this and addressing https://github.com/samtools/htsjdk/issues/1202 separately.

PR #1212, opened 2018-10-28
Getter for generic fields in VCFSimpleHeaderLine

Wathis:
### Description

Setting a getter for accessing to values like "Description" 

### Checklist

- [] Code compiles correctly
- [] New tests covering changes and new functionality
- [] All tests passing
- [] Extended the README / documentation, if necessary
- [] Is not backward compatible (breaks binary or source compatibility)

### DIFF
--- a/src/main/java/htsjdk/variant/vcf/VCFSimpleHeaderLine.java
+++ b/src/main/java/htsjdk/variant/vcf/VCFSimpleHeaderLine.java
@@ -126,4 +126,8 @@ public class VCFSimpleHeaderLine extends VCFHeaderLine implements VCFIDHeaderLin
     public String getID() {
         return name;
     }
-}
+	
+    public String getGenericFields() {
+	return genericFields;    
+    }	    
+ }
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/variant/vcf/VCFSimpleHeaderLine.java
```diff
@@ -126,4 +126,8 @@ public int hashCode() {
     public String getID() {
         return name;
     }
-}
\ No newline at end of file
+	
+    public String getGenericFields() {
```
lbergelson:
```suggestion
    /**
     * @return a map of all pairs of fields and values in this header line
     */
    public Map<String, String> getGenericFields() {
	    return Collections.unmodifiableMap(genericFields);
    }
```

Wathis:
@lbergelson Yes and sorry for String ;). For unit test, no problem i will do it when i have time!

cmnbroad:
I think its fine sense given the current implementation. Ideally, generic field support would live higher in the class hierarchy since all (ID) header lines can have custom attributes. The existing hierarchy only implements them for VCFSimpleHeaderLine derivatives (contig and filter), which is strange. (This is another area that was unified and fixed in my old PR that never got merged).

Wathis:
I added some test and changed the getter too ! Now it's working for me

PR #1219, opened 2018-11-01
CRAM DataSeries/DataReader/DataWriter refactor

jmthibault79:
### Description

Code simplification, new tests, and removal of unneeded reflection

### Checklist

- [x] Code compiles correctly
- [x] New tests covering changes and new functionality
- [x] All tests passing
- [x] Extended the README / documentation, if necessary
- [x] Is not backward compatible (breaks binary or source compatibility)

PR #1221, opened 2018-11-06
CRAM: Be explicit about spec IDs instead of using ordinal()

jmthibault79:
### Description

Using `ordinal()` is brittle because code will break mysteriously if the orders of enums ever change.  These Encoding, Block Compression Method, and BlockContentType IDs are mandated by the spec.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/cram/structure/BlockCompressionMethod.java
+++ b/src/main/java/htsjdk/samtools/cram/structure/BlockCompressionMethod.java
@@ -17,6 +17,9 @@
  */
 package htsjdk.samtools.cram.structure;
 
+/**
+ * The block compression methods specified by Section 8 of the CRAM spec
+ */
 public enum BlockCompressionMethod {
     RAW(0),
     GZIP(1),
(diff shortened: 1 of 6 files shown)


jmthibault79:
Inspired by review comments in #1219 which cover the use of `ordinal()` for `EncodingKey`/`DataSeries`

src/main/java/htsjdk/samtools/cram/structure/BlockCompressionMethod.java
```diff
@@ -18,5 +18,23 @@
...
+    LZMA(3),
+    RANS(4);
+
+    private final int specValue;
+
+    /**
+     * The block compression methods specified by Section 8 of the CRAM spec
+     * @param value the number assigned to each block compression method in the CRAM spec
+     */
+    BlockCompressionMethod(final int value) {
```
lbergelson: The naming of these is inconsistent and "spec" seems a bit strange in the name.  I would drop spec and just go with `getId` or `getTypeId` as appropriate.  The javadoc is a good addition, it would make sense to put it as class level javadoc  and on the getter, as well since users will never actually see the enum constructor.

PR #1223, opened 2018-11-09
Adding getPathToDataFile default method to FeatureCodec

lbergelson:
* adding a new method FeatureCodec.getPathToDataFile()
  - This method allows a special class of codec where the files that they
      accept are a configuration file and the actual data is stored elsewhere.
  - The default implementation means that existing codecs do not need to be modified in any way
* modifying AbstractFeatureReader and it's subclasses to use getPathToDataFile
* downstream tools that implement their own FeatureReader subclasses may
  want to update them to be aware of getPathToDataFile if they want to
  understand these codecs

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


lbergelson:
@jrobinso As a heads up, this touches `FeatureCodec`, but I believe it should have no effect on IGV.

jrobinso:
@lbergelson thanks for the heads up,  I agree I don't see any impact.

src/main/java/htsjdk/tribble/AbstractFeatureReader.java
```diff
@@ -145,7 +146,7 @@ protected AbstractFeatureReader(final String path, final FeatureCodec<T, SOURCE>
     protected AbstractFeatureReader(final String path, final FeatureCodec<T, SOURCE> codec,
                                     final Function<SeekableByteChannel, SeekableByteChannel> wrapper,
                                     final Function<SeekableByteChannel, SeekableByteChannel> indexWrapper) {
-        this.path = path;
+        this.path = codec.getPathToDataFile(path);
```
droazen: Add a comment with a one-line explanation of why we call this method here.

src/main/java/htsjdk/tribble/FeatureCodec.java
```diff
@@ -156,4 +156,21 @@
     default public TabixFormat getTabixFormat() {
         throw new TribbleException(this.getClass().getSimpleName() + "does not have defined tabix format");
     }
+
+    /**
+     * Codecs may override this method if they file that they recognize with {@link #canDecode(String)} is different than
```
droazen: "they file" -> "the file"

src/main/java/htsjdk/tribble/FeatureCodec.java
```diff
@@ -156,4 +156,21 @@
...
+     * the file that contains the data they pars.
```
droazen: "pars" -> "parse"

src/main/java/htsjdk/tribble/FeatureCodec.java
```diff
@@ -156,4 +156,21 @@
...
+     *
+     * This enables a class of codecs where the input file is a configuration that defines how to locate and handle the
+     * datafile.
+     *
+     * The default implementation returns the same path which was passed in.
+     *
+     * @param path the path to a file that this codec {@link #canDecode}
+     * @return the path to the data file that should be parsed by this codec to produce Features}.
```
droazen: remove stray '}'

src/main/java/htsjdk/tribble/AbstractFeatureReader.java
droazen: Also add a comment above the declaration of the `path` instance variable clarifying that this is the path to the data file, not necessarily the file recognized by `canDecode()`

src/test/java/htsjdk/tribble/AbstractFeatureReaderTest.java
```diff
@@ -228,4 +229,45 @@ public SeekableByteChannel truncate(long size) throws IOException {
...
+    @DataProvider
+    public Object[][] getVcfRedirects(){
+        return new Object[][]{
+          {"src/test/resources/htsjdk/tribble/AbstractFeatureReaderTest/vcf.redirect"},
+          {"src/test/resources/htsjdk/tribble/AbstractFeatureReaderTest/vcf.gz.redirect"}
+        };
+    }
+
+    @Test(dataProvider = "getVcfRedirects")
+    public void testRedirectCode(String vcfRedirect) throws IOException {
```
droazen: Add an explanatory comment above this test explaining what it's testing, since the name doesn't make it obvious

src/test/java/htsjdk/tribble/AbstractFeatureReaderTest.java
```diff
@@ -228,4 +229,45 @@ public SeekableByteChannel truncate(long size) throws IOException {
...
+            Assert.assertTrue( redirectVcfSize > 0, "iterator found " + redirectVcfSize + " records");
+            Assert.assertEquals(redirectVcfSize, directReader.iterator().toList().size());
+
+            final int redirectQuerySize = redirectReader.query("20", 1, 20000).toList().size();
+            Assert.assertTrue(redirectQuerySize > 0, "query found " + redirectVcfSize + " records");
+            Assert.assertEquals(redirectQuerySize, directReader.query("20", 1, 20000).toList().size() );
+        }
+    }
+
+    public static class VcfRedirectCodec extends VCFCodec{
```
droazen: Shouldn't this be `private`?

src/test/resources/htsjdk/tribble/AbstractFeatureReaderTest/test.vcf
```diff
@@ -0,0 +1,24 @@
+##fileformat=VCFv4.1
```
droazen: Can you give this file a better/more specific name than "test.vcf", since it's only used in the redirect test (or put it in a subdirectory specific to that test)?

src/main/java/htsjdk/tribble/AbstractFeatureReader.java
lbergelson: Good idea.. Done.

src/main/java/htsjdk/tribble/AbstractFeatureReader.java
```diff
@@ -102,8 +105,12 @@
      */
     public static <FEATURE extends Feature, SOURCE> AbstractFeatureReader<FEATURE, SOURCE> getFeatureReader(final String featureResource, String indexResource, final FeatureCodec<FEATURE, SOURCE> codec, final boolean requireIndex, Function<SeekableByteChannel, SeekableByteChannel> wrapper, Function<SeekableByteChannel, SeekableByteChannel> indexWrapper) throws TribbleException {
         try {
-            // Test for tabix index
-            if (methods.isTabix(featureResource, indexResource)) {
+
+            // Test for tabix index.
+            // Note that we use pathToDataFile here when determining the file type, but featureResource when constructing the readers.
+            // This is because the readers constructor will convert the path and it needs to be converted exactly once.
```
lbergelson:
```suggestion
            // This is because the reader's constructor will convert the path and it needs to be converted exactly once.
```

src/main/java/htsjdk/tribble/AbstractFeatureReader.java
```diff
@@ -102,8 +105,12 @@
      */
     public static <FEATURE extends Feature, SOURCE> AbstractFeatureReader<FEATURE, SOURCE> getFeatureReader(final String featureResource, String indexResource, final FeatureCodec<FEATURE, SOURCE> codec, final boolean requireIndex, Function<SeekableByteChannel, SeekableByteChannel> wrapper, Function<SeekableByteChannel, SeekableByteChannel> indexWrapper) throws TribbleException {
         try {
-            // Test for tabix index
-            if (methods.isTabix(featureResource, indexResource)) {
+
+            // Test for tabix index.
+            // Note that we use pathToDataFile here when determining the file type, but featureResource when constructing the readers.
+            // This is because the reader's constructor will convert the path and it needs to be converted exactly once.
```
lbergelson: testing apply batch...
```suggestion
            // This is because the reader's constructor will convert the path and it needs to be converted exactly once.
```

PR #1227, opened 2018-11-14
more changes to SAMTagUtil

lbergelson:
* reverting the change from constants to static variables, deprecating
the constants instead

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMTag.java
+++ b/src/main/java/htsjdk/samtools/SAMTag.java
@@ -23,6 +23,8 @@
  */
 package htsjdk.samtools;
 
+import htsjdk.samtools.util.StringUtil;
+
 /**
  * The standard tags for a SAM record that are defined in the SAM spec.
  */
@@ -105,7 +107,11 @@ public enum SAMTag {
     U2,
     UQ;
 
-    private final short shortValue = SAMTag.makeBinaryTag(this.name());;
+    // Cache of already-converted tags.  Should speed up SAM text generation.
+    // Not synchronized because race condition is not a problem.
+    private static final String[] stringTags = new String[Short.MAX_VALUE];
+
+    private final short shortValue = SAMTag.makeBinaryTag(this.name());
 
     /**
      * Convert from String representation of tag name to short representation.
@@ -113,7 +119,7 @@ public enum SAMTag {
      * @param tag 2-character String representation of a tag name.
      * @return Tag name packed as 2 ASCII bytes in a short.
      */
-    static short makeBinaryTag(String tag) {
+    public static short makeBinaryTag(String tag) {
         if (tag.length() != 2) {
             throw new IllegalArgumentException("String tag does not have length() == 2: " + tag);
         }
@@ -121,8 +127,26 @@ public enum SAMTag {
     }
 
     /**
+     * Convert from short representation of tag name to String representation.
+     *
+     * @param tag Tag name packed as 2 ASCII bytes in a short.
+     * @return 2-character String representation of a tag name.
+     */
+    public static String makeStringTag(final short tag) {
+        String ret = stringTags[tag];
+        if (ret == null) {
+            final byte[] stringConversionBuf = new byte[2];
+            stringConversionBuf[0] = (byte)(tag & 0xff);
+            stringConversionBuf[1] = (byte)((tag >> 8) & 0xff);
+            ret = StringUtil.bytesToString(stringConversionBuf);
+            stringTags[tag] = ret;
+        }
+        return ret;
+    }
+
+    /**
      * Get the binary representation of this tag name.
-     * @see SAMTagUtil#makeBinaryTag(String)
+     * @see SAMTag#makeBinaryTag(String)
      *
      * @return the binary representation of this tag name
      */
--- a/src/main/java/htsjdk/samtools/SAMTag.java
+++ b/src/main/java/htsjdk/samtools/SAMTag.java
@@ -146,7 +146,7 @@ public enum SAMTag {
 
     /**
      * Get the binary representation of this tag name.
-     * @see SAMTag#makeBinaryTag(String)
+     * @see #makeBinaryTag(String)
      *
      * @return the binary representation of this tag name
      */
--- a/src/main/java/htsjdk/samtools/SAMTagUtil.java
+++ b/src/main/java/htsjdk/samtools/SAMTagUtil.java
@@ -27,9 +27,11 @@ import htsjdk.samtools.util.StringUtil;
 
 /**
  * Facility for converting between String and short representation of a SAM tag.  short representation
- * is used by SAM JDK internally and is much more efficient.  Callers are encouraged to obtain the short
+ * is used by Htsjdk internally and is much more efficient.  Callers are encouraged to obtain the short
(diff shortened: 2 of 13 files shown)


src/main/java/htsjdk/samtools/SAMTagUtil.java
```diff
@@ -41,82 +43,141 @@
...
     /** @deprecated use BC instead */
     @Deprecated
-    public static final short RT = SAMTag.RT.getBinaryTag();
-    public static final short RX = SAMTag.RX.getBinaryTag();
+    public final short RT = SAMTag.RT.getBinaryTag();
+
+    /** @deprecated use {@link SAMTag#getBinaryTag()} instead. */
+    @Deprecated public final short RX = SAMTag.RX.getBinaryTag();
     /** @deprecated reserved tag for backwards compatibility only */
+
```
droazen: Stray whitespace separating this comment from the thing it's commenting on

src/main/java/htsjdk/samtools/SAMTagUtil.java
```diff
@@ -27,10 +27,12 @@
 
 /**
  * Facility for converting between String and short representation of a SAM tag.  short representation
- * is used by SAM JDK internally and is much more efficient.  Callers are encouraged to obtain the short
+ * is used by Htsjdk internally and is much more efficient.  Callers are encouraged to obtain the short
```
droazen: Htsjdk -> HTSJDK?

src/main/java/htsjdk/samtools/SAMTagUtil.java
```diff
@@ -27,10 +27,12 @@
...
  * value for a tag of interest once, and then use the SAMRecord attribute API that takes shorts rather than
  * Strings.
  *
+ * Tags that are defined by the SAM spec are included in the enum {@link SAMTag} along with their precomputed short tag.
+ *
  * @author alecw@broadinstitute.org
  */
 public final class SAMTagUtil {
```
droazen: You should perhaps consider deprecating this entire class, and moving the 2 non-deprecated methods to a different location.

lbergelson: done

src/main/java/htsjdk/samtools/SAMTagUtil.java
lbergelson: done

src/main/java/htsjdk/samtools/SAMTagUtil.java
lbergelson: gone

lbergelson:
@cmnbroad Could you take a look at this?  I just deprecated the SamTagUtil entirely and moved all the functionality to SAMTag so that we wouldn't have this weird split between them.

src/main/java/htsjdk/samtools/SAMTagUtil.java
```diff
@@ -140,8 +179,10 @@ public static SAMTagUtil getSingleton() {
      *
      * @param tag 2-character String representation of a tag name.
      * @return Tag name packed as 2 ASCII bytes in a short.
+     * @deprecated u
```
cmnbroad: Forgot to finish the message....

lbergelson: woops, good catch

src/main/java/htsjdk/samtools/SAMTag.java
```diff
@@ -105,24 +107,46 @@
     U2,
     UQ;
 
-    private final short shortValue = SAMTag.makeBinaryTag(this.name());;
+    // Cache of already-converted tags.  Should speed up SAM text generation.
+    // Not synchronized because race condition is not a problem.
+    private static final String[] stringTags = new String[Short.MAX_VALUE];
```
pshapiro4broad: static fields should be uppercase
```suggestion
    private static final String[] STRING_TAGS = new String[Short.MAX_VALUE];
```

src/main/java/htsjdk/samtools/SAMTag.java
```diff
@@ -105,24 +107,46 @@
...
+            ret = StringUtil.bytesToString(stringConversionBuf);
+            stringTags[tag] = ret;
+        }
+        return ret;
+    }
+
     /**
      * Get the binary representation of this tag name.
-     * @see SAMTagUtil#makeBinaryTag(String)
+     * @see SAMTag#makeBinaryTag(String)
```
pshapiro4broad: A minor point, but the class prefix is unnecessary if you're linking to another member of the same class.
```suggestion
     * @see #makeBinaryTag(String)
```

src/main/java/htsjdk/samtools/SAMTagUtil.java
```diff
@@ -150,16 +191,10 @@ public static short makeBinaryTag(final String tag) {
...
-        if (ret == null) {
-            final byte[] stringConversionBuf = new byte[2];
-            stringConversionBuf[0] = (byte)(tag & 0xff);
-            stringConversionBuf[1] = (byte)((tag >> 8) & 0xff);
-            ret = StringUtil.bytesToString(stringConversionBuf);
-            stringTags[tag] = ret;
-        }
-        return ret;
+    @Deprecated
+    public String makeStringTag(final short tag) {
```
pshapiro4broad: can be static
```suggestion
    public static String makeStringTag(final short tag) {
```

Unless it needs to be non-static for backward compatibility?

src/main/java/htsjdk/samtools/SAMTagUtil.java
```diff
@@ -140,8 +179,10 @@ public static SAMTagUtil getSingleton() {
      *
      * @param tag 2-character String representation of a tag name.
      * @return Tag name packed as 2 ASCII bytes in a short.
+     * @deprecated use {@link SAMTag#makeBinaryTag(String) instead}
      */
-    public static short makeBinaryTag(final String tag) {
+    @Deprecated
+    public short makeBinaryTag(final String tag) {
```
pshapiro4broad: can be static
```suggestion
    public static short makeBinaryTag(final String tag) {
```

src/main/java/htsjdk/samtools/SAMTag.java
lbergelson: This isn't really a constant though.  My thought is that things that are constant are final, but this is a static cache that's updated with new values.  It seems more confusing to make it caps since people assume those are immutable.

src/main/java/htsjdk/samtools/SAMTag.java
lbergelson: good point.

src/main/java/htsjdk/samtools/SAMTagUtil.java
lbergelson: Keeping it non-static for backwards compatibility.  I was changing them to be static but I realized that would cause confusing breaks in binary compatibility.

lbergelson: This pr is basically me fixing some breaks in compatibility that I had though would be harmless but on further inspection seemd problematic.

lbergelson:
@pshapiro4broad @cmnbroad Thank you for the reviews!

PR #1231, opened 2018-11-28
CRAM: Immutable Block

jmthibault79:
### Description

Refactoring CRAM code for future maintainability

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


src/main/java/htsjdk/samtools/cram/io/ExternalCompression.java
```diff
@@ -149,4 +150,25 @@
         buffer.get(bytes);
         return bytes;
     }
+
+    public static byte[] uncompress(final BlockCompressionMethod method, final byte[] compressedContent) {
+        try {
+            switch (method) {
+                case RAW:
+                    return compressedContent;
+                case GZIP:
```
jmthibault79: There's an opportunity for refactoring the compression classes here.  Leaving alone for this PR.

src/main/java/htsjdk/samtools/cram/structure/Block.java
```diff
@@ -1,323 +0,0 @@
-/**
```
jmthibault79: Moved to `htsjdk.samtools.cram.structure.block.Block`

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
```diff
@@ -0,0 +1,234 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ****************************************************************************
+ */
+package htsjdk.samtools.cram.structure.block;
```
jmthibault79: moved from `package htsjdk.samtools.cram.structure`

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
```diff
@@ -0,0 +1,234 @@
+/**
```
jmthibault79: Block is now the abstract base class for the 2 concrete implementations `RawBlock` and `CompressibleBlock`

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    /**
+     * Create a new file header block with the given uncompressed content.
+     * The block will have RAW (no) compression and FILE_HEADER content type.
+     *
+     * @param rawContent the uncompressed content of the block
+     * @return a new file header {@link RawBlock} object
+     */
+    public static RawBlock buildNewFileHeaderBlock(final byte[] rawContent) {
```
jmthibault79: alternate implementation: `FileHeaderBlock`, `CoreBlock`, etc

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
```diff
@@ -0,0 +1,234 @@
...
+     * Identifies whether this is a header or data block, and which kind
+     * @return the CRAM content type of the block
+     */
+    public final BlockContentType getContentType() {
+        return contentType;
+    }
+
+    /**
+     * Return the External Content ID for this block.
+     * By default this is NO_CONTENT_ID (0) but CompressibleBlocks override this
```
jmthibault79: CompressibleBlocks can still have Content ID 0 if they are not external.  Should I clarify that here?

src/main/java/htsjdk/samtools/cram/structure/block/CompressibleBlock.java
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * For external blocks, the external block Content ID
+     * Otherwise, NO_CONTENT_ID (0)
+     */
+    private final int contentId;
+
+    /**
+     * The content stored in this block, in uncompressed form
+     */
+    private final byte[] rawContent;
```
jmthibault79: alternate implementation: only store the compressed data, and uncompress on demand

src/main/java/htsjdk/samtools/cram/structure/block/CompressibleBlock.java
```diff
@@ -0,0 +1,86 @@
...
+    public final int getContentId() {
+        return contentId;
+    }
+
+    /**
+     * Return a copy of the uncompressed block content
+     */
+    @Override
+    public final byte[] getRawContent() {
+        return rawContent.clone();
```
jmthibault79: Need to copy defensively for strict immutability.  Is this overkill?

src/main/java/htsjdk/samtools/cram/structure/block/RawBlock.java
```diff
@@ -0,0 +1,52 @@
...
+    public final BlockCompressionMethod getMethod() {
+        return BlockCompressionMethod.RAW;
+    }
+
+    /**
+     * Return a copy of the uncompressed block content
+     */
+    @Override
+    public final byte[] getRawContent() {
+        return content.clone();
```
jmthibault79: Defensive copy for strict immutability. Is this overkill?

src/main/java/htsjdk/samtools/cram/structure/block/CompressibleBlock.java
```diff
@@ -0,0 +1,94 @@
...
+     * @param compressedContent the compressed form of the data to be stored in this block
+     */
+    public CompressibleBlock(final BlockCompressionMethod method,
+                      final BlockContentType type,
+                      final int contentId,
+                      final byte[] rawContent,
+                      final byte[] compressedContent) {
+        super(type);
+
+        if (type == BlockContentType.EXTERNAL && contentId == Block.NO_CONTENT_ID) {
```
jmthibault79: This check is causing many test failures - I'm investigating.

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
```diff
@@ -0,0 +1,236 @@
...
+     * Only external blocks have meaningful Content IDs
+     * Other blocks are required to have a Content ID of 0
+     */
+    public static final int NO_CONTENT_ID = 0;
+
+    /**
+     * Identifies CRAM content type of the block.
+     */
+    private final BlockContentType contentType;
+
```
cmnbroad: Would it make sense to have the BlockCompressionMethod kept here as a final, and just have `RawBlock` enforce it to always be RAW ? Same question about rawContent (since both subclasses have that). Then this can have default implementations for everything; `RawBlock` would be simplified, and `CompressibleBlock` would only override the compressedContent part.

src/main/java/htsjdk/samtools/cram/structure/block/CompressibleBlock.java
```diff
@@ -0,0 +1,97 @@
...
+        // causes test failures.  https://github.com/samtools/htsjdk/issues/1232
+//        if (type == BlockContentType.EXTERNAL && contentId == Block.NO_CONTENT_ID) {
+//            throw new CRAMException("Valid Content ID required for external blocks.");
+//        }
+
+        if (type != BlockContentType.EXTERNAL && contentId != Block.NO_CONTENT_ID) {
+            throw new CRAMException("Cannot set a Content ID for non-external blocks.");
+        }
+
+        this.method = method;
```
cmnbroad: So is `method` allowed to be `raw` here ? If so, its not clear to me what really distinguishes a `RawBlock` from a `CompressibleBlock`; since this case can be represented either way  ?

cmnbroad:
@jmthibault79 Was just looking at this again - is it ready for review (wasn't sure if you were still making changes) ?

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
jmthibault79: I found a way to make this work well (including the static factory) and I'll be committing in a moment.

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
```diff
@@ -0,0 +1,235 @@
...
+    /**
+     * Deserialize the Block from the {@link InputStream}. The reading is parametrized by the major CRAM version number.
+     *
+     * @param major CRAM version major number
+     * @param inputStream    input stream to read the block from
+     * @return a new {@link CompressibleBlock} object with fields and content from the input stream
+     * @throws IOException as per java IO contract
+     */
+    public static CompressibleBlock readFromInputStream(final int major, InputStream inputStream) throws IOException {
```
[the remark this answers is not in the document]
jmthibault79: Now returns `Block` subclasses.

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
```diff
@@ -0,0 +1,235 @@
...
+    public abstract int getCompressedContentSize();
+
+    /**
+     * Write the block out to the the specified {@link OutputStream}. The method is parametrized with CRAM major version number.
+     *
+     * @param major CRAM version major number
+     * @param outputStream    output stream to write to
+     * @throws IOException as per java IO contract
+     */
+    public final void write(final int major, final OutputStream outputStream) throws IOException {
```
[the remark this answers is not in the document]
jmthibault79: How about the alternate: rename the other to `read()` ?

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
jmthibault79: Yes - and we do not actually need `CompressibleBlock`

jmthibault79:
@cmnbroad ready for re-review now, after the subclasses commit

cmnbroad:
@jmthibault79 I still feel like we don't have the `Block` hierarchy factored quite correctly yet. It doesn't seem like any useful polymorphism results from having the class hierarchy based on content type, type casting is still required, and the actual "content type" of a given block is redundantly represented three different ways now (in the class hierarchy itself; in the base class `BlockContentType` instance field; and in the static `BlockContentType` constant maintained by each subclass). Since the call sites all seem to know what content type they're reading or writing (i.e., SliceIO knows it wants to extract, say, a Slice header), most of this seems to be about enforcing/validating the compression/content-type combination for each block type.

What do you think about this proposal: go back to just `Block` (possibly with protected subclass `CompressibleBlock` since there is some useful polymorphism there wrt/compressed vs. raw content), where `Block` has factory methods like it does now for creating/validating each of type of block from either raw content or from a stream. The factory methods would enforce or validate the correct combination of "Block" params (compressible, etc.), and possibly substitute `Compressible` block as appropriate - but the only class type callers would only ever deal with is `Block`. 

Then in subsequent PRs, we can create classes that model things like `SliceHeader`, with fields and methods specific to it, and constructors for those classes read/write from/to plain old `Block`.

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
```diff
@@ -0,0 +1,294 @@
...
+            if (v3OrHigher) {
+                final int actualChecksum = ((CRC32InputStream) inputStream).getCRC32();
+                final int checksum = CramInt.readInt32(inputStream);
+                if (checksum != actualChecksum) {
+                    throw new RuntimeException(String.format("Block CRC32 mismatch: %04x vs %04x", checksum, actualChecksum));
+                }
+            }
+
+            // TODO: is this check worthwhile?  it may be expensive.
+            final byte[] uncompressedContent = ExternalCompression.uncompress(method, compressedContent);
```
cmnbroad: The validation seems like a good idea, but it does seem like it would be expensive to do speculatively. Maybe we should do it lazily on "un-compression" instead. We'd have to retain the original size from the stream, but thats cheap.

jmthibault79: good idea, thanks

jmthibault79:
I see your point about the hierarchy.  Here's another attempt, with `ExternalDataBlock` as the only subclass, because it has the meaningful distinction of being associated with a Content ID.

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
```diff
@@ -0,0 +1,272 @@
...
+     * @param uncompressedLength the length of the content stored in this block when uncompressed
+     */
+    protected Block(final BlockCompressionMethod method,
+                    final BlockContentType type,
+                    final byte[] compressedContent,
+                    final int uncompressedLength) {
+        this.method = method;
+        this.contentType = type;
+        this.compressedContent = compressedContent;
+        this.uncompressedLength = uncompressedLength;
```
cmnbroad: It seems error prone to mix the uncompressed content and compressed lengths like this, especially when many of the calls to this pass in raw content as the compressed content.

Maybe this should have two separate constructors:

`Block (final BlockContentType type, final byte[] rawContent)` (for creating raw blocks from raw content)

and:

`Block (final BlockContentType type, final BlockCompressionMethod method, final byte[] compressedContent, final int expectedUncompressedLength)`

and just have corresponding factory methods (the call sites can pass in the contentType). Then have the implementation keep the expectedUncompressedLength length separate from the actual length, and manage the content and length inter-conversions on demand. We'd need to do a similar thing for `External`.

src/main/java/htsjdk/samtools/cram/build/ContainerFactory.java
```diff
@@ -170,21 +170,13 @@ private static Slice buildSlice(final List<CramCompressionRecord> records,
         writer.writeCramCompressionRecords(records, slice.alignmentStart);
 
         bitOutputStream.close();
-        slice.coreBlock = Block.buildNewCore(bitBAOS.toByteArray());
+        slice.coreBlock = Block.uncompressedCoreBlock(bitBAOS.toByteArray());
```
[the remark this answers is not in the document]
jmthibault79: ok

src/main/java/htsjdk/samtools/cram/build/ContainerFactory.java
```diff
@@ -170,21 +170,13 @@ private static Slice buildSlice(final List<CramCompressionRecord> records,
...
-            externalBlock.setContentId(key);
-            externalBlock.setContentType(BlockContentType.EXTERNAL);
-
             final ExternalCompressor compressor = header.externalCompressors.get(key);
-            final byte[] rawData = os.toByteArray();
-            final byte[] compressed = compressor.compress(rawData);
-            externalBlock.setContent(rawData, compressed);
-            externalBlock.setMethod(compressor.getMethod());
+            final byte[] rawData = map.get(key).toByteArray();
+            final ExternalDataBlock externalBlock = new ExternalDataBlock(key, compressor, rawData);
```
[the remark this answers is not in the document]
jmthibault79: Yes definitely

src/main/java/htsjdk/samtools/cram/structure/ContainerIO.java
```diff
@@ -228,4 +217,21 @@ public void write(final int b) throws IOException {
...
     }
+
+    /**
+     * Read a COMPRESSION_HEADER Block from an InputStream and return its contents as a CompressionHeader
+     *
+     * @param major the CRAM major version number
+     * @param inputStream the stream to read from
+     * @return a new CompressionHeader from the input
+     */
+    private static CompressionHeader readAsCompressionHeader(final int major, final InputStream inputStream) {
```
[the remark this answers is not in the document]
jmthibault79: makes sense

src/main/java/htsjdk/samtools/cram/build/CramIO.java
```diff
@@ -279,18 +279,18 @@ private static SAMFileHeader readSAMFileHeader(final Version version, InputStrea
...
                  * pending issue: container.containerByteSize inputStream 2 bytes shorter
 				 * then needed in the v21 test cram files.
 				 */
-                block = Block.readFromInputStream(version.major, inputStream);
+                block = Block.read(version.major, inputStream);
             }
         }
 
-        inputStream = new ByteArrayInputStream(block.getRawContent());
+        inputStream = new ByteArrayInputStream(block.getUncompressedContent());
```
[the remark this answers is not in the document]
jmthibault79: I would prefer to leave the term "raw" for Blocks which have RAW compression method.  This performs an decompression on the data if compressed, or a no-op if RAW.

src/main/java/htsjdk/samtools/cram/compression/ExternalCompression.java
```diff
@@ -150,4 +151,25 @@
         buffer.get(bytes);
         return bytes;
     }
+
+    public static byte[] uncompress(final BlockCompressionMethod method, final byte[] compressedContent) {
```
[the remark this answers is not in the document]
jmthibault79: I like it here, as a generic counterpart to methods like `gunzip()`.  I have a general "compression" refactoring in my mental backlog, so we can revisit this then.

src/main/java/htsjdk/samtools/cram/structure/ContainerIO.java
```diff
@@ -165,17 +164,7 @@ public static int writeContainer(final Version version, final Container containe
...
-        block.setContentId(0);
-        block.setMethod(BlockCompressionMethod.RAW);
-        final byte[] bytes;
-        try {
-            bytes = container.header.toByteArray();
-        } catch (final IOException e) {
-            throw new RuntimeException("This should have never happened.");
-        }
-        block.setRawContent(bytes);
+        final Block block = Block.uncompressedCompressionHeaderBlock(container.header.toByteArray());
```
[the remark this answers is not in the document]
jmthibault79: I like it.

src/main/java/htsjdk/samtools/cram/structure/SliceIO.java
```diff
@@ -36,27 +38,26 @@
...
     private static final Log log = Log.getInstance(SliceIO.class);
 
-    private static void readSliceHeadBlock(final int major, final Slice slice, final InputStream inputStream) throws IOException {
-        slice.headerBlock = Block.readFromInputStream(major, inputStream);
-        parseSliceHeaderBlock(major, slice);
-    }
+    private static void readSliceHeader(final int major, final Slice slice, final InputStream readInputStream) throws IOException {
+        slice.headerBlock = Block.read(major, readInputStream);
+        if (slice.headerBlock.getContentType() != BlockContentType.MAPPED_SLICE)
+            throw new RuntimeException("Content type does not match: " + slice.headerBlock.getContentType().name());
```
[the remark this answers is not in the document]
jmthibault79: yes

src/main/java/htsjdk/samtools/cram/structure/SliceIO.java
```diff
@@ -105,17 +106,20 @@ private static void parseSliceHeaderBlock(final int major, final Slice slice) th
...
+            final Block block = Block.read(major, inputStream);
 
             switch (block.getContentType()) {
                 case CORE:
                     slice.coreBlock = block;
                     break;
                 case EXTERNAL:
-                    if (slice.embeddedRefBlockContentID == block.getContentId()) slice.embeddedRefBlock = block;
-                    slice.external.put(block.getContentId(), block);
+                    final ExternalDataBlock extBlock = (ExternalDataBlock) block;
```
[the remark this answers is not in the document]
jmthibault79: It does, because this `block` comes from the generic `Block.read()`.  If we made `external` and `embeddedRefBlock` into simple `Block`s we could remove this.  I prefer them as `ExternalDataBlock`s.

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
```diff
@@ -0,0 +1,272 @@
...
+    public int getContentId() {
+        return NO_CONTENT_ID;
+    }
+
+    public final byte[] getUncompressedContent() {
+        final byte[] uncompressedContent = ExternalCompression.uncompress(method, compressedContent);
+        if (uncompressedContent.length != uncompressedLength) {
+            throw new CRAMException(String.format("Block uncompressed length did not match expected length: %04x vs %04x", uncompressedLength, uncompressedContent.length));
+        }
+        return ExternalCompression.uncompress(method, compressedContent);
```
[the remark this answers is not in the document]
jmthibault79: doh, thanks

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
```diff
@@ -0,0 +1,272 @@
...
+     */
+    public static Block read(final int major, InputStream inputStream) {
+        final boolean v3OrHigher = major >= CramVersions.CRAM_v3.major;
+        if (v3OrHigher) {
+            inputStream = new CRC32InputStream(inputStream);
+        }
+
+        try {
+            final BlockCompressionMethod method = BlockCompressionMethod.byId(inputStream.read());
+            final BlockContentType type = BlockContentType.byId(inputStream.read());
```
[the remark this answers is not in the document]
jmthibault79: ok

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
jmthibault79: ok for Block.  Not sure this makes sense for External - which now has two constructors for different purposes already.

src/main/java/htsjdk/samtools/cram/build/CramIO.java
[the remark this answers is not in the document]
cmnbroad: Yes, thats exactly the right thing to do and what I intended - not sure why I put this comment on this line.

src/main/java/htsjdk/samtools/cram/compression/ExternalCompression.java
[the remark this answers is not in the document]
cmnbroad: ok.

src/main/java/htsjdk/samtools/cram/structure/SliceIO.java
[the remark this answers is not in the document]
jmthibault79: changed it to `Block`

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
```diff
@@ -0,0 +1,284 @@
...
+//            throw new CRAMException("Valid Content ID required for external blocks.");
+//        }
+
+        if (contentType != BlockContentType.EXTERNAL && getContentId() != Block.NO_CONTENT_ID) {
+            throw new CRAMException("Cannot set a Content ID for non-external blocks.");
+        }
+    }
+
+    /**
+     * Private constructor of a generic Block, to be called by static factory methods and subclasses.
```
cmnbroad: Private -> protected.

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
```diff
@@ -0,0 +1,284 @@
...
+     *
+     * Only ExternalBlocks have a meaningful Content ID, so that class overrides this method.
+     *
+     * @return the External Content ID, or NO_CONTENT_ID
+     */
+    public int getContentId() {
+        return NO_CONTENT_ID;
+    }
+
+    public final byte[] getUncompressedContent() {
```
cmnbroad: needs javadoc

src/main/java/htsjdk/samtools/cram/structure/block/ExternalBlock.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Create a new external data block with the given content ID, compressor, and uncompressed content.
+     * The block will have EXTERNAL content type.
+     *
+     * @param contentId the external identifier for the block
+     * @param compressor which external compressor to use on this block
+     * @param rawContent the uncompressed content of the block
+     */
+    public ExternalBlock(final int contentId, final ExternalCompressor compressor, final byte[] rawContent) {
```
cmnbroad: I think it would be more consistent to have this take a `BlockCompressionMethod` like the other Block constructors, instead of an ExternalCompressor instance. ContainerFactory/ContainerHeader currently stores the instances, but once those are refactored maybe they just need to store the method ? Not sure if that has other ramifications. Also, the other Block constructor methods have the compression method as the first param, followed by contentID for the ones that have that, rather than the other way around like this.

src/test/java/htsjdk/samtools/cram/structure/BlockTest.java
```diff
@@ -0,0 +1,145 @@
...
+        final Block core = Block.createRawCoreDataBlock(testData);
+        contentCheck(core, testData, testData);
+    }
+
+    private Block roundTrip(final Block in, final Version version) {
+        byte[] written;
+        try (final ByteArrayOutputStream os = new ByteArrayOutputStream()) {
+            in.write(version.major, os);
+            written = os.toByteArray();
+        } catch (final IOException e) {
```
cmnbroad: In tests we can just declare that these methods and the test methods "throw IOException" rather than catching and wrapping them, and just get rid of the catch clause.

src/test/java/htsjdk/samtools/cram/structure/BlockTest.java
```diff
@@ -0,0 +1,145 @@
...
+        final byte[] testData = "TEST STRING".getBytes();
+
+        final Block shBlock = Block.createRawSliceHeaderBlock(testData);
+
+        final Block rtBlock2 = roundTrip(shBlock, CramVersions.CRAM_v2_1);
+        contentCheck(rtBlock2, testData, testData);
+
+        final Block rtBlock3 = roundTrip(shBlock, CramVersions.CRAM_v3);
+        contentCheck(rtBlock3, testData, testData);
+    }
```
cmnbroad: Factoring out a data provider with {version, testdata} entries would eliminate the redundancy and reduce a lot of these test methods to just 2 or 3 lines.

src/main/java/htsjdk/samtools/cram/structure/CompressionHeader.java
```diff
@@ -201,13 +193,25 @@ else if (TD_tagIdsDictionary.equals(key)) {
...
-        write(byteArrayOutputStream);
-        return byteArrayOutputStream.toByteArray();
+    /**
+     * Write this CompressionHeader out to an internal OutputStream, wrap it in a Block, and write that
+     * Block out to the passed-in OutputStream.
+     *
+     * @param cramVersion the CRAM major version number
+     * @param blockStream the stream to write to
+     */
+    public void write(final Version cramVersion, final ByteArrayOutputStream blockStream) {
```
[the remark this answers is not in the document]
jmthibault79: true

src/main/java/htsjdk/samtools/cram/build/ContainerFactory.java
```diff
@@ -116,11 +116,11 @@ private static void calculateAlignmentBoundaries(final Container container) {
 
     private static Slice buildSlice(final List<CramCompressionRecord> records,
                                     final CompressionHeader header)
-            throws IllegalArgumentException, IllegalAccessException,
+            throws IllegalArgumentException,
```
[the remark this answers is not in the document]
jmthibault79: nice

src/test/java/htsjdk/samtools/cram/structure/BlockTest.java
jmthibault79: ok

src/test/java/htsjdk/samtools/cram/structure/BlockTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+        // raw and compressed data are equal to what was given
+
+        Assert.assertEquals(actual.getUncompressedContent(), expectedRaw);
+        Assert.assertEquals(actual.getUncompressedContentSize(), expectedRaw.length);
+        Assert.assertEquals(actual.getCompressedContent(), expectedCompressed);
+        Assert.assertEquals(actual.getCompressedContentSize(), expectedCompressed.length);
+    }
+
+    private void rtCheck(final Block srcBlock, final byte[] testData, final Version[] versions) throws IOException {
```
cmnbroad: Part of the reason I suggested using a data provider was to eliminate this need for this kind of loop.

src/test/java/htsjdk/samtools/cram/structure/BlockTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+        try (final InputStream is = new ByteArrayInputStream(written)) {
+            return Block.read(version.major, is);
+        }
+    }
+
+    @DataProvider(name = "RoundTripTest")
+    public static Object[][] rtProvider() {
+        return new Object[][]{
+                {"TEST STRING".getBytes(), new Version[]{CramVersions.CRAM_v2_1, CramVersions.CRAM_v3}}
```
cmnbroad: A data provider with a single entry is questionable. I was thinking the version would be separate test cases in the provider (you'd have to duplicate the test string which is fine, and probably putting the getBytes call into the test method body) .That way, you could eliminate the rtCheck loop, and maybe the whole method.

jmthibault79: that makes more sense

PR #1234, opened 2018-12-03
Replaced several for-each loops in VariantContext.Make() based on HaplotypeCaller profiling

jamesemery:
These are micro optimizations that add up to a pretty small change in the HaplotypeCaller (Exome +GVCF) profile when they are implemented. Here are some profiling snippets:

Before: 
screen shot 2018-12-03 at 3 45 29 pm

After:
screen shot 2018-12-03 at 3 45 07 pm

Then some runtime examples on my laptop just chromosome 15 of the same bam:
Before:  `4m26s, 4m16s, 4m18s, 4m15s, 4m28s`
After:     `4m04s, 4m09s, 4m15s, 4m11s, 4m07`

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/VariantContext.java
+++ b/src/main/java/htsjdk/variant/variantcontext/VariantContext.java
@@ -822,8 +822,9 @@ public class VariantContext implements Feature, Serializable {
             return true;
 
         final List<Allele> allelesToConsider = considerRefAllele ? getAlleles() : getAlternateAlleles();
-        for ( int i = 0; i < allelesToConsider.size(); i++) {
-            if ( allelesToConsider.get(i).equals(allele, ignoreRefState) )
+        for (int i = 0, allelesToConsiderSize = allelesToConsider.size(); i < allelesToConsiderSize; i++) {
+            Allele anAllelesToConsider = allelesToConsider.get(i);
+            if (anAllelesToConsider.equals(allele, ignoreRefState))
                 return true;
         }
 
@@ -1356,7 +1357,7 @@ public class VariantContext implements Feature, Serializable {
             Genotype genotype = genotypes.get(i);
             if ( genotype.isAvailable() ) {
                 final List<Allele> alleles = genotype.getAlleles();
-                for ( int j = 0; j < alleles.size(); j++ ) {
+                for ( int j = 0, size = alleles.size(); j < size; j++ ) {
                     final Allele gAllele = alleles.get(j);
                     if ( ! hasAllele(gAllele) && gAllele.isCalled() )
                         throw new IllegalStateException("Allele in genotype " + gAllele + " not in the variant context " + alleles);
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1485,9 +1485,10 @@ public String toStringWithoutGenotypes() {
 
         boolean sawRef = false;
         for ( final Allele a : alleles ) {
```
lbergelson: should this be made into a doubly indexed array? wouldn't that be faster?

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1353,9 +1353,9 @@ private void validateAlleles() {
     private void validateGenotypes() {
         if ( this.genotypes == null ) throw new IllegalStateException("Genotypes is null");
 
-        for ( final Genotype g : this.genotypes ) {
-            if ( g.isAvailable() ) {
-                for ( Allele gAllele : g.getAlleles() ) {
+        for ( int i = 0; i < genotypes.size(); i++ ) {
+            if ( genotypes.get(i).isAvailable() ) {
```
lbergelson: pull out the get(i) as a variable if speed is so essential here.

lbergelson: shouldn't the inner loop be made index because that will be the most essential one

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -823,8 +823,8 @@ private boolean hasAllele(final Allele allele, final boolean ignoreRefState, fin
             return true;
 
         final List<Allele> allelesToConsider = considerRefAllele ? getAlleles() : getAlternateAlleles();
-        for ( Allele a : allelesToConsider ) {
-            if ( a.equals(allele, ignoreRefState) )
+        for ( int i = 0; i < allelesToConsider.size(); i++) {
```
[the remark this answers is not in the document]
jamesemery: thats a good point, the code is ~identical too it looks like

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
jamesemery: It would, but the input to that method is a collection, not a list so I don't actually have the means to access it by index and I don't want to change the VariantContext API to only accept a list. Getting rid of the iteration over the second list seems to have accounted for ~half of the runtime this method was taking up.

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
jamesemery: Yes, I missed that. Though its worth mentioning that most of the improvement in this method came from genotypes.isAvailible which was more expensive than the inner loop.  I'll extract it out however since we are here

jamesemery:
@lbergelson Responded to your comments. Let me know if there is anything else I should do for the branch. And I am very sorry about the readability cost, the doubly indexed list is a sad change that I wasn't necessarily going to implement but you did want it...

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
jamesemery: I would say this is probably the grosses of the unreadable code after this change...

PR #1238, opened 2018-12-05
- implementing the spec change in 

yfarjoun:
https://github.com/samtools/hts-specs/pull/333

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMSequenceRecord.java
+++ b/src/main/java/htsjdk/samtools/SAMSequenceRecord.java
@@ -74,7 +74,7 @@ public class SAMSequenceRecord extends AbstractSAMHeaderRecord implements Clonea
     // These are the chars matched by \\s.
     private static final char[] WHITESPACE_CHARS = {' ', '\t', '\n', '\013', '\f', '\r'}; // \013 is vertical tab
 
-    private static Pattern ILLEGAL_RNAME_CHARS = Pattern.compile("[\\\\,\"\'`()<>{}\\]\\[]");
+    private static Pattern LEGAL_RNAME_PATTERN = Pattern.compile("[0-9A-Za-z!#$%&+./:;?@^_|~-][0-9A-Za-z!#$%&*+./:;=?@^_|~-]*");
 
     /**
      * @deprecated Use {@link #SAMSequenceRecord(String, int)} instead.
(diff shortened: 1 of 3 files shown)


src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -57,10 +57,10 @@
 
 
     /**
-     * This is not a valid sequence name, because it is reserved in the MRNM field of SAM text format
+     * This is not a valid sequence name, because it is reserved in the RNEXT field of SAM text format
      * to mean "same reference as RNAME field."
      */
-    public static final String RESERVED_MRNM_SEQUENCE_NAME = "=";
+    public static final String RESERVED_RNEXT_SEQUENCE_NAME = "=";
```
lbergelson: deprecate this instead of just renaming it

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -85,8 +87,8 @@ public SAMSequenceRecord(final String name) {
 
     public SAMSequenceRecord(final String name, final int sequenceLength) {
         if (name != null) {
-            if (SEQUENCE_NAME_SPLITTER.matcher(name).find()) {
-                throw new SAMException("Sequence name contains invalid character: " + name);
+            if (!LEGAL_RNAME_PATTERN.matcher(name).useAnchoringBounds(true).matches()) {
```
lbergelson: move this into the validate method

nh13: `SEQUENCE_NAME_SPLITTER` in both code and comments can be removed correct?  I still see it elsewhere.

jmarshall:
I notice that this enforces RNAME compliance with the regexp regardless of any declared **\@HD-VN**.

Did HTSJDK consider making this check dependent on the file's VN header and allowing such RNAMEs in existing files with HD-VN≤1.5 or so? There was discussion in https://github.com/samtools/hts-specs/pull/333#issuecomment-422835741 that it seems important not to gratuitously reject existing data files (if any…) while allowing for diagnostics on newly-produced files.

yfarjoun:
hmmm. this is a good point. will look into making is 1.5 and on only (not sure how....)

yfarjoun:
@jmarshall, we have decided (in htsjdk) to implement a blanket rejection of non-compliant sam files. if this turns out to be an undue burden on our users, we will consider putting to work to make htsjdk more version-aware. right now we are working on the next version which should include more version-aware features and it should be easier to do such a thing there...I still think that we were right to keep the **SPEC** version specific and not declare all old SAM files as non-compliant.

jmarshall:
Fair enough. Please take a look at the text added in that new commit on samtools/hts-specs#333 too :smile:

PR #1245, opened 2018-12-19
new Interface VCFIterator extends CloseableIterator<VariantContext> (continued)

romanzenka:
### Description

This is a PR based on #837 - the only difference is addressing the reviewer comments + a little bit of extra testing.

My purpose is to continue "carrying the torch" of the original PR by @lindenb until successful completion, as I could REALLY use the VCFIterator interface.

### Checklist

- [x] Code compiles correctly
- [x] New tests covering changes and new functionality
- [x] All tests passing
- [x] Extended the README / documentation, if necessary
- [ ] Is not backward compatible (breaks binary or source compatibility)

PR #1246, opened 2018-12-19
resolving some nitpicks from pr #1245

lbergelson:
resolving some nitpicks left over from pr #1245

### DIFF
--- a/src/main/java/htsjdk/samtools/util/IOUtil.java
+++ b/src/main/java/htsjdk/samtools/util/IOUtil.java
@@ -98,9 +98,6 @@ public class IOUtil {
     public static final String COMPRESSED_VCF_FILE_EXTENSION = ".vcf.gz";
     public static final String COMPRESSED_VCF_INDEX_EXTENSION = ".tbi";
 
-
-
-
     /** Possible extensions for VCF files and related formats. */
     public static final String[] VCF_EXTENSIONS = {VCF_FILE_EXTENSION, COMPRESSED_VCF_FILE_EXTENSION, BCF_FILE_EXTENSION};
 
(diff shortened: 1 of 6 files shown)


romanzenka:
Looks like there is a tiiiny coverage drop thanks to an exception being thrown when stream does not have "markSupported". I can try to make a test case for this if you want me to.

src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -99,6 +99,8 @@
     public static final String COMPRESSED_VCF_INDEX_EXTENSION = ".tbi";
 
 
+
```
yfarjoun: why extra nl?

src/main/java/htsjdk/variant/utils/VCFHeaderReader.java
```diff
@@ -50,7 +51,7 @@ public static VCFHeader readHeaderFrom(final SeekableStream in) throws IOExcepti
     private static InputStream bufferAndDecompressIfNecessary(final InputStream in) throws IOException {
         BufferedInputStream bis = new BufferedInputStream(in);
         // despite the name, SamStreams.isGzippedSAMFile looks for any gzipped stream (including block compressed)
```
yfarjoun: update comment

src/main/java/htsjdk/samtools/util/IOUtil.java
lbergelson: accident...

src/main/java/htsjdk/variant/utils/VCFHeaderReader.java
lbergelson: done

PR #1248, opened 2018-12-28
Parse valid VCF 4.2 with ##INFO containing Source + Version

romanzenka:
### Description

As per bug #517 - despite VCFv4.2 spec allowing for "Source" and "Version" fields in ##INFO, htsjdk would throw an error. This fixes the issue.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


romanzenka:
I will work on getting coverage back in shape

lbergelson:
@cmnbroad Could you take a look at this?  It's one of the tickets that was slated to be closed by your giant refactor branch that never went int.

lbergelson:
@romanzenka Thanks for doing this!  This warning might be too late, but the code coverage is really oversensitive and tends to fail at random due to some non-determinism in the test suite.  So don't worry too much about pleasing it.

cmnbroad:
Yes, the original PR refactored the entire header line class hierarchy, so it would be hard to extract just a piece of it. I'll take a look at the changes here.

src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
```diff
@@ -160,9 +163,11 @@ protected VCFCompoundHeaderLine(String line, VCFHeaderVersion version, Supported
         super(lineType.toString(), "");
 
         final ArrayList<String> expectedTags = new ArrayList(Arrays.asList("ID", "Number", "Type", "Description"));
-        if (version.isAtLeastAsRecentAs(VCFHeaderVersion.VCF4_2))
-            expectedTags.add("Version");
-        final Map<String, String> mapping = VCFHeaderLineTranslator.parseLine(version, line, expectedTags);
+        Set<String> optionalTags = Collections.emptySet();  
+        if (version.isAtLeastAsRecentAs(VCFHeaderVersion.VCF4_2)) {
+            optionalTags = new HashSet<String>(Arrays.asList("Source", "Version"));
```
pshapiro4broad:
```suggestion
            optionalTags = new HashSet<>(Arrays.asList("Source", "Version"));
```

pshapiro4broad: Also, not sure of the benefit vs overhead of using `Set` vs `List` here if the collection only contains two elements.

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
```diff
@@ -48,13 +50,19 @@
     }
 
     public static Map<String,String> parseLine(VCFHeaderVersion version, String valueLine, List<String> expectedTagOrder) {
-        return mapping.get(version).parseLine(valueLine,expectedTagOrder);
+        return mapping.get(version).parseLine(valueLine,expectedTagOrder,Collections.emptySet());
```
pshapiro4broad: IMO better to forward here to avoid code duplication.
```suggestion
        return parseLine(version, valueLine, expectedTagOrder, Collections.emptySet());
```

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
```diff
@@ -48,13 +50,19 @@
...
+    }
+    
+    public static Map<String,String> parseLine(VCFHeaderVersion version, String valueLine, List<String> expectedTagOrder, Set<String> optionalTags) {
+        return mapping.get(version).parseLine(valueLine,expectedTagOrder,optionalTags);
```
pshapiro4broad:
```suggestion
        return mapping.get(version).parseLine(valueLine,expectedTagOrder, optionalTags);
```

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
```diff
@@ -65,10 +73,23 @@
     /**
      * parse a VCF4 line
      * @param valueLine the line
+     * @param expectedTagOrder List of expected tags
      * @return a mapping of the tags parsed out
      */
     @Override
     public Map<String, String> parseLine(String valueLine, List<String> expectedTagOrder) {
+        return parseLine(valueLine, expectedTagOrder, Collections.emptySet());
```
pshapiro4broad: This is a good candidate for a default method in `VCFLineParser`. Once you have that, you no longer need a method here.

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
```diff
@@ -48,13 +50,19 @@
...
+    public static Map<String,String> parseLine(VCFHeaderVersion version, String valueLine, List<String> expectedTagOrder, Set<String> optionalTags) {
+        return mapping.get(version).parseLine(valueLine,expectedTagOrder,optionalTags);
     }
 }
 
 
 interface VCFLineParser {
     public Map<String,String> parseLine(String valueLine, List<String> expectedTagOrder);
+
+    public Map<String,String> parseLine(String valueLine, List<String> expectedTagOrder, Set<String> optionalTags);
```
pshapiro4broad: Note that this will break any code that currently subclasses `VCFLineParser` so it is a breaking API change

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
```diff
@@ -65,10 +73,23 @@
...
     public Map<String, String> parseLine(String valueLine, List<String> expectedTagOrder) {
+        return parseLine(valueLine, expectedTagOrder, Collections.emptySet());
+    }
+    
+    /**
+     * parse a VCF4 line
+     * @param valueLine the line
+     * @param expectedTagOrder List of expected tags
+     * @param optionalTags Set of tags that may or may not be present.
+     * @return a mapping of the tags parsed out
```
pshapiro4broad: The interface should have this javadoc, so it's present on all subclass implementations.

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
```diff
@@ -115,7 +136,7 @@
                 escape = false;
                 switch (c) {
                     case ('<') : if (index == 0) break; // if we see a open bracket at the beginning, ignore it
-                    case ('>') : if (index == valueLine.length()-1) ret.put(key,builder.toString().trim()); break; // if we see a close bracket, and we're at the end, add an entry to our list
+                    case ('>') : if (index == valueLine.length()-1 && !key.isEmpty()) ret.put(key,builder.toString().trim()); break; // if we see a close bracket, and we're at the end, add an entry to our list
```
pshapiro4broad: All the cases in this `switch` should be reformatted to follow the code conventions
```
case '>':
    if (index == valueLine.length() - 1 && !key.isEmpty()) {
        // if we see a close bracket, and we're at the end, add an entry to our list
        ret.put(key, builder.toString().trim());
        break;
    }
```

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
```diff
@@ -132,11 +153,16 @@
         // validate the tags against the expected list
         index = 0;
         if ( expectedTagOrder != null ) {
-            if ( ret.size() > expectedTagOrder.size() )
+            if ( ret.size() > expectedTagOrder.size() + optionalTags.size())
                 throw new TribbleException.InvalidHeader("unexpected tag count " + ret.size() + " in line " + valueLine);
             for ( String str : ret.keySet() ) {
-                if ( !expectedTagOrder.get(index).equals(str) )
+                if (index < expectedTagOrder.size()) {
```
pshapiro4broad: This should also be updated to follow code conventions
```
            for (String str : ret.keySet()) {
                if (index < expectedTagOrder.size()) {
                    if (!expectedTagOrder.get(index).equals(str)) {
                        throw new TribbleException.InvalidHeader("Unexpected tag " + str + " in line " + valueLine);
                    }
                } else if (!optionalTags.contains(str)) {
                    throw new TribbleException.InvalidHeader("Unexpected tag " + str + " in line " + valueLine);
                }
                index++;
            }
```

src/test/java/htsjdk/variant/vcf/VCFHeaderLineTranslatorUnitTest.java
```diff
@@ -14,49 +20,143 @@ public void testParseVCF4HeaderLine() {
...
+                        line,
+                        expectedTagOrder,
+                        optionalTags
+                );
+            }
+            if (!error.isEmpty()) {
+                Assert.fail("Expected failure: '" + error + "', got success");
+            }
+        }
+        catch (Exception e) {
```
pshapiro4broad: format on previous line, e.g. `} catch (Exception e) {`

src/test/java/htsjdk/variant/vcf/VCFHeaderLineTranslatorUnitTest.java
```diff
@@ -14,49 +20,143 @@ public void testParseVCF4HeaderLine() {
...
+            if (error.isEmpty()) {
```
pshapiro4broad: Why have a value in `error` if it is unused? Is there a way to assert that the error we get is the error we expect to get?

src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
romanzenka: I was trying to make an interface that can survive into the future as the number of flag grows.
The optional flags are being tested for presence/absence, their order does not matter in this case, which is a Set operation. If you think this is extremely unlikely to grow to the point when a Set would matter, I can switch back to list. Would not want to have a breaking change in the interface later.

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
romanzenka: I agree, but my goal was not to break the library interface. So I did not want to simply add a parameter to a function, I was forced to duplicate code. What are your policies when it comes to changing the interfaces?

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
romanzenka: I am a bit confused - why that one extra space, while the rest of the line does not have spaces? I was just trying to be consistent here.

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
romanzenka: Good idea, sorry still mentally stuck in Java 7 world.

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
romanzenka: How can I possibly solve this?

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
romanzenka: Will do

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
romanzenka: Mixing reformats with logic changes in a single PR is a bad practice IMHO. If you want to do a "pure reformat" PR, we can make a separate one.

src/test/java/htsjdk/variant/vcf/VCFHeaderLineTranslatorUnitTest.java
romanzenka: Ok!

src/test/java/htsjdk/variant/vcf/VCFHeaderLineTranslatorUnitTest.java
romanzenka: I am using a DataProvider for giving inputs together with an "error" field. If the error field is empty, the test was meant to succeed. If the error field is NOT empty, it tells me what kind of error I am expecting to get. That way I am simultaneously testing inputs that are okay, and those that are meant to fail. I could turn the "error" into "errorMessage" for more clarity I guess? Or do you want these sort of tests split into two parts - one for successful ones, one for failures?

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
romanzenka: I added spaces before commas in this code

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
```diff
@@ -114,11 +142,31 @@
...
-                    case ('>') : if (index == valueLine.length()-1) ret.put(key,builder.toString().trim()); break; // if we see a close bracket, and we're at the end, add an entry to our list
-                    case ('=') : key = builder.toString().trim(); builder = new StringBuilder(); break; // at an equals, copy the key and reset the builder
-                    case (',') : ret.put(key,builder.toString().trim()); builder = new StringBuilder(); break; // drop the current key value to the return map
-                    default: builder.append(c); // otherwise simply append to the current string
+                    case '<':
+                        // if we see a open bracket at the beginning, ignore it
+                        if (index == 0) {
+                            break;
+                        }
+                        // GOTCHA: no break here means we fall through to the 'default:' branch and append <
```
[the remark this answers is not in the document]
romanzenka: Okay if you do not mind, I'll fix that so it does not do fancy falthrough 'break'.

cmnbroad:
@romanzenka Thanks for taking this on - we've been wanting to fix it for a while. It looks you've made changes to make "source" and "version" tolerated on read, but the values aren't retained, accessible to callers, or round-tripped on write. Would you be able to add getters so these values can be retrieved from the header lines, and update the encoder to propagate them on write, with corresponding tests ? Ideally we'd implement generalized attribute support, but thats a lot more work. For now it would be ok to handle these two attributes specifically since they're recommended in the spec, but I'd be reluctant to take a change that accepts and then silently drops them.

romanzenka:
> @romanzenka Would you be able to add getters so these values can be retrieved from the header lines, and update the encoder to propagate them on write, with corresponding tests ?

Roundtripping the source/version is low priority - these are optional metadata fields after all. If you drop them - not great, but everything will keep working.

Getting an exception when parsing a valid VCF is a high priority. If you cannot parse VCFs, not even partially, that is a big problem.

I would be more comfortable with having this PR in just to get the big issue fixed. 
Then I can make a separate PR that does full round-trip and all bells and whistles.
Would that be acceptable?

src/test/java/htsjdk/variant/vcf/VCFHeaderLineTranslatorUnitTest.java
```diff
@@ -14,49 +18,147 @@ public void testParseVCF4HeaderLine() {
...
+                {line, "ID", "Description,Optional2", ""},
+                {line, "", "Description,ID", ""},
+                {line, "Description,ID", "", "Tag ID in wrong order"},
+                {line, "Description", "ID", "Optional tag ID must be listed after all expected tags"},
+                {line, "ID,Desc", "", "Unexpected tag Description"},
+                {line, "ID", "Desc", "Unexpected tag Description"},
+                {line, "ID,Description,Extra", "", ""},
+                {"<>", "", "", ""},
+                {"<>", "", "ID,Description", ""},
+                {"<>", "ID", "", ""},
```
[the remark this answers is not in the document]
romanzenka: The empty tag cases <> did pass previously. They would create an output with empty key=value pair. I considered THAT a bug, so I fixed that. You want an exception thrown in this situation?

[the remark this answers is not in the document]
romanzenka: > It would be much cleaner to separate the passing and failing test cases into separate data providers and test methods

I offered that before, seems like you're on a same page here.

src/test/java/htsjdk/variant/vcf/VCFHeaderLineTranslatorUnitTest.java
```diff
@@ -14,49 +18,147 @@ public void testParseVCF4HeaderLine() {
...
+        catch (TribbleException.InvalidHeader e) {
+        }
+
+    }
+
+    @DataProvider(name = "testData")
+    private Object[][] getTestData() {
+        String line = "<ID=X,Description=\"Y\">";
+        return new Object[][]{
+                {line, "ID,Description", "", ""},
```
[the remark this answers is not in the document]
romanzenka: I know, but I did this very specifically to make the test data as readable and easy to reason about as possible. Idea being there's much more test data than there is splitting code.

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
```diff
@@ -114,11 +142,35 @@
...
+                    case '<':
+                        // if we see a open bracket at the beginning, ignore it, otherwise add it
+                        if (index != 0) { 
+                            builder.append(c);
+                        }
+                        break;
+                    case '>': 
+                        if (index == valueLine.length()-1) {
+                            // if we see a close bracket, and we're at the end, add an entry to our list
+                            if (!key.isEmpty()) { // Properly handle empty <> sequence
```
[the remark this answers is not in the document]
romanzenka: I can throw an exception here. Idea was that if you have only optional fields, and none are present - that is legal. This simply fixes a bug when an empty key=value pair was added to the resulting map.

There are a few more scenarios that we may want to think about if you want to nail this down really precisely:

<key=>
<key>
<=>
<=value>
<,key=value>
<key=value,>

src/main/java/htsjdk/variant/vcf/VCFSimpleHeaderLine.java
```diff
@@ -64,7 +64,20 @@ public VCFSimpleHeaderLine(String key, String name, String description) {
...
+     * create a VCF info header line
+     *
+     * @param line      the header line
+     * @param version   the vcf header version
+     * @param key            the key for this header line
+     * @param expectedTagOrdering the tag ordering expected for this header line
+     * @param optionalTags tags that are optional for this header line                            
+     */
+    public VCFSimpleHeaderLine(final String line, final VCFHeaderVersion version, final String key, final List<String> expectedTagOrdering, final List<String> optionalTags) {
+        this(key, VCFHeaderLineTranslator.parseLine(version, line, expectedTagOrdering, optionalTags));
```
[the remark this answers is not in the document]
romanzenka: I was trying to go for minimally invasive action - if it ain't broke, don't fix it. The `VCFContigHeaderLine` and `VCFFilterHeaderLine` do not need optional fields, so they use the old constructor. I can change that.

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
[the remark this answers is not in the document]
romanzenka: I made it fail if you specify expected attributes and you get <>.
It does not fail if all attributes are optional - but nobody calls it that way, so it is the same as if it always failed.

romanzenka:
I have added round-trip code.
There is a test exercising the code paths, but I am not sure where to put a VCF to have it round-trip tested. Is there such a test already?

cmnbroad:
@romanzenka The existing tests are not very well partitioned, but I'd suggest maybe putting the round trip test as a new test in VCFHeaderUnitTest.

romanzenka:
@cmnbroad 
Added a roundtrip test

src/test/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
```diff
@@ -503,6 +516,39 @@ public void testVCFHeaderQuoteEscaping() throws Exception {
...
+                .setOptions(EnumSet.of(Options.ALLOW_MISSING_FIELDS_IN_HEADER, Options.INDEX_ON_THE_FLY))
+                .build();
+        copyWriter.writeHeader(originalHeader);
+        final CloseableIterator<VariantContext> firstCopyVariantIterator = originalFileReader.iterator();
+        while (firstCopyVariantIterator.hasNext()) {
+            VariantContext variantContext = firstCopyVariantIterator.next();
+            copyWriter.add(variantContext);
+        }
+        originalFileReader.close();
+        copyWriter.close();
```
cmnbroad: Use try-with-resources for both of these to avoid manual closing.

src/test/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
```diff
@@ -503,6 +516,39 @@ public void testVCFHeaderQuoteEscaping() throws Exception {
...
+        final File actualFile = File.createTempFile("testVcf4.2roundtrip.", IOUtil.VCF_FILE_EXTENSION);
+        actualFile.deleteOnExit();
+
+        final VariantContextWriter copyWriter = new VariantContextWriterBuilder()
+                .setOutputFile(actualFile)
+                .setReferenceDictionary(createArtificialSequenceDictionary())
+                .setOptions(EnumSet.of(Options.ALLOW_MISSING_FIELDS_IN_HEADER, Options.INDEX_ON_THE_FLY))
+                .build();
+        copyWriter.writeHeader(originalHeader);
+        final CloseableIterator<VariantContext> firstCopyVariantIterator = originalFileReader.iterator();
```
cmnbroad: `firstCopyVariantIterator` -> `inputIterator` or `originalIterator` or something.

src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
```diff
@@ -151,78 +189,126 @@ protected VCFCompoundHeaderLine(String name, VCFHeaderLineCount count, VCFHeader
...
-                (!version.isAtLeastAsRecentAs(VCFHeaderVersion.VCF4_0) && numberStr.equals(VCFConstants.UNBOUNDED_ENCODING_v3))) {
+        }
+        else if ((
+                         version.isAtLeastAsRecentAs(VCFHeaderVersion.VCF4_0) &&
+                         numberStr.equals(VCFConstants.UNBOUNDED_ENCODING_v4)
+                 ) ||
+                 (
+                         !version.isAtLeastAsRecentAs(VCFHeaderVersion.VCF4_0) &&
+                         numberStr.equals(VCFConstants.UNBOUNDED_ENCODING_v3)
+                 )) {
```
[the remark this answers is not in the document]
romanzenka: Gah, IntelliJ. If you published code style settings that I can load into IntelliJ to prevent these issues, that would be a huge time saver + you would get cleaner code.

src/test/java/htsjdk/variant/vcf/VCFFileReaderTest.java
```diff
@@ -69,6 +69,8 @@ public void testIsQueriable(final File vcf, final boolean expectedQueryable) thr
                 {TEST_DATA_DIR + "VcfThatLacksAnIndex.bcf", null, true, false},
                 {TEST_DATA_DIR + "VcfThatLacksAnIndex.vcf.bgz", null, true, false},
 
+                // testing that v4.2 parses Source/Version fields, see issue #517
+                {TEST_DATA_DIR + "Vcf4.2WithSourceVersionInfoFields.vcf", null, false, true}
```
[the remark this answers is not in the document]
romanzenka: I could, but that is not what the VCF spec actually allows. Did I miss some mouseprint somewhere?

src/test/java/htsjdk/variant/vcf/VCFHeaderLineTranslatorUnitTest.java
```diff
@@ -14,49 +17,149 @@ public void testParseVCF4HeaderLine() {
...
+    @Test(dataProvider = "invalidHeaderLines")
+    public void testParseVCF4HeaderLineWithTagsInvalid(final String line,
+                                                     final List<String> expectedTagOrder,
+                                                     final List<String> optionalTags,
+                                                     final String error) {
+        try {
+            callTranslator(line, expectedTagOrder, optionalTags);
+            Assert.fail("Expected failure: '" + error + "', got success");
+        }
+        catch (Exception e) {
```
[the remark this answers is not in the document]
romanzenka: Done

src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
```diff
@@ -277,30 +370,52 @@ public int hashCode() {
...
+     * <p>
+     * These values are optional starting with VCFv4.2. Do not set them for older VCF versions.
+     * There is a single method instead of traditional pair of setters to encourage specification
+     * of both source and version simultaneously. You can pass 'null' as either parameter if not known.
+     *
+     * @param source  annotation source (case-insensitive, e.g. "dbsnp")
+     * @param version exact version (e.g. "138")
+     * @return This line for easier chaining
+     */
+    public VCFCompoundHeaderLine setSourceAndVersion(String source, String version) {
```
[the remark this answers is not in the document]
romanzenka: Split into separate setters + added constructor versions that fill in source/version

src/test/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
romanzenka: I copy&pasted your own code here...

src/test/java/htsjdk/variant/vcf/VCFHeaderUnitTest.java
romanzenka: I do not think this is a good idea. This code is not mine, there is first copy and a second copy happening here, and the `firstCopyVariantIterator` goes with the `firstCopyVCFFile` and `firstCopyWriter`

src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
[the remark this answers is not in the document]
pshapiro4broad: I think using the default intellij settings will work, with the exception of `case` statements, which are indented (vs K&R style where they align with the `switch`). Generally the code uses oracle/sun conventions.

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -212,7 +212,7 @@ protected VCFHeader parseHeaderFromLines( final List<String> headerStrings, fina
                     final VCFContigHeaderLine contig = new VCFContigHeaderLine(str.substring(9), version, VCFConstants.CONTIG_HEADER_START.substring(2), contigCounter++);
                     metaData.add(contig);
                 } else if ( str.startsWith(VCFConstants.ALT_HEADER_START) ) {
-                    final VCFSimpleHeaderLine alt = new VCFSimpleHeaderLine(str.substring(6), version, VCFConstants.ALT_HEADER_START.substring(2), Arrays.asList("ID", "Description"));
+                    final VCFSimpleHeaderLine alt = new VCFSimpleHeaderLine(str.substring(6), version, VCFConstants.ALT_HEADER_START.substring(2), Arrays.asList("ID", "Description"), Collections.emptyList());
```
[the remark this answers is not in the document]
cmnbroad: Well, its a bit ambiguous. The spec allows for all "structured" header lines to have optional fields, but you're right in that "Source" and "Version" are recommended specifically only for INFO. Technically, they should be accepted for ALT as well, but since we're still not implementing a fully generalized mechanism for optional attributes, it would probably be misleading to special case them for the other header lines.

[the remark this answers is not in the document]
romanzenka: Okay, I see. I'll update the test file to exercise the valid options and make sure it works.
My understanding was that they want `expected` then `recommended` and only then `optional`, so technically if I parse ALT with Source and Version as `recommended`, I would disallow valid `##ALT<ID,Description,Random,Version,Random2,Source>`

Speaking of which, in my functions I have named Source/Version fields as "optional", while "recommended" is the more appropriate word. I will refactor so there is space for "truly optional".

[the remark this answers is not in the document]
cmnbroad: I think it will be sufficient to just make Source/Version work. Other attributes won't round trip since there is no place to store them, and thats a much bigger change, so we should try to keep this PR constrained to those.

src/test/java/htsjdk/variant/vcf/VCFHeaderLineTranslatorUnitTest.java
```diff
@@ -46,17 +49,102 @@ public void testParseVCF4HeaderLine() {
...
+        };
+    }
+
+    @DataProvider(name = "invalidHeaderLines")
+    private Object[][] getInvalidHeaderLines() {
+        String line = "<ID=X,Description=\"Y\">";
+        return new Object[][]{
+                // to parse, required, recommended, error message
+                {line, Arrays.asList("Description", "ID"), Arrays.asList(), "Tag ID in wrong order"},
+                {line, Arrays.asList("Description"), Arrays.asList("ID"), "Recommended tag ID must be listed after all expected tags"},
```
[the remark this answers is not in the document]
romanzenka: Let me go about it the other way around. I wanted to reuse the same line and play with different inputs. Instead I will use different lines and reasonable looking inputs to get my point across. Will be a bit more verbose, but more readable.

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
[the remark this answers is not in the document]
romanzenka: To my surprise, Source/Version seems to work and round-trip just fine. Even though the fields are not listed as recommended (technically they aren't), the same parsing code is used and if the fields are present, they are whisked into `source` and `version` members and round-tripped just fine. Since these values are not technically "recommended", I am not adding them to the `recommendedTags` list - but it still works fine.

romanzenka:
I think I addressed everything - the only remaining issue is deprecation of those two constructors (that are not used anymore within the library). I'll remove the `@deprecated` for now, it can always be added back later.

src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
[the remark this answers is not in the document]
pshapiro4broad: > I was reverting things like Map<String, String> to Map<String,String>, there are missing { }s around if statements, max line length is not honored, things are indented strangely in places... If you simply run "reformat file" and you get dozens of changes, the development becomes quite cumbersome.

Instead of reformatting an entire file, select the new/modified code and use `Code/Reformat Code`. If existing code isn't formatted correctly but isn't being modified by your PR, you can leave it for now.

> I would recommend taking default IntelliJ settings, actually reformat everything to that (you will be surprised to see all the changes), publish the settings (you can export them as a .jar), and then just keep things clean going forward.

It would be nice to fix the formatting of the entire code base, but as yet no-one has taken that work on. In the meantime, we can avoid adding new code that doesn't conform to the conventions, and fix code along the way as it's modified.

> As a reviewer, do not waste time policing coding style, we have tools for that. Just enable people to use those tools.

I agree that it would be better if there's less code style discussions in code reviews. Submitters can help by making sure that any new/modified code follows the code style..
  
While code reviews are a valuable tool for identifying problems, sharing knowledge, etc, they are also a tool to ensure that the code stays in a form that's easy for anyone to maintain. Moving to and/or preserving a consistent coding style is part of that process.

Regarding tooling there have been some suggestions but again, no-one has done the work to enable, say, checkstyle on the codebase. Using something like checkstyle would help a lot but requires an investment to get working and integrated in the build.

src/main/java/htsjdk/variant/vcf/VCFHeaderLineTranslator.java
```diff
@@ -114,11 +142,35 @@
...
+                        builder = new StringBuilder(); 
+                        break; 
+                    case ',':
+                        // drop the current key value to the return map
+                        ret.put(key,builder.toString().trim()); 
+                        builder = new StringBuilder(); 
+                        break; 
+                    default:
+                        // otherwise simply append to the current string
+                        builder.append(c);
```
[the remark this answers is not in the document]
romanzenka: Reverting

src/test/java/htsjdk/variant/vcf/VCFHeaderLineTranslatorUnitTest.java
```diff
@@ -46,17 +50,116 @@ public void testParseVCF4HeaderLine() {
...
+    @DataProvider(name = "invalidHeaderLines")
+    private Object[][] getInvalidHeaderLines() {
+        List<String> idDesc = Arrays.asList("ID", "Description");
+        List<Object> none = Collections.emptyList();
+        List<String> sourceVersion = Arrays.asList("Source", "Version");
+        return new Object[][]{
+                // to parse, expected, recommended, error message
+                {"<Description=\"Y\",ID=X>", idDesc, none, "Tag Description in wrong order (was #1, expected #2)"},
+                {"<ID=X,Desc=\"Y\">", idDesc, none, "Unexpected tag Desc"},
+                {"<>", idDesc, none, "Header with no tags is not supported when there are expected tags"},
```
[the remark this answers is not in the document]
romanzenka: That is not my experience.
I reverted that switch back and the tests still pass.
The main change to the switch was 
a) formatting 
b) handling of embedded <> symbols - (which I do not actually test for, yikes)
c) handling of the empty case <> - (again I do not test for it, the old code will create a map with null -> null entry, the new code created an empty map)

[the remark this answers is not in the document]
cmnbroad: @romanzenka I didn't mean for you to remove all of the new tests, only the one specific *test case* I commented on, because it will no longer pass with the switch statement reverted. All the rest of the tests are a welcome addition.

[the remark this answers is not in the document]
romanzenka: Sorry, my misunderstanding. You made it sound like some other test cases cover this already.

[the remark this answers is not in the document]
romanzenka: Pushed the cases back, made sure they pass

romanzenka:
I added a freebie that you may/may not like - I am tired of the MD5CalculatingOutputStream spuriously messing up my coverage, so I made it deterministic.

lbergelson:
@romanzenka 👍 to fixing that.  Don't worry too much about the coverage though... it's weirdly oversensitive.  We need to change the settings to be more lenient.

romanzenka:
@lbergelson I do not think it is even sensitivity - there is something random about how the coverage gets compared. The latest report makes no sense to me - I have not done anything with those 21 reported files. I suspect the code coverage tool itself might be confused?
What you would want to compare is coverage as it is right now versus coverage as it would be after you merge the PR. I do not think that is happening.

lbergelson:
Yeah, it choose funny comparisons sometimes and I'm not certain exactly what happens.  I think it compares the coverage in the last run of master vs the coverage in the pr build, which isn't necessarily in sync although it should be close.

romanzenka:
@lbergelson I am simply pulling from master for now to see what happens... my bad, that's probably best practice before accepting a PR anyway

romanzenka:
@cmnbroad Do you think this could be merged before Friday 2019-01-25? We are releasing an app and it would be nice to build it on top of an official commit instead of our own branch.
I am not hoping for an official release, although that would of course be awesome!

src/test/java/htsjdk/samtools/util/Md5CalculatingOutputStreamTest.java
```diff
@@ -0,0 +1,57 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2014 The Broad Institute
```
cmnbroad: Hm - thats a bit out of date.

romanzenka: I copy&pasted this from elsewhere in the project, can find a newer one.

romanzenka: Ha, so the issue with headers is a bit more severe - I saw several somewhat different styles in files. So I simply bumped the year to 2019.

cmnbroad:
@romanzenka Thanks. Add yes, we should be able to get this merged in the next 24 hours.

lbergelson:
@cmnbroad @romanzenka 👍 

Thanks for working through this one.  It turned out to be more work than I expected.  I don't know if we'll be able to get another release out by friday.  I'll see what I can do...

romanzenka:
> @cmnbroad @romanzenka 👍
> 
> Thanks for working through this one. It turned out to be more work than I expected. I don't know if we'll be able to get another release out by friday. I'll see what I can do...

Don't worry about it. Release when it is appropriate to release. We'll build off a particular commit.

cmnbroad:
Thanks again for working through all the issues @romanzenka.

PR #1252, opened 2019-01-04
Makes IntervalKeepPairFilter filter out single ended reads

meganshand:
### Description

When a BAM contains both single ended and paired ended reads this filter throws an exception (`java.lang.IllegalStateException: Inappropriate call if not paired read`). Since the filter is only meant to keep paired reads it now also filters out single ended reads and doesn't throw an exception.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/test/java/htsjdk/samtools/filter/IntervalKeepPairFilterTest.java
+++ b/src/test/java/htsjdk/samtools/filter/IntervalKeepPairFilterTest.java
@@ -48,6 +48,8 @@ public class IntervalKeepPairFilterTest extends HtsjdkTest {
         builder.addFrag("mapped_pair_chr1", 0, 1, false, false, "151M", null, -1, true, false);
         // Supplementary alignment are never kept by the interval filter.
         builder.addFrag("mapped_pair_chr1", 0, 1, false, false, "151M", null, -1, false, true);
+        // Single ended read should never be kept by the interval filter.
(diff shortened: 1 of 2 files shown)


meganshand:
@lbergelson could you please take a look at this?

src/test/java/htsjdk/samtools/filter/IntervalKeepPairFilterTest.java
```diff
@@ -96,6 +98,21 @@ public void testNotPrimaryReads() {
...
 
+    @Test
+    public void testSingleEndedReads() {
+        final List<Interval> intervalList = new ArrayList<>();
+        final Interval interval1 = new Interval("chr1", 1, 999);
+        intervalList.add(interval1);
+
+        final IntervalKeepPairFilter filter = new IntervalKeepPairFilter(intervalList);
+
+        boolean singleEnded = StreamSupport.stream(builder.spliterator(), false)
```
lbergelson: It looks like this might be cleaner if you use `builder.getRecords().stream()` instead of this spliterator path.

Also, instead of assertingFalse it might be cleaner to use `noneMatch` and have one less inversion.

lbergelson:
On further inspection you're just copying the pattern around it... maybe we can fix them all in a separate PR.

lbergelson:
@meganshand Thank you for this!

PR #1255, opened 2019-01-09
Fix to ensure that SeekableStream#available() never returns a negative value

tomwhite:
### Description

For files that are larger than 2GB (e.g. an uncompressed FASTA) it is possible that
`length - position` in `SeekableStream#available()` overflows an int, resulting in a negative value being returned.

This fixes the problem described here: https://github.com/broadinstitute/gatk/issues/5547

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/seekablestream/SeekableStream.java
+++ b/src/main/java/htsjdk/samtools/seekablestream/SeekableStream.java
@@ -107,8 +107,13 @@ public abstract class SeekableStream extends InputStream {
             return 0;
         }
         final long remaining = length() - position();
-        // the remaining might be negative if the length is not available (0)
-        return (remaining < 0) ? 0 : (int) remaining;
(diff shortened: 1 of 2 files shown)


tomwhite:
@cmnbroad @lbergelson can you review please?

src/main/java/htsjdk/samtools/seekablestream/SeekableStream.java
```diff
@@ -107,8 +107,13 @@ public int available() throws IOException {
             return 0;
```
lbergelson: This method is cursed....  @droazen

lbergelson:
I really wish that java would throw exceptions on overflow...

lbergelson:
I wonder if there are any static analysis tools that would have pointed this out as a probable error...

PR #1256, opened 2019-01-09
Immutable CRAIEntry

jmthibault79:
### Description

Code cleanup for CRAIEntry

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/cram/CRAIEntry.java
+++ b/src/main/java/htsjdk/samtools/cram/CRAIEntry.java
@@ -7,27 +7,36 @@ import htsjdk.samtools.util.RuntimeIOException;
 
 import java.io.IOException;
 import java.io.OutputStream;
-import java.util.ArrayList;
-import java.util.Comparator;
-import java.util.List;
+import java.util.*;
+import java.util.stream.Collectors;
 
 /**
  * A class representing CRAI index entry: file and alignment offsets for each slice.
  * Created by vadim on 10/08/2015.
  */
-public class CRAIEntry implements Comparable<CRAIEntry>, Cloneable {
-    public int sequenceId;
-    public int alignmentStart;
-    public int alignmentSpan;
-    public long containerStartOffset;
-    public int sliceOffset;
-    public int sliceSize;
-    public int sliceIndex;
+public class CRAIEntry implements Comparable<CRAIEntry> {
+    private final int sequenceId;
+    private final int alignmentStart;
+    private final int alignmentSpan;
+    private final long containerStartOffset;
+    private final int sliceByteOffset;
+    private final int sliceByteSize;
 
-    private static int CRAI_INDEX_COLUMNS = 6;
-    private static String entryFormat = "%d\t%d\t%d\t%d\t%d\t%d";
+    private static final int CRAI_INDEX_COLUMNS = 6;
+    private static final String ENTRY_FORMAT = "%d\t%d\t%d\t%d\t%d\t%d";
 
-    public CRAIEntry() {
+    public CRAIEntry(final int sequenceId,
+                     final int alignmentStart,
+                     final int alignmentSpan,
+                     final long containerStartOffset,
+                     final int sliceByteOffset,
+                     final int sliceByteSize) {
+        this.sequenceId = sequenceId;
+        this.alignmentStart = alignmentStart;
+        this.alignmentSpan = alignmentSpan;
+        this.containerStartOffset = containerStartOffset;
+        this.sliceByteOffset = sliceByteOffset;
+        this.sliceByteSize = sliceByteSize;
     }
 
     /**
@@ -36,7 +45,7 @@ public class CRAIEntry implements Comparable<CRAIEntry>, Cloneable {
      * @param line string formatted as a CRAI index entry
      * @throws CRAIIndex.CRAIIndexException
      */
-    public CRAIEntry(final String line) throws CRAIIndex.CRAIIndexException {
+    public CRAIEntry(final String line) throws CRAIIndex.CRAIIndexException  {
         final String[] chunks = line.split("\t");
         if (chunks.length != CRAI_INDEX_COLUMNS) {
             throw new CRAIIndex.CRAIIndexException(
@@ -48,8 +57,8 @@ public class CRAIEntry implements Comparable<CRAIEntry>, Cloneable {
             alignmentStart = Integer.parseInt(chunks[1]);
             alignmentSpan = Integer.parseInt(chunks[2]);
             containerStartOffset = Long.parseLong(chunks[3]);
-            sliceOffset = Integer.parseInt(chunks[4]);
-            sliceSize = Integer.parseInt(chunks[5]);
+            sliceByteOffset = Integer.parseInt(chunks[4]);
+            sliceByteSize = Integer.parseInt(chunks[5]);
         } catch (final NumberFormatException e) {
             throw new CRAIIndex.CRAIIndexException(e);
         }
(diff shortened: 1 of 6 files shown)


src/main/java/htsjdk/samtools/cram/CRAIEntry.java
```diff
@@ -7,27 +7,36 @@
...
  * Created by vadim on 10/08/2015.
  */
-public class CRAIEntry implements Comparable<CRAIEntry>, Cloneable {
-    public int sequenceId;
-    public int alignmentStart;
-    public int alignmentSpan;
-    public long containerStartOffset;
-    public int sliceOffset;
-    public int sliceSize;
-    public int sliceIndex;
```
jmthibault79: Removed sliceIndex because it can't be represented in CRAI.  We were incorrectly using this uninitialized field to populate Slices.

src/main/java/htsjdk/samtools/cram/CRAIEntry.java
```diff
@@ -73,30 +82,18 @@ public void writeToStream(OutputStream os) {
...
-            e.containerStartOffset = s.containerOffset;
-            e.sliceOffset = container.landmarks[i];
-            e.sliceSize = s.size;
-
-            e.sliceIndex = i;
-            entries.add(e);
-        }
-        return entries;
+        return Arrays.stream(container.slices)
+                .map(slice -> slice.getCRAIEntry(slice.containerOffset))
```
jmthibault79: Can't necessarily derive the container offset from the Slice because sometimes it conflict's with the Container's own offset - will need to look into why.

src/main/java/htsjdk/samtools/cram/CRAIEntry.java
```diff
@@ -114,19 +111,6 @@ public int compareTo(final CRAIEntry o) {
         return (int) (containerStartOffset - o.containerStartOffset);
     }
 
-    @Override
-    public CRAIEntry clone() throws CloneNotSupportedException {
```
jmthibault79: only tests were using this

src/main/java/htsjdk/samtools/cram/CRAIIndex.java
```diff
@@ -55,63 +46,30 @@ public void writeIndex(final OutputStream os) {
...
-                    e.sequenceId = c.sequenceId;
-                    e.alignmentStart = s.alignmentStart;
-                    e.alignmentSpan = s.alignmentSpan;
-                    e.containerStartOffset = c.offset;
-                    e.sliceOffset = c.landmarks[i];
-                    e.sliceSize = s.size;
-                    e.sliceIndex = i;
-
-                    entries.add(e);
+                    final Map<Integer, AlignmentSpan> spans = s.getMultiRefAlignmentSpans(container.header, ValidationStringency.DEFAULT_STRINGENCY);
```
jmthibault79: equivalent to what `getCRAIEntriesForMultiRefSlice()` did

src/test/java/htsjdk/samtools/SamIndexesTest.java
```diff
@@ -71,13 +69,7 @@ public void testCraiInMemory() throws IOException {
...
         header.setSortOrder(SAMFileHeader.SortOrder.coordinate);
         CRAMCRAIIndexer indexer = new CRAMCRAIIndexer(baos, header);
-        final CRAIEntry entry = new CRAIEntry();
-        entry.sequenceId = 0;
-        entry.alignmentStart = 1;
-        entry.alignmentSpan = 2;
-        entry.sliceOffset = 3;
-        entry.sliceSize = 4;
-        entry.containerStartOffset = 5;
+        final CRAIEntry entry = CRAIEntryTest.newEntry(0, 1, 2, 5, 3, 4);
```
jmthibault79: preserves the values set above rather than the order

src/main/java/htsjdk/samtools/cram/CRAIEntry.java
cmnbroad: There are still references to slice.index in indexing code. Maybe for the BAI ?

src/main/java/htsjdk/samtools/cram/CRAIEntry.java
cmnbroad: Also, any reason not to move this method to container `getCRAIEntries` just as it is, and have  it return the list of entries. Eventually it should end up there I think, and would be similar to the new one you added to `Slice`.

src/main/java/htsjdk/samtools/cram/CRAIIndex.java
```diff
@@ -128,16 +86,16 @@ public static SeekableStream openCraiFileAsBaiStream(final InputStream indexStre
...
+            slice.alignmentSpan = entry.getAlignmentSpan();
+            slice.sequenceId = entry.getSequenceId();
+            // NOTE: the recordCount and sliceIndex fields can't be derived from the CRAM index
+            // so we can only set them to zero
+            // see https://github.com/samtools/htsjdk/issues/531
+            slice.nofRecords = 0;
+            slice.index = 0;
+            slice.offset = entry.getSliceByteOffset();
 
             indexer.processSingleReferenceSlice(slice);
```
cmnbroad: I was just starting to come around to your thinking about IndexableSlice, but the more I look at this, I wonder if this indexing code should be working right off of the CRAIEntry, instead of creating a partial slice, and then using the exact subset of slice that was populated from the CRAIEntry ? I wonder if we can cut out the slice altogether ?

cmnbroad: Just to clarify, in this case its converting to BAI, so there may be no need to use a slice for that. There are probably still other cases where we want to create an index from a real slice, in which case we'd use the real thing.

src/main/java/htsjdk/samtools/cram/CRAIEntry.java
jmthibault79: Yes, it's still used for BAI

src/main/java/htsjdk/samtools/cram/CRAIEntry.java
jmthibault79: OK, these suggestions make sense.

Slice.containerOffset is the byte offset of the Container from the beginning of the stream.

Slice.offset is the byte offset of the Slice from the beginning of its Container.

I'll add comments making this more clear to `Slice`

src/main/java/htsjdk/samtools/cram/structure/Slice.java
```diff
@@ -291,4 +292,15 @@ public CramRecordReader createCramRecordReader(final CompressionHeader header,
...
 
+    /**
+     * Generate a CRAI Index entry from this Slice and the container offset.
+     *
+     * TODO: investigate why we can't simply use the Slice's own containerOffset here
+     *
+     * @param containerStartOffset the byte offset of this Slice's Container
+     * @return a new CRAI Index Entry
+     */
+    public CRAIEntry getCRAIEntry(final long containerStartOffset) {
```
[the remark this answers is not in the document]
jmthibault79: I'm not sure that we can before we work through the "Slice Types" spec questions - it's not clear to me whether unmapped Slices should be indexed, for example.

src/main/java/htsjdk/samtools/cram/CRAIEntry.java
```diff
@@ -2,32 +2,40 @@
...
-    public int sliceIndex;
-
-    private static int CRAI_INDEX_COLUMNS = 6;
-    private static String entryFormat = "%d\t%d\t%d\t%d\t%d\t%d";
-
-    public CRAIEntry() {
+public class CRAIEntry implements Comparable<CRAIEntry> {
+    private final int sequenceId;
+    private final int alignmentStart;
+    private final int alignmentSpan;
```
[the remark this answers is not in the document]
jmthibault79: AlignmentSpan has a record count, and CRAIEntry does not.  I think that has the potential to be misleading.

src/main/java/htsjdk/samtools/cram/CRAIIndex.java
jmthibault79: In my Slice PR I have `SliceBAIMetadata` playing this role: essentially just `CRAIEntry` with the necessary addition of `nofRecords` and `index`.

lbergelson:
@cmnbroad @jmthibault79 👍 Thank you.

PR #1258, opened 2019-01-10
Only compute BIN on BAM write and on index building

tfenne:
### Description

Prior to this change if you had a SAM or BAM file with records mapped to references > 512mb, SAMRecord would log a warning per record mapped to those contigs!!  It would also blindly compute an invalid BIN value and store it.

This change removes the stored `indexingBin` field in `SAMRecord` and has it computed only when needed - i.e. when writing a BAM file in `BAMRecordCodec` and during index building.  In `BAMRecordCoded` it will warn once when the first read mapped to a long contig is found, and will write a bin value of `0`, the same as we do for unmapped reads.  When index building it will throw an exception since BAI cannot support reads mapped beyond 512mb.

I ran into a lot of issues with how this interacted with the long cigar code.  @yfarjoun I ended up resolving those by modifying the code in `BAMRecordCodec` slightly.  Instead of calling `setCigar()` with the sentinal cigar, it creates a local variable called `cigarToWrite` and assigns either the real cigar or the sentinal cigar as appropriate.  I also added a line at the end of the `encode()` to remove the `CG` tag from the record post-encoding, so that a pass through `BAMRecordCodec` now leaves the record largely untouched.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1589,26 +1570,10 @@ int computeIndexingBin() {
...
-        return GenomicIndexUtil.regionToBin(alignmentStart, alignmentEnd);
-    }
 
-
-    /**
-     * Gets indexing bin if it is present otherwise compute it.
-     * @param alignment to compute indexing bin for.
-     * @return indexing bin.
-     */
-    public int computeIndexingBinIfAbsent(final SAMRecord alignment) {
```
yfarjoun: change in public API...perhaps deprecate first?

src/main/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -1254,7 +1254,7 @@ public static boolean isValidUnsignedIntegerAttribute(long value) {
      * Checks if reference sequence is compatible with BAI indexing format.
      * @param sequence reference sequence.
      */
-    public static boolean isReferenceSequenceCompatibleWithBAI(final SAMSequenceRecord sequence) {
```
yfarjoun: Another change in public API....perhaps keep the old, misleading name around (deprecated), with a comment?

src/test/java/htsjdk/samtools/BAMFileWriterTest.java
```diff
@@ -342,7 +339,7 @@ public void testBinNotNullWhenLargeCigarIsLoaded(final int numOps) throws Except
         try (final SamReader reader = SamReaderFactory.makeDefault().validationStringency(ValidationStringency.SILENT).open(bamFile)) {
             reader.iterator().forEachRemaining(samRecord -> {
                 samRecord.getCigar();
-                Assert.assertNotNull(samRecord.getIndexingBin());
+                samRecord.computeIndexingBin();
```
yfarjoun: you removed the assertion...what does this check now?

src/test/java/htsjdk/samtools/CRAMComplianceTest.java
```diff
@@ -294,9 +293,6 @@ public void testBAMThroughCRAMRoundTrip() throws IOException, NoSuchAlgorithmExc
         // retrieve all records from the bam and reset the indexing bins to keep comparisons with
```
yfarjoun: update the comment?

src/test/java/htsjdk/samtools/CRAMComplianceTest.java
```diff
@@ -327,9 +323,6 @@ public void testBAMThroughCRAMRoundTripViaPath() throws IOException, NoSuchAlgor
...
```
yfarjoun: update comment

src/test/java/htsjdk/samtools/BAMFileWriterTest.java
yfarjoun: (add a comment if there might be a thrown exception that you are checking for)

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1564,23 +1553,15 @@ public SAMTagAndValue(final String tag, final Object value) {
...
      * @return indexing bin based on alignment start & end.
      */
     int computeIndexingBin() {
+        if (getAlignmentStart() > GenomicIndexUtil.BIN_GENOMIC_SPAN || getAlignmentEnd() > GenomicIndexUtil.BIN_GENOMIC_SPAN) {
+            throw new IllegalStateException("Read position too high for BAI bin indexing.");
+        }
+
         // regionToBin has zero-based, half-open API
         final int alignmentStart = getAlignmentStart()-1;
         int alignmentEnd = getAlignmentEnd();
```
yfarjoun: Move these two lines up so that getAlignmentStart() and getAlignmentEnd() aren't called twice.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -1564,23 +1553,15 @@ public SAMTagAndValue(final String tag, final Object value) {
...
-    /**
-     * Used internally when writing BAMRecords.
-     * @param mIndexingBin c.f. http://samtools.sourceforge.net/SAM1.pdf
-     */
-    void setIndexingBin(final Integer mIndexingBin) {
-        this.mIndexingBin = mIndexingBin;
-    }
-
     /**
      * Does not change state of this.
```
yfarjoun: update comment to reflect the fact that there is a possible side-effect due to the need to parse the cigar.

tfenne: Done.

src/main/java/htsjdk/samtools/SAMRecord.java
tfenne: Done.

src/main/java/htsjdk/samtools/SAMRecord.java
tfenne: Done.  Its such a horrible method and should never have been public in the first place, but just in case.

src/main/java/htsjdk/samtools/SAMUtils.java
tfenne: Done.

src/test/java/htsjdk/samtools/BAMFileWriterTest.java
tfenne: I have no real idea what it was checking before - the only thing it can have been checking is that a value got returned.  Since we don't store it any more but calculate it on demand, the return type of `computeIndexingBin()` is `int`, so null checking didn't seem appropriate.  But the method will throw an exception if it can't be calculated for some reason.  So it's testing that it can be generated.  Much like the `samRecord.getCigar()` directly above it that someone else (ahem) wrote ;)

src/test/java/htsjdk/samtools/CRAMComplianceTest.java
tfenne: I've just removed the comment as all that line now does is load the records from a file and that didn't seem comment-worthy.

src/test/java/htsjdk/samtools/CRAMComplianceTest.java
tfenne: Ditto

yfarjoun:
@lbergelson  case to cast a glance?

tfenne:
@lbergelson I'd be very grateful if you could let me know, in the next few days, whether you'd like to review this PR or if you're happy enough with @yfarjoun's review.  Thanks!

src/main/java/htsjdk/samtools/BAMRecordCodec.java
```diff
@@ -108,18 +108,20 @@ public void encode(final SAMRecord alignment) {
         final int readLength = alignment.getReadLength();
 
         // if cigar is too long, put into CG tag and replace with sentinel value
-        if (alignment.getCigarLength() > BAMRecord.MAX_CIGAR_OPERATORS) {
+        final Cigar cigarToWrite;
+        final boolean cigarSwitcharoo = alignment.getCigar().numCigarElements() > BAMRecord.MAX_CIGAR_OPERATORS;
```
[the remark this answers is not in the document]
tfenne: The short answer is that some of the tests fail if I leave it as `getCigarLength()` and they all pass if I just call `getCigar().numCigarElements()`.  I'm honestly not 100% sure why and that code has gotten horribly complicated with all the long cigar support.  I'll add a comment.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2112,27 +2080,6 @@ public int hashCode() {
...
         }
 
-        if (this.getAlignmentStart() != NO_ALIGNMENT_START) {
-            final SAMSequenceRecord sequence = getHeader() != null ? getHeader().getSequence(getReferenceName()) : null;
-            if (sequence != null && SAMUtils.isReferenceSequenceCompatibleWithBAI(sequence)) {
-                if (getValidationStringency() != ValidationStringency.SILENT) {
-                    if (getReferenceName() != null) {
-                        LOG.warn(String.format("%s reference length is too large for BAM bin field. %s record bin field value is incorrect.",
-                                getReferenceName(), getReadName()));
-                    } else {
```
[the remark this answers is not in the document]
tfenne: This is actually my primary reason for the PR in the first place!  With this check here, unless you set validation stringency to silent, you get a warning logged _per record_ when reading.  So I'm getting millions of warnings!  
My take is that it's perfectly valid SAM/BAM to have records mapped to references that are longer than BAI can handle, but it's problematic when you try to index with BAI.  That's why I removed the warnings, and made sure it was a clear failure when building a BAI.  If I leave this in, then anyone using either unindexed BAMs or CSI-indexed BAMs will see millions of unnecessary warnings.

src/main/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -1250,11 +1250,19 @@ public static boolean isValidUnsignedIntegerAttribute(long value) {
         return alignments;
     }
 
+    /**
+     * @deprecated because the method does the exact opposite of what it says.  Use the correctly named
+     *             isReferenceSequenceIncompatibleWithBAI() instead.
+     */
+    @Deprecated public static boolean isReferenceSequenceCompatibleWithBAI(final SAMSequenceRecord sequence) {
```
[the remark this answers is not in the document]
tfenne: Indeed.  I was really confused initially.

src/main/java/htsjdk/samtools/SAMRecord.java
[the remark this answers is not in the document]
lbergelson: Oh.. of course.. that makes sense.  Maybe there's a way we can have ValidateSam warn exactly once if it encounters reads like that but not have it happen in the general case of just processing a sam file with validation on.

src/main/java/htsjdk/samtools/BAMRecordCodec.java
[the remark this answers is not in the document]
lbergelson: That scares the hell out of me...  those seem like they should be identical..

src/test/java/htsjdk/samtools/BAMCigarOverflowTest.java
```diff
@@ -24,6 +24,6 @@ public void testCigarOverflow() throws Exception {
 
         //The BAM file that exposed the bug triggered a SAM validation error because the bin field of the BAM record did not equal the computed value. Here we test for this error.
         //Cast to int to avoid an ambiguity in the assertEquals() call between assertEquals(int,int) and assertEquals(Object,Object).
-        assertEquals(testBAMRecord.computeIndexingBin(), (int) testBAMRecord.getIndexingBin());
+        assertEquals(testBAMRecord.computeIndexingBin(), (int) 0);
```
pshapiro4broad: cast and its associated comment no longer necessary
```suggestion
        assertEquals(testBAMRecord.computeIndexingBin(), 0);
```

src/test/java/htsjdk/samtools/BAMFileWriterTest.java
pshapiro4broad: The way it's written now the test will pass even if `computeIndexingBin()` is never called, e.g. if `reader.iterator()` returns an empty stream. Is there a way to improve this?

> the method will throw an exception if it can't be calculated for some reason

Are there any negative tests that ensure that this will happen?

src/main/java/htsjdk/samtools/BAMRecordCodec.java
```diff
@@ -223,15 +231,21 @@ public static Cigar makeSentinelCigar(final Cigar cigar) {
...
         final SAMSequenceRecord sequence = rec.getHeader() != null ? rec.getHeader().getSequence(rec.getReferenceName()) : null;
-        if (!isReferenceSizeWarningShowed
-                && sequence != null
-                && SAMUtils.isReferenceSequenceCompatibleWithBAI(sequence)
-                && rec.getValidationStringency() != ValidationStringency.SILENT) {
-            LOG.warn("Reference length is too large for BAM bin field. Values in the bin field could be incorrect.");
+        final boolean tooLarge = sequence != null && SAMUtils.isReferenceSequenceIncompatibleWithBAI(sequence);
+        if (!isReferenceSizeWarningShowed & tooLarge && rec.getValidationStringency() != ValidationStringency.SILENT) {
+            LOG.warn("Reference length is too large for BAM bin field.");
+            LOG.warn("Reads on references longer than " + GenomicIndexUtil.BIN_GENOMIC_SPAN + "bp will have bin set to 0.");
```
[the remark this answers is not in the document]
tfenne: This is intentional.  It's normal to write `1234bp` vs. `1234 bp`.

PR #1262, opened 2019-01-23
Rename LogLevel.WARNING to LogLevel.WARN.

### LINKED ISSUE #1261: Non-standard logging level WARNING prevents downstream configuration
`htsjdk.samtools.util.Log.WARNING` is a non-standard logging level and should be renamed `htsjdk.samtools.util.Log.WARN`.  Or better yet, `htsjdk.samtools.util.Log` should be removed altogether.

As is, downstream configuration using slf4j `jcl-over-slf4j` cannot silence these WARNING log statements.

heuermh:
Fixes #1261

### DIFF
--- a/src/main/java/htsjdk/samtools/util/Log.java
+++ b/src/main/java/htsjdk/samtools/util/Log.java
@@ -41,7 +41,7 @@ import java.util.Date;
  */
 public final class Log {
     /** Enumeration for setting log levels. */
-    public enum LogLevel { ERROR, WARNING, INFO, DEBUG }
+    public enum LogLevel { ERROR, WARN, INFO, DEBUG }
 
     private static LogLevel globalLogLevel = LogLevel.INFO;
     private static PrintStream out = System.err;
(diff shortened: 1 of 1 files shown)


heuermh:
Ping @lbergelson for review.  This is helpful for Disq and ADAM and other downstream tools.

lbergelson:
@heuermh Sorry, lots of stuff built up here that I haven't gotten a chance to review yet. 

This is a breaking change that would likely cause a lot of people to have to patch their source code so I want to understand the need for it.  

I'm not sure I understand the benefit of this.  If this is renamed will it enable automatic integration with slf4j?  Is it just to standardize the language?  Log4j / slf4j use `WARN`, but `java.util.logging` defines `WARNING` so it's not exactly an abberation.

lbergelson:
We could do this rename seemlessly without breaking anyone by deprecating the `WARNING` level and then making the log priority explicit instead of the ordinal.  Then we can remove uses of WARN in htsjdk itself without affecting downstream projects that rely on this enum.

yfarjoun:
closing in lieu of #1270. please reopen if that is not a good replacement.

PR #1264, opened 2019-01-23
Fixed memory leak in VCFCodec

d-cameron:
### Description

VCFCodec caches previously seen allele strings which I presume is so that the millions of identical allele strings (e.g A, AA, AT, T) are weakly interned in the string HashMap<>. Unfortunately, when parsing large VCF with SV alleles, this causes a memory leak.

Since each allele string is retained until the VCFCodec is GCed, streaming a VCF containing SVs (e.g just counting # records) results in a continual increase in memory usage even when the VCF records are no longer used. For SVs, the allele string can easily be thousands of bytes in size.

Changes:
- Excluding breakpoint, single breakend allele strings from the cache.
- Excluding ALT allele strings containing multiple alleles from the cache (String.split() gets called so there's no point in caching these at all)
- I also put a safety check to exclude long alleles from the cache since they're very likely to be unique and not benefit from the caching at all.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/test/java/htsjdk/variant/variantcontext/AlleleUnitTest.java
+++ b/src/test/java/htsjdk/variant/variantcontext/AlleleUnitTest.java
@@ -277,4 +277,34 @@ public class AlleleUnitTest extends VariantBaseTest {
         Assert.assertEquals("ATCGA", Allele.extend(Allele.create("AT"), "CGA".getBytes()).toString());
         Assert.assertEquals("ATCGA", Allele.extend(Allele.create("ATC"), "GA".getBytes()).toString());
     }
+    @Test
+    public void testWouldBeSymbolic() {
+        Assert.assertTrue(Allele.wouldBeSymbolicAllele("<DEL>".getBytes()));
+        Assert.assertTrue(Allele.wouldBeSymbolicAllele("AAAAAA[chr1:1234[".getBytes()));
+        Assert.assertTrue(Allele.wouldBeSymbolicAllele("AAAAAA]chr1:1234]".getBytes()));
+        Assert.assertTrue(Allele.wouldBeSymbolicAllele("A.".getBytes()));
+        Assert.assertTrue(Allele.wouldBeSymbolicAllele(".A".getBytes()));
+        Assert.assertFalse(Allele.wouldBeSymbolicAllele("AA".getBytes()));
+        Assert.assertFalse(Allele.wouldBeSymbolicAllele("A".getBytes()));
+    }
+    @Test
+    public void testWouldBeBreakpoint() {
+        Assert.assertFalse(Allele.wouldBeBreakpoint("<DEL>".getBytes()));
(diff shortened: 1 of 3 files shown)


d-cameron:
Also performed a minor refactor to removed the duplicated symbolic alllele detection code in Allele and VCFCodec

d-cameron:
Alternative approaches:
- just remove the HashMap entirely and to have a fixed size cache of the small alleles you're likely to encounter. 680 Allele records covers all A/C/G/T ref and alt alleles up to 4bp in length. 2,728 covers up to 5bp. A fixed static cache works better when you're opening multiple files as the existing cache is per-codec anyway.

- String.intern() the short strings

- Replace the map with a `WeakHashMap<String, WeakReference<String>>` so it doesn't leak

lbergelson:
@d-cameron We had a similar proposal a while ago https://github.com/samtools/htsjdk/pull/845 which fell apart because no one could agree about the right thing to do so in the end we did nothing.

However, looking at this more it looks to me that the caching / interning of the allele strings is entirely pointless.  It just gets passed into a call to Allele.create() which then re-encodes the string into bytes.  I can't tell exactly, because the string encoding is complex and varies by local, but it doesn't look to me that it has any sort of string -> byte cache in the encoding stage.  It looks like maybe the right thing to do would be to not bother caching any of the allele strings.  Then in addition we could create a cache of String -> Allele's which would be used to reduce byte[] overhead from many duplicate `AA` alleles...

@yfarjoun Can you take a look and see if you agree with my interpretation of what's happening?  How did we not see that before when we looked at this?

yfarjoun:
I agree, @lbergelson however, I would like to note that the internedString map is also used for filters, reference names and genotypes.  For those objects, the current solution might be OK.

src/test/java/htsjdk/variant/variantcontext/AlleleUnitTest.java
```diff
@@ -277,4 +277,34 @@ public void testExtend() {
         Assert.assertEquals("ATCGA", Allele.extend(Allele.create("AT"), "CGA".getBytes()).toString());
         Assert.assertEquals("ATCGA", Allele.extend(Allele.create("ATC"), "GA".getBytes()).toString());
     }
+    @Test
+    public void testWouldBeSymbolic() {
```
lbergelson: Thank you for adding these tests!

lbergelson:
Hi @d-cameron,

Are you interested in continuing to work on this branch or would you like us to take it over for you?

d-cameron:
@lbergelson happy for you to take it over. The change is relatively small but the code reviewing will be much faster if you only have to do it internally and not bounce back and forward to Australia for each revision. I have a few other projects on the go and since I already have this patch packaged into GRIDSS, it's less urgent that the other work.

Thanks

lbergelson:
Closing this because it's been replace by #1282

PR #1265, opened 2019-01-23
- Make Unique intervals keep both interval names even when intervals …

yfarjoun:
…are the same (assuming names are different)

- added a test

### Description

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


src/main/java/htsjdk/samtools/util/IntervalList.java
```diff
@@ -383,11 +383,12 @@ public void unique(final boolean concatenateNames) {
     /**
      * Merges a sorted collection of intervals and optionally concatenates unique names or takes the first name.
      */
-    static Interval merge(final SortedSet<Interval> intervals, final boolean concatenateNames) {
-        final String chrom = intervals.first().getContig();
-        int start = intervals.first().getStart();
-        int end = intervals.last().getEnd();
-        final boolean neg = intervals.first().isNegativeStrand();
+    static Interval merge(final Iterable<Interval> intervals, final boolean concatenateNames) {
```
lbergelson: We should consider lifting it into Locatable as something like `getSpanningLocatable`.  We'd have to decide on a concrete representation to instantiate though, so maybe that should wait until a different pr.

PR #1266, opened 2019-01-23
CRAM: Container and Slice states

jmthibault79:
### Description

Add tests for Container and Slice states and associated code changes.

Related to the discussion at https://github.com/samtools/hts-specs/pull/372

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


src/main/java/htsjdk/samtools/cram/build/ContainerFactory.java
```diff
@@ -77,116 +65,51 @@ Container buildContainer(final List<CramCompressionRecord> records,
...
                     Math.min(records.size(), i + recordsPerSlice));
-            final Slice slice = buildSlice(sliceRecords, header);
+            final Slice slice = Slice.buildSlice(sliceRecords, header);
             slice.globalRecordCounter = lastGlobalRecordCounter;
             lastGlobalRecordCounter += slice.nofRecords;
             container.bases += slice.bases;
             slices.add(slice);
-
-            // assuming one sequence per container max:
-            if (container.sequenceId == -1 && slice.sequenceId != -1)
```
jmthibault79: moved into `calculateAlignmentBoundaries()`

src/main/java/htsjdk/samtools/cram/build/ContainerFactory.java
```diff
@@ -77,116 +65,51 @@ Container buildContainer(final List<CramCompressionRecord> records,
...
-        container.slices = slices.toArray(new Slice[slices.size()]);
+        container.slices = slices.toArray(new Slice[0]);
         calculateAlignmentBoundaries(container);
 
         globalRecordCounter += records.size();
         return container;
     }
 
-    private static void calculateAlignmentBoundaries(final Container container) {
+    static void calculateAlignmentBoundaries(final Container container) {
```
jmthibault79: to make it testable

src/main/java/htsjdk/samtools/cram/build/ContainerFactory.java
```diff
@@ -77,116 +65,51 @@ Container buildContainer(final List<CramCompressionRecord> records,
...
-            }
-        }
 
-        if (start < Integer.MAX_VALUE) {
-            container.alignmentStart = start;
-            container.alignmentSpan = end - start;
-        }
-    }
-
-    private static Slice buildSlice(final List<CramCompressionRecord> records,
```
jmthibault79: moved to Slice

src/main/java/htsjdk/samtools/cram/build/ContainerParser.java
```diff
@@ -77,7 +77,7 @@ private static void addSpan(final int seqId, final int start, final int span, fi
 
     private static Map<Integer, AlignmentSpan> addAllSpans(final Map<Integer, AlignmentSpan> spanMap, final Map<Integer, AlignmentSpan> addition) {
         for (final Map.Entry<Integer, AlignmentSpan> entry:addition.entrySet()) {
-            addSpan(entry.getKey(), entry.getValue().getStart(), entry.getValue().getCount(), entry.getValue().getSpan(), spanMap);
+            addSpan(entry.getKey(), entry.getValue().getStart(), entry.getValue().getSpan(), entry.getValue().getCount(), spanMap);
```
jmthibault79: yikes!  who knows what subtle bugs this was causing?

src/main/java/htsjdk/samtools/cram/structure/Slice.java
```diff
@@ -246,10 +257,14 @@ void setAttribute(final short tag, final Object value, final boolean isUnsignedA
         }
     }
 
-    public boolean isMapped() {
+    public boolean isMappedSingleRef() {
         return sequenceId > SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX;
```
jmthibault79: alternate impl: `! isUnmapped() && ! isMultiref()`

src/main/java/htsjdk/samtools/cram/structure/Slice.java
```diff
@@ -318,4 +333,113 @@ public CRAIEntry getCRAIEntry() {
...
     }
+
+    /**
+     * Create a Slice from CRAM Compression Records and a Compression Header
+     *
+     * @param records input CRAM Compression Records
+     * @param header the enclosing {@link Container}'s Compression Header
+     * @return a Slice corresponding to the given records
+     */
+    public static Slice buildSlice(final List<CramCompressionRecord> records,
```
jmthibault79: moved here from ContainerFactory.  May change this to a constructor in future Slice refactoring.

src/test/java/htsjdk/samtools/cram/build/ContainerFactoryTest.java
```diff
@@ -21,105 +19,499 @@
  */
 public class ContainerFactoryTest extends HtsjdkTest {
 
-    @Test
-    public void testUnmapped() {
-        SAMFileHeader header = new SAMFileHeader();
+    public static Container getSingleRefContainer(final SAMFileHeader samFileHeader) {
```
jmthibault79: these are public because they're reused by ContainerParserTest

src/test/java/htsjdk/samtools/cram/build/ContainerFactoryTest.java
```diff
@@ -21,105 +19,499 @@
...
+        Assert.assertEquals(container.nofRecords, 10);
+        Assert.assertEquals(container.sequenceId, 0);
+        Assert.assertEquals(container.slices.length, 1);
+        Assert.assertEquals(container.slices[0].sequenceId, container.sequenceId);
+        return container;
+    }
 
-        List<CramCompressionRecord> records = new ArrayList<>();
-        for (int i = 0; i < recordsPerContainer; i++) {
+    public static Container getUnmappedNoRefContainer(final SAMFileHeader samFileHeader) {
```
jmthibault79: There are two ways to be unmapped/unplaced.  No Ref ID is one, no Alignment Start is the other

src/test/java/htsjdk/samtools/cram/build/ContainerFactoryTest.java
```diff
@@ -21,105 +19,499 @@
...
+        slice1.alignmentStart = 10;
+        slice1.alignmentSpan = 15;
+
+        final Slice slice2 = new Slice();
+        slice2.sequenceId = slice1.sequenceId;
+        slice2.alignmentStart = 20;
+        slice2.alignmentSpan = 20;
+
+        final Container container = new Container();
+        container.slices = new Slice[]{slice1, slice2};
```
jmthibault79: a multiple-slice container!

src/test/java/htsjdk/samtools/cram/build/ContainerFactoryTest.java
```diff
@@ -21,105 +19,499 @@
...
+        single.alignmentStart = 10;
+        single.alignmentSpan = 15;
+
+        final Slice unmapped = new Slice();
+        unmapped.sequenceId = SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX;
+        unmapped.alignmentStart = Slice.NO_ALIGNMENT_START;
+        unmapped.alignmentSpan = Slice.NO_ALIGNMENT_SPAN;
+
+        final Container container = new Container();
+        container.slices = new Slice[]{single, unmapped};
```
jmthibault79: can't combine different Slice Types

src/test/java/htsjdk/samtools/cram/build/ContainerParserTest.java
```diff
@@ -28,159 +24,134 @@
 public class ContainerParserTest extends HtsjdkTest {
 
     @Test
-    public void testEOF() throws IOException {
```
jmthibault79: split into 2 tests

src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
```diff
@@ -210,6 +210,17 @@ public void setSegmentUnmapped(final boolean segmentUnmapped) {
         flags = segmentUnmapped ? flags | SEGMENT_UNMAPPED_FLAG : flags & ~SEGMENT_UNMAPPED_FLAG;
     }
 
+    /**
+     * Does this record have a valid placement / alignment location?
+     * It must have both a valid reference sequence ID and alignment start position to qualify.
+     *
+     * @return true is the record is placed
```
cmnbroad: is -> if

src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
```diff
@@ -210,6 +210,17 @@ public void setSegmentUnmapped(final boolean segmentUnmapped) {
         flags = segmentUnmapped ? flags | SEGMENT_UNMAPPED_FLAG : flags & ~SEGMENT_UNMAPPED_FLAG;
     }
 
+    /**
+     * Does this record have a valid placement / alignment location?
+     * It must have both a valid reference sequence ID and alignment start position to qualify.
```
cmnbroad: Might be worth explicitly stating that this returns the placed status independent of mapped status (that is, it can return true even when the record is mapped).

src/main/java/htsjdk/samtools/cram/structure/Slice.java
```diff
@@ -246,10 +257,14 @@ void setAttribute(final short tag, final Object value, final boolean isUnsignedA
         }
     }
 
-    public boolean isMapped() {
+    public boolean isMappedSingleRef() {
         return sequenceId > SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX;
     }
 
+    public boolean isUnmapped() {
```
cmnbroad: This could use some comments on the possible interpretations (only unmapped records, or mixed, etc), especially since it uses a value from SAMRecord sequence ID state, but it doesn't mean the same thing as `SAMRecord` `unmapped` flag. Ultimately it seems like `Slice` and `Container` could each have a single state variable, possibly drawn from (separate ?) enums, that dictate which state they're in.

src/main/java/htsjdk/samtools/cram/structure/Slice.java
```diff
@@ -318,4 +333,113 @@ public CRAIEntry getCRAIEntry() {
...
+            // @formatter:on
+
+            // Valid Slice states, by individual record contents:
+            //
+            // Single Reference: all records have valid placements/alignments on the same reference sequence
+            //      * records can be unmapped-but-placed
+            //      * reference can be external or embedded
+            // Multiple Reference: records may be placed or not, and may have differing reference sequences
+            //      * reference must not be embedded (not checked here)
+            // Unmapped: all records are unmapped and unplaced
```
cmnbroad: This is the definition of the "slice state" enum and state variable I was asking about above. Should there just be an enum for these values, and a single state variable that changes as needed ? And also one for container, perhaps with slightly different states. That would codify the whole state thing nicely.

src/test/java/htsjdk/samtools/cram/build/ContainerFactoryTest.java
cmnbroad: These get methods are `public static`, and are not annotated as tests, but have asserts. Its a little hard to tell whats test code and whats utility code in here - is it possible to separate them ?

cmnbroad: Tthere is a lot of code that is duplicated many times in this file (ie., the code that creates `CramCompressionRecord`s) that could be factored out. 

Some of the utility code might be something we can move to CramTestUtils if its useful across tests.

src/main/java/htsjdk/samtools/cram/build/ContainerFactory.java
```diff
@@ -77,116 +65,51 @@ Container buildContainer(final List<CramCompressionRecord> records,
...
-                } else if (record.alignmentStart != SAMRecord.NO_ALIGNMENT_START) {
-                    minAlStart = Math.min(record.alignmentStart, minAlStart);
-                    maxAlEnd = Math.max(record.getAlignmentEnd(), maxAlEnd);
-                }
+        for (final Slice slice : container.slices) {
+            if (slice.sequenceId != container.sequenceId) {
+                final String msg = String.format(
+                        "Slices in Container have conflicting reference sequences: %d and %d ",
+                        slice.sequenceId, container.sequenceId);
+                throw new CRAMException(msg);
```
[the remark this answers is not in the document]
jmthibault79: The github diff is confusing here.  This code section *is* the part of calculateAlignmentBoundaries that ensures this.  Let's Slack about this.

src/main/java/htsjdk/samtools/cram/build/ContainerFactory.java
jmthibault79: Excellent points.  Added commit 5447ca665b9f87db9355b2ccefb233e82f13b915 for this.

src/main/java/htsjdk/samtools/cram/structure/Slice.java
jmthibault79: Yes - the intention here is that this list of records all goes into the same slice.

`buildContainer()` handles dividing the list of records into slices, calling this once per result of that division.

I can add a comment about this.

src/test/java/htsjdk/samtools/cram/build/ContainerParserTest.java
jmthibault79: yep, will do

src/test/java/htsjdk/samtools/cram/build/ContainerParserTest.java
```diff
@@ -10,177 +10,144 @@
...
 
-        ContainerParser parser = new ContainerParser(samFileHeader);
         final Map<Integer, AlignmentSpan> referenceSet = parser.getReferences(container, ValidationStringency.STRICT);
-        Assert.assertNotNull(referenceSet);
-        Assert.assertEquals(referenceSet.size(), 1);
-        Assert.assertTrue(referenceSet.containsKey(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX));
+        Assert.assertEquals(referenceSet.keySet(), expectedKeys);
 
+        // TODO this fails.  return to this when refactoring Container and CramCompressionRecord
+        //
```
cmnbroad: Any idea why ? For now it might be worth adding an assert that the list sizes match (hopefully....)

src/test/java/htsjdk/samtools/cram/build/ContainerFactoryTest.java
```diff
@@ -20,115 +18,257 @@
...
+        header.addSequence(new SAMSequenceRecord("8", 10));
+        header.addSequence(new SAMSequenceRecord("9", 10));
+        return header;
+    }
+
+    static SAMFileHeader getSAMFileHeaderForTests() {
+        return header;
+    }
+
+    static List<CramCompressionRecord> getSingleRefRecords() {
```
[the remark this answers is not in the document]
jmthibault79: Good idea

src/test/java/htsjdk/samtools/cram/build/ContainerParserTest.java
jmthibault79: I think it's due to CramNormalizer?  There's a lot going on there which I don't understand.

More generally, there are many ways to construct a CramCompressionRecord which don't fulfill the requirements of various consumers.  I think some of these are rewriting the records to be more compliant.

Sizes do match, fortunately.

PR #1269, opened 2019-01-28
Test on Java 11

tomwhite:
### Description

Runs htsjdk's unit tests on Java 11.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ cache:
 jdk:
   - oraclejdk8
   - openjdk8
+  - openjdk11
 matrix:
   fast_finish: true
   allow_failures: 
--- a/build.gradle
+++ b/build.gradle
@@ -38,7 +38,7 @@ dependencies {
     compile "org.tukaani:xz:1.5"
     compile "gov.nih.nlm.ncbi:ngs-java:2.9.0"
 
-    testCompile "org.scala-lang:scala-library:2.12.6"
+    testCompile "org.scala-lang:scala-library:2.12.8"
     testCompile "org.scalatest:scalatest_2.12:3.0.5"
     testRuntime 'org.pegdown:pegdown:1.6.0' // Necessary for generating HTML reports with ScalaTest
     testCompile "org.testng:testng:6.14.3"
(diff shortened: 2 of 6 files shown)


yfarjoun:
did you only open this to get the tests? is this a serious candidate for merging? in the current state, of course that would not be possible...

tomwhite:
@yfarjoun the goal here is to check we can build and run htsjdk tests with Java 11. The source and target Java versions are *not* changed - they are still Java 8 - so this is just to ensure that we can use a Java 11 compiler and runtime with the current htsjdk source. It does not replace the Java 8-based build.

There were a few things I had to change to get this to work:
* Upgrade to Gradle 5 for Java 11 support
* Change a Scala method call in the tests to avoid a conflict with Java (see  https://github.com/scala/bug/issues/11125).
* Remove two javadoc comments that were causing the Java 11 javadoc tool to fail.

This is a serious candidate for merging, since none of these changes interfere with the current Java 8 build. I also added a `openjdk11` target to Travis so that every change is checked on Java 11. This will mean that htsjdk doesn't regress in its ability to be built with and run on Java 11.

lbergelson:
Wow, those two javadoc comments were the issue?  I assumed it would be much worse than that.  Bizarre.  That shouldn't cause a javadoc null point exception...

lbergelson:
@tomwhite This still had some issues with it, I pushed lb_java11 which should resolve them.  Would you like to incorporate those changes into yours or should I create a new pr?

tomwhite:
@lbergelson thanks for looking at this. I had a look at your branch, but it wasn't clear how the changes related to Java 11. That said, please go ahead and merge as you see fit.

lbergelson:
@tomwhite They fix problems introduce by moving to gradle 5.  It broke the `maven` publication so I switched it to the newer `maven-publish` which seemed to fix the problem.   I bumped a few versions of things and did a little cleanup to remove newly deprecated stuff as well.

lbergelson:
closing this since I've taken it over in #1291

PR #1272, opened 2019-02-02
Moving the SRA tests to a separate env

yfarjoun:
### Description

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -108,13 +108,12 @@ test {
         jvmArgs += '-Dsamjdk.sra_libraries_download=true'
     }
 
-    excludeGroups 'sra'
-
     tags {
         exclude "slow"
         exclude "broken"
         exclude "defaultReference"
         exclude "ftp"
+        exclude 'sra'
         //if (System.env.CI == "false") exclude "sra"
         if (!OperatingSystem.current().isUnix()) exclude "unix"
     }
--- a/build.gradle
+++ b/build.gradle
@@ -104,16 +104,14 @@ test {
         events "failed", "skipped"
     }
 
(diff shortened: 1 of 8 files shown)


yfarjoun:
This should fix the breakage of the master branch.

src/test/java/htsjdk/samtools/cram/ref/EnaRefServiceTest.java
```diff
@@ -7,6 +7,8 @@
 
 import java.io.IOException;
 
+//calling this an FTP test so that it could fail in travis
+@Test(groups="ftp")
 public class EnaRefServiceTest extends HtsjdkTest {
```
lbergelson: I don't like labelling this as an FTP test.  Maybe instead of a new SRA block we can have an EXTERNAL_APIS test type that covers SRA and ENA.

build.gradle
```diff
@@ -128,11 +126,13 @@ task testFTP(type: Test) {
     }
 }
 
-task testSRA(type: Test) {
-    description = "Run the SRA tests"
+task testExternalApis(type: Test) {
+    description = "Run the SRA and ENA tests (external APIs"
```
lbergelson:
```suggestion
    description = "Run the SRA and ENA tests (tests that interact with external APIs)"
```

src/test/java/htsjdk/samtools/cram/ref/EnaRefServiceTest.java
yfarjoun: done.

PR #1274, opened 2019-02-11
CRAM: formalize Slice Type with an enum

jmthibault79:
### Description

Add a ReferenceContext member to Slice and Container, based on the discussion at https://github.com/samtools/hts-specs/issues/370 and the related spec update at https://github.com/samtools/hts-specs/pull/372.  

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
+++ b/src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
@@ -132,7 +132,7 @@ public class CRAMBAIIndexer {
                 /**
                  * Unmapped span must be processed after mapped spans:
                  */
-                AlignmentSpan unmappedSpan = spanMap.remove(ReferenceContext.UNMAPPED);
+                AlignmentSpan unmappedSpan = spanMap.remove(ReferenceContext.UNMAPPED_UNPLACED_CONTEXT);
                 for (final ReferenceContext refContext : new TreeSet<>(spanMap.keySet())) {
                     final AlignmentSpan span = spanMap.get(refContext);
                     final Slice fakeSlice = new Slice(refContext);
@@ -147,7 +147,7 @@ public class CRAMBAIIndexer {
                 }
 
                 if (unmappedSpan != null) {
-                    final Slice fakeSlice = new Slice(ReferenceContext.UNMAPPED);
+                    final Slice fakeSlice = new Slice(ReferenceContext.UNMAPPED_UNPLACED_CONTEXT);
                     fakeSlice.containerOffset = slice.containerOffset;
                     fakeSlice.offset = slice.offset;
                     fakeSlice.index = slice.index;
@@ -450,10 +450,10 @@ public class CRAMBAIIndexer {
                     final ReferenceContext containerContext = container.getReferenceContext();
                     String sequenceName;
                     switch (containerContext.getType()) {
-                        case UNMAPPED_UNPLACED:
+                        case UNMAPPED_UNPLACED_TYPE:
                             sequenceName = "?";
                             break;
-                        case MULTI_REFERENCE:
+                        case MULTIPLE_REFERENCE_TYPE:
                             sequenceName = "???";
                             break;
                         default:
--- a/src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
+++ b/src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
@@ -40,6 +40,7 @@ package htsjdk.samtools;
 
 import htsjdk.samtools.cram.build.ContainerParser;
 import htsjdk.samtools.cram.build.CramIO;
+import htsjdk.samtools.cram.ref.ReferenceContext;
 import htsjdk.samtools.cram.structure.AlignmentSpan;
 import htsjdk.samtools.cram.structure.Container;
 import htsjdk.samtools.cram.structure.ContainerIO;
@@ -49,7 +50,6 @@ import htsjdk.samtools.seekablestream.SeekableStream;
 import htsjdk.samtools.util.BlockCompressedFilePointerUtil;
 import htsjdk.samtools.util.Log;
 import htsjdk.samtools.util.ProgressLogger;
-import htsjdk.samtools.util.RuntimeIOException;
 
 import java.io.File;
 import java.io.IOException;
@@ -123,19 +123,19 @@ public class CRAMBAIIndexer {
         for (final Slice slice : container.slices) {
             slice.containerOffset = container.offset;
             slice.index = sliceIndex++;
-            if (slice.isMultiref()) {
+            if (slice.getReferenceContext().isMultiRef()) {
                 final ContainerParser parser = new ContainerParser(indexBuilder.bamHeader);
-                final Map<Integer, AlignmentSpan> refSet = parser.getReferences(container, validationStringency);
-                final Slice fakeSlice = new Slice();
+                final Map<ReferenceContext, AlignmentSpan> spanMap = parser.getReferences(container, validationStringency);
+
                 slice.containerOffset = container.offset;
                 slice.index = sliceIndex++;
                 /**
                  * Unmapped span must be processed after mapped spans:
                  */
-                AlignmentSpan unmappedSpan = refSet.remove(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX);
-                for (final int refId : new TreeSet<>(refSet.keySet())) {
-                    final AlignmentSpan span = refSet.get(refId);
-                    fakeSlice.sequenceId = refId;
+                AlignmentSpan unmappedSpan = spanMap.remove(ReferenceContext.UNMAPPED);
+                for (final ReferenceContext refContext : new TreeSet<>(spanMap.keySet())) {
+                    final AlignmentSpan span = spanMap.get(refContext);
+                    final Slice fakeSlice = new Slice(refContext);
                     fakeSlice.containerOffset = slice.containerOffset;
                     fakeSlice.offset = slice.offset;
                     fakeSlice.index = slice.index;
@@ -145,16 +145,16 @@ public class CRAMBAIIndexer {
                     fakeSlice.nofRecords = span.getCount();
                     processSingleReferenceSlice(fakeSlice);
                 }
+
                 if (unmappedSpan != null) {
-                    final AlignmentSpan span = unmappedSpan;
-                    fakeSlice.sequenceId = SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX;
+                    final Slice fakeSlice = new Slice(ReferenceContext.UNMAPPED);
                     fakeSlice.containerOffset = slice.containerOffset;
                     fakeSlice.offset = slice.offset;
                     fakeSlice.index = slice.index;
 
                     fakeSlice.alignmentStart = SAMRecord.NO_ALIGNMENT_START;
                     fakeSlice.alignmentSpan = 0;
-                    fakeSlice.nofRecords = span.getCount();
+                    fakeSlice.nofRecords = unmappedSpan.getCount();
                     processSingleReferenceSlice(fakeSlice);
                 }
             } else {
@@ -164,8 +164,8 @@ public class CRAMBAIIndexer {
     }
 
     /**
-     * Record index information for a given CRAM slice that contains either unmapped reads or
-     * reads mapped to a single reference.
+     * Record index information for a given CRAM slice that contains either
+     * unmapped-unplaced reads or reads mapped to a single reference.
      * If this alignment starts a new reference, write out the old reference.
      *
      * @param slice The CRAM slice, single ref or unmapped only.
@@ -173,13 +173,16 @@ public class CRAMBAIIndexer {
      */
     public void processSingleReferenceSlice(final Slice slice) {
         try {
-            final int reference = slice.sequenceId;
-            if (reference == SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX) {
+            final ReferenceContext sliceContext = slice.getReferenceContext();
+            if (sliceContext.isUnmappedUnplaced()) {
                 return;
             }
-            if (slice.sequenceId == Slice.MULTI_REFERENCE) {
+
+            if (sliceContext.isMultiRef()) {
                 throw new SAMException("Expecting a single reference slice.");
             }
+
+            final int reference = sliceContext.getSequenceId();
             if (reference != currentReference) {
                 // process any completed references
                 advanceToReference(reference);
@@ -264,13 +267,13 @@ public class CRAMBAIIndexer {
             // metadata
             indexStats.recordMetaData(slice);
 
-            final int alignmentStart = slice.alignmentStart;
-            if (alignmentStart == SAMRecord.NO_ALIGNMENT_START) {
+            final ReferenceContext sliceContext = slice.getReferenceContext();
+            if (! sliceContext.isMappedSingleRef()) {
                 return; // do nothing for records without coordinates, but count them
             }
 
             // various checks
-            final int reference = slice.sequenceId;
+            final int reference = sliceContext.getSequenceId();
             if (reference != currentReference) {
                 throw new SAMException("Unexpected reference " + reference +
                         " when constructing index for " + currentReference + " for record " + slice);
@@ -328,6 +331,7 @@ public class CRAMBAIIndexer {
             // process linear index
 
             // the smallest file offset that appears in the 16k window for this bin
+            final int alignmentStart = slice.alignmentStart;
             final int alignmentEnd = slice.alignmentStart + slice.alignmentSpan;
             int startWindow = LinearIndex.convertToLinearIndexOffset(alignmentStart); // the 16k window
             final int endWindow;
@@ -443,16 +447,17 @@ public class CRAMBAIIndexer {
                 indexer.processContainer(container, validationStringency);
 
                 if (null != log) {
+                    final ReferenceContext containerContext = container.getReferenceContext();
                     String sequenceName;
-                    switch (container.sequenceId) {
-                        case SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX:
+                    switch (containerContext.getType()) {
+                        case UNMAPPED_UNPLACED:
                             sequenceName = "?";
                             break;
-                        case Slice.MULTI_REFERENCE:
+                        case MULTI_REFERENCE:
                             sequenceName = "???";
                             break;
                         default:
-                            sequenceName = cramHeader.getSamFileHeader().getSequence(container.sequenceId).getSequenceName();
+                            sequenceName = cramHeader.getSamFileHeader().getSequence(containerContext.getSequenceId()).getSequenceName();
                             break;
                     }
                     progressLogger.record(sequenceName, container.alignmentStart);
--- a/src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
+++ b/src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
@@ -201,7 +201,7 @@ public class CRAMContainerStreamWriter {
             return true;
         }
 
-        if (refSeqIndex == ReferenceContext.REF_ID_MULTIPLE) {
+        if (refSeqIndex == ReferenceContext.MULTIPLE_REFERENCE_ID) {
             return false;
         }
 
@@ -216,7 +216,7 @@ public class CRAMContainerStreamWriter {
         if (samRecords.size() > MIN_SINGLE_REF_RECORDS) {
             return true;
         } else {
-            refSeqIndex = ReferenceContext.REF_ID_MULTIPLE;
+            refSeqIndex = ReferenceContext.MULTIPLE_REFERENCE_ID;
             return false;
         }
     }
@@ -265,13 +265,13 @@ public class CRAMContainerStreamWriter {
         final byte[] referenceBases;
         String refSeqName = null;
         switch (refSeqIndex) {
-            case ReferenceContext.REF_ID_MULTIPLE:
+            case ReferenceContext.MULTIPLE_REFERENCE_ID:
                 if (preservation != null && preservation.areReferenceTracksRequired()) {
                     throw new SAMException("Cannot apply reference-based lossy compression on non-coordinate sorted reads.");
                 }
                 referenceBases = new byte[0];
                 break;
-            case ReferenceContext.REF_ID_UNMAPPED:
+            case ReferenceContext.UNMAPPED_UNPLACED_ID:
                 referenceBases = new byte[0];
                 break;
             default:
@@ -480,14 +480,14 @@ public class CRAMContainerStreamWriter {
      * @param samRecordReferenceIndex index of the new reference sequence
      */
     private void updateReferenceContext(final int samRecordReferenceIndex) {
-        if (refSeqIndex == ReferenceContext.REF_ID_MULTIPLE) {
+        if (refSeqIndex == ReferenceContext.MULTIPLE_REFERENCE_ID) {
             return;
         }
 
         if (refSeqIndex == REF_SEQ_INDEX_NOT_INITIALIZED) {
             refSeqIndex = samRecordReferenceIndex;
         } else if (refSeqIndex != samRecordReferenceIndex) {
-            refSeqIndex = ReferenceContext.REF_ID_MULTIPLE;
+            refSeqIndex = ReferenceContext.MULTIPLE_REFERENCE_ID;
     }
     }
 
--- a/src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
+++ b/src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
@@ -10,6 +10,7 @@ import htsjdk.samtools.cram.common.Version;
 import htsjdk.samtools.cram.lossy.PreservationPolicy;
 import htsjdk.samtools.cram.lossy.QualityScorePreservation;
 import htsjdk.samtools.cram.ref.CRAMReferenceSource;
+import htsjdk.samtools.cram.ref.ReferenceContext;
 import htsjdk.samtools.cram.ref.ReferenceTracks;
 import htsjdk.samtools.cram.structure.Container;
 import htsjdk.samtools.cram.structure.ContainerIO;
@@ -55,8 +56,8 @@ public class CRAMContainerStreamWriter {
     private boolean preserveReadNames = true;
     private QualityScorePreservation preservation = null;
     private boolean captureAllTags = true;
-    private Set<String> captureTags = new TreeSet<String>();
-    private Set<String> ignoreTags = new TreeSet<String>();
+    private Set<String> captureTags = new TreeSet<>();
+    private Set<String> ignoreTags = new TreeSet<>();
 
     private CRAMBAIIndexer indexer;
     private long offset;
@@ -93,13 +94,7 @@ public class CRAMContainerStreamWriter {
      */
     public void writeAlignment(final SAMRecord alignment) {
         if (shouldFlushContainer(alignment)) {
-            try {
-                flushContainer();
-            } catch (IOException e) {
-                throw new RuntimeIOException(e);
-            } catch (IllegalAccessException e) {
-                throw new RuntimeException(e);
-            }
+            flushContainer();
         }
 
         updateReferenceContext(alignment.getReferenceIndex());
@@ -137,8 +132,6 @@ public class CRAMContainerStreamWriter {
             outputStream.close();
         } catch (final IOException e) {
             throw new RuntimeIOException(e);
-        } catch (final IllegalAccessException e) {
-            throw new RuntimeException(e);
         }
     }
 
@@ -208,7 +201,7 @@ public class CRAMContainerStreamWriter {
             return true;
         }
 
-        if (refSeqIndex == Slice.MULTI_REFERENCE) {
+        if (refSeqIndex == ReferenceContext.REF_ID_MULTIPLE) {
             return false;
         }
 
@@ -223,7 +216,7 @@ public class CRAMContainerStreamWriter {
         if (samRecords.size() > MIN_SINGLE_REF_RECORDS) {
             return true;
         } else {
-            refSeqIndex = Slice.MULTI_REFERENCE;
+            refSeqIndex = ReferenceContext.REF_ID_MULTIPLE;
             return false;
         }
     }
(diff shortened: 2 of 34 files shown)


src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
```diff
@@ -135,7 +134,7 @@ public void processContainer(final Container container, final ValidationStringen
                 AlignmentSpan unmappedSpan = refSet.remove(SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX);
                 for (final int refId : new TreeSet<>(refSet.keySet())) {
                     final AlignmentSpan span = refSet.get(refId);
-                    fakeSlice.sequenceId = refId;
+                    fakeSlice.setSequenceId(refId);
```
jmthibault79: also sets the SliceType, to ensure they stay in sync

src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
```diff
@@ -93,13 +94,7 @@ public CRAMContainerStreamWriter(
...
     public void writeAlignment(final SAMRecord alignment) {
         if (shouldFlushContainer(alignment)) {
-            try {
-                flushContainer();
-            } catch (IOException e) {
-                throw new RuntimeIOException(e);
-            } catch (IllegalAccessException e) {
-                throw new RuntimeException(e);
-            }
+            flushContainer();
```
jmthibault79: no longer throws anything we need to catch

src/main/java/htsjdk/samtools/CRAMIterator.java
```diff
@@ -157,40 +157,45 @@ void nextContainer() throws IllegalArgumentException, CRAMException {
...
+            case UNMAPPED:
+                referenceBases = new byte[]{};
+                prevSeqId = SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX;
+                break;
+            case MULTI_REFERENCE:
+                referenceBases = null;
+                prevSeqId = Slice.MULTI_REFERENCE;
+                break;
+            default:
+                if (prevSeqId != container.getSequenceId()) {
```
jmthibault79: removed the `prevSeqId < 0` check: since we know that the slice type is single-ref, we also know that the current sequence ID here is 0+.  So it will differ from any negative values.

src/main/java/htsjdk/samtools/cram/structure/Slice.java
```diff
@@ -54,8 +55,10 @@
     public static final int NO_ALIGNMENT_SPAN = 0;
     private static final Log log = Log.getInstance(Slice.class);
 
+    private SliceType sliceType;
+
     // as defined in the specs:
-    public int sequenceId = -1;
+    private int sequenceId = -1;
```
jmthibault79: can't modify this without also modifying type

src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
```diff
@@ -15,6 +15,7 @@
 import htsjdk.samtools.cram.structure.ContainerIO;
 import htsjdk.samtools.cram.structure.CramCompressionRecord;
 import htsjdk.samtools.cram.structure.Slice;
+import htsjdk.samtools.cram.structure.slice.SliceType;
```
cmnbroad: Someday the `slice` package eventually contain other stuff ? Otherwise it seems strange to have `SliceType` but not `Slice` in there...

src/main/java/htsjdk/samtools/cram/structure/Container.java
```diff
@@ -59,6 +63,27 @@
      */
     public long offset;
 
+
+    /**
+     * Get this container's serializable sequence ID:
+     * 0 or greater for a single reference
+     * -1 for unmapped slice
+     * -2 for multiple reference slice
```
cmnbroad: Can this be narrowed so it only returns valid (actual) sequence IDs ? I think having it return an int that is overloaded with the slice type sentinel values diminishes the value of having a type-safe `getSliceType` method in the first place. We want to encourage callers to use `getSliceType` *instead* of the sequence ID to determine what type of Slice it is, and to determine whether or not it has a valid sequence ID (or valid alignment span), and only call `getSequenceID` when they want/expect an actual legitimate sequenceID value. I would go so far as to have this throw if the `sliceType` isn't `SINGLE_REFERENCE`. We might have to change some other call sites that depend on the current overloading of these values.

src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
jmthibault79: yes - I hope to use that for the overall Slice refactor.  I can put elsewhere (structure?) if you prefer that I wait until then.

src/main/java/htsjdk/samtools/cram/structure/Container.java
jmthibault79: We do need a method to return the literal value for serialization, as required by the spec.

How about something like this: `public getValidSequenceId()` which checks for validity, and a separate `getSequenceIdForHeader()` which can return -1 and -2?  The later can eventually be `private` when we refactor `Container` to handle its own I/O.

src/main/java/htsjdk/samtools/cram/structure/slice/SliceType.java
```diff
@@ -0,0 +1,31 @@
...
+ *      -1: Unmapped Slice
+ *      Any positive integer (including zero): Single Reference Slice
+ */
+public enum SliceType {
+    MULTI_REFERENCE,
+    UNMAPPED,
+    SINGLE_REFERENCE;
+
+    public static SliceType bySequenceId(final int sequenceId) {
+        if (sequenceId == -2) {
```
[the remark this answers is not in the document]
jmthibault79: doh.  of course.

src/main/java/htsjdk/samtools/cram/structure/slice/SliceType.java
```diff
@@ -0,0 +1,31 @@
...
+            return MULTI_REFERENCE;
+        } else if (sequenceId == -1) {
```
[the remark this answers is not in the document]
jmthibault79: I like that.  added such a constant.

src/main/java/htsjdk/samtools/cram/structure/Container.java
```diff
@@ -43,6 +44,9 @@
     public int[] landmarks;
     public int checksum = 0;
 
+    // all slices in this container must be of the same type
+    private SliceType sliceType;
```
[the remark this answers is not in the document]
jmthibault79: I had `containerType` in a previous version.  Now I have switched it back to that.

I have some lifecycle thoughts.  Let's chat on Slack.

src/main/java/htsjdk/samtools/cram/structure/ContainerHeaderIO.java
```diff
@@ -53,7 +53,7 @@ public boolean readContainerHeader(final int major, final Container container, f
         }
 
         container.containerByteSize = CramInt.readInt32(peek);
-        container.sequenceId = ITF8.readUnsignedITF8(inputStream);
+        container.setSequenceId(ITF8.readUnsignedITF8(inputStream));
         container.alignmentStart = ITF8.readUnsignedITF8(inputStream);
         container.alignmentSpan = ITF8.readUnsignedITF8(inputStream);
```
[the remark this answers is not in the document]
jmthibault79: My assumption is that we have MULTI and UNMAPPED Slices/Containers in the wild which do have alignment starts and spans, so I'm concerned at breakage if we enforce this.

src/main/java/htsjdk/samtools/cram/structure/Container.java
jmthibault79: Added `getValidSequenceId()` and `getSequenceIdOrSentinel()` to both Slice and Container.

src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
```diff
@@ -328,6 +328,7 @@ private void processSingleReferenceSlice(final Slice slice) {
             // process linear index
 
             // the smallest file offset that appears in the 16k window for this bin
+            final int alignmentStart = slice.alignmentStart;
```
cmnbroad: Yay!

src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
cmnbroad: If this is the right place its fine - just wanted to check.

src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
```diff
@@ -267,24 +260,24 @@ private static void updateTracks(final List<SAMRecord> samRecords, final Referen
...
      */
-    protected void flushContainer() throws IllegalArgumentException, IllegalAccessException, IOException {
+    protected void flushContainer() throws IllegalArgumentException {
 
-        final byte[] refs;
+        final byte[] referenceBases;
         String refSeqName = null;
-        switch (refSeqIndex) {
-            case Slice.MULTI_REFERENCE:
+        switch (SliceType.bySequenceId(refSeqIndex)) {
```
cmnbroad: The `refSeqIndex` variable in this class takes on all of the values of SliceType, plus legitimate seq indices, plus the sentinel value `REF_SEQ_INDEX_NOT_INITIALIZED`. I wonder if we can unify those somehow. If not now, then certainly in a future PR..

src/main/java/htsjdk/samtools/cram/structure/Container.java
```diff
@@ -59,10 +63,57 @@
...
+    /**
+     * Set this container's sequence ID and derive its {@link SliceType} from that value
+     * @param sequenceId the sequence ID
+     */
+    public Container(final int sequenceId) {
+        this.sequenceId = sequenceId;
+        this.containerType = SliceType.bySequenceId(sequenceId);
+    }
+
+    public static Container newMultiRefContainer() {
```
cmnbroad: As long as we continue to have ` ContainerFactory` class I wonder if these should live there.

src/main/java/htsjdk/samtools/cram/structure/Container.java
```diff
@@ -59,10 +63,57 @@
...
+    public int getSequenceIdOrSentinel() {
+        return sequenceId;
+    }
+
+    /**
+     * Get this single-reference container's valid sequence ID
+     * @throw CRAMException if this container is not single-ref
+     * @return the sequence ID
+     */
+    public int getValidSequenceId() {
```
cmnbroad: How about `getReferenceSequenceID` ?

src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
cmnbroad: The more I see how this code is evolving, the more I think we need a `ReferenceContext` class that wraps what is currently called `SliceType`, plus a legitimate reference ID, that can be used by `Container`, `Slice` and all of these other classes that need to track that same set of components.

jmthibault79: My `SliceTyped` idea at 05d46996ced1e85e28d0dc61e04713a3ac92b77d is similar, except inheritance instead of composition.

src/main/java/htsjdk/samtools/cram/structure/Container.java
jmthibault79: Hm, interesting.  My inclination is to remove ContainerFactory in a future Container iteration.

src/main/java/htsjdk/samtools/cram/structure/Container.java
jmthibault79: Open to changing the name but I would like some signifier that this also validates

src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
jmthibault79: I've added ReferenceContext here, but I'm not thrilled about all the duplicated pass-through methods.  What would you think about a `HasReferenceContext` interface or ABC to interact with a ReferenceContext?

src/main/java/htsjdk/samtools/cram/structure/Container.java
cmnbroad: Ok - I was thinking that "valid" should be implied, and that if a qualifier is needed somewhere, it should be on the method that *isn't* valid (i.e., what is currently called `getSequenceIdOrSentinel`, which might return a validID or it might return a sentinel).

Thinking out loud here though, I would argue we should seek to not have `getSequenceIdOrSentinel` at all now that we have `SliceType` and `ReferenceContext`. Part of the goal is to eliminate the need for code (at least outside of `Slice` and `Container`) from needing to overload integer sequence ID variables with sentinel values. Ideally the sentinels could be kept private.

Maybe `SliceType` should be called `ReferenceContextType`; and `ReferenceContext` should have a constructor that takes a `ReferenceContextType` and one that takes an integer ID. Then evrything would be type-safe, and his method could be called `getReferenceSequence`, which would only be valid if the `ReferenceContextType` for a `ReferenceContext` is SINGLE_REFERENCE.

src/main/java/htsjdk/samtools/cram/structure/Container.java
cmnbroad: Agreed - I just thought in the meantime it was suboptimal to not have them all together, but its not that important if its temporary.

cmnbroad: See my comment below - I think it would eliminate most of the pass-through method if we just make the one concrete implementation of `ReferenceContext` be the currency used everywhere, but if not let me know and we can revisit this.

src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
cmnbroad: See my comment below - I think it would eliminate most of the pass-through methods if we just make the one concrete implementation of `ReferenceContext` be the currency used everywhere, but if not let me know and we can revisit this.

src/main/java/htsjdk/samtools/CRAMIterator.java
```diff
@@ -153,40 +154,47 @@ void nextContainer() throws IllegalArgumentException, CRAMException {
...
-            refs = referenceSource.getReferenceBases(sequence, true);
-            if (refs == null) {
-                throw new CRAMException(String.format("Contig %s not found in the reference file.", sequence.getSequenceName()));
-            }
-            prevSeqId = container.sequenceId;
+        final ReferenceContext containerContext = container.getReferenceContext();
+        switch (containerContext.getType()) {
+            case UNMAPPED:
+                referenceBases = new byte[]{};
+                prevSeqId = SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX;
```
jmthibault79: future candidate for turning into a ReferenceContext

src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
```diff
@@ -208,7 +201,7 @@ protected boolean shouldFlushContainer(final SAMRecord nextRecord) {
             return true;
         }
 
-        if (refSeqIndex == Slice.MULTI_REFERENCE) {
+        if (refSeqIndex == Slice.REF_ID_MULTIPLE) {
```
jmthibault79: refSeqIndex is a future candidate for ReferenceContext

src/main/java/htsjdk/samtools/cram/build/ContainerParser.java
```diff
@@ -59,60 +60,55 @@ public ContainerParser(final SAMFileHeader samFileHeader) {
         return records;
     }
 
-    public Map<Integer, AlignmentSpan> getReferences(final Container container, final ValidationStringency validationStringency) {
```
jmthibault79: someday we will merge a PR that collapses all these stupid things.  I may even have such a thing open?

src/main/java/htsjdk/samtools/cram/encoding/reader/CramRecordReader.java
```diff
@@ -173,10 +175,10 @@ public void read(final CramCompressionRecord cramRecord) {
 
             cramRecord.flags = bitFlagsCodec.readData();
             cramRecord.compressionFlags = compressionBitFlagsCodec.readData();
-            if (refId == Slice.MULTI_REFERENCE) {
+            if (refContext.isMultiRef()) {
                 cramRecord.sequenceId = refIdCodec.readData();
             } else {
-                cramRecord.sequenceId = refId;
+                cramRecord.sequenceId = refContext.getSequenceIdOrSentinel();
```
jmthibault79: watch out: can be unmapped here

src/main/java/htsjdk/samtools/cram/build/CramIO.java
```diff
@@ -257,17 +258,14 @@ private static long writeContainerForSamFileHeader(final int major, final SAMFil
         System.arraycopy(data, 0, blockContent, 0, Math.min(data.length, length));
         final Block block = Block.createRawFileHeaderBlock(blockContent);
 
-        final Container container = new Container();
+        final Container container = new Container(new ReferenceContext(0));
         container.blockCount = 1;
         container.blocks = new Block[]{block};
         container.landmarks = new int[0];
         container.slices = new Slice[0];
-        container.alignmentSpan = Slice.NO_ALIGNMENT_SPAN;
```
jmthibault79: these fields are now initialized to these values

src/main/java/htsjdk/samtools/cram/structure/ContainerHeaderIO.java
```diff
@@ -17,29 +17,32 @@
...
     }
 
-    public boolean readContainerHeader(final int major, final Container container, final InputStream inputStream) {
+    public Container readContainerHeader(final int major, final InputStream inputStream) {
         final byte[] peek = new byte[4];
         try {
             int character = inputStream.read();
             if (character == -1)
-                return false;
+                return readContainerHeader(new ByteArrayInputStream((major >= 3 ? CramIO.ZERO_F_EOF_MARKER : CramIO.ZERO_B_EOF_MARKER)));
```
jmthibault79: This was what the caller did when it received a `false` from this method.

src/main/java/htsjdk/samtools/cram/structure/ContainerIO.java
```diff
@@ -56,13 +56,7 @@ private static Container readContainer(final int major, final InputStream inputS
      * @return a new {@link Container} object with container header values filled out but empty body (no slices and blocks).
      */
     public static Container readContainerHeader(final int major, final InputStream inputStream) {
-        final Container container = new Container();
```
jmthibault79: logic moved inside `ContainerHeaderIO.readContainerHeader()`

src/main/java/htsjdk/samtools/cram/build/ContainerParser.java
jmthibault79: Yes: that's over at #1276

src/main/java/htsjdk/samtools/cram/structure/ContainerHeaderIO.java
cmnbroad: Mildly not crazy about the recursion, but this change would make it directly recursive instead of indirectly, allow the otherwise unused method above to be deleted:

```suggestion
                return readContainerHeader(major, new ByteArrayInputStream((major >= 3 ? CramIO.ZERO_F_EOF_MARKER : CramIO.ZERO_B_EOF_MARKER)));
```

src/main/java/htsjdk/samtools/cram/ref/ReferenceContext.java
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * Are we handling MULTI_REFERENCE records (-2, from the CRAM spec)?
+ *
+ * Are we handling UNMAPPED_UNPLACED records (-1, from the CRAM spec)?
+ *
+ * Or are we handing a known SINGLE_REFERENCE sequence (0 or higher, from the CRAM spec))?
+ *
+ */
+public class ReferenceContext implements Comparable<ReferenceContext> {
+    // TODO: merge with Slice constants
```
cmnbroad: Is there any reason we can't do this now (see my comments below) ? It seems like the ultimate authority for these values should be `ReferenceContext` ?

src/main/java/htsjdk/samtools/cram/ref/ReferenceContext.java
```diff
@@ -0,0 +1,141 @@
...
+     * -1 for unmapped-unplaced
+     * -2 for multiple reference
+     *
+     * @param serializableSequenceId the sequence ID or sentinel value for constructing this ReferenceContext
+     */
+    public ReferenceContext(final int serializableSequenceId) {
+        this.serializableSequenceId = serializableSequenceId;
+
+        switch (serializableSequenceId) {
+            case INTERNAL_INT_MULTIPLE:
```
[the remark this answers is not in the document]
cmnbroad: Since you already have the public static `ReferenceContext` values `MULTIPLE` and `UNMAPPED` anyway, why not reserve this public constructor for only constructing `SINGLE_REFERENCE` contexts, where a legitimate (int >= 0) sequence ID value is required. Then have a separate, private constructor that takes a `ReferenceContextType` for creating the public statics, which can then be treated/referenced as true singletons everywhere they're needed.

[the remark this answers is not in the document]
jmthibault79: Agree with the first comment - I'll merge the constants.

Disagree with the second - this check is necessary.  Outside of tests, this constructor is only used by callers which don't know whether it's single-ref or not.  The singletons are used in cases where the caller is aware.

src/main/java/htsjdk/samtools/cram/ref/ReferenceContext.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /**
+     * Get the ReferenceContext sequence ID or sentinel value, suitable for serialization:
+     * 0 or greater for single reference
+     * -1 for unmapped
+     * -2 for multiple reference
+     * @return the sequence ID
+     */
+    public int getSequenceIdOrSentinel() {
```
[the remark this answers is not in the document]
jmthibault79: Renamed and added toString

src/main/java/htsjdk/samtools/cram/ref/ReferenceContext.java
jmthibault79: Nope.  Done now.  I'm guessing something in a previous iteration made this easier.

src/main/java/htsjdk/samtools/cram/build/ContainerFactory.java
```diff
@@ -53,25 +53,24 @@ Container buildContainer(final List<CramCompressionRecord> records,
...
             slices.add(slice);
         }
 
-        container.finalizeContainerState(slices.toArray(new Slice[0]));
+        final Container container = Container.initializeFromSlices(slices);
+        container.header = header;
+        container.nofRecords = records.size();
+        container.globalRecordCounter = globalRecordCounter;
+        container.blockCount = 0;
+        container.bases += baseCount;
```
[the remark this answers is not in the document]
jmthibault79: No, I'm not seeing much.  I'll add some.

src/main/java/htsjdk/samtools/cram/ref/ReferenceContext.java
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * Or are we handing a known SINGLE_REFERENCE sequence (0 or higher, from the CRAM spec))?
+ *
+ */
+public class ReferenceContext implements Comparable<ReferenceContext> {
+    public static final int REF_ID_MULTIPLE = -2;
+    public static final int REF_ID_UNMAPPED = SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX; // -1
+
+    public static final ReferenceContext MULTIPLE = new ReferenceContext(REF_ID_MULTIPLE);
+    public static final ReferenceContext UNMAPPED = new ReferenceContext(REF_ID_UNMAPPED);
```
cmnbroad: Suggest aligning the names for the integer constants, these static objects, and the enum names as much as possible. For example rather than `REF_ID_MULTIPLE`, `MULTIPLE` , and `MULTI_REFERENCE`, which use differing abbreviations and ordering of ref/multi, maybe use `REFERENCE_MULTIPLE_ID`, `REFERENCE_MULTIPLE`, and `REFERENCE_MULTIPLE`. Same then the same for corresponding `SINGLE` and `UNMAPPED` ones, etc. Or whatever makes sense.

src/test/java/htsjdk/samtools/cram/ref/ReferenceContextTest.java
```diff
@@ -0,0 +1,76 @@
...
+        Assert.assertEquals(multi.getSerializableId(), ReferenceContext.REF_ID_MULTIPLE);
+    }
+
+    @Test(expectedExceptions = CRAMException.class)
+    public void testInvalidSetSequenceId() {
+        new ReferenceContext(-3);
+    }
+
+    @DataProvider(name = "sentinels")
+    private static Object[][] invalidGet() {
```
cmnbroad: We also usually try to align the name of the data provider with the function. I believe they can be the same. (I always use a name, but you also can just refer to the data provider by the method name, and not use a name property at all).

src/test/java/htsjdk/samtools/cram/ref/ReferenceContextTest.java
```diff
@@ -0,0 +1,76 @@
...
+        final ReferenceContext multi = new ReferenceContext(ReferenceContext.REF_ID_MULTIPLE);
+        final ReferenceContext multiStatic = ReferenceContext.MULTIPLE;
+        Assert.assertEquals(multiStatic, multi);
+
+        Assert.assertFalse(multi.isMappedSingleRef());
+        Assert.assertFalse(multi.isUnmappedUnplaced());
+        Assert.assertTrue(multi.isMultiRef());
+        Assert.assertEquals(multi.getType(), ReferenceContextType.MULTI_REFERENCE);
+        Assert.assertEquals(multi.getSerializableId(), ReferenceContext.REF_ID_MULTIPLE);
+    }
```
cmnbroad: I know there are a lot of existing tests written in this style, but in general we try not to gang up tests in a single method like this. These could be organized with a data provider (not sure I captured all the cases), but something like this:
```suggestion
    }    @DataProvider(name = "validRefContexts")
    private static Object[][] refContexts() {
        return new Object[][] {
                //sequenceID, expectedType, isSingle, isUnplaced, isMultiple, serializableID
                {0, ReferenceContextType.SINGLE_REFERENCE, true, false, false, 0},
                {1, ReferenceContextType.SINGLE_REFERENCE, true, false, false, 1},
                {ReferenceContext.REF_ID_MULTIPLE, ReferenceContextType.MULTI_REFERENCE, false, true, false, ReferenceContext.REF_ID_MULTIPLE},
                {ReferenceContext.REF_ID_UNMAPPED, ReferenceContextType.UNMAPPED_UNPLACED, false, false, true, ReferenceContext.REF_ID_UNMAPPED}
        };
    }

    @Test(dataProvider = "validRefContexts")
    public void referenceContextTest(
            int serializableID,
            ReferenceContextType expectedType,
            boolean expectedSingle,
            boolean expectedMulti,
            boolean expectedUnplaced,
            int expectedID) {
        final ReferenceContext refContext = new ReferenceContext(serializableID);
        Assert.assertEquals(refContext.isMappedSingleRef(), expectedSingle);
        Assert.assertEquals(refContext.isUnmappedUnplaced(), expectedUnplaced);
        Assert.assertEquals(refContext.isMultiRef(), expectedMulti);
        Assert.assertEquals(refContext.getType(), expectedType);
        Assert.assertEquals(refContext.getSerializableId(), expectedID);
    }
```

src/main/java/htsjdk/samtools/cram/build/ContainerFactory.java
```diff
@@ -39,41 +39,33 @@ public ContainerFactory(final SAMFileHeader samFileHeader, final int recordsPerS
...
-        return buildContainer(records, null);
-    }
-
-    Container buildContainer(final List<CramCompressionRecord> records,
-                             final SubstitutionMatrix substitutionMatrix) {
-
         // sets header APDelta
         final boolean coordinateSorted = samFileHeader.getSortOrder() == SAMFileHeader.SortOrder.coordinate;
+        final SubstitutionMatrix substitutionMatrix = null;     // TODO
         final CompressionHeader header = new CompressionHeaderFactory().build(records, substitutionMatrix, coordinateSorted);
```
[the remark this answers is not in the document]
jmthibault79: The "TODO" is that I don't know what this does, so I'm just naming the parameter.

I'll just pass null here instead.

src/main/java/htsjdk/samtools/cram/structure/Container.java
```diff
@@ -80,40 +94,44 @@
...
+                endPlusOne = Math.max(endPlusOne, slice.alignmentStart + slice.alignmentSpan);
             }
-        }
 
-        // because we enforce identical ref seq IDs, this checks that all slices are mapped
-        if (firstSlice.isMappedSingleRef()) {
-            alignmentStart = start;
-            alignmentSpan = end - start;
+            container.alignmentStart = start;
+            container.alignmentSpan = endPlusOne - start;
```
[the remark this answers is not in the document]
jmthibault79: Not in pathological cases.  We do need max(end) rather than just lastStart and lastSpan.  Here's an example: 1: 1-10 followed by 1: 2-5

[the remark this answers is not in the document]
cmnbroad: Oh, right.

src/test/java/htsjdk/samtools/cram/ref/ReferenceContextTest.java
```diff
@@ -0,0 +1,76 @@
...
+    @Test(expectedExceptions = CRAMException.class)
+    public void testInvalidSetSequenceId() {
+        new ReferenceContext(-3);
+    }
+
+    @DataProvider(name = "sentinels")
+    private static Object[][] invalidGet() {
+        return new Object[][] {
+                {-2},
+                {-1}
```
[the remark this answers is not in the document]
jmthibault79: yes

src/test/java/htsjdk/samtools/cram/ref/ReferenceContextTest.java
jmthibault79: oh cool - I was not aware of that

src/test/java/htsjdk/samtools/cram/build/ContainerFactoryTest.java
```diff
@@ -266,14 +283,14 @@ public void multiRefWithUnmappedNoRef() {
...
+        assertContainerState(container, ReferenceContext.MULTIPLE, Slice.NO_ALIGNMENT_START, Slice.NO_ALIGNMENT_SPAN);
     }
 
     @Test
     public void multiRefWithUnmappedNoStart() {
-        final ContainerFactory factory = new ContainerFactory(getSAMFileHeaderForTests(), 10);
+        final ContainerFactory factory = new ContainerFactory(getSAMFileHeaderForTests(), TEST_RECORD_COUNT);
         final List<CramCompressionRecord> records = new ArrayList<>();
-        for (int i = 0; i < 10; i++) {
+        for (int i = 0; i < TEST_RECORD_COUNT; i++) {
```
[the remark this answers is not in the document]
jmthibault79: Refactored these in the latest commit

src/test/java/htsjdk/samtools/cram/structure/ContainerTest.java
```diff
@@ -1,149 +1,104 @@
...
         slice.alignmentSpan = 15;
 
-        final Container container = new Container();
-        container.finalizeContainerState(slice);
+        final Container container = Container.initializeFromSlices(Arrays.asList(slice));
 
-        Assert.assertEquals(container.sequenceId, slice.sequenceId);
+        Assert.assertTrue(container.getReferenceContext().isMappedSingleRef());
+        Assert.assertEquals(container.getReferenceContext().getType(), ReferenceContextType.SINGLE_REFERENCE_TYPE);
+        Assert.assertEquals(container.getReferenceContext(), slice.getReferenceContext());
```
cmnbroad: Many of the blocks of asserts in these tests could be replaced with one of the `ContainerFactoryTest.assertContainerState` methods. Not sure if `SliceTests.assertSliceState` could be reuseable as well.

Issue #1220, opened 2018-11-02
CramCompressionRecord defines equals but not hashCode


lbergelson:
This could cause all sorts of strange behavior.  It should be implemented.

jmthibault79:
This problem is all over CRAM actually.  I'll fix it.

cmnbroad:
This as fixed in #1222.

Issue #1267, opened 2019-01-28
Address issues found by SpotBugs


tomwhite:
I ran SpotBugs on the htsjdk source and it found a number of issues that should be looked at.

To gnerate the files, checkout https://github.com/tomwhite/htsjdk/commit/8d2f7255ca062fded660676c5fa6c6635c80da08, and type 

```
./gradlew check -x test
```

Here are the reports that it generated (converted to PDF since GitHub doesn't support HTML attachments; however this means you can't drill down into each issue):
[main.pdf](https://github.com/samtools/htsjdk/files/2802757/main.pdf)
[test.pdf](https://github.com/samtools/htsjdk/files/2802758/test.pdf)

cmnbroad:
Thanks Tom. @jmthibault79 - Might be worth taking a look at the CRAM package part of this as you're refactoring.

tomwhite:
With the latest PRs all the spotbugs issues in the tests have been resolved, and there are now 19 remaining issues in the main code. I've opened issues for the CRAM ones, but there are a few more (with less obvious fixes) that need looking at - they can either be fixed or an exclusion can be added to gradle/spotbugs-exclude.xml. @lbergelson it would be good to get your opinion on these. (The easiest way to get the latest report is to checkout https://github.com/tomwhite/htsjdk/tree/spotbugs and run `./gradlew spotbugsMain`.)

When all the issues have been fixed then we can merge https://github.com/tomwhite/htsjdk/tree/spotbugs to run spotbugs on each new PR to ensure no new issues are introduced.

tomwhite:
See more fixes in #1303

lbergelson:
This was completed and spotbugs now runs automatically on pull requests.

PR #1222, opened 2018-11-07
Add hashCode() to classes with equals() and clean up equals()

jmthibault79:
### Description

`equals()` without `hashCode()` is bad news.  Also, some `equals()` have problems too.

### Checklist

- [x] Code compiles correctly
- [x] New tests covering changes and new functionality
- [x] All tests passing
- [x] Extended the README / documentation, if necessary
- [ ] Is not backward compatible (breaks binary or source compatibility)

PR #1278, opened 2019-02-13
A few fixes for issues found by spotbugs

tomwhite:
Address more of #1267 

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/cram/encoding/readfeatures/Substitution.java
+++ b/src/main/java/htsjdk/samtools/cram/encoding/readfeatures/Substitution.java
@@ -122,7 +122,7 @@ public class Substitution implements Serializable, ReadFeature {
         if (code == NO_CODE) {
             return Objects.hash(position);
         }
-        return Objects.hash(position, base, referenceBase, code);
+        return Objects.hash(position, base, referenceBase);
     }
 
     @Override
--- a/src/main/java/htsjdk/samtools/cram/encoding/readfeatures/Substitution.java
+++ b/src/main/java/htsjdk/samtools/cram/encoding/readfeatures/Substitution.java
@@ -18,6 +18,7 @@
 package htsjdk.samtools.cram.encoding.readfeatures;
 
 import java.io.Serializable;
+import java.util.Objects;
 
 /**
  * A substitution event captured in read coordinates. It is characterized by position in read, read base and reference base.
@@ -114,6 +115,11 @@ public class Substitution implements Serializable, ReadFeature {
             if (base != substitution.base) return false;
         }
         return true;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(position, base, referenceBase, code);
     }
 
     @Override
--- a/src/main/java/htsjdk/samtools/cram/encoding/readfeatures/Substitution.java
+++ b/src/main/java/htsjdk/samtools/cram/encoding/readfeatures/Substitution.java
@@ -119,6 +119,9 @@ public class Substitution implements Serializable, ReadFeature {
 
     @Override
     public int hashCode() {
+        if (code == NO_CODE) {
+            return Objects.hash(position);
+        }
         return Objects.hash(position, base, referenceBase, code);
     }
 
--- a/src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
+++ b/src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
@@ -28,6 +28,7 @@ import htsjdk.samtools.cram.encoding.readfeatures.SoftClip;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
+import java.util.Objects;
 
 public class CramCompressionRecord {
     private static final int MULTI_FRAGMENT_FLAG = 0x1;
@@ -131,6 +132,14 @@ public class CramCompressionRecord {
 
     private boolean deepEquals(final Collection<?> c1, final Collection<?> c2) {
         return (c1 == null || c1.isEmpty()) && (c2 == null || c2.isEmpty()) || c1 != null && c1.equals(c2);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = Objects.hash(alignmentStart, readLength, recordsToNextFragment, mappingQuality, readFeatures, flags, readName);
+        result = 31 * result + Arrays.hashCode(readBases);
+        result = 31 * result + Arrays.hashCode(qualityScores);
+        return result;
     }
 
     @Override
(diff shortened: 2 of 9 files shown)


src/main/java/htsjdk/samtools/cram/encoding/readfeatures/Substitution.java
```diff
@@ -116,6 +117,11 @@ public boolean equals(final Object obj) {
         return true;
     }
 
+    @Override
+    public int hashCode() {
+        return Objects.hash(position, base, referenceBase, code);
```
jmthibault79: From an previous attempt to fix this: https://github.com/samtools/htsjdk/pull/1222#discussion_r231674367

jmthibault79: Specifically: what is that weird `NO_CODE` stuff in `equals()` and do we need to match its behavior here?

src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
```diff
@@ -133,6 +134,14 @@ private boolean deepEquals(final Collection<?> c1, final Collection<?> c2) {
         return (c1 == null || c1.isEmpty()) && (c2 == null || c2.isEmpty()) || c1 != null && c1.equals(c2);
     }
 
+    @Override
```
jmthibault79: This is another class with a strange `equals()` method which should be carefully checked.

src/main/java/htsjdk/samtools/cram/encoding/readfeatures/Substitution.java
tomwhite: From the [contract](https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#hashCode--) for `hashCode`:

> If two objects are equal according to the `equals(Object)` method, then calling the `hashCode` method on each of the two objects must produce the same integer result.

If two `Substitution` objects are equal then they must have equal `position`, `base`, `referenceBase`, and `code` fields (since if any of them differ then `equals` returns false), and the implementation of `hashCode` will produce the same result for both objects.

Furthermore

> It is _not_ required that if two objects are unequal according to the `equals(java.lang.Object)` method, then calling the `hashCode` method on each of the two objects must produce distinct integer results.

In this case, if the two objects have `NO_CODE` codes then `equals` will return false, but `hashCode` may return `true` - which is consistent with the contract.

So, my reading is that this fix is OK.

src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
tomwhite: I think this is OK too. The `equals` method will return true only if all of the fields listed here in the `hashCode` method are equal. I think the second part of the contract is actually satisfied in this case too.

tomwhite:
Thanks for taking a look @jmthibault79

src/main/java/htsjdk/samtools/cram/encoding/readfeatures/Substitution.java
jmthibault79: I think I've found a violation: when `a.code == b.code == NO_CODE` and `a.base != b.base`, then `equals()` returns true but `hashCode()`s would differ.

src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
jmthibault79: Here's a violation: `a.readFeatures == null` and `b.readFeatures.isEmpty()`: the `deepEquals()` method means `equals()` is true, but `Objects.hash()` assigns 0 to null and 1 to empty list.

src/main/java/htsjdk/samtools/cram/encoding/readfeatures/Substitution.java
tomwhite: @jmthibault79 you're right - and I am spectacularly wrong! I've fixed and added a comprehensive test.

src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
tomwhite: Ditto. Fixed and added a test which covers a lot of cases.

src/test/java/htsjdk/samtools/cram/encoding/ReadFeaturesTest.java
```diff
@@ -25,4 +29,29 @@ public void faultyEquality() {
...
         Assert.assertEquals(sc1, sc2);
     }
+
+    @Test
+    public void testSubstitutionEqualsAndHashCodeAreConsistent() {
+        final List<Substitution> substitutions = new ArrayList<>();
+        for (int position : new int[] {0, 1}) {
+            for (byte base : new byte[] {(byte) -1, (byte) 'A'}) {
+                for (byte referenceBase : new byte[] {(byte) -1, (byte) 'C'}) {
+                    for (byte code : new byte[] {Substitution.NO_CODE, (byte) 2}) {
```
jmthibault79: Add a second valid code here (like 1).  This makes the test fail.

src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
jmthibault79: Looks good.  Would you mind adding a real `readFeatures` list and an empty-string `readName` to the test as well?

src/test/java/htsjdk/samtools/cram/encoding/ReadFeaturesTest.java
tomwhite: This is because two objects with different codes (neither of which is `NO_CODE`) will always compare equal. I think the fix is not to use `code` in the computation of `hashCode`.

src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
tomwhite: Done

src/test/java/htsjdk/samtools/cram/encoding/ReadFeaturesTest.java
jmthibault79: I think that works.  If the only difference between two of these is that one code is `NO_CODE` but the other isn't, equals() says false but hashCode() is the same.  That's OK in the contract.

tomwhite: Right. Technically `hashCode()` could always return the same number and be compliant (that might have saved a lot of this bother!), but then it's not very good for hash tables.

jmthibault79:
Resolves #1220

lbergelson:
Thank you @tomwhite and @jmthibault79

PR #1282, opened 2019-02-13
remove caching of alleles in AbstractVCFCodec

lbergelson:
This replaces https://github.com/samtools/htsjdk/pull/1264 and responds to my own comments.

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/Allele.java
+++ b/src/main/java/htsjdk/variant/variantcontext/Allele.java
@@ -118,6 +118,11 @@ public class Allele implements Comparable<Allele>, Serializable {
     public static final long serialVersionUID = 1L;
 
     private static final byte[] EMPTY_ALLELE_BASES = new byte[0];
+    private static final char SINGLE_BREAKEND_INDICATOR = '.';
+    private static final char BREAKEND_EXTENDING_RIGHT = '[';
+    private static final char BREAKEND_EXTENDING_LEFT = ']';
+    private static final char SYMBOLIC_ALLELE_START = '<';
+    private static final char SYMBOLIC_ALLELE_END = '>';
 
     private boolean isRef = false;
     private boolean isNoCall = false;
@@ -300,7 +305,7 @@ public class Allele implements Comparable<Allele>, Serializable {
     	if ( bases.length <= 1 )
             return false;
         else {
-            return bases[0] == '<' || bases[bases.length - 1] == '>' ||
+            return bases[0] == SYMBOLIC_ALLELE_START || bases[bases.length - 1] == SYMBOLIC_ALLELE_END ||
                     wouldBeBreakpoint(bases) ||
                     wouldBeSingleBreakend(bases);
         }
@@ -313,15 +318,14 @@ public class Allele implements Comparable<Allele>, Serializable {
     public static boolean wouldBeBreakpoint(final byte[] bases) {
         if (bases.length <= 1) {
             return false;
-        else {
-            for (int i = 0; i < bases.length; i++) {
-                final byte base = bases[i];
-                if (base == '[' || base == ']') {
-                    return true;
-                }
-            }
-            return false;
         }
+        for (int i = 0; i < bases.length; i++) {
+            final byte base = bases[i];
+            if (base == BREAKEND_EXTENDING_LEFT || base == BREAKEND_EXTENDING_RIGHT) {
+                return true;
+            }
+        }
+        return false;
     }
     /**
      * @param bases  bases representing an allele
@@ -331,7 +335,7 @@ public class Allele implements Comparable<Allele>, Serializable {
         if ( bases.length <= 1 )
             return false;
         else {
-            return bases[0] == '.' || bases[bases.length - 1] == '.';
+            return bases[0] == SINGLE_BREAKEND_INDICATOR || bases[bases.length - 1] == SINGLE_BREAKEND_INDICATOR;
         }
     }
 
--- a/src/main/java/htsjdk/variant/variantcontext/Allele.java
+++ b/src/main/java/htsjdk/variant/variantcontext/Allele.java
@@ -311,7 +311,7 @@ public class Allele implements Comparable<Allele>, Serializable {
      * @return true if the bases represent a symbolic allele in breakpoint notation, (ex: G]17:198982] or ]13:123456]T )
      */
     public static boolean wouldBeBreakpoint(final byte[] bases) {
-        if ( bases.length <= 1 )
+        if (bases.length <= 1) {
             return false;
         else {
             for (int i = 0; i < bases.length; i++) {
--- a/src/main/java/htsjdk/variant/variantcontext/Allele.java
+++ b/src/main/java/htsjdk/variant/variantcontext/Allele.java
@@ -300,10 +300,36 @@ public class Allele implements Comparable<Allele>, Serializable {
     	if ( bases.length <= 1 )
             return false;
         else {
-            final String strBases = new String(bases);
-            return (bases[0] == '<' || bases[bases.length-1] == '>') || // symbolic or large insertion
-            		(bases[0] == '.' || bases[bases.length-1] == '.') || // single breakend
-                    (strBases.contains("[") || strBases.contains("]")); // mated breakend
+            return bases[0] == '<' || bases[bases.length - 1] == '>' ||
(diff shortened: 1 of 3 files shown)


lbergelson:
@pshapiro4broad Could you review this branch?

src/main/java/htsjdk/variant/variantcontext/Allele.java
```diff
@@ -294,16 +294,44 @@ public static boolean wouldBeNoCallAllele(final byte[] bases) {
...
+                    wouldBeSingleBreakend(bases);
+        }
+    }
+
+    /**
+     * @param bases  bases representing an allele
+     * @return true if the bases represent a symbolic allele in breakpoint notation, (ex: G]17:198982] or ]13:123456]T )
+     */
+    public static boolean wouldBeBreakpoint(final byte[] bases) {
+        if ( bases.length <= 1 )
```
pshapiro4broad: Code is inconsistent in spacing/braces. To make it consistent I would add `{` `}` here and remove the extra spaces, on this line and the other lines that duplicate this, e.g. 300 and 331.
```suggestion
        if (bases.length <= 1) {
```

src/main/java/htsjdk/variant/variantcontext/Allele.java
```diff
@@ -294,16 +294,44 @@ public static boolean wouldBeNoCallAllele(final byte[] bases) {
...
+            return false;
+        else {
```
pshapiro4broad: unnecessary `else` after `return`

src/main/java/htsjdk/variant/variantcontext/Allele.java
```diff
@@ -294,16 +294,44 @@ public static boolean wouldBeNoCallAllele(final byte[] bases) {
...
+            for (int i = 0; i < bases.length; i++) {
```
pshapiro4broad: can use java foreach here
```suggestion
            for (byte base : bases) {
```

src/main/java/htsjdk/variant/variantcontext/Allele.java
```diff
@@ -294,16 +294,44 @@ public static boolean wouldBeNoCallAllele(final byte[] bases) {
...
+    }
+    /**
+     * @param bases  bases representing an allele
+     * @return true if the bases represent a symbolic allele in single breakend notation (ex: .A or A. )
+     */
+    public static boolean wouldBeSingleBreakend(final byte[] bases) {
+        if ( bases.length <= 1 )
+            return false;
+        else {
+            return bases[0] == '.' || bases[bases.length - 1] == '.';
```
pshapiro4broad: It would be nice to use constants instead of the magic characters `>` `<` `[` `]` `.`

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -332,12 +332,12 @@ else if ( parts[2].equals(VCFConstants.EMPTY_ID_FIELD) )
...
 
-        final String ref = getCachedString(parts[3].toUpperCase());
-        final String alts = getCachedString(parts[4]);
+        final String ref = parts[3].toUpperCase();
+        final String alts = parts[4];
         builder.log10PError(parseQual(parts[5]));
 
         final List<String> filters = parseFilters(getCachedString(parts[6]));
-        if ( filters != null ) builder.filters(new HashSet<String>(filters));
+        if ( filters != null ) builder.filters(new HashSet<>(filters));
```
pshapiro4broad: may as well format the `if` across lines and use `{` `}` if you're changing this line

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -577,7 +577,7 @@ private static void checkAllele(String allele, boolean isRef, int lineNo) {
             System.err.println(String.format("Allele detected with length %d exceeding max size %d at approximately line %d, likely resulting in degraded VCF processing performance", allele.length(), MAX_ALLELE_SIZE_BEFORE_WARNING, lineNo));
         }
 
-        if ( isSymbolicAllele(allele) ) {
+        if (Allele.wouldBeSymbolicAllele(allele.getBytes())) {
```
pshapiro4broad: This check is done before calling the `Allele` constructor, but it appears to be ensuring that we create a valid Allele. Is there a reason why this code isn't part of the `Allele` constructor? We would have to wrap the constructor call in a try/catch to get a message printed out with the VCF line number, though.

src/main/java/htsjdk/variant/variantcontext/Allele.java
```diff
@@ -294,16 +294,44 @@ public static boolean wouldBeNoCallAllele(final byte[] bases) {
...
+                    wouldBeBreakpoint(bases) ||
+                    wouldBeSingleBreakend(bases);
+        }
+    }
+
+    /**
+     * @param bases  bases representing an allele
+     * @return true if the bases represent a symbolic allele in breakpoint notation, (ex: G]17:198982] or ]13:123456]T )
+     */
+    public static boolean wouldBeBreakpoint(final byte[] bases) {
```
pshapiro4broad: The use of `would` here confuses me. If this is a breakpoint allele string, then why not `is`?

src/test/java/htsjdk/variant/variantcontext/AlleleUnitTest.java
```diff
@@ -277,4 +277,38 @@ public void testExtend() {
         Assert.assertEquals("ATCGA", Allele.extend(Allele.create("AT"), "CGA".getBytes()).toString());
         Assert.assertEquals("ATCGA", Allele.extend(Allele.create("ATC"), "GA".getBytes()).toString());
     }
+    @Test
+    public void testWouldBeSymbolic() {
```
pshapiro4broad: Could rewrite these tests using a testng data provider
```suggestion
    public void testWouldBeSymbolic(String allele, boolean expectedSymbolic) {
```

src/main/java/htsjdk/variant/variantcontext/Allele.java
lbergelson: Yeah, it's weird, but it matches the existing methods. I think changing 1 of them is worse than having this weird convention.  We could change all of them though...

src/main/java/htsjdk/variant/variantcontext/Allele.java
lbergelson: We've been finding noticeable runtime costs for using foreach in hotspots (seemingly because it instantiates a new iterator each time which is much more costly than a fast for loop).  Since Alleles are created in a lot of places I figured I'd stick with the for loop even though it's gross... That said, I haven't profiled this particular instance in any sensible way so I can't prove that this is any faster.

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
lbergelson: Yeah, there's a big redundant section of allele checking code here that I believe is redundant with the constructor.  I wanted to make this as small a change set as possible though.  I started a follow up branch that cleans up some of this stuff but I figured I'd get this one in before that.

src/main/java/htsjdk/variant/variantcontext/Allele.java
lbergelson: done

src/main/java/htsjdk/variant/variantcontext/Allele.java
lbergelson: I've extracted some... I can't decide if it's an improvement or not... You can take a look when I push the changes.

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
lbergelson: done

src/test/java/htsjdk/variant/variantcontext/AlleleUnitTest.java
lbergelson: Done, I didn't do all the tests in this file though, just the bottom ones..

src/main/java/htsjdk/variant/variantcontext/Allele.java
pshapiro4broad: OK, I didn't see the rest of them.

PR #1283, opened 2019-02-14
Fixes for issues found by SpotBugs

### LINKED ISSUE #1279: [SpotBugs] contentId isn't initialized in htsjdk.samtools.cram.structure.block.ExternalBlock.getContentId() when invoked from constructor for superclass
This looks like a real bug in CRAM. /cc @jmthibault79, @cmnbroad 

See https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#ur-uninitialized-read-of-field-method-called-from-constructor-of-superclass-ur-uninit-read-called-from-super-constructor

Part of #1267 

Report:
```
In class htsjdk.samtools.cram.structure.block.ExternalBlock
In method htsjdk.samtools.cram.structure.block.ExternalBlock.getContentId()
Field htsjdk.samtools.cram.structure.block.ExternalBlock.contentId
Superclass constructor is new htsjdk.samtools.cram.structure.block.Block(BlockCompressionMethod, BlockContentType, byte[], int)
Call from superclass constructor at Block.java:[line 86]
Constructor is new htsjdk.samtools.cram.structure.block.ExternalBlock(BlockCompressionMethod, int, byte[], int)
Field set (too late) at ExternalBlock.java:[line 26]
At ExternalBlock.java:[line 31]
```

### LINKED ISSUE #1280: [SpotBugs] Field only ever set to null: htsjdk.samtools.CRAMIterator.nextRecord
This looks like a real bug in CRAM. /cc @jmthibault79, @cmnbroad

See https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#uwf-field-only-ever-set-to-null-uwf-null-field

Part of #1267

Report:
```
In class htsjdk.samtools.CRAMIterator
Field htsjdk.samtools.CRAMIterator.nextRecord
In CRAMIterator.java
```

### LINKED ISSUE #1281: [SpotBugs] htsjdk.samtools.cram.structure.CramHeader defines clone() but doesn't implement Cloneable
Cloneable is not used very much so maybe deprecate and remove the clone() method? /cc @jmthibault79, @cmnbroad

See https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#cn-class-defines-clone-but-doesn-t-implement-cloneable-cn-implements-clone-but-not-cloneable

Part of #1267

Report:
```
In class htsjdk.samtools.cram.structure.CramHeader
In method htsjdk.samtools.cram.structure.CramHeader.clone()
At CramHeader.java:[lines 80-85]
```

jmthibault79:
### Description

Resolves #1279 - removed the ExternalBlock class which was not truly necessary
Resolves #1280 - removed the unused `CramIterator.nextRecord` field
Resolves #1281 - removed the unused `CramHeader.clone()` method

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


src/main/java/htsjdk/samtools/CRAMIterator.java
```diff
@@ -41,7 +41,6 @@
     private final CountingInputStream countingInputStream;
     private final CramHeader cramHeader;
     private final ArrayList<SAMRecord> records;
-    private SAMRecord nextRecord = null;
```
jmthibault79: This would resolve #1280 - but what was this *intended* to do?

src/main/java/htsjdk/samtools/CRAMIterator.java
```diff
@@ -209,10 +205,9 @@ void nextContainer() throws IllegalArgumentException, CRAMException {
             if (mReader != null) {
                 final long chunkStart = (container.offset << 16) | cramRecord.sliceIndex;
                 final long chunkEnd = ((container.offset << 16) | cramRecord.sliceIndex) + 1;
-                nextRecord.setFileSource(new SAMFileSource(mReader,
-                        new BAMFileSpan(new Chunk(chunkStart, chunkEnd))));
+                samRecord.setFileSource(new SAMFileSource(mReader, new BAMFileSpan(new Chunk(chunkStart, chunkEnd))));
```
jmthibault79: This appears to be what was intended here.

src/main/java/htsjdk/samtools/CRAMIterator.java
cmnbroad: Probably cut&pasted from some other iterator, if I had to guess.

src/main/java/htsjdk/samtools/CRAMIterator.java
cmnbroad: Apparently this code block has never been executed before, since it would have gotten an NPE. Can you add a test that exercises this (calls `enableFileSource` on a CRAMReader, and verifies that `SAMRecord.getFileSource` returns a `SAMFileSource` in that case) ?

src/main/java/htsjdk/samtools/cram/structure/block/Block.java
```diff
@@ -66,15 +71,18 @@
...
      * @param uncompressedLength the length of the content stored in this block when uncompressed
      */
     protected Block(final BlockCompressionMethod compressionMethod,
                     final BlockContentType contentType,
+                    final int contentId,
                     final byte[] compressedContent,
                     final int uncompressedLength) {
         this.compressionMethod = compressionMethod;
         this.contentType = contentType;
+        this.contentId = contentId;
```
cmnbroad: We should add a unit test in `BlockTest` with a data provider for all of the block types, and a negative test that verifies that this method won't create a block with a content ID unless its an external block.

src/main/java/htsjdk/samtools/CRAMIterator.java
jmthibault79: Writing this test was a bit tricky because I found that this path is only triggered when `cramFile` is null - i.e. the `CRAMIterator` was constructed using an `InputStream` instead of a file.  Maybe it's the rarity of that situation which hid this bug for so long.

src/test/java/htsjdk/samtools/CRAMFileReaderTest.java
```diff
@@ -240,4 +240,35 @@ public void testCramIteratorWithoutCallingHasNextFirst() throws IOException {
...
+            // which may get replaced.  is this correct behavior?
+
+            cramReader.enableFileSource(samReader, true);
+
+            final SAMRecord record = cramReader.getIterator().next();
+            Assert.assertEquals(record.getFileSource().getReader(), samReader);
+
+            cramReader.close();
+        }
+    }
```
cmnbroad: Yeah, it looks like you had to really jump through hoops to get that code block to execute. But I see the problem now - the SamReader is only propagated by `CRAMFileReader` to the `CRAMIterator` thats in place when enableFileSource is called, which is almost always immediately replaced by a new iterator instance as soon as an iterator is called for.

Given that its so hard to even get this code block to execute, and how unnatural the test code is (requiring opening the same file twice), I'd say we should just remove this test; I created https://github.com/samtools/htsjdk/issues/1293 to deal with the bigger issue.

src/test/java/htsjdk/samtools/cram/structure/BlockTest.java
```diff
@@ -102,12 +103,54 @@ public void testExternalBlockRoundTrips() throws IOException {
...
+
+    private class BlockTester extends Block {
+        private BlockTester(final BlockCompressionMethod compressionMethod,
+                           final BlockContentType contentType,
+                           final int contentId,
+                           final byte[] compressedContent,
+                           final int uncompressedLength) {
+            super(compressionMethod, contentType, contentId, compressedContent, uncompressedLength);
+        }
+    }
```
cmnbroad: If you moved this file into the block package I think you wouldn't need this ?

src/test/java/htsjdk/samtools/cram/structure/BlockTest.java
```diff
@@ -102,12 +103,54 @@ public void testExternalBlockRoundTrips() throws IOException {
...
+    public void nonExternalContentId(final BlockContentType contentType, final int ignored) {
+        new BlockTester(BlockCompressionMethod.RAW, contentType, Block.NO_CONTENT_ID, new byte[0], 0);
+    }
+
+    // show that non-external content types cannot have valid contentIds
+
+    @Test(dataProvider = "nonExternalTypes", expectedExceptions = CRAMException.class)
+    public void nonExternalContentIdNegative(final BlockContentType contentType, final int blockContentId) {
+        new BlockTester(BlockCompressionMethod.RAW, contentType, blockContentId, new byte[0], 0);
+    }
```
cmnbroad:
```suggestion
    @DataProvider(name = "nonExternalTypes")
    private Object[][] nonExternalTypes() {
        return new Object[][] {
                {BlockContentType.COMPRESSION_HEADER},
                {BlockContentType.CORE},
                {BlockContentType.MAPPED_SLICE},
                {BlockContentType.FILE_HEADER},
                {BlockContentType.RESERVED},
        };
    }

    // show that we can build all non-external Block types with NO_CONTENT_ID
    // but we can't if they have other values

    @Test(dataProvider = "nonExternalTypes")
    public void nonExternalContentId(final BlockContentType contentType) {
        new BlockTester(BlockCompressionMethod.RAW, contentType, Block.NO_CONTENT_ID, new byte[0], 0);
    }

    // show that non-external content types cannot have valid contentIds

    @Test(dataProvider = "nonExternalTypes", expectedExceptions = CRAMException.class)
    public void nonExternalContentIdNegative(final BlockContentType contentType) {
        final int VALID_CONTENT_ID = 1;
        new BlockTester(BlockCompressionMethod.RAW, contentType, VALID_CONTENT_ID, new byte[0], 0);
    }
```

src/test/java/htsjdk/samtools/cram/structure/BlockTest.java
jmthibault79: Right!

PR #1284, opened 2019-02-14
update CramCompressionRecord.isPlaced() to make it APDelta-aware

jmthibault79:
### Description

isPlaced() was not checking for the APDelta flag, thus inappropriately marking some records as unplaced.

Error introduced on 1/30 with #1266

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
+++ b/src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
@@ -379,7 +379,7 @@ public class CRAMContainerStreamWriter {
                     while (last.next != null) last = last.next;
 
                     if (cramRecord.isFirstSegment() && last.isLastSegment()) {
-                        final boolean coordinateSorted = (samFileHeader.getSortOrder() == SAMFileHeader.SortOrder.unsorted);
+                        final boolean coordinateSorted = (samFileHeader.getSortOrder() == SAMFileHeader.SortOrder.coordinate);
                         final int templateLength = CramNormalizer.computeInsertSize(cramRecord, last, coordinateSorted);
 
                         if (cramRecord.templateSize == templateLength) {
--- a/src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
+++ b/src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
@@ -379,8 +379,8 @@ public class CRAMContainerStreamWriter {
                     while (last.next != null) last = last.next;
 
                     if (cramRecord.isFirstSegment() && last.isLastSegment()) {
-
-                        final int templateLength = CramNormalizer.computeInsertSize(cramRecord, last);
+                        final boolean coordinateSorted = (samFileHeader.getSortOrder() == SAMFileHeader.SortOrder.unsorted);
+                        final int templateLength = CramNormalizer.computeInsertSize(cramRecord, last, coordinateSorted);
 
                         if (cramRecord.templateSize == templateLength) {
                             last = cramRecord.next;
--- a/src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
+++ b/src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
@@ -272,7 +272,7 @@ public class CramCompressionRecord {
     // check placement without regard to mapping; helper method for isPlaced()
     private boolean isPlacedInternal(final boolean useAbsolutePositionEncoding) {
         // placement requires a valid sequence ID
-        if (sequenceId != SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX) {
+        if (sequenceId == SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX) {
             return false;
         }
 
--- a/src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
+++ b/src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
@@ -184,8 +184,8 @@ public class CramCompressionRecord {
 
     private void intializeAlignmentBoundaries(final boolean usePositionDeltaEncoding) {
         if (!isPlaced(usePositionDeltaEncoding)) {
-            alignmentSpan = 0;
-            alignmentEnd = SAMRecord.NO_ALIGNMENT_START;
+            alignmentSpan = Slice.NO_ALIGNMENT_SPAN;
+            alignmentEnd = Slice.NO_ALIGNMENT_END;
             return;
         }
 
--- a/src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
+++ b/src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
@@ -183,7 +183,7 @@ public class CramCompressionRecord {
     }
 
     private void intializeAlignmentBoundaries(final boolean usePositionDeltaEncoding) {
-        if (! isPlaced(usePositionDeltaEncoding)) {
+        if (!isPlaced(usePositionDeltaEncoding)) {
             alignmentSpan = 0;
             alignmentEnd = SAMRecord.NO_ALIGNMENT_START;
             return;
@@ -211,7 +211,6 @@ public class CramCompressionRecord {
                         break;
                 }
             }
-
         }
 
         alignmentEnd = alignmentStart + alignmentSpan - 1;
--- a/src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
+++ b/src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
@@ -24,6 +24,7 @@ import htsjdk.samtools.cram.encoding.readfeatures.InsertBase;
 import htsjdk.samtools.cram.encoding.readfeatures.Insertion;
 import htsjdk.samtools.cram.encoding.readfeatures.ReadFeature;
 import htsjdk.samtools.cram.encoding.readfeatures.SoftClip;
+import htsjdk.samtools.util.Log;
 
 import java.util.Arrays;
 import java.util.Collection;
@@ -51,6 +52,8 @@ public class CramCompressionRecord {
 
     private static final int UNINITIALIZED_END = -1;
     private static final int UNINITIALIZED_SPAN = -1;
+
+    private static final Log log = Log.getInstance(CramCompressionRecord.class);
 
     // sequential index of the record in a stream:
     public int index = 0;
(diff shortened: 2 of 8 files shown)


src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
```diff
@@ -152,13 +152,22 @@ public String toString() {
...
+        if (alignmentSpan < 0) {
+            calculateAlignmentBoundaries(APDelta);
+        }
         return alignmentSpan;
     }
 
-    void calculateAlignmentBoundaries() {
-        if (isSegmentUnmapped()) {
+    public int getAlignmentEnd(final boolean APDelta) {
+        if (alignmentEnd < 0) {
```
cmnbroad: What are the possible interpretations of `alignmentEnd < 0` ? Is it possible to replace tests like this with something stricter that checks for a specific value ? Could we always initialize `alignmentEnd` to a specific value indicating that not yet known due to APDelta weirdness ? Either way this could use some beefed up comments describing what its doing.

src/test/java/htsjdk/samtools/cram/structure/CramCompressionRecordTest.java
```diff
@@ -12,19 +13,24 @@
  * Created by vadim on 28/09/2015.
  */
 public class CramCompressionRecordTest extends HtsjdkTest {
-    @Test
-    public void test_getAlignmentEnd() {
+    @DataProvider(name = "tf")
```
cmnbroad: Can we give this a name that indicates what its intended to toggle (i.e. APDelta).

src/test/java/htsjdk/samtools/cram/structure/CramCompressionRecordTest.java
```diff
@@ -12,19 +13,24 @@
...
+    private Object[][] tf() {
+        return new Object[][] { {true}, {false}};
+    }
+
+    @Test(dataProvider = "tf")
+    public void test_getAlignmentEnd(final boolean APDelta) {
```
cmnbroad: Except when referring to the actual APDelta data series, it might be more useful (in the tests and in the code) to call these arguments `usePositionDeltaEncoding`, or something like that, in order to underscore what its actually referring to.

src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
jmthibault79: This is a check to see if these values have been initialized.  I'll make that more explicit.

More generally, this is a part of the horrible mess of `CramCompressionRecord` alignment boundary calculation which I hope to get to during the CramCompressionRecord refactoring.

src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
```diff
@@ -379,8 +379,8 @@ protected void flushContainer() throws IllegalArgumentException, IllegalAccessEx
                     while (last.next != null) last = last.next;
 
                     if (cramRecord.isFirstSegment() && last.isLastSegment()) {
-
-                        final int templateLength = CramNormalizer.computeInsertSize(cramRecord, last);
+                        final boolean coordinateSorted = (samFileHeader.getSortOrder() == SAMFileHeader.SortOrder.unsorted);
```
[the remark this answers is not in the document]
jmthibault79: oops, good catch.

Unfortunately, I believe that it would take me a substantial amount of time to make sense of CramNormalizer and write a reasonable test for it, so I don't think that's feasible with the quick turnaround time we want for this PR.

src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
```diff
@@ -220,15 +224,20 @@ public void setSegmentUnmapped(final boolean segmentUnmapped) {
...
      *
      * @see #isSegmentUnmapped()
      * @return true if the record is placed
      */
-    public boolean isPlaced() {
-        return sequenceId != SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX &&
-                alignmentStart != SAMRecord.NO_ALIGNMENT_START;
+    public boolean isPlaced(final boolean APDelta) {
+        // if an absolute alignment start coordinate is required but we have none, it's unplaced
+        if (! APDelta && alignmentStart == SAMRecord.NO_ALIGNMENT_START) {
```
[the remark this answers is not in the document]
jmthibault79: updated logic and comments based on our offline conversation

src/test/java/htsjdk/samtools/cram/structure/SliceTests.java
```diff
@@ -72,238 +75,180 @@ public void testFailsMD5Check() throws IOException {
...
+                          final int expectedAlignmentStart,
+                          final int expectedAlignmentSpan) {
+        final CompressionHeader header = new CompressionHeaderFactory().build(records, null, coordinateSorted);
         final Slice slice = Slice.buildSlice(records, header);
-
-        Assert.assertEquals(slice.nofRecords, 10);
-        Assert.assertEquals(slice.sequenceId, Slice.MULTI_REFERENCE);
-        Assert.assertEquals(slice.alignmentStart, Slice.NO_ALIGNMENT_START);
-        Assert.assertEquals(slice.alignmentSpan, Slice.NO_ALIGNMENT_SPAN);
+        assertSliceState(slice, expectedRecordCount, expectedSequenceId, expectedAlignmentStart, expectedAlignmentSpan);
```
cmnbroad: I like the way this test/provider combination is written - simple and declarative - thx.

src/test/java/htsjdk/samtools/cram/structure/SliceTests.java
```diff
@@ -72,238 +75,180 @@ public void testFailsMD5Check() throws IOException {
...
 
-        Assert.assertEquals(slice.nofRecords, 2);
-        Assert.assertEquals(slice.sequenceId, Slice.MULTI_REFERENCE);
-        Assert.assertEquals(slice.alignmentStart, Slice.NO_ALIGNMENT_START);
-        Assert.assertEquals(slice.alignmentSpan, Slice.NO_ALIGNMENT_SPAN);
+    private List<CramCompressionRecord> getNoStartRecords() {
+        final List<CramCompressionRecord> records = new ArrayList<>();
+        for (int i = 0; i < TEST_RECORD_COUNT; i++) {
+            final CramCompressionRecord rec = createRecord(i, i);
+            rec.alignmentStart = SAMRecord.NO_ALIGNMENT_START;
```
cmnbroad: Same comment here. Some of these records are mapped/unplaced (though for a different reason than the one above, since these have a seq index but no start).

src/test/java/htsjdk/samtools/cram/structure/SliceTests.java
```diff
@@ -72,238 +75,180 @@ public void testFailsMD5Check() throws IOException {
...
+            records.add(createRecord(i, i));
+        }
+        return records;
+    }
 
-        final Slice slice = Slice.buildSlice(records, header);
+    private List<CramCompressionRecord> getNoRefRecords() {
+        final List<CramCompressionRecord> records = new ArrayList<>();
+        for (int i = 0; i < TEST_RECORD_COUNT; i++) {
+            records.add(createRecord(i, SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX));
```
cmnbroad: This test data is a bit suspicious. `createRecord` alternately creates mapped and unmapped records, which means some of the records created here will wind up as "mapped/half-unplaced" and others will wind up as "unmapped/half-placed", due to having an alignment start but no reference index. Those are weird states for test data, and its not at all clear from looking at this method whether its intentional. Not that we don't want to test the "aberrant" cases, but we need a cleaner separation between valid test data and invalid.

src/test/java/htsjdk/samtools/cram/structure/SliceTests.java
```diff
@@ -72,238 +75,180 @@ public void testFailsMD5Check() throws IOException {
...
-        record1.qualityScores = "!!!".getBytes();
-        record1.readLength = 3;
-        record1.readName = "1";
-        record1.sequenceId = 1;
-        record1.alignmentStart = 1;
-        record1.setLastSegment(true);
-        record1.readFeatures = Collections.emptyList();
-        record1.setSegmentUnmapped(false);
-
+        final CramCompressionRecord record1 = createRecord(0, 0);
```
cmnbroad: It seems like this test has a lot of duplicated code, and is doing something very similar it what `getSingleRefRecords`/`getMultiRefRecords` is doing. It is possible to rewrite it test a provide and those primitives ?

src/test/java/htsjdk/samtools/cram/structure/CramCompressionRecordTest.java
```diff
@@ -12,19 +13,24 @@
...
  */
 public class CramCompressionRecordTest extends HtsjdkTest {
-    @Test
-    public void test_getAlignmentEnd() {
+    @DataProvider(name = "deltaEncodingTrueFalse")
+    private Object[][] tf() {
+        return new Object[][] { {true}, {false}};
+    }
+
+    @Test(dataProvider = "deltaEncodingTrueFalse")
```
cmnbroad: Can we add tests for the various cases of `getAlignmentSpan` (and UNINITIALIZED_SPAN if appropriate) ?

src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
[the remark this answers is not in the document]
cmnbroad: Is there any test in this PR that would have failed before without the fix for APDelta, but passes now so at least we can be sure that we've fixed it ? I assume we didn't have one before this bug was introduced, since no test caught the issue ? Without that, I'm not even sure I'd recognize the failure if I saw it. Ideally we'd create a ticket describing it (that will  be fixed by this PR) so we have a record in case anyone sees it using the interim version that had the issue.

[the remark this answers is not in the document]
jmthibault79: Yes.  The pair of SliceTests cases starting with the comment on line 97 calls this out explicitly.

On a lower level, the combinations in CramCompressionRecordTest.placedTests show that the APDelta flag does affect results, and some would fail without this fix.

src/test/java/htsjdk/samtools/cram/structure/SliceTests.java
```diff
@@ -72,238 +75,194 @@ public void testFailsMD5Check() throws IOException {
...
-        Assert.assertEquals(slice.sequenceId, Slice.MULTI_REFERENCE);
-        Assert.assertEquals(slice.alignmentStart, Slice.NO_ALIGNMENT_START);
-        Assert.assertEquals(slice.alignmentSpan, Slice.NO_ALIGNMENT_SPAN);
+    private List<CramCompressionRecord> getNoStartRecords() {
+        final List<CramCompressionRecord> records = new ArrayList<>();
+        for (int index = 0; index < TEST_RECORD_COUNT; index++) {
+            final CramCompressionRecord record = createRecord(index, index);
+            record.alignmentStart = SAMRecord.NO_ALIGNMENT_START;
+            record.setSegmentUnmapped(true);
+            records.add(record);
```
cmnbroad: The concern I was expressing in the last round is that these methods create a mix of records, some of which are in "undefined" territory (which I referred to as "half-x"). This method creates some records (half of them, but thats not what I meant by "half") that have the unmapped flag set; but have a valid reference index; and not a valid start. Although we want to be able to handle those, they're outside the normal range of records we expect to see. I'd like to find a way to make that explicit (or preferably, separate such data from the normal cases), or at least make it clear that its intentional. Let me know what you think about this issue or if it doesnt make sense.

jmthibault79: Oh I see now.  Yes, it is intentional - I want to cover all of these half-cases.  Would adding comments like this work for you?

> These records have valid startAlignments but don't have valid references.  This is not something we expect to see in reads, but we want to treat them as unplaced if we do.  These tests cover that rare case.

cmnbroad: That will work for the short term, sure.

src/test/java/htsjdk/samtools/cram/structure/CramCompressionRecordTest.java
```diff
@@ -4,85 +4,189 @@
...
+        Assert.assertEquals(r.getAlignmentSpan(usePositionDeltaEncoding), expectedSpan);
+        Assert.assertEquals(r.getAlignmentEnd(usePositionDeltaEncoding), expectedEnd);
     }
 
-    @Test
-    public void test_isPlaced() {
+    // show that alignmentEnd and alignmentSpan are set once only and do not update
+    // to reflect the state of the record
+
+    // TODO: is this the behavior we actually want?
```
cmnbroad: Ohh - I didn't see this before. Not updating the span when either start or length is updated seems dangerous and totally wrong to me. Not sure what you think, but because they're publicly accessible, we have to assume they're mutable, so I think getAlignmentEnd/Span needs to always calculate on demand.

This is a great example of why I feel like we need to bite the bullet and refactor the underlying classes (`CramCompressionRecord` in this case) in order to make progress and write reasonable tests with confidence, rather than making incremental changes and point fixes. Its really hard to know what to do in cases like this.

cmnbroad: In looking back at previous versions, it looks like this behavior always existed and isn't new, so we can defer a fix to another PR.

PR #1285, opened 2019-02-15
Changes to FastaReferenceWriter

lbergelson:
* make sure the OutputStreams are always buffered
* change close behavior to be more likely to close all output streams in
the event of an early shutdown
* loosen restriction from BufferedWriter to Writer in
SAMSequenceDictionaryCodec and SAMTextHeaderCodec

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMSequenceDictionaryCodec.java
+++ b/src/main/java/htsjdk/samtools/SAMSequenceDictionaryCodec.java
@@ -25,7 +25,7 @@
 package htsjdk.samtools;
 
 import htsjdk.samtools.util.LineReader;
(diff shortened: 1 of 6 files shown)


src/main/java/htsjdk/samtools/SAMSequenceDictionaryCodec.java
```diff
@@ -26,6 +26,7 @@
 
 import htsjdk.samtools.util.LineReader;
 import java.io.BufferedWriter;
```
yfarjoun: check imports

lbergelson: good catch

PR #1288, opened 2019-02-16
Add an IntervalCodec that use useful for sorting large sets of

nh13:
intervals.

### Description

This will be useful for using with a `SortingCollection` when we are sorting over MANY intervals.  My intention is to submit a PR to Picard to use this codec and `SortingCollection` when lifting over an interval list, as I sometimes run out of memory with `LiftoverIntervalList` even with 32GB of memory.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


nh13:
I may need to also make a `IntervalListWriter` so we don't need the whole interval list in memory... Sigh...

src/main/java/htsjdk/samtools/util/IntervalCodec.java
```diff
@@ -0,0 +1,119 @@
...
+ */
+package htsjdk.samtools.util;
+
+import htsjdk.samtools.SAMSequenceDictionary;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+public class IntervalCodec implements SortingCollection.Codec<Interval> {
+
+    private SAMSequenceDictionary dict;
```
lbergelson:
```suggestion
    private final SAMSequenceDictionary dict;
```

src/test/java/htsjdk/samtools/util/IntervalCodecTest.java
```diff
@@ -0,0 +1,89 @@
...
+        final IntervalList actualList = new IntervalList(this.dict);
+
+        expectedList.add(new Interval("chr1", 50, 150));
+        expectedList.add(new Interval("chr1", 150, 250));
+        expectedList.add(new Interval("chr2", 50, 150));
+        expectedList.add(new Interval("chr3", 50, 150));
+        expectedList.add(new Interval("chr1", 50, 150, true, "number-5"));
+        expectedList.add(new Interval("chr1", 150, 250, false, "number-6"));
+
+        final OutputStream outputStream = new FileOutputStream(tempFile);
```
lbergelson: why not try-with-resoureces here?

nh13: Copy and pasted from another test file.

nh13:
@lbergelson commit bd872b5 is needed for this PR to be fully useful for the PR in Picard.  Thanks for the review!

src/main/java/htsjdk/samtools/util/IntervalListWriter.java
```diff
@@ -0,0 +1,90 @@
...
+import java.io.IOException;
+
+/**
+ * Writes out the list of intervals to the supplied file.  This class is particularly useful if we have many intervals
+ * to write, such that they all cannot be held in memory, for example in an {@link IntervalList}.
+ */
+public class IntervalListWriter implements Closeable {
+
+    private final BufferedWriter out;
+    private final FormatUtil format = new FormatUtil();
```
lbergelson: I know this is just moving existing code, but this use of FormatUtil seems unnecessary.   Wouldn't Integer.toString() serve the same purpose with less overhead?

src/test/java/htsjdk/samtools/util/IntervalListWriterTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+        final IntervalListWriter writer = new IntervalListWriter(tempFile, new SAMFileHeader(this.dict));
+        for (final Interval interval : expectedList.getIntervals()) {
+            writer.write(interval);
+        }
+        writer.close();
+
+        final IntervalList actualList = IntervalList.fromFile(tempFile);
+
+        Assert.assertEquals(
```
lbergelson: This should assert that the header is correct as well.

src/main/java/htsjdk/samtools/util/IntervalCoordinateComparator.java
```diff
@@ -0,0 +1,81 @@
...
+                return 1;
+            } else {
+                return lhs.getName().compareTo(rhs.getName());
+            }
+        }
+
+        return retval;
+    }
+}
+}
```
lbergelson: I think there's an extra `}`.  It's not compiling.

src/main/java/htsjdk/samtools/util/IntervalListWriter.java
```diff
@@ -41,18 +42,18 @@
...
 
     /** Creates a new writer, writing a header to the file.
-     * @param file a file to write to.  If exists it will be overwritten.
+     * @param path a file to write to.  If exists it will be overwritten.
      * @param header the header to write.
      */
-    public IntervalListWriter(final File file, final SAMFileHeader header) {
-        out = IOUtil.openFileForBufferedWriting(file);
+    public IntervalListWriter(final  Path path, final SAMFileHeader header) {
+        out = IOUtil.openFileForBufferedWriting(path.toFile());
```
lbergelson: This kind of defeats the purpose since it means that you can't actually use generic Paths, only ones that represent files.  I hadn't realized that we were missing path equivalents of this method though.  I've opened #1294 and https://github.com/samtools/htsjdk/pull/1296 to fix it.

PR #1291, opened 2019-02-19
Build and test on java 11

lbergelson:
### Description
This is an extension of #1269 which resolves some problems with the build in gradle 5+

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


lbergelson:
@cmnbroad Could you take a look at this when you get a chance?  It has a bunch of gradle changes because the `maven` publishing plugin broke when we moved to gradle 5 and rather than figure out what was going wrong I updated it to use the `maven-publish` plugin instead which is the new recommended one.

build.gradle
```diff
@@ -141,92 +142,79 @@ task testExternalApis(type: Test) {
...
+                url = 'http://samtools.github.io/htsjdk/'
 
                 developers {
                     developer {
-                        id 'picard'
-                        name 'Picard Team'
-                        url 'http://broadinstitute.github.io/picard'
+                        id = 'picard'
+                        name = 'Picard Team'
+                        url = 'http://broadinstitute.github.io/picard'
```
cmnbroad: Unrelated and tangential, but this seems like ancient history. And if you look there, you see nothing about htsjdk.

build.gradle
```diff
@@ -141,92 +142,79 @@ task testExternalApis(type: Test) {
...
-    archives sourcesJar
+    archiveClassifier.set('sources')
 }
 
 /**
  * Sign non-snapshot releases with our secret key.  This should never need to be invoked directly.
  */
 signing {
-    required { isRelease && gradle.taskGraph.hasTask("uploadArchives") }
+    required { isRelease && gradle.taskGraph.hasTask("publishHtsjdkPublicationToMavenRepository") }
```
cmnbroad: Is the uppercase 'H' in "Htsjdk" significant? It seems to work ok locally (it winds up in "htsjdk"ok).

gradlew
```diff
@@ -28,7 +28,7 @@ APP_NAME="Gradle"
 APP_BASE_NAME=`basename "$0"`
 
 # Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS=""
+DEFAULT_JVM_OPTS='"-Xmx64m"'
```
cmnbroad: Wondering what triggered this change ?

build.gradle
lbergelson: Yeah, I noticed that too but decided to leave it for later...  I'll open a follow up to change it.  Not sure to what though...

build.gradle
lbergelson: I don't believe it's meaningful.  I think it's case insensitive but the examples on the gradle website use camel case so I did as well.

gradlew
lbergelson: Rebuilding the wrapper triggered it.  I think they reduced gradle's memory requirement so they added a small default one.

PR #1277, opened 2019-02-13
Fix spotbugs issues in tests.

tomwhite:
Addresses #1267 for the tests (except for some valid violation that will be excluded in another PR).

### Checklist

- [ ] Code compiles correctly
- [ ] New tests covering changes and new functionality
- [ ] All tests passing
- [ ] Extended the README / documentation, if necessary
- [ ] Is not backward compatible (breaks binary or source compatibility)

PR #1303, opened 2019-02-25
Spotbugs remaining fixes

tomwhite:
Fix all the remaining issues found by Spotbugs (in addition to those in #1277 and #1278). In some cases it is not clear if there is a fix, or what the best fix would be that maintains existing behaviour, so I've added these cases to an exclude list. With the changes to the Gradle build file new violations will be picked up by causing the build to fail.

/cc @lbergelson 

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/TextualBAMIndexWriter.java
+++ b/src/main/java/htsjdk/samtools/TextualBAMIndexWriter.java
@@ -71,13 +71,13 @@ class TextualBAMIndexWriter implements BAMIndexWriter {
     @Override
     public void writeReference(final BAMIndexContent content) {
 
-        final int reference = content.getReferenceSequence();
-
         if (content == null) {
-            writeNullContent(reference);
+            writeNullContent();
             count++;
             return;
         }
+
+        final int reference = content.getReferenceSequence();
 
         if (reference != count){
             throw new SAMException("Reference on content is " + reference + " but expecting reference " + count);
@@ -161,6 +161,11 @@ class TextualBAMIndexWriter implements BAMIndexWriter {
                     " end: " + Long.toString(metaData.getUnalignedRecordCount(), 16));
         }
 
+    }
+
+    private void writeNullContent() {
+        pw.println("Reference <unknown> has n_bin=0");
+        pw.println("Reference <unknown> has n_intv=0");
     }
        
     private void writeNullContent(final int reference) {
--- a/src/main/java/htsjdk/samtools/TextualBAMIndexWriter.java
+++ b/src/main/java/htsjdk/samtools/TextualBAMIndexWriter.java
@@ -72,9 +72,7 @@ class TextualBAMIndexWriter implements BAMIndexWriter {
     public void writeReference(final BAMIndexContent content) {
 
         if (content == null) {
-            writeNullContent();
-            count++;
-            return;
+            throw new NullPointerException("BAMIndexContent cannot be null");
         }
 
         final int reference = content.getReferenceSequence();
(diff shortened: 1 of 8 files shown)


src/main/java/htsjdk/samtools/cram/compression/rans/E14.java
```diff
@@ -43,7 +43,7 @@ static int compress(final ByteBuffer in, final RansEncSymbol[][] syms,
         // Deal with the remainder
         l3 = 0xFF & in.get(in_size - 1);
         for (i3 = in_size - 2; i3 > 4 * isz4 - 2 && i3 >= 0; i3--) {
-            final int c3 = 0xFF & in.get(i3 > -1 ? i3 : 0);
+            final int c3 = 0xFF & in.get(i3);
```
pshapiro4broad: I thought this looked suspicious but I missed that the `for` loop directly restricts `i3`, e.g. `i3 >= 0`

src/main/java/htsjdk/tribble/readers/AsciiLineReaderIterator.java
```diff
@@ -75,7 +75,7 @@ public String peek() {
     private class TupleIterator extends AbstractIterator<Tuple<String, Long>> implements LocationAware {
         
         public TupleIterator() {
-            hasNext(); // Initialize the iterator, which appears to be a requirement of the parent class.  TODO: Really?
+            super.hasNext(); // Initialize the iterator, which appears to be a requirement of the parent class.  TODO: Really?
```
pshapiro4broad: Why add `super.`? This class doesn't implement `hasNext()` and it's `private`.

tomwhite: It's to make it clear that it's `TupleIterator`'s superclass (`AbstractIterator`) that the `hasNext()` method resolves to, as opposed to `hasNext()` on the outer class, `AsciiLineReaderIterator`.

The default is the superclass, so this doesn't change the behaviour, it just makes it clear (and stops Spotbugs complaining). See https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#ia-potentially-ambiguous-invocation-of-either-an-inherited-or-outer-method-ia-ambiguous-invocation-of-inherited-or-outer-method

src/main/java/htsjdk/samtools/TextualBAMIndexWriter.java
```diff
@@ -71,14 +71,14 @@ private void writeHeader() {
     @Override
     public void writeReference(final BAMIndexContent content) {
 
-        final int reference = content.getReferenceSequence();
-
         if (content == null) {
-            writeNullContent(reference);
+            writeNullContent();
```
pshapiro4broad: This changes the behavior; before this would throw a `NullPointerException`. Is the new behavior the correct/desired behavior of this API?

tomwhite: The NPE was introduced in https://github.com/samtools/htsjdk/commit/ee10696af44fcf6d0f2166b14884f7ae7b0e6aeb#diff-d029b962e0444d3ad90c9c69372591bcL70. This fix removes the NPE, which seems like the right thing to do.

pshapiro4broad: It was introduced a month after the initial checkin of the code, which was 8 years ago. So it's worked this way for the lifetime of this code. If `content == null` is not a valid state, it would be better to remove this code rather than add a new feature.

tomwhite: Fair enough. I've fixed this to explicitly throw a NPE in this case.

lbergelson:
Rebased and merged this as #1331

PR #1309, opened 2019-02-27
making changes to reduce size of giant interval lists

lbergelson:
### Description
Two separate changes which should reduce the in-memory size of IntervalList and OverlapDetector.

1: Cache the last seen contig name when loading IntervalLists and reuse the same String to avoid keeping many duplicate copies of "chr1" in memory. This should have very minor overhead since interval files are sorted it should be about as effective as a more complicated caching scheme.

2.  Use SingletonSets as the leaf objects in OverlapDetector when constructing it.  These are substantially smaller than HashSets of the same size.  Change to HashSet only when we find 2 or more identically positioned elements.  This will have a minor performance cost while constructing the OverlapDetector but should result in dramatically decreased size in memory.  

### Checklist

- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/IntervalTree.java
+++ b/src/main/java/htsjdk/samtools/util/IntervalTree.java
@@ -60,8 +60,6 @@ public class IntervalTree<V> implements Iterable<IntervalTree.Node<V>>
         mRoot = null;
     }
 
-
-
     /**
      * Put a new interval into the tree (or update the value associated with an existing interval).
      * If the interval is novel, the special sentinel value is returned.
@@ -129,15 +127,17 @@ public class IntervalTree<V> implements Iterable<IntervalTree.Node<V>>
      * @param start interval start position
      * @param end interval end position
      * @param value value to merge into the tree, must not be equal to the sentinel value
-     * @param remappingFunction a function that merges the new value with the existing value for the same start and end position
-     * @return the updated value that is now stored in the tree or sentinel
+     * @param remappingFunction a function that merges the new value with the existing value for the same start and end position,
+     *                          if the function returns the sentinel value then the mapping will be unset
+     * @return the updated value that is stored in the tree after the completion of this merge operation, this will
+     * be the sentinel value if nothing ended up being stored
      */
     public V merge(int start, int end, V value, BiFunction<? super V,  ? super V, ? extends V> remappingFunction) {
         ValidationUtils.validateArg(!Objects.equals(value, mSentinel), "Values equal to the sentinel value may not be merged");
         final V alreadyPresent = put(start, end, value);
         if (!Objects.equals(alreadyPresent, mSentinel)) {
             final V newComputedValue = remappingFunction.apply(value, alreadyPresent);
-            if(Objects.equals(newComputedValue, mSentinel)){
+            if (Objects.equals(newComputedValue, mSentinel)) {
                 remove(start, end);
             } else {
                 put(start, end, newComputedValue);
--- a/src/main/java/htsjdk/samtools/util/IntervalTree.java
+++ b/src/main/java/htsjdk/samtools/util/IntervalTree.java
@@ -23,9 +23,13 @@
  */
 package htsjdk.samtools.util;
 
+import htsjdk.utils.ValidationUtils;
+
 import java.util.ConcurrentModificationException;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.function.BiFunction;
 
 /**
  * A Red-Black tree with intervals for keys.
@@ -56,6 +60,8 @@ public class IntervalTree<V> implements Iterable<IntervalTree.Node<V>>
         mRoot = null;
     }
 
+
+
     /**
      * Put a new interval into the tree (or update the value associated with an existing interval).
      * If the interval is novel, the special sentinel value is returned.
@@ -64,7 +70,6 @@ public class IntervalTree<V> implements Iterable<IntervalTree.Node<V>>
      * @param value The associated value.
      * @return The old value associated with that interval, or the sentinel.
      */
-    @SuppressWarnings("null")
     public V put( final int start, final int end, final V value )
     {
         if ( start > end )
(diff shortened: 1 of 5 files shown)


lbergelson:
@samuelklee Simple changes to reduce the in memory size of interval lists. 
@nh13 This might benefit you as well since it sounds like you're dealing with very large interval lists as well.

samuelklee:
Thanks for making these changes!  I think the per-contig OverlapDetector you suggested was enough to bring down memory in CollectReadCounts, so I didn't see any change there.  However, I do see a substantial improvement in ModelSegments, which still uses global OverlapDetectors.  I was able to run with ~10M bins and <8GB memory, while master required ~10GB.  Runtime also actually went down slightly from 18 minutes to 17 minutes, but that might be within the noise.

src/main/java/htsjdk/samtools/util/OverlapDetector.java
```diff
@@ -82,13 +82,19 @@ public void addLhs(final T object, final Locatable interval) {
...
-        final Set<T> objects = new HashSet<>(1);
-        objects.add(object);
+        final Set<T> newValue = Collections.singleton(object);
         if (start <= end) {  // Don't put in sequences that have no overlappable bases
-            final Set<T> alreadyThere = tree.put(start, end, objects);
+            final Set<T> alreadyThere = tree.put(start, end, newValue);
             if (alreadyThere != null) {
-                alreadyThere.add(object);
-                tree.put(start, end, alreadyThere);
+                if( alreadyThere.size() == 1){
```
yfarjoun: I think that this should be implemented within the IntervalTree class as a `computeIfPresent` method...

src/main/java/htsjdk/samtools/util/OverlapDetector.java
```diff
@@ -82,13 +82,19 @@ public void addLhs(final T object, final Locatable interval) {
...
+                    Set<T> mutableSet = new HashSet<>(2);
```
yfarjoun: hmmm. better find out if you got a mutable one and just add the new element if so...no?

lbergelson: You mean with an instanceof check?  As it's written it should be immutable always if it's exactly size 1 and mutable otherwise.  It's a bit awkward but I thought maybe relying on the size was better than relying on the class.  I can change that though.

lbergelson:
@yfarjoun Could you take a look.  The implementation in interval tree is slightly more complicated because it has to deal with arbitrary sentinel values instead of just null, and I added some tests because I realized that the conditions I cared about didn't look well covered.

src/main/java/htsjdk/samtools/util/IntervalTree.java
```diff
@@ -114,6 +119,34 @@ public V put( final int start, final int end, final V value )
...
+     * If the specified start and end positions are not already associated with a value or are
+     * associated with the sentinel ( see {@link #getSentinel()}, associates it with the given (non-sentinel) value.
+     * Otherwise, replaces the associated value with the results of the given
+     * remapping function, or removes if the result is equal to the sentinel value. This
+     * method may be of use when combining multiple values that have the same start and end position.
+     *
+     * @param start interval start position
+     * @param end interval end position
+     * @param value value to merge into the tree, must not be equal to the sentinel value
+     * @param remappingFunction a function that merges the new value with the existing value for the same start and end position
```
yfarjoun: note that if function returns null the region will be left empty

src/main/java/htsjdk/samtools/util/IntervalTree.java
```diff
@@ -114,6 +119,34 @@ public V put( final int start, final int end, final V value )
...
+     * @return the updated value that is now stored in the tree or sentinel
```
yfarjoun: `now` is not clear enough...(does it refer to before or after the application of the method?) 
perhaps `the updated value that is eventually stored in the tree or the sentinel value if nothing ends up being stored`

src/main/java/htsjdk/samtools/util/OverlapDetector.java
```diff
@@ -80,25 +81,35 @@ public void addLhs(final T object, final Locatable interval) {
...
-                    mutableSet.addAll(alreadyThere);
-                    mutableSet.add(object);
-                    tree.put(start, end, mutableSet);
-                } else {
-                    alreadyThere.add(object);
-                    tree.put(start, end, alreadyThere);
-                }
-            }
+            tree.merge(start, end,
+                                  Collections.singleton(object),
```
yfarjoun: wierd spacing

src/main/java/htsjdk/samtools/util/IntervalTree.java
```diff
@@ -56,6 +60,8 @@ public void clear()
         mRoot = null;
     }
 
+
```
yfarjoun: extra nls

src/main/java/htsjdk/samtools/util/OverlapDetector.java
```diff
@@ -80,25 +81,35 @@ public void addLhs(final T object, final Locatable interval) {
...
 
+    /**
+     * merge two Sets, assumes sets of size 1 are immutale
+     */
+    private static <T> BiFunction<Set<T>, Set<T>, Set<T>> mergeSetsAccountingForSingletons() {
+        return (newValue, oldValue) -> {
+            // Sets of size 1 are immutable SingletonSets so we have to make a new
+            // mutable one to add to
+            if (oldValue.size() == 1) {
+                Set<T> newMutableSet = new HashSet<>();
```
yfarjoun: can replace entire if statement with:
```
final Set<T> newMutableSet = oldValue.size() == 1 ? new HashSet<>() : oldValue;
newMutableSet.addAll(oldValue);
newMutableSet.addAll(newValue);
return newMutableSet;
```

src/main/java/htsjdk/samtools/util/IntervalTree.java
```diff
@@ -114,6 +119,34 @@ public V put( final int start, final int end, final V value )
...
+     * @param value value to merge into the tree, must not be equal to the sentinel value
+     * @param remappingFunction a function that merges the new value with the existing value for the same start and end position
+     * @return the updated value that is now stored in the tree or sentinel
+     */
+    public V merge(int start, int end, V value, BiFunction<? super V,  ? super V, ? extends V> remappingFunction) {
+        ValidationUtils.validateArg(!Objects.equals(value, mSentinel), "Values equal to the sentinel value may not be merged");
+        final V alreadyPresent = put(start, end, value);
+        if (!Objects.equals(alreadyPresent, mSentinel)) {
+            final V newComputedValue = remappingFunction.apply(value, alreadyPresent);
+            if(Objects.equals(newComputedValue, mSentinel)){
```
yfarjoun: spaces

src/main/java/htsjdk/samtools/util/IntervalTree.java
lbergelson: done

src/main/java/htsjdk/samtools/util/IntervalTree.java
lbergelson: done

src/main/java/htsjdk/samtools/util/IntervalTree.java
lbergelson: I went with a slight modification:
```
the updated value that is stored in the tree after the completion of this merge operation, this will be the sentinel value if nothing ended up being stored
``

src/main/java/htsjdk/samtools/util/IntervalTree.java
lbergelson: done

src/main/java/htsjdk/samtools/util/OverlapDetector.java
lbergelson: fixed

src/main/java/htsjdk/samtools/util/OverlapDetector.java
lbergelson: Oh right, clever, it's a set so you can unconditionally add the same thing over again...

lbergelson:
@yfarjoun It's an improvement.  Thanks for the review.

lbergelson:
@yfarjoun Thanks for the reviews!

PR #1310, opened 2019-02-28
Fix bug in BlockCompressedInputStream.checkTermination()

### LINKED ISSUE #1251: BlockCompressedInputStream.readFully seems likely to be buggy
My implementation of `BlockCompressedInputStream.readFully` seems like it will potentially incorrectly throw on channels that don't always provide all requested bytes when reading, it should probably read in a loop until 0 bytes are read and maybe needs to assert that the channel is in synchronous mode as well.

I'm not sure what I was thinking when I wrote it, maybe I looked deeper and found an explanation for why it's ok which I've now forgotten.

It should either be changed or documented better to explain why it's ok.

lbergelson:
* Fixing a bug in BlockCompressedInputStream.checkTermination()
  The internal method readFully would throw in cases where less bytes were read than were available in the stream.
  We haven't observed this in practice but it's a valid behavior of SeekableByteChannel.read() so we should honor it.
* Fixes #1251

### Description

### Checklist

- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
+++ b/src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
@@ -689,14 +689,15 @@ public class BlockCompressedInputStream extends InputStream implements LocationA
 
     /**
      * read as many bytes as dst's capacity into dst or throw if that's not possible
+     *
      * @throws EOFException if channel has fewer bytes available than dst's capacity
      */
     static void readFully(SeekableByteChannel channel, ByteBuffer dst) throws IOException {
         int totalBytesRead = 0;
         final int capacity = dst.capacity();
-        while( totalBytesRead < capacity){
+        while (totalBytesRead < capacity) {
             final int bytesRead = channel.read(dst);
-            if( bytesRead == -1 ){
+            if (bytesRead == -1) {
                 throw new EOFException();
             }
             totalBytesRead += bytesRead;
--- a/src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
+++ b/src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
@@ -692,9 +692,14 @@ public class BlockCompressedInputStream extends InputStream implements LocationA
      * @throws EOFException if channel has fewer bytes available than dst's capacity
      */
     static void readFully(SeekableByteChannel channel, ByteBuffer dst) throws IOException {
-        final int bytesRead = channel.read(dst);
-        if (bytesRead < dst.capacity()){
-            throw new EOFException();
+        int totalBytesRead = 0;
+        final int capacity = dst.capacity();
+        while( totalBytesRead < capacity){
+            final int bytesRead = channel.read(dst);
+            if( bytesRead == -1 ){
+                throw new EOFException();
+            }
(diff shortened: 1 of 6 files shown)


src/main/java/htsjdk/samtools/util/BlockCompressedInputStream.java
```diff
@@ -692,9 +692,14 @@ public static FileTermination checkTermination(SeekableByteChannel channel) thro
      * @throws EOFException if channel has fewer bytes available than dst's capacity
      */
     static void readFully(SeekableByteChannel channel, ByteBuffer dst) throws IOException {
-        final int bytesRead = channel.read(dst);
-        if (bytesRead < dst.capacity()){
-            throw new EOFException();
+        int totalBytesRead = 0;
```
pshapiro4broad: Please reformat using IDE to fix whitespace, in all three files.

src/test/java/htsjdk/samtools/util/BlockCompressedTerminatorTest.java
```diff
@@ -120,6 +120,35 @@ public void testReadFullyThrowWhenItCantReadEnough() throws IOException {
...
     }
 
+    @Test
+    public void testReadFullyReadsBytesCorrectlyWhenPartialReadOccurs() throws IOException {
+        final byte[] expected = "something to test reading from".getBytes();
+        final ByteBuffer buffer = ByteBuffer.wrap(expected);
+        try(final SeekableByteChannel channel = new OneByteAtATimeChannel(buffer)){
+            final int readBufferSize = 10;
+            final ByteBuffer readBuffer = ByteBuffer.allocate(readBufferSize);
+           // Assert.assertTrue(channel.size() >= readBuffer.capacity());
```
pshapiro4broad: Delete or uncomment commented code

src/test/java/htsjdk/samtools/SeekableByteChannelFromBuffer.java
```diff
@@ -77,6 +77,10 @@ public void close() throws IOException {
     open = false;
   }
 
+  protected ByteBuffer getBuffer(){
```
pshapiro4broad: Seems a shame to expose this for testing purposes only.  If the subclass type was defined in the same package, you could use package (default) scope instead, which is more restrictive than protected, and doesn't lead to API compatibility requirements later on.

So you could use default scope here if you defined `OneByteAtATimeChannel` in the same package as `SeekableByteChannelFromBuffer`. You could put the source in the test directory so it wouldn't be part of the public API.

lbergelson: This class is already in the test directories so it isn't part of the public API.  I don't mind moving them both to a subpackage though and restricting it to default scope.

lbergelson: moved both to htsjdk.testutil.streams

src/test/java/htsjdk/samtools/util/BlockCompressedTerminatorTest.java
lbergelson: done

PR #1311, opened 2019-02-28
Fix bug when loading indexed bgzip fasta file.

### LINKED ISSUE #1290: BlockCompressedIndexedFastaSequenceFile checks for a gzi even if an index is passed in.
It looks like ` new BlockCompressedIndexedFastaSequenceFile( Path, FastaSequenceIndex, GZIIndex)` inappropriately `Files.exists(GZIIndex.resolveIndexNameForBgzipFile(path)` even though the GZIIndex has been preloaded and passed in as part of `canCreateBlockCompresedIndexedFastaSequence`

lbergelson:
* There was a bug that prevented loading an indexed bgzip fasta file if the index location was specified but wasn't next to the fasta in the file system.
* Fixes #1290

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/reference/BlockCompressedIndexedFastaSequenceFile.java
+++ b/src/main/java/htsjdk/samtools/reference/BlockCompressedIndexedFastaSequenceFile.java
@@ -100,8 +100,8 @@ public class BlockCompressedIndexedFastaSequenceFile extends AbstractIndexedFast
     private static void assertIsBlockCompressed(final Path path) {
         try {
             // check if the it is a valid block-compressed file and if the .gzi index exits
-            if(!IOUtil.isBlockCompressed(path, true)){
-              throw new SAMException("Invalid block-compressed Fasta file: " + path);
+            if (!IOUtil.isBlockCompressed(path, true)) {
+                throw new SAMException("Invalid block-compressed Fasta file: " + path);
             }
         } catch (IOException e) {
             throw new SAMException("Invalid block-compressed Fasta file: " + path, e);
--- a/src/main/java/htsjdk/samtools/reference/BlockCompressedIndexedFastaSequenceFile.java
+++ b/src/main/java/htsjdk/samtools/reference/BlockCompressedIndexedFastaSequenceFile.java
@@ -26,19 +26,15 @@ package htsjdk.samtools.reference;
 
 import htsjdk.samtools.SAMException;
 import htsjdk.samtools.SAMSequenceDictionary;
-import htsjdk.samtools.seekablestream.ReadableSeekableStreamByteChannel;
 import htsjdk.samtools.seekablestream.SeekablePathStream;
 import htsjdk.samtools.seekablestream.SeekableStream;
 import htsjdk.samtools.util.BlockCompressedInputStream;
 import htsjdk.samtools.util.GZIIndex;
 import htsjdk.samtools.util.IOUtil;
 
-import java.io.BufferedInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
-import java.io.InputStream;
 import java.nio.ByteBuffer;
-import java.nio.file.Files;
 import java.nio.file.Path;
 
 /**
@@ -55,7 +51,7 @@ public class BlockCompressedIndexedFastaSequenceFile extends AbstractIndexedFast
 
     public BlockCompressedIndexedFastaSequenceFile(final Path path)
             throws FileNotFoundException {
-        this(path,new FastaSequenceIndex((findRequiredFastaIndexFile(path))));
+        this(path, new FastaSequenceIndex((findRequiredFastaIndexFile(path))));
     }
 
     public BlockCompressedIndexedFastaSequenceFile(final Path path, final FastaSequenceIndex index) {
@@ -67,9 +63,7 @@ public class BlockCompressedIndexedFastaSequenceFile extends AbstractIndexedFast
         if (gziIndex == null) {
             throw new IllegalArgumentException("null gzi index");
         }
-        if (!canCreateBlockCompresedIndexedFastaSequence(path)) {
-            throw new SAMException("Invalid block-compressed Fasta file");
-        }
+        assertIsBlockCompressed(path);
         try {
             stream = new BlockCompressedInputStream(new SeekablePathStream(path));
             gzindex = gziIndex;
@@ -103,12 +97,14 @@ public class BlockCompressedIndexedFastaSequenceFile extends AbstractIndexedFast
         }
     }
 
-    private static boolean canCreateBlockCompresedIndexedFastaSequence(final Path path) {
+    private static void assertIsBlockCompressed(final Path path) {
         try {
             // check if the it is a valid block-compressed file and if the .gzi index exits
-            return IOUtil.isBlockCompressed(path, true) && Files.exists(GZIIndex.resolveIndexNameForBgzipFile(path));
+            if(!IOUtil.isBlockCompressed(path, true)){
+              throw new SAMException("Invalid block-compressed Fasta file: " + path);
+            }
         } catch (IOException e) {
-            return false;
+            throw new SAMException("Invalid block-compressed Fasta file: " + path, e);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/reference/BlockCompressedIndexedFastaSequenceFile.java
```diff
@@ -103,12 +97,14 @@ private static GZIIndex loadFastaGziIndex(final Path path) {
         }
     }
 
-    private static boolean canCreateBlockCompresedIndexedFastaSequence(final Path path) {
+    private static void assertIsBlockCompressed(final Path path) {
         try {
             // check if the it is a valid block-compressed file and if the .gzi index exits
-            return IOUtil.isBlockCompressed(path, true) && Files.exists(GZIIndex.resolveIndexNameForBgzipFile(path));
+            if(!IOUtil.isBlockCompressed(path, true)){
```
pshapiro4broad:
```suggestion
            if (!IOUtil.isBlockCompressed(path, true)) {
```

src/main/java/htsjdk/samtools/reference/BlockCompressedIndexedFastaSequenceFile.java
```diff
@@ -103,12 +97,14 @@ private static GZIIndex loadFastaGziIndex(final Path path) {
...
+              throw new SAMException("Invalid block-compressed Fasta file: " + path);
```
pshapiro4broad:
```suggestion
                throw new SAMException("Invalid block-compressed Fasta file: " + path);
```

src/test/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFileTest.java
```diff
@@ -27,10 +27,7 @@
 import htsjdk.HtsjdkTest;
 import htsjdk.samtools.SAMException;
 import htsjdk.samtools.seekablestream.SeekableFileStream;
-import htsjdk.samtools.util.CloserUtil;
-import htsjdk.samtools.util.GZIIndex;
-import htsjdk.samtools.util.RuntimeIOException;
-import htsjdk.samtools.util.StringUtil;
+import htsjdk.samtools.util.*;
```
pshapiro4broad: I think we should avoid using `*` imports as it makes it harder to see which package an import came from outside of an IDE. There's an IDE setting that will automatically fix this when adding/removing imports.

src/main/java/htsjdk/samtools/reference/BlockCompressedIndexedFastaSequenceFile.java
```diff
@@ -103,12 +97,14 @@ private static GZIIndex loadFastaGziIndex(final Path path) {
         }
     }
 
-    private static boolean canCreateBlockCompresedIndexedFastaSequence(final Path path) {
+    private static void assertIsBlockCompressed(final Path path) {
         try {
             // check if the it is a valid block-compressed file and if the .gzi index exits
-            return IOUtil.isBlockCompressed(path, true) && Files.exists(GZIIndex.resolveIndexNameForBgzipFile(path));
```
pshapiro4broad: Is the presence of the index file asserted elsewhere? Or is it not required as we've already asserted `gziIndex != null`?

src/test/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFileTest.java
```diff
@@ -39,13 +36,17 @@
...
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 
 /**
  * Test the indexed fasta sequence file reader.
  */
 public class AbstractIndexedFastaSequenceFileTest extends HtsjdkTest {
     private static final File TEST_DATA_DIR = new File("src/test/resources/htsjdk/samtools/reference");
     private static final File SEQUENCE_FILE = new File(TEST_DATA_DIR,"Homo_sapiens_assembly18.trimmed.fasta");
+    private static final File SEQUENCE_FILE_DICT = new File(TEST_DATA_DIR,"Homo_sapiens_assembly18.trimmed.dict");
```
pshapiro4broad: New field is unused in this PR, is that intentional? If so, you may want to add a comment as to why it's defined here so it's not deleted later on.

src/test/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFileTest.java
```diff
@@ -353,4 +354,16 @@ public void testBadInputForIndexedFastaSequenceFile() throws Exception {
     public void testBadInputForBlockCompressedIndexedFastaSequenceFile() throws Exception {
         new BlockCompressedIndexedFastaSequenceFile(SEQUENCE_FILE.toPath());
     }
+
+    @Test
+    public void testCanCreateBlockCompressedIndexedWithSpecifiedGZIAndDict() throws IOException {
+        final Path moved = Files.createTempFile("moved", ".fasta.gz");
+        Files.copy(SEQUENCE_FILE_BGZ.toPath(), moved, StandardCopyOption.REPLACE_EXISTING);
+        IOUtil.deleteOnExit(moved);
+        try( ReferenceSequenceFile withNoAdacentIndex = new BlockCompressedIndexedFastaSequenceFile(moved, new FastaSequenceIndex(SEQUENCE_FILE_INDEX), GZIIndex.loadIndex(SEQUENCE_FILE_GZI.toPath()));
```
pshapiro4broad:
```suggestion
        try (ReferenceSequenceFile withNoAdacentIndex = new BlockCompressedIndexedFastaSequenceFile(moved, new FastaSequenceIndex(SEQUENCE_FILE_INDEX), GZIIndex.loadIndex(SEQUENCE_FILE_GZI.toPath()));
```

src/test/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFileTest.java
```diff
@@ -353,4 +354,16 @@ public void testBadInputForIndexedFastaSequenceFile() throws Exception {
...
+             ReferenceSequenceFile withFilesAdjacent = new BlockCompressedIndexedFastaSequenceFile(SEQUENCE_FILE_BGZ.toPath())){
```
pshapiro4broad: Indent looks wrong on this line but it's hard to tell for sure in the PR. I would expect it to line up with the text after the `(` on the previous line.

src/test/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFileTest.java
```diff
@@ -353,4 +354,16 @@ public void testBadInputForIndexedFastaSequenceFile() throws Exception {
...
+            Assert.assertEquals(withNoAdacentIndex.getSubsequenceAt("chrM", 100, 1000).getBases(),
+            withFilesAdjacent.getSubsequenceAt("chrM", 100, 1000).getBases());
```
pshapiro4broad:
```suggestion
                withFilesAdjacent.getSubsequenceAt("chrM", 100, 1000).getBases());
```

src/main/java/htsjdk/samtools/reference/BlockCompressedIndexedFastaSequenceFile.java
lbergelson: Yeah, the check for gziIndex != null is the correct one here.  The bug was that this was checking for an index next to your file even if you passed in one that was somewhere else.

src/test/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFileTest.java
lbergelson: good catch, it was used in an earlier version of the test but it turned out to be unecessary

src/test/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFileTest.java
lbergelson: It's lined up with the beginning of the other ReferenceSequenceFile, so it's 1 space after the `(` I think it's what you expect.  Auto format doesn't change anything.

lbergelson:
@pshapiro4broad Thanks for the review!

PR #1314, opened 2019-03-05
Fix CSI bug when querying files with long references.

### LINKED ISSUE #1047: Trouble loading bam file in IGV
Hello,

I work with Barley where most of the sequences are longer than 2^29-1. I have RNA-Seq data which map to regions beyond 2^29 on the chromosomes. I generally provide the .bam file to IGV. For this I need to generate the .bai file using samtools index. But this command fails since the there are regions which cannot be included in the bai index. Instead samtools recommends me to generate a .csi index file. IGV on the other hand cannot work with .csi index files. IGV used htsjdk for loading the files. Could you please suggest an alternative?

Thank you.

lbergelson:
* Fixing a bug that caused querying a bam to incorrectly return no results when reading from a CSI.
  This occurred in cases when the reference was longer than those supported by bai files.
  It was caused by an integer overflow in GenomicIndexRegionUtils.regionToBins().

I've increased the size of the intermediate values used in the computation to `long` which seems to fix the problem.

* Refactoring GenomicIndexRegionUtils.regionToBins() for clarity, it's more javaish now.
* Adding additional test cases.
* Refactoring tests to use dataproviders when possible.
* This should fix #1047

fixing query bug where some csi files would silently fail queries do to an overflow

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


lbergelson:
@valeriuo If you have a chance could you look at the changes here and see if they seem sane to you?  The meat of the fix is changing some of the intermediate values to long  since those were overflowing, but then I did a little bit of refactoring on the code as well.

src/test/java/htsjdk/samtools/CSIIndexTest.java
```diff
@@ -210,73 +257,110 @@ public static void testGetParentBinNumber() {
...
+    }
+
+    @DataProvider
+    public Object[][] getTestRegionToBinsData() {
+        return new Object[][]{
+                //start, end, minShift, binDepth, expectedBins
+                {(1 << 12) * 98403 + 4094, (1 << 12) * 98404 + 1, 12, 7, Arrays.asList(0, 4, 33, 265, 2122, bin9.getBinNumber() - 1, bin9.getBinNumber(), ucsi.getParentBinNumber(bin9.getBinNumber()))},
+                {939520000, 939529000, 12, 7, Arrays.asList(0, 7, 8, 64, 65, 520, 521, 4168, 4169, 33352, 33353, 266823, 266824, 266825, 266826)},
+                //this case tests https://github.com/samtools/htsjdk/issues/1047
+                {558419286, 558424804, 14, 7, Arrays.asList(0, 2, 17, 139, 1117, 8941, 71532)}
```
droazen: Add comment here noting that these values test integer overflow conditions

src/test/java/htsjdk/samtools/CSIIndexTest.java
```diff
@@ -28,18 +32,20 @@
     private static Bin bin8 = new Bin(1, 4689);
     private static Bin bin9 = new Bin(1, 135853);
     private static Bin bin10 = new Bin(1, 163);
+    public static final int BAI_EQUIVALENT_MIN_SHIFT = 14;
+    public static final int BAI_EQUIVALENT_BIN_DEPTH = 6;
```
droazen: Add comments explaining what these values mean.

src/main/java/htsjdk/samtools/GenomicIndexUtil.java
```diff
@@ -110,44 +112,47 @@ public static BitSet regionToBins(final int startPos, final int endPos) {
...
-            binWidth = binWidth - 3;
-            level = level + 1;
+        int firstBinOnLevel = 1;
+        int binWidth = minShift + 3 * (binDepth - 2);
+
+        final BitSet bins = new BitSet(((1 << 3 * binDepth) - 1) / 7);
+        bins.set(0);
+        for (int level = 1; level < binDepth; level++) {
+            final int startBin = (int) (start >> binWidth) + firstBinOnLevel;
+            final int endBin = (int) (end >> binWidth) + firstBinOnLevel;
```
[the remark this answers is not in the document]
valeriuo: You would have to go to a depth larger than 11 to overflow the bin number and there is hardly any reason to go that far. Even if you want to put the entire human genome in bin 0, going to level 11 would mean indexing every 4 bases.

lbergelson:
@valeriuo Thank you for taking a look.  Changing everything to support `long` is going to be a much bigger change than what we can make right here unfortunately.

Issue #1316, opened 2019-03-06
Changing to use SingletonList in Sam


lbergelson:
Testing the newest version of htsjdk against picard fails with:
```
picard.analysis.CollectGcBiasMetricsTest.setupBuilder FAILED
    java.lang.UnsupportedOperationException
        at java.util.AbstractList.add(AbstractList.java:148)
        at java.util.AbstractList.add(AbstractList.java:108)
        at htsjdk.samtools.SAMFileHeader.addReadGroup(SAMFileHeader.java:192)
        at htsjdk.samtools.SAMRecordSetBuilder.setReadGroup(SAMRecordSetBuilder.java:163)
        at picard.analysis.CollectGcBiasMetricsTest.setupTest1(CollectGcBiasMetricsTest.java:407)
        at picard.analysis.CollectGcBiasMetricsTest.setupBuilder(CollectGcBiasMetricsTest.java:109)
```

This happens because of the change in https://github.com/samtools/htsjdk/pull/1286/ which changed ArrayList -> Collections.singletonList.  

```
    final List<SAMReadGroupRecord> readGroups = Collections.singletonList(readGroupRecord);
```

PR #1317, opened 2019-03-06
Revert singletonList optimization in SamRecordSetBuilder

lbergelson:
* A trivial optimization to use a singletonList instead of an ArrayList in SamRecordSet builder broke downstream tests because SamFileHeader.setReadGroups explicitly doesn't copy the list.
* Tests downstream expect the list to be mutable.
* Fixes https://github.com/samtools/htsjdk/issues/1316


lbergelson:
@yfarjoun @pshapiro4broad Even the most benign changes can break something...

PR #1318, opened 2019-03-06
Fixing zero-length interval bug in IntervalList.merge

lbergelson:
* Fixing a bug in IntervalList.merge() that caused it to break when merging 0 length intervals
* Improving tests for merge
* The bug was introduced in #1265 and hasn't been in any release

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/util/IntervalList.java
+++ b/src/main/java/htsjdk/samtools/util/IntervalList.java
@@ -379,13 +379,14 @@ public class IntervalList implements Iterable<Interval> {
     }
 
     /**
-     * Merges a sorted collection of intervals and optionally concatenates unique names or takes the first name.
+     * Merges a collection of intervals and optionally concatenates unique names or takes the first name.
+     * Uses the contig and strand of the first interval seen without validating that the others match.
      */
     static Interval merge(final Iterable<Interval> intervals, final boolean concatenateNames) {
         final Interval first = intervals.iterator().next();
         final String chrom = first.getContig();
         int start = first.getStart();
-        int end = start;
+        int end = first.getEnd();
         final boolean neg = first.isNegativeStrand();
         final LinkedHashSet<String> names = new LinkedHashSet<>();
         final String name;
--- a/src/main/java/htsjdk/samtools/util/IntervalList.java
+++ b/src/main/java/htsjdk/samtools/util/IntervalList.java
@@ -380,7 +380,11 @@ public class IntervalList implements Iterable<Interval> {
 
     /**
      * Merges a collection of intervals and optionally concatenates unique names or takes the first name.
-     * Uses the contig and strand of the first interval seen without validating that the others match.
+     * *
+     * @param concatenateNames if true, combine the names of all the intervals with |, otherwise use the name of the first interval.
+     * @return a single interval which spans from the minimum input start position to the maximum input end position.
+     * The resulting strandedness and contig are those of the first input with no validation.
+     *
      */
     static Interval merge(final Iterable<Interval> intervals, final boolean concatenateNames) {
         final Interval first = intervals.iterator().next();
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/util/IntervalList.java
```diff
@@ -379,13 +379,14 @@ public void unique(final boolean concatenateNames) {
...
-     * Merges a sorted collection of intervals and optionally concatenates unique names or takes the first name.
+     * Merges a collection of intervals and optionally concatenates unique names or takes the first name.
+     * Uses the contig and strand of the first interval seen without validating that the others match.
      */
     static Interval merge(final Iterable<Interval> intervals, final boolean concatenateNames) {
         final Interval first = intervals.iterator().next();
         final String chrom = first.getContig();
         int start = first.getStart();
-        int end = start;
+        int end = first.getEnd();
```
yfarjoun: what should happen when we merge a zero-length interval with a non-zerolength one after it?

so, 
10-9 merged with 11-15  should remain those two intervals
10-9 merged with 10-15 should become one interval 10-15
10-9 merged with 9-15 should become one interval 9-15

yfarjoun: nice catch. sorry.

lbergelson: I definitely didn't notice it in review.  It makes sense, you can't have an end before start..

lbergelson: The first case, you mean it should become 10-15?  Merge never leaves them separate.

lbergelson:
I added some more test cases to cover the cases you mentioned

lbergelson:
@yfarjoun I added the tests I think you wanted.  Is there anything else?

src/main/java/htsjdk/samtools/util/IntervalList.java
yfarjoun: no...I think that it should not become 10-15....since the size of the first interval is 0 it shouldn't change the total number of bases covered...

yfarjoun: I think that zero length intervals need to either be dropped when merging, or ignored...normal rules of merging do not apply.

src/test/java/htsjdk/samtools/util/IntervalListTest.java
```diff
@@ -564,22 +564,47 @@ public void testFromSequenceName(final Path intervalList, final String reference
...
+                {Collections.singletonList(normalInterval), true, normalInterval},
+                {Collections.singletonList(normalInterval), false, normalInterval},
+                {Collections.singletonList(zeroLengthInterval), true, zeroLengthInterval},
+                {Collections.singletonList(zeroLengthInterval), false, zeroLengthInterval},
+                {Arrays.asList(zeroLengthInterval, interval600To601), true, interval600To625},
+                {Arrays.asList(zeroLengthInterval, interval600To601), false, interval600To625},
+                {Arrays.asList(zeroLengthInterval, interval600To601), true, interval600To625},
+                {Arrays.asList(interval600To601, new Interval(contig, 100, 200, false, "hasName")), true, new Interval(contig, 100, 601, false, "hasName")},
+                {Arrays.asList(interval600To601, new Interval(contig, 100, 200, false, "hasName")), false, new Interval(contig, 100, 601, false, "hasName")},
+                {Arrays.asList(zeroInterval10To9, new Interval(contig, 11, 15)), false, new Interval(contig, 10, 15)},
```
yfarjoun: I don't think that this is right. adding a zero-length interval shouldn't add bases to another interval.

lbergelson:
@yfarjoun There exists problems with 0-length intervals... Unsurprisingly... I've opened a pr draft #1320 that shows some of them.  I'm not going to resolve them in this PR.

yfarjoun:
👍

PR #1319, opened 2019-03-08
CRAM: correct CT and MF encoding datatypes from byte to int

jmthibault79:
Also add a test demonstrating that the old byte-based code did not cause problems because byte and int codecs are identical for the full range of the fields 

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/cram/encoding/reader/CramRecordReader.java
+++ b/src/main/java/htsjdk/samtools/cram/encoding/reader/CramRecordReader.java
@@ -34,7 +34,7 @@ import java.util.stream.Collectors;
 
 public class CramRecordReader {
     private final DataSeriesReader<Integer> bitFlagsCodec;
-    private final DataSeriesReader<Byte> compressionBitFlagsCodec;
+    private final DataSeriesReader<Integer> compressionBitFlagsCodec;
     private final DataSeriesReader<Integer> readLengthCodec;
     private final DataSeriesReader<Integer> alignmentStartCodec;
     private final DataSeriesReader<Integer> readGroupCodec;
@@ -54,7 +54,7 @@ public class CramRecordReader {
     private final DataSeriesReader<Integer> paddingCodec;
     private final DataSeriesReader<Integer> deletionLengthCodec;
     private final DataSeriesReader<Integer> mappingScoreCodec;
-    private final DataSeriesReader<Byte> mateBitFlagCodec;
+    private final DataSeriesReader<Integer> mateBitFlagCodec;
     private final DataSeriesReader<Integer> mateReferenceIdCodec;
     private final DataSeriesReader<Integer> mateAlignmentStartCodec;
     private final DataSeriesReader<Integer> insertSizeCodec;
--- a/src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
+++ b/src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
@@ -99,12 +99,12 @@ public class CramCompressionRecord {
 
     public int sliceIndex = 0;
 
-    public byte getMateFlags() {
-        return (byte) (0xFF & mateFlags);
+    public int getMateFlags() {
+        return (0xFF & mateFlags);
     }
 
-    public byte getCompressionFlags() {
-        return (byte) (0xFF & compressionFlags);
+    public int getCompressionFlags() {
+        return (0xFF & compressionFlags);
     }
 
     @SuppressWarnings("SimplifiableIfStatement")
--- a/src/test/java/htsjdk/samtools/cram/encoding/external/ExternalCodecEquivalenceTest.java
+++ b/src/test/java/htsjdk/samtools/cram/encoding/external/ExternalCodecEquivalenceTest.java
@@ -14,6 +14,26 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 
+/**
+ * Motivation for this test: we were incorrectly encoding a few CRAM record
+ * fields using the wrong data type.  This was surprising, since we would
+ * expect catastrophic "frame shift" type failures from this.
+ *
+ * Instead, we noticed that this conflict causes no problems under the following conditions:
+ *
+ * 1. External Byte, External Integer, and External Long Encodings
+ * create identical output streams when values are limited to the
+ * range 0 to 0x7F because the respective underlying encodings
+ * (raw bytes, ITF8, and LTF8 respectively) result in byte-for-byte
+ * identical outputs.
+ *
+ * 2. External Integer and External Long Encodings also create identical
+ * output streams when values are limited to 28 unsigned bits
+ * (0 to 0x0F FF FF FF) because ITF8 ands LTF8 are equivalent over that range.
+ *
+ * Demonstrate that these conditions do in fact create identical streams.
+ *
+ */
 public class ExternalCodecEquivalenceTest extends HtsjdkTest {
     // show that External codecs are equivalent within certain ranges
 
--- a/src/test/java/htsjdk/samtools/cram/encoding/external/ExternalCodecEquivalenceTest.java
+++ b/src/test/java/htsjdk/samtools/cram/encoding/external/ExternalCodecEquivalenceTest.java
@@ -42,8 +42,8 @@ public class ExternalCodecEquivalenceTest extends HtsjdkTest {
 
     @Test(dataProvider = "testPositiveByteLists", dataProviderClass = IOTestCases.class)
     public void byteEquivalenceTest(final List<Byte> values) {
-        codecPairTest(values, this::externalByteCodecWrite,      this::externalIntegerCodecRead);
-        codecPairTest(values, this::externalByteCodecWrite,      this::externalLongCodecRead);
+        codecPairTest(values, this::externalByteCodecWrite,     this::externalIntegerCodecRead);
+        codecPairTest(values, this::externalByteCodecWrite,     this::externalLongCodecRead);
         codecPairTest(values, this::externalIntegerCodecWrite,  this::externalByteCodecRead);
         codecPairTest(values, this::externalIntegerCodecWrite,  this::externalLongCodecRead);
         codecPairTest(values, this::externalLongCodecWrite,     this::externalByteCodecRead);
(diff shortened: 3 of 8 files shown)


src/main/java/htsjdk/samtools/cram/structure/CramCompressionRecord.java
```diff
@@ -99,12 +99,12 @@
 
     public int sliceIndex = 0;
 
-    public byte getMateFlags() {
```
jmthibault79: updated because now codecs are expecting ints

src/test/java/htsjdk/samtools/cram/encoding/external/ExternalCodecEquivalenceTest.java
```diff
@@ -0,0 +1,224 @@
...
+public class ExternalCodecEquivalenceTest extends HtsjdkTest {
+    // show that External codecs are equivalent within certain ranges
+
+    // if all values are 0 <= v <= 0x7F (high bit 7)
+    // ExternalByteCodec, ExternalIntegerCodec, and ExternalLongCodec streams are identical
+    // because raw bytes, ITF8 and LTF8 are identical
+
+    @Test(dataProvider = "testPositiveByteLists", dataProviderClass = IOTestCases.class)
+    public void byteEquivalenceTest(final List<Byte> values) {
+        codecPairTestForByte(values, this::writeByteForByte, this::readIntegerForByte);
```
jmthibault79: I'd like to make this into a DataProvider somehow, but that doesn't appear to be possible with functional interfaces

cmnbroad: Yeah, unfortunately the `FunctionalInterface`->`Object` conversion doesn't usually go well.

src/main/java/htsjdk/samtools/cram/encoding/reader/CramRecordReader.java
```diff
@@ -34,7 +34,7 @@
 
 public class CramRecordReader {
     private final DataSeriesReader<Integer> bitFlagsCodec;
-    private final DataSeriesReader<Byte> compressionBitFlagsCodec;
+    private final DataSeriesReader<Integer> compressionBitFlagsCodec;
```
cmnbroad: Good catch on these. Did you by any chance compare the rest of them to the spec to see if there are any others ?

src/test/java/htsjdk/samtools/cram/encoding/external/ExternalCodecEquivalenceTest.java
```diff
@@ -0,0 +1,224 @@
...
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ExternalCodecEquivalenceTest extends HtsjdkTest {
```
cmnbroad: This class needs a comment explaining the motivation for including it. Its essentially validating a property that several of the codecs happen to share, which is that for small values, the encoded streams are identical. But that property isn't exploited by, or even mentioned in the spec (?); it was only the *previous*, but incorrect encoding of the CF/MF data series incorrect that was accidentally exploiting it. Once this PR goes in, the incorrect implementation will be gone.

So I think its fine to retain the test, but it needs an explanation saying something like that its exploiting that property in order to cross validate the encoding implementations, or something like similar.

src/test/java/htsjdk/samtools/cram/encoding/external/ExternalCodecEquivalenceTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+    private interface ByteWriter {
+        byte[] write(final List<Byte> toWrite);
+    }
+
+    private interface ByteReader {
+        List<Byte> read(final byte[] toRead, final int count);
+    }
+
+    private byte[] writeByteForByte(List<Byte> values) {
```
cmnbroad: It would be super helpful if each of these methods that uses a particular codec had a name that reflected/included the codec its using.

src/test/java/htsjdk/samtools/cram/io/IOTestCases.java
```diff
@@ -140,6 +157,28 @@
         };
     }
 
+    // for testing ITF8 and LTF8 equivalence: only true in the range 0 - (high bit = 28)
```
cmnbroad: Same here, expand a bit on this comment to explain why these tests cases are included.

src/main/java/htsjdk/samtools/cram/encoding/reader/CramRecordReader.java
jmthibault79: Thx - the rest of the spec matches.  Though QS and BA are a little weird because they're sometimes byte and sometimes byte array.  (See 10.6 vs 10.7)

src/test/java/htsjdk/samtools/cram/encoding/external/ExternalCodecEquivalenceTest.java
```diff
@@ -0,0 +1,176 @@
...
+    // ExternalIntegerCodec, and ExternalLongCodec streams are identical
+    // because ITF8 and LTF8 are identical
+
+    @Test(dataProvider = "testUint28Lists", dataProviderClass = IOTestCases.class)
+    public void intEquivalenceTest(final List<Integer> values) {
+        codecPairTest(values, this::externalIntegerCodecWrite,  this::externalLongCodecRead);
+        codecPairTest(values, this::externalLongCodecWrite,     this::externalIntegerCodecRead);
+    }
+
+    private <T extends Number> void codecPairTest(final List<T> values, Writer<T> writer, Reader reader) {
```
lbergelson: It's  a bit confusing to define custom `Reader/Writer` interfaces here since there a really common `java.io` interfaces already with those names.  It's not a big deal either way though.

src/test/java/htsjdk/samtools/cram/io/IOTestCases.java
```diff
@@ -72,6 +73,22 @@
...
 
+    @DataProvider(name = "testPositiveByteLists")
+    public static Object[][] testPositiveByteValues() {
+        final List<Byte> positiveByteTests = IOTestCases.byteTests()
+                .stream()
+                .filter(b -> b >= 0)
+                .collect(Collectors.toList());
+
+        final List<Byte> shuffled = new ArrayList<>(positiveByteTests);
+        Collections.shuffle(shuffled);
```
lbergelson: Why shuffle these?  If you need to shuffle them can we use a set random seed?  `TestUtil.RANDOM_SEED` is a good option.

src/test/java/htsjdk/samtools/cram/encoding/external/ExternalCodecEquivalenceTest.java
```diff
@@ -0,0 +1,176 @@
...
+        return retval;
+    }
+
+    // TODO: move these into the Codec classes?
+
+    private interface ReadCodecConstructor<T> {
+        CRAMCodec<T> reader(final ByteArrayInputStream is);
+    }
+
+    private CRAMCodec<Byte> externalByteCodecReaderConstructor(final ByteArrayInputStream is) {
```
lbergelson: These methods seem like good candidates for inlining?  Did you want to add an interface method to access these instead somehow? The todo makes it sound that way.

src/test/java/htsjdk/samtools/cram/encoding/external/ExternalCodecEquivalenceTest.java
```diff
@@ -0,0 +1,176 @@
...
+            os.flush();
+            written = os.toByteArray();
+        } catch (final IOException e) {
+            throw new RuntimeIOException(e);
+        }
+        return written;
+    }
+
+    private <T extends Number> byte[] externalIntegerCodecWrite(List<T> values) {
+        byte[] written;
```
lbergelson: These look refactorable to reduce the size of the repeated code, but it's always hard to do that with numeric types than I think.

src/test/java/htsjdk/samtools/cram/encoding/external/ExternalCodecEquivalenceTest.java
jmthibault79: true, will fix

src/test/java/htsjdk/samtools/cram/encoding/external/ExternalCodecEquivalenceTest.java
jmthibault79: Yeah, I tried to do so, but could never get the casting/conversion right

src/test/java/htsjdk/samtools/cram/encoding/external/ExternalCodecEquivalenceTest.java
jmthibault79: like this?  
`codecPairTest(values, this::externalIntegerCodecWrite,  ((bytes, len) -> read(bytes, len, this::externalByteCodecReaderConstructor)));`

lbergelson: Oh, bleh, I think I misread where these were used.

src/test/java/htsjdk/samtools/cram/io/IOTestCases.java
jmthibault79: results may be order-dependent, so I'm shuffling in order to give a good balance there.

I'll update to use the seed.

src/test/java/htsjdk/samtools/cram/encoding/external/ExternalCodecEquivalenceTest.java
jmthibault79: That looks less readable to me, and I don't think it matters much for a small test like this

lbergelson: No.. these are used only in the 1 method above them:
```
    private List<Long> externalLongCodecRead(byte[] written, final int length) {
        return read(written, length, is -> new ExternalLongCodec(is, null));
    }
```

It would stay `codecPairTest(values, this::externalIntegerCodecWrite,  this::externalByteCodecRead); `

lbergelson: A lot of these methods can be made static too while we're here.

lbergelson: If you think it's less readable then leave it how it is.

jmthibault79: oh I see - yes I'll make these inline.

What is the benefit of static for test classes?  I have not generally been doing that.

PR #1320, opened 2019-03-08
Adding tests revealing 0-length interval nonsense

lbergelson:
This PR adds failing tests that test intersection/abutment using 0-length intervals.

Many are failing...

I've added an ascii diagram in the test to show my belief of what the various intervals represent.

The equivalent methods in `Interval` and `Locatable` have slightly different failure modes.

We should decide what the right answer is, and then fix the various methods to implement those.

We should also check related methods like `encloses` and `overlap` in `CoordMath`.

I'm sick of this issue and want it to be resolved so we never have to think about it again.

### DIFF
--- a/src/test/java/htsjdk/samtools/util/IntervalTest.java
+++ b/src/test/java/htsjdk/samtools/util/IntervalTest.java
@@ -1,10 +1,14 @@
 package htsjdk.samtools.util;
 
+import com.google.common.collect.Streams;
 import htsjdk.HtsjdkTest;
 import htsjdk.tribble.annotation.Strand;
 import org.testng.Assert;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
+
+import java.util.Arrays;
+import java.util.stream.Stream;
 
 public class IntervalTest extends HtsjdkTest {
     @DataProvider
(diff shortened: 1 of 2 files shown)


src/test/java/htsjdk/samtools/util/IntervalTest.java
```diff
@@ -19,4 +23,86 @@ public void testGetStrand(boolean isNegativeStrand){
...
+                {chr1_10_20, zeroLength_chr1_22_21, false, false},
+                {chr1_10_21, zeroLength_chr1_22_21, false, false},
+                {chr1_10_22, zeroLength_chr1_22_21, false, true},
+                {chr1_10_23, zeroLength_chr1_22_21, true, false},
+                {chr1_20_30, zeroLength_chr1_22_21, true, false},
+                {chr1_21_30, zeroLength_chr1_22_21, true, false},
+                {chr1_22_30, zeroLength_chr1_22_21, true, false},
+                {chr1_23_30, zeroLength_chr1_22_21, false, true},
+                {chr1_24_30, zeroLength_chr1_22_21, false, false},
+                {zeroLength_chr1_22_21, zeroLength_chr1_22_21, true, false},
```
yfarjoun: hmmm. to answer his question we need a definition of "intersects" and "abuts"

Intersects: There exists at least one base in the reference that is present in both intervals.
Abuts: Two intervals that do not intersect (c.f.) , however, the union of bases they cover can be written as one interval. 

According to this definition, two empty intervals in the same position do not intersect, but they do abut.

yfarjoun: we have both intersects and overlaps? huh?

yfarjoun: it's the same function, implemented twice!

lbergelson: We have more in other places I'm sure.   Because of  fear or weird details like this we never merged the implementations of a lot of these.

PR #1322, opened 2019-03-12
Fixed issue #1111 in broadinstitute/picard

Sonik-zirael:
### Description

Here you can see our fix for issue https://github.com/broadinstitute/picard/issues/1111. We decided to create an index file if it does not exist. We wrote several tests in htsjdk to check if index file was created correctly. But now there is one test (testIsQueryable in VCFFileReaderTest) which has not been passed successfully. This test is for hasIndex() (Boolean) function. And now this function always returns true. Maybe it would be better to rewrite this test. 

### Checklist

- [ ] Code compiles correctly


lbergelson:
@Sonik-zirael Hello and thank you for your interest and work in creating this PR.  I especially appreciate that you've added tests!  

I'm not sure that it's the right solution for this problem though.  The decision to not automatically generate the index in htsjdk when loading files was a deliberate one.  Usually an index is used to provide fast access to a file.  Regenerating it incurs the cost of reading the entire file which negates the point of having the index.  So if htsjdk is told to load a file with an index we deliberately fail when it's missing to alert the user that there is a problem.  It's the responsibility of downstream tool authors to decide when they require the index and when they don't.  In this case it seems like picard is accidentally requiring an index when none is necessary, and we can fix that much more easily in picard.  

I think that this [line in picard](https://github.com/broadinstitute/picard/blob/1531278515aaf39c9049b5792ae569b92326dd90/src/main/java/picard/vcf/RenameSampleInVcf.java#L118) is the problem:

```
final VCFFileReader in = new VCFFileReader(INPUT);
```
changing it to
```
final VCFFileReader in = new VCFFileReader(INPUT, false);
```
Should allow the tool to run without an index.

lbergelson:
I'm going to close this since I believe this issue should be fixed in picard instead.  Feel to reopen if you have further comments.

PR #1328, opened 2019-03-15
Make VariantContextWriterBuilder warn when indexing-on-the-fly is enabled for stream output

jamesemery:
This is the first stage in resolving https://github.com/broadinstitute/gatk/issues/5779. Currently in when we provide a stream (or any non-regular file to resolve) to the `VariantContextWriterBuilder` we set `.setOutputFileType(OutputType.VCF)`. This sidesteps the code in the switch statement designed to this case and throw an exception if indexing on the fly is enabled. It turns out to be a pain point for the users because indexing on the fly is enabled by default. I propose that we push the responsibility of deciding if the output is a stream up to gatk, and then make the stream code in `VariantContextWriterBuilder` consistent with the BAM writer, which only warns and disables indexing if it suspects the file might be a stream.

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
@@ -190,7 +190,7 @@ public class VariantContextWriterBuilder {
     /**
      * Set the output file type for the next <code>VariantContextWriter</code> created by this builder.
      *
-     * @param outType the type of file or stream the <code>VariantContextWriter</code> will write to
+     * @param outType the type of file the <code>VariantContextWriter</code> will write to
      * @return this <code>VariantContextWriterBuilder</code>
      */
     public VariantContextWriterBuilder setOutputFileType(final OutputType outType) {
--- a/src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
@@ -194,6 +194,8 @@ public class VariantContextWriterBuilder {
      * @return this <code>VariantContextWriterBuilder</code>
      */
     public VariantContextWriterBuilder setOutputFileType(final OutputType outType) {
+        if (!FILE_TYPES.contains(outType))
+            throw new IllegalArgumentException("Must choose a file type, not other output types.");
 
         if (this.outPath == null || this.outStream != null)
             throw new IllegalArgumentException("Cannot set a file type if the output is not to a file.");
--- a/src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
+++ b/src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
@@ -27,7 +27,11 @@ package htsjdk.variant.variantcontext.writer;
 
 import htsjdk.samtools.Defaults;
 import htsjdk.samtools.SAMSequenceDictionary;
-import htsjdk.samtools.util.*;
+import htsjdk.samtools.util.BlockCompressedOutputStream;
+import htsjdk.samtools.util.IOUtil;
+import htsjdk.samtools.util.Md5CalculatingOutputStream;
+import htsjdk.samtools.util.Log;
+import htsjdk.samtools.util.RuntimeIOException;
 import htsjdk.tribble.index.IndexCreator;
 import htsjdk.tribble.index.tabix.TabixFormat;
 import htsjdk.tribble.index.tabix.TabixIndexCreator;
@@ -104,7 +108,7 @@ public class VariantContextWriterBuilder {
     public static final EnumSet<Options> DEFAULT_OPTIONS = EnumSet.of(Options.INDEX_ON_THE_FLY);
     public static final EnumSet<Options> NO_OPTIONS = EnumSet.noneOf(Options.class);
     private static final OpenOption[] EMPTY_OPEN_OPTION_ARRAY = new OpenOption[0];
-    private final static Log log = Log.getInstance(VariantContextWriter.class);
+    private static final Log log = Log.getInstance(VariantContextWriter.class);
 
     public enum OutputType {
         UNSPECIFIED,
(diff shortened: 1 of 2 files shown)


jamesemery:
@cmnbroad

src/test/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilderUnitTest.java
```diff
@@ -347,6 +347,25 @@ public void testIndexingOnTheFly() {
     }
   }
 
+  @Test
+  public void testIndexingOnTheFlyForPathStream() throws IOException {
+    final VariantContextWriterBuilder builder = new VariantContextWriterBuilder()
```
pshapiro4broad: Indent should be 4 spaces, not 2. I realize the rest of the file is using 2, but IMO we should move to a common convention for all files, starting with new/updated code.

src/test/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilderUnitTest.java
```diff
@@ -347,6 +347,25 @@ public void testIndexingOnTheFly() {
...
+            .setReferenceDictionary(dictionary)
+            .setOption(Options.INDEX_ON_THE_FLY);
+
+    try (FileSystem fs = Jimfs.newFileSystem("test", Configuration.unix())) {
+      final Path vcfPath = IOUtil.getPath("/dev/null");
+      final Path vcfIdxPath = Tribble.indexPath(vcfPath);
+      // We explicitly setOutputFileType here to mimic gatk behavior where it pretends files that have no extension
```
pshapiro4broad: Incomplete sentence, or extra `that`?

src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
```diff
@@ -107,6 +104,7 @@
     public static final EnumSet<Options> DEFAULT_OPTIONS = EnumSet.of(Options.INDEX_ON_THE_FLY);
     public static final EnumSet<Options> NO_OPTIONS = EnumSet.noneOf(Options.class);
     private static final OpenOption[] EMPTY_OPEN_OPTION_ARRAY = new OpenOption[0];
+    private final static Log log = Log.getInstance(VariantContextWriter.class);
```
pshapiro4broad: incorrect modifier order
```suggestion
    private static final Log log = Log.getInstance(VariantContextWriter.class);
```

src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
```diff
@@ -476,14 +472,18 @@ else if (STREAM_TYPES.contains(this.outType))
                 writer = createBCFWriter(outPath, outStreamFromFile);
                 break;
             case VCF_STREAM:
-                if (options.contains(Options.INDEX_ON_THE_FLY))
-                    throw new IllegalArgumentException("VCF index creation not supported for stream output.");
+                if (options.contains(Options.INDEX_ON_THE_FLY)) {
```
pshapiro4broad: to avoid code duplication this check can be done on line 433 above, after checking `STREAM_TYPES.contains(this.outType)`

src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
```diff
@@ -27,10 +27,7 @@
 
 import htsjdk.samtools.Defaults;
 import htsjdk.samtools.SAMSequenceDictionary;
-import htsjdk.samtools.util.BlockCompressedOutputStream;
-import htsjdk.samtools.util.IOUtil;
-import htsjdk.samtools.util.Md5CalculatingOutputStream;
-import htsjdk.samtools.util.RuntimeIOException;
+import htsjdk.samtools.util.*;
```
pshapiro4broad: IMO it's better to avoid wildcard imports as it makes it harder to see what the code is doing when you're not using an IDE. You can change the IDE settings so it doesn't create wildcard imports.

src/test/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilderUnitTest.java
```diff
@@ -347,6 +347,25 @@ public void testIndexingOnTheFly() {
...
+      try (final VariantContextWriter writer = builder.setOutputPath(vcfPath).setOutputFileType(OutputType.VCF_STREAM).build()) {
```
yfarjoun: will this also work if you do not set the output type to a stream?

src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
yfarjoun: can you use htsjdk.samtools.util.IOUtil#isRegularPath(java.io.File)?

src/test/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilderUnitTest.java
jamesemery: I can reindent the file pretty easily if you would like? It seems orthogonal to this branch however.

src/main/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilder.java
jamesemery: The issue with this approach is that the user of this builder can set the output path directly. If this happens then there is no way to extract what or where the stream points to and consequently we can't reliably determine the output value here. The plan was to just to call htsjdk.samtools.util.IOUtil#isRegularPath(java.io.File) in gatk to sidestep this issue. Either way disabling index-on-the-fly would save users/tool authors the trouble of having their streams crash if outputting to an unindexable place like "/dev/null"

src/test/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilderUnitTest.java
jamesemery: No it won't. In that case it will end up in the same place where we are as of right now, where an output stream location will end up "UNSPECIFIED" and throw an exception unless the user explicitly defines what type of output it is.

src/test/java/htsjdk/variant/variantcontext/writer/VariantContextWriterBuilderUnitTest.java
pshapiro4broad: I would just fix the lines that you're adding/modifying.

PR #1276, opened 2019-02-12
CRAM indexing tests and associated fixes and updates

jmthibault79:
### Description

Adding better test coverage for CRAM indexing, which inspired:
* a CRAIQuery class with the necessary fields from CRAIEntry for querying
* AlignmentSpan immutability and distinguishing between mapped and unmapped read counts
* Recording mapped, unmapped, and unplaced read counts per slice
* requiring Coordinate-Sorted for CRAM BAI generation

### Checklist

- [x] Code compiles correctly
- [x] New tests covering changes and new functionality
- [x] All tests passing
- [x] Extended the README / documentation, if necessary
- [ ] Is not backward compatible (breaks binary or source compatibility)

PR #1326, opened 2019-03-14
Test ContainerIO.calculateSliceOffsetsAndSizes() and fix the slice size calculation

jmthibault79:
### Description

Problem:
calculateSliceOffsetsAndSizes() was calculating Slice.size incorrectly.

Impact: 
Slices created via ContainerIO.readContainer() would produce incorrect sliceByteSize values for their CRAI entries.

### Checklist

- [x] Code compiles correctly
- [x] New tests covering changes and new functionality
- [x] All tests passing
- [x] Extended the README / documentation, if necessary
- [ ] Is not backward compatible (breaks binary or source compatibility)

PR #1306, opened 2019-02-26
DO NOT MERGE: CRAM improve test coverage and fix a few bugs

jmthibault79:
Add tests to improve test coverage and find bugs.

### Checklist

- [x] Code compiles correctly
- [x] New tests covering changes and new functionality
- [x] All tests passing
- [x] Extended the README / documentation, if necessary
- [ ] Is not backward compatible (breaks binary or source compatibility)

PR #1329, opened 2019-03-15
Follow-up to #1276: Remove multiple versions of Slice/Container getCRAIEntries()

jmthibault79:
- Encapsulate Container.byteOffset and distribute to Slices on set
- enforce prerequisites to index entry creation

### Description

### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/BAMIndexMetaData.java
+++ b/src/main/java/htsjdk/samtools/BAMIndexMetaData.java
@@ -159,7 +159,7 @@ public class BAMIndexMetaData {
             unAlignedRecords += slice.unmappedReadsCount;
         }
 
-        final long start = slice.offset;
+        final long start = slice.byteOffsetFromContainer;
 
         if (BlockCompressedFilePointerUtil.compare(start, firstOffset) < 1 || firstOffset == -1) {
             this.firstOffset = start;
--- a/src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
+++ b/src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
@@ -134,14 +134,12 @@ public class CRAMBAIIndexer {
 
         int sliceIndex = 0;
         for (final Slice slice : container.slices) {
-            slice.containerOffset = container.offset;
             slice.index = sliceIndex++;
             if (slice.getReferenceContext().isMultiRef()) {
                 final Map<ReferenceContext, AlignmentSpan> spanMap = container.getSpans(validationStringency);
 
                 // TODO why are we updating the original slice here?
 
-                slice.containerOffset = container.offset;
                 slice.index = sliceIndex++;
 
                 /**
@@ -151,8 +149,8 @@ public class CRAMBAIIndexer {
                 for (final ReferenceContext refContext : new TreeSet<>(spanMap.keySet())) {
                     final AlignmentSpan span = spanMap.get(refContext);
                     final Slice fakeSlice = new Slice(refContext);
-                    fakeSlice.containerOffset = slice.containerOffset;
-                    fakeSlice.offset = slice.offset;
+                    fakeSlice.containerByteOffset = slice.containerByteOffset;
+                    fakeSlice.byteOffsetFromContainer = slice.byteOffsetFromContainer;
                     fakeSlice.index = slice.index;
 
                     fakeSlice.alignmentStart = span.getStart();
@@ -165,8 +163,8 @@ public class CRAMBAIIndexer {
 
                 if (unmappedSpan != null) {
                     final Slice fakeSlice = new Slice(ReferenceContext.UNMAPPED_UNPLACED_CONTEXT);
-                    fakeSlice.containerOffset = slice.containerOffset;
-                    fakeSlice.offset = slice.offset;
+                    fakeSlice.containerByteOffset = slice.containerByteOffset;
+                    fakeSlice.byteOffsetFromContainer = slice.byteOffsetFromContainer;
                     fakeSlice.index = slice.index;
 
                     fakeSlice.alignmentStart = SAMRecord.NO_ALIGNMENT_START;
@@ -273,7 +271,7 @@ public class CRAMBAIIndexer {
                     break;
                 }
 
-                container.offset = offset;
+                container.setByteOffset(offset);
 
                 indexer.processContainer(container, validationStringency);
 
@@ -368,7 +366,7 @@ public class CRAMBAIIndexer {
          * Record any index information for a given CRAM slice
          *
          * Reads these Slice fields:
-         * sequenceId, alignmentStart, alignmentSpan, containerOffset, index
+         * sequenceId, alignmentStart, alignmentSpan, containerByteOffset, index
          *
          * @param slice CRAM slice, single ref only.
          */
@@ -411,8 +409,8 @@ public class CRAMBAIIndexer {
 
             // process chunks
 
-            final long chunkStart = (slice.containerOffset << 16) | slice.index;
-            final long chunkEnd = ((slice.containerOffset << 16) | slice.index) + 1;
+            final long chunkStart = (slice.containerByteOffset << 16) | slice.index;
+            final long chunkEnd = ((slice.containerByteOffset << 16) | slice.index) + 1;
 
             final Chunk newChunk = new Chunk(chunkStart, chunkEnd);
 
--- a/src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
+++ b/src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
@@ -74,6 +74,10 @@ import java.util.TreeSet;
  * and proceeds by processing {@link CRAIEntry} elements obtained from
  * {@link CRAMCRAIIndexer#readIndex(InputStream)}.  {@link CRAMBAIIndexer#processAsSingleReferenceSlice(Slice)}
  * is called on each {@link CRAIEntry} and {@link CRAMBAIIndexer#finish()} is called at the end.
+ *
+ * NOTE: a third pattern of building a BAI from a CRAM file is also supported by this class,
+ * but it is unused.  This would be accomplished via {@link #createIndex(SeekableStream, File, Log, ValidationStringency)}.
+ *
  */
 public class CRAMBAIIndexer {
 
--- a/src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
+++ b/src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
@@ -53,7 +53,6 @@ import htsjdk.samtools.util.Log;
 import htsjdk.samtools.util.ProgressLogger;
 
 import java.io.File;
-import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.Arrays;
@@ -133,7 +132,7 @@ public class CRAMBAIIndexer {
         }
 
         int sliceIndex = 0;
-        for (final Slice slice : container.slices) {
+        for (final Slice slice : container.getSlices()) {
             slice.index = sliceIndex++;
             if (slice.getReferenceContext().isMultiRef()) {
                 final Map<ReferenceContext, AlignmentSpan> spanMap = container.getSpans(validationStringency);
@@ -260,40 +259,31 @@ public class CRAMBAIIndexer {
         }
         final CRAMBAIIndexer indexer = new CRAMBAIIndexer(output, cramHeader.getSamFileHeader());
 
-        int totalRecords = 0;
         Container container = null;
         ProgressLogger progressLogger = new ProgressLogger(log, 1, "indexed", "slices");
         do {
-            try {
-                final long offset = stream.position();
-                container = ContainerIO.readContainer(cramHeader.getVersion(), stream);
-                if (container == null || container.isEOF()) {
-                    break;
+            container = ContainerIO.readContainer(cramHeader.getVersion(), stream);
+            if (container == null || container.isEOF()) {
+                break;
+            }
+
+            indexer.processContainer(container, validationStringency);
+
+            if (null != log) {
+                String sequenceName;
+                final ReferenceContext containerContext = container.getReferenceContext();
+                switch (containerContext.getType()) {
+                    case UNMAPPED_UNPLACED_TYPE:
+                        sequenceName = "?";
+                        break;
+                    case MULTIPLE_REFERENCE_TYPE:
+                        sequenceName = "???";
+                        break;
+                    default:
+                        sequenceName = cramHeader.getSamFileHeader().getSequence(containerContext.getSequenceId()).getSequenceName();
+                        break;
                 }
-
-                container.setByteOffset(offset);
-
-                indexer.processContainer(container, validationStringency);
-
-                if (null != log) {
-                    String sequenceName;
-                    final ReferenceContext containerContext = container.getReferenceContext();
-                    switch (containerContext.getType()) {
-                        case UNMAPPED_UNPLACED_TYPE:
-                            sequenceName = "?";
-                            break;
-                        case MULTIPLE_REFERENCE_TYPE:
-                            sequenceName = "???";
-                            break;
-                        default:
-                            sequenceName = cramHeader.getSamFileHeader().getSequence(containerContext.getSequenceId()).getSequenceName();
-                            break;
-                    }
-                    progressLogger.record(sequenceName, container.alignmentStart);
-                }
-
-            } catch (final IOException e) {
-                throw new RuntimeException("Failed to read cram container", e);
+                progressLogger.record(sequenceName, container.alignmentStart);
             }
 
         } while (!container.isEOF());
--- a/src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
+++ b/src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
@@ -38,7 +38,8 @@
  */
 package htsjdk.samtools;
 
-import htsjdk.samtools.cram.build.ContainerParser;
+import htsjdk.samtools.cram.CRAIEntry;
+import htsjdk.samtools.cram.CRAIIndex;
 import htsjdk.samtools.cram.build.CramIO;
 import htsjdk.samtools.cram.ref.ReferenceContext;
 import htsjdk.samtools.cram.structure.AlignmentSpan;
@@ -53,6 +54,7 @@ import htsjdk.samtools.util.ProgressLogger;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.Arrays;
 import java.util.List;
@@ -61,11 +63,18 @@ import java.util.TreeSet;
 
 /**
  * Class for both constructing BAM index content and writing it out.
+ *
  * There are two usage patterns:
- * 1) Building a bam index from an existing cram file
- * 2) Building a bam index while building the cram file
- * In both cases, processAlignment is called for each cram slice and
- * finish() is called at the end.
+ * 1) Building a bam index (BAI) while building the CRAM file
+ * 2) Building a bam index (BAI) from an existing CRAI file
+ *
+ * 1) is driven by {@link CRAMContainerStreamWriter} and proceeds by calling {@link CRAMBAIIndexer#processContainer}
+ * after each {@link Container} is built, and {@link CRAMBAIIndexer#finish()} is called at the end.
+ *
+ * 2) is driven by {@link CRAIIndex#openCraiFileAsBaiStream(InputStream, SAMSequenceDictionary)}
+ * and proceeds by processing {@link CRAIEntry} elements obtained from
+ * {@link CRAMCRAIIndexer#readIndex(InputStream)}.  {@link CRAMBAIIndexer#processAsSingleReferenceSlice(Slice)}
+ * is called on each {@link CRAIEntry} and {@link CRAMBAIIndexer#finish()} is called at the end.
  */
 public class CRAMBAIIndexer {
 
@@ -78,7 +87,7 @@ public class CRAMBAIIndexer {
     private int currentReference = 0;
 
     // content is built up from the input bam file using this
-    private final BAMIndexBuilder indexBuilder;
+    private final CRAMBAIIndexBuilder indexBuilder;
 
     /**
      * Create a CRAM indexer that writes BAI to a file.
@@ -91,7 +100,7 @@ public class CRAMBAIIndexer {
             throw new SAMException("CRAM file must be coordinate-sorted for indexing.");
         }
         numReferences = fileHeader.getSequenceDictionary().size();
-        indexBuilder = new BAMIndexBuilder(fileHeader);
+        indexBuilder = new CRAMBAIIndexBuilder(fileHeader);
         outputWriter = new BinaryBAMIndexWriter(numReferences, output);
     }
 
@@ -106,7 +115,7 @@ public class CRAMBAIIndexer {
             throw new SAMException("CRAM file must be coordinate-sorted for indexing.");
         }
         numReferences = fileHeader.getSequenceDictionary().size();
-        indexBuilder = new BAMIndexBuilder(fileHeader);
+        indexBuilder = new CRAMBAIIndexBuilder(fileHeader);
         outputWriter = new BinaryBAMIndexWriter(numReferences, output);
     }
 
@@ -183,6 +192,9 @@ public class CRAMBAIIndexer {
      * @throws htsjdk.samtools.SAMException if slice refers to multiple reference sequences.
      */
     public void processAsSingleReferenceSlice(final Slice slice) {
+        // validate that this Slice is ready to be indexed
+        slice.baiIndexInitializationCheck();
+
         final ReferenceContext sliceContext = slice.getReferenceContext();
         if (sliceContext.isMultiRef()) {
             throw new SAMException("Expecting a single reference or unmapped slice.");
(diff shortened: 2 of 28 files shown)


src/main/java/htsjdk/samtools/BAMIndexMetaData.java
```diff
@@ -159,7 +159,7 @@ void recordMetaData(final Slice slice) {
             unAlignedRecords += slice.unmappedReadsCount;
         }
 
-        final long start = slice.offset;
```
jmthibault79: renamed the Slice indexing parameters to make their function clearer.

src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
```diff
@@ -125,14 +124,12 @@ public void processContainer(final Container container, final ValidationStringen
 
         int sliceIndex = 0;
         for (final Slice slice : container.slices) {
-            slice.containerOffset = container.offset;
```
jmthibault79: this is now set elsewhere

src/main/java/htsjdk/samtools/cram/CRAIIndex.java
```diff
@@ -59,7 +59,7 @@ public void writeIndex(final OutputStream os) {
      * @param container the container to index
      */
     public void processContainer(final Container container) {
-        addEntries(container.getCRAIEntriesSplittingMultiRef());
+        addEntries(container.getCRAIEntries());
```
jmthibault79: we always split multiref now

src/main/java/htsjdk/samtools/cram/build/ContainerFactory.java
```diff
@@ -58,8 +58,7 @@ public Container buildContainer(final List<CramCompressionRecord> records) {
             slices.add(slice);
         }
 
-        final Container container = Container.initializeFromSlices(slices);
-        container.compressionHeader = compressionHeader;
+        final Container container = Container.initializeFromSlices(slices, compressionHeader);
```
jmthibault79: We need the compressionHeader when we call `getCraiEntries()` so requiring it's here helps ensure it's valid

src/main/java/htsjdk/samtools/cram/structure/Container.java
```diff
@@ -137,34 +156,16 @@ else if (sliceRefContexts.size() > 1) {
...
     public List<CRAIEntry> getCRAIEntries() {
-        return Arrays.stream(slices)
-                .map(Slice::getCRAIEntry)
-                .collect(Collectors.toList());
-    }
-
-    /**
-     * Retrieve the list of CRAI Index entries corresponding to this Container.
-     *
-     * TODO: investigate why we sometimes split multi-ref Slices
```
jmthibault79: now we always split

src/main/java/htsjdk/samtools/cram/structure/Container.java
```diff
@@ -137,34 +156,16 @@ else if (sliceRefContexts.size() > 1) {
...
-     * into different entries and sometimes do not
-     *
-     * TODO: clearly identify and enforce preconditions, e.g.
```
jmthibault79: precondition enforcement is in `craiIndexInitializationCheck()`

src/main/java/htsjdk/samtools/cram/structure/ContainerIO.java
```diff
@@ -95,18 +95,20 @@ private static Container readContainer(final int major, final InputStream inputS
     }
 
     private static void calculateSliceOffsetsAndSizes(final Container container) {
```
jmthibault79: Make sure to carefully coordinate this change with #1326

src/main/java/htsjdk/samtools/cram/structure/Slice.java
```diff
@@ -337,51 +372,42 @@ public CramRecordReader createCramRecordReader(final CompressionHeader header,
...
+        craiIndexInitializationCheck();
+
         if (referenceContext.isMultiRef()) {
             final Map<ReferenceContext, AlignmentSpan> spans = getMultiRefAlignmentSpans(compressionHeader, ValidationStringency.DEFAULT_STRINGENCY);
 
             return spans.entrySet().stream()
                     .map(e -> new CRAIEntry(e.getKey().getSerializableId(),
                             e.getValue().getStart(),
                             e.getValue().getSpan(),
-                            containerStartByteOffset,
```
jmthibault79: can trust the Slice now instead of relying on the Container, for these 2 values

src/main/java/htsjdk/samtools/cram/structure/Slice.java
```diff
@@ -337,51 +372,42 @@ public CramRecordReader createCramRecordReader(final CompressionHeader header,
...
-                            landmarks[index],
-                            size))
```
jmthibault79: same member variable.  just renamed.

src/main/java/htsjdk/samtools/cram/structure/Slice.java
```diff
@@ -337,51 +372,42 @@ public CramRecordReader createCramRecordReader(final CompressionHeader header,
...
                     .collect(Collectors.toList());
         } else {
             // single ref or unmapped
             final int sequenceId = referenceContext.getSerializableId();
-            return Collections.singletonList(new CRAIEntry(sequenceId, alignmentStart, alignmentSpan, containerStartByteOffset, offset, size));
+            return Collections.singletonList(new CRAIEntry(
+                    sequenceId,
+                    alignmentStart,
+                    alignmentSpan,
+                    containerByteOffset,
```
jmthibault79: `containerByteOffset` newly uses the Slice member variable.
`byteOffsetFromContainer` and `byteSize` are continuing uses of the Slice member variables, renamed.

jmthibault79:
This is going to conflict with #1326 so let's hold off on review for now, and wait for a rebase.

src/main/java/htsjdk/samtools/cram/structure/Container.java
```diff
@@ -67,11 +67,28 @@
...
+     * @param byteOffset the byte location in the stream where this Container begins
      */
-    public long offset;
+    public void setByteOffset(final long byteOffset) {
+        this.byteOffset = byteOffset;
+        if (slices != null) {
+            for (final Slice slice : slices) {
+                slice.containerByteOffset = byteOffset;
+            }
+        }
```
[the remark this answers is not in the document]
cmnbroad: @jmthibault79 It seems like it should be possible to reduce the number of call sites that need to update these offsets. Every call to this method (except for the one in ContainerStreamWriter, when a new container is being created from records) is immediately preceded by a call (either directly or indirectly), to ContainerIO.readContainer. It would be much  simpler to make ContainerIO responsible for all of those cases, even if it means passing the offset in (although I think it could often derive the offset from the stream position). Would that work ? I think it would reduce the complexity around understanding who/when the slice offsets get propagated, etc.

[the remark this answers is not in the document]
jmthibault79: I do think this is possible, and I have started an attempt at this.  I'll work on that some more.

Good point about propagation.  I will have to think more about that.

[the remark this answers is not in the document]
jmthibault79: Addressed both of these in a new commit.

src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
jmthibault79: `Container.setSlices()` sets this value.  If the slices exist prior to the setting of this value, then `Container.setByteOffset()` sets the slices' values.

src/main/java/htsjdk/samtools/CRAMCRAIIndexer.java
```diff
@@ -94,27 +93,19 @@ public void finish() {
...
+        final CramHeader cramHeader = CramIO.readCramHeader(cramStream);
+        final CRAMCRAIIndexer indexer = new CRAMCRAIIndexer(craiStream, cramHeader.getSamFileHeader());
+        final Version cramVersion = cramHeader.getVersion();
 
-            indexer.finish();
-        }
-        catch (IOException e) {
-            throw new RuntimeIOException("Error writing CRAI index to output stream");
+        // get the first container
+        Container container = ContainerIO.readContainer(cramVersion, cramStream);
```
jmthibault79: offset is set inside `readContainer()` by querying the stream

src/main/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -400,7 +400,7 @@ public SAMRecordIterator assertSorted(final SortOrder sortOrder) {
             seekableStream.seek(0);
             newIterator = new CRAMIterator(seekableStream, referenceSource, validationStringency);
             seekableStream.seek(startOfLastLinearBin >>> 16);
-            final Container container = ContainerIO.readContainerHeader(newIterator.getCramHeader().getVersion().major, seekableStream);
+            final Container container = ContainerHeaderIO.readContainerHeader(newIterator.getCramHeader().getVersion().major, seekableStream);
```
jmthibault79: this was driving me nuts

jmthibault79: TODO obviously: move this to a new `ContainerHeader` class.  Just not today.

src/main/java/htsjdk/samtools/cram/build/ContainerFactory.java
```diff
@@ -58,8 +64,7 @@ public Container buildContainer(final List<CramCompressionRecord> records) {
             slices.add(slice);
         }
 
-        final Container container = Container.initializeFromSlices(slices);
-        container.compressionHeader = compressionHeader;
```
jmthibault79: I don't recall why I moved this inside the call to `initializeFromSlices()`.  I can take it back out - they seem pretty equal to me.

src/main/java/htsjdk/samtools/cram/build/CramContainerHeaderIterator.java
```diff
@@ -22,8 +22,19 @@ public CramContainerHeaderIterator(final InputStream inputStream) {
...
+     * Consume the entirety of the next container from the stream,
+     * but retain only the header.
+     *
+     * @see CramContainerIterator#containerFromStream(Version, CountingInputStream)
+     *
+     * @param cramVersion the expected CRAM version of the stream
+     * @param countingStream the {@link CountingInputStream} to read from
+     * @return The next Container's header from the stream, returned as a Container.
+     */
+    @Override
```
jmthibault79: This tripped me up!

Hopefully this explicit Override and Javadoc will save some future grief.

src/main/java/htsjdk/samtools/cram/build/CramContainerIterator.java
```diff
@@ -17,20 +17,14 @@
     private CountingInputStream countingInputStream;
     private Container nextContainer;
     private boolean eof = false;
-    private long offset = 0;
```
jmthibault79: containers can now take care of this.  don't need it here.

src/main/java/htsjdk/samtools/cram/build/CramIO.java
```diff
@@ -293,36 +297,39 @@ private static SAMFileHeader readSAMFileHeader(final Version version, InputStrea
             } else {
                 /*
                  * pending issue: container.containerByteSize inputStream 2 bytes shorter
```
jmthibault79: I have not investigated this

src/main/java/htsjdk/samtools/cram/build/CramIO.java
```diff
@@ -293,36 +297,39 @@ private static SAMFileHeader readSAMFileHeader(final Version version, InputStrea
...
-				 * then needed in the v21 test cram files.
-				 */
+                 * than needed in the v21 test cram files.
+                 */
                 block = Block.read(version.major, inputStream);
             }
         }
 
-        inputStream = new ByteArrayInputStream(block.getUncompressedContent());
+        byte[] bytes;
```
jmthibault79: not really related to this PR.  Just a little cleanup while I'm here.

src/main/java/htsjdk/samtools/cram/build/CramSpanContainerIterator.java
```diff
@@ -99,10 +99,7 @@ public Container next() {
                     throw new RuntimeException("No more containers in this boundary.");
                 }
                 
-                final long offset = seekableStream.position();
-                final Container c = ContainerIO.readContainer(cramHeader.getVersion(), seekableStream);
-                c.offset = offset;
-                return c;
+                return ContainerIO.readContainer(cramHeader.getVersion(), seekableStream);
```
jmthibault79: `readContainer()` takes care of setting the offset by querying the `SeekableStream`

src/main/java/htsjdk/samtools/cram/structure/Container.java
```diff
@@ -65,13 +65,55 @@
     public CompressionHeader compressionHeader;
 
     // slices found in the container:
-    public Slice[] slices;
+    private Slice[] slices;
```
jmthibault79: Can't set the slices without also setting their byteOffsets now

src/main/java/htsjdk/samtools/cram/structure/ContainerHeaderIO.java
```diff
@@ -70,9 +77,42 @@ public static Container readContainerHeader(final int major, final InputStream i
         if (major >= 3)
             container.checksum = CramInt.readInt32(inputStream);
 
+        container.setByteOffset(containerByteOffset);
```
jmthibault79: this is now the only remaining external call to `setByteOffset()`

src/main/java/htsjdk/samtools/cram/structure/ContainerHeaderIO.java
```diff
@@ -70,9 +77,42 @@ public static Container readContainerHeader(final int major, final InputStream i
...
         return container;
     }
 
+    // convenience methods for SeekableStream and CountingInputStream
+    // TODO: merge these two classes?
```
jmthibault79: these 2 seem VERY similar.  It's annoying to have to account for both.

jmthibault79: Maybe Seekable can inherit from Counting?

src/main/java/htsjdk/samtools/cram/structure/ContainerIO.java
```diff
@@ -19,89 +21,99 @@
...
+     * @param inputStream the {@link InputStream} stream to read from
+     * @param containerByteOffset the byte offset from the start of the stream
      * @return a new container object read from the stream
      */
-    public static Container readContainer(final Version version, final InputStream inputStream) {
-        final Container container = readContainer(version.major, inputStream);
+    public static Container readContainer(final Version version,
+                                          final InputStream inputStream,
+                                          final long containerByteOffset) {
+        Container container = readContainerInternal(version.major, inputStream, containerByteOffset);
```
jmthibault79: Another version of this `readContainer()` consolidation is in #1306

src/main/java/htsjdk/samtools/cram/structure/ContainerIO.java
```diff
@@ -19,89 +21,99 @@
...
 
-        final Container container = readContainerHeader(major, inputStream);
+        final Container container = ContainerHeaderIO.readContainerHeader(major, inputStream, containerByteOffset);
         if (container.isEOF()) {
             return container;
         }
 
         container.compressionHeader = CompressionHeader.read(major, inputStream);
 
-        howManySlices = Math.min(container.landmarks.length, howManySlices);
```
jmthibault79: `howManySlices` and `fromSlice` were never used.  See #1306 for more details.

src/main/java/htsjdk/samtools/cram/structure/ContainerIO.java
```diff
@@ -19,89 +21,99 @@
...
 
-    /**
-     * Writes a {@link Container} header information to a {@link OutputStream}.
-     *
-     * @param major     the CRAM version to assume
-     * @param container the container holding the header to write
-     * @param outputStream        the stream to write to
-     * @return the number of bytes written
-     */
-    public static int writeContainerHeader(final int major, final Container container, final OutputStream outputStream) {
```
jmthibault79: why did this exist

src/test/java/htsjdk/samtools/CRAMSliceMD5Test.java
```diff
@@ -30,10 +31,14 @@ public void testSliceMD5() throws IOException {
         final CramTestCase test = new CramTestCase();
 
         // read the CRAM:
-        final ByteArrayInputStream bais = new ByteArrayInputStream(test.cramData);
-        final CramHeader cramHeader = CramIO.readCramHeader(bais);
-        final Container container = ContainerIO.readContainer(cramHeader.getVersion(), bais);
-        final Slice slice = container.slices[0];
+        Container container;
```
jmthibault79: more unrelated misc cleanup

src/main/java/htsjdk/samtools/cram/structure/Slice.java
```diff
@@ -75,16 +75,51 @@
...
-    public int size = -1;
-    // this Slice's index within its Container
-    public int index = -1;
+    // CRAI only
+    public int byteSize = UNINITIALIZED_INDEXING_PARAMETER;
+    // this Slice's index number within its Container
+    // BAI only
+    public int index = UNINITIALIZED_INDEXING_PARAMETER;
+
+    void baiIndexInitializationCheck() {
```
cmnbroad: This looks like a good check, but it doesn't appear to be called from anywhere except within one test (`craiIndexInitializationCheck` below *is* called before indexing).

cmnbroad: I started looking at this a bit more. It looks like `populateSlicesAndIndexingParameters` is only called when we're creating an index for a CRAM file that we've read in from disk, but I'm unclear on how the slice index info gets populated in the case where we're creating the index on-the-fly, when writing a new CRAM file. Since the tests were passing, I assumed it was happening somehow. So, I naively tried adding a call to this method from within CRAMBAIIndexer.processContainer, and then ran as many of the indexing tests as I easily could. 6 of them failed. So this led to a few questions:
- Where should the call to `baiIndexInitializationCheck` method go ?
- Does the equivalent of `populateSlicesAndIndexingParameters` happen in the case where we're writing a cram out from SAMRecords ?
- If we aren't populating the slice index info when writing a new CRAM file on this branch, how/why is it that all of the index tests are still passing ?
- Is there a similar issue already on master that isn't being caught by tests ?

src/main/java/htsjdk/samtools/cram/build/ContainerFactory.java
```diff
@@ -37,7 +37,13 @@ public ContainerFactory(final SAMFileHeader samFileHeader, final int recordsPerS
         this.recordsPerSlice = recordsPerSlice;
     }
 
-    public Container buildContainer(final List<CramCompressionRecord> records) {
+    /**
+     *
```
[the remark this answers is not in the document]
jmthibault79: Oops, yeah

src/main/java/htsjdk/samtools/cram/structure/ContainerIO.java
```diff
@@ -19,89 +21,97 @@
...
-        try {
-            if (fromSlice > 0) //noinspection ResultOfMethodCallIgnored
-                inputStream.skip(container.landmarks[fromSlice]);
-        } catch (final IOException e) {
-            throw new RuntimeIOException(e);
-        }
-
         final ArrayList<Slice> slices = new ArrayList<>();
-        for (int sliceCount = fromSlice; sliceCount < howManySlices - fromSlice; sliceCount++) {
+        for (final int landmark : container.landmarks) {
```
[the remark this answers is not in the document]
jmthibault79: Yeah I went back and forth about this ... I'll make it an int counter.

src/main/java/htsjdk/samtools/cram/structure/ContainerIO.java
```diff
@@ -19,89 +21,97 @@
...
             slices.add(SliceIO.read(major, inputStream));
         }
 
-        container.populateSlicesAndIndexingParameters(slices);
+        container.populateSlicesAndIndexingParameters(slices, containerByteOffset);
```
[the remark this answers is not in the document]
jmthibault79: Updated by removing its parameters entirely (and renaming to distributeIndexingParametersToSlices).  Slices must be now be set before calling this method.

Added tests to ContainerTest and CRAMBAiIndexerTest

src/main/java/htsjdk/samtools/cram/build/CramContainerHeaderIterator.java
```diff
@@ -22,8 +22,19 @@ public CramContainerHeaderIterator(final InputStream inputStream) {
...
     protected Container containerFromStream(final Version cramVersion, final CountingInputStream countingStream) {
-        final Container container = ContainerIO.readContainerHeader(cramVersion.major, countingStream);
```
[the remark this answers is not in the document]
cmnbroad: Actually, this comment is in the wrong place - it was intended for CramContainerIterator.containerFromStream, not this method.

src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
```diff
@@ -118,21 +126,21 @@ public CRAMBAIIndexer(final OutputStream output, final SAMFileHeader fileHeader)
      *
      * @param container container to be indexed
      */
-    public void processContainer(final Container container, final ValidationStringency validationStringency) {
+    void processContainer(final Container container, final ValidationStringency validationStringency) {
         if (container == null || container.isEOF()) {
             return;
         }
 
+        container.distributeIndexingParametersToSlices();
```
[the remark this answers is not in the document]
jmthibault79: Good catch: yes this can be removed.

There is a validation check later on, in `processAsSingleReferenceSlice()` because that is where all BAI code paths end up.

Since the indexing values all live in Slice (currently) I feel that's the appropriate place to validate them.  Maybe in the future that will change, but that's beyond the scope of this PR.

src/main/java/htsjdk/samtools/cram/structure/Container.java
```diff
@@ -65,13 +65,55 @@
...
+
+    /**
+     * Assign {@link Slice}s to this Container.
+     * Also distribute the Container's byte offset to the {@link Slice}s, for indexing.
+     * @param slices the Slices belonging to this container
+     * @param byteOffset the byte location in the stream where this Container begins
+     */
+    void setSlices(final List<Slice> slices, final long byteOffset) {
+        this.slices = slices.toArray(new Slice[0]);
+        setByteOffsetInSlices(byteOffset);
```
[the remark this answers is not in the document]
jmthibault79: sounds good

src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
```diff
@@ -296,8 +296,11 @@ public static void createIndex(final SeekableStream stream,
      * One instance is used to construct an entire index.
      * processAlignment is called for each alignment until a new reference is encountered, then
      * processReference is called when all records for the reference have been processed.
+     *
+     * TODO: this was copied from {@link htsjdk.samtools.BAMIndexer.BAMIndexBuilder}
```
cmnbroad: This link is broken because this class is not accessible from here; I think either just remove the `@link` or replace with `@see`.

src/test/java/htsjdk/samtools/cram/CRAIEntryTest.java
```diff
@@ -13,13 +14,50 @@
...
+        noSize.byteSize = Slice.UNINITIALIZED_INDEXING_PARAMETER;
+        noSize.containerByteOffset = 1;
+        noSize.byteOffsetFromContainer = 1;
+
+        return new Object[][] {
+                { noContainerOffset },
+                { noOffsetFromContainer },
+                { noSize }
+        };
+    }
```
cmnbroad: This code is all duplicated in `SliceTests`, and should be shared.

src/main/java/htsjdk/samtools/cram/structure/Slice.java
```diff
@@ -113,6 +119,46 @@ public ReferenceContext getReferenceContext() {
...
+        if (index == UNINITIALIZED_INDEXING_PARAMETER) {
+            error.append("Cannot index this Slice for BAI because its index is unknown.").append(System.lineSeparator());
+        }
+
+        if (error.length() > 0) {
+            throw new CRAMException(error.toString());
+        }
+    }
+
+    public void craiIndexInitializationCheck() {
```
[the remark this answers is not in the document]
jmthibault79: I'll add a comment to make the difference clearer.

I'd prefer to keep them separate, since they both test for exactly what is needed under their own requirements.

src/main/java/htsjdk/samtools/CRAMBAIIndexer.java
```diff
@@ -61,11 +62,18 @@
...
  * Class for both constructing BAM index content and writing it out.
+ *
  * There are two usage patterns:
- * 1) Building a bam index from an existing cram file
- * 2) Building a bam index while building the cram file
- * In both cases, processAlignment is called for each cram slice and
- * finish() is called at the end.
+ * 1) Building a bam index (BAI) while building the CRAM file
+ * 2) Building a bam index (BAI) from an existing CRAI file
+ *
```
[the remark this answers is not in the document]
jmthibault79: Right - that's why I didn't include it.  I can add it here with a note that it's unused.

src/test/java/htsjdk/samtools/cram/CRAIEntryTest.java
jmthibault79: Good call.

src/main/java/htsjdk/samtools/cram/structure/Slice.java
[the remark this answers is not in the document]
cmnbroad: I don't see why the distinction, or the increased complexity and code duplication, is useful - there is only one `distributeIndexingParametersToSlices` thats establishes these values. But ok.

PR #1330, opened 2019-03-18
Spotbugs fix tests

lbergelson:
copy of https://github.com/samtools/htsjdk/pull/1277 with a fix for the dataprovider meta test

Addresses #1267 for the tests (except for some valid violation that will be excluded in another PR).


lbergelson:
@tomwhite Thanks for this.  I took it over so we wouldn't have to go back to you for another round.

PR #1334, opened 2019-03-20
SortingCollection logging should only be if DEBUG

alecw:
### Description

A utility like SortingCollection should not fill log output with unnecessary output by default.  Caller can enable debug output if they want to see these messages.  We sometimes create a lot of SortingCollections, and these messages become annoying.

### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing


tfenne:
No objections here.

lbergelson:
@alecw Thank you.

PR #1340, opened 2019-03-26
Added fasta.gz write support to FastaReferenceWriterBuilder

jamesemery:
Added the ability for the FastaReferenceWriterBuilder to emit block compressed output streams. Hooked up some basic checks for this into the existing test code as well.

### DIFF
--- a/src/main/java/htsjdk/samtools/reference/FastaReferenceWriterBuilder.java
+++ b/src/main/java/htsjdk/samtools/reference/FastaReferenceWriterBuilder.java
@@ -38,13 +38,13 @@ import java.nio.file.Path;
 /**
  * Buider for a {@link htsjdk.samtools.reference.FastaReferenceWriter}
  * <p>
- * You can set each of the three outputs (fasta, dictionary and index) to a file or a stream.
- * by default if you provide a file to the fasta an accompanying index and dictionary will be created.
+ * You can set each of the four outputs (fasta, dictionary and index, both gzi and fai) to a file or a stream.
+ * by default if you provide a file to the fasta an accompanying set of indexes and dictionary will be created.
  * This behaviour can be controlled by {@link #setMakeDictOutput(boolean)} and {@link #setMakeFaiOutput(boolean)}
- * <p>
+ * and {@link #setMakeGziOutput(boolean)}<p>
  * The default bases-per-line is {@value FastaReferenceWriter#DEFAULT_BASES_PER_LINE}.
  * <p>
- * Setting a file or an output stream for any of the three outputs (fasta, index or dict) will invalidate the other
+ * Setting a file or an output stream for any of the four outputs (fasta, fai, gzi or dict) will invalidate the other
  * output type (i.e. setting a file output will invalidate a previous stream and vice-versa).
  * </p>
  */
@@ -82,8 +82,9 @@ public class FastaReferenceWriterBuilder {
     }
 
     /**
-     * Set the output fasta file to write to. Doesn't (currently) support compressed filenames.
+     * Set the output fasta file to write to. Will automatically write a blockCompressed output for files ending with ".gz" extension.
      * If the index file and output stream are both null and makeFaiOutput is true (default), a default index file will be created as well.
+     * If the gzi index file and output stream are both null and makeGziOutput is true (default), a default gzi index file will be created if the output is block compressed.
      * If the dictionary file and output stream are both null and makeDictOutput is true (default), a default dictionary file will be created as well.
      *
      * You can only provide a compressed stream to the fastaOutput, and only in the case that an index isn't written.
@@ -103,7 +104,7 @@ public class FastaReferenceWriterBuilder {
 
     /**
      * Sets whether to automatically generate an index file from the name of the fasta-file (assuming it is given
-     * as a file). This can only happen if both the index file and output stream are null.
+     * as a file). This can only happen if both the index file and output stream have not been provided.
      *
      * @param makeFaiOutput a boolean flag
      * @return this builder
@@ -115,7 +116,7 @@ public class FastaReferenceWriterBuilder {
 
     /**
      * Sets whether to automatically generate an dictionary file from the name of the fasta-file (assuming it is given
-     * as a file). This can only happen if both the dictionary file and output stream are null.
+     * as a file). This can only happen if both the dictionary file and output stream have not been provided.
      *
      * @param makeDictOutput a boolean flag
      * @return this builder
@@ -128,7 +129,7 @@ public class FastaReferenceWriterBuilder {
 
     /**
      * Sets whether to automatically generate a gzi index file from the name of the fasta-file (assuming it is given
-     * a valid BCF as a file). This can only happen if both the index file and output stream are null.
+     * a valid bgzipped as a file). This can only happen if both the index file and output stream have not been provided.
      *
      * @param makeGziOutput a boolean flag
      * @return this builder
@@ -153,7 +154,7 @@ public class FastaReferenceWriterBuilder {
     }
 
     /**
-     * Set the output index file to write to. Doesn't (currently) support compressed filenames.
+     * Set the output fai index file to write to.
      */
     public FastaReferenceWriterBuilder setIndexFile(final Path faiIndexFile) {
         this.faiIndexFile = faiIndexFile;
@@ -162,7 +163,7 @@ public class FastaReferenceWriterBuilder {
     }
 
     /**
-     * Set the output index file to write to. Doesn't (currently) support compressed filenames.
+     * Set the output gzi index file to write to.
      */
     public FastaReferenceWriterBuilder setGziIndexFile(final Path gziIndexFile) {
         this.gziIndexFile = gziIndexFile;
@@ -182,6 +183,9 @@ public class FastaReferenceWriterBuilder {
     /**
      * Set the output stream for writing the reference. Doesn't support compressed streams.
      *
+     * NOTE: If you would like to output a BlockCompressed fasta file it is recommended you use {@link #setFastaFile(Path)}
+     *       as that codepath will handle generation of a gzi index automatically.
+     *
      * @param fastaOutput a {@link OutputStream} for the output fasta file.
      * @return this builder
      */
@@ -189,6 +193,7 @@ public class FastaReferenceWriterBuilder {
     public FastaReferenceWriterBuilder setFastaOutput(final OutputStream fastaOutput) {
         this.fastaOutput = fastaOutput;
         this.fastaFile = null;
+        this.gzippedFastaFile = false;
         this.makeGziOutput = false;
         return this;
     }
@@ -248,24 +253,36 @@ public class FastaReferenceWriterBuilder {
             throw new IllegalArgumentException("Both fastaFile and fastaOutput were null. Please set one of them to be non-null.");
         }
         if(fastaFile != null) {
+
             if (faiIndexFile == null && faiIndexOutput == null) {
                 faiIndexFile = defaultFaiFile(makeFaiOutput, fastaFile);
+                // Asserting that we were not provided two paths for fai output files
             } else if (faiIndexFile != null && faiIndexOutput != null) {
                 throw new IllegalArgumentException("Both faiIndexFile and faiIndexOutput were non-null. Please set one of them to be null.");
             }
+
             if (dictFile == null && dictOutput == null) {
                 dictFile = defaultDictFile(makeDictOutput, fastaFile);
+                // Asserting that we were not provided two paths for dict output files
             } else if (dictFile != null && dictOutput != null) {
                 throw new IllegalArgumentException("Both dictFile and dictOutput were non-null. Please set one of them to be null.");
             }
+
             if (gzippedFastaFile && gziIndexFile == null && gziIndexOutput == null) {
                 gziIndexFile = defaultGziFile(makeGziOutput, fastaFile);
-            } else if (gziIndexFile != null && gziIndexOutput != null) {
-                throw new IllegalArgumentException("Both dictFile and dictOutput were non-null. Please set one of them to be null.");
+                // Asserting that we were not given a path for a gzi output for a non-block compressed output
             } else if (!gzippedFastaFile && (gziIndexFile != null || gziIndexOutput != null)) {
                 throw new IllegalArgumentException("Requested a gzi index but the output format fasta file was not a block compressed gzip file");
             }
-            if (faiIndexFile != null && gzippedFastaFile && gziIndexFile == null) {
+
+            // Asserting that we were not provided two paths for gzi output files
+            if (gziIndexFile != null && gziIndexOutput != null) {
+                throw new IllegalArgumentException("Both dictFile and dictOutput were non-null. Please set one of them to be null.");
+            }
+            // Asserting that that a Gzi output must accompany the fai output for block compressed output
+            if ((faiIndexFile != null || faiIndexOutput != null) &&
+                    gzippedFastaFile &&
+                    (gziIndexFile == null || gziIndexOutput != null)) {
                 throw new IllegalArgumentException("Requested a fai index file for a block compressed output file. This index is unusable without a gzi index file as well");
             }
 
--- a/src/main/java/htsjdk/samtools/reference/FastaReferenceWriterBuilder.java
+++ b/src/main/java/htsjdk/samtools/reference/FastaReferenceWriterBuilder.java
@@ -50,7 +50,7 @@ import java.nio.file.Path;
  */
 public class FastaReferenceWriterBuilder {
     private Path fastaFile;
-    private boolean gzippedFastaFile = false;
+    private boolean gzippedFastaFile;
     private boolean makeGziOutput = true;
     private boolean makeFaiOutput = true;
     private boolean makeDictOutput = true;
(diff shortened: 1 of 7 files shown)


jamesemery:
@droazen Would you have time to take a look at this branch?

src/main/java/htsjdk/samtools/reference/FastaReferenceWriterBuilder.java
```diff
@@ -127,9 +155,18 @@ public FastaReferenceWriterBuilder setBasesPerLine(final int basesPerLine) {
...
+        this.faiIndexOutput = null;
+        return this;
+    }
+
+    /**
+     * Set the output index file to write to. Doesn't (currently) support compressed filenames.
+     */
+    public FastaReferenceWriterBuilder setGziIndexFile(final Path gziIndexFile) {
+        this.gziIndexFile = gziIndexFile;
+        this.gziIndexOutput = null;
```
cmnbroad: Shouldn't this set makeGziOutput = true for consistency ? Also, does the filename have to conform (".gzi") ? Will the index work if its something else ?

src/main/java/htsjdk/samtools/reference/FastaReferenceWriterBuilder.java
```diff
@@ -198,31 +248,50 @@ public FastaReferenceWriter build() throws IOException {
             throw new IllegalArgumentException("Both fastaFile and fastaOutput were null. Please set one of them to be non-null.");
         }
         if(fastaFile != null) {
-            if (indexFile == null && indexOutput == null) {
-                indexFile = defaultFaiFile(makeFaiOutput, fastaFile);
-            } else if (indexFile != null && indexOutput != null) {
-                throw new IllegalArgumentException("Both indexFile and indexOutput were non-null. Please set one of them to be null.");
+            if (faiIndexFile == null && faiIndexOutput == null) {
+                faiIndexFile = defaultFaiFile(makeFaiOutput, fastaFile);
```
cmnbroad: what if gzippedFastaFile == true ?

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -431,4 +431,34 @@ private static final ByteBuffer allocateBuffer(final int numberOfEntries,
...
+        public GZIIndexer(final OutputStream outputStream) {
+            output = outputStream;
+        }
+
+        public GZIIndexer(final Path outputFile) throws IOException {
+            output = Files.newOutputStream(outputFile);
+        }
+
+        // Adds a new index location given the compressed file offset and a running tally based on the uncompressed block sizes
+        public void addGzipBlock(final long compressedFileileOffset, final long uncompressedBlockSize) {
```
cmnbroad: compressedFileileOffset -> compressedFileOffset

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -431,4 +431,34 @@ private static final ByteBuffer allocateBuffer(final int numberOfEntries,
...
+            IndexEntry indexEntry = new IndexEntry(compressedFileileOffset, uncompressedFileOffset);
+            uncompressedFileOffset += uncompressedBlockSize;
+            entries.add(indexEntry);
+        }
+
+        @Override
+        public void close() throws IOException {
+            GZIIndex index = new GZIIndex(entries);
+            index.writeIndex(output);
```
cmnbroad: This should close the stream it opened in the constructor.

src/test/java/htsjdk/samtools/reference/FastaReferenceWriterTest.java
```diff
@@ -195,6 +192,18 @@ public void testAddingInvalidSequenceName(final String invalidName) throws IOExc
...
+    @Test(expectedExceptions = IllegalArgumentException.class)
+    public void testAddingGziIndexToNonBCFFile() throws IOException {
+        final Path testOutput = File.createTempFile("fwr-test", ".fasta").toPath();
+        final Path testOutputGZI = File.createTempFile("fwr-test", ".fasta.gzi").toPath();
+        IOUtil.deleteOnExit(testOutput);
+        Files.delete(testOutput);
+
+        try (FastaReferenceWriter writer = new FastaReferenceWriterBuilder().setFastaFile(testOutput).setMakeFaiOutput(false).setMakeDictOutput(false).setGziIndexFile(testOutputGZI).build()) {
+            writer.startSequence(testOutput.toString());
+        }
```
cmnbroad: The `writer.startSequence` call is unnecessary/redundant.

src/test/java/htsjdk/samtools/reference/FastaReferenceWriterTest.java
```diff
@@ -195,6 +192,18 @@ public void testAddingInvalidSequenceName(final String invalidName) throws IOExc
         }
     }
 
+    @Test(expectedExceptions = IllegalArgumentException.class)
+    public void testAddingGziIndexToNonBCFFile() throws IOException {
+        final Path testOutput = File.createTempFile("fwr-test", ".fasta").toPath();
+        final Path testOutputGZI = File.createTempFile("fwr-test", ".fasta.gzi").toPath();
+        IOUtil.deleteOnExit(testOutput);
+        Files.delete(testOutput);
```
cmnbroad: Not sure these files need to actually exist for this test, but they only need to be deleted once. If you're trying to get the index deleted as well, I think you need to create a temp dir (I don't think there is any guarantee that two temp files are siblings) ?

src/main/java/htsjdk/samtools/util/BlockCompressedOutputStream.java
```diff
@@ -392,14 +406,19 @@ private int deflateBlock() {
...
         assert(bytesToCompress <= numUncompressedBytes);
 
+        // Call out to the indexer if it exists
+        if (indexer != null) {
+            indexer.addGzipBlock(mBlockAddress, numUncompressedBytes);
+        }
+
         // Clear out from uncompressedBuffer the data that was written
-        if (bytesToCompress == numUncompressedBytes) {
+//        if (bytesToCompress == numUncompressedBytes) {
```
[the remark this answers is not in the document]
jamesemery: Yes, though it should have been deleted not commented out. That condition was a holdover from a much older version of this code where bytesToCompress might have been mutated when there was an artificial throttle on the number of bytes per block before the noCompressionDeflater was added. Having stared at this code for a while I think eliminating confusing and unreachable branches helps readability so I'll delete this code instead.

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -225,27 +224,27 @@ public String toString() {
...
             // implementation of entry ensures that the offsets are no negative
-            buffer.putLong(entry.getCompressedOffset());
-            buffer.putLong(entry.getUncompressedOffset());
+            codec.writeLong(entry.getCompressedOffset());
+            codec.writeLong(entry.getUncompressedOffset());
         }
 
         // write into the output
-        Files.write(output, buffer.array());
+        codec.close();
```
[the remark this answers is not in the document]
jamesemery: done

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -389,6 +388,7 @@ public static final GZIIndex buildIndex(final Path bgzipFile) throws IOException
                     // gets the block address (compressed offset) - requires to parse with the file pointer utils
                     final long compressed = BlockCompressedFilePointerUtil.getBlockAddress(bgzipStream.getFilePointer());
                     // add a new IndexEntry
+                    //TODO unify this with the Indexer code
```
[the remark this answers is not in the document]
jamesemery: I should have been more careful about vetting this branch for spurious out of date todos?

src/main/java/htsjdk/samtools/reference/FastaReferenceWriterBuilder.java
```diff
@@ -111,6 +126,19 @@ public FastaReferenceWriterBuilder setMakeDictOutput(final boolean makeDictOutpu
         return this;
     }
 
+    /**
+     * Sets whether to automatically generate a gzi index file from the name of the fasta-file (assuming it is given
+     * a valid BCF as a file). This can only happen if both the index file and output stream are null.
```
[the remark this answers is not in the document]
jamesemery: This is based on the comments for the other files. I've updated them to hopefully be somewhat clearer

src/main/java/htsjdk/samtools/reference/FastaReferenceWriterBuilder.java
```diff
@@ -152,18 +189,31 @@ public FastaReferenceWriterBuilder setDictFile(final Path dictFile) {
     public FastaReferenceWriterBuilder setFastaOutput(final OutputStream fastaOutput) {
         this.fastaOutput = fastaOutput;
         this.fastaFile = null;
+        this.makeGziOutput = false;
```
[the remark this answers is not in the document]
jamesemery: I commented this behavior explicitly in the javadocs. The GZI index must be manually attached to the block compressed output stream. Consequently if the user provides their own output stream we should just assume nothing is block compressed and disable that code .

src/main/java/htsjdk/samtools/reference/FastaReferenceWriterBuilder.java
```diff
@@ -127,9 +155,18 @@ public FastaReferenceWriterBuilder setBasesPerLine(final int basesPerLine) {
     /**
      * Set the output index file to write to. Doesn't (currently) support compressed filenames.
      */
-    public FastaReferenceWriterBuilder setIndexFile(final Path indexFile) {
-        this.indexFile = indexFile;
-        this.indexOutput = null;
+    public FastaReferenceWriterBuilder setIndexFile(final Path faiIndexFile) {
+        this.faiIndexFile = faiIndexFile;
+        this.faiIndexOutput = null;
+        return this;
```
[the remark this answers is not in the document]
jamesemery: I vetted all of the javadocs in this class for out of date allusions to compressed output. It is entirely acceptable behavior for a user to have a fasta.gz file with both a gzi index and a fai index. There is code to handle both of those indexes together invisibly elsewhere. Note I didn't update `setIndexFile()` -> `setFaiIndexFile()` because that seemed like it would be likely to break compatibility to dependent projects when most of the rest of this branch was purely additive.

src/main/java/htsjdk/samtools/reference/FastaReferenceWriterBuilder.java
```diff
@@ -47,16 +50,24 @@
...
 public class FastaReferenceWriterBuilder {
     private Path fastaFile;
+    private boolean gzippedFastaFile = false;
+    private boolean makeGziOutput = true;
     private boolean makeFaiOutput = true;
     private boolean makeDictOutput = true;
     private boolean emitMd5 = true;
     private int basesPerLine = FastaReferenceWriter.DEFAULT_BASES_PER_LINE;
-    private Path indexFile;
+    private Path gziIndexFile;
```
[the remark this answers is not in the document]
jamesemery: Yes, as we have discussed in person it is valid to produce both output files together.

src/main/java/htsjdk/samtools/reference/FastaReferenceWriterBuilder.java
jamesemery: Hmm... no a gzi index named anything other than fastafile.gzi will fail. I don't see this as a unique problem given that the user could set the fai index file to any arbitrary name elsewhere. My inclination is to simply leave it as is.

jamesemery: Also we don't need to set makeGziOutput = true as the make<FILE>Output flags can be more accurately thought of as "should this tool automatically generate this output given that the user hasn't provided me with a path for this file explicitly"

src/test/java/htsjdk/samtools/reference/FastaReferenceWriterTest.java
jamesemery: fair point

src/test/java/htsjdk/samtools/reference/FastaReferenceWriterTest.java
jamesemery: I set both files to delete on exit to be sure. It doesn't matter for this test since this is just asserting validation behavior on the `.build()` step.

jamesemery:
@cmnbroad Responded to your comments based on our discussion.

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -221,31 +220,33 @@ public String toString() {
...
      * Writes this index into the requested path.
      *
+     * NOTE: This method will close out the provided output stream when it finishes writing the index
+     *
      * @param output the output file.
      *
      * @throws IOException if an I/O error occurs.
      */
-    public void writeIndex(final Path output) throws IOException {
+    void writeIndex(final OutputStream output) throws IOException {
```
cmnbroad: We shouldn't just remove an existing public method without at least deprecating it first. However, it seems unnecessary to do when you could just add the stream overload and delegate to that.

src/main/java/htsjdk/samtools/util/BlockCompressedOutputStream.java
```diff
@@ -258,6 +259,19 @@ public static BlockCompressedOutputStream maybeBgzfWrapOutputStream(final File l
         }
     }
 
+    /**
+     * Adds a GZIIndexer to the block compressed output stream to be written to the specified output stream. See
+     * {@link GZIIndex} for details on the index. Note that the stream will be written to disk entirely when close()
+     * is called.
+     * @throws RuntimeException if attempting to add an indexer when this stream has already accepted input.
```
cmnbroad:
```suggestion
     * @throws RuntimeException if this method is called after output has already been written to the stream.
```

src/main/java/htsjdk/samtools/reference/FastaReferenceWriterBuilder.java
```diff
@@ -35,28 +38,36 @@
...
  * The default bases-per-line is {@value FastaReferenceWriter#DEFAULT_BASES_PER_LINE}.
  * <p>
- * Setting a file or an output stream for any of the three outputs (fasta, index or dict) will invalidate the other
+ * Setting a file or an output stream for any of the four outputs (fasta, fai, gzi or dict) will invalidate the other
  * output type (i.e. setting a file output will invalidate a previous stream and vice-versa).
  * </p>
  */
 public class FastaReferenceWriterBuilder {
     private Path fastaFile;
+    private boolean gzippedFastaFile = false;
```
pshapiro4broad: Initializer unnecessary as `false` is already the value of an uninitialized boolean.
```suggestion
    private boolean gzippedFastaFile;
```

src/main/java/htsjdk/samtools/reference/FastaReferenceWriterBuilder.java
```diff
@@ -145,25 +183,42 @@ public FastaReferenceWriterBuilder setDictFile(final Path dictFile) {
     /**
      * Set the output stream for writing the reference. Doesn't support compressed streams.
      *
```
pshapiro4broad: needs a `<p>` to format correctly as javadoc
```suggestion
     * <p>
```

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -431,4 +432,34 @@ private static final ByteBuffer allocateBuffer(final int numberOfEntries,
         // creates a byte buffer in little-endian
         return ByteBuffer.allocate(size).order(ByteOrder.LITTLE_ENDIAN);
     }
+
+    /**
+     *
+     */
+    public static final class GZIIndexer implements Closeable {
+        private int uncompressedFileOffset = 0;
```
pshapiro4broad: unnecessary initializer
```suggestion
        private int uncompressedFileOffset;
```

src/main/java/htsjdk/samtools/reference/FastaReferenceWriterBuilder.java
```diff
@@ -145,25 +183,42 @@ public FastaReferenceWriterBuilder setDictFile(final Path dictFile) {
...
-        this.indexOutput = indexOutput;
-        this.indexFile = null;
+    public FastaReferenceWriterBuilder setIndexOutput(final OutputStream faiIndexOutput) {
+        this.faiIndexOutput = faiIndexOutput;
+        this.faiIndexFile = null;
+        return this;
+    }
+
+    /**
+     * Set the output stream for writing the index. Doesn't support compressed streams.
```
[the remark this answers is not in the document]
jamesemery: This was copied from above. The statement I take to imply that the provided output stream to the method is not a compressed stream. I will clarify that. (Honestly i'm not sure how the GZI index code will handle a gzipped gzi index file)

src/main/java/htsjdk/samtools/util/GZIIndex.java
jamesemery: Whoops, this was accidental. I agree keeping the old api should be fine (despite it not being overwhelmingly old).

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -221,31 +220,33 @@ public String toString() {
...
      * Writes this index into the requested path.
      *
+     * NOTE: This method will close out the provided output stream when it finishes writing the index
+     *
      * @param output the output file.
      *
      * @throws IOException if an I/O error occurs.
      */
-    public void writeIndex(final Path output) throws IOException {
+    public void writeIndex(final OutputStream output) throws IOException {
```
cmnbroad: My comment in the last round was about the fact that you changed the signature for this method, which was already public, which amounts to removing an existing method without first deprecating it. So although its probably correct for this new one be public, the old method that takes a Path needs to be restored, and changed to delegate to this one.

jamesemery:
@cmnbroad Added the method back in, do you have any other comments to this?

src/main/java/htsjdk/samtools/util/GZIIndex.java
```diff
@@ -217,6 +217,19 @@ public String toString() {
         return "GZIIndex:" + StringUtil.join(", ", entries);
     }
 
+    /**
+     * Writes this index into the requested path.
+     *
+     * NOTE: This method will close out the provided output stream when it finishes writing the index
```
cmnbroad: @jamesemery This comment is misplaced here- this method has no `provided output stream`.

Issue #255, opened 2015-05-19
VariantContextBuilder does not perform a deep copy


d-cameron:
When using VariantContextBuilder to create a new variant based on an existing variant, the underlying objects and collections are not copied nor are they wrapped with copy-on-write semantics. Making changes to a copied variant change the original variant as well.

Failing unit test for VariantContextUnitTest.java
```
@Test
    public void testFilterUnaffectedByClonedVariants() {
    	VariantContext vc1 = new VariantContextBuilder("source", "contig", 1, 1, Arrays.asList(Tref, A)).filter("TEST").make();
    	VariantContext vc2 = new VariantContextBuilder(vc1).filter("TEST2").make();
    	Assert.assertFalse(vc1.getFilters().equals(vc2.getFilters()));
    }
```

yfarjoun:
attributes *are* protected against this, I guess filters need the same
treatment.

d-cameron:
Yes, attributesCanBeModified does look pretty obvious when reviewing the code again. Depending on when GenotypesContext is/is not immutable, genotypes *might* also be an issue.

yfarjoun:
also include a test without calling make():
```
@Test
    public void testFilterUnaffectedByClonedVariants() {
        VariantContext vc1 = new VariantContextBuilder("source", "contig", 1, 1, Arrays.asList(Tref, A)).filter("TEST").make();
        VariantContext vc2 = new VariantContextBuilder(vc1).filter("TEST2");
        Assert.assertFalse(vc1.getFilters().equals(vc2.getFilters()));
    }
```

yfarjoun:
this is actually worse. I can get it to fail on filters (as reported) attributes, genotypes and alleles....

yfarjoun:
submitting a PR with tests for comments.

tfenne:
Fixed by #1344 (with follow-up #1403). VariantContextBuilder now copies filters/attributes on write, and the failing test from this issue is covered by testFiltersUnaffectedByClonedVariants. Closing as resolved.

PR #1344, opened 2019-04-03
make VariantContextBuilder safer

### LINKED ISSUE #1365: VariantContextBuilder : putAttributes fails if Map cannot be modified
### Subject of the issue

I got an exception with `VariantContextBuilder.putAttributes`: UnsupportedOperationException

in https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java#L273

if `this.attributes` is not null and was created from `public VariantContextBuilder(final VariantContext parent)` , as far as I can see `this.attributes` is **unmodifieable**.

I could write a PR but I don't really understand this `attributesCanBeModified`.

### Your environment
* version of htsjdk  2.2.4
* version of java 1.8
* which OS linux

yfarjoun:
Adds tests and fixes unsafe building patterns that were enabled by VariantContextBuilder. 

This required a small change in the API around adding a genotype context. Now when .genotypes() is called with a GenotypeContext object the builder will call immutable() on that object, to make sure that the user doesn't change the GC prior to the creation of the VariantContext...If folks really need to reuse the GenotypeContext objects across multiple build calls, we could add a .genotypesKeepMutable() method with a scary javaDoc note.....

Another change in the API (somewhat) is that validation was added for the filters as per the hts-spec.

fixes #1365


yfarjoun:
Several of these use-patterns fail (as you can see in the test results...) my question is..are they reasonable use-patterns and should we fix them?

yfarjoun:
@d-cameron this fixes the issue you reported 4 years ago: #255   

Care to take a look?

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -225,7 +226,7 @@
     protected CommonInfo commonInfo = null;
     public final static double NO_LOG10_PERROR = CommonInfo.NO_LOG10_PERROR;
 
-    public final static Set<String> PASSES_FILTERS = Collections.unmodifiableSet(new LinkedHashSet<String>());
+    public final static Set<String> PASSES_FILTERS = Collections.unmodifiableSet(new LinkedHashSet<>());
```
pshapiro4broad: correct modifier sort order is
```suggestion
    public static final Set<String> PASSES_FILTERS = Collections.unmodifiableSet(new LinkedHashSet<>());
```
Although I don't see what the point of this is. It's an empty set that can't be modified? If that's what's needed, it already exists as `Collections.emptySet()`.

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1366,6 +1370,21 @@ private void validateGenotypes() {
         }
     }
 
+    //no controls and white-spaces characters, no semicolon.
+    static final public Pattern VALID_FILTER = Pattern.compile("^[!-:<-~]+$");
```
pshapiro4broad:
```suggestion
    public static final Pattern VALID_FILTER = Pattern.compile("^[!-:<-~]+$");
```

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1366,6 +1370,21 @@ private void validateGenotypes() {
...
+    private void validateFilters() {
+        final Set<String> filters = this.getFilters();
+        if (filters == null) return;
+
+        for (String filter : filters) {
+            if (!VALID_FILTER.matcher(filter).matches()){
+                throw new IllegalStateException("Filter '" + filter +
+                        "' contains an illegal character. It must conform to the regex ;'" + VALID_FILTER.toString());
```
pshapiro4broad: unnecessary `toString()`
```suggestion
                        "' contains an illegal character. It must conform to the regex ;'" + VALID_FILTER);
```

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
```diff
@@ -28,16 +28,7 @@
...
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
pshapiro4broad: Using wildcard imports makes it harder to figure out where a name came from when reading the code outside of an IDE. Changing an IDE setting can enforce the rule to keep imports, instead of replacing them with a wildcard.

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
```diff
@@ -167,25 +159,33 @@ public GenotypesContext getGenotypes() {
      * @param parent  Cannot be null
      */
     public VariantContextBuilder(final VariantContext parent) {
-        if ( parent == null ) throw new IllegalArgumentException("BUG: VariantContextBuilder parent argument cannot be null in VariantContextBuilder");
+        if( parent == null ) throw new IllegalArgumentException("BUG: VariantContextBuilder parent argument cannot be null in VariantContextBuilder");
```
pshapiro4broad: fix whitespace, format across lines
```
if (parent == null) {
    throw new ...;
}
```

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
```diff
@@ -323,12 +323,26 @@ public VariantContextBuilder rmAttributes(final List<String> keys) {
...
+                this.attributes = new HashMap<>(attributes);
+            }
+        }
+    }
+
+    /**
+     * Makes the filters modifiable.
+     */
+    private void makeFiltersModifiable() {
+        if ( !filtersCanBeModified) {
```
pshapiro4broad: whitespace
```suggestion
        if (!filtersCanBeModified) {
```

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
```diff
@@ -323,12 +323,26 @@ public VariantContextBuilder rmAttributes(final List<String> keys) {
...
+            this.filtersCanBeModified = true;
+            if (filters == null) {
```
pshapiro4broad: could be written as
```
this.filters = new HashSet<>();
if (filters != null) {
    this.filters.addAll(filters);
}
```

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
```diff
@@ -342,23 +356,52 @@ private void makeAttributesModifiable() {
      * @return this builder
      */
     public VariantContextBuilder filters(final Set<String> filters) {
-        this.filters = filters;
+        makeFiltersModifiable();
+        if (filters == null) {
+            unfiltered();
```
pshapiro4broad: This nulls out `filters`; if this path is taken than there's no point in calling `makeFiltersModifiable()`

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
```diff
@@ -342,23 +356,52 @@ private void makeAttributesModifiable() {
      * @return this builder
      */
     public VariantContextBuilder filters(final Set<String> filters) {
-        this.filters = filters;
+        makeFiltersModifiable();
```
pshapiro4broad: Maybe a better approach would be to pass `filters` to `makeFiltersModifiable()`? It seems like you've got similar logic split across different methods here.

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
```diff
@@ -342,23 +356,52 @@ private void makeAttributesModifiable() {
...
+    /** Adds the given filter to the list of filters
+     *
+     * @param filter
+     * @return
+     */
     public VariantContextBuilder filter(final String filter) {
-        if ( this.filters == null ) this.filters = new LinkedHashSet<String>(1);
+        makeFiltersModifiable();
+        if ( this.filters == null ) {
+            filtersAsIs(new LinkedHashSet<>(1));
```
pshapiro4broad: isn't this already done by `makeFiltersModifiable()`?

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
```diff
@@ -342,23 +356,52 @@ private void makeAttributesModifiable() {
...
+
     /**
      * {@link #filters}
      *
      * @param filters  Set of strings to set as the filters for this builder
      * @return this builder
      */
     public VariantContextBuilder filters(final String ... filters) {
-        filters(new LinkedHashSet<String>(Arrays.asList(filters)));
+        filtersAsIs(new LinkedHashSet<>(Arrays.asList(filters)));
```
pshapiro4broad: How is this different than `filters(new LinkedHashSet<>(Arrays.asList(filters)))`?

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
yfarjoun: hmmm. that will not work as written, since filters will never be null.

yfarjoun: but I get your drift.

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
yfarjoun: doesn't bother call make filters modifyable.

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
yfarjoun: yup. right on.

yfarjoun:
back to you @pshapiro4broad

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
yfarjoun: 👍

yfarjoun:
@pshapiro4broad are you OK with this PR?

yfarjoun:
@pshapiro4broad 👀 please?

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
```diff
@@ -339,26 +366,55 @@ private void makeAttributesModifiable() {
...
+     *                This set will be copied so that external set can be
+     *                safely changed.
      * @return this builder
      */
     public VariantContextBuilder filters(final Set<String> filters) {
-        this.filters = filters;
+        if (filters == null) {
+            unfiltered();
+        } else {
+            makeFiltersModifiable();
```
pshapiro4broad: I think these lines
```
            makeFiltersModifiable();
            this.filters.clear();
            this.filters.addAll(filters);
            toValidate.add(VariantContext.Validation.FILTERS);
```
can be replaced with
```
this.filtersCanBeModified = true;
filtersAsIs(new HashSet<>(filters));
```

There's a lot of overlapping methods so it's hard for me to keep clear exactly what method does what thing. On the other hand, maybe it's OK to merge it up without cleaning it up too much...

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
```diff
@@ -323,12 +335,27 @@ public VariantContextBuilder rmAttributes(final List<String> keys) {
...
+                this.attributes = new HashMap<>(tempAttributes);
+            }
+        }
+    }
+
+    /**
+     * Makes the filters modifiable.
+     */
+    private void makeFiltersModifiable() {
+        if (!filtersCanBeModified) {
```
pshapiro4broad: Instead of `filtersCanBeModified` could you instead always make the filters modifiable? E.g. whenever setting filters always create a copy, so you don't have to worry about making it modifiable later on. Or is the desire here to avoid copying `filters` unless you have to, e.g. due to memory or other concerns?

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
```diff
@@ -393,6 +452,7 @@ public VariantContextBuilder unfiltered() {
     public VariantContextBuilder genotypes(final GenotypesContext genotypes) {
         this.genotypes = genotypes;
         if ( genotypes != null )
```
pshapiro4broad: Here's a bug! Introduced by the lack of `{` `}`!

Please add some `{` `}`s.

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
```diff
@@ -574,7 +634,10 @@ public VariantContext make() {
     }
 
     public VariantContext make(final boolean leaveModifyableAsIs) {
-        if(!leaveModifyableAsIs) attributesCanBeModified = false;
+        if(!leaveModifyableAsIs){
```
pshapiro4broad: Add whitespace please
```suggestion
        if (!leaveModifyableAsIs) {
```

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
yfarjoun: yes, this is an optimization, since the normal pattern doesn't involve changing the filters after instantiating...

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
yfarjoun: I'm not sure I understand what you mean. I see that the pattern of `makeFiltersModifiable` is identical to that of `makeAttributesModifiable`, but making one method out of that seems a bit contrived....

yfarjoun: oh. nm, I did this :-)

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
yfarjoun: :+1:

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
yfarjoun: it wouldn't be that bad to validate a null genotypeContext...but sure, not as intended.

yfarjoun:
Thanks again @pshapiro4broad back to you.

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1292,17 +1383,12 @@ public void validateChromosomeCounts() {
...
     private boolean validate(final EnumSet<Validation> validationToPerform) {
         validateStop();
-        for (final Validation val : validationToPerform ) {
-            switch (val) {
-                case ALLELES: validateAlleles(); break;
-                case GENOTYPES: validateGenotypes(); break;
-                default: throw new IllegalArgumentException("Unexpected validation mode " + val);
-            }
-        }
+        validationToPerform.forEach(v->v.validate(this));
```
pshapiro4broad: whitespace
```suggestion
        validationToPerform.forEach(v -> v.validate(this));
```

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -310,9 +322,88 @@
...
+
+            if (genotypes == null) {
+                throw new IllegalStateException("Genotypes is null");
+            }
+
+            for (int i = 0; i < genotypes.size(); i++) {
+                final Genotype genotype = genotypes.get(i);
+                if (genotype.isAvailable()) {
+                    final List<Allele> alleles = genotype.getAlleles();
+                    for (int j = 0, size = alleles.size(); j < size; j++) {
```
pshapiro4broad: Now that I'm reading this code, due to the move, I have a comment. It looks like this can be replaced with a foreach loop, e.g.
```
for (Allele gAllele : genotype.getAlleles()) {
    ...
}
```

src/main/java/htsjdk/variant/variantcontext/CommonInfo.java
```diff
@@ -114,7 +114,7 @@ public boolean isNotFiltered() {
 
     public void addFilter(String filter) {
         if ( filters == null ) // immutable -> mutable
-            filters = new HashSet<String>();
+            filters = new HashSet<>();
```
lbergelson: {}

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -310,9 +322,88 @@
...
+                        final Allele gAllele = alleles.get(j);
+                        if (!variantContext.hasAllele(gAllele) && gAllele.isCalled()) {
+                            throw new IllegalStateException("Allele in genotype " + gAllele + " not in the variant context " + alleles);
+                        }
+                    }
+                }
+            }
+        }
+
+        static private void validateFilters(final VariantContext variantContext) {
```
lbergelson:
```suggestion
        private static void validateFilters(final VariantContext variantContext) {
```

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -310,9 +322,88 @@
...
+        FILTERS {
+            void validate(VariantContext variantContext) {
+                validateFilters(variantContext);
+            }
+        };
+
+        abstract void validate(VariantContext variantContext);
+
+
+        static private void validateAlleles(final VariantContext vc) {
```
lbergelson:
```suggestion
        private static void validateAlleles(final VariantContext vc) {
```

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -310,9 +322,88 @@
...
+                }
+            }
+
+            // make sure there's one reference allele
+            if (!alreadySeenRef) {
+                throw new IllegalArgumentException("No reference allele found in VariantContext");
+            }
+        }
+
+        static private void validateGenotypes(final VariantContext variantContext) {
```
lbergelson:
```suggestion
        private static void validateGenotypes(final VariantContext variantContext) {
```

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -260,6 +270,8 @@
 * Determine which genotype fields are in use in the genotypes in VC
 * @return an ordered list of genotype fields in use in VC.  If vc has genotypes this will always include GT first
 */
+
+
```
lbergelson: weird extra linebreaks.

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
lbergelson: also  validation(**s**)ToPerform would make more sense

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
```diff
@@ -339,26 +367,53 @@ private void makeAttributesModifiable() {
...
 
+    /**
+     * This builder's filters are set to this value
+     *
+     * filters can be <code>null</code> -&gt; meaning there are no filters
+     *
+     * @param filters Set of strings to set as the filters for this builder
+     * @return this builder
+     */
+    private void filtersAsIs(final Set<String> filters) {
```
lbergelson: should this just be called set filters?

src/test/java/htsjdk/variant/variantcontext/VariantContextBuilderTest.java
```diff
@@ -21,6 +27,8 @@ public void before() {
 
         C = Allele.create("C");
         Aref = Allele.create("A", true);
+        G = Allele.create("G");
+        Tref = Allele.create("T", true);
```
lbergelson: we need to just expose these as public members of allele.  it's stupid to have to redefine them in every test

src/test/java/htsjdk/variant/variantcontext/VariantContextBuilderTest.java
```diff
@@ -47,19 +55,207 @@ public void testAttributeResettingWorks(final boolean leaveModifyableAsIs) {
         }
     }
 
+
     @Test()
-    public void testAttributeResettingWorks() {
-        final VariantContextBuilder root1 = new VariantContextBuilder(snpSource, snpLoc, snpLocStart, snpLocStop, Arrays.asList(Aref, C));
-        final VariantContextBuilder root2 = new VariantContextBuilder(snpSource, snpLoc, snpLocStart, snpLocStop, Arrays.asList(Aref, C));
+    public void testBuilkAttributeResettingWorks() {
```
lbergelson: typo
```suggestion
    public void testBulkAttributeResettingWorks() {
```

src/test/java/htsjdk/variant/variantcontext/VariantContextBuilderTest.java
```diff
@@ -47,19 +55,207 @@ public void testAttributeResettingWorks(final boolean leaveModifyableAsIs) {
...
+        Assert.assertEquals(result1.getAttribute("AN"), result2.getAttribute("AN"));
+
+        final HashMap<String, Object> map = new HashMap<>();
+        map.put("AC", 2);
+        map.put("AN", 4);
+        root2.attributes(map);
+
+        final VariantContext result3 = root2.make();
+
+        Assert.assertEquals(result3.getAttribute("AC"), 2);
```
lbergelson: should this also assert that result1 and 2 are not modified?

src/test/java/htsjdk/variant/variantcontext/VariantContextBuilderTest.java
```diff
@@ -47,19 +55,207 @@ public void testAttributeResettingWorks(final boolean leaveModifyableAsIs) {
...
+            VariantContextBuilder getOtherBuilder(final VariantContextBuilder builder, final VariantContext vc) {
+                return builder;
+            }
+        };
+
+        abstract VariantContextBuilder getOtherBuilder(final VariantContextBuilder builder, VariantContext vc);
+    }
+
+    @DataProvider
+    public Object[][] BuilderSchemes() {
```
lbergelson:
```suggestion
    public Object[][] builderSchemes() {
```

src/test/java/htsjdk/variant/variantcontext/VariantContextBuilderTest.java
```diff
@@ -47,19 +55,207 @@ public void testAttributeResettingWorks(final boolean leaveModifyableAsIs) {
...
+
+        filters.clear();
+        filters.add("TEST2");
+
+        final VariantContext vc2 = builderScheme.getOtherBuilder(builder, vc1).filters(filters).make();
+        Assert.assertNotEquals(vc2.getFilters(), vc1.getFilters(), "The two lists of filters should be different");
+    }
+
+    @Test
+    public void testFilterCanUseUnmodifyableSet() {
```
lbergelson:
```suggestion
    public void testFilterCanUseUnmodifiableSet() {
```

src/test/java/htsjdk/variant/variantcontext/VariantContextBuilderTest.java
```diff
@@ -47,19 +55,207 @@ public void testAttributeResettingWorks(final boolean leaveModifyableAsIs) {
...
+        return new Object[][]{
+                {"Tab\t"},
+                {"newLine\n"},
+                {"space "},
+                {"semicolon;"},
+                {"carriage return\r"}
+        };
+    }
+
+    @Test(dataProvider = "illegalFilterStrings",expectedExceptions = IllegalStateException.class)
```
lbergelson:
```suggestion
    @Test(dataProvider = "illegalFilterStrings", expectedExceptions = IllegalStateException.class)
```

src/test/java/htsjdk/variant/variantcontext/VariantContextBuilderTest.java
```diff
@@ -47,19 +55,207 @@ public void testAttributeResettingWorks(final boolean leaveModifyableAsIs) {
...
+        final VariantContext result1 = root1.attribute("AC", 1).make();
+        final VariantContextBuilder root2 = builderScheme.getOtherBuilder(root1, result1)
+                .source(snpSource).chr(snpChr).start(snpLocStart).stop(snpLocStop).alleles(Arrays.asList(Aref, C));
+
+        //this is a red-herring and should not change anything.
+        final VariantContext ignored = root1.attribute("AC", 2).make();
         final VariantContext result2 = root2.attribute("AC", 1).make();
 
-       Assert.assertEquals(result1.getAttribute("AC"), result2.getAttribute("AC"));
+        Assert.assertEquals(result1.getAttribute("AC"), result2.getAttribute("AC"));
```
lbergelson: One of the potential issues around the builder is that there can be subtle bugs that relate to the order of instantiation vs alteration.

```
with this initialization:
builder1.attribute("AC", 1)
builder2 = new Builder(builder1)

these two following operations could be different:

case1: 
result1 = builder1.make()
builder2.attribute("AC", 2)
result2= builder2.make()

case2:
builder2.attribute("AC", 2)
result1 = builder1.make()
result2= builder2.make()
```

That could be the case if the builder has an internal state that is accidentally shared/modified but the `make()` operation performs a copy. In both cases I would expect `result1` to have ac of 1 but it's possible that case 1 would be fine but case 2 would be broken.  I don't see a test for that type of difference here.  I think in practice the only place that make() copies anything is with the alleles, so this may be unnecessary / overkill to tests.

src/test/java/htsjdk/variant/variantcontext/VariantContextBuilderTest.java
```diff
@@ -47,19 +55,207 @@ public void testAttributeResettingWorks(final boolean leaveModifyableAsIs) {
...
+
+        final Genotype sample1 = GenotypeBuilder.create("sample1", Arrays.asList(Tref, C));
+        final Genotype sample2 = GenotypeBuilder.create("sample2", Arrays.asList(Tref, G));
+
+        GenotypesContext gc = GenotypesContext.create(sample1);
+        builder.genotypes(gc);
+        try {
+            gc.add(sample2);
+        } catch (IllegalAccessError e) {
+            // nice work...
```
lbergelson: Wow, I didn't realize that this throws `IllegalAccessError`.  That's a subclass of `LinkageError`.  We should absolutely not be throwing or catching that here.   

Could you change it to `IllegalStateException` instead?

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
yfarjoun: well, it's private, and I wanted to be clear that it does less work than the .filters() method.

src/test/java/htsjdk/variant/variantcontext/VariantContextBuilderTest.java
yfarjoun: case 1 is tested already, I believe, but I'll add it explicitly, also case 2 works, but I added a test.

src/test/java/htsjdk/variant/variantcontext/VariantContextBuilderTest.java
yfarjoun: no.. the issue is that the sample map is immutable, so when I try to add a sample to it I get this exception (as I should!)

yfarjoun: oh. I see what you mean. sure.

PR #1354, opened 2019-04-18
- Add a function to calculate the value of the OA tag

yfarjoun:
The OA tag can be used to record the previous ("Original") alignment information in a SAM record. This PR add a method that will construct the contents of the OA tag from a record, to make the creation of the OA tag easy and uniform.

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMUtils.java
+++ b/src/main/java/htsjdk/samtools/SAMUtils.java
@@ -1268,7 +1268,13 @@ public final class SAMUtils {
         return sequence.getSequenceLength() > GenomicIndexUtil.BIN_GENOMIC_SPAN;
     }
 
-    public static String calculateOATagValue(SAMRecord record){
+    /**
+     * Function to create the OA tag value from a record. The OA tag contains the mapping information
+     * of a record encoded as a comma-separated string (REF,POS,STRAND,CIGAR,MAPPING_QUALITY,NM_TAG_VALUE)
+     * @param record to use for generating the OA tag
+     * @return the OA tag string value
+     */
+    public static String calculateOATagValue(SAMRecord record) {
         if (record.getReferenceName().contains(",")) {
             throw new SAMException(String.format("Reference name for record %s contains a comma character.", record.getReadName()));
         }
--- a/src/main/java/htsjdk/samtools/SAMUtils.java
+++ b/src/main/java/htsjdk/samtools/SAMUtils.java
@@ -30,6 +30,7 @@ import htsjdk.samtools.util.CloserUtil;
 import htsjdk.samtools.util.CoordMath;
 import htsjdk.samtools.util.RuntimeEOFException;
 import htsjdk.samtools.util.StringUtil;
+import htsjdk.tribble.annotation.Strand;
 
 import java.io.File;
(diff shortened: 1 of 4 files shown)


src/main/java/htsjdk/samtools/util/SequenceUtil.java
```diff
@@ -34,6 +35,7 @@
 import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Optional;
```
lbergelson: accidental inclusion?

src/main/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -1265,4 +1267,26 @@ public static boolean isValidUnsignedIntegerAttribute(long value) {
     public static boolean isReferenceSequenceIncompatibleWithBAI(final SAMSequenceRecord sequence) {
         return sequence.getSequenceLength() > GenomicIndexUtil.BIN_GENOMIC_SPAN;
     }
+
+    public static String calculateOATagValue(SAMRecord record){
```
lbergelson: can you add a javadoc here?

src/main/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -1265,4 +1267,26 @@ public static boolean isValidUnsignedIntegerAttribute(long value) {
...
+                    record.getReadNegativeStrandFlag() ? Strand.NEGATIVE : Strand.POSITIVE,
+                    record.getMappingQuality());
+        } else {
+            oaValue = String.format("%s,%s,%s,%s,%s,%s",
+                    record.getReferenceName(),
+                    record.getAlignmentStart(),
+                    record.getReadNegativeStrandFlag() ? Strand.NEGATIVE : Strand.POSITIVE,
+                    record.getCigarString(),
+                    record.getMappingQuality(),
+                    Optional.ofNullable(record.getAttribute(SAMTag.NM.name())).orElse(""));
```
lbergelson: Annoyingly this actually creates a new `Optional` object which adds needless overhead.  A ternery expression is actually more performant here I believe.  That said, this overhead is probably negligible and unlikely to ever show up in a profiler ever.  The string formatter probably has a lot more overhead than an extra Optional.

src/test/java/htsjdk/samtools/SAMUtilsTest.java
```diff
@@ -284,4 +286,46 @@ public void testCompressedBasesToBytes() {
...
+    @DataProvider()
+    public Iterator<Object[]> getOAValues(){
+        final SAMRecordSetBuilder builder = new SAMRecordSetBuilder();
+        final List<Object[]> tests = new ArrayList<>();
+
+        {
+            final SAMRecord record = builder.addFrag("test1", 0, 15, false, false, "36M", null, 45);
+
+            tests.add(new Object[]{record, "chr1,15,+,36M,255,"});
+        }
```
lbergelson: Very weird, we don't often open scopes just to reuse variable names, but I don't think there's anything wrong with it here, and it's convenient to reuse the name record...  Maybe we should use this more when doing repetitive construction of things in a dataprovider...

PR #1355, opened 2019-04-20
Removed redundant readability test in unrollPaths method

yfarjoun:
Cloud paths are slow to provide readability. This PR removes a readability check that comes right before opening a file. A failure to read the file will result in an error in both cases.

### DIFF
--- a/src/main/java/htsjdk/samtools/util/IOUtil.java
+++ b/src/main/java/htsjdk/samtools/util/IOUtil.java
@@ -23,7 +23,6 @@
  */
 package htsjdk.samtools.util;
 
-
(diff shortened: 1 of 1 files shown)


src/main/java/htsjdk/samtools/util/IOUtil.java
```diff
@@ -1107,7 +1137,7 @@ public static String slurp(final InputStream is, final Charset charSet) {
                     throw new IllegalArgumentException("had trouble reading from " + p.toUri().toString());
```
lbergelson: Could you add the exception `e` as the cause to this newly thrown exception.

PR #1358, opened 2019-04-25
Added name of offending record to error message in SamPairUtil

yfarjoun:
This PR was inspired by https://github.com/broadinstitute/picard/issues/201 where the error doesn't allow the user to investigate the offending record. By adding the record name to the exception message, the user can investigate more easily and fix the input.

### DIFF
--- a/src/main/java/htsjdk/samtools/SamPairUtil.java
+++ b/src/main/java/htsjdk/samtools/SamPairUtil.java
@@ -437,9 +437,9 @@ public class SamPairUtil {
             // No need to advance if we have records remaining
             if (!records.isEmpty()) return;
 
-            /**
-             * Get all records with the same name, and then identify the canonical first and second end to which we
-             * want to set mate info.
+            /*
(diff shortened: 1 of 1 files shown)


yfarjoun:
Apologies about the CRLF->LF change. please use "hide whitespaces" to view the actual code change.

src/main/java/htsjdk/samtools/SamPairUtil.java
```diff
@@ -1,513 +1,513 @@
...
+
+        /**
+         * Populate this.records if necessary.
+         */
+        private void advance() {
+
+            // No need to advance if we have records remaining
+            if (!records.isEmpty()) return;
+
+            /*
```
pshapiro4broad: It's good to fix this -- it's not a valid javadoc so it shouldn't start with ** -- but somehow the entire file is showing up as changed, as a whitespace change.

Can you make another PR that doesn't show every line as being different? Also, multi-line comments usually have a `*` character at the start of each line. Or, you could use `//` for each line of a multi-line comment.

Issue #1174, opened 2018-08-11
Zero-length read without FZ, CS or CQ tag


nh13:
@tfenne @lbergelson @yfarjoun I wanted to get your thoughts on relaxing this validation, which I think is stricter than the SAM spec.  I have a use case to represent some data in SAM/BAM/CRAM, but with the bases and qualities removed, with an example below with most extraneous information removed (i.e. other tags).  The bases and qualities are `*`, so  no assumptions should be made about how the bases and cigar relate.  There was a bug fixed here: 9d724c3a2dfd6928ea24a91443cd7f2bf6eb6c09.  But I think that requiring FZ/CS/CQ tags when the bases are empty is not warranted (McCowan!):
https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/samtools/SAMRecord.java#L2081-L2109

Example:

<details>

```
htsjdk.samtools.SAMFormatException: Error parsing text SAM file. Zero-length read without FZ, CS or CQ tag; File ...
Line: name	0	chr1	1	60	7M	*	0	0	*	*	RG:Z:Id
   at htsjdk.samtools.SAMLineParser.reportErrorParsingLine(SAMLineParser.java:457)
   at htsjdk.samtools.SAMLineParser.parseLine(SAMLineParser.java:355)
   at htsjdk.samtools.SAMTextReader$RecordIterator.parseLine(SAMTextReader.java:268)
   at htsjdk.samtools.SAMTextReader$RecordIterator.next(SAMTextReader.java:255)
   at htsjdk.samtools.SAMTextReader$RecordIterator.next(SAMTextReader.java:228)
   at htsjdk.samtools.SamReader$AssertingIterator.next(SamReader.java:576)
   at htsjdk.samtools.SamReader$AssertingIterator.next(SamReader.java:548)
```
</details>

yfarjoun:
Do you need this to pass ValidateSamFile, or are you having trouble reading/writing the file?

There has been discussions in HTS-Spec about SAM-valid and SAM-strict specifications and I would tend to agree that this falls in the "sam-strict" side of things, but my sense is that for normal use of BAM this would indicate a problem that I'd like to know about...

tfenne:
I think the problem is that the reader in STRICT mode (which generally you want), doesn't allow reads that are missing bases, qualities _and_ don't have color-space related tags.

[...]

PR #1360, opened 2019-05-01
Remove requirement that zero length reads need color space or flow si…

nh13:
…gnal data.

See https://github.com/samtools/htsjdk/issues/1174

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMRecord.java
+++ b/src/main/java/htsjdk/samtools/SAMRecord.java
@@ -2044,35 +2044,7 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
             ret.addAll(errors);
             if (firstOnly) return ret;
         }
-        // TODO(mccowan): Is this asking "is this the primary alignment"?
-        if (this.getReadLength() == 0 && !this.isSecondaryAlignment()) {
-            final Object fz = getAttribute(SAMTag.FZ.getBinaryTag());
-            if (fz == null) {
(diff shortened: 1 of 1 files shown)


src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2044,35 +2044,7 @@ public int hashCode() {
             ret.addAll(errors);
             if (firstOnly) return ret;
         }
-        // TODO(mccowan): Is this asking "is this the primary alignment"?
```
nh13: @mccowan !

mccowan: Thank you for cleaning up my loose ends!  😁. Hope all is well!

yfarjoun:
I'm not happy about removing this sanity test...

nh13:
@yfarjoun I am not happy about you being not happy.  Lets discuss on the original issue.

yfarjoun:
Well... while I'm not 100% happy about it, I think that if you convert the deletion to a commented code (like all the other skeletons in that method) and add a comment about when it could go back into the code (i.e. when there's a system in place for disabling various errors) I'll relent.... :-)

nh13:
@yfarjoun I don't like leaving dead code, so I guess we both aren't 100% happy, which is a good compromise.  Thank-you!

yfarjoun:
:+1: thanks for compromising.

PR #1362, opened 2019-05-02
Moved loading of sequence dictionary into a separate, overriddeable me…

tfenne:
…thod.

### Description

In rare cases it may be desirable to either not load the sequence dictionary for a reference, or to customize how it's loaded.  The change here simply moves some code out of the constructor into a protected method that can be easily overridden to customize the behaviour.
### Checklist

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
+++ b/src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
@@ -78,17 +78,14 @@ abstract class AbstractFastaSequenceFile implements ReferenceSequenceFile {
     /** Attempts to find and load the sequence dictionary if present. */
     protected SAMSequenceDictionary findAndLoadSequenceDictionary(final Path fasta) {
         final Path dictPath = findSequenceDictionary(path);
-        if (dictPath == null) {
-            return null;
+        if (dictPath == null) return null;
+
+        IOUtil.assertFileIsReadable(dictPath);
+        try (InputStream dictionaryIn = IOUtil.openFileForReading(dictPath)) {
+            return ReferenceSequenceFileFactory.loadDictionary(dictionaryIn);
         }
-        else {
-            IOUtil.assertFileIsReadable(dictPath);
-            try (InputStream dictionaryIn = IOUtil.openFileForReading(dictPath)) {
-                return ReferenceSequenceFileFactory.loadDictionary(dictionaryIn);
-            }
-            catch (Exception e) {
-                throw new SAMException("Could not open sequence dictionary file: " + dictPath, e);
-            }
+        catch (Exception e) {
+            throw new SAMException("Could not open sequence dictionary file: " + dictPath, e);
         }
     }
 
--- a/src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
+++ b/src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
@@ -60,17 +60,7 @@ abstract class AbstractFastaSequenceFile implements ReferenceSequenceFile {
     AbstractFastaSequenceFile(final Path path) {
         this.path = path;
         this.source = path == null ? "unknown" : path.toAbsolutePath().toString();
-        final Path dictionary = findSequenceDictionary(path);
-
-        if (dictionary != null) {
-            IOUtil.assertFileIsReadable(dictionary);
-            try (InputStream dictionaryIn = Files.newInputStream(dictionary)) {
-                this.sequenceDictionary = ReferenceSequenceFileFactory.loadDictionary(dictionaryIn);
-            }
-            catch (Exception e) {
-                throw new SAMException("Could not open sequence dictionary file: " + dictionary, e);
-            }
-        }
+        this.sequenceDictionary = findAndLoadSequenceDictionary(path);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
```diff
@@ -60,17 +60,7 @@
...
-        if (dictionary != null) {
-            IOUtil.assertFileIsReadable(dictionary);
-            try (InputStream dictionaryIn = Files.newInputStream(dictionary)) {
-                this.sequenceDictionary = ReferenceSequenceFileFactory.loadDictionary(dictionaryIn);
-            }
-            catch (Exception e) {
-                throw new SAMException("Could not open sequence dictionary file: " + dictionary, e);
-            }
-        }
+        this.sequenceDictionary = findAndLoadSequenceDictionary(path);
```
pshapiro4broad: Putting a method call to an overridable method in a constructor means that you can end up calling a method in an object whose constructor hasn't been called yet. This can lead to some very hard to debug errors.

One way to avoid this kind of dependency is to make a field lazily initialized, e.g. in its getter.

src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
```diff
@@ -85,29 +75,41 @@
...
+        if (dictPath == null) {
             return null;
         }
-        return dictionary.toFile();
+        else {
+            IOUtil.assertFileIsReadable(dictPath);
+            try (InputStream dictionaryIn = IOUtil.openFileForReading(dictPath)) {
+                return ReferenceSequenceFileFactory.loadDictionary(dictionaryIn);
+            }
+            catch (Exception e) {
```
pshapiro4broad: `catch` should be on previous line

src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
```diff
@@ -85,29 +75,41 @@
...
-        final Path dictionary = findSequenceDictionary(IOUtil.toPath(file));
-        if (dictionary == null) {
+    /** Attempts to find and load the sequence dictionary if present. */
+    protected SAMSequenceDictionary findAndLoadSequenceDictionary(final Path fasta) {
+        final Path dictPath = findSequenceDictionary(path);
+        if (dictPath == null) {
             return null;
         }
-        return dictionary.toFile();
+        else {
```
pshapiro4broad: `else` unnecessary after `return`

src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
tfenne: My first thought was indeed to make this lazy loading, but that's not possible without a breaking change.  Unfortunately we have:

```
    protected SAMSequenceDictionary sequenceDictionary;
```

and subclasses (including any outside htsjdk) can access via property access.  If others (@yfarjoun , @lbergelson) are happy to have a breaking change, I'd be happy to rework this PR to make that a `private Lazy<SAMSequenceDictionary>` and move access to be via a method.  Otherwise I'm sorta stuck doing it in the manner I have.

src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
tfenne: Probably?  I was just moving this code around, not actually changing it.

lbergelson:
@tfenne Maybe you could motivate this use case a little more?  It's not clear to me what the use case is that you can't use the alternate constructor that explicitly lets you specify the sequence dictionary.  Are you trying to modify the loading behavior of some subclass that doesn't expose a path to that constructor cleanly?

tfenne:
@lbergelson I'm really trying to avoid loading the dictionary at all.  In my case I've been handed a reference that has millions of sequences (someone has basically created ~100bp "chromosomes") from hg19 with each possible combination of variants from 1KG inserted into them.  So the dictionary is massive - and since I just want to iterate through the reference sequences, I don't need the dictionary and loading it is sucking up gigs of memory and lots of time.  Given the class hierarchy, and use of `ReferenceSequenceFileFactory` I could not see another sensible way to prevent loading of the dictionary.

I think the ideal would be to do as @pshapiro4broad suggests and make the dictionary lazily loaded, if you are on board with changing `sequenceDictionary` from `protected` to `private`.  Thoughts?

src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
yfarjoun: sounds good to me.

tfenne: Ok, I've pushed up a change that makes it lazy.  Fixed any problems in htsjdk, and double checked that there aren't classes in Picard or GATK that access that property directly.  @lbergelson are you on board with this?

src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
```diff
@@ -82,32 +73,41 @@
...
 
-    protected static Path findSequenceDictionary(final Path path) {
-        if (path == null) {
+    /** Attempts to locate the sequence dictionary file adjacent to the reference fasta file. */
+    protected static Path findSequenceDictionary(final Path fastaPath) {
+        if (fastaPath == null) {
             return null;
         }
         // Try and locate the dictionary with the default method
-        final Path dictionary = ReferenceSequenceFileFactory.getDefaultDictionaryForReferenceSequence(path); path.toAbsolutePath();
```
lbergelson: weird pointless code there you discovered....

src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
```diff
@@ -82,32 +73,41 @@
...
-        final Path dictionary = findSequenceDictionary(IOUtil.toPath(file));
-        if (dictionary == null) {
-            return null;
+    /** Attempts to find and load the sequence dictionary if present. */
+    protected SAMSequenceDictionary findAndLoadSequenceDictionary(final Path fasta) {
+        final Path dictPath = findSequenceDictionary(path);
+        if (dictPath == null) return null;
+
+        IOUtil.assertFileIsReadable(dictPath);
+        try (InputStream dictionaryIn = IOUtil.openFileForReading(dictPath)) {
```
lbergelson: Is this change to support dict.gz files?  Should findSequenceDictionary below try to locate them as well?

src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
```diff
@@ -82,32 +73,41 @@
     AbstractFastaSequenceFile(final Path path, final String source, final SAMSequenceDictionary sequenceDictionary) {
         this.path = path;
         this.source = source;
-        this.sequenceDictionary = sequenceDictionary;
+        this.dictionary = new Lazy<>(() -> sequenceDictionary);
     }
 
-    protected static File findSequenceDictionary(final File file) {
```
lbergelson: Getting rid of the file version of this?  I'm fine with that... but we should make note of it.

src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
tfenne: Yeah.  +1 for IntelliJ graying out unused code.

src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
tfenne: Not really, although I suppose it now could.  I just like for every file/path open to vector through IOUtil since it standardizes how things are done and how buffering is applied.

src/main/java/htsjdk/samtools/reference/AbstractFastaSequenceFile.java
tfenne: It's mostly because the constructor now calls `findAndLoadSequenceDictionary(final Path fasta)` rather than just `findSequenceDictionary()` so it seemed kinda pointless.  I'll restore it for now and add deprecation to it so we can remove it later.

tfenne:
@lbergelson I was thinking that I could do as you suggest, but the ReferenceSequenceFactory doesn't have methods that take in sequence dictionaries, and I didn't want a) to proliferate a whole new set of methods that take a dictionary too, and b) it seemed like a lot of work.  I could instantiate a `FastaSequenceFile` directly, and I may still end up doing that, but that seemed kinda dirty and also it is still a little hard to determine if/when the dictionary is accessed on that path.

tfenne:
I should also mention that I suspect GATK requires an fasta index, and if one is present then a subclass of `AbstractIndexedFastaFile` is created, and in the constructor for that a call is made to validate the dictionary matches the index, which will cause the dictionary to be loaded, so I think for GATK the delay is going to be measured in ms or ns!

PR #1363, opened 2019-05-02
Add support for the TP (topology) tag in the sequence dictionary

nh13:
See https://github.com/samtools/hts-specs/pull/405

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMRecord.java
+++ b/src/main/java/htsjdk/samtools/SAMRecord.java
@@ -2300,7 +2300,7 @@ public class SAMRecord implements HtsRecord, Cloneable, Locatable, Serializable 
                     ret.add(new SAMValidationError(SAMValidationError.Type.INVALID_REFERENCE_INDEX, buildMessage("Reference sequence not found in sequence dictionary.", isMate), getReadName()));
                     if (firstOnly) return ret;
                 } else {
-                    if (alignmentStart > sequence.getSequenceLength()) {
+                    if (alignmentStart > sequence.getSequenceLength() && !sequence.isCircular()) {
                         if (ret == null) ret = new ArrayList<>();
                         ret.add(new SAMValidationError(SAMValidationError.Type.INVALID_ALIGNMENT_START, buildMessage("Alignment start (" + alignmentStart + ") must be <= reference sequence length (" +
                                 sequence.getSequenceLength() + ") on reference " + sequence.getSequenceName(), isMate), getReadName()));
--- a/src/main/java/htsjdk/samtools/SAMSequenceRecord.java
+++ b/src/main/java/htsjdk/samtools/SAMSequenceRecord.java
@@ -58,6 +58,14 @@ public class SAMSequenceRecord extends AbstractSAMHeaderRecord implements Clonea
     public static final String URI_TAG = "UR";
     public static final String SPECIES_TAG = "SP";
     public static final String DESCRIPTION_TAG = "DS";
+    public static final String TOPOLOGY_TAG = "TP";
+
+    /**
+     * Topology values.
+     */
+    public enum Topology {
+        linear, circular
+    }
 
     /**
      * If one sequence has this length, and another sequence had a different length, isSameSequence will
@@ -80,7 +88,7 @@ public class SAMSequenceRecord extends AbstractSAMHeaderRecord implements Clonea
      * The standard tags are stored in text header without type information, because the type of these tags is known.
      */
     public static final Set<String> STANDARD_TAGS =
-            new HashSet<>(Arrays.asList(SEQUENCE_NAME_TAG, SEQUENCE_LENGTH_TAG, ASSEMBLY_TAG, ALTERNATIVE_SEQUENCE_NAME_TAG, MD5_TAG, URI_TAG, SPECIES_TAG));
+            new HashSet<>(Arrays.asList(SEQUENCE_NAME_TAG, SEQUENCE_LENGTH_TAG, ASSEMBLY_TAG, ALTERNATIVE_SEQUENCE_NAME_TAG, MD5_TAG, URI_TAG, SPECIES_TAG, TOPOLOGY_TAG));
 
     // These are the chars matched by \\s.
     private static final char[] WHITESPACE_CHARS = {' ', '\t', '\n', '\013', '\f', '\r'}; // \013 is vertical tab
@@ -155,6 +163,20 @@ public class SAMSequenceRecord extends AbstractSAMHeaderRecord implements Clonea
     public SAMSequenceRecord setDescription(final String value) {
         setAttribute(DESCRIPTION_TAG, value);
         return this;
+    }
+
+    public Topology getTopology() {
+        final String value = getAttribute(TOPOLOGY_TAG);
+        return (value == null) ? null : Topology.valueOf(value);
+    }
+
+    public SAMSequenceRecord setTopology(final Topology value) {
+        setAttribute(TOPOLOGY_TAG, (value == null) ? null : value.name());
+        return this;
+    }
+
+    public boolean isCircular() {
+        return getTopology() == Topology.circular;
     }
 
     /**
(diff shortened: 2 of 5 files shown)


cmnbroad:
Holding off until samtools/hts-specs#405 is merged.

jmarshall:
From the commit message:

> as well as if the start (or mate start) is off the end.

Um… not sure anyone has made a case for allowing this?

nh13:
@jmarshall 
> Um… not sure anyone has made a case for allowing this?
That's definitely my mistake.  I think what I _want_ is to say that it's ok for the mate's start to be past the end of the reference as long as the current start is not.  So we can have a template span the origin, but not have a template that goes past it.  If that makes sense, I'll fix the implementation.

yfarjoun (changes requested):

yfarjoun:
I suggest the non-controversial parts be extracted into a separate PR.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2161,7 +2161,7 @@ protected void setFileSource(final SAMFileSource fileSource) {
                     ret.add(new SAMValidationError(SAMValidationError.Type.INVALID_REFERENCE_INDEX, buildMessage("Reference sequence not found in sequence dictionary.", isMate), getReadName()));
                     if (firstOnly) return ret;
                 } else {
-                    if (alignmentStart > sequence.getSequenceLength()) {
+                    if (alignmentStart > sequence.getSequenceLength() && sequence.getTopology() != SAMSequenceRecord.Topology.circular) {
```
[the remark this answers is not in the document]
nh13: I'll defer changing this until we agree on the representation on https://github.com/samtools/hts-specs/pull/405.  Note that this does force alignments to have `1 <= POS <= REFLEN`, even though that hasn't been decided upon by the spec.

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -113,8 +121,14 @@ public SAMSequenceRecord(final String name, final int sequenceLength) {
     public String getMd5() { return (String) getAttribute(MD5_TAG); }
     public void setMd5(final String value) { setAttribute(MD5_TAG, value); }
 
-    public String getDescription() { return getAttribute(DESCRIPTION_TAG);}
-    public void setDescription(final String value) { setAttribute(DESCRIPTION_TAG, value);}
+    public String getDescription() { return getAttribute(DESCRIPTION_TAG); }
+    public void setDescription(final String value) { setAttribute(DESCRIPTION_TAG, value); }
+
+    public Topology getTopology() {
```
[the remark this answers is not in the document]
nh13: Done

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -113,8 +121,14 @@ public SAMSequenceRecord(final String name, final int sequenceLength) {
...
+        final String value = getAttribute(TOPOLOGY_TAG);
+        return (value == null) ? null : Topology.valueOf(value);
+    }
+    public void setTopology(final Topology value) { setAttribute(TOPOLOGY_TAG, (value == null) ? null : value.name()); }
```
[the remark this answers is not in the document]
nh13: Look at the lines before it, they are all one-liners, and there are methods are grouped by topic.  Leaving as is.

nh13 (commented):
@yfarjoun back to you

yfarjoun (approved):
Thanks!

yfarjoun:
note that testLongReferenceQuery is failing in Travis.

yfarjoun (changes requested):
tests need to pass before actually approved...

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2172,12 +2172,10 @@ protected void setFileSource(final SAMFileSource fileSource) {
                     ret.add(new SAMValidationError(SAMValidationError.Type.INVALID_REFERENCE_INDEX, buildMessage("Reference sequence not found in sequence dictionary.", isMate), getReadName()));
                     if (firstOnly) return ret;
                 } else {
-                    if (alignmentStart > sequence.getSequenceLength()) {
```
yfarjoun: why did you get rid of this if statement?

src/main/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -943,13 +943,16 @@ public static int getMateUnclippedEnd(final SAMRecord rec) {
...
                 final SAMSequenceRecord sequence = samHeader.getSequence(referenceIndex);
-                final int referenceSequenceLength = sequence.getSequenceLength();
-                for (final AlignmentBlock alignmentBlock : alignmentBlocks) {
-                    if (alignmentBlock.getReferenceStart() + alignmentBlock.getLength() - 1 > referenceSequenceLength) {
-                        if (ret == null) ret = new ArrayList<>();
-                        ret.add(new SAMValidationError(SAMValidationError.Type.CIGAR_MAPS_OFF_REFERENCE,
-                                cigarTypeName + " M operator maps off end of reference", rec.getReadName(), recordNumber));
-                        break;
+                final SAMSequenceRecord.Topology topology = sequence.getTopology();
+                if (topology != SAMSequenceRecord.Topology.circular) {
```
yfarjoun: please use !isCircular() here.

src/main/java/htsjdk/samtools/SAMRecord.java
```diff
@@ -2300,7 +2300,7 @@ private List<SAMValidationError> isValidReferenceIndexAndPosition(final Integer
...
+                    if (alignmentStart > sequence.getSequenceLength() && !sequence.isCircular()) {
```
yfarjoun, line 2303: My understanding of the spec is that POS must still be between 1 and LEN, so why the allowance for greater than LEN in the circular case?

src/test/java/htsjdk/samtools/ValidateSamFileTest.java
```diff
@@ -150,29 +150,32 @@ public void testUnpairedRecords() throws IOException {
...
-        for (int i = 0; i < 5; i++) {
+        final SAMSequenceDictionary dict = samBuilder.getHeader().getSequenceDictionary();
+        for (int i = 0; i < 6; i++) {
             samBuilder.addPair(String.valueOf(i), i, i, i + 100);
+            dict.getSequence(i).setTopology(topology);
         }
         final Iterator<SAMRecord> records = samBuilder.iterator();
         records.next().setMateReferenceName("*");
-        records.next().setMateAlignmentStart(Integer.MAX_VALUE);
+        records.next().setMateAlignmentStart(Integer.MAX_VALUE); // mate start off the end
```
yfarjoun: My understanding of the spec is that POS≤LEN must still be upheld, so this doesn't seem right.

src/test/java/htsjdk/samtools/ValidateSamFileTest.java
```diff
@@ -150,29 +150,32 @@ public void testUnpairedRecords() throws IOException {
...
         records.next().setMateReferenceIndex(records.next().getReferenceIndex() + 1);
         records.next().setMateUnmappedFlag(!records.next().getReadUnmappedFlag());
-
+        final int start = dict.getSequence(records.next().getContig()).getSequenceLength() + 1;
+        records.next().setAlignmentStart(start); // rec start off the end
 
         final Histogram<String> results = executeValidation(samBuilder.getSamReader(), null, IndexValidationStringency.EXHAUSTIVE);
 
-        Assert.assertEquals(results.get(SAMValidationError.Type.INVALID_ALIGNMENT_START.getHistogramString()).getValue(), 3.0);
+        Assert.assertEquals(results.get(SAMValidationError.Type.INVALID_ALIGNMENT_START.getHistogramString()).getValue(), (topology == SAMSequenceRecord.Topology.circular) ? 2.0 : 4.0);
```
yfarjoun, line 174: My understanding of the spec is that POS≤LEN must still be upheld, so this doesn't seem right, as there shouldn't be any alignements with POS>LEN

yfarjoun (changes requested):
I think POS≤LEN is still a requirement in `TP:circular`

PR #1364, opened 2019-05-08
Tolerate lower-case nans in QUAL

karenfeng:
### Description

We sometimes see VCFs with lower-case nan's in QUAL. This doesn't match the Java-style NaN, and therefore causes VCF parsing to break.

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


yfarjoun:
does the spec allow for Nan in the QUAL field?

karenfeng:
VCF reading doesn't break on `NaN` in the `QUAL` or `GQ` fields, but I'm not completely confident on how spec-compliant that is. The spec seems to suggest that because `QUAL` is a Float, it should be ok; the reason `GQ` doesn't break on `NaN` today is because it's parsed via `Double` rather than `Integer`.

jmarshall:
The 4.3 spec is clear that NaNs are valid in QUAL (§1.6.1 states QUAL is a Float; §1.3 says VCF Floats include NaN). The earlier specs don't have §1.3 and appear to be silent on what text a Float might allow.

See also `NaN` vs. `nan` discussions in https://github.com/samtools/bcftools/issues/755#issuecomment-455240246 and https://github.com/samtools/hts-specs/issues/89#issuecomment-455503658 and nearby.

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -760,6 +761,17 @@ private static boolean isVCFStream(final InputStream stream, final String MAGIC_
         return values;
     }
 
+    /**
+     * Parses a String as a Double, being tolerant for lower-case NaN (nan).
+     */
+    private static final Double decodeDouble(final String string) {
+        if (Pattern.matches("[+-]?nan", string)) {
```
yfarjoun: I don't see -+ nans defined anywhere, only "nan" without a sign...have you seen that in the wild? -+Infinity I have seen...but not NaN.

Could you please:
1. first attempt to call `Double.valueOf(string)` and  catch the exception if it fails (for performance)
2. `compile` the pattern into a static final (also for performance)
3.  make the pattern match case-independent `Pattern.compile(regex, Pattern.CASE_INSENSITIVE)` and remove the `[+-]` part.
4. also add checks (and tests) for `[-+]?inf(inity)?` and encode into either Double.POSITIVE_INFINITY or Double.NEGATIVE_INFINITY

Thanks!

jmarshall: NaNs are NaNs and their sign is surely meaningless, but nonetheless the IEEE NaN format has a sign bit and C `printf` will print it (in glibc and other implementations, but not the BSD/macOS one — though IMHO that's not standard compliant). And Java will [parse it](https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#valueOf-java.lang.String-).

So IMHO you should keep the `[+-]?` part. My suggestion for what the VCF spec should allow for NaN/∞ is samtools/hts-specs#409.

lbergelson: I'd like to object to allowing `∞`...

lbergelson: Some additional comments:

1.  Replying to @yfarjoun point 1:  Usually I'm not a fan of catching exceptions as part of the standard flow of a method, but in this case I agree that  it makes sense for performance reasons.
2.  This should be a replacement for `Double.parseDouble` and return a primitive `double` instead of the boxed `Double`

src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java
```diff
@@ -760,6 +761,17 @@ private static boolean isVCFStream(final InputStream stream, final String MAGIC_
         return values;
     }
 
+    /**
+     * Parses a String as a Double, being tolerant for lower-case NaN (nan).
+     */
+    private static final Double decodeDouble(final String string) {
```
lbergelson: I would expose this somewhere public and maybe rename it to be clearer how it's different from the standard methods.  Maybe `VCFUtils.parseDoubleAccordingToVcfSpec`.

lbergelson: There are a number of additional places that this should be used as well I believe:

`GenotypeLikelihoods.parseDeprecatedGLString` line 270 parseDouble`
 `CommonInfo` 299 and 323 that use `Double.valueOf`
`VariantContext` 1632
`Genotype` 529

karenfeng:
@lbergelson Thank you for the feedback; let me know if there any further changes you would like me to make.

src/main/java/htsjdk/variant/vcf/VCFUtils.java
```diff
@@ -228,6 +229,17 @@ else if (vcfFile.getAbsolutePath().endsWith(IOUtil.COMPRESSED_VCF_FILE_EXTENSION
...
     }
 
+    /**
+     * Parses a String as a Double, being tolerant for lower-case NaN (nan).
+     */
+    public static double parseDoubleAccordingToVcfSpec(final String str) {
+        try {
+            return Double.valueOf(str);
+        } catch (NumberFormatException e) {
+            return Double.valueOf(str.replaceAll("[nN][aA][nN]", "NaN"));
```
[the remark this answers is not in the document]
yfarjoun: Though there's a PR (https://github.com/samtools/hts-specs/pulls) to allow all sorts of craziness...I think that a compiled regular expression would be better than using replaceAll.

something like 
```java
//in class
 Pattern infOrNanPattern = Pattern
                .compile("^(sign:[-+])?(inf:(INF|INFINITY))|(nan:NAN)$",
                        Pattern.CASE_INSENSITIVE);

//in method
        final Matcher matcher = infOrNanPattern.matcher(str);
        if (matcher.matches()) {
            final double ret;
            if (matcher.group("inf") == null) {
                ret = sign + Double.NaN;
            } else {
                if (matcher.group("sign").equals("-")) {
                    ret = Double.NEGATIVE_INFINITY;
                } else {
                    ret = Double.POSITIVE_INFINITY;
                }
            }
            return ret;
        }

```

src/test/resources/htsjdk/variant/test_withNanQual.vcf
```diff
@@ -0,0 +1,8 @@
+##fileformat=VCFv4.1
+##FORMAT=<ID=GQ,Number=1,Type=Integer,Description="Genotype Quality">
+##contig=<ID=chr1,length=14640000>
+#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	NA12878
+chr1	1	.	C	T	NaN	.	.	GT:GQ	./.:NaN
+chr1	2	.	C	T	nan	.	.	GT:GQ	./.:nan
+chr1	3	.	C	T	-nan	.	.	GT:GQ	./.:-nan
+chr1	4	.	C	T	+nan	.	.	GT:GQ	./.:+nan
```
yfarjoun: While were' at it, please also test inf and infinity and the various capitalizations....

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1629,7 +1629,7 @@ private final Object decodeOne(final String field, final String string, final VC
                         return b;
                     case String:    return string;
                     case Integer:   return Integer.valueOf(string);
-                    case Float:     return Double.valueOf(string);
+                    case Float:     return Double.parseDouble(string);
```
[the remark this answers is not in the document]
alecw: Yes, the change @yfarjoun suggests is needed to address https://github.com/samtools/htsjdk/pull/1383

src/main/java/htsjdk/variant/vcf/VCFUtils.java
```diff
@@ -37,10 +37,14 @@
 import java.io.File;
 import java.io.IOException;
 import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 public class VCFUtils {
 
+    private static Pattern infOrNanPattern = Pattern.compile("^(?<sign>[-+]?)((?<inf>(INF|INFINITY))|(?<nan>NAN))$", Pattern.CASE_INSENSITIVE);
```
pshapiro4broad: this should be `final`, and should have an uppercase name since it's a static, e.g.
```suggestion
    private static final Pattern INF_OR_NAN_PATTERN = Pattern.compile("^(?<sign>[-+]?)((?<inf>(INF|INFINITY))|(?<nan>NAN))$", Pattern.CASE_INSENSITIVE);
```

src/main/java/htsjdk/variant/vcf/VCFUtils.java
```diff
@@ -250,6 +254,32 @@ else if (vcfFile.getAbsolutePath().endsWith(IOUtil.COMPRESSED_VCF_FILE_EXTENSION
         return output;
     }
 
+    /**
+     * Parses a String as a Double, being tolerant for lower-case NaN (nan).
+     */
+    public static double parseDoubleAccordingToVcfSpec(final String str) {
```
pshapiro4broad: I realize I'm late to this PR, but how about `parseVcfDouble()`? Saying "according to spec" seems redundant to me, unless we need to distinguish this from a VCF double parsing that's not done according to a spec.

src/main/java/htsjdk/variant/vcf/VCFUtils.java
```diff
@@ -250,6 +254,32 @@ else if (vcfFile.getAbsolutePath().endsWith(IOUtil.COMPRESSED_VCF_FILE_EXTENSION
         return output;
     }
 
+    /**
+     * Parses a String as a Double, being tolerant for lower-case NaN (nan).
```
pshapiro4broad: This javadoc doesn't seem sufficient given that this supports INF and NaN. It would be nice to document its behavior more completely, and/or provide a link to part of the spec that this code is implementing.

src/main/java/htsjdk/variant/vcf/VCFUtils.java
```diff
@@ -250,6 +254,32 @@ else if (vcfFile.getAbsolutePath().endsWith(IOUtil.COMPRESSED_VCF_FILE_EXTENSION
...
+                    ret = Double.NaN;
+                } else {
+                    if (matcher.group("sign").equals("-")) {
+                        ret = Double.NEGATIVE_INFINITY;
+                    } else {
+                        ret = Double.POSITIVE_INFINITY;
+                    }
+                }
+                return ret;
+            } else {
```
pshapiro4broad: unnecessary `else` after `return`

src/main/java/htsjdk/variant/vcf/VCFUtils.java
```diff
@@ -250,6 +254,32 @@ else if (vcfFile.getAbsolutePath().endsWith(IOUtil.COMPRESSED_VCF_FILE_EXTENSION
...
+            return Double.parseDouble(str);
+        } catch (NumberFormatException e) {
+            final Matcher matcher = infOrNanPattern.matcher(str);
+            if (matcher.matches()) {
+                final double ret;
+                if (matcher.group("inf") == null) {
+                    ret = Double.NaN;
+                } else {
+                    if (matcher.group("sign").equals("-")) {
+                        ret = Double.NEGATIVE_INFINITY;
```
pshapiro4broad: The test below appears to test this case but it's showing up as not covered. Are you sure that this code path is being tested by the code below?

src/test/java/htsjdk/variant/vcf/AbstractVCFCodecTest.java
```diff
@@ -66,4 +67,30 @@ public void testGLnotOverridePL() {
...
+    public Object[][] getCaseIntolerantDoubles() {
+        return new Object[][]{
+                {"src/test/resources/htsjdk/variant/test_withNanQual.vcf", Double.NaN},
+                {"src/test/resources/htsjdk/variant/test_withPosInfQual.vcf", Double.POSITIVE_INFINITY},
+                {"src/test/resources/htsjdk/variant/test_withNegInfQual.vcf", Double.NEGATIVE_INFINITY},
+        };
+    }
+
+    @Test(dataProvider = "caseIntolerantDoubles")
+    public void testCaseIntolerantDoubles(String vcfInput, Double value) {
```
pshapiro4broad: unnecessary to use a boxed type here as `value` is never `null`
```suggestion
    public void testCaseIntolerantDoubles(String vcfInput, double value) {
```

src/main/java/htsjdk/variant/vcf/VCFUtils.java
```diff
@@ -250,6 +254,32 @@ else if (vcfFile.getAbsolutePath().endsWith(IOUtil.COMPRESSED_VCF_FILE_EXTENSION
...
+                    } else {
+                        ret = Double.POSITIVE_INFINITY;
+                    }
+                }
+                return ret;
+            } else {
+                throw e;
```
pshapiro4broad: Can you add a unit test to `VCFUtilsTest` that verifies this negative case?

alecw:
Hi @lbergelson , is a release with this change coming any time soon?  We would really appreciate it.

Thanks, Alec

lbergelson:
Yes.  I've wanted to do a release for a while but I've been busy with personal stuff (mostly baby related) and because of incompatibilities I've had trouble running the downstream tests successfully so it's taken longer than intended.  I hope to have one out this week but next week at the latest.  I can publish a release candidate that hasn't been fully tested if that would be helpful.

PR #1366, opened 2019-05-13
SubstitutionMatrix class cleanup - part 1.

cmnbroad:
This is mostly comments documenting what this code is doing, and some name changes, but there are no functional changes. A separate follow-on PR will pick up where this one ends.

### DIFF
--- a/src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
+++ b/src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
@@ -34,17 +34,18 @@ import java.util.Comparator;
 public class SubstitutionMatrix {
     // List of bases subject to substitution. The upper and lower case arrays must have the bases
     // in the same order so that the index for a given base is the same for upper and lower case
-    public static final byte[] SUBSTITUTION_BASES_UPPER = new byte[]{'A', 'C', 'G', 'T', 'N'};
+    static final byte[] SUBSTITUTION_BASES_UPPER = new byte[]{'A', 'C', 'G', 'T', 'N'};
+
     private static final byte[] SUBSTITUTION_BASES_LOWER = new byte[]{'a', 'c', 'g', 't', 'n'};
-    private static int BASES_SIZE = SUBSTITUTION_BASES_UPPER.length;
+    private static final int BASES_SIZE = SUBSTITUTION_BASES_UPPER.length;
 
     // Since bases are represented as bytes, there are theoretically 256 possible symbols in the
     // symbol space, though in reality we only care about 10 of these (5 upper and 5 lower case
     // reference bases).
-    private static int SYMBOL_SPACE_SIZE = 256;
+    private static final int SYMBOL_SPACE_SIZE = 256;
 
     // number of possible substitution codes per base
-    private static int CODES_PER_BASE = BASES_SIZE - 1;
+    private static final int CODES_PER_BASE = BASES_SIZE - 1;
 
     // The order in which substitution codes are encoded in the serialized matrix, as prescribed
     // by the CRAM spec:
--- a/src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
+++ b/src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
@@ -20,38 +20,86 @@ package htsjdk.samtools.cram.structure;
 import java.util.Arrays;
 import java.util.Comparator;
 
+/**
+ * Substitution matrix, used to represent base substitutions for reference-based CRAM
+ * compression.
+ *
+ * The matrix is stored internally in two forms; the packed/encoded form used for serialization,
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -20,160 +20,251 @@
...
+ * This implementation substitutes both upper and lower case versions of a give base with the
+ * same (upper case) substitute base.
+ */
 public class SubstitutionMatrix {
-    public static final byte[] BASES = new byte[]{'A', 'C', 'G', 'T', 'N'};
-    private static final byte[] BASES_LC = new byte[]{'a', 'c', 'g', 't', 'n'};
-    private static final byte[] ORDER;
+    // List of bases subject to substitution. The upper and lower case arrays must have the bases
+    // in the same order so that the index for a given base is the same for upper and lower case
+    public static final byte[] SUBSTITUTION_BASES_UPPER = new byte[]{'A', 'C', 'G', 'T', 'N'};
```
lbergelson: This continues to be unsafe and may be rewritten by the caller.  Is there a strong reason that this one is public and the others aren't?

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -20,160 +20,251 @@
...
+    private static final byte[] SUBSTITUTION_BASES_LOWER = new byte[]{'a', 'c', 'g', 't', 'n'};
+    private static int BASES_SIZE = SUBSTITUTION_BASES_UPPER.length;
```
lbergelson: This should be final.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -20,160 +20,251 @@
...
+
+    // Since bases are represented as bytes, there are theoretically 256 possible symbols in the
+    // symbol space, though in reality we only care about 10 of these (5 upper and 5 lower case
+    // reference bases).
+    private static int SYMBOL_SPACE_SIZE = 256;
```
lbergelson: final

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -20,160 +20,251 @@
...
 
+    // number of possible substitution codes per base
+    private static int CODES_PER_BASE = BASES_SIZE - 1;
```
lbergelson: final

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -20,160 +20,251 @@
...
+    private static final byte[] CODE_ORDER;
     static {
-        ORDER = new byte[255];
-        Arrays.fill(ORDER, (byte) -1);
-        ORDER['A'] = 0;
-        ORDER['C'] = 1;
-        ORDER['G'] = 2;
-        ORDER['T'] = 3;
-        ORDER['N'] = 4;
+        CODE_ORDER = new byte[SYMBOL_SPACE_SIZE];
```
lbergelson: This whole thing is very fragile if we ever change the bases (which we probably don't...) it could be rewritten to actually use the values  and indexes from the `SUBSTITUTION_BASES_UPPER` automatically.  That might be less clear though...

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
lbergelson: Maybe make this package protected if it needs to be used by tests.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
cmnbroad: Yes, I plan to do exactly that in the next PR .

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
cmnbroad: Yes, it should be package protected.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
cmnbroad: Done.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
cmnbroad: Done.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
cmnbroad: Done.

PR #1369, opened 2019-05-17
add documention for VariantContext.getStart() regarding telomeric events

SHuang-Broad:
### Description

The VCF spec allows `POS` column to take value 0 (or chromosome length + 1), when the event is at a telomere.

Currently the documentation for `public int VariantContext.getStart()` claims it returns a 1-based value, where it seems that 0 is an invalid value.

This PR intends to clarify that, by adding documentation.

### Checklist

(Documentation change, most of the following does not apply)

- [x] Code compiles correctly
- [x] All tests passing

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/VariantContext.java
+++ b/src/main/java/htsjdk/variant/variantcontext/VariantContext.java
@@ -1663,20 +1663,34 @@ public class VariantContext implements Feature, Serializable {
     }
 
     /**
-     * @return 1-based inclusive start position of the Variant
-     * INDEL events usually start on the first unaltered reference base before the INDEL.
+     * Returns 1-based inclusive start position of the variant, 0 or greater.
+     * See below for explanation on "0".
      *
-     * Note also that for the VCF spec allows 0 and N + 1 for POS field, where N is the length of the chromosome,
-     * for telomeric event.
+     * <p>
+     *     <strong>Warning:</strong>
+     *     be aware that the start position of the VariantContext is defined
+     *     in terms of the start position specified in the underlying vcf file,
+     *     VariantContexts representing the same biological event may have different
+     *     start positions depending on the specifics of the vcf file they are derived from.
+     * </p>
+     *
+     * <p>
+     *     INDEL events usually start on the first unaltered reference base before the INDEL.
+     * </p>
+     *
+     * <p>
+     *     Note also that the VCF spec allows 0 and N + 1 for POS field for telomeric event,
+     *     where N is the length of the chromosome.
+     *     The "0" value returned should be interpreted as telomere, and does not violate the above "1-based" comment.
+     *     It's the responsibility of code generating such variants to make sure {@code start} is populated correctly.
+     *     And code consuming the returned {@code start} should be prepared for such out-of-the-ordinary values.
+     * </p>
+     *
      * The given example is in section 5.4.5 (see example event illustrated in Figure 6 and VCF records below the figure).
      * Another possible scenario is when a telomere on the p-arm of a chromosome is deleted.
-     * In these cases, the "0" value returned should be interpreted as telomere, and does not violate the above "1-based" comment.
-     * It's the responsibility of code generating such variants to make sure {@code start} is populated correctly.
-     * And code consuming the returned {@code start} should be prepared for such out-of-the-ordinary values.
-     * 
-     * <strong>Warning:</strong> be aware that the start position of the VariantContext is defined in terms of the start position specified in the
-     * underlying vcf file, VariantContexts representing the same biological event may have different start positions depending on the
-     * specifics of the vcf file they are derived from
+     * In these cases, .
+     *
+     * @return 1-based inclusive start position of the Variant
      */
     @Override
     public int getStart() {
--- a/src/main/java/htsjdk/variant/variantcontext/VariantContext.java
+++ b/src/main/java/htsjdk/variant/variantcontext/VariantContext.java
@@ -1686,10 +1686,6 @@ public class VariantContext implements Feature, Serializable {
      *     And code consuming the returned {@code start} should be prepared for such out-of-the-ordinary values.
      * </p>
      *
-     * The given example is in section 5.4.5 (see example event illustrated in Figure 6 and VCF records below the figure).
-     * Another possible scenario is when a telomere on the p-arm of a chromosome is deleted.
-     * In these cases, .
-     *
      * @return 1-based inclusive start position of the Variant
      */
     @Override
--- a/src/main/java/htsjdk/variant/variantcontext/VariantContext.java
+++ b/src/main/java/htsjdk/variant/variantcontext/VariantContext.java
@@ -1664,7 +1664,15 @@ public class VariantContext implements Feature, Serializable {
 
     /**
      * @return 1-based inclusive start position of the Variant
-     * INDEL events usually start on the first unaltered reference base before the INDEL
+     * INDEL events usually start on the first unaltered reference base before the INDEL.
+     *
+     * Note also that for the VCF spec allows 0 and N + 1 for POS field, where N is the length of the chromosome,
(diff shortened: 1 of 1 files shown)


src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1664,7 +1664,15 @@ public String getContig() {
 
     /**
      * @return 1-based inclusive start position of the Variant
```
vruano: Since you are here, I would reformat the javadoc like this:

```
/**
 * Summary sentence.
 *<p>
*    explanation of what it does without loosing time in details of particular edge cases.
 *</p>
* <p>
*   edge case-1
*</p>
* <p>
*   edge case-2
* </p>
*
* @return the text above should had made clear wha is returned... here you report on the possible range of values very briefly.
*/
```

vruano:
```
For example in this case something like this:

/**
  * Returns the (start) position of the variant.
  * <p>
  *   Main blah blah ... 1-based  ... blah blah.
  *</p>
  *<p>
  * For telomeres  blah blah can be 0  or N+1 blah blah
  *</p>
  * @return 0 or greater, never a negative number.
 */
```

vruano: Here I'm just putting emphasis in the summary sentence (first one finished with .) and the statement in the [at]return tag, the details in the section in between is up to you (that is way I add those "blah" "blah").

That said I think you should be more concise... The programmer don't want to expend to much time to understand what may happen in 0.1% percent of cases.   For example I would not include details on what is said or depicted in the spec, just simply refer to it for further reading/details.

SHuang-Broad:
@vruano  Thanks for taking a look!
Please take a look at the re-formatted doc.

vruano:
In my view is just too much text; think about the 99.5% of people that don't care about telomeres. but I rather you get some other opinion as perhaps I'm just too pedantic.  

About the summary and [at]return annotation. I mean something like this:

```
/**
 * Returns the reporting position for this variant.
 *
 * @return 0 or greater.
 */
```

The summary sentence is "what it does" very briefly whareas the [at]return is just give info about the possible range of values without semantics (this goes into the summary and the rest of the javadoc. For example if the return was Object the often is either ```@return never {@code null}``` or ```@return might be {@code null}.``` Sometimes if the method is know to return ```this``` (e.g. StringBuilder append methods) then I also say that ```@return this very same instance```.

vruano:
I would add ```<strong>warning</strong>``` people that work to this level with VCF data are supposed to be aware of this.

SHuang-Broad:
I agree with you that the summary was too long.
So I shrunk it further and added a warning to the new note about telomere.

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1663,8 +1663,11 @@ public String getContig() {
     }
 
     /**
-     * Returns 1-based inclusive start position of the variant, 0 or greater.
-     * See below for explanation on "0".
+     * Returns 1-based inclusive start position of the variant.
```
vruano: I would drop the "start" since this is in fact controversial, despite that it is the method's name.

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1663,8 +1663,11 @@ public String getContig() {
...
+     *
+     * <p>
+     *     INDEL events usually start on the first unaltered reference base before the INDEL.
```
vruano: This is a bit confussing... INDEL event actual start in the bae before the INDEL... not really ... the indel start where it starts... perhaps you mean to say that it is reported one base before.
Also perhaps you should try to be more general here... instead of talking about INDEL alone you could say something like:
"Notice that for some types of variant events the actual start position may not be this value (e.g. deletions are reported on the base before the first base deleted)."

So you are not giving the impression that it may only happen with deletions or indels.

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
vruano: Also would people know what "1-based inclusive" mean? perhaps that should be remove from here and or explained in another "<p/>" block... but since this not the only place  we use 1-based indexes ... so do you even need to explain it here.....  I guess it won't hurt.

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1675,18 +1678,14 @@ public String getContig() {
...
      * <p>
-     *     INDEL events usually start on the first unaltered reference base before the INDEL.
-     * </p>
-     *
-     * <p>
+     *     <strong>Warning:</strong>
      *     Note also that the VCF spec allows 0 and N + 1 for POS field for telomeric event,
      *     where N is the length of the chromosome.
      *     The "0" value returned should be interpreted as telomere, and does not violate the above "1-based" comment.
-     *     It's the responsibility of code generating such variants to make sure {@code start} is populated correctly.
```
vruano: Is unfortunate the the spec talks about "Telomeres" only. It is assuming that the CHROM (thus the referene) only can contain full non-circular chromosomes.... total BS. (e.g. the MT chromosome in humans or nearly every contig in unfinished genomic references) 
Yes CHROM as a name wasn't a good choice to start with but we don't need to keep up to that mistake. Notice that API using "contig" instead of "chr" or "chromosome; it is still assuming too much but is closer to the truth..  

So I would refrain to make it seems as it can only be telomerees.

What about something like:

"
This property can take on "0" and "N+1" (where N is the last base in the enclosing contig) when this variant record makes references to events that happen before or at the beginning or after or at the end of the enclosing contig."

vruano:
Sorry perhaps I went to far giving examples ... can borrow mine but you can/should use your own words.

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
SHuang-Broad: > I would drop the "start" since this is in fact controversial, despite that it is the method's name.

Why is "start" controversial?

I don't quite understand the rest of the comments

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
SHuang-Broad: This is directly copied from the old comment.

But I agree that it can be more general.

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
SHuang-Broad: I hope this PR does not turn into a loath about the spec itself. I can only work within the current spec for this PR.
So, since the latest spec talks about telomere, I'm happy to use a single word "telomere" to avoid such long sentence.

vruano: It won't, I leave it up to you at this point.

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1663,12 +1663,29 @@ public String getContig() {
...
-     * INDEL events usually start on the first unaltered reference base before the INDEL
-     * 
-     * <strong>Warning:</strong> be aware that the start position of the VariantContext is defined in terms of the start position specified in the
-     * underlying vcf file, VariantContexts representing the same biological event may have different start positions depending on the
-     * specifics of the vcf file they are derived from
+     * Returns 1-based inclusive start position of the variant.
+     *
+     * <p>
+     *     INDEL events usually start on the first unaltered reference base before the INDEL.
+     * </p>
```
pshapiro4broad: `</p>` markup is unnecessary as the `<p>` tag closes the previous open paragraph tag.

Issue #18, opened 2014-05-23
Extended API for symbolic alleles


d-cameron:
Extending the htsjdk to parse symbolic alt alleles and return meaningful Alleles based on the parsed result would be useful to structural variation users of the API. In particular the following features would be useful:
- separation of symbolic alleles into named structural variations, named contig insertions (both with sequence such as "A<ctg>" and without "<ctg>"), breakends and breakpoints
- methods to expose the parsed breakpoint contig, position, directions
- incorporation of all 31 SV headers into VCFStandardHeaderLines (or split out into VCFStructuralVariationHeaderLines)
- methods to return the base call portions of symbolic alleles (ie: extract the ACGT from ACGT[chr1:1[ )

One possible design is to subclass Allele:
abstract Allele
Allele <- BaseCallAllele
Allele <- SymbolicAllele
Allele <- NoCallAllele
Allele <- MissingAllele
SymbolicAllele <- StructuralVariationAllele (with enum, subclass hierarchy or getters for SV type and subtypes (subsubtypes?) such as <DUP> and <DUP:TANDEM>)
SymbolicAllele <- BreakendAllele
SymbolicAllele <- BreakpointAllele
SymbolicAllele <- NamedContigAllele

The low-level design and implementation depends on resolution of ambiguities in the VCF specifications as the are multiple points in which the VCF specifications are under-defined and ambiguous. For example, it is unclear as to whether "AA<ctg>", "<ctg><ctg2>", "AA<INS>" are valid alleles.

PR #1370, opened 2019-05-18
Addresses part of issue #18, Allele class refactoring.

vruano:
More concretely it refactor the Allele class:

* Is now an interface.
* Added several methods to access properties from several allele types like breakends and whole contig insertions.
* Disentangles the current mixture of "display" bytes vs base bytes in the bases array... now only base are stored like that if it applies.
* Adds a "hidden" hierarchy of implementations for different allele subtypes.
* Improves reusing instances for common variants (single base and common symbolic alleles)
* Mark as deprecated several method that seem unnecessary.
* Disable the possibility of changing an allele bases and adds an API (BaseSequence interface) to access these efficiently
  wihtout the need of cloning a byte[] array each time.
* Makes impossible to create alleles that have invalid base (IUPAC codes that are not allowed by the VCF spec).
* Fixes (and also deprecates) some not well-though Allele method that only make sense for some subtypes of alleles.
* Breakend API, access to contig, position type of breakend and base.

### Description

### Checklist

- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,7 @@ jacocoTestReport {
 }
 
 dependencies {
+    compile "org.apache.commons:commons-lang3:3.9"
     compile "org.apache.commons:commons-jexl:2.1.1"
     compile "commons-logging:commons-logging:1.1.1"
     compile "org.xerial.snappy:snappy-java:1.1.4"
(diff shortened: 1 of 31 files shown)


vruano:
Allele.java diffs rather messy perhaps reviewers should consider to assess it on the changed version of the file and go back to the github diff just report what they don't like.

vruano:
Current test pass but need to add new ones, please review in the meantime.

vruano:
I'm adding test code and fixing issues in the code as I come across them, you can start reviewing the first commit only for stability.

tfenne:
I think this work is largely superseded by #1454 and #1576.  If that's not the case and you (or anyone else) are interested in resuscitating this PR please re-open.

Issue #1371, opened 2019-05-20
GL field handling inconsistent between encoder and decoder


mitochon:
### Verify
Somewhat related [issue](https://github.com/samtools/htsjdk/issues/340) but not the same.

### Subject of the issue
Getting an error reading GL fields written using htsjdk for multi-sample VCF.

### Your environment
* version of htsjdk = 2.18.0 (also in master branch)
* version of java = 1.8
* which OS = Mac OS X

### Steps to reproduce

When creating multi sample VCF, the encoder puts in ".,.,." for missing GL fields, e.g.

```
chr1	123	.	G	A	.	.	.	GT:GL:MIN_DP	0/1:0,0.9,0.1	./.:.,.,.:10
```

Specifically there's this comment in the encoder code
https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/variant/vcf/VCFEncoder.java#L346

```
// If we have a missing field but multiple values are expected, we need to construct a new string with all fields.
// For example, if Number=2, the string has to be ".,."
```

The VCFCodec does not honor this contract and I get an error when trying to parse the above line:

```
java.lang.NumberFormatException: For input string: "."
	at sun.misc.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:2043)
	at sun.misc.FloatingDecimal.parseDouble(FloatingDecimal.java:110)
	at java.lang.Double.parseDouble(Double.java:538)
	at htsjdk.variant.variantcontext.GenotypeLikelihoods.parseDeprecatedGLString(GenotypeLikelihoods.java:264)
	at htsjdk.variant.variantcontext.GenotypeLikelihoods.fromGLField(GenotypeLikelihoods.java:90)
	at htsjdk.variant.vcf.AbstractVCFCodec.createGenotypeMap(AbstractVCFCodec.java:722)
	at htsjdk.variant.vcf.AbstractVCFCodec$LazyVCFGenotypesParser.parse(AbstractVCFCodec.java:132)
	at htsjdk.variant.variantcontext.LazyGenotypesContext.decode(LazyGenotypesContext.java:158)
	at htsjdk.variant.vcf.AbstractVCFCodec.parseVCFLine(AbstractVCFCodec.java:352)
	at htsjdk.variant.vcf.AbstractVCFCodec.decodeLine(AbstractVCFCodec.java:284)
	at htsjdk.variant.vcf.AbstractVCFCodec.decode(AbstractVCFCodec.java:262)
```

[...]

PR #1372, opened 2019-05-20
VCF codec should handle multiple missing GL fields

mitochon:
### Description

This change attempts to solve https://github.com/samtools/htsjdk/issues/1371 with the least amount of change. Currently the htsjdk encoder will write `.,.,.` for multiple missing GL fields, but the decoder will throw an error when parsing this output because it assumes a single `.` and the rest must be double.

I noticed the actual method is marked a deprecated but somehow its still used by `AbstractVCFCodec`. This is a simple workaround that handles this case and is backwards compatible.

### Checklist

- [x] Code compiles correctly


mitochon:
@lbergelson @tomwhite you are the last people who touched this file. Would you be available to review this PR or designate assignees?

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -267,7 +267,11 @@ public static double getGQLog10FromLikelihoods(int iOfChoosenGenotype, double[]
             String[] strings = GLString.split(",");
             double[] likelihoodsAsVector = new double[strings.length];
             for ( int i = 0; i < strings.length; i++ ) {
-                likelihoodsAsVector[i] = Double.parseDouble(strings[i]);
+                if ( strings[i].equals(VCFConstants.MISSING_VALUE_v4) ) {
+                  return null; // one or more entries are missing
```
yfarjoun: hmmm. so 1,.,2 will return a null value? I guess with primitive doubles we have, to...though we could use NaNs for that. maybe when we support 4.3....

src/test/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
```diff
@@ -82,6 +82,13 @@ public void testFromString2() {
         Assert.assertEquals(gl.getAsString(), vPLString);
     }
 
+    @Test
+    public void testFromString3() {
```
yfarjoun: please explicitly test what would happen in the case that some of the elements are missing and others not (so that there's an explicit example of lost data.

mitochon:
Incorporated feedback. Pull from latest.

mitochon:
@yfarjoun thanks for the review. I've updated as requested.

yfarjoun:
@lbergelson what do you think about this? explicitly accepting a missing element in a genotype array and converting the entire array to missing...I agree that the spec isn't very clear about it....but I'm not sure that this is the right way to resolve it..

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -242,7 +242,7 @@ public static double getGQLog10FromLikelihoods(int iOfChoosenGenotype, double[]
     }
 
     private final static double[] parsePLsIntoLikelihoods(String likelihoodsAsString_PLs) {
-        if ( !likelihoodsAsString_PLs.equals(VCFConstants.MISSING_VALUE_v4) ) {
+        if ( likelihoodsAsString_PLs != null && !likelihoodsAsString_PLs.equals(VCFConstants.MISSING_VALUE_v4) ) {
```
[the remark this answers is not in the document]
mitochon: Without this change, this test (added for this PR) will fail -> https://github.com/mitochon/htsjdk/blob/1371-decoder-handle-multiple-missing-GL-fields/src/test/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java#L86

More generally, this method assumes `likelihoodsAsString_PLs` cannot be null which is not true even before this PR.

mitochon:
@pshapiro4broad I didn't want to pollute the unit test for `GenotypeLikelihood` with imports from other modules, i.e.

```
import htsjdk.variant.variantcontext.*
import htsjdk.variant.vcf.*
```

I don't know if there's a place in the repo to do some sort of integration test between different modules.
Also the actual method that's driving the change here has been deprecated
```
    @Deprecated
    public final static GenotypeLikelihoods fromGLField(String GLs) {
        return new GenotypeLikelihoods(parseDeprecatedGLString(GLs));
    }
```

I'm not sure when is the plan to remove the corresponding call from `AbstractVCFCodec`. I am hoping this PR will fix the inconsistency between the VCF writer and reader with minimum disruption (no signature changes, etc) that will solve my use case and hopefully others' too without investing too much effort given this is a deprecated method and all.

Just for additional context, I'm using `htsjdk` to write multi-sample VCF's which then gets loaded to [hail](https://hail.is) ... hail also uses `htsjdk` to read VCF's and currently its not able to do so due to this mismatch between reader and writer.

mitochon:
@yfarjoun what's your thought on this PR? If your concern is `null` is returned when there are only one missing value I can instead put `Double.NaN` in the returned array as you suggested, along with any other non-missing values. 
Will that work?

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
lbergelson: I think a mixed case like `1,.,2` should probably throw an exception.

lbergelson:
I'd advocate for throwing an exception if we have partially missing GL.

lbergelson:
I'd advocate for throwing an exception if we have partially missing GL.  We don't handle partially missing arrays values anywhere but in GT and I'm not sure that we should start here.  

I think it would probably make sense to change the vcf writer to just write a single missing value instead of `.,.,.` while we're at it.  I don't see the value of writing an array of missing values versus just making the array missing.  I can see an argument that they are different, but we're not handling them as different, we're just always outputting any missing array as an array of missing values.

mitochon:
Thanks for the review @lbergelson. I'll update the PR for this change -> ` throwing an exception if we have partially missing GL.` Hopefully all parties will agree.

lbergelson:
So I ended up looking at what we currently do with other values, and looks like for AD and PL we silently null out partially missing fields.  I hadn't realized that.  So making you error on a partially missing value is actually less consistent.  That said, I think it's probably the right thing.  @yfarjoun What do you think?  Should we make treatment of AD and PL more stringent as well?

mitochon:
@lbergelson for `AD` at least throwing hard error on missing values may be too stringent. When multi-allelic variants get split or merged in a multi-sample VCF there could be use cases where its acceptable to have missing values.

`silently null out partially missing fields` -> doing this should be easy also. Let me know either way.

lbergelson:
@yfarjoun Any further comments? Looks good to me.  I'm going to open a separate PR to prevent writing `.,.,.`

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -266,10 +266,21 @@ public static double getGQLog10FromLikelihoods(int iOfChoosenGenotype, double[]
         if ( !GLString.equals(VCFConstants.MISSING_VALUE_v4) ) {
             String[] strings = GLString.split(",");
             double[] likelihoodsAsVector = new double[strings.length];
+            int missing = 0;
             for ( int i = 0; i < strings.length; i++ ) {
-                likelihoodsAsVector[i] = Double.parseDouble(strings[i]);
+                if ( strings[i].equals(VCFConstants.MISSING_VALUE_v4) ) {
```
yfarjoun:
```suggestion
                if (strings[i].equals(VCFConstants.MISSING_VALUE_v4)) {
```

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -266,10 +266,21 @@ public static double getGQLog10FromLikelihoods(int iOfChoosenGenotype, double[]
...
+                  missing ++;
```
yfarjoun:
```suggestion
                  missing++;
```

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -266,10 +266,21 @@ public static double getGQLog10FromLikelihoods(int iOfChoosenGenotype, double[]
...
+                } else {
+                  likelihoodsAsVector[i] = Double.parseDouble(strings[i]);
+                }
+            }
+            if ( missing == 0 ) {
```
yfarjoun:
```suggestion
            if (missing == 0) {
```

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -266,10 +266,21 @@ public static double getGQLog10FromLikelihoods(int iOfChoosenGenotype, double[]
...
+              return likelihoodsAsVector;
+            } else if ( likelihoodsAsVector.length == missing ) {
```
yfarjoun:
```suggestion
            } else if (likelihoodsAsVector.length == missing) {
```

mitochon:
@yfarjoun updated with formatting changes as requested.

`but I'm a little worried about silently dropping data` -> with the latest implementation there's no data dropped, the code will throw an exception for partial missing values.

Also, speaking of formatting, please check out my PR for a IDE formatter #1386 =)

yfarjoun:
Thanks!

PR #1374, opened 2019-05-22
Substitution matrix refactoring and tests

cmnbroad:
There are 4 commits:

- use an enum for substitution bases, tighten up types and matrix code, range checking
- a minor reordering of methods so all public methods come first
- move creation of SubstitutionMatrix inside of CompressionHeaderFactory, rather than having all callers create it
- change the SubstitutionMatrix constructor to take CramCompressionRecords directly, rather than an array of frequencies, and add a bunch of tests

A subsequent PR will continue with CramCompressionHeader/Factory refactoring.

I recommend reviewing the commits individually, in order, since I think that it be easier to follow that way.


src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -164,44 +140,117 @@ public SubstitutionMatrix(final byte[] matrix) {
...
+            }
+            stringBuilder.append('\t');
+        }
+        return stringBuilder.toString();
+    }
+
+    // Initialize the baseByCode array to a sentinel value
+    private void initializeBaseByCode() {
+        for (final byte[] base : baseByCode) {
+            Arrays.fill(base, NO_BASE);
```
vruano: NO_BASE is 0 so all this filling actually does nothing because those arrays are initialized with zero by default. If this method is called somewhere else apart from the beginning of the constructor code then it should be call ```clearBaseByCode()``` instead.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -164,44 +140,117 @@ public SubstitutionMatrix(final byte[] matrix) {
...
+    private static long[][] buildFrequencies(final List<CramCompressionRecord> cramRecords) {
+        final long[][] frequencies = new long[SYMBOL_SPACE_SIZE][SYMBOL_SPACE_SIZE];
+        for (final CramCompressionRecord record : cramRecords) {
+            if (record.readFeatures != null) {
+                for (final ReadFeature readFeature : record.readFeatures) {
+                    if (readFeature.getOperator() == Substitution.operator) {
+                        final Substitution substitution = ((Substitution) readFeature);
+                        final byte refBase = substitution.getReferenceBase();
+                        final byte base = substitution.getBase();
+                        frequencies[refBase][base]++;
```
vruano: refBase and base are typed as bytes... they won't ever take values over 127 (as the highest byte is set to 1 then they are negative). However frequencies is a [256][256] 2d array. ... perhaps despite than there are 256 bytes you actually wan to create this arrays to only 128 in size and simply disregard negative base values? Need to check in other places but is quite possible that this is a common problem with other arrays around here like codeByByte and byteByCode.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -71,52 +64,37 @@
...
-        }
-        for (int i = 0; i < SUBSTITUTION_BASES_UPPER.length; i++) {
-            final byte r = SUBSTITUTION_BASES_UPPER[i];
-            for (final byte b : SUBSTITUTION_BASES_UPPER) {
-                if (r == b) {
-                    continue;
+        for (final SubstitutionBase r : SubstitutionBase.values()) {
+            for (final SubstitutionBase b : SubstitutionBase.values()) {
+                if (r != b) {
+                    baseByCode[r.getBase()][codeByBase[r.getBase()][b.getBase()]] = b.getBase();
```
vruano: getBase() will return a ```byte``` whose value is -128 to 127. Obviously negatives are going to cause a problem here therefore In practice here you only get 0 to 127 however these arrays are initialized to store from 0 to 256. So you can change the array lengths to 128 without further changes and add the "documented" constraint that getBase() won't ever return a negative... or change its return type to ```char``` that is never a negative (is the unsigned short equivalent in Java)...

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -164,44 +140,117 @@ public SubstitutionMatrix(final byte[] matrix) {
...
 
+    /**
+     * Given a reference base and a read base, find the corresponding substitution code
+     * @param refBase reference base being substituted
+     * @param readBase read base to substitute for the reference base
+     * @return code to be used for this refBase/readBase pair
+     */
+    public byte code(final byte refBase, final byte readBase) {
+        if (Character.isLowerCase((char) refBase)) {
+            log.warn(String.format("CRAM: Attempt to generate a substitution code for a lower case reference base '%c'", (char) refBase));
```
vruano: Is this log potentially going to penalize performance too much with some real-life datasets or you consider them as pathological? There is already and additional ```if``` evaluation penalty for everyone and for those that have a lower case base a string format and a log.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -164,44 +140,117 @@ public SubstitutionMatrix(final byte[] matrix) {
...
-     * @return
+     * @return the encoded matrix in serialized form
      */
     public byte[] getEncodedMatrix() {
         return encodedMatrixBytes;
     }
 
+    @Override
+    public String toString() {
+        final StringBuilder stringBuilder = new StringBuilder();
```
vruano: Can you pass a final size estimate to the builder constructor?

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -164,44 +140,117 @@ public SubstitutionMatrix(final byte[] matrix) {
...
+            stringBuilder.append((char) r.getBase());
+            stringBuilder.append(':');
+            for (int i = 0; i < CODES_PER_BASE; i++) {
+                stringBuilder.append((char) baseByCode[r.getBase()][i]);
+            }
+            stringBuilder.append('\t');
+        }
+        // lower case substitutions
+        for (final SubstitutionBase r : SubstitutionBase.values()) {
+            stringBuilder.append(Character.toLowerCase((char)r.getBase()));
```
vruano: Perhaps you should consider to cache the "toLowerCase" in the enum... is just 4-5 values so the amount of memory need is nothing and saves you this toLowerCase calls here and there.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -164,44 +140,117 @@ public SubstitutionMatrix(final byte[] matrix) {
...
+            stringBuilder.append(':');
+            for (int i = 0; i < CODES_PER_BASE; i++) {
+                stringBuilder.append((char) baseByCode[Character.toLowerCase((char)r.getBase())][i]);
```
vruano: you can cache here the return from ```baseByCode[Character.toLowerCase((char)r.getBase())]```... in fact you could write this for as ```stringBuilder.append(new String(baseByCode[Character.toLowerCase((char)r.getBase())], 0, CODES_PER_BASE))``` if CODES_PER_BASE is the same length as the array you can skip the two last int arguments.

vruano: That said I don't know if performance is better adding the bytes (casted into chars) one by one, is possible.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -71,52 +64,37 @@
...
 
-        for (final byte[] base : baseByCode) {
-            Arrays.fill(base, (byte) 'N');
-        }
-        for (int i = 0; i < SUBSTITUTION_BASES_UPPER.length; i++) {
-            final byte r = SUBSTITUTION_BASES_UPPER[i];
-            for (final byte b : SUBSTITUTION_BASES_UPPER) {
-                if (r == b) {
-                    continue;
+        for (final SubstitutionBase r : SubstitutionBase.values()) {
```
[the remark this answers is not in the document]
cmnbroad: Yeah that probably worth doing. The array that was previously used was public, but it was only used by tests. Now that I have the enum, I can switch to using a `List` internally, and let the tests continue to use the `.values` call on the enum.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
lbergelson: This does seem weird.  If it's abnormal lets throw and we'll find out that it's happening.  If it's expected than logging on  evert read * base is a bad idea.  If it's expected than we can of course just do the substitution without comment.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -164,44 +140,117 @@ public SubstitutionMatrix(final byte[] matrix) {
...
-            // primary sort by frequency
-            if (o1.freq != o2.freq) {
-                return (int) (o2.freq - o1.freq);
-            }
-            // same frequency; compare based on spec tie-breaking rule (use base order prescribed by the spec)
-            return CODE_ORDER[o1.base] - CODE_ORDER[o2.base];
-        }
-    };
+    private static final Comparator<SubstitutionFrequency> comparator =
+            (o1, o2) -> {
```
lbergelson: If you're rewriting this you could use the Comparator.comparing methods instead, but I'm not sure it really buys you any clarity in this case.

src/test/java/htsjdk/samtools/cram/structure/SubstitutionMatrixTest.java
```diff
@@ -1,103 +1,193 @@
...
+            final int nC,
+            final int nG,
+            final int nT,
+            final int nN,
+            final byte[] expectedMatrixBytes) {
+        // make sure the matrix generated for the given substitution frequencies has the codes ordered as expected
+        final SubstitutionMatrix substitutionMatrix = getSubstitutionMatrixForFrequencies(refBase, nA, nC, nG, nT, nN);
+        Assert.assertEquals(substitutionMatrix.getEncodedMatrix(), expectedMatrixBytes);
+
+        // create a "roundTrip" matrix using the expected bytesas input, and cross-validate that the codes match
```
lbergelson:
```suggestion
        // create a "roundTrip" matrix using the expected bytes as input, and cross-validate that the codes match
```

src/main/java/htsjdk/samtools/cram/structure/SubstitutionBase.java
```diff
@@ -0,0 +1,29 @@
+package htsjdk.samtools.cram.structure;
+
+/**
+ * Bases from the base symbol space that are subject to substitution via the CRAM SubstitutionMatrix, in the order
+ * in which they're serialized in the SubstitutionMatrix.
+ */
+enum SubstitutionBase {
+    A('A'),
```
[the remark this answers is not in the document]
cmnbroad: Yeah,  done.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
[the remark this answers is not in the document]
cmnbroad: Done.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -71,52 +64,37 @@
...
+     * @param records array of CramCompressionRecord with Substitutions
      */
-    public SubstitutionMatrix(final long[][] frequencies) {
-        // get the substitution code vector for each base
-        for (final byte base : SUBSTITUTION_BASES_UPPER) {
+    public SubstitutionMatrix(final List<CramCompressionRecord> records) {
+        initializeBaseByCode();
+
+        final long[][] frequencies = buildFrequencies(records);
+        for (final SubstitutionBase b : SubstitutionBase.values()) {
```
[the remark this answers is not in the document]
cmnbroad: Done.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
cmnbroad: Yes, let throw. I'm hoping that at some point soon we can remove the lower case code entirely, since it really shouldn't happen. Having said that, I'm unsure why the original author went out of their way to make it work.

cmnbroad: Done.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
cmnbroad: I added the return value caching to improve readability, but otherwise perf is not really an issue for this method, which code is only used in debugging anyway.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
cmnbroad: Done.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
cmnbroad: Right - this used to use a non-zero value, not now its redundant. Removed.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
cmnbroad: True. I reduced the size to 128 and added guards to check for negative values which we shuld never see.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionBase.java
```diff
@@ -0,0 +1,29 @@
...
+    C('C'),
+    G('G'),
+    T('T'),
+    N('N');
+
+    // The base this substitution represents
+    private final byte base;
+
+    /**
+     * @param base the character represneting the base for this substitution
```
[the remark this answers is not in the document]
cmnbroad: This whole thing was removed per a separate comment.

src/test/java/htsjdk/samtools/cram/structure/SubstitutionMatrixTest.java
cmnbroad: Done.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
cmnbroad: Also, in the next round I'll add similar (non-negative) restrictions to the `Substitution` code.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -164,44 +140,125 @@ public SubstitutionMatrix(final byte[] matrix) {
...
     }
 
+    /**
+     * Given a reference base and a read base, find the corresponding substitution code
+     * @param refBase reference base being substituted
+     * @param readBase read base to substitute for the reference base
+     * @return code to be used for this refBase/readBase pair
+     */
+    public byte code(final byte refBase, final byte readBase) {
+        if (refBase <= 0 || Character.isLowerCase((char) refBase)) {
```
vruano: What if the refBase is either upper-case non valid base e.g. 'J' or some symbol chracter like '<'... How is that we care so much so censore lower-case letters (not only valid bases) but nothing else that in fact is not a valid base call?

Since we go to so much trouble to create this big 128, 128 matrices we could use them to detect bad ref bases without calling isLowerCase.... if (codeByBase[refBase] == null is a bad refBase... that cause a throw... and only then we can check whether refBase was a lower-case version of an otherwise valid base in which case you customize the error message accordingly. 

Alternatively, if we accepted '>' or 'J' as "good" or "ok" ref bases what would happen to these case down-the-road... was an exception thrown on them later one, perhaps we should use the same approach with 'a', 'c', 'g', 't' and  'n'.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -164,44 +140,125 @@ public SubstitutionMatrix(final byte[] matrix) {
...
+            throw new IllegalArgumentException(
+                    String.format("CRAM: Attempt to generate a substitution code for invalid or lower case reference base '%c'", (char) refBase));
+        }
+        if (readBase <= 0) {
```
vruano: I wonder... why the asymmetric treatment for read bases... cannot these also be erroneous (e.g. lowercase or symbols?) why not croak on those with we do it with refBase.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -164,44 +140,125 @@ public SubstitutionMatrix(final byte[] matrix) {
...
+    }
+
+    /**
+     * Given a reference base and a substitution code, return the corresponding substitution base.
+     * @param refBase reference base being substituted
+     * @param code substitution code
+     * @return base to be substituted for this (refBase, code) pair
+     */
+    public byte base(final byte refBase, final byte code) {
+        if (refBase <= 0) {
```
vruano: Why don't care about lower vs upper-case here?

vruano: No check is done on ```code``` is that ok?

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -164,44 +140,125 @@ public SubstitutionMatrix(final byte[] matrix) {
...
-     * @return
+     * @return the encoded matrix in serialized form
      */
     public byte[] getEncodedMatrix() {
         return encodedMatrixBytes;
     }
 
+    @Override
+    public String toString() {
+        final int DISPLAY_SIZE = BASES_SIZE * (CODES_PER_BASE + 3) * 2;
```
vruano: if you want you can move this into a static constant field since is in upper case... if its a local variable I would make it lowercase or simply inline the code in the ```new StringBuilder(...)``` statement.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
cmnbroad: You're correct that the parameter checking here is somewhat selective. Previously this code would have silently returned a code given any pair of byte values, and I've added only selective checks that reject completely nonsensical inputs (such as `refBase <= 0`), or attempts to use lower case bases (since we're unclear why code was specifically included to handle them when they're not allowed by the spec, and want to find out if we have any code paths or data that depends on it).

More thoroughly checking of the reference and read bases *is* something we should do, but it should be done at the site(s) where `Substitution` class instances are created in the first place. The CRAM refactoring is being done in a series PRs rather than all at once in order to keep the reviews manageable and prevent them from getting stale. The `Substitution` changes, will be in one of the next refactoring rounds, along with changes to use more strongly typed arguments for these methods, i.e., `Substitution` instead of (`byte, byte`), after which the matrix indexing will be completely safe.

vruano: For you are saying here you actually added this additional lower-case check but at the same time is not clear what we should be doing about them in the end (after those several PRs)... or simply the problem will disappear by using specialized input types rather than standard byte primitives ....  in that case wouldn't make more sense simply error on the side of simplicity and omit this change i.e. don't check for lower-case as it used to be?

cmnbroad: @vruano Initially I was only logging a warning here, since I still want a have a way to find out if there are cases where we are using lower case bases (especially, other places in the code that aren't normalizing the bases to upper case, as they should be). Changing it so it throws was based on review feedback, and I think it makes sense. Its the surest way to find any offending code paths. So I think throwing for now is the right thing to do.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -28,38 +33,29 @@
...
  */
-public class SubstitutionMatrix {
-    // List of bases subject to substitution. The upper and lower case arrays must have the bases
-    // in the same order so that the index for a given base is the same for upper and lower case
-    static final byte[] SUBSTITUTION_BASES_UPPER = new byte[]{'A', 'C', 'G', 'T', 'N'};
+ public class SubstitutionMatrix {
+    private static final Log log = Log.getInstance(Substitution.class);
+
+    // substitution bases, in the order in which they're stored in the substitution matrix
+    private static List<SubstitutionBase> BASES = Arrays.asList(SubstitutionBase.values());
```
pshapiro4broad: Is there a reason to not use `SubstitutionBase[] BASES` here? It looks like all uses of `BASES` are in `for` loops which would support arrays.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -164,44 +140,125 @@ public SubstitutionMatrix(final byte[] matrix) {
...
+        for (final CramCompressionRecord record : cramRecords) {
+            if (record.readFeatures != null) {
+                for (final ReadFeature readFeature : record.readFeatures) {
+                    if (readFeature.getOperator() == Substitution.operator) {
+                        final Substitution substitution = ((Substitution) readFeature);
+                        final byte refBase = substitution.getReferenceBase();
+                        final byte base = substitution.getBase();
+                        if (refBase <= 0 || base <= 0) {
+                            throw new IllegalArgumentException(
+                                    String.format("CRAM: Attempt to generate a substitution code for invalid reference base '%c'", (char) refBase));
```
pshapiro4broad: If the values are negative, printing them as a character doesn't seem like a great way to provide feedback to the user. Unless the user entered this value as a character in the first place.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
cmnbroad: @pshapiro4broad You're right that it could go either way based on the usages, and I don't mind changing it, but its is a static list - is there a compelling reason that array is better here ?

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
cmnbroad: Yeah, true that.

vruano: If this comes from a careless ```char``` -> ```byte``` cast, a negative here could come from a extended asciii characters like 'ä' (with a code 128 or above). So they are actually printable .... the case that would need to worry are control characters (e.g. tabs new-lines) ... ```java.lang.Character``` query method should give you the ability to tell those apart. Perhaps always print the unsigned code and for those that are printable, you can include the printable version between () beside the code.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
pshapiro4broad: It avoids the need to create an `Arrays.asList()` wrapper.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
```diff
@@ -164,44 +140,125 @@ public SubstitutionMatrix(final byte[] matrix) {
...
-        public int compare(final SubstitutionFrequency o1, final SubstitutionFrequency o2) {
-            // primary sort by frequency
-            if (o1.freq != o2.freq) {
-                return (int) (o2.freq - o1.freq);
-            }
-            // same frequency; compare based on spec tie-breaking rule (use base order prescribed by the spec)
-            return CODE_ORDER[o1.base] - CODE_ORDER[o2.base];
-        }
-    };
+    private static final Comparator<SubstitutionFrequency> comparator =
```
pshapiro4broad: Use uppercase for statics
```suggestion
    private static final Comparator<SubstitutionFrequency> COMPARATOR =
```

cmnbroad: Done.

src/main/java/htsjdk/samtools/cram/structure/SubstitutionMatrix.java
cmnbroad: Changed. If this ever happens, its an internal error, not one the user can recover from, so changing this to display as a signed integer value.

cmnbroad:
@vruano @lbergelson @pshapiro4broad Any more comments on this ? I'd like to get it approved/merged since I have many commits piled up behind it.

vruano:
I don't have any further comments.

pshapiro4broad:
LGTM

cmnbroad:
Thanks for the reviews @vruano @pshapiro4broad .

PR #1376, opened 2019-05-24
Fix to ensure we only publish snapshots compiled with openjdk-8.

tfenne:
### Description

With the travis builds now running on multiple Java versions, whichever one finishes last wins at publishing the snapshot build.  This is problematic as there are situations where compiling on Java10 (targeting Java 1.8) still causes issues.  Specifically there are changes to `ByteBuffer` to make some methods return `ByteBuffer` where in Java 8 they returned `Buffer`, and at runtime this causes link/no-such-method problems running on Java 8!

The solution is only to publish if the Java Home is set to openjdk-8.

### Checklist

- [x] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,5 +41,7 @@ after_success:
   - echo "TRAVIS_BRANCH='$TRAVIS_BRANCH'";
     echo "JAVA_HOME='$JAVA_HOME'";
     if [ "$TRAVIS_BRANCH" == "master" ]; then
-       ./gradlew publish;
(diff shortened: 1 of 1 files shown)


tfenne:
@lbergelson I'd appreciate a quick thumbs up/down on this as I'm blocked from using the snapshot release with my latest two PRs until I fix this.  Given the bash code only executes when on master, I will have to merge this just to test it, and if it doesn't work may have to push to master to fix it :(

tfenne:
Thanks @lbergelson!

PR #1349, opened 2019-04-04
CRAM WIP: expand ReferenceContext to include Alignment Start and Span

jmthibault79:
### Description

ReferenceContext often travels with Alignment Start and Span.  Let's combine them.

Added new AlignmentContext to Slice, Container, and CRAIEntry.

Also adds ReferenceContext to CramIterator.

TODO:
- AlignmentSpan
- CramCompressionRecord (see #1332)
- CRAMContainerStreamWriter (see #1333)
- Maybe: BAMIndexer, BAMIndexMetadata, CRAMBAIIndexer, CRAMCRAIIndexer, CramIterator

### Checklist

- [x] Code compiles correctly
- [x] New tests covering changes and new functionality
- [x] All tests passing
- [x] Extended the README / documentation, if necessary
- [ ] Is not backward compatible (breaks binary or source compatibility)

PR #1348, opened 2019-04-03
WIP: Progress toward Container encapsulation

jmthibault79:
### Checklist

- [ ] Code compiles correctly
- [ ] New tests covering changes and new functionality
- [ ] All tests passing
- [ ] Extended the README / documentation, if necessary
- [ ] Is not backward compatible (breaks binary or source compatibility)

PR #1377, opened 2019-05-27
Read no_ref crams

sd4B75bJ:
### Description

Adding support for reading no_ref CRAMs

samtools#721

### Checklist

- [x] Code compiles correctly
- [x] All tests passing


cmnbroad:
@sd4B75bJ Thanks for submitting a PR to htsjdk. Unfortunately, the problems with RR go deeper than just the MD5 check, so relaxing the MD5 validation fixes only the symptom, not the underlying problem. The CRAM code as it stands doesn't correctly handle read features for RR=0. If you look at what happens in your test case, the bases for the reads are being pulled directly from the reference (which the caller is still required to provide) rather than from the `Bases` read features that are embedded in the CRAM file.

A refactoring of much of the CRAM code is currently in progress on another branch in order to address those and other issues. Once thats done and tests are put in place to verify that the read bases are correctly round tripped, then the  MD5 check, along with the requirement to always provide a reference, can relaxed.

cmnbroad:
@sd4B75bJ Closing based on previous comments - feel free to reopen/comment if I've missed something.

sd4B75bJ:
@cmnbroad okay sounds reasonable, thanks for reviewing. Sorry, could you point me at the main CRAM refactoring branch that you refer to above?

cmnbroad:
@sd4B75bJ The refactoring is being done in waves, so there are actually a series of PRs and branches underway, in various states of readiness. i.e., https://github.com/samtools/htsjdk/pull/1374, https://github.com/cmnbroad/htsjdk/tree/cn_cram_compression_header_factory, https://github.com/samtools/htsjdk/pull/1164, https://github.com/samtools/htsjdk/pull/1348, https://github.com/samtools/htsjdk/pull/1349, to name a few.

PR #1384, opened 2019-06-13
Prevent integer overflow in Interval countBases method.

cmnbroad:
This fixes a regression introduced in https://github.com/samtools/htsjdk/pull/1356. We should get this in before we release. Caught by Picard test `IntervalListTool` testActionsWithInvert.


lbergelson:
add a test?

cmnbroad:
Yes, it should have a test.

Issue #1196, opened 2018-10-11
Adopt automatic code style formatting


nh13:
@tfenne @lbergelson @jacarey @pshapiro4broad

From recent memory, I have seen most pull requests have style issues (mine included) with anywhere between 20-80% of the comments related to code style.  I suggest moving to using a code reformatter, for example [google-java-format](https://github.com/google/google-java-format) or an equivalent (top hit on google for me, hah!).  I don't particularly care which one, as long as it's IDE-independent.  Yes, the code may get uglier to some, but as always, there's a trade-off that were are currently making so I'd like to revisit that.

I think this would go a long way to reduce chatter on issues about style, and focus the conversations around substance.

pshapiro4broad:
I think that's a great idea, and I agree that the solution shouldn't require an IDE.  

`google-java-format` looks interesting, and there are notes [here](https://bjurr.com/java-code-formatting-with-google-java-format/) about how to add it to a gradle build.

In the past I've used [checkstyle](http://checkstyle.sourceforge.net/) which works quite well, but it doesn't fix your code, it only tells you what needs fixing. (although there are plugins for ecplise and intellij that will auto-fix checkstyle errors)

nh13:
For fun, I have a branch with code style updated on compile: https://github.com/samtools/htsjdk/pull/1197.  Compile-time formatting is a little dangerous, as syntax errors can cause imports to be removed, but this forces folks to always have well formatted code.  The other option is to use travis-ci to enforce it via the `verifyGoogleJavaFormat` task.

nh13:
I know @tfenne will return next week from vacation, but @lbergelson or @jacarey care to weigh in?

lbergelson:
I like the idea of having code style be more automated.  I have a few concerns though.

1.  Reformatting the whole of htsjdk at once is going to make the history much more annoying, this isn't necessarily a deal breaker but it's not great.
2.  We want to be sure that it's uncommon for people to submit a PR and then discover they forgot to format things, ideally, we'd have easy to set up git hooks and instructions for how to set it up so that it always formats before committing.
3.  Auto formatters can be really annoying if you and the formatter disagree and it's not very configurable.  They're also not great when there are special cases that need special formatting.

yfarjoun:
so, I'm not @lbergelson or @jacarey...but I have to say that I don't like the 2-space version...also the linter seems to chop lines in weird places...

nh13:
@yfarjoun any reason you closed this?

[...]

PR #1386, opened 2019-06-14
add style guide for IDE's

mitochon:
### Description

I recently submitted my first PR to this project. I find that the code style is heterogenous which sometimes distracts me from working.

I am used to having an IDE format things but I realize this may not work for everybody. I see some discussion already in #1196 about enforcing certain styles.

My proposal is the following:
- Add a non-binding style guide and people can adopt it if they like.
- This is stolen from [Google's styleguide](https://github.com/samtools/htsjdk/issues/1196) with one mofication => change indent size from 2 to 4 which seem to be the default here.

I'm relying on inertia, with the assumption that people won't reformat files that's already reformatted, and new changes will mostly be in-line with the proposed format.

Hopefully in the long run the style will converge, but initially there may be a lot of PR's with a lot of formatting changes when people start using their IDE to reformat stuff.

Side note: I asked about this earlier in the gitter channel and @cmnbroad suggested to try (see logs on May 21 and May 22)

### Checklist

- [n/a ] Code compiles correctly
- [ n/a] New tests covering changes and new functionality
- [ n/a] All tests passing
- Extended the README / documentation, if necessary <- let me know if any is needed
- [ n/a] Is not backward compatible (breaks binary or source compatibility)

### References

Diff between proposed version and Google's version (TLDR; increase indent from 2 -> 4)

[eclipse formatter](https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml)
```
3c3
< <profile kind="CodeFormatterProfile" name="htsjdkStyle" version="1">
---
> <profile kind="CodeFormatterProfile" name="GoogleStyle" version="13">
170c170
< <setting id="org.eclipse.jdt.core.formatter.tabulation.size" value="4"/>
---
> <setting id="org.eclipse.jdt.core.formatter.tabulation.size" value="2"/>
```

[intellij formatter](https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml)
```
< <code_scheme name="htsjdkStyle">
---
> <code_scheme name="GoogleStyle">
5c5
<       <option name="INDENT_SIZE" value="4" />
---
>       <option name="INDENT_SIZE" value="2" />
```

### DIFF
--- a/java-style-eclipse.xml
+++ b/java-style-eclipse.xml
@@ -1,0 +1,337 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<profiles version="13">
+<profile kind="CodeFormatterProfile" name="htsjdkStyle" version="1">
+<setting id="org.eclipse.jdt.core.formatter.comment.insert_new_line_before_root_tags" value="insert"/>
+<setting id="org.eclipse.jdt.core.formatter.disabling_tag" value="@formatter:off"/>
+<setting id="org.eclipse.jdt.core.formatter.insert_space_after_comma_in_annotation" value="insert"/>
+<setting id="org.eclipse.jdt.core.formatter.insert_space_before_comma_in_type_parameters" value="do not insert"/>
+<setting id="org.eclipse.jdt.core.formatter.insert_space_before_opening_brace_in_type_declaration" value="insert"/>
+<setting id="org.eclipse.jdt.core.formatter.insert_space_after_comma_in_type_arguments" value="insert"/>
+<setting id="org.eclipse.jdt.core.formatter.brace_position_for_anonymous_type_declaration" value="end_of_line"/>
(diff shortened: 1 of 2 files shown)


yfarjoun:
Given that we are not aware of maintainers who use eclipse, perhaps we only need the IntelliJ version?

mitochon:
@yfarjoun updated to remove eclipse formatter

lindenb:
@lindenb I am too late for this ? I use eclipse : I think it would be good for people writing PR

mitochon:
Someone can make a call 😃  

I can undo the last commit and put back the eclipse style guide. My personal opinion is its better to support more IDE options, as long as they are compatible with each other.

yfarjoun:
as long as everything is compatible with current guidelines, and we have active folks that use the style_guide, I think it would be good to have them....thanks @lindenb for chiming in!

mitochon:
All right, so I'll put the Eclipse style formatter back in per the above comment.

PR #1389, opened 2019-06-20
Output missing vcf fields as a single .

lbergelson:
### Description
* Previously vcf genotype fields with a specific number of elements were output as a comma separated list of dots
* This is valid, but uneccessarily long and doesn't provide any additional information.
* Now only a single dot will be emitted.
* ex:  .,.,. -> .

This is a follow up to #1371 
### Checklist

- [ ] Code compiles correctly
- [ ] All tests passing

### DIFF
--- a/src/test/java/htsjdk/variant/vcf/VCFEncoderTest.java
+++ b/src/test/java/htsjdk/variant/vcf/VCFEncoderTest.java
@@ -23,7 +23,7 @@ public class VCFEncoderTest extends HtsjdkTest {
 
 	@DataProvider(name = "VCFWriterDoubleFormatTestData")
 	public Object[][] makeVCFWriterDoubleFormatTestData() {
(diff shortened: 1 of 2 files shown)


src/test/java/htsjdk/variant/vcf/VCFEncoderTest.java
```diff
@@ -90,6 +90,11 @@ public void testVCFWriterDoubleFormatTestData(final double d, final String expec
         tests.add(new Object[]{keepMissing, vc, "./.:.:2", alleleMap, formatKeys});
         baseGT.noAttributes();
 
+        //check that we only produce a single . when writing attributes with multiple values instead of .,.
```
pshapiro4broad: please add whitespace before text in `//` comments
```suggestion
        // check that we only produce a single . when writing attributes with multiple values instead of .,.
```

PR #1390, opened 2019-06-21
Adding a code of conduct document for htsjdk.

lbergelson:
I think it's good to explicitly state that we intend to be an inclusive and welcoming project and that part of the responsibility of the maintainers is to maintain a positive community as well as just functional code.

This is the Contributor Covenant code of conduct which is the same one adopted by htsjdk-next.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,46 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as contributors and maintainers pledge to making participation in our project and our community a harassment-free experience for everyone, regardless of age, body size, disability, ethnicity, gender identity and expression, level of experience, nationality, personal appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Scope
+
+This Code of Conduct applies both within project spaces and in public spaces when an individual is representing the project or its community. Examples of representing a project or community include using an official project e-mail address, posting via an official social media account, or acting as an appointed representative at an online or offline event. Representation of a project may be further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be reported by contacting the project team at <contact@htsjdk.org>. The project team will review and investigate all complaints, and will respond in a way that it deems appropriate to the circumstances. The project team is obligated to maintain confidentiality with regard to the reporter of an incident. Further details of specific enforcement policies may be posted separately.
```
cmnbroad: Looks good, except we need to reify contact@htsjdk.org. Do we have an existing distribution we can use ?

cmnbroad: Its the same as the one in htsjdk-next ? Is it a real thing ?

lbergelson: Currently that email address goes to Me, MagicDGS, Jay, and Tim, we should probably update it to include whoever we think should be included on it. Or create a new one.

cmnbroad: Oh ok.

PR #1396, opened 2019-07-01
Optimize AbstractBAMFileIndex#query to avoid quadratic behavior when …

tomwhite:
…reading sequences in order.

### Description

Reading index sequences in the order they are defined in the header is a very common access pattern. However, the way it is implemented in `AbstractBAMFileIndex` means that it has quadratic complexity, since when reading sequence _i_ in `skipToSequence` that is not in the cache, it will start from the beginning, even if sequence _i_ - 1 is in the cache.

The effect is particularly stark when there are many reference sequences, such as hg38 which has over 3000. See https://github.com/disq-bio/disq/pull/109 for an example.

The fix is simple - look to see if sequence _i_ - 1 is in the cache, and start from that position if it is. The change is covered by existing tests.

/cc @lbergelson

### DIFF
--- a/src/main/java/htsjdk/samtools/AbstractBAMFileIndex.java
+++ b/src/main/java/htsjdk/samtools/AbstractBAMFileIndex.java
@@ -398,14 +398,23 @@ public abstract class AbstractBAMFileIndex implements BAMIndex {
         }
     }
 
-    private void skipToSequence(final int sequenceIndex) {
+    protected void skipToSequence(final int sequenceIndex) {
     	//Use sequence position cache if available
     	if(sequenceIndexes[sequenceIndex] != -1){
     		seek(sequenceIndexes[sequenceIndex]);
     		return;
     	}
+
+        // Use previous sequence position if in cache
+        final int startSequenceIndex;
+        if (sequenceIndex > 0 && sequenceIndexes[sequenceIndex - 1] != -1) {
+            seek(sequenceIndexes[sequenceIndex - 1]);
+            startSequenceIndex = sequenceIndex - 1;
+        } else {
+            startSequenceIndex = 0;
+        }
     	
-        for (int i = 0; i < sequenceIndex; i++) {
+        for (int i = startSequenceIndex; i < sequenceIndex; i++) {
             // System.out.println("# Sequence TID: " + i);
             final int nBins = readInteger();
             // System.out.println("# nBins: " + nBins);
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/AbstractBAMFileIndex.java
```diff
@@ -398,14 +398,23 @@ protected void readChunks(int nChunks, List<Chunk> chunks) {
...
 
-    private void skipToSequence(final int sequenceIndex) {
+    protected void skipToSequence(final int sequenceIndex) {
     	//Use sequence position cache if available
     	if(sequenceIndexes[sequenceIndex] != -1){
     		seek(sequenceIndexes[sequenceIndex]);
     		return;
     	}
+
+        // Use previous sequence position if in cache
```
lbergelson: Could you add an extra line pointing out how this helps in the common case of querying in sequential order.

src/main/java/htsjdk/samtools/AbstractBAMFileIndex.java
```diff
@@ -398,14 +398,23 @@ protected void readChunks(int nChunks, List<Chunk> chunks) {
         }
     }
 
-    private void skipToSequence(final int sequenceIndex) {
+    protected void skipToSequence(final int sequenceIndex) {
```
lbergelson: This introduced an name conflict.
```
executing task ':compileJava' ...
/home/travis/build/samtools/htsjdk/src/main/java/htsjdk/samtools/CSIIndex.java:473: error: skipToSequence(int) in CSIIndex cannot override skipToSequence(int) in AbstractBAMFileIndex
    private void skipToSequence(final int sequenceIndex) {
                 ^
  attempting to assign weaker access privileges; was protected
```

src/main/java/htsjdk/samtools/AbstractBAMFileIndex.java
tomwhite: Done

src/main/java/htsjdk/samtools/AbstractBAMFileIndex.java
tomwhite: Oops - fixed now

tomwhite:
I've fixed this now. I wasn't sure how to add a direct test. Any suggestions, or is the existing coverage sufficient?

lbergelson:
@tomwhite I wrote a test here:  https://github.com/samtools/htsjdk/tree/skip-to-sequence-optimization

lbergelson:
Closing since I merged  #1397 which includes this.

PR #1408, opened 2019-08-13
Make AbstractFastaSequenceFile serializable by Kryo for Spark.

tomwhite:
See explanation at https://github.com/EsotericSoftware/kryo/issues/469

This change is a fix for https://github.com/broadinstitute/gatk/issues/6091, which is a regression in GATK running on Spark. I've tested the fix manually with GATK.

/cc @lbergelson


lbergelson:
@tomwhite Is this happening now because of the kryo update or because of htsjdk changes?

tomwhite:
@lbergelson htsjdk certainly has a change in the last release that caused this. Whether going from Kryo 3 to 4 also made a difference, I'm not sure. (Thanks for merging.)

PR #1409, opened 2019-08-14
Suppress all exceptions during DeleteOnExitPathHook

lbergelson:
### Description

* Currently we were not catching some commonly thrown exceptions like RuntimeIOException while deleting files at shutdown.
  This results in confusing error messages at program exit which can mask the actual cause of failures.
  Now we catch all exceptions and output a log message at debug level in case anyone is interested in seeing the failures.


lbergelson:
I don't know why it wouldn't.  It's basically just writing to system.err so unless you're 100% out of memory it doesn't seem like it would be problematic.  I googled for restrictions on shutdown hooks and the only thing that's definitely not allowed is registering a new shutdown hook.   I'm definitely not an expert on it though, so I might be wrong...

PR #1410, opened 2019-08-19
Updated the max isize to be Integer.MAX_VALUE.

tfenne:
### Description

I'm processing some data with a reference sequence that has long chromosomes, over the BAI limit.  I ran into this issue where one (likely chimeric) read-pair has an insert size of over 536,870,912 (aka `1<<29`), which causes HTSJDK to throw validation errors.

I looked at the spec and can find nothing supporting a limit on insert size this low.  The SAM section defines it as between `[−2^31 + 1, 2^31 − 1]`, and the BAM section dedicates a full `int32 t` for tlen.

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMRecord.java
+++ b/src/main/java/htsjdk/samtools/SAMRecord.java
@@ -158,7 +158,7 @@ public class SAMRecord implements Cloneable, Locatable, Serializable {
     /**
      * abs(insertSize) must be <= this
      */
-    public static final int MAX_INSERT_SIZE = 1<<29;
+    public static final int MAX_INSERT_SIZE = Integer.MAX_VALUE;
 
     /**
      * Tags that are known to need the reverse complement if the read is reverse complemented.


lbergelson:
Seems likely to be an artifact of the bai restrictions, but the history of its creation is lost to git history.

lbergelson:
Lets see if @yfarjoun can remember any objections too this.  I can't think of any though.

jmarshall:
Historical specification note: this field was initially neglected and only lifted from 2<sup>29</sup> to 2<sup>31</sup> a couple of years after most of the other fields (see samtools/hts-specs@6f8dfe41814df32de75bf16cf2719148c939ae3f). This may go some way to explaining the still-present HTSJDK limit on this field — but it was still lifted almost six years ago!

lbergelson:
@jmarshall Thanks you.  Sometimes you just have to give things time to, you don't want to rush anything.

tfenne:
Thanks @jmarshall - I figured it was something like that and I appreciate the archeological help!

jmarshall:
Tragically I still remember making that change :smile:

PR #1412, opened 2019-08-22
Add LiftOver constructor that takes an input stream

kevinpetersavage:
**Desrciption**

Add LiftOver constructor that takes an input stream. This is so that it can be used in situations where the config is not on a file system. In my case it is on AWS S3. 

**Checklist**

- [x]  Code compiles correctly
- [x]  New tests covering changes and new functionality
- [x]  All tests passing

### DIFF
--- a/src/main/java/htsjdk/samtools/liftover/Chain.java
+++ b/src/main/java/htsjdk/samtools/liftover/Chain.java
@@ -36,6 +36,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.regex.Pattern;
 
+import static htsjdk.samtools.util.IOUtil.openFileForReading;
+
 /**
  * Holds a single chain from a UCSC chain file.  Chain file format is described here: http://genome.ucsc.edu/goldenPath/help/chain.html
  *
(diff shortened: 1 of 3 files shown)


src/main/java/htsjdk/samtools/liftover/LiftOver.java
```diff
@@ -103,6 +111,11 @@ public LiftOver(File chainFile) {
         }
     }
 
+    private static InputStream toInputStream(File chainFile) {
```
lbergelson: If you make the change I suggested above you can remove this method, but you probably want to push the assertFileIsReadable check down into `Chain.loadChains(File)`

src/main/java/htsjdk/samtools/liftover/LiftOver.java
```diff
@@ -84,9 +85,16 @@ public long getFailedIntervalsBelowThreshold() {
     /**
      * Load UCSC chain file in order to lift over Intervals.
      */
-    public LiftOver(File chainFile) {
-        IOUtil.assertFileIsReadable(chainFile);
-        chains = Chain.loadChains(chainFile);
+    public LiftOver(File chainFile){
+        this(toInputStream(chainFile), chainFile.toString());
```
lbergelson: It seems like java constructor constraints have forced some slightly awkward behavior here.

I would probably restructure like this in order to avoid some awkwardness.

```
 /**
     * Load UCSC chain file in order to lift over Intervals.
     */
    public LiftOver(File chainFile){
        this(Chain.loadChains(chainFile));
    }

    /**
     * Load UCSC chain file in order to lift over Intervals.
     */
    public LiftOver(InputStream chainFileInputStream, String sourceName) {
        this(Chain.loadChains(new BufferedLineReader(chainFileInputStream), sourceName));
    }

    private LiftOver(OverlapDetector<Chain> chains) {
        this.chains = chains;
        for (final Chain chain : this.chains.getAll()) {
            final String from = chain.fromSequenceName;
            final String to   = chain.toSequenceName;
            final Set<String> names;
            if (contigMap.containsKey(from)) {
                names = contigMap.get(from);
            }
            else {
                names = new HashSet<>();
                contigMap.put(from, names);
            }
            names.add(to);
        }
    }
```

That way there's no question about who's responsibility it is to close streams.  (sorry to harp on this particular issue, it's just something that's come up a bunch recently so it's in my mind...)

src/main/java/htsjdk/samtools/liftover/Chain.java
```diff
@@ -306,16 +308,19 @@ public int hashCode() {
         return result;
     }
 
+    static OverlapDetector<Chain> loadChains(final File chainFile) {
+        return loadChains(new BufferedLineReader(openFileForReading(chainFile)), chainFile.toString());
```
lbergelson: As mentioned below you probably want to push the IOUtil.assertFileIsReadable check into this method.

kevinpetersavage:
@lbergelson thanks for the really good feedback, definitely improved the code!

PR #1413, opened 2019-08-26
Allow partially spanned alleles.

tfenne:
### Description

@lbergelson this is likely to be a slightly contentious change despite the small PR.  The goal here is to support partially spanned alleles in the form they are used in [octopus](https://github.com/luntergroup/octopus).  See https://github.com/luntergroup/octopus/issues/75 for a description of what's going on, but the short version is that Octopus outputs alleles like `*AT` and `**GCGA` to represent cases where the first or first two (respectively) bases of the allele are spanned by an upstream event.

This isn't prohibited by the VCF spec, although it seems like it's an unintentional thing that alleles like this are spec compliant.  Still, I'd like to be able to use HTSJDK's VCF API to read octopus VCFs and it currently blows up on records with alleles like this.

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/Allele.java
+++ b/src/main/java/htsjdk/variant/variantcontext/Allele.java
@@ -382,7 +382,7 @@ public class Allele implements Comparable<Allele>, Serializable {
 
         for (byte base :  bases ) {
             switch (base) {
-                case 'A': case 'C': case 'G': case 'T':  case 'a': case 'c': case 'g': case 't': case 'N' : case 'n' :
+                case 'A': case 'C': case 'G': case 'T': case 'a': case 'c': case 'g': case 't': case 'N': case 'n': case '*':
                     break;
                 default:
                     return false;
--- a/src/test/java/htsjdk/variant/variantcontext/AlleleUnitTest.java
+++ b/src/test/java/htsjdk/variant/variantcontext/AlleleUnitTest.java
@@ -104,6 +104,12 @@ public class AlleleUnitTest extends VariantBaseTest {
     }
 
     @Test
+    public void testPartiallySpannedAllele() {
+        final Allele partial = new Allele("*ATAT", false);
+        Assert.assertEquals(partial.getDisplayString(), "*ATAT");
+    }
+
+    @Test
     public void testCreatingIndelAlleles() {
         Assert.assertEquals(ATIns.length(), 2);
         Assert.assertEquals(ATCIns.length(), 3);


dancooke:
I think the best case that I can make is that it provides a neater definition that is more in line with the rest of the specification. The working definition that I've used for `*` is something like

    This base is explained by an upstream (or matching `POS`) record.

While the meaning currently being used by GATK *appears* to be more like

    Some or all of the leading reference bases in this allele are deleted - as explained by an upstream (or matching `POS`) record.

So suddenly we go from defining `ALT` alleles in terms of nucleotides to defining them in terms of both nucleotides and symbolic *alleles* (but distinct from the other symbolic alleles the spec permits). I'm not suggesting both definitions can't result in semantically equivalent information (w.r.t describing haplotypes), but I think the former is both clearer and more expressive.

A more practical benefit of the base-`*` definition is that reconstructing haplotypes from VCF is a little easier as there's no need to keep track of information from previous records (other than the sequence), since each record specifies exactly how many reference bases must be ignored.

jmarshall:
@dancooke: What do you mean by ‘upstream record’? Another record earlier in the same VCF file?

(Referring to other records seems problematic if the VCF file later gets filtered…)

dancooke:
@jmarshall A record with smaller or equal `POS` (so yes, an earlier record in the same VCF if it's co-ordinate sorted). Problematic how, for reconstructing haplotypes? How is this different to the 'symbolic' interpretation? If a previous record is filtered, the reference is implicitly implied. RTGTools vcfeval (3.10.1) supports partially overlapping alleles with `*` and I haven't encountered any issues with filtered records.

lbergelson:
@tfenne Sorry, I didn't mean to totally shut this down in htsjdk.  I figured you were going to come back with a change that adds a flag in defaults.   

I'm happy to accept a pr that adds an opt in for this, I just don't want picard/gatk suddenly falling over in confusing ways when they encounter these things.  If we decide to not ban it in the spec we can come back and add better support in a later version.  

I don't want to make any decision about it in the spec before I talk with a few people who are on vacation right now.  People are trickling back in after labor day.

tfenne:
@lbergelson no worries - I wasn't taking this as 100% blocked, but rather figured it was useful to try and push the spec discussion forward some because it would be better to have the spec clarified and then just do the right thing here.  If I get to the point where I'm blocked before the spec issue resolves I'll take a look at making this opt in.  Sorry if I wasn't clear.

yfarjoun:
I like the idea of having a STAR_IS_A_BASE flag in the Defaults class to enable reading of Octopus vcfs. Concurrently there should be discussion in hts-specs about the base-status of star.

yfarjoun:
@tfenne this is in your court now to implement a "Defaults" parameter to unlock these "malformed" inputs.

PR #1415, opened 2019-08-28
Fix issue 1414- add proper behavior of clear() to DiskBackedQueue

biotinker:
### Description

Fix bug in DiskBackedQueue.

DiskBackedQueue.closeIOResources() was using IOUtil.deletePaths() to clean up its temp file, but IOUtil.deletePaths() splits a given path into an array of filenames, and there is no function in IOUtil that would properly remove this one path (without needlessly wrapping the temp file in an iterator).

I copied the try/catch that does the file deletion in IOUtil.deletePaths() to the correct place within closeIOResources, and it now works as expected.

### DIFF
--- a/src/main/java/htsjdk/samtools/util/DiskBackedQueue.java
+++ b/src/main/java/htsjdk/samtools/util/DiskBackedQueue.java
@@ -339,9 +339,14 @@ public class DiskBackedQueue<E> implements Queue<E> {
     private void closeIOResources() {
         CloserUtil.close(this.outputStream);
         CloserUtil.close(this.inputStream);
-        if (this.diskRecords != null) IOUtil.deletePaths(this.diskRecords);
+        if (this.diskRecords != null){
+            try {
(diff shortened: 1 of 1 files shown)


lbergelson:
@biotinker Thank you for identifying this problem and producing a patch!  This pr looks like it solves the problem, but I think in this case changing the underlying problem in the method is justified.

The behavior you identified in `IOUtil.deletePaths(Iterable<Path>)` is REALLY bad.  It was essentially a copy of the old method `IOUtil.deleteFile` and no one noticed that Path is itself and `Iterator<Path>`.  

Rather than working around the pathological behavior I'm opening a new PR to fix the underlying problem with the method since it's completely broken.  

Would you mind reviewing https://github.com/samtools/htsjdk/pull/1416 and making sure it solves your problem?

lbergelson:
@biotinker I merged the alternate pr that made fixes to the underlying method.  Thank you for identifying this and opening a pr to fix it!

Issue #389, opened 2015-11-18
VCFContigHeaderLine throwing Exception although the spec doesn't require it


jmmut:
The field "length" in a contig header line, is documented as required:

> getSequenceDictionary
> 
> public SAMSequenceDictionary getSequenceDictionary()
> 
>    Returns the contigs in this VCF file as a SAMSequenceDictionary. Returns null if contigs lines are not present in the header. Throws SAMException if one or more contig lines do not have length information. 

but the specifications of VCF 4.2 and 4.3 don't require it:

4.2:

> 1.2.7    Contig Field format
> As with chromosomal sequences it is highly recommended (but not required) that the header include tags describing
> the contigs referred to in the VCF file.  [...]

4.3:

> 1.4.7 Contig field format 
> The structured contig field must include the ID attribute and typically includes also sequence length, [...]

PR #1418, opened 2019-09-19
Make VCFHeader not throw exception if contig header lines lack length field

cwhelan:
### Description

`VCFHeader.getSequenceDictionary` currently throws an exception if one or more of the `Contig` header lines does not have a length field, despite the fact that:

- The VCF spec (v 4.2 and 4.3) does not require a length field (although it states that they are typically present)
- If contig lines are not present in the header it simply returns null

This generally has the effect of making htsjdk tools unable to process files missing length fields without re-headering, which is not always feasible, for example in the case of large multi-sample files in cloud storage.

Fixes https://github.com/samtools/htsjdk/issues/389

- [ ] Check your code style.

### DIFF
--- a/src/main/java/htsjdk/variant/vcf/VCFContigHeaderLine.java
+++ b/src/main/java/htsjdk/variant/vcf/VCFContigHeaderLine.java
@@ -76,9 +76,14 @@ public class VCFContigHeaderLine extends VCFSimpleHeaderLine {
         return contigIndex;
     }
 
+    /**
+     * Get the SAMSequenceRecord that corresponds to this VCF header line.
+     * @return The SAMSequenceRecord containing the ID, length, assembly, and index of this contig. Returns null if the
+     * contig header line does not have a length.
+     */
 	public SAMSequenceRecord getSAMSequenceRecord() {
 		final String lengthString = this.getGenericFieldValue("length");
-		if (lengthString == null) throw new TribbleException("Contig " + this.getID() + " does not have a length field.");
+		if (lengthString == null) return null;
 		final SAMSequenceRecord record = new SAMSequenceRecord(this.getID(), Integer.parseInt(lengthString));
         record.setAssembly(this.getGenericFieldValue("assembly"));
 		record.setSequenceIndex(this.contigIndex);
--- a/src/main/java/htsjdk/variant/vcf/VCFContigHeaderLine.java
+++ b/src/main/java/htsjdk/variant/vcf/VCFContigHeaderLine.java
@@ -78,13 +78,21 @@ public class VCFContigHeaderLine extends VCFSimpleHeaderLine {
 
     /**
      * Get the SAMSequenceRecord that corresponds to this VCF header line.
+     * If the VCF header line does not have a length tag, the SAMSequenceRecord returned will be set to have a length of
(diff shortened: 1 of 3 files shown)


src/main/java/htsjdk/variant/vcf/VCFContigHeaderLine.java
```diff
@@ -76,9 +76,14 @@ public Integer getContigIndex() {
...
 
+    /**
+     * Get the SAMSequenceRecord that corresponds to this VCF header line.
+     * @return The SAMSequenceRecord containing the ID, length, assembly, and index of this contig. Returns null if the
+     * contig header line does not have a length.
+     */
 	public SAMSequenceRecord getSAMSequenceRecord() {
 		final String lengthString = this.getGenericFieldValue("length");
-		if (lengthString == null) throw new TribbleException("Contig " + this.getID() + " does not have a length field.");
+		if (lengthString == null) return null;
```
cmnbroad: SAMSequenceRecord has an `UNKNOWN_SEQUENCE_LENGTH` value specifically for the case where the length is missing. It would be preferable to use that here and return a `SAMSequenceRecord`, rather than returning null in cases where it previously would have thrown.

cmnbroad: It might be worth adding a comment noting the SAMSequenceRecord.isSameSequence treats a record created this way (with a length of `UNKNOWN_SEQUENCE_LENGTH`) as matching another sequence record with the same name but a valid length.

cwhelan:
Thanks for pointing me to `SAMSequenceRecord.UNKNOWN_SEQUENCE_LENGTH` @cmnbroad . I've changed this to return a `SAMSequenceRecord` with length set to that.

src/main/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -270,7 +270,9 @@ public SAMSequenceDictionary getSequenceDictionary() {
 
         final List<SAMSequenceRecord> sequenceRecords = new ArrayList<SAMSequenceRecord>(contigHeaderLines.size());
         for (final VCFContigHeaderLine contigHeaderLine : contigHeaderLines) {
-            sequenceRecords.add(contigHeaderLine.getSAMSequenceRecord());
+            final SAMSequenceRecord samSequenceRecord = contigHeaderLine.getSAMSequenceRecord();
```
lbergelson: How do you end up with nulls here?  Shouldn't it be returning lines with UNKNOWN_SEQUENCE_LENGTH?

src/main/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -261,7 +261,7 @@ public void addMetaDataLine(final VCFHeaderLine headerLine) {
 
     /**
      * Returns the contigs in this VCF file as a SAMSequenceDictionary. Returns null if contigs lines are
-     * not present in the header. Throws SAMException if one or more contig lines do not have length
+     * not present in the header. Returns null if one or more contig lines do not have length
```
lbergelson: I don't think this comment is true.

lbergelson:
Looks good otherwise.

cwhelan:
Sorry for the oversight @lbergelson. Pushed a new commit to `VCFHeader` to bring it up to date with the change to using `UNKNOWN_LENGTH_SEQUENCE`.

PR #1427, opened 2019-10-17
Yf add platforms

yfarjoun:
Adding legal values for the PL tag of the @rg SAM header line (BGI and OTHER)


yfarjoun:
not sure why...afterall, we have things that are not in the spec (like sortorder=duplicate)

PR #1430, opened 2019-10-25
Path-enable IndexFactory.

cmnbroad:
Prerequisite for https://github.com/broadinstitute/gatk/issues/6161.

### DIFF
--- a/src/main/java/htsjdk/tribble/index/IndexFactory.java
+++ b/src/main/java/htsjdk/tribble/index/IndexFactory.java
@@ -46,15 +46,17 @@ import htsjdk.tribble.index.tabix.TabixIndexCreator;
 import htsjdk.tribble.readers.PositionalBufferedStream;
 import htsjdk.tribble.util.LittleEndianInputStream;
 import htsjdk.tribble.util.ParsingUtils;
+import htsjdk.utils.ValidationUtils;
 
 import java.io.BufferedInputStream;
 import java.io.EOFException;
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.channels.SeekableByteChannel;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/tribble/index/IndexFactory.java
```diff
@@ -224,7 +229,20 @@ else if (indexFile.endsWith(FileExtensions.TABIX_INDEX)) {
      * @param codec     the codec to use for decoding records
      */
     public static LinearIndex createLinearIndex(final File inputFile, final FeatureCodec codec) {
-        return createLinearIndex(inputFile, codec, LinearIndexCreator.DEFAULT_BIN_WIDTH);
+        return createLinearIndex(
```
lbergelson: You can use IOUtil.toPath(file) which does a null safe conversion to path and then just have the null check in the final one.

src/main/java/htsjdk/tribble/index/IndexFactory.java
```diff
@@ -434,17 +540,30 @@ private static void checkSorted(final File inputFile, final Feature lastFeature,
...
+        public FeatureIterator(final Path inputPath, final FeatureCodec<FEATURE_TYPE, SOURCE> codec) {
+            if (inputPath == null) {
                 throw new IllegalArgumentException("FeatureIterator input file cannot be null");
             }
             this.codec = codec;
-            this.inputFile = inputFile;
+
+            // We _MUST_ call getPathToDataFile here to preserve behavior of codecs that rely on config files
+            // that point to other files to be indexed. This is CRITICALLY IMPORTANT to downstream tools.
+            this.inputPath = Paths.get(codec.getPathToDataFile(inputPath.toUri().toString()));
```
lbergelson: ugh...  path -> uri -> string -> string -> path   Not much to do about it though :(

src/test/java/htsjdk/tribble/index/IndexFactoryTest.java
```diff
@@ -63,7 +65,7 @@
     }
 
     @Test(dataProvider = "bedDataProvider")
-    public void testCreateLinearIndexFromBED(final File inputBEDFIle) throws Exception {
+    public void testCreateLinearIndexFromBED(final File inputBEDFIle) {
```
pshapiro4broad: Is it intentional to leave the `File` test code in place? I was expecting to see all tests updated to use `Path` instead.

src/main/java/htsjdk/tribble/index/IndexFactory.java
cmnbroad: Yeah, I considered changing `getPathToDataFile` to return a `Path` instead of a String, but decided that was too much scope creep.

src/test/java/htsjdk/tribble/index/IndexFactoryTest.java
cmnbroad: I could easily be persuaded, but the `File` overloads are still used by the rest of the world, including other (non-IndexFactory) htsjdk tests, Picard, and for the moment GATK. Until we remove or deprecate them (should we ?), these tests exercise both code paths.

lbergelson: I think we should still test file paths, they get forwarded to the path versions anyway so it's not like we're losing coverage on the path implementation.  People still use the file versions so if we break them it's a problem.

cmnbroad:
@lbergelson I don't think the order in which we merge these matters, but we should probably wait for @jonn-smith 's redirect test before we merge either one.

src/main/java/htsjdk/tribble/index/IndexFactory.java
cmnbroad: Oh yeah - done.

lbergelson:
Sorry, I merged mine which caused a conflict :(

lbergelson:
@cmnbroad I merged the xsv fix.

cmnbroad:
@lbergelson Rebased and conflicts resolved.

PR #1431, opened 2019-10-25
Replace IndexFactory reflection.

lbergelson:
* IndexFactory used old reflection code that was unnecessarily brittle and may cause problems in java 11.
  Replaced it by using direct references to Index constructor methods instead of finding them with reflection.
* Breaking changes: Removed public methods IndexType.getIndexCreator() and IndexType.getIndexType().  The first of these was broken and always resulted in an exception and the second is mostly useless and also unlikely to be missed.  

* It would be good to look at canCreate and understand what it's purpose is and if it still makes sense.
The whole group of IndexConstructors/Indexes/IndexFactory methods seems like there is something fundamentally wrong with the design.

- [ ] Check your code style.


lbergelson:
@cmnbroad Some cleanup in this class.  It's a similar change to https://github.com/samtools/htsjdk/pull/1421

src/main/java/htsjdk/tribble/index/IndexFactory.java
```diff
@@ -29,27 +29,22 @@
...
-import java.io.BufferedInputStream;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.io.EOFException;
-import java.io.InputStream;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
+import java.io.*;
```
pshapiro4broad: In general I think we should prefer non-wildcard imports to make it easier to find out where a symbol comes from.

src/main/java/htsjdk/tribble/index/IndexFactory.java
```diff
@@ -73,47 +68,41 @@
...
-            }
-        }
-
         public boolean canCreate() {
-            return indexCreatorClass != null;
+            return canCreate;
         }
 
-        IndexType(final int magicNumber, final Integer tribbleIndexType, final Class creator, final Class indexClass, final int defaultBinSize) {
+        private interface IOFunction<T,R>{ R apply(T t) throws IOException;}
```
pshapiro4broad: This would be easier to read if it's formatted across lines
```
private interface IOFunction<T, R> {
    R apply(T t) throws IOException;
}
```

Also isn't `T` `InputStream` and `R` is always a subtype of `AbstractIndex`? Is there a reason to use a type abstraction here?

src/main/java/htsjdk/tribble/index/IndexFactory.java
```diff
@@ -306,8 +292,8 @@ public static LinearIndex createLinearIndex(final File inputFile, final FeatureC
             case INTERVAL_TREE: return createIntervalIndex(inputFile, codec);
             case LINEAR:        return createLinearIndex(inputFile, codec);
             case TABIX:         return createTabixIndex(inputFile, codec, sequenceDictionary);
+            default: throw new IllegalArgumentException("Unrecognized IndexType " + type);
```
pshapiro4broad: If this was a method call on `IndexType` then you wouldn't need a `switch` here or a `default` case.

src/main/java/htsjdk/tribble/index/IndexFactory.java
```diff
@@ -189,21 +178,18 @@ public static Index loadIndex(final String indexFile, Function<SeekableByteChann
...
-            if (ex.getCause() instanceof EOFException) {
+            return createIndex(bufferedInputStream);
+        } catch (final EOFException ex) {
                 throw new TribbleException.CorruptedIndexFile("Index file is corrupted", source, ex);
-            }
-            throw new RuntimeException(ex);
-        } catch (final Exception ex) {
-            throw new RuntimeException(ex);
+        } catch (final IOException ex) {
+            throw new RuntimeIOException("Failed to read index file: " + source, ex);
```
pshapiro4broad: Is there a reason why this is `RuntimeException` not `TribbleException`?

src/main/java/htsjdk/tribble/index/IndexFactory.java
```diff
@@ -73,47 +68,41 @@
...
         }
 
         public Integer getTribbleIndexType() {
             return tribbleIndexType;
         }
 
-        public Class<Index> getIndexType() {
-            return indexType;
+        public Index createIndex(InputStream in) throws IOException{
+           return this.createFromInputStream.apply(in);
```
pshapiro4broad: This class doesn't seem to use `this.` prefix except when necessary, e.g. in constructors.

src/main/java/htsjdk/tribble/index/IndexFactory.java
lbergelson: Oh, good call, I changed computers and lost intellij preferences.

src/main/java/htsjdk/tribble/index/IndexFactory.java
lbergelson: If I call it IOFunction and make it abstract then I don't have to think of an name for the concretized type.  I'll change.

src/main/java/htsjdk/tribble/index/IndexFactory.java
lbergelson: Historical weirdness.  Could be a TribbleException instead.

src/main/java/htsjdk/tribble/index/IndexFactory.java
lbergelson: done

src/main/java/htsjdk/tribble/index/IndexFactory.java
lbergelson: Yeah.  That's a really good point.  The whole structure here is super weird and could be rationalized.  I was doing the most braindead fix I could, but maybe I should move more stuff around.

src/main/java/htsjdk/tribble/index/IndexFactory.java
```diff
@@ -73,47 +78,43 @@
...
             this.defaultBinSize = defaultBinSize;
         }
 
         public Integer getTribbleIndexType() {
             return tribbleIndexType;
         }
 
-        public Class<Index> getIndexType() {
-            return indexType;
+        public Index createIndex(InputStream in) throws IOException{
```
cmnbroad: It seems like this should reject creation of a TABIX index (honor `canCreate()`) since it doesn't have the sequence dictionary? I think the previous (index creator) method would have blown up on TABIX, but now I'm not sure what will happen with this one. Also, since its public perhaps it should wrap IOException with RuntimeIOException, and `final` on arg.

lbergelson: I don't really understand what `canCreate()` is supposed to mean.  You can totally create a tabixIndex from a stream without providing a sequence dictionary.  The previous method that used getIndexType() worked on tabix as far as I can see and didn't check canCreate().   I suspect that canCreate may just be completely outdated and irrelevant, but I didn't want to figure it out as part of stripping out the bad reflection.

cmnbroad: Ok, fair enough.

PR #1433, opened 2019-10-30
add method we killed accidentally

lbergelson:
We accidentally deleted a method when adding path options, adding it and a missed Pathification.


lbergelson:
@cmnbroad Can you take a quick look at this?

lbergelson:
No problem.  This is why I run automated comparisons before doing a release...

PR #1437, opened 2019-11-20
Improve calculation of available memory when sizing buffers in SortingCollection

tfenne:
### Description

SortingCollection keeps insisting that it doesn't have enough memory to buffer files for reading. But this is testably false - if you keep the same number of items in memory, same number of files, and increase the heap size, it continues to insist there is not enough space to buffer.

The existing implementation uses a method that returns the amount of free memory _in the currently allocated heap_.  It does not take into account that the heap may not have been expanded to it's max size as specified by `-Xmx`.

- [ ] Check your code style.

### DIFF
--- a/src/main/java/htsjdk/samtools/util/SortingCollection.java
+++ b/src/main/java/htsjdk/samtools/util/SortingCollection.java
@@ -524,11 +524,16 @@ public class SortingCollection<T> implements Iterable<T> {
         // enough memory for buffering it will return zero and all reading will be unbuffered.
         private int checkMemoryAndAdjustBuffer(int numFiles) {
             int bufferSize = Defaults.BUFFER_SIZE;
+
             // garbage collect so that our calculation is accurate.
-            Runtime.getRuntime().gc();
+            final Runtime rt = Runtime.getRuntime();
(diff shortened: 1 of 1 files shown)


tfenne:
Hey @lbergelson.  For your first question, I think that's really old javadoc that is true but misleading.  My understanding is that most JVMs have a default maximum heap size that is not infinity or the physical memory on a machine.  E.g. for me running oracle jdk8 with no -Xmx (and no JAVA_OPTS) I see a max memory of ~7.1GB (on a machine with 32GB of RAM).

Re: (2) my reading of that post and experience is that `maxMemory()` actually gives you an underestimate.  I.e. you say `-Xmx8g` and `maxMemory()` reports 7.1GB.

Honestly I think this whole strategy of trying to estimate the available memory is questionable.  `Runtime.gc()` is asynchronous, so even if you call it there's zero guarantee that space is reclaimed before calling other methods.  Probably the correct solution here is to iteratively attempt to allocate the buffers, reducing size until it succeeds.  But that's a bigger change and I'm just trying to make it so that I stop getting unbuffered readers all the time, which is really killing performance.

PR #1441, opened 2019-11-27
Reduce test output from BlockCompressedInputStreamTest test method args.

cmnbroad:
We're within about 20k of the 5MB log output limit on travis - with my cram branch we go way over. A lot of it is the expansion of test method arg values. This change suppresses about 500k of output by wrapping large list inputs in a `Supplier`.


cmnbroad:
Not sure if there is a better solution, but these changes reduce the log output by about 1MB.

PR #1447, opened 2019-12-20
Modified numLikelihoods calculation for performance

### LINKED ISSUE #1446: GenotypeLikelihoods.calcNumLikelihoods() should not be recursive
`GenotypeLikelihoods.calcNumLikelihoods(numAlleles, ploidy)`, which is the method used to calculate the length of the PL array given the number of alleles and the ploidy, is currently written using an inefficient recursive implementation:

```
    private static final int calcNumLikelihoods(int numAlleles, int ploidy) {
        if (numAlleles == 1) {
            return 1;
        } else if (ploidy == 1) {
            return numAlleles;
        } else {
            int acc = 0;

            for(int k = 0; k <= ploidy; ++k) {
                acc += calcNumLikelihoods(numAlleles - 1, ploidy - k);
            }

            return acc;
        }
    }
```

However, this method has an easy closed-form combinatorics solution using the formula for selection with replacement:

image

So (for example), for numAlleles = 5 and ploidy = 2, it is just `(5 + 2 - 1) choose 2` = `6 choose 2` = 15

(Thanks to @fleharty and @kachulis for their help in deriving this!)

Since this method is called every time we create a PL array (which is very frequently!), it seems worth it to get rid of the recursion in favor of a one-line call out to a standard combinatorics function.

KevinCLydon:
Fixes #1446

### Description

- [ ] Check your code style.

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
+++ b/src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
@@ -39,8 +39,6 @@ import java.util.Map;
 import java.util.HashMap;
 
 public class GenotypeLikelihoods {
-    private final static int NUM_LIKELIHOODS_CACHE_N_ALLELES = 5;
-    private final static int NUM_LIKELIHOODS_CACHE_PLOIDY = 10;
     // caches likelihoods
     private final static GenotypeNumLikelihoodsCache numLikelihoodCache = new GenotypeNumLikelihoodsCache();
 
@@ -53,18 +51,6 @@ public class GenotypeLikelihoods {
     //
     private double[] log10Likelihoods = null;
     private String likelihoodsAsString_PLs = null;
-
-    /**
-     * initialize num likelihoods cache to up to 5 alleles and 10 ploidy
-     */
-    static {
-        // must be done before PLIndexToAlleleIndex
-        for ( int numAlleles = 1; numAlleles < NUM_LIKELIHOODS_CACHE_N_ALLELES; numAlleles++ ) {
-            for ( int ploidy = 1; ploidy < NUM_LIKELIHOODS_CACHE_PLOIDY; ploidy++ ) {
-                numLikelihoodCache.put(numAlleles,ploidy, calcNumLikelihoods(numAlleles, ploidy));
-            }
-        }
-    }
 
     /**
      * The maximum number of diploid alternate alleles that we can represent as genotype likelihoods
@@ -442,10 +428,10 @@ public class GenotypeLikelihoods {
      * @param ploidy        number of chromosomes
      * @return  number of likelihoods
      */
-    private static final int calcNumLikelihoods(final int numAlleles, final int ploidy) {
+    static final int calcNumLikelihoods(final int numAlleles, final int ploidy) {
         //Note: Casting to int instead instead of returning long because values above Integer.MAX_VALUE would not be valid array indices,
         // and would cause other problems if a PL array needed to be that size
-        return (int)(BinomialCoefficientUtil.binomialCoefficient((numAlleles + ploidy - 1),ploidy));
+        return Math.toIntExact(BinomialCoefficientUtil.binomialCoefficient((numAlleles + ploidy - 1),ploidy));
     }
 
     /**
@@ -472,16 +458,9 @@ public class GenotypeLikelihoods {
      *   @param  ploidy          Ploidy, or number of chromosomes in set
      *   @return    Number of likelihood elements we need to hold.
      */
-    public static int numLikelihoods(final int numAlleles, final int ploidy) {
-        //Check cache
-        Integer numLikelihoods = numLikelihoodCache.get(numAlleles, ploidy);
-        if(numLikelihoods != null){
-            return numLikelihoods;
-        }
-        // have to calculate on the fly
-        numLikelihoods = calcNumLikelihoods(numAlleles, ploidy);
-        numLikelihoodCache.put(numAlleles, ploidy, numLikelihoods);
-        return numLikelihoods;
+    public static synchronized int numLikelihoods(final int numAlleles, final int ploidy) {
+        //Get the value from the cache
+        return numLikelihoodCache.get(numAlleles, ploidy);
     }
 
     // As per the VCF spec: "the ordering of genotypes for the likelihoods is given by: F(j/k) = (k*(k+1)/2)+j.
--- a/src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
+++ b/src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
@@ -26,6 +26,7 @@
 package htsjdk.variant.variantcontext;
 
 import htsjdk.tribble.TribbleException;
+import htsjdk.variant.utils.BinomialCoefficientUtil;
 import htsjdk.variant.utils.GeneralUtils;
 import htsjdk.variant.vcf.VCFConstants;
 import htsjdk.variant.vcf.VCFUtils;
@@ -36,15 +37,12 @@ import java.util.EnumMap;
 import java.util.List;
 import java.util.Map;
 import java.util.HashMap;
-import java.util.Objects;
-
-import org.apache.commons.math3.util.CombinatoricsUtils;
 
 public class GenotypeLikelihoods {
     private final static int NUM_LIKELIHOODS_CACHE_N_ALLELES = 5;
     private final static int NUM_LIKELIHOODS_CACHE_PLOIDY = 10;
-    // caches likelihoods up to 5 alleles and up to 10 ploidy
-    private final static int[][] numLikelihoodCache = new int[NUM_LIKELIHOODS_CACHE_N_ALLELES][NUM_LIKELIHOODS_CACHE_PLOIDY];
+    // caches likelihoods
+    private final static GenotypeNumLikelihoodsCache numLikelihoodCache = new GenotypeNumLikelihoodsCache();
 
     public final static int MAX_PL = Integer.MAX_VALUE;
 
@@ -57,13 +55,13 @@ public class GenotypeLikelihoods {
     private String likelihoodsAsString_PLs = null;
 
     /**
-     * initialize num likelihoods cache
+     * initialize num likelihoods cache to up to 5 alleles and 10 ploidy
      */
     static {
         // must be done before PLIndexToAlleleIndex
         for ( int numAlleles = 1; numAlleles < NUM_LIKELIHOODS_CACHE_N_ALLELES; numAlleles++ ) {
             for ( int ploidy = 1; ploidy < NUM_LIKELIHOODS_CACHE_PLOIDY; ploidy++ ) {
-                numLikelihoodCache[numAlleles][ploidy] = calcNumLikelihoods(numAlleles, ploidy);
+                numLikelihoodCache.put(numAlleles,ploidy, calcNumLikelihoods(numAlleles, ploidy));
             }
         }
     }
@@ -447,7 +445,7 @@ public class GenotypeLikelihoods {
     private static final int calcNumLikelihoods(final int numAlleles, final int ploidy) {
         //Note: Casting to int instead instead of returning long because values above Integer.MAX_VALUE would not be valid array indices,
         // and would cause other problems if a PL array needed to be that size
-        return (int)(CombinatoricsUtils.binomialCoefficient((numAlleles + ploidy - 1),ploidy));
+        return (int)(BinomialCoefficientUtil.binomialCoefficient((numAlleles + ploidy - 1),ploidy));
     }
 
     /**
@@ -475,13 +473,15 @@ public class GenotypeLikelihoods {
      *   @return    Number of likelihood elements we need to hold.
      */
     public static int numLikelihoods(final int numAlleles, final int ploidy) {
-        if ( numAlleles < NUM_LIKELIHOODS_CACHE_N_ALLELES
-                && ploidy < NUM_LIKELIHOODS_CACHE_PLOIDY )
-            return numLikelihoodCache[numAlleles][ploidy];
-        else {
-            // have to calculate on the fly
-            return calcNumLikelihoods(numAlleles, ploidy);
+        //Check cache
+        Integer numLikelihoods = numLikelihoodCache.get(numAlleles, ploidy);
+        if(numLikelihoods != null){
+            return numLikelihoods;
         }
+        // have to calculate on the fly
+        numLikelihoods = calcNumLikelihoods(numAlleles, ploidy);
+        numLikelihoodCache.put(numAlleles, ploidy, numLikelihoods);
+        return numLikelihoods;
     }
 
     // As per the VCF spec: "the ordering of genotypes for the likelihoods is given by: F(j/k) = (k*(k+1)/2)+j.
@@ -604,4 +604,5 @@ public class GenotypeLikelihoods {
         indexes[2] = calculatePLindex(allele2Index, allele2Index);
         return indexes;
     }
+
 }
--- a/src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
+++ b/src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
@@ -38,6 +38,8 @@ import java.util.Map;
 import java.util.HashMap;
 import java.util.Objects;
 
+import org.apache.commons.math3.util.CombinatoricsUtils;
+
 public class GenotypeLikelihoods {
     private final static int NUM_LIKELIHOODS_CACHE_N_ALLELES = 5;
     private final static int NUM_LIKELIHOODS_CACHE_PLOIDY = 10;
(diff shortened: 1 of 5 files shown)


lbergelson:
@KevinCLydon @droazen This adds a new dependency which isn't ideal to avoid implementing  1 function which already exists in our codebase.   The implementation in the apache library is also recursive so it's not clear to me that that is going to be a performance improvement.  

Without profiling showing this is a meaningful improvement I don't think we should make this change.  If we believe the apache implementation is better than ours maybe we can adopt their function without including the entirety of commons-math3:3.3.

src/test/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
```diff
@@ -128,11 +128,16 @@ public void testGetAsMap(){
...
         for (int nAlleles=2; nAlleles<=5; nAlleles++)
-            // simplest case: diploid
+        {
             Assert.assertEquals(GenotypeLikelihoods.numLikelihoods(nAlleles, 2), nAlleles*(nAlleles+1)/2);
+        }
 
         // some special cases: ploidy = 20, #alleles = 4
         Assert.assertEquals(GenotypeLikelihoods.numLikelihoods(4, 20), 1771);
+        // ploidy = 10, alleles = 5
+        Assert.assertEquals(GenotypeLikelihoods.numLikelihoods(5, 10), 1001);
```
lbergelson: 👍 to more tests

droazen:
@lbergelson If you examine the Apache implementation, you'll find that it makes *many* fewer recursive calls. The recursion in our implementation happens within a loop, and can produce a crazy number of recursive calls with realistic values for the number of alleles and ploidy. 

I think we should certainly do some benchmarking to illustrate the problems with the current implementation, and to test whether the Apache implementation is actually faster. 

I know that HTSJDK is opposed in general to new dependencies, but if we want to maintain a high level of code quality we really shouldn't be implementing our own subpar versions of standard mathematical functions, when high-quality and well-tested implementations exist in standard libraries, and these libraries themselves are reasonably lightweight.

lbergelson:
@droazen It does look more efficient than our implementation. I would still rather import that 1 function than add all of commons-math3.  All of htsjdk and it's dependencies right now are 4.6 mb, with math 3 it's 6.5, so it's a large proportional increase in stuff.   If there was a whole bunch of stuff we wanted from it then I would say import the whole thing, but we just want 1 small function, why not just take that 1?  It's not like we don't have apache licensed stuff in htsjdk already....

Commons math is essentially [abandonware](https://www.orekit.org/mailing-list-archives/orekit-users/msg00031.html), so it's not like it's getting regular updates anyway.

vruano:
In gatk we have such calculation effectively cached in a big bidimensional matrix although in ```GenotypeLikelihoodsCalculator(s)``` but is a bit obfuscated because that/those classes do more things apart from that. 

If this is something that is done repetively I would say that caching is the better alternative in terms of performance. I don't think in practice would use that much memory and time to construct. No apache-common dependency required.

vruano:
Perhaps you would like to porr that GATK class/classes in a "decaffeinated" form into htsjdk. I created it in the first place in order to accommodate for non diploid large allele-number calculations so a deficiency in htsjdk back in the day and perhaps now as well.

vruano:
I think it could benefit of some refactoring ... but part of the design idea that you get a calculator for a specific ploidy and number of alleles and then you ask that calculator several things, for example the number of genotypes is that often is the case that ofthen you are working with a specific combination of ploidy/allele-number and you are better of querying this info to a object that is efficient responding to queries related to that combination as supposed to some static object that must handle any combo.

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -55,13 +55,13 @@
     private String likelihoodsAsString_PLs = null;
 
     /**
-     * initialize num likelihoods cache
+     * initialize num likelihoods cache to up to 5 alleles and 10 ploidy
      */
     static {
```
lbergelson: This warm up should be moved into the constructor of the cache instead of a static block now that it has it's own place to live.

src/main/java/htsjdk/variant/variantcontext/GenotypeNumLikelihoodsCache.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.HashMap;
+
+/**
+ * Hybrid dynamic cache for genotype likelihood counts
+ */
+
+public class GenotypeNumLikelihoodsCache {
+
+    private final static int N_ALLELES = 5;
```
lbergelson: These should be specified in 1 place, either in GenotypeLikelihoodUtils or here but not both.

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -443,16 +443,9 @@ private static void calculatePLIndexToAlleleIndices(final int altAlleles, final
...
-            return numAlleles;
-        else {
-            int acc =0;
-            for (int k=0; k <= ploidy; k++ )
-                acc += calcNumLikelihoods(numAlleles - 1, ploidy - k);
-            return acc;
-        }
+        //Note: Casting to int instead instead of returning long because values above Integer.MAX_VALUE would not be valid array indices,
+        // and would cause other problems if a PL array needed to be that size
+        return (int)(BinomialCoefficientUtil.binomialCoefficient((numAlleles + ploidy - 1),ploidy));
```
lbergelson: I would probably cast this with Math.toIntExact() so we know if we ever overflow.

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -55,13 +55,13 @@
     private String likelihoodsAsString_PLs = null;
 
     /**
-     * initialize num likelihoods cache
+     * initialize num likelihoods cache to up to 5 alleles and 10 ploidy
```
lbergelson: you can reference the constants in the javadoc so it doesn't accidentally get out of date.

src/main/java/htsjdk/variant/variantcontext/GenotypeNumLikelihoodsCache.java
```diff
@@ -0,0 +1,43 @@
...
+    private final static int PLOIDY = 10;
+
+    private final int[][] staticCache = new int[5][10];
+    private final HashMap<Integer, HashMap<Integer, Integer>> dynamicCache = new HashMap<>();
```
lbergelson: I don't know if we want to change this cache to do the calculations itself if missing.  I.e. instead of get/put, we could just have get which would handle the caching internally and always return an answer.  it might make things a bit cleaner.

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -471,25 +464,24 @@ else if (ploidy == 1)
...
-     *   S(1,P) = 1 (only way to have 1 integer add to P is with that integer P itself).
+     * For non-cached values, the result is calculated via a call to calcNumLikelihoods,
+     * which uses the Apache Commons CombinatoricsUtils class
+     * using the formula (numAlleles + ploidy - 1) choose ploidy
      *
      *   @param  numAlleles      Number of alleles (including ref)
      *   @param  ploidy          Ploidy, or number of chromosomes in set
      *   @return    Number of likelihood elements we need to hold.
      */
     public static int numLikelihoods(final int numAlleles, final int ploidy) {
```
lbergelson: now that the cache is dynamic this needs to be synchronized or we risk weird intermediate states.

src/main/java/htsjdk/variant/variantcontext/GenotypeNumLikelihoodsCache.java
lbergelson: It would probably be faster to use a compound key and a single layer of HashMap, but it would be more complicated if we wanted to avoid creating a new key object on every lookup.

src/main/java/htsjdk/variant/variantcontext/GenotypeNumLikelihoodsCache.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public void put(final int numAlleles, final int ploidy, final int numLikelihoods) {
+        if(numAlleles < N_ALLELES && ploidy < PLOIDY) {
+            staticCache[numAlleles][ploidy] = numLikelihoods;
+        }
+        else{
+            final HashMap<Integer, Integer> cache = dynamicCache.containsKey(ploidy) ? dynamicCache.get(ploidy) : new HashMap<>();
```
lbergelson: This is kind of nitpicky, but since we're saying this is important to performance ( which hopefully it's much less important now that we have a faster implementation), but we might want to avoid redundant hash look ups by using get and saving / checking the result rather than the nicer contains followed by a get.  I might skip that though since I don't think it will be a bottleneck at all...

src/test/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
```diff
@@ -125,16 +125,36 @@ public void testGetAsMap(){
...
+        // ploidy = 16, alleles = 10
+        Assert.assertEquals(GenotypeLikelihoods.numLikelihoods(10, 16), 2042975);
+    }
+
+    @Test
+    public void testNumLikelihoodsCache() {
+        GenotypeNumLikelihoodsCache cache = new GenotypeNumLikelihoodsCache();
+
+        cache.put(2, 5, 10);
+        cache.put(10, 16, 2042975);
```
lbergelson: since there's a bit of slightly tricky multilayered map access stuff going on, could we add an additional test case where we put a value that shares the same first level.  i.e. add `(2, 10, whateverthevalueis)` . In general we tend to use dataproviders for cases like this where we have a bunch of similar tests with different values, but this is fine if you don't want to change it over.

src/test/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
```diff
@@ -125,16 +125,36 @@ public void testGetAsMap(){
...
+
+        Assert.assertEquals(cache.get(2, 5), new Integer(10));
```
lbergelson: not sure why we need to manually box this.

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -36,15 +37,12 @@
 import java.util.List;
 import java.util.Map;
 import java.util.HashMap;
-import java.util.Objects;
-
-import org.apache.commons.math3.util.CombinatoricsUtils;
 
 public class GenotypeLikelihoods {
     private final static int NUM_LIKELIHOODS_CACHE_N_ALLELES = 5;
```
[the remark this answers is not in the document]
KevinCLydon: Makes sense.  I'll move them to the cache class

src/main/java/htsjdk/variant/variantcontext/GenotypeNumLikelihoodsCache.java
```diff
@@ -0,0 +1,43 @@
+package htsjdk.variant.variantcontext;
+
+import java.util.HashMap;
+
+/**
+ * Hybrid dynamic cache for genotype likelihood counts
+ */
+
+public class GenotypeNumLikelihoodsCache {
```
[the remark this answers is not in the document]
KevinCLydon: Good point.  I'll switch it to default

src/main/java/htsjdk/variant/variantcontext/GenotypeNumLikelihoodsCache.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private final static int N_ALLELES = 5;
+    private final static int PLOIDY = 10;
+
+    private final int[][] staticCache = new int[5][10];
```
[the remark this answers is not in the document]
KevinCLydon: Good catch.  Weird oversight on my part

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
KevinCLydon: Yeah, that's definitely more organized.  I'll move it

src/main/java/htsjdk/variant/variantcontext/GenotypeNumLikelihoodsCache.java
KevinCLydon: I think I'll move them out of GenotypeLikelihoods so they're just here

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
KevinCLydon: Oh I hadn't thought of that.  Thanks, I'll make that change

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
KevinCLydon: Oh true. I'll move the cache initialization to the cache class constructor and reference them there

src/main/java/htsjdk/variant/variantcontext/GenotypeNumLikelihoodsCache.java
KevinCLydon: I like that idea.  I'll make that change

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
KevinCLydon: Good catch; I'll add that

src/main/java/htsjdk/variant/variantcontext/GenotypeNumLikelihoodsCache.java
KevinCLydon: I'm on board with using a compound key

src/main/java/htsjdk/variant/variantcontext/GenotypeNumLikelihoodsCache.java
KevinCLydon: This is an easy enough change.  Although switching to a compound will probably remove this

src/test/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
KevinCLydon: I'll switch to a data provider.  And I think switching to the single-layered map will remove the potential of issues that could result from the multilayered one.  I'll probably make a few changes to these tests to account for the other changes

src/test/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
KevinCLydon: We shouldn't need to.  I'm not actually sure why I added that

src/main/java/htsjdk/variant/variantcontext/GenotypeNumLikelihoodsCache.java
```diff
@@ -6,38 +6,90 @@
...
         }
     }
 
-    public boolean containsKey(final int numAlleles, final int ploidy) {
-        return (numAlleles < N_ALLELES && ploidy < PLOIDY && staticCache[numAlleles][ploidy] > 0)
-                || (dynamicCache.containsKey(ploidy) && dynamicCache.get(ploidy).containsKey(numAlleles));
+    /**
+     * Key for the hash map, made up of numAlleles and ploidy
+     */
+    private class CacheKey{
```
lbergelson: this should be static

src/main/java/htsjdk/variant/variantcontext/GenotypeNumLikelihoodsCache.java
```diff
@@ -6,38 +6,90 @@
...
+    /**
+     * Returns the number of likelihoods for the specified allele count and ploidy
+     * Values not present yet in the cache will be calculated and cached when get is called
+     * @param numAlleles
+     * @param ploidy
+     * @return number of likelihoods
+     */
+    int get(final int numAlleles, final int ploidy) {
+        if(numAlleles <= 0 || ploidy <= 0){
+            throw new IllegalArgumentException("numAlleles and ploidy must both exceed 0");
```
lbergelson: include their actual values here in case anyone actually hits this error

src/main/java/htsjdk/variant/variantcontext/GenotypeNumLikelihoodsCache.java
```diff
@@ -6,38 +6,90 @@
...
+    }
+
+    GenotypeNumLikelihoodsCache(int numAlleles, int ploidy){
+        staticCache = new int[numAlleles][ploidy];
+
+        fillCache();
+    }
+
+    private void fillCache(){
+        for ( int numAlleles = 1; numAlleles < staticCache.length; numAlleles++ ) {
```
lbergelson: as discussed, shift by 1

PR #1448, opened 2020-01-01
Some improvements on CigarOperator enum.

vruano:
### Description

* Clean up a bit the code.
   - Removed per value switches.
   - String, character initialization.
* Optimize some operations
   - Fast look-up of values from primitive codes.
   - Static to member methods to generate primitive codes.
* Deprecated methods and update their uses to new ones.
* Clear most of the warnings on the main and test code.

- [ ] Check your code style.

### DIFF
--- a/src/main/java/htsjdk/samtools/BinaryCigarCodec.java
+++ b/src/main/java/htsjdk/samtools/BinaryCigarCodec.java
@@ -46,7 +46,7 @@ class BinaryCigarCodec {
         int binaryCigarLength = 0;
         for (int i = 0; i < cigar.numCigarElements(); ++i) {
             final CigarElement cigarElement = cigar.getCigarElement(i);
-            final int op = cigarElement.getOperator().bamCode();
+            final int op = cigarElement.getOperator().getBamEncoding();
             binaryCigar[binaryCigarLength++] = cigarElement.getLength() << 4 | op;
         }
         return binaryCigar;
@@ -84,6 +84,6 @@ class BinaryCigarCodec {
     private static CigarElement binaryCigarToCigarElement(final int cigarette) {
         final int binaryOp = cigarette & 0xf;
         final int length = cigarette >>> 4;
-        return new CigarElement(length, CigarOperator.fromBamCode(binaryOp));
+        return new CigarElement(length, CigarOperator.fromBamEncoding(binaryOp));
     }
 }
--- a/src/main/java/htsjdk/samtools/BinaryCigarCodec.java
+++ b/src/main/java/htsjdk/samtools/BinaryCigarCodec.java
@@ -46,7 +46,7 @@ class BinaryCigarCodec {
         int binaryCigarLength = 0;
         for (int i = 0; i < cigar.numCigarElements(); ++i) {
             final CigarElement cigarElement = cigar.getCigarElement(i);
-            final int op = CigarOperator.enumToBinary(cigarElement.getOperator());
+            final int op = cigarElement.getOperator().bamCode();
             binaryCigar[binaryCigarLength++] = cigarElement.getLength() << 4 | op;
         }
         return binaryCigar;
@@ -84,6 +84,6 @@ class BinaryCigarCodec {
     private static CigarElement binaryCigarToCigarElement(final int cigarette) {
         final int binaryOp = cigarette & 0xf;
         final int length = cigarette >>> 4;
-        return new CigarElement(length, CigarOperator.binaryToEnum(binaryOp));
+        return new CigarElement(length, CigarOperator.fromBamCode(binaryOp));
     }
 }
--- a/src/main/java/htsjdk/samtools/CigarOperator.java
+++ b/src/main/java/htsjdk/samtools/CigarOperator.java
@@ -23,52 +23,59 @@
  */
 package htsjdk.samtools;
 
+/* Disjoint categorization of cigar-operators in coarser classes. */
+enum CigarOperatorClass {
+    ALIGNMENT, CLIPPING, INDEL, SKIP, PADDING
+}
+
 /**
  * The operators that can appear in a cigar string, and information about their disk representations.
  */
 public enum CigarOperator {
+
     /** Match or mismatch */
-    M(true, true),
+    M(true, true, "M", 0, CigarOperatorClass.ALIGNMENT),
     /** Insertion vs. the reference. */
-    I(true, false),
+    I(true, false, "I", 1, CigarOperatorClass.INDEL),
     /** Deletion vs. the reference. */
-    D(false, true),
+    D(false, true, "D", 2, CigarOperatorClass.INDEL),
     /** Skipped region from the reference. */
-    N(false, true),
+    N(false, true, "N", 3, CigarOperatorClass.SKIP),
     /** Soft clip. */
-    S(true, false),
+    S(true, false, "S", 4, CigarOperatorClass.CLIPPING),
     /** Hard clip. */
-    H(false, false),
+    H(false, false, "H", 5, CigarOperatorClass.CLIPPING),
     /** Padding. */
-    P(false, false),
+    P(false, false, "P", 6, CigarOperatorClass.PADDING),
     /** Matches the reference. */
-    EQ(true, true,  "="),
+    EQ(true, true,  "=", 7, CigarOperatorClass.ALIGNMENT),
     /** Mismatches the reference. */
-    X(true, true)
+    X(true, true, "X", 8, CigarOperatorClass.ALIGNMENT)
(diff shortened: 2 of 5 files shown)


src/main/java/htsjdk/samtools/CigarOperator.java
```diff
@@ -89,114 +117,116 @@
...
-            case P:
-                return OP_P;
-            case EQ:
-                return OP_EQ;
-            case X:
-                return OP_X;
-            default:
-                throw new IllegalArgumentException("Unrecognized CigarOperator: " + e);
-        }
+        return e.ordinal();
```
yfarjoun: I don't think that it is safe to use ordinal() for encoding and decoding. as it means that two implicit decision (the order of the enum definition, and the way java converts an enum class into ordinals) affects how the cigar is encoded and decoded. I think that it is much preferable to have it written explicitly. 

@tfenne @lbergelson opinions?

vruano: I added unit test to check that the transcription into bam-codes is correct. So a change in order of the enum constant would result in a test failure. 
However I'm happy to change it to explicitly indicate the code as a additional constructor argument.

lbergelson: I think it's unlikely that this will be a problem in practice, but typically it's considered bad practice to rely on the ordinal value in java because it's brittle if you change the ordering or add new enum values.  I would say keep it as a constant in the constructor.

src/main/java/htsjdk/samtools/CigarOperator.java
```diff
@@ -72,12 +94,18 @@
...
+    }
+
     /** Default constructor. */
-    CigarOperator(boolean consumesReadBases, boolean consumesReferenceBases, char character) {
+    CigarOperator(boolean consumesReadBases, boolean consumesReferenceBases, final String string) {
         this.consumesReadBases = consumesReadBases;
         this.consumesReferenceBases = consumesReferenceBases;
-        this.character = (byte) character;
-        this.string = new String(new char[] {character}).intern();
+        this.string = string == null ? name() : string; // since we will provide a constant it is internalized already.
```
lbergelson: This is clever but I'd rather just have it be specified in the constructor than have logic for determining it from the enum value name.

src/main/java/htsjdk/samtools/BinaryCigarCodec.java
```diff
@@ -46,7 +46,7 @@
         int binaryCigarLength = 0;
         for (int i = 0; i < cigar.numCigarElements(); ++i) {
             final CigarElement cigarElement = cigar.getCigarElement(i);
-            final int op = CigarOperator.enumToBinary(cigarElement.getOperator());
+            final int op = cigarElement.getOperator().bamCode();
```
lbergelson: Maybe `bamCode` would be clearer as `getBamEncoding`

and `fromBamCode` -> `fromBamEncoding` .

src/main/java/htsjdk/samtools/CigarOperator.java
```diff
@@ -89,114 +117,116 @@
...
     /** Returns true if the operator is a Insertion or Deletion operator */
     public boolean isIndel() {
-        return this == I || this == D;
+        return (INDEL_MASK & mask) != 0;
     }
 
     /** Returns true if the operator is a Skipped Region Insertion or Deletion operator */
     public boolean isIndelOrSkippedRegion() {
-        return this == N || isIndel();
+        return (mask & INDEL_OR_SKIP_MASK) != 0;
```
lbergelson: Do you ever see these show up in a profiler?  This looks like it should be a bit faster if the compiler can't write clever if statements.  I'm curious if this is a real performance problem though.

src/main/java/htsjdk/samtools/BinaryCigarCodec.java
lbergelson: Yes!  I've looked at this and been annoyed by the fact that this is computed rather than hardcoded before.

src/main/java/htsjdk/samtools/CigarOperator.java
```diff
@@ -28,40 +28,62 @@
...
+    N(false, true),
     /** Soft clip. */
-    S(true, false,  'S'),
+    S(true, false),
     /** Hard clip. */
-    H(false, false, 'H'),
+    H(false, false),
     /** Padding. */
-    P(false, false, 'P'),
+    P(false, false),
```
lbergelson: I would just have 1 constructor that takes the  character and the binary representation as parameters.  It's a bit redundant but it seems clearer to me.

src/main/java/htsjdk/samtools/CigarOperator.java
```diff
@@ -89,114 +117,116 @@
...
+     * <p>
+     *     It is guaranteed that {@code CigarOperator.fromBamCode(op.bamCode()) == op}.
+     * </p>
+     *
+     * @param bamCode the query BAM code.
+     * @return never {@code null}.
+     * @throws IllegalArgumentException if the input BAM code is not valid.
+     * @see #bamCode
+     */
+    public static CigarOperator fromBamCode(final int bamCode) {
```
lbergelson: Is this actually faster than the switch statement?  It's slightly less verbose but it seems likely that it will be implemented pretty similarly as a lookup in a small precomputed table.

lbergelson (changes requested):
@vruano The fact that you had to compute the binary value has annoyed me whenever I looked at this so definitely a good change to hardcode code that.  

 I'm not totally sold on the other optimizations but I can go either way.  It's not clear to me that the lookup arrays are actually faster than a switch statement, it seems likely that they are compiled to something very similar either way, although the java compiler has surprised me with poor optimization before.  It seems like we're trading a stupid but extremely simple method for a more complicated and *maybe* more performant one.  

Deprecating the old method names seems reasonable, this seems more standard.  

I suggested some slight changes to method names.

I agree with yossi that we shouldn't be using the ordinal value even though for all practical purposes it's unlikely to be a problem.

src/main/java/htsjdk/samtools/CigarOperator.java
vruano: If the code that use these enum it itself efficiently written, this probably won't show in a profiler. This mask based solution is an the Inlined version of a EnumSet... . 

I was curious myself so I did some bench marking. if you allow JIT to do its magic the mask and using an EnumSet is about the same cost say (1000 units) whereas compare with three constants (== || == || ==) was around 5000 2 constants 4500 (== || ==) and one only 2000.  
If you simply add a extra boolean field to identify members of a group then it is ~750. The same if your set happens to be begining or end of the enum const list so that you can use this.ordinal() <= X.ordinal() then it was around 750. It is easy to see how the compiler may transform the latter into the former and so the similar performance. 

Now disabling JIT (java -Djava.compiler=NONE)  all the solutions were the same cost except from using the EnumSet that was 4 or 5 times slower... so it seems that another factor was dominating for the faster alternatives (perhaps memory access). 

Anyway the bottom line is that using additional boolean fields is probably the faster way closely followed by using a mask and or EnumSet (but only if we can assume JIT is active).

Which one do you prefer?

lbergelson: Interesting.  Let's just embed a bunch of constants then, that seems fast and readable.  We can set them in the constructor using the clear logic that exists now.

vruano: Not sure what you mean... do you mean create additional ```final boolean``` fields (one per set) caching whether the value belongs to a particular set?

src/main/java/htsjdk/samtools/CigarOperator.java
vruano: will leave it as it is.

vruano:
I have addressed some the comments and thought of another way to address broader operator categories meaning alignment, indel, clips, etc... It turns out that Alignment, Indel, Clipping, Padding and Skip (N) are disjointed and exhaustive categorization of the operators.

I added a non-public enum for those categories and the boolean final fields for each status accessor (e.g. isAlignment, isIndel... etc) is initialized using a single param in the constructor that indicates the class/category for that operator. 

What do you think? @lbergelson @yfarjoun ...

You know what they say ... if you feel you need to ask to do somethings is because you know you shouldn't do it :P

src/main/java/htsjdk/samtools/CigarOperator.java
vruano: In the new commit I done something different to make the constructor expression a bit more readable... is difficult to get what is going on with so many ""anonymous"" additional boolean arguments

PR #1449, opened 2020-01-07
Fix IndexFactory on non-default file system.

cmnbroad:
Fixes the issue seen [here](https://github.com/broadinstitute/gatk/issues/6348), along with a new test that fails without the fix. The actual fix is a one line change, the rest of the changes are to fix error messages, and there is a second commit with some other redundant code removed.

### DIFF
--- a/src/main/java/htsjdk/tribble/index/IndexFactory.java
+++ b/src/main/java/htsjdk/tribble/index/IndexFactory.java
@@ -520,7 +520,7 @@ public class IndexFactory {
                     msg += " Saw feature " + featToString(visitedChromos.get(curChr));
                     msg += " followed later by " + featToString(lastFeature);
                     msg += " and then " + featToString(currentFeature);
-                    throw new TribbleException.MalformedFeatureFile(msg, inputPath.toAbsolutePath().toString());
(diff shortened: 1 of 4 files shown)


src/main/java/htsjdk/tribble/index/IndexFactory.java
```diff
@@ -625,17 +625,17 @@ private static BlockCompressedInputStream initIndexableBlockCompressedStream(fin
             try {
                 if (!IOUtil.isBlockCompressed(inputPath, true)) {
                     throw new TribbleException.MalformedFeatureFile("Input file is not in valid block compressed format.",
-                            inputPath.toAbsolutePath().toString());
+                            inputPath.toString());
                 }
 
                 final ISeekableStreamFactory ssf = SeekableStreamFactory.getInstance();
-                final SeekableStream seekableStream = ssf.getStreamFor(inputPath.toAbsolutePath().toString());
+                final SeekableStream seekableStream = ssf.getStreamFor(inputPath.toUri().toString());
```
cmnbroad: This is the actual fix.

cmnbroad:
Thanks for the quick review @lbergelson.

PR #1454, opened 2020-01-25
Extract Allele into a interface with only one implementation as is currently the case.

vruano:
This is the first of a few commits/PR to refactor Allele as intended in the "Large" stale PR https://github.com/samtools/htsjdk/pull/1370

This first commit extracts an interface from Allele that becomes the new Allele common . class ejem interface, and the sole implementation is ByteArrayAllele that as the name states, uses a byte[] array to encode the allele bases, symbolic ID etc like the original Allele class. This change was suggested by @droazen.

ByteArrayAllele is not mean to be use publicly (package protection level). To instantiate it you should use the original Allele static create methods.

It also does some deprecation marking of unused or smelly methods like the "wouldBe*" and acceptableBases families.

@lbergelson Please review.

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/ByteArrayAllele.java
+++ b/src/main/java/htsjdk/variant/variantcontext/ByteArrayAllele.java
@@ -1,0 +1,305 @@
+/*
+* Copyright (c) 2012 The Broad Institute
+* 
+* Permission is hereby granted, free of charge, to any person
+* obtaining a copy of this software and associated documentation
+* files (the "Software"), to deal in the Software without
+* restriction, including without limitation the rights to use,
+* copy, modify, merge, publish, distribute, sublicense, and/or sell
+* copies of the Software, and to permit persons to whom the
+* Software is furnished to do so, subject to the following
+* conditions:
+* 
+* The above copyright notice and this permission notice shall be
+* included in all copies or substantial portions of the Software.
+* 
+* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+* EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
+* OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+* NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+* HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+* WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+* FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR
+* THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
+package htsjdk.variant.variantcontext;
+
+import htsjdk.samtools.util.StringUtil;
+
+import java.util.Arrays;
+
+class ByteArrayAllele implements Allele {
+
+    private static long serialVersionUID = 1L;
+
+    private static final byte[] EMPTY_ALLELE_BASES = new byte[0];
+
+    private boolean isRef = false;
+    private boolean isNoCall = false;
+    private boolean isSymbolic = false;
+
+    private byte[] bases = null;
+
+    // no public way to create an allele
+    protected ByteArrayAllele(final byte[] bases, final boolean isRef) {
+        // null alleles are no longer allowed
+        if ( Allele.wouldBeNullAllele(bases) ) {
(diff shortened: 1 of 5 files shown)


src/main/java/htsjdk/variant/variantcontext/ByteArrayAllele.java
```diff
@@ -0,0 +1,305 @@
...
+    public byte[] getBases() { return isSymbolic ? EMPTY_ALLELE_BASES : bases; }
+
+    /**
+     * Return the DNA bases segregating in this allele in String format.
+     * This is useful, because toString() adds a '*' to reference alleles and getBases() returns garbage when you call toString() on it.
+     *
+     * @return the segregating bases
+     */
+    @Override
+    public String getBaseString() { return isNoCall() ? NO_CALL_STRING : new String(getBases()); }
```
lbergelson: This should really specify a charset.

src/main/java/htsjdk/variant/variantcontext/ByteArrayAllele.java
```diff
@@ -0,0 +1,305 @@
...
+        else if (other instanceof ByteArrayAllele) {
+            final ByteArrayAllele baOther = (ByteArrayAllele) other;
+            return compareBases(baOther.bases);
+        } else {
+            return compareBases(other.getBases());
+        }
+    }
+
+    private int compareBases(final byte[] otherBases) {
+        final int stop = otherBases.length < bases.length ? otherBases.length : bases.length;
```
lbergelson: Should this have a fast return if bases = otherBases?

lbergelson: can't this stop calculation be replaced with Math.min?

src/main/java/htsjdk/variant/variantcontext/ByteArrayAllele.java
```diff
@@ -0,0 +1,305 @@
...
+    public byte[] getDisplayBases() { return bases; }
+
+    /**
+     * @param other  the other allele
+     *
+     * @return true if these alleles are equal
+     */
+    @Override
+    public boolean equals(Object other) {
+        return ( ! (other instanceof ByteArrayAllele) ? false : equals((ByteArrayAllele)other, false) );
```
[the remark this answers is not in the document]
vruano: good point.

src/main/java/htsjdk/variant/variantcontext/ByteArrayAllele.java
vruano: Is it really necessary considering that we should not encounter non-ascii characters (bytes)... Honestly I've never dealled with mixtures of charsets so perhaps you can mention an scenario in that it would be useful to use something different to the system default ... for example... if someone runs this in Japan, will he get random kanji?

src/main/java/htsjdk/variant/variantcontext/ByteArrayAllele.java
vruano: yeah... I kinda imagin that it has an added cost but perhaps it is inlined.

src/main/java/htsjdk/variant/variantcontext/ByteArrayAllele.java
lbergelson: It's probably not not strictly necessary because it's unlikely that anyone will use non-ascii bases and almost every system default will keep ascii in the same position I hope.  I think it's unlikely we'd run into any problems here, we haven't in years, but there's always someone with a really crazy setting on their computer somewhere.    It's fine if you don't want to change it here.

src/main/java/htsjdk/variant/variantcontext/ByteArrayAllele.java
lbergelson: I would hope it gets inline and has no added cost, but you can never be sure... Math.min is annotated as a hotspot intrinsic candidate which makes me suspect it will probably be very fast.  I think in any case it's clearer though.

mwalker174:
@vruano Were you planning to follow through on this? As the first step towards integrating symbolic alleles, it is very much needed for gatk-sv development.

vruano:
@lbergelson addressed comments and it passes tests.

lbergelson:
@vruano There's something about this branch that is making the diff show up with lots changes that are already in master.  Have you rebased it recently?

vruano:
Seems that the last rebase has fixed the issue with the additional master commit showing up. I never did anything different to the usual when rebasing so is kind of mystery to why it happened .... but now is gone :D

vruano:
@lbergelson seems that the many-commit issues has dissipated by simply rebasing. Can you give it the ok and merge?

lbergelson (approved):
Looks good.  Thank you @vruano

PR #1458, opened 2020-02-19
interval_list with gz compression

lindenb:
### Description

This PR allows a, '.interval_list' to be **gzipped**. It's now identified by SAMSequenceDictionaryExtractor. 

I also added the new extension in ` public boolean canDecode(final String s) ` . I don't know tribble enough: is it enough to be identified by gatk/picard ?

- [x] Check your code style.


src/main/java/htsjdk/tribble/IntervalList/IntervalListCodec.java
```diff
@@ -158,7 +159,8 @@ public void close() { }
     }
 
     @Override
-    public boolean canDecode(String s) {
-        return s.endsWith(".interval_list");
+    public boolean canDecode(final String s) {
+        return s.endsWith(FileExtensions.INTERVAL_LIST) ||
```
lbergelson: This should in theory enable this is in picard and gatk.  I think it will work without issue in picard, but I think GATK has old outdated handling of interval_list's and it might not get picked up correctly.  We can check though.

PR #1463, opened 2020-03-14
Better tribble error message

### LINKED ISSUE #1462: Tribble parsing error message
Parsing errors in Tribble result in the following error message

```
Error parsing line at byte position: LineIteratorImpl(SynchronousLineReader), for input source: /Users/jrobinso/Downloads/ARHI0A_peaks_std_500.bed
```

This is not helpful, and even if the "byte position" were displayed it would be only slightly more helpful.  I think either the line number, or the text of the line itself, should be displayed.   

This is related but more narrow than https://github.com/samtools/htsjdk/issues/709

yfarjoun:
### Description
fixes #1462


src/main/java/htsjdk/tribble/TribbleIndexedFeatureReader.java
```diff
@@ -383,7 +385,14 @@ private void readNextRecord() throws IOException {
...
                     throw e;
                 } catch (NumberFormatException e) {
-                    final String error = "Error parsing line at byte position: " + source;
+
+                    final String error;
+                    if (previousRecord == null) {
+                        error = String.format("Error parsing %s at the first record", source);
+                    } else {
+                        error = String.format("Error parsing %s just after record at: %s:%d-%d",
+                                source.toString(), previousRecord.getContig(), previousRecord.getStart(), previousRecord.getEnd());
```
lbergelson: This seems like an improvement.

src/main/java/htsjdk/tribble/TribbleIndexedFeatureReader.java
```diff
@@ -383,7 +385,14 @@ private void readNextRecord() throws IOException {
                     e.setSource(path);
                     throw e;
                 } catch (NumberFormatException e) {
-                    final String error = "Error parsing line at byte position: " + source;
+
+                    final String error;
+                    if (previousRecord == null) {
+                        error = String.format("Error parsing %s at the first record", source);
+                    } else {
+                        error = String.format("Error parsing %s just after record at: %s:%d-%d",
```
lbergelson: It's not clear to me that source.toString() will be meaningful at all.  I might just drop that.  Path is useful but that should be included already.

PR #1467, opened 2020-03-18
Add another improvement to tribble error code

yfarjoun:
- cleaned up some unneeded files

### Description

- [ ] Check your code style.


src/main/java/htsjdk/tribble/TribbleIndexedFeatureReader.java
```diff
@@ -418,9 +418,9 @@ public WFIterator iterator() {
      * Iterator for a query interval
      */
     class QueryIterator implements CloseableTribbleIterator<T> {
-        private String chrAlias;
-        int start;
-        int end;
+        String chrAlias;
```
lbergelson: we lost a private here for some reason?

src/main/java/htsjdk/tribble/TribbleIndexedFeatureReader.java
```diff
@@ -429,6 +429,8 @@ public WFIterator iterator() {
         public QueryIterator(final String chr, final int start, final int end, final List<Block> blocks) throws IOException {
             this.start = start;
             this.end = end;
+            // for error message
+            chrAlias = chr;
```
lbergelson: this is already set below and should probably be final

src/main/java/htsjdk/tribble/TribbleIndexedFeatureReader.java
```diff
@@ -461,7 +462,7 @@ private void advanceBlock() throws IOException {
             while (blockIterator != null && blockIterator.hasNext()) {
                 final Block block = blockIterator.next();
                 if (block.getSize() > 0) {
-                    final int bufferSize = Math.min(2000000, block.getSize() > 100000000 ? 10000000 : (int) block.getSize());
+                    final int bufferSize = Math.min(2_000_000, block.getSize() > 100_000_000 ? 10_000_000 : (int) block.getSize());
```
lbergelson: Is this a bug here?  I totally assumed those were the same number.

lbergelson: On closer inspection, this whole line seems busted.  
It seems like what they meant to write was `Math.min(2_000_000, Math.min(10_000_000, block.getSize()))`, but then typoed the number of 0's.  But I think as it's currently written (or as I think they intended it, it just reduces to `Math.min(2_000_000, block.getSize())` anyway.  

I'm not sure what the correct behavior is, but this seems wrong in any case.

src/main/java/htsjdk/tribble/TribbleIndexedFeatureReader.java
yfarjoun: I actually need this for the case that an error happens on the first line that is being read....happy to explain more, but this is not a mistake.

src/main/java/htsjdk/tribble/TribbleIndexedFeatureReader.java
yfarjoun: I just added '_' so it (human)readable....if you want to change the numbers, please use a followup PR.

src/main/java/htsjdk/tribble/TribbleIndexedFeatureReader.java
lbergelson: Oh.  I see.  Maybe write that into the comment?

src/main/java/htsjdk/tribble/TribbleIndexedFeatureReader.java
lbergelson: I know, it just surprised me that those numbers weren't the same, which made me look at it.

yfarjoun:
better now @lbergelson

PR #1469, opened 2020-04-02
Add VCFHeader javadoc

bbimber:
I recently got burned (my own fault) using VCFHeader.getSampleNamesInOrder().  In retrospect I should have seen it, but 'InOrder' could reasonably be interpreted to mean 'VCF order' instead of 'sorted'.  This PR adds javadoc to this method to clarify.


lbergelson:
@bbimber Thank you!  We should have done this long ago, I can never remember what order it means without looking at the code.

PR #1473, opened 2020-04-16
new interface VCFReader (+ VCFFIleReader implements VCFReader)

lindenb:
### Description

I created a  new ` interface VCFReader  extends Closeable, Iterable<VariantContext>` and I changed VCFFileReader to `class VCFFileReader extends VCFReader` 

The idea is to  use an interface instead of a File-based class to read+query the VCF files

``` 
public interface VCFReader extends Closeable, Iterable<VariantContext> {
    public VCFHeader getHeader();
    public CloseableIterator<VariantContext> query(final String chrom, final int start, final int end); 
    public default CloseableIterator<VariantContext> query(final Locatable locatable) {...}
    public boolean isQueryable();
    public CloseableIterator<VariantContext> iterator();
}
```

I would like to be able to use some different sources of VCF (one could imagine a **SQL** database, plink ... ) with the same interface. 

My current need is to use a System-call to `bcftools view` to parse the lastest version of the bcf format.

I use `getHeader` instead of `getFileHeader` because it's not always file-based.

I also wondered i I should create a even more generic interface like

``` 
public interface QuerySource<HEADER,ITEM> extends Closeable, Iterable<ITEM> {
    public HEADER getHeader();
    public CloseableIterator<ITEM> query(final String chrom, final int start, final int end); 
    public default CloseableIterator<ITEM> query(final Locatable locatable) {...}
    public boolean isQueryable();
    public CloseableIterator<ITEM> iterator();
}
``` 
... but everything in its time ?

- [x] Check your code style.

### DIFF
--- a/src/main/java/htsjdk/variant/vcf/VCFFileReader.java
+++ b/src/main/java/htsjdk/variant/vcf/VCFFileReader.java
@@ -24,13 +24,11 @@
 
 package htsjdk.variant.vcf;
 
-import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.samtools.util.CloseableIterator;
 import htsjdk.samtools.util.FileExtensions;
 import htsjdk.samtools.util.Interval;
 import htsjdk.samtools.util.IntervalList;
-import htsjdk.samtools.util.Locatable;
 import htsjdk.tribble.AbstractFeatureReader;
 import htsjdk.tribble.FeatureCodec;
 import htsjdk.tribble.FeatureReader;
@@ -38,7 +36,6 @@ import htsjdk.tribble.TribbleException;
 import htsjdk.variant.bcf2.BCF2Codec;
 import htsjdk.variant.variantcontext.VariantContext;
 
-import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
@@ -48,7 +45,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 /**
  * Simplified interface for reading from VCF/BCF files.
  */
-public class VCFFileReader implements Closeable, Iterable<VariantContext> {
+public class VCFFileReader implements VCFReader {
 
     private final FeatureReader<VariantContext> reader;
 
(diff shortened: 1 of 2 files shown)


lbergelson:
We'd like to move to a more generic FeatureReader<Feature,Header> like you're suggesting.  Probably good to save that for a later PR since there are more naming issues and such.

lbergelson:
Wait.. what am I talking about.  We should consider how this relates to FeatureReader already.    This should probably extend FeatureReader<VariantContext> since it's nearly an identical interface.

lindenb:
> Wait.. what am I talking about. We should consider how this relates to FeatureReader already. This should probably extend FeatureReader since it's nearly an identical interface.

I'm not sure I understand your last comment. You meant this new interface "VCFReader should extends FeatureReader" in the current PR ? or in the future ?

lbergelson:
@lindenb Sorry, let's make it extend FeatureReader in this PR.  It seems like a natural thing for it to do and ties it together with the existing interfaces.

lindenb:
> @lindenb Sorry, let's make it extend FeatureReader in this PR.

@lbergelson ok, but it's not clear to me how I should do this. While the signatures of the functions are similar they don't return the same things. For example 

FeatureReader :  ` CloseableTribbleIterator<T> query(Locatable locus)`

VCFFIleReader: `  CloseableIterator<VariantContext> query(final Locatable locatable)  throws IOException ` 

how can I write the interface without breaking the current code ?

or ... the current  ` interface FeatureReader` would extends the interface i described in my first  comment (+ `throws IOException` )

```
public interface QuerySource<HEADER,ITEM> extends Closeable, Iterable<ITEM> {
    public HEADER getHeader();
    public CloseableIterator<ITEM> query(final String chrom, final int start, final int end) throws IOException; 
    public default CloseableIterator<ITEM> query(final Locatable locatable) throws IOException  {...}
    public boolean isQueryable() throws IOException;
    public CloseableIterator<ITEM> iterator() throws IOException;
}
```

and then

```
interface FeatureReader extends QuerySource<Object,T extends Feature> { ... }
```

what do you think ?

lbergelson:
You're right.  It's not as clean as I thought with a quick look.  I'm fine adding what you have here, as long as you'd be open to changing it slightly in the future if we decide to unify things better.

PR #1474, opened 2020-04-23
Add support for SQ-AN in SAMSequenceDictionary #956

yfarjoun:
This is a replacement of PR #956 

see discussion and comments there....

one major difference from that PR: 
- I removed potential problem when providing list of SAMSequenceRecords (didn't make a private copy of list) in commit 30181fb

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
+++ b/src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
@@ -215,7 +215,7 @@ public class SAMSequenceDictionary implements Serializable {
      * <code>MT,chrM</code>
      *
      * NOTE: this method does not add the alias to the alternative sequence name tag (AN) in the SAMSequenceRecord.
-     * If you would like to add it to the AN tag, use {@link #setAlternativeSequenceName(String, String)} instead.
+     * If you would like to add it to the AN tag, use {@link #addAlternativeSequenceName(String, String)} instead.
      *
      * @param originalName
      *            existing contig name
--- a/src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
+++ b/src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
@@ -79,7 +79,7 @@ public class SAMSequenceDictionary implements Serializable {
                 throw new IllegalArgumentException("Cannot add sequence that already exists in SAMSequenceDictionary: " +
                         record.getSequenceName());
             } else {
-                record.getAlternativeSequeneNames().forEach(an -> addSequenceAlias(record.getSequenceName(), an));
+                record.getAlternativeSequenceNames().forEach(an -> addSequenceAlias(record.getSequenceName(), an));
             }
         }
     }
@@ -92,7 +92,7 @@ public class SAMSequenceDictionary implements Serializable {
         sequenceRecord.setSequenceIndex(mSequences.size());
         mSequences.add(sequenceRecord);
         mSequenceMap.put(sequenceRecord.getSequenceName(), sequenceRecord);
-        sequenceRecord.getAlternativeSequeneNames().forEach(an -> addSequenceAlias(sequenceRecord.getSequenceName(), an));
+        sequenceRecord.getAlternativeSequenceNames().forEach(an -> addSequenceAlias(sequenceRecord.getSequenceName(), an));
     }
 
     /**
@@ -197,7 +197,11 @@ public class SAMSequenceDictionary implements Serializable {
         return !thatSequences.hasNext();
     }
 
-    /** returns true if the two dictionaries are the same, aliases are NOT considered */
+    /**
+     * Returns {@code true} if the two dictionaries are the same.
+     *
+     * <p>NOTE: Aliases are NOT considered, but alternative sequence names (AN tag) names ARE.
+     */
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
@@ -214,13 +218,11 @@ public class SAMSequenceDictionary implements Serializable {
      * <code>1,chr1,chr01,01,CM000663,NC_000001.10</code> e.g:
      * <code>MT,chrM</code>
      *
-     * NOTE: this method does not add the alias to the alternative sequence name tag (AN) in the SAMSequenceRecord.
+     * <p>NOTE: this method does not add the alias to the alternative sequence name tag (AN) in the SAMSequenceRecord.
      * If you would like to add it to the AN tag, use {@link #addAlternativeSequenceName(String, String)} instead.
      *
-     * @param originalName
-     *            existing contig name
-     * @param altName
-     *            new contig name
+     * @param originalName  existing contig name
+     * @param altName       new contig name
      * @return the contig associated to the 'originalName/altName'
      */
     public SAMSequenceRecord addSequenceAlias(final String originalName,
@@ -248,13 +250,11 @@ public class SAMSequenceDictionary implements Serializable {
      * to retrieve the contigs (as with {@link #addSequenceAlias(String, String)}.
      *
      * <p>This can be use to provide some alternate names fo a given contig. e.g:
-     * <code>1,chr1,chr01,01,CM000663,NC_000001.10</code> or
+     * <code>1,chr1,chr01,01,CM000663</code> or
      * <code>MT,chrM</code>.
      *
-     * @param originalName
-     *            existing contig name
-     * @param altName
-     *            new contig name
+     * @param originalName  existing contig name
+     * @param altName       new contig name
      * @return the contig associated to the 'originalName/altName', with the AN tag including the altName
      */
     public SAMSequenceRecord addAlternativeSequenceName(final String originalName,
--- a/src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
+++ b/src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
@@ -339,8 +339,8 @@ public class SAMSequenceDictionary implements Serializable {
 
             throw new IllegalArgumentException(String.format("Do not use this function to merge dictionaries with " +
                             "different sequences in them. Sequences must be in the same order as well. Found [%s] and [%s].",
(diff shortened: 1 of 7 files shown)


src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -98,68 +107,185 @@ public SAMSequenceRecord(final String name, final int sequenceLength) {
...
+        mSequenceIndex = value;
+    }
+
+    /**
+     * Returns unmodifiable set with alternative sequence names.
+     */
+    public Set<String> getAlternativeSequenceNames() {
+        final String anTag = getAttribute(ALTERNATIVE_SEQUENCE_NAME_TAG);
+        return (anTag == null) ? Collections.emptySet()
+                : Collections.unmodifiableSet(new HashSet<>(Arrays.asList(anTag.split(ALTERNATIVE_SEQUENCE_NAME_SEPARATOR))));
```
lbergelson, line 172: This shouldn't be coerced through a HashSet, it will shuffle the order.  Use a LinkedHashSet instead.

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -98,68 +107,185 @@ public SAMSequenceRecord(final String name, final int sequenceLength) {
...
+        }
+    }
+
+    private static void validateAltRegExp(final String name) {
+        if (!LEGAL_RNAME_PATTERN.matcher(name).matches()) {
+            throw new IllegalArgumentException(String.format("Invalid alternative sequence name '%s': do not match the pattern %s", name, LEGAL_RNAME_PATTERN));
+        }
+    }
+
+    private void encodeAltSequences(final Collection<String> alternativeSequences) {
```
lbergelson, line 207: I think you could probably save a bunch of code/ potential error paths if the error checking / deduplication is all done here.

Something like:

encoded ==   alternativeSequences.isEmpty() ? null : alternativeSequences.stream()
                .sorted()
                .distinct()
                .peek(SAMSequenceRecord::validateAltRegExp)
                .collect(Collectors.joining(ALTERNATIVE_SEQUENCE_NAME_SEPARATOR));

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -98,68 +107,185 @@ public SAMSequenceRecord(final String name, final int sequenceLength) {
...
+    public void setSequenceIndex(final int value) {
+        mSequenceIndex = value;
+    }
+
+    /**
+     * Returns unmodifiable set with alternative sequence names.
+     */
+    public Set<String> getAlternativeSequenceNames() {
+        final String anTag = getAttribute(ALTERNATIVE_SEQUENCE_NAME_TAG);
+        return (anTag == null) ? Collections.emptySet()
```
lbergelson, line 171: Should we worry about empty string here?

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -306,7 +324,7 @@ public String toString() {
      * @param tagsToMatch list of tags that must be equal if present in both sequence. Must contain MD, and LN
      * @return dictionary consisting of the same sequences as the two inputs with the merged values of tags.
      */
-    static public SAMSequenceDictionary mergeDictionaries(final SAMSequenceDictionary dict1,
+    public static SAMSequenceDictionary mergeDictionaries(final SAMSequenceDictionary dict1,
```
lbergelson, line 327: Hmn, do we have to worry about alias name collisions when merging dicionaries?

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
yfarjoun: empty string shouldn't be valid....

yfarjoun: according to the regexp. I added a test to that effect

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
yfarjoun: yes. :-)

yfarjoun: as written, it will only merge dictionaries if all the tags agree (when present). this (implicitly) includes the AN tag, so the two dictionaries will have to have the same alternative names. So as written it is safe. 

If we want the method to merge the following contigs:

dict1: 
@SN ID=name1 AN=n1

and dict2:
@SN ID=n1 AN=name1

or even with dict3:
@SN ID=nana1 AN=name1

then some additional smarts would have to be written....

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
yfarjoun: I did it...but I'm not sure what code you think I can remove now....

src/test/java/htsjdk/samtools/SAMSequenceRecordTest.java
```diff
@@ -96,6 +103,104 @@ public void testIsSameSequence(final SAMSequenceRecord rec1 , final SAMSequenceR
...
+
+        try (SAMFileWriter samWriter = new SAMFileWriterFactory().setCreateIndex(true).makeWriter(fileHeader, false, outputFile, null)) {
+            samReader.forEach(samWriter::addAlignment);
+        }
+        AccessFileWithAlternateContigName = outputFile;
+    }
+
+    @Test(dataProvider = "testAccessFileWithAlternateContigNameData")
+    public void testAccessFileWithAlternateContigName(final String contigName, final int expectedRecords) throws IOException {
+        try(SamReader bamReader = SamReaderFactory.make().open(AccessFileWithAlternateContigName);
```
yfarjoun, line 138: this tests querying a bam file with alternative names

src/test/java/htsjdk/samtools/SAMSequenceRecordTest.java
```diff
@@ -118,7 +226,9 @@ public void testSetAndCheckDescription() {
                 {"lparen("},
                 {"rparen)"},
                 {"lbracket{"},
-                {"rbracket}"}};
+                {"rbracket}"},
+                {""}
```
yfarjoun, line 230: this shows that an empty name is illegal

yfarjoun:
back to you @lbergelson

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
lbergelson: That's probably a good idea provided the md5's are the same, but it can be a different pr...

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
lbergelson: I thought you could remove the various guards in the code that call into this and let this be the single source of error checking.

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -98,68 +108,186 @@ public SAMSequenceRecord(final String name, final int sequenceLength) {
...
+
+    /**
+     * Sets the alternative sequence names in the order provided by iteration, removing the previous values.
+     */
+    public void setAlternativeSequenceName(final Collection<String> alternativeSequences) {
+        if (alternativeSequences == null) {
+            setAttribute(ALTERNATIVE_SEQUENCE_NAME_TAG, null);
+        } else {
+            // validate all the names and encode afterwards
+            alternativeSequences.forEach(SAMSequenceRecord::validateAltRegExp);
```
lbergelson, line 197: we can remove this validation since it's done encodeAltSequences now

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -98,68 +108,186 @@ public SAMSequenceRecord(final String name, final int sequenceLength) {
...
+        final String anTag = getAttribute(ALTERNATIVE_SEQUENCE_NAME_TAG);
+        return (anTag == null) ? Collections.emptySet()
+                : Collections.unmodifiableSet(new LinkedHashSet<>(Arrays.asList(anTag.split(ALTERNATIVE_SEQUENCE_NAME_SEPARATOR))));
+    }
+
+    /**
+     * Adds an alternative sequence name if it is not the same as the sequence name or it is not present already.
+     */
+    public void addAlternativeSequenceName(final String name) {
+        validateAltRegExp(name);
```
lbergelson, line 180: we can remove this validation since it's done encodeAltSequences now

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -98,68 +108,186 @@ public SAMSequenceRecord(final String name, final int sequenceLength) {
...
+        }
+        if (that == null) {
+            return false;
+        }
 
-        if (mSequenceIndex != that.mSequenceIndex) return false;
+        if (mSequenceIndex != that.mSequenceIndex) {
+            return false;
+        }
         // PIC-439.  Allow undefined length.
```
lbergelson, line 240: Do we know if this is still an issue?

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
yfarjoun: 👍

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
yfarjoun: 👍

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
yfarjoun: Not sure what the numbering scheme refers to….and I can't find it online either….however, it does not pertain to this PR…

PR #1475, opened 2020-05-02
Mf get read position from ref

fleharty:
### Description

Adding functionality to optionally include softclips in getAlignmentBlocks

- [ ] Check your code style.


tfenne:
@fleharty I really don't like this - it adds a bunch more public methods to `SAMRecord` for something that is rather questionable.  If you really want to treat soft clips as `M`s I feel like the more correct thing to do would be to update the cigar on the record, and then use the existing getReadPositionAtRefPosition() method(s).

fleharty:
@tfenne That's an interesting idea.  I'll try to accomplish this by modifying the CIGAR.  Thanks.

fleharty:
Closing this PR, I believe there is a better solution.

PR #1476, opened 2020-05-03
Switch to Java 11, Add GA4GH Crypt4H SeekableStream [WIP]

AlexanderSenf:
This is just for demonstration!

This is a follow-up on adding GA4GH Crypt4GH directly to HTSJDK. This is based on a new implementation of the Crypt4GH library, based on Java 11. It uses native Java functionality to support ChaCha20-Poly1305, the dependency to Google TINK is removed.

### Description

This is a discussion point, to determine the best way forward. There exists now a version of the Crypt4GH library that is no longer dependent on Google TINK, but uses native Java functionality; although the Blake2B dependency remains.

This will be dependent on the whole library switching to Java 11+. Is there any timeline for this?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,7 +37,8 @@ dependencies {
     compile "org.apache.commons:commons-compress:1.19"
     compile 'org.tukaani:xz:1.8'
     compile "gov.nih.nlm.ncbi:ngs-java:2.9.0"
-
+    implementation files('libs/crypt4gh-2.3.0-shaded.jar')
+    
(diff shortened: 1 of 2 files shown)


cmnbroad:
@AlexanderSenf Thanks for the update on this. We are making some progress on being able to upgrade to Java 11, but we're not quite there yet. Any idea how this new dependency affects the size of the htsjdk jar ?

tfenne:
Closing as stale / not followed up on.

PR #1480, opened 2020-05-12
Gff3BaseData precompute hash and Gff3Codec decodeShallow method

kachulis:
Four basic changes:
1) Since Gff3BaseData hash is based only on final fields, compute hash once when object is initialized.  This speeds up any code which uses Gff3Feature or Gff3BaseData as a key in a HashMap.

2) Make Gff3BaseData constructor public, (previously package-private) fields private, and add public getter methods to access those fields.

3) Add `decodeShallow` method, which returns only a flat Gff3Feature, which is NOT linked to parent/child/cofeatures.  While this method should not be used for general analysis, it is necessary to preprocess (specifically to add flush directives to) large gff3 files which do not already have flush directives.  Otherwise, the standard decode method will lead to the entire file being held in memory, and GC overhead problems.  

4) Adds score field to Gff3Feature.  This column is included in the spec, was somehow left out of the original Gff3 commit.

### DIFF
--- a/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
+++ b/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
@@ -3,6 +3,7 @@ package htsjdk.tribble.gff;
 import htsjdk.tribble.annotation.Strand;
 
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.Map;
 
 public class Gff3BaseData {
@@ -14,7 +15,7 @@ public class Gff3BaseData {
     private final String type;
     private final int start;
     private final int end;
-    private final Double score;
+    private final double score;
     private final Strand strand;
     private final int phase;
     private final Map<String, String> attributes;
@@ -34,11 +35,11 @@ public class Gff3BaseData {
         this.score = score;
         this.phase = phase;
         this.strand = strand;
-        this.attributes = Collections.unmodifiableMap(attributes);
+        this.attributes = Collections.unmodifiableMap(new LinkedHashMap<>(attributes));
         this.id = attributes.get(ID_ATTRIBUTE_KEY);
         this.name = attributes.get(NAME_ATTRIBUTE_KEY);
         this.alias = attributes.get(ALIAS_ATTRIBUTE_KEY);
-        hashCode = computeHashCode();
+        this.hashCode = computeHashCode();
     }
 
     @Override
@@ -56,7 +57,7 @@ public class Gff3BaseData {
                 otherBaseData.getType().equals(getType()) &&
                 otherBaseData.getStart() == getStart() &&
                 otherBaseData.getEnd() == getEnd() &&
-                otherBaseData.getScore().equals(score) &&
+                ((Double)otherBaseData.getScore()).equals(score) &&
                 otherBaseData.getPhase() == getPhase() &&
                 otherBaseData.getStrand().equals(getStrand()) &&
                 otherBaseData.getAttributes().equals(getAttributes());
@@ -92,7 +93,7 @@ public class Gff3BaseData {
         hash = 31 * hash + getType().hashCode();
         hash = 31 * hash + getStart();
         hash = 31 * hash + getEnd();
-        hash = 31 * hash + getScore().hashCode();
+        hash = 31 * hash + Double.hashCode(getScore());
         hash = 31 * hash + getPhase();
         hash = 31 * hash + getStrand().hashCode();
         hash = 31 * hash + getAttributes().hashCode();
@@ -131,7 +132,9 @@ public class Gff3BaseData {
         return end;
     }
 
-    public Double getScore() {return score;}
+    public double getScore() {
+        return score;
+    }
 
     public Strand getStrand() {
         return strand;
--- a/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
+++ b/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
@@ -1,11 +1,12 @@
 package htsjdk.tribble.gff;
 
+import htsjdk.samtools.util.Locatable;
 import htsjdk.tribble.annotation.Strand;
 
 import java.util.Collections;
 import java.util.Map;
 
-public class Gff3BaseData {
+public class Gff3BaseData implements Locatable {
     private static final String ID_ATTRIBUTE_KEY = "ID";
     private static final String NAME_ATTRIBUTE_KEY = "Name";
     private static final String ALIAS_ATTRIBUTE_KEY = "Alias";
--- a/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
+++ b/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
@@ -1,7 +1,5 @@
 package htsjdk.tribble.gff;
 
-import htsjdk.samtools.util.Locatable;
-import htsjdk.tribble.Feature;
 import htsjdk.tribble.annotation.Strand;
 
 import java.util.Collections;
@@ -22,7 +20,7 @@ public class Gff3BaseData {
     private final String id;
     private final String name;
     private final String alias;
-    final int hashCode;
+    private final int hashCode;
 
     public Gff3BaseData(final String contig, final String source, final String type,
                  final int start, final int end, final Strand strand, final int phase,
--- a/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
+++ b/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
@@ -14,6 +14,7 @@ public class Gff3BaseData {
     private final String type;
     private final int start;
     private final int end;
+    private final Double score;
     private final Strand strand;
     private final int phase;
     private final Map<String, String> attributes;
@@ -23,13 +24,14 @@ public class Gff3BaseData {
     private final int hashCode;
 
     public Gff3BaseData(final String contig, final String source, final String type,
-                 final int start, final int end, final Strand strand, final int phase,
+                 final int start, final int end, final Double score, final Strand strand, final int phase,
                  final Map<String, String> attributes) {
         this.contig = contig;
         this.source = source;
         this.type = type;
         this.start = start;
         this.end = end;
+        this.score = score;
         this.phase = phase;
         this.strand = strand;
         this.attributes = Collections.unmodifiableMap(attributes);
@@ -54,6 +56,7 @@ public class Gff3BaseData {
                 otherBaseData.getType().equals(getType()) &&
                 otherBaseData.getStart() == getStart() &&
                 otherBaseData.getEnd() == getEnd() &&
+                otherBaseData.getScore().equals(score) &&
                 otherBaseData.getPhase() == getPhase() &&
                 otherBaseData.getStrand().equals(getStrand()) &&
                 otherBaseData.getAttributes().equals(getAttributes());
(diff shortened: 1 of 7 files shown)


src/main/java/htsjdk/tribble/gff/Gff3Codec.java
```diff
@@ -49,6 +47,7 @@
     private static final int FEATURE_TYPE_INDEX = 2;
     private static final int START_LOCATION_INDEX = 3;
     private static final int END_LOCATION_INDEX = 4;
+    private static final int SCORE_INDEX = 5;
```
lbergelson, line 50: Woops, I probably should have used the power of counting here during the initial review...

src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
```diff
@@ -47,57 +51,109 @@ public boolean equals(Object other) {
...
+
+    public int getStart() {
+        return start;
+    }
+
+    public int getEnd() {
+        return end;
+    }
+
+    public Double getScore() {return score;}
```
lbergelson, line 134: linebreak style is inconsistent here

src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
```diff
@@ -47,57 +51,109 @@ public boolean equals(Object other) {
...
-                otherBaseData.strand.equals(strand) &&
-                otherBaseData.attributes.equals(attributes);
-        if (id == null) {
-            ret = ret && otherBaseData.id == null;
+        boolean ret = otherBaseData.getContig().equals(getContig()) &&
+                otherBaseData.getSource().equals(getSource()) &&
+                otherBaseData.getType().equals(getType()) &&
+                otherBaseData.getStart() == getStart() &&
+                otherBaseData.getEnd() == getEnd() &&
+                otherBaseData.getScore().equals(score) &&
```
lbergelson, line 59: Did you autogenerate these?  I think you need to check for null since score is Double.

src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
```diff
@@ -47,57 +51,109 @@ public boolean equals(Object other) {
...
+        return hashCode;
+    }
+
+    private int computeHashCode() {
+        int hash = getContig().hashCode();
+        hash = 31 * hash + getSource().hashCode();
+        hash = 31 * hash + getType().hashCode();
+        hash = 31 * hash + getStart();
+        hash = 31 * hash + getEnd();
+        hash = 31 * hash + getScore().hashCode();
```
lbergelson, line 95: null check?

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
```diff
@@ -198,6 +180,31 @@ public Gff3Feature decode(final LineIterator lineIterator) throws IOException {
...
+            throw new TribbleException("Found an invalid number of columns in the given Gff3 file- Given: " + splitLine.size() + " Expected: " + NUM_FIELDS + " : " + line);
+        }
+
+        try {
+            final String contig = URLDecoder.decode(splitLine.get(CHROMOSOME_NAME_INDEX), "UTF-8");
+            final String source = URLDecoder.decode(splitLine.get(ANNOTATION_SOURCE_INDEX), "UTF-8");
+            final String type = URLDecoder.decode(splitLine.get(FEATURE_TYPE_INDEX), "UTF-8");
+            final int start = Integer.parseInt(splitLine.get(START_LOCATION_INDEX));
+            final int end = Integer.parseInt(splitLine.get(END_LOCATION_INDEX));
+            final Double score = splitLine.get(SCORE_INDEX).equals(".") ? -1 : Double.parseDouble(splitLine.get(SCORE_INDEX));
```
lbergelson, line 196: looks like it defaults to -1, so lets make score lowercase double instead to avoid any null issues.

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
```diff
@@ -198,6 +180,31 @@ public Gff3Feature decode(final LineIterator lineIterator) throws IOException {
...
+            final int phase = splitLine.get(GENOMIC_PHASE_INDEX).equals(".") ? -1 : Integer.parseInt(splitLine.get(GENOMIC_PHASE_INDEX));
+            final Strand strand = Strand.decode(splitLine.get(GENOMIC_STRAND_INDEX));
+            final Map<String, String> attributes = parseAttributes(splitLine.get(EXTRA_FIELDS_INDEX));
+            return new Gff3BaseData(contig, source, type, start, end, score, strand, phase, attributes);
+        } catch (final NumberFormatException ex ) {
+            throw new TribbleException("Cannot read integer value for start/end position!", ex);
```
lbergelson, line 202: You should include the at least the entirety of the line here, and I would probably try to include the line number as well.

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
```diff
@@ -198,6 +180,31 @@ public Gff3Feature decode(final LineIterator lineIterator) throws IOException {
         return attributes;
     }
 
+    static private Gff3BaseData parseLine(final String line) {
+        final List<String> splitLine = ParsingUtils.split(line, FIELD_DELIMITER);
+
+        if (splitLine.size() != NUM_FIELDS) {
+            throw new TribbleException("Found an invalid number of columns in the given Gff3 file- Given: " + splitLine.size() + " Expected: " + NUM_FIELDS + " : " + line);
```
lbergelson, line 187: I would pass the line number in here still.  It's useful enough for debugging that I think it's worth the cruft.

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
```diff
@@ -198,6 +180,31 @@ public Gff3Feature decode(final LineIterator lineIterator) throws IOException {
...
+            final int end = Integer.parseInt(splitLine.get(END_LOCATION_INDEX));
+            final Double score = splitLine.get(SCORE_INDEX).equals(".") ? -1 : Double.parseDouble(splitLine.get(SCORE_INDEX));
+            final int phase = splitLine.get(GENOMIC_PHASE_INDEX).equals(".") ? -1 : Integer.parseInt(splitLine.get(GENOMIC_PHASE_INDEX));
+            final Strand strand = Strand.decode(splitLine.get(GENOMIC_STRAND_INDEX));
+            final Map<String, String> attributes = parseAttributes(splitLine.get(EXTRA_FIELDS_INDEX));
+            return new Gff3BaseData(contig, source, type, start, end, score, strand, phase, attributes);
+        } catch (final NumberFormatException ex ) {
+            throw new TribbleException("Cannot read integer value for start/end position!", ex);
+        } catch (final IOException ex) {
+            throw new TribbleException("Cannot decode feature info", ex);
```
lbergelson, line 204: Likewise.

src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
```diff
@@ -9,32 +9,36 @@
...
-    final String alias;
-
-    Gff3BaseData(final String contig, final String source, final String type,
-                 final int start, final int end, final Strand strand, final int phase,
+    private final String contig;
+    private final String source;
+    private final String type;
+    private final int start;
+    private final int end;
+    private final Double score;
```
lbergelson, line 17: It looks like this is set to -1 as a sentinel for missing value.  That makes me thing this should be `double` so you don't have to worry about NPE anywhere.

src/test/java/htsjdk/tribble/gff/Gff3FeatureTest.java
```diff
@@ -232,12 +232,12 @@ public void testDerivesFrom() {
...
     @Test
     public void testFeatureWithUnLoadedParent() {
-        final Gff3FeatureImpl gene01 = new Gff3FeatureImpl("chrX", ".", "gene", 1, 35, Strand.POSITIVE, -1, ImmutableMap.of("ID", "gene01", "Parent", "region01"));
-        final Gff3FeatureImpl gene02 = new Gff3FeatureImpl("chrX", ".", "gene", 1, 35, Strand.POSITIVE, -1, ImmutableMap.of("ID", "gene01", "Parent", "region01"));
+        final Gff3FeatureImpl gene01 = new Gff3FeatureImpl("chrX", ".", "gene", 1, 35, -1d, Strand.POSITIVE, -1, ImmutableMap.of("ID", "gene01", "Parent", "region01"));
+        final Gff3FeatureImpl gene02 = new Gff3FeatureImpl("chrX", ".", "gene", 1, 35, -1d, Strand.POSITIVE, -1, ImmutableMap.of("ID", "gene01", "Parent", "region01"));
 
         Assert.assertEquals(gene01, gene02);
 
-        final Gff3FeatureImpl mRNA01 = new Gff3FeatureImpl("chrX", ".", "mRNA", 1, 100, Strand.POSITIVE, -1 , ImmutableMap.of("ID", "mRNA01", "Parent", "gene01, gene02"));
```
lbergelson, line 240: I didn't see any tests with a non -1 score?  Should there be?  Also, is decodeLoc tested anywhere

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
```diff
@@ -220,19 +227,22 @@ private void validateFeature(final Gff3Feature feature) {
 
     @Override
     public Feature decodeLoc(LineIterator lineIterator) {
+        return decodeShallow(lineIterator);
```
lbergelson, line 230: This is probably more expensive than the previous decodeLoc implementation, but I think it's a pretty underused operation anyway so it is unlikely to make a difference anywhere.  I think it's only used in certain reindexing operations.

src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
```diff
@@ -35,6 +36,7 @@
         this.id = attributes.get(ID_ATTRIBUTE_KEY);
         this.name = attributes.get(NAME_ATTRIBUTE_KEY);
         this.alias = attributes.get(ALIAS_ATTRIBUTE_KEY);
+        hashCode = computeHashCode();
```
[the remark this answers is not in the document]
kachulis: oh, that's tricky.  good catch, changed to copy.

src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
kachulis: fixed

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
kachulis: done

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
kachulis: done

src/test/java/htsjdk/tribble/gff/Gff3FeatureTest.java
kachulis: added some tests

kachulis:
@lbergelson thanks for the review!  I think I have addressed all of you comments.  Also made a bit of a change to how the shallow decoding is implemented, instead of having a 'decodeShallow' method, the class can now be either in shallow or deep (default) mode, and the decode method returns either a shallow feature or a fully linked feature accordingly.  The point is that this way I can still use the `AbstractFeatureReader.getFeatureReader` method for reading in features shallowly, instead of having to duplicate the code from the different FeatureReaders.

Just a heads up, there's actually 2 new commits  38a38bb as well as 4ae8ff0.  I first one I just made before your comments posted, so it's gotten buried back at the top of the thread.

back to you!

lbergelson:
@kachulis I think making it a mode in the decoder is an improvement!  Good idea.

PR #1486, opened 2020-06-23
Gff3 Writer

kachulis:
Adds a Gff3 Writer to write out Gff3 files.  Also adjusts the attributes field to account for the fact that attributes map a key to a list of strings, which are separated by commas.  As previously written, the comma separated list was left as a single string, which could later be processed into a list by the user.  This was vulnerable to situations where one of the entries in the list contained an encoded comma (ie %2C), so that after decoding the decoded comma could not be differentiated from the list separating commas.

### DIFF
--- a/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
+++ b/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
@@ -1,8 +1,8 @@
 package htsjdk.tribble.gff;
 
-import htsjdk.tribble.TribbleException;
 import htsjdk.tribble.annotation.Strand;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -37,10 +37,22 @@ public class Gff3BaseData {
         this.score = score;
         this.phase = phase;
         this.strand = strand;
-        this.attributes = Collections.unmodifiableMap(new LinkedHashMap<>(attributes));
+        this.attributes = copyAttributesSafely(attributes);
         this.id = Gff3Codec.extractSingleAttribute(attributes.get(ID_ATTRIBUTE_KEY));
         this.name = Gff3Codec.extractSingleAttribute(attributes.get(NAME_ATTRIBUTE_KEY));
-        this.aliases = attributes.containsKey(ALIAS_ATTRIBUTE_KEY)? attributes.get(ALIAS_ATTRIBUTE_KEY) : Collections.emptyList();   this.hashCode = computeHashCode();
+        this.aliases = attributes.getOrDefault(ALIAS_ATTRIBUTE_KEY, Collections.emptyList());
+        this.hashCode = computeHashCode();
+    }
+
+    private static Map<String, List<String>> copyAttributesSafely(final Map<String, List<String>> attributes) {
+        final Map<String, List<String>> modifiableDeepMap = new LinkedHashMap<>();
+
+        for (final Map.Entry<String, List<String>> entry : attributes.entrySet()) {
+            final List<String> unmodifiableDeepList = Collections.unmodifiableList(new ArrayList<>(entry.getValue()));
+            modifiableDeepMap.put(entry.getKey(), unmodifiableDeepList);
+        }
+
+        return Collections.unmodifiableMap(modifiableDeepMap);
     }
 
     @Override
@@ -102,9 +114,7 @@ public class Gff3BaseData {
             hash = 31 * hash + getName().hashCode();
         }
 
-        for (final String alias : aliases) {
-            hash = 31 * hash + alias.hashCode();
-        }
+        hash = 31 * hash + aliases.hashCode();
 
         return hash;
     }
@@ -146,7 +156,7 @@ public class Gff3BaseData {
     }
 
     public List<String> getAttribute(final String key) {
-        return attributes.containsKey(key)? attributes.get(key) : Collections.emptyList();
+        return attributes.getOrDefault(key, Collections.emptyList());
     }
 
     public String getId() {
--- a/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
+++ b/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
@@ -1,12 +1,13 @@
 package htsjdk.tribble.gff;
 
+import htsjdk.samtools.util.Locatable;
 import htsjdk.tribble.annotation.Strand;
 
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
-public class Gff3BaseData {
+public class Gff3BaseData implements Locatable {
     private static final String ID_ATTRIBUTE_KEY = "ID";
     private static final String NAME_ATTRIBUTE_KEY = "Name";
     private static final String ALIAS_ATTRIBUTE_KEY = "Alias";
--- a/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
+++ b/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
@@ -1,9 +1,11 @@
 package htsjdk.tribble.gff;
 
+import htsjdk.tribble.TribbleException;
 import htsjdk.tribble.annotation.Strand;
 
 import java.util.Collections;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 
 public class Gff3BaseData {
@@ -18,15 +20,15 @@ public class Gff3BaseData {
     private final double score;
     private final Strand strand;
     private final int phase;
-    private final Map<String, String> attributes;
+    private final Map<String, List<String>> attributes;
     private final String id;
     private final String name;
-    private final String alias;
+    private final List<String> aliases;
     private final int hashCode;
 
     public Gff3BaseData(final String contig, final String source, final String type,
                  final int start, final int end, final Double score, final Strand strand, final int phase,
-                 final Map<String, String> attributes) {
+                 final Map<String, List<String>> attributes) {
         this.contig = contig;
         this.source = source;
         this.type = type;
@@ -36,10 +38,9 @@ public class Gff3BaseData {
         this.phase = phase;
         this.strand = strand;
         this.attributes = Collections.unmodifiableMap(new LinkedHashMap<>(attributes));
-        this.id = attributes.get(ID_ATTRIBUTE_KEY);
-        this.name = attributes.get(NAME_ATTRIBUTE_KEY);
-        this.alias = attributes.get(ALIAS_ATTRIBUTE_KEY);
-        this.hashCode = computeHashCode();
+        this.id = Gff3Codec.extractSingleAttribute(attributes.get(ID_ATTRIBUTE_KEY));
+        this.name = Gff3Codec.extractSingleAttribute(attributes.get(NAME_ATTRIBUTE_KEY));
+        this.aliases = attributes.containsKey(ALIAS_ATTRIBUTE_KEY)? attributes.get(ALIAS_ATTRIBUTE_KEY) : Collections.emptyList();   this.hashCode = computeHashCode();
     }
 
     @Override
@@ -73,11 +74,7 @@ public class Gff3BaseData {
             ret = ret && otherBaseData.getName() != null && otherBaseData.getName().equals(getName());
         }
 
-        if (getAlias() == null) {
-            ret = ret && otherBaseData.getAlias() == null;
-        } else {
-            ret = ret && otherBaseData.getAlias() != null && otherBaseData.getAlias().equals(getAlias());
-        }
+        ret = ret && otherBaseData.getAliases().equals(getAliases());
 
         return ret;
     }
@@ -105,8 +102,8 @@ public class Gff3BaseData {
             hash = 31 * hash + getName().hashCode();
         }
 
-        if (getAlias() != null) {
-            hash = 31 * hash + getAlias().hashCode();
+        for (final String alias : aliases) {
+            hash = 31 * hash + alias.hashCode();
         }
 
         return hash;
@@ -144,8 +141,12 @@ public class Gff3BaseData {
         return phase;
     }
 
-    public Map<String, String> getAttributes() {
+    public Map<String, List<String>> getAttributes() {
         return attributes;
+    }
+
+    public List<String> getAttribute(final String key) {
+        return attributes.containsKey(key)? attributes.get(key) : Collections.emptyList();
     }
 
     public String getId() {
@@ -156,7 +157,7 @@ public class Gff3BaseData {
         return name;
     }
 
-    public String getAlias() {
-        return alias;
+    public List<String> getAliases() {
+        return aliases;
     }
 }
--- a/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
+++ b/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
@@ -1,13 +1,14 @@
 package htsjdk.tribble.gff;
 
 import htsjdk.samtools.util.Locatable;
+import htsjdk.tribble.Feature;
 import htsjdk.tribble.annotation.Strand;
 
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
-public class Gff3BaseData implements Locatable {
+public class Gff3BaseData {
     private static final String ID_ATTRIBUTE_KEY = "ID";
     private static final String NAME_ATTRIBUTE_KEY = "Name";
     private static final String ALIAS_ATTRIBUTE_KEY = "Alias";
--- a/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
+++ b/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
@@ -1,7 +1,5 @@
 package htsjdk.tribble.gff;
 
-import htsjdk.samtools.util.Locatable;
-import htsjdk.tribble.Feature;
 import htsjdk.tribble.annotation.Strand;
 
 import java.util.Collections;
--- a/src/main/java/htsjdk/tribble/gff/Gff3Codec.java
+++ b/src/main/java/htsjdk/tribble/gff/Gff3Codec.java
@@ -67,7 +67,7 @@ public class Gff3Codec extends AbstractFeatureCodec<Gff3Feature, LineIterator> {
     private final Map<String, Set<Gff3FeatureImpl>> activeFeaturesWithIDs = new HashMap<>();
     private final Map<String, Set<Gff3FeatureImpl>> activeParentIDs = new HashMap<>();
 
-    private final Map<String, SequenceRegion> sequenceRegionMap = new HashMap<>();
+    private final Map<String, SequenceRegion> sequenceRegionMap = new LinkedHashMap<>();
     private final List<String> comments = new ArrayList<>();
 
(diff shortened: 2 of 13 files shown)


src/main/java/htsjdk/tribble/gff/Gff3FeatureImpl.java
```diff
@@ -116,8 +116,8 @@ public Gff3BaseData getBaseData() {
     }
 
     private Set<Gff3FeatureImpl> getDescendents(final Set<String> idsInLineage) {
-        final List<Gff3FeatureImpl> childrenToAdd = children.stream().filter(c -> c.getAttribute(DERIVES_FROM_ATTRIBUTE_KEY) == null ||
-                idsInLineage.contains(c.getAttribute(DERIVES_FROM_ATTRIBUTE_KEY))).
+        final List<Gff3FeatureImpl> childrenToAdd = children.stream().filter(c -> c.getAttribute(DERIVES_FROM_ATTRIBUTE_KEY).isEmpty() ||
+                !Collections.disjoint(idsInLineage, new HashSet<>(c.getAttribute(DERIVES_FROM_ATTRIBUTE_KEY)))).
```
lbergelson, line 120: I don't think it's necessary to create a set here since it accepts collections and then tries to figure out the fastest way to check them.

src/main/java/htsjdk/tribble/gff/Gff3Writer.java
```diff
@@ -0,0 +1,103 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.zip.GZIPOutputStream;
+
+
+public class Gff3Writer implements Closeable {
```
lbergelson, line 22: javadoc

src/main/java/htsjdk/tribble/gff/Gff3Writer.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private final PrintStream out;
```
lbergelson, line 24: I would not use a PrintStream.  It doesn't throw exceptions when things go wrong during writing but instead sets a flag you have to manually check.  It's designed for writing to interactive consoles and should basically never be used when writing files.  It is used badly in  a number of places in htsjdk but we should avoid new ones and work on stamping out the old ones..  Maybe just a BufferedOutputStream and write linebreaks manually?   Does gff3 define a specific newline character or is it platform specific?

src/main/java/htsjdk/tribble/gff/Gff3Writer.java
```diff
@@ -0,0 +1,103 @@
...
+    private final static String version = "3.1.25";
+
+    public Gff3Writer(final Path path) throws IOException {
+        if (!FileExtensions.GFF3.stream().anyMatch(e -> path.toString().endsWith(e))) {
```
lbergelson, line 28: ! anyMatch -> noneMatch for clarity

src/main/java/htsjdk/tribble/gff/Gff3Writer.java
```diff
@@ -0,0 +1,103 @@
...
+            throw new TribbleException("File " + path + " does not have extension consistent with gff3");
+        }
+
+        final OutputStream outputStream = IOUtil.hasGzipFileExtension(path)? new GZIPOutputStream(Files.newOutputStream(path)) : Files.newOutputStream(path);
```
lbergelson, line 32: The result of Files.newOutputStream needs to be wrapped in a buffering stream

src/main/java/htsjdk/tribble/gff/Gff3Writer.java
```diff
@@ -0,0 +1,103 @@
...
+        out = stream;
+        initialize();
+    }
+
+    private void initialize() {
+        out.println(Gff3Codec.Gff3Directive.VERSION3_DIRECTIVE.encode(version));
+    }
+
+    public void addFeature(final Gff3Feature feature) {
+        final String lineNoAttributes = String.join("\t",
```
lbergelson, line 48: It's more performant to directly `.append` each encoded value rather than convert them to a bigger string first.  That just adds layers of internal copies.  If you think this is more readable that's fine though.

src/main/java/htsjdk/tribble/gff/Gff3Writer.java
```diff
@@ -0,0 +1,103 @@
...
+    static String encodeForNinthColumn(final List<String> values) {
+        final List<String> encodedValues = values.stream().map(Gff3Writer::encodeString).collect(Collectors.toList());
+
+        return String.join(",", encodedValues);
+    }
+
+
+    static String encodeString(final String s) {
+        try {
+            return URLEncoder.encode(s, "UTF-8").replace("+", " ");
```
lbergelson, line 80: what's with removing +?

src/main/java/htsjdk/tribble/gff/SequenceRegion.java
```diff
@@ -9,7 +9,8 @@
     private final int start;
     private final int end;
     private final String contig;
-    private boolean isCircular;
+    private Boolean isCircular;
+    private int hashCode;
```
lbergelson, line 13: this should be final

src/main/java/htsjdk/tribble/gff/SequenceRegion.java
```diff
@@ -41,9 +43,28 @@ void setCircular() {
...
+
+        if (!(other instanceof SequenceRegion)) {
+            return false;
+        }
+
+        final SequenceRegion otherSequenceRegion = (SequenceRegion) other;
+        return otherSequenceRegion.start == start && otherSequenceRegion.end==end && otherSequenceRegion.contig.equals(contig) && otherSequenceRegion.isCircular == isCircular;
     }
 
+    private int computeHashCode() {
```
lbergelson, line 60: This is cached but the value of isCircular can change which invalidates it.  I think you should either force recalculation of the hashcode when setting isCircular or leave it out of the hash code.

src/test/resources/htsjdk/tribble/gff/url_encoding.gff3
```diff
@@ -1,3 +1,3 @@
 ##gff-version 3
 ##sequence-region   1 1 10
-the%20contig	a%20source%20%26%20also%20a%20str*)%25nge%20source	a%20region	1	10	.	.	.	ID=this%20is%20the%20ID%20of%20this%20wacky%20feature%5E%26%25%23%23%24%25*%26%3E%2C.%20%2C.; Another%20key=Another%3Dvalue
\ No newline at end of file
+the contig	a source %26 also a str*)%25nge source	a region	1	10	.	.	.	ID=this is the ID of this wacky feature%5E%26%25%23%23%24%25*%26%3E%2C. %2C.; Another key=Another%3Dvalue,And a second%2C value
```
lbergelson, line 3: Why did this test have to change?

src/test/java/htsjdk/tribble/gff/Gff3WriterTest.java
```diff
@@ -0,0 +1,155 @@
...
+    @Test(dataProvider = "roundTripDataProvider")
+    public void testRoundTrip(final Path path, final int expectedCompression) {
+
+        final HashSet<String> comments1 = new HashSet<>();
+        final HashSet<SequenceRegion> regions1 = new HashSet<>();
+        final LinkedHashSet<Gff3Feature> features1 = readFromFile(path, comments1, regions1);
+
+            //write out to temp files (one gzipped, on not)
+        try {
+            final Path tempFile = Files.createTempFile("gff3Writer", ".gff3");
```
lbergelson, line 54: Use one of the IOUtil.newTempPath methods to be sure the files are deleted.

src/main/java/htsjdk/tribble/gff/Gff3Writer.java
lbergelson, line 32: Avoid GZIPOutputStream.  GZIP streams in java are moderately cursed.  If you really mean GZIP and not BGZIP you can get a stream with IOUtil.openGzipFileForWriting.  

Are these files not indexable in any way?  I can't remember... Are they not sorted?  If they are meant to be randomly accessible you should really be using a bgzip output.

src/test/java/htsjdk/tribble/gff/Gff3WriterTest.java
lbergelson: I know you want to use a dataprovider and not duplicate everything, but maybe you can factor out out a function that runs on each file instead of just doubling the code, and then call it twice here?

src/test/java/htsjdk/tribble/gff/Gff3WriterTest.java
```diff
@@ -0,0 +1,155 @@
...
+                throw new TribbleException("Error writing gff3 files", ex);
+            }
+
+            //read temp files back in
+
+
+            final long unzippedSize = tempFile.toFile().length();
+            final long zippedSize = tempFileGzip.toFile().length();
+
+            Assert.assertTrue(unzippedSize > zippedSize * expectedCompression);
```
lbergelson, line 82: This is kind of weird to be testing the compression ratio here?  Why not just test that the file is zipped?

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
```diff
@@ -123,6 +125,7 @@ private Gff3Feature decode(final LineIterator lineIterator, final DecodeDepth de
         }
 
         if (line.startsWith(COMMENT_START) && !line.startsWith(DIRECTIVE_START)) {
+            comments.add(line.substring(1));
```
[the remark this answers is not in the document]
kachulis: good call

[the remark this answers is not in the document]
kachulis: I can't think of an obvious reason to want the line number of a particular comment, but that probably means that at some point someone will want it.  So I'll include it as a pair with the comment text.

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
```diff
@@ -190,15 +191,15 @@ private Gff3Feature decode(final LineIterator lineIterator, final DecodeDepth de
...
      * @throws UnsupportedEncodingException
      */
-    static private Map<String,String> parseAttributes(final String attributesString) throws UnsupportedEncodingException {
-        final Map<String, String> attributes = new LinkedHashMap<>();
+    static private Map<String, List<String>> parseAttributes(final String attributesString) throws UnsupportedEncodingException {
+        final Map<String, List<String>> attributes = new LinkedHashMap<>();
         final List<String> splitLine = ParsingUtils.split(attributesString,ATTRIBUTE_DELIMITER);
         for(String attribute : splitLine) {
             final List<String> key_value = ParsingUtils.split(attribute,KEY_VALUE_SEPARATOR);
             if (key_value.size()<2) {
```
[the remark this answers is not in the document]
kachulis: That does seem reasonable.  Honestly, I can't see why failure of this check leads to a continue as opposed to throwing an exception, since this indicates a invalid file.

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
```diff
@@ -228,6 +229,12 @@ static private Gff3BaseData parseLine(final String line, final int currentLine)
         }
     }
 
+    public Collection<SequenceRegion> getSequenceRegions() {
+        return sequenceRegionMap.values();
+    }
+
+    public List<String> getComments() {return comments;}
```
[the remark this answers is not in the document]
kachulis: Added javadoc and made unmodifiable.  I believe the comment actually excludes the leading #

src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
```diff
@@ -36,10 +38,9 @@ public Gff3BaseData(final String contig, final String source, final String type,
         this.phase = phase;
         this.strand = strand;
         this.attributes = Collections.unmodifiableMap(new LinkedHashMap<>(attributes));
```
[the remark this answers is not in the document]
kachulis: uggh, you're right, that's nasty.  I have implemented a deep unmodifiable copy to do this safely.

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
```diff
@@ -465,6 +546,12 @@ public static Gff3Directive toDirective(final String line) {
         public Object decode(final String line) throws IOException {
             return null;
         }
+
+        abstract public String encode(final Object object);
+
+        public String encode() {
```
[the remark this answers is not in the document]
kachulis: fair question, removed

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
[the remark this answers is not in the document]
lbergelson: It would be useful if you want to roundrip the file with modifications to keep comments near the lines they came from.

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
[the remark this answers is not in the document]
lbergelson: Lets make an exception then..

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
```diff
@@ -465,6 +546,12 @@ public static Gff3Directive toDirective(final String line) {
         public Object decode(final String line) throws IOException {
```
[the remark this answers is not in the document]
kachulis: I don't think there's any reason for these to be public, so made them protected instead

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
```diff
@@ -425,9 +456,34 @@ public TabixFormat getTabixFormat() {
...
-        VERSION3_DIRECTIVE("##gff-version\\s+3(?:.\\d)*(?:\\.\\d)*$"),
+        VERSION3_DIRECTIVE("##gff-version\\s+3(?:\\.\\d*)*$") {
+            @Override
+            public Object decode(final String line) throws IOException {
+                final String[] splitLine = line.split("\\s+");
+                return splitLine[1];
+            }
+
+            @Override
+            public String encode(final Object object) {
```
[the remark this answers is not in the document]
kachulis: yes, though not for the reason I initially thought.  Apparently `instanceof` will handle a null object just fine, but the `object.getClass()` will of course be a problem.

src/test/java/htsjdk/tribble/gff/Gff3WriterTest.java
kachulis: done

src/main/java/htsjdk/tribble/gff/Gff3Writer.java
kachulis: Good to know, changes to BufferedOutputStream.

src/main/java/htsjdk/tribble/gff/Gff3Writer.java
kachulis: done

src/main/java/htsjdk/tribble/gff/Gff3Writer.java
kachulis: As discussed offline, the gff3 spec is a bit inconsistent about which characters should and shouldn't be escaped.  Instead of trying to implement exactly all the different combinations mentioned in the spec, I settled on some reasonable middle ground.  The result is that I want to leave spaces unescaped.  Unfortunately, `URLEncoder.encode` changes all spaces to `+`, and this is in fact hardcoded in.  There is actually a pretty good solution to this in guava's "PercentEscaper" class, which allows space to + conversion to be turned on and off, as well as particular safe and unsafe characters to be specified.  However in our offline discussion we decided that this single situation isn't worth adding a guava dependency to htsjdk.  So this seems to be the simplest solution to leave spaces unescaped.  Also, I check and "+" ends up being escaped to "%2B", so there is no loss of information there.

src/test/resources/htsjdk/tribble/gff/url_encoding.gff3
kachulis: Just to model correct formatting.  Spaces shouldn't be escaped (though the codec actually doesn't care either way).

src/main/java/htsjdk/tribble/gff/Gff3Writer.java
kachulis: They can be sorted, in which case they can be indexed.  Changed to bgzip

src/test/java/htsjdk/tribble/gff/Gff3WriterTest.java
kachulis: agreed, that's much nicer.  done

kachulis:
@lbergelson thanks for the review, back to you.

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
```diff
@@ -217,8 +213,8 @@ static private Gff3BaseData parseLine(final String line, final int currentLine)
             final String type = URLDecoder.decode(splitLine.get(FEATURE_TYPE_INDEX), "UTF-8");
             final int start = Integer.parseInt(splitLine.get(START_LOCATION_INDEX));
             final int end = Integer.parseInt(splitLine.get(END_LOCATION_INDEX));
-            final double score = splitLine.get(SCORE_INDEX).equals(".") ? -1 : Double.parseDouble(splitLine.get(SCORE_INDEX));
-            final int phase = splitLine.get(GENOMIC_PHASE_INDEX).equals(".") ? -1 : Integer.parseInt(splitLine.get(GENOMIC_PHASE_INDEX));
+            final double score = splitLine.get(SCORE_INDEX).equals(Gff3Constants.UNDEFINED_FIELD_VALUE) ? -1 : Double.parseDouble(splitLine.get(SCORE_INDEX));
```
lbergelson, line 216: These read much better now, thank you.

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
```diff
@@ -229,11 +225,29 @@ static private Gff3BaseData parseLine(final String line, final int currentLine)
...
+     * Get list of sequence regions parsed by the codec.
+     * @return list of sequence regions
+     */
+    public List<SequenceRegion> getSequenceRegions() {
+        return Collections.unmodifiableList(new ArrayList<>(sequenceRegionMap.values()));
     }
 
-    public List<String> getComments() {return comments;}
+    /**
+     * Gets map from line number to comment found on that line.  The text of the comment EXCLUDES the leading # which indicates a comment line.
```
lbergelson, line 237: 👍

src/main/java/htsjdk/tribble/gff/Gff3Codec.java
```diff
@@ -229,11 +225,29 @@ static private Gff3BaseData parseLine(final String line, final int currentLine)
...
+     * @return Map from line number to comment found on line
+     */
+    public Map<Integer, String> getCommentsWithLineNumbers() {
+        return Collections.unmodifiableMap(new LinkedHashMap<>(commentsWithLineNumbers));
```
lbergelson, line 241: I think it's unnecessary to copy the map AND make it unmodifiable, but i seriously doubt this is going to be a performance issue ever

PR #1491, opened 2020-07-13
Remove inaccurate downstream status badges

lbergelson:
### Description

Removing downstream status badges which haven't been working for a while due to me not fixing our jenkins server configuration.  We shut down the server for now rather than burning money on it.  I should fix it but I'm hoping Carrot will fix the problem for us soon.


yfarjoun:
"carrot"? I don't think that folks here have any idea what you are talking about.....perhaps we do need to set up a public-facing service that will test downstream products using the master htsjdk.....

lbergelson:
@yfarjoun I agree.  The downstream tests are super useful to me but I don't have the bandwidth to fix them right now.  Carrot is the name of testing service that is being developed by the GATK team that should be useable for this sort of thing.

PR #1494, opened 2020-07-23
Add HtsgetBAMFileReader

andersleung:
### Description

Add a new SamReader type that is able to request reads from HTSget sources, as well as classes defining the content of an HTSget request and response. This change allows consumers of the htsjdk library to use HTSget sources like any other type of reads source.

- [x] Check your code style.


src/main/java/htsjdk/samtools/HtsgetBAMFileReader.java
```diff
@@ -0,0 +1,561 @@
...
+            if (!this.iterators.hasNext()) {
+                this.currentIterator = null;
+                return;
+            }
+            final Interval currInterval = this.intervals[this.currentIntervalIndex];
+            final Interval prevInterval = this.currentIntervalIndex == 0 ? null : this.intervals[this.currentIntervalIndex - 1];
+            final SamRecordFilter filter = new SamRecordFilter() {
+                @Override
+                public boolean filterOut(final SAMRecord record) {
+                    return record.getReadUnmappedFlag() && record.getAlignmentStart() != SAMRecord.NO_ALIGNMENT_START
```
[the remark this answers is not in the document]
lbergelson: I would make this anonymous class a named static class and explicitly pass the current/previous intervals and contained as constructor parameters.  Currently you have to read it carefully in order to understand if it has any state, how it interacts with the containing class, etc.  A static class is much easier to ignore the internals of.

src/main/java/htsjdk/samtools/HtsgetBAMFileReader.java
```diff
@@ -0,0 +1,561 @@
...
+
+    public static class BAMQueryFilteringIterator implements CloseableIterator<SAMRecord> {
+        /**
+         * The wrapped iterator.
+         */
+        private final CloseableIterator<SAMRecord> wrappedIterator;
+        /**
+         * The next record to be returned.  Will be null if no such record exists.
+         */
+        private SAMRecord mNextRecord;
```
lbergelson, line 503: I would just make this nextRecord.  The m to indicate it's a field is not something we do anymore (in new code at least).

src/main/java/htsjdk/samtools/HtsgetBAMFileReader.java
```diff
@@ -0,0 +1,561 @@
...
+                            CoordMath.encloses(prevInterval.getStart(), prevInterval.getEnd(), start, start);
+                    return matchesCurrInterval && !matchesPrevInterval;
+                }
+            };
+            this.currentIterator = new FilteringSamIterator(this.iterators.next().get(), filter, false);
+            this.currentIntervalIndex++;
+        }
+    }
+
+    public static class BAMQueryFilteringIterator implements CloseableIterator<SAMRecord> {
```
lbergelson, line 495: It's a shame there isn't a good way to avoid duplicating this code identically from BamReader

src/main/java/htsjdk/samtools/HtsgetBAMFileReader.java
```diff
@@ -0,0 +1,561 @@
...
+        SAMRecord advance() {
+            while (true) {
+                // Pull next record from stream
+                if (!this.wrappedIterator.hasNext()) {
+                    return null;
+                }
+
+                final SAMRecord record = this.wrappedIterator.next();
+                switch (this.iteratorFilter.compareToFilter(record)) {
+                    case MATCHES_FILTER:
```
lbergelson, line 544: in case statements like this I actually think putting the case and code on the same line is fine.

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -323,7 +325,25 @@ public SamReader open(final SamInputResource resource) {
                     return reader;
                   }
                 }
-                if (type == InputResource.Type.SEEKABLE_STREAM || type == InputResource.Type.URL) {
+                if (type == InputResource.Type.URI) {
+                    final URI uri = ((UriInputResource) data).uri;
+                    if (uri.getScheme().equalsIgnoreCase("htsget")) {
+                        final URI httpUri = new URI(
+                            "http", uri.getUserInfo(), uri.getHost(), uri.getPort(), uri.getPath(),
```
lbergelson, line 332: https?

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -323,7 +325,25 @@ public SamReader open(final SamInputResource resource) {
                     return reader;
                   }
                 }
-                if (type == InputResource.Type.SEEKABLE_STREAM || type == InputResource.Type.URL) {
+                if (type == InputResource.Type.URI) {
+                    final URI uri = ((UriInputResource) data).uri;
+                    if (uri.getScheme().equalsIgnoreCase("htsget")) {
+                        final URI httpUri = new URI(
```
lbergelson, line 331: I would extract out a function `HtsgetBAMFileReader.convertHtsgetUriToHttps` to document what this is doing.

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -323,7 +325,25 @@ public SamReader open(final SamInputResource resource) {
                     return reader;
                   }
                 }
-                if (type == InputResource.Type.SEEKABLE_STREAM || type == InputResource.Type.URL) {
+                if (type == InputResource.Type.URI) {
+                    final URI uri = ((UriInputResource) data).uri;
+                    if (uri.getScheme().equalsIgnoreCase("htsget")) {
```
lbergelson, line 330: this should reference a constant in the htsget reader or another htsget file.

src/main/java/htsjdk/samtools/SamInputResource.java
```diff
@@ -551,3 +559,48 @@ public SRAAccession asSRAAccession() {
         return accession;
     }
 }
+
+// TODO: replace this with an InputResource type taking PathSpecifier once this interface is pushed down to htsjdk
+class UriInputResource extends InputResource {
```
lbergelson, line 564: So I'm not totally sure what we should be doing here.  I'm thinking we might want to actually just add an HtsgetInputResource which specifically handles htsget and nothing else.  Then the of(URI, wrapper, wrapper) factory method can check if the URI is an htsget input and create one of those, or create one of the other existing types.  I think the logic would be:
1.  URI.scheme == htsget -> HtsgetInput
2. IOUtils.createPath succeeds -> PathInput
3. everything else: -> URLInput

The issue is that in GATK we exclusively use path and have special handling for it with the wrapper objects, so your idea of converting to URL and retrying is good and makes a lot of sense, but I think we need to go through Path as our first attempt.  

Maybe we can chat about it if what I'm saying doesn't make sense to you.

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -323,7 +325,25 @@ public SamReader open(final SamInputResource resource) {
                     return reader;
                   }
                 }
-                if (type == InputResource.Type.SEEKABLE_STREAM || type == InputResource.Type.URL) {
+                if (type == InputResource.Type.URI) {
```
lbergelson, line 328: Some of the comments here are might be irrelevant if we change how we do the input type like I mentioned above.

src/main/java/htsjdk/samtools/SamReaderFactory.java
```diff
@@ -523,6 +548,11 @@ void applyTo(final CRAMFileReader underlyingReader, final SamReader reader) {
             void applyTo(final SRAFileReader underlyingReader, final SamReader reader) {
                 underlyingReader.enableIndexCaching(true);
             }
+
+            @Override
+            void applyTo(final HtsgetBAMFileReader underlyingReader, final SamReader reader) {
```
lbergelson, line 553: This design is so crufty.  It seems like someone started refactoring it to use ReaderImplementation so we don't need these custom handlers here as well but then never actually did it.  (Just me venting... nothing to do here...)

src/main/java/htsjdk/samtools/util/htsget/HtsgetErrorResponse.java
```diff
@@ -0,0 +1,35 @@
+package htsjdk.samtools.util.htsget;
+
+
+/**
+ * Class allowing deserialization from json htsget error response
+ */
```
lbergelson, line 6: Note to self: stopping here for now.

src/main/java/htsjdk/samtools/HtsgetBAMFileReader.java
```diff
@@ -0,0 +1,561 @@
...
+    }
+
+    /**
+     * Generally loads data at a given point in the file.  Unsupported for HtsgetBAMFileReaders.
+     *
+     * @param fileSpan The file span.
+     * @return An iterator over the given file span.
+     */
+    @Override
+    public CloseableIterator<SAMRecord> getIterator(final SAMFileSpan fileSpan) {
```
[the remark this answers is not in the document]
andersleung: The javadoc mentions it's unsupported and the body throws UnsupportedOperationException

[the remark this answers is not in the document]
lbergelson: Sorry, I was just agreeing with your decision to not support it.

src/main/java/htsjdk/samtools/HtsgetBAMFileReader.java
```diff
@@ -0,0 +1,561 @@
...
+        final InputStream stream = req.getResponse().getDataStream();
+
+        final BlockCompressedInputStream compressedInputStream = this.mUseAsynchronousIO
+            ? new AsyncBlockCompressedInputStream(stream, this.mInflaterFactory)
+            : new BlockCompressedInputStream(stream, this.mInflaterFactory);
+        if (this.mCheckCRC) {
+            compressedInputStream.setCheckCrcs(true);
+        }
+
+        // Skip over the header
```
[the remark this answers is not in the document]
andersleung: If no class is specified then the entire file is returned, including header. There is an option to return only the body but it's not supported by the reference server yet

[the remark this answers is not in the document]
lbergelson: Oh, huh.  That could actually be really expensive for small queries because the header can be huge in some files.

src/main/java/htsjdk/samtools/util/htsget/HtsgetErrorResponse.java
```diff
@@ -0,0 +1,35 @@
...
+public class HtsgetErrorResponse {
```
lbergelson, line 7: Maybe put an example of what the json scheme looks like in a comment here?  Or a link to the spec.

src/main/java/htsjdk/samtools/HtsgetBAMFileReader.java
```diff
@@ -0,0 +1,561 @@
...
+import java.io.DataInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.util.*;
+
+/**
+ * Class for reading and querying BAM files from an htsget source
+ */
+public class HtsgetBAMFileReader implements SamReader.PrimitiveSamReader {
```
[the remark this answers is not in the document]
andersleung: The documentation for ReaderImplementation mentions that it's specifically for file based readers as opposed to non file based readers, and that the class might need to be removed anyway, so I thought it would make more sense to extend directly from PrimitiveSamReader

src/main/java/htsjdk/samtools/SamReaderFactory.java
andersleung: This is probably something that needs to be brought up at the next htsget call, but there isn't a way defined in the spec to indicate that a source is an htsget server and we're just using the scheme as a workaround, but this doesn't allow us to specify whether the actual backing source should be http or https. Another workaround to this might be to split the scheme into htsget and htsgets, corresponding to http and https. For the time being this defaults to http because not all servers might support https

src/main/java/htsjdk/samtools/util/htsget/HtsgetResponse.java
```diff
@@ -0,0 +1,174 @@
...
+        /**
+         * Parse a single data block from a json value
+         *
+         * @param blockJson json value representing a block
+         * @return parsed block object
+         */
+        public static Block parse(final mjson.Json blockJson) {
+            final Block block = new Block();
+
+            final mjson.Json uri = blockJson.at("url");
```
lbergelson, line 74: Are these .at operations case insensitive?

src/main/java/htsjdk/samtools/HtsgetBAMFileReader.java
```diff
@@ -0,0 +1,561 @@
...
+                               final boolean useAsynchronousIO,
+                               final InflaterFactory inflaterFactory) throws IOException {
+        this.mSource = source;
+        this.mEagerDecode = eagerDecode;
+        this.mValidationStringency = validationStringency;
+        this.mSamRecordFactory = samRecordFactory;
+        this.mUseAsynchronousIO = useAsynchronousIO;
+        this.mInflaterFactory = inflaterFactory;
+
+        final HtsgetRequest req = new HtsgetRequest(this.mSource).withDataClass(HtsgetClass.header);
```
lbergelson, line 84: We should always be setting the format here to BAM shouldn't we?  Otherwise we might get back cram header/ blocks which we can't parse? 
If we set the format to BAM is the server allowed to return CRAM anyway?

src/main/java/htsjdk/samtools/HtsgetBAMFileReader.java
```diff
@@ -0,0 +1,561 @@
...
+        // Request only the header and use it to construct a SAMFileHeader for this reader
+        try (final InputStream headerStream = req.getResponse().getDataStream()) {
```
lbergelson, line 86: I think we have to be checking the response objects for the FORMAT to be sure that they are BAM. (not sure if it matters for the headers, but definitely matters for the data).

src/main/java/htsjdk/samtools/util/htsget/HtsgetResponse.java
```diff
@@ -0,0 +1,174 @@
...
+        return this.md5;
+    }
+
+    /**
+     * Parses HtsgetResponse object from json string
+     *
+     * @param s json string
+     * @return parsed HtsgetResponse object
+     */
+    public static HtsgetResponse parse(final String s) {
```
lbergelson, line 126: We should be checking the API version header value IF we get it back and warning if it's not compatible with our version.

src/test/resources/htsjdk/samtools/BAMFileIndexTest/htsget_config.json
```diff
@@ -0,0 +1,10 @@
+{
```
lbergelson, line 1: Should this live with the script?  Can we add a README explaining this somewhere?

src/test/java/htsjdk/samtools/HtsgetBAMFileReaderTest.java
```diff
@@ -0,0 +1,255 @@
...
+             final SAMRecordIterator samRecordIterator = samReader.iterator()) {
+            Assert.assertEquals(samReader.getFileHeader().getSortOrder(), SAMFileHeader.SortOrder.coordinate);
+
+            int counter = 0;
+            while (samRecordIterator.hasNext()) {
+                final SAMRecord samRecord = samRecordIterator.next();
+                if (samRecord.getReferenceIndex() == SAMRecord.NO_ALIGNMENT_REFERENCE_INDEX) {
+                    break;
+                }
+                if (counter++ % 100 > 1) { // test only 1st and 2nd in every 100 to speed the test up:
```
lbergelson, line 57: How slow is it without this?

src/test/java/htsjdk/samtools/HtsgetBAMFileReaderTest.java
```diff
@@ -0,0 +1,255 @@
...
+                    samRecord.getReferenceName(),
+                    samRecord.getAlignmentStart());
+
+                Assert.assertTrue(iterator.hasNext(), counter + ": " + sam1);
+                final SAMRecord bamRecord = iterator.next();
+                final String sam2 = bamRecord.getSAMString();
+                Assert.assertEquals(samRecord.getReferenceName(), bamRecord.getReferenceName(), sam1 + sam2);
+
+                // default 'overlap' is true, so test records intersect the query:
+                Assert.assertTrue(CoordMath.overlaps(
```
lbergelson, line 72: You can do bamRecord.overlaps(samRecord) I think

src/test/java/htsjdk/samtools/HtsgetBAMFileReaderTest.java
```diff
@@ -0,0 +1,255 @@
...
+import htsjdk.samtools.util.CoordMath;
+import org.testng.Assert;
+import org.testng.annotations.BeforeTest;
+import org.testng.annotations.Test;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URI;
+
+public class HtsgetBAMFileReaderTest extends HtsjdkTest {
```
lbergelson, line 14: It would be good to run these tests with a variety of settings on the htsget reader.  Particularly I'd like to see them with async on for the htsget reader since it seems plausible that there could be potential issues there.  You could add a datprovider that has multiple readers.

The readers also need to be closed at the end.  

Typically I'm in favor of not reusing the same objects for multiple tests and generating them fresh each time, it helps avoid test ordering problems or if 1 test puts the objects into a broken state.  This should be OK though.

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -38,6 +39,9 @@
 public class SamReaderFactoryTest extends HtsjdkTest {
     private static final File TEST_DATA_DIR = new File("src/test/resources/htsjdk/samtools");
 
+    private static final String HTSGET_ENDPOINT = "htsget://127.0.0.1:3000/reads/";
```
lbergelson, line 42: Lets hardcore these in 1 place only and reference them here instead.  Otherwise it's a pain if we change the location.

src/test/java/htsjdk/samtools/util/HtsgetResponseUnitTest.java
```diff
@@ -0,0 +1,89 @@
...
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.URI;
+
+public class HtsgetResponseUnitTest extends HtsjdkTest {
+    @Test
+    public void testDeserialization() {
+        final String respJson = "{\"htsget\":{\"format\":\"BAM\",\"urls\":[{\"url\":\"data:application/vnd.ga4gh.bam;base64,QkFNAQ==\",\"class\":\"header\"},{\"url\":\"https://htsget.blocksrv.example/sample1234/header\",\"class\":\"header\"},{\"url\":\"https://htsget.blocksrv.example/sample1234/run1.bam\",\"headers\":{\"Authorization\":\"Bearer xxxx\",\"Range\":\"bytes=65536-1003750\"},\"class\":\"body\"},{\"url\":\"https://htsget.blocksrv.example/sample1234/run1.bam\",\"headers\":{\"Authorization\":\"Bearer xxxx\",\"Range\":\"bytes=2744831-9375732\"},\"class\":\"body\"}],\"md5\":\"blah\"}}";
```
lbergelson, line 21: I can't wait until java adds multiline strings...

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -38,6 +39,9 @@
 public class SamReaderFactoryTest extends HtsjdkTest {
```
lbergelson, line 39: We should also have a test here somehow for the fallback from URI -> URL or whatever we ultimately decide to do.

src/main/java/htsjdk/samtools/SamReaderFactory.java
lbergelson: Will the servers automatically redirect to https supported and you request http?  Defaulting to https seems good because it offers security but possibly more importantly data integrity which I don't think HTTP does?  Maybe we should implement a failover where we try https and then fall back to http if it fails?  Most of the time bam data is private and legally protected so defaulting to http seems a bit scary.

src/test/java/htsjdk/samtools/HtsgetBAMFileReaderTest.java
andersleung: It takes a few minutes before actually causing the process inside the docker container to hang so I think keeping this would be a good idea. The test including the skip was copied from the BAMFileReaderTest so presumably it was taking a while there too

src/test/java/htsjdk/samtools/HtsgetBAMFileReaderTest.java
andersleung: The try with resources block will close the iterators. I rewrote them this way because I was actually having the exact problem you mentioned, where earlier tests failing before they closed the iterator caused later tests to fail. The HtsgetBAMFileReader doesn't have the "one iterator open at a time" restriction like the regular reader so that shouldn't happen. 

I'll add the provider with async io since I would've needed to anyway to test the prefetching

src/main/java/htsjdk/samtools/HtsgetBAMFileReader.java
andersleung: BAM is the default so it should be interpreted as BAM without setting it explicitly. The spec says the server SHOULD return an error if it can't serve the request format, so no

src/main/java/htsjdk/samtools/util/htsget/HtsgetRequest.java
```diff
@@ -0,0 +1,273 @@
...
+            this.interval != null ||
+                !this.fields.isEmpty() ||
+                !this.tags.isEmpty() ||
+                !this.notags.isEmpty())) {
+            throw new IllegalArgumentException("Invalid request: no query parameters except `format` may be specified when class=header");
+        }
+
+        if (this.format != null) {
+            final String path = this.endpoint.getPath();
+            if ((path.endsWith(FileExtensions.BAM) || path.endsWith(FileExtensions.SAM) && (
```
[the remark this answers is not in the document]
andersleung: It should've been CRAM since htsget doesn't return SAM, fixed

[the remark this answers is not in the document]
andersleung: The path returned by getPath doesn't include queries or fragments so that shouldn't be a problem

src/main/java/htsjdk/samtools/SamReaderFactory.java
andersleung: Changed to default to https and fall back to http

src/main/java/htsjdk/samtools/util/htsget/HtsgetResponse.java
andersleung: Added a check for this

src/main/java/htsjdk/samtools/HtsgetBAMFileReader.java
```diff
@@ -0,0 +1,561 @@
...
+                    return false;
+                }
+
+                @Override
+                public SAMRecord next() {
+                    return null;
+                }
+            };
+        } else {
+            final HtsgetRequest req = new HtsgetRequest(this.mSource).withInterval(new Interval(sequence, start, Integer.MAX_VALUE));
```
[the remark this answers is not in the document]
andersleung: There's no way in htsget of just requesting only the reads that start at a specific site so the only thing we can do is request the whole contig starting from the site. The filter will stop processing reads as soon as it hits one that doesn't start at the site so it won't read the whole contig

src/test/resources/htsjdk/samtools/BAMFileIndexTest/htsget_config.json
andersleung: Moved to a folder with the script and added a readme

src/main/java/htsjdk/samtools/HtsgetBAMFileReader.java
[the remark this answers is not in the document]
lbergelson: Why not ask for the range (start, start+1)?  Instead of (start, end of contig).   Won't that give you a tighter range around what you want?

src/main/java/htsjdk/samtools/util/htsget/SAMRecordPrefetchingIterator.java
```diff
@@ -0,0 +1,85 @@
...
+
+import htsjdk.samtools.SAMRecord;
+import htsjdk.samtools.util.CloseableIterator;
+import htsjdk.samtools.util.PeekableIterator;
+
+import java.util.NoSuchElementException;
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class SAMRecordPrefetchingIterator implements CloseableIterator<SAMRecord> {
```
lbergelson, line 11: I think this class could use specific tests for it.  I think we could write tests that use a mocked input iterator that allows us to test various cases, (things like the reading throwing exceptions), very large reads compared to the number of tickets, etc

src/main/java/htsjdk/samtools/util/SAMRecordPrefetchingIterator.java
```diff
@@ -65,6 +67,7 @@ private void prefetch() {
                 // InterruptedException is expected if the iterator is being closed
                 return;
             } catch (final Throwable t) {
+                t.printStackTrace();
```
lbergelson, line 70: I think this shouldy be printed in the case of an Error, not in the normal case of an Exception.

lbergelson:
@andersleung 👍

PR #1499, opened 2020-08-13
Fix genotype validation error message

ldgauthier:
Actually report the VC alleles as the VC alleles, not the genotype alleles. Current error message is incorrect and not helpful.


src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -379,7 +379,7 @@ private static void validateGenotypes(final VariantContext variantContext) {
                     for (int j = 0, size = alleles.size(); j < size; j++) {
                         final Allele gAllele = alleles.get(j);
                         if (!variantContext.hasAllele(gAllele) && gAllele.isCalled()) {
-                            throw new IllegalStateException("Allele in genotype " + gAllele + " not in the variant context " + alleles);
+                            throw new IllegalStateException("Allele in genotype " + gAllele + " not in the variant context " + variantContext.getAlleles());
                         }
```
lbergelson, line 383: Oh no, that's extremely unhelpful.

PR #1500, opened 2020-08-17
No longer require external call to `initializeAnyploidPLIndexToAlleleIndices`

kachulis:
### Description
 
Instead of requiring anyploidy PLIndex caches to be initialized by an external call to `initializeAnyploidPLIndexToAlleleIndices`, calculate and cache PLIndex allele lists on an as needed basis.  Since `initializeAnyploidPLIndexToAlleleIndices` is public and `anyploidPloidyToPLIndexToAlleleIndices` is protected, they remain in the code to avoid breaking the API, but are deprecated.

- [ ] Check your code style.

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
+++ b/src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
@@ -502,7 +502,8 @@ public class GenotypeLikelihoods {
      * Initialize cache of allele anyploid indices
      * If initialized multiple times with the same ploidy, the alternate alleles from the last initialization will be used
      *
-     * This method is no longer necessary for this class, and in general should be avoided.  However it reamins
+     * This method is no longer necessary for this class, and in general should be avoided.  However it remains to avoid
+     * breaking the API
      *
      * @param altAlleles number of alternate alleles
      * @param ploidy    number of chromosomes
@@ -522,7 +523,6 @@ public class GenotypeLikelihoods {
 
     /**
      * Get the allele ploidy indices for the given PL index
-     * Must use the same ploidy as @see #initializeAnyploidPLIndexToAlleleIndices
      *
      * @param PLindex   the PL index
      * @param ploidy    number of chromosomes
--- a/src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
+++ b/src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
@@ -533,8 +533,8 @@ public class GenotypeLikelihoods {
             throw new IllegalStateException("The PL index " + PLindex + " cannot be negative");
         }
 
-        if (ploidy < 0) {
-            throw new IllegalStateException("The ploidy " + ploidy + " cannot be negative");
+        if (ploidy <= 0) {
+            throw new IllegalStateException("The ploidy " + ploidy + " must be greater than zero");
         }
 
         if ( ploidy == 2 ) { // diploid
--- a/src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
+++ b/src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
@@ -527,7 +527,7 @@ public class GenotypeLikelihoods {
      * @param PLindex   the PL index
      * @param ploidy    number of chromosomes
      * @return the ploidy allele indices
-     * @throws IllegalStateException if @see #anyploidPloidyToPLIndexAllelesList does not contain the requested ploidy or PL index
+     * @throws IllegalStateException if PLindex < 0 or ploidy < 0
      */
     public static synchronized List<Integer> getAlleles(final int PLindex, final int ploidy) {
         if ( PLindex < 0) {
--- a/src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
+++ b/src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
@@ -70,11 +70,10 @@ public class GenotypeLikelihoods {
      * For example, for a ploidy of 3, the allele lists for each PL index is:
      * {0,0,0}, {0,0,1}, {0,1,1}, {1,1,1}, {0,0,2}, {0,1,2}, {1,1,2}, {0,2,2}, {1,2,2}, {2,2,2}
      */
-    protected final static Map<Integer, PLIndexAllelesList> anyploidPloidyToPLIndexAllelesList = new HashMap<>();
+    private final static Map<Integer, PLIndexAllelesList> anyploidPloidyToPLIndexAllelesList = new HashMap<>();
 
     /**
-     * This field is no longer used in this class, and in general should be avoided in favor of anyploidPloidyToPLIndexAllelesList.  However, it remains to
-     * avoid breaking API since it is protected.
+     * This field is no longer used in this class.  However, it remains to avoid breaking API since it is protected.
      */
     @Deprecated
     protected final static Map<Integer, List<List<Integer>>> anyploidPloidyToPLIndexToAlleleIndices = new HashMap<Integer, List<List<Integer>>>();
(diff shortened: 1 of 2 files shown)


kachulis:
Some clarification on the justification for this PR.  While the PR was motivated as part of a performance improvement in gatk (broadinstitute/gatk#6754), I believe it is a reasonable change to htsjdk in its own right as well.

Currently (before this PR) in order to use the static method `GenotypeLikelihoods.getAlleles` with any ploidy other than 2, any calling code must first call `GenotypeLikelihoods.initializeAnyploidPLIndexToAlleleIndices`.  This increases the complexity of using GenotypeLikelihoods in non ploidy-2 situations.   Removing this additional complexity of use without sacrificing performance is the objective of this PR.

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -509,28 +522,98 @@ public static synchronized void initializeAnyploidPLIndexToAlleleIndices(final i
...
+         * are reset to 0, and all alleles to the right are left unchanged.
+         *
+         * The algorithm is initialized in thr constructor for PLIndex 0 with a list of P 0's and by setting the next allele to increment as ploidy - 1 (ie, the rightmost allele).
+         * @param PLIndex
+         * @return
+         */
+
+        private List<Integer> getAlleles(final int PLIndex) {
+
+            while(PLIndex >= allelesList.size()) {
```
lbergelson, line 596: This looks good to me but I don't have high confidence that I'm able to find errors in it by inspection.  There are a lot of possible off-by-one error sites.

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -67,6 +70,12 @@
      * For example, for a ploidy of 3, the allele lists for each PL index is:
      * {0,0,0}, {0,0,1}, {0,1,1}, {1,1,1}, {0,0,2}, {0,1,2}, {1,1,2}, {0,2,2}, {1,2,2}, {2,2,2}
      */
+    private final static Map<Integer, PLIndexAllelesList> anyploidPloidyToPLIndexAllelesList = new HashMap<>();
+
+    /**
+     * This field is no longer used in this class.  However, it remains to avoid breaking API since it is protected.
+     */
+    @Deprecated
```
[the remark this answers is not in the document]
lbergelson: Let's mark the class final as well since it effectively is already.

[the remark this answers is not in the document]
kachulis: oh nice!  Done.

src/test/java/htsjdk/variant/variantcontext/GenotypeLikelihoodsUnitTest.java
```diff
@@ -346,32 +346,44 @@ public void testInitializeAnyploidPLIndexToAlleleIndicesBadData(final int altAll
...
-                {7, 2, 3, Arrays.asList(0,2,2)},
-                {8, 2, 3, Arrays.asList(1,2,2)},
-                {9, 2, 3, Arrays.asList(2,2,2)},
-                {1, 2, 1, Arrays.asList(1)}
+                {0, 3, Arrays.asList(0,0,0)},
+                {1, 3, Arrays.asList(0,0,1)},
+                {2, 3, Arrays.asList(0,1,1)},
+                {3, 3, Arrays.asList(1,1,1)},
+                {4, 3, Arrays.asList(0,0,2)},
+                {5, 3, Arrays.asList(0,1,2)},
```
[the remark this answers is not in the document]
kachulis: good call, swapped up the order a bit and added some reruns of some earlier ones at the end.

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
kachulis: Agreed on the possibility for off-by-one errors.  I think the large "derived in reverse" tests in `GenotypeLikelihoodsUnitTests::testGetAllelesData` are a pretty good guard against this, and I added a bunch more randomly generated similar tests to be safe.  It seems quite unlikely that none of them would fail if there were an off-by-one error floating around here.

lbergelson: That was my thought too.  If it works empirically for a large set of things then it should continue to work.  It doesn't have a lot of obvious edge cases that don't also get hit on the way to a larger number.

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -492,10 +501,14 @@ public static GenotypeLikelihoodsAllelePair getAllelePair(final int PLindex) {
      * Initialize cache of allele anyploid indices
      * If initialized multiple times with the same ploidy, the alternate alleles from the last initialization will be used
      *
+     * This method is no longer necessary for this class, and in general should be avoided.  However it remains to avoid
+     * breaking the API
```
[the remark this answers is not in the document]
kachulis: I think those tests should just be dropped.  `testGetAlleles` does all the necessary testing, since we aren't building a cache separately from the `getAlleles` method anymore.

kachulis:
Thanks for the review @lbergelson!

Regarding performance, as far as I can tell, this code is essentially unused in gatk, picard, or the rest of htsjdk (save for the one issue referred to in broadinstitute/gatk#6754, which only uses code removed by this pr).  The only place in either gatk, picard, or htsjdk that uses `GenotypeLikelihoods::getAlleles` is in picard's `AlleleSubsettingUtils::subsetAlleles`, which is only implemented for ploidy 2 currently, so doesn't hit any of the anyPloidy code.  GATK appears to generally use its own `GenotypeLikelihoodCalculator` class instead of `GenotypeLikelihoods`.  So, to confirm performance, I wrote up a quick and hacky performance test (`GenotypeLikelihoodsUnitTest::performanceTest` in the `ck_anyploidy_new_performance_test` and `ck_anyploidy_new_performance_test` branches on my fork).  Ran with a few different ploidies, with results below.

ploidy 3:
old 3.52 s
new 2.02 s

ploidy 4:
old 2.74 s
new 3.37 s

ploidy 5:
old 3.18 s
new 2.08 s

which leads me to conclude that the performance is basically equivalent.

back to you!

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -534,7 +461,7 @@ public static synchronized void initializeAnyploidPLIndexToAlleleIndices(final i
         }
 
         if (ploidy <= 0) {
-            throw new IllegalStateException("The ploidy " + ploidy + " must be greater than zero");
+            throw new IllegalStateException("The ploidy must be greater than zero");
```
lbergelson, line 464: I would keep the actual value in the error message.  If it's ever triggered we'll want to know.

src/main/java/htsjdk/variant/variantcontext/GenotypeLikelihoods.java
```diff
@@ -489,48 +439,108 @@ public static GenotypeLikelihoodsAllelePair getAllelePair(final int PLindex) {
...
-     * @throws IllegalStateException if @see #anyploidPloidyToPLIndexToAlleleIndices does not contain the requested ploidy or PL index
+     * @throws IllegalStateException if PLindex < 0 or ploidy < 0
      */
     public static synchronized List<Integer> getAlleles(final int PLindex, final int ploidy) {
+        if ( PLindex < 0) {
+            throw new IllegalStateException("The PL index " + PLindex + " cannot be negative");
+        }
+
+        if (ploidy <= 0) {
+            throw new IllegalStateException("The ploidy must be greater than zero");
```
lbergelson, line 464:
```suggestion
            throw new IllegalStateException("The ploidy " + ploidy +" must be greater than zero");
```

lbergelson (approved):
@kachulis This looks good to me.  Thanks for looking into the performance.  I thought this was more widely used... Maybe it should be now that it isn't terrible.  I added a tiny commit to change an error message.  I'll merge when tests pass.

lbergelson:
@kachulis This is failing tests because of an issue in master, could you rebase? I can't push to your branch directly.

kachulis:
Thanks @lbergelson.  Just rebased, so hopefully tests will pass now.

PR #1501, opened 2020-08-21
Support very large sequence dictionaries/headers

nh13:
This PR is meant to start a discussion; nitpicks will be ignored.

When we have very large sequence headers, tools will fail with OOM.

Quoting @jacarey's issue 
> My issue was in a `readHeader`. BufferedLineReader fromString reads into a byte array of `len * max` bytes per char for UTF8, which in my case was greater than `2^31 -1`.

```bash
Exception in thread "main" java.lang.OutOfMemoryError: Requested array size exceeds VM limit
        at java.lang.StringCoding$StringEncoder.encode(StringCoding.java:300)
        at java.lang.StringCoding.encode(StringCoding.java:344)
        at java.lang.StringCoding.encode(StringCoding.java:387)
        at java.lang.String.getBytes(String.java:958)
        at htsjdk.samtools.util.BufferedLineReader.fromString(BufferedLineReader.java:57)
        at htsjdk.samtools.BAMFileReader.readHeader(BAMFileReader.java:667)
```

### DIFF
--- a/src/main/java/htsjdk/samtools/util/BufferedLineReader.java
+++ b/src/main/java/htsjdk/samtools/util/BufferedLineReader.java
@@ -31,7 +31,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.LineNumberReader;
 import java.nio.charset.Charset;
-import java.nio.charset.CharsetEncoder; 
+import java.nio.charset. StandardCharsets; 
 
 /**
  * Implementation of LineReader that is a thin wrapper around BufferedReader.  On Linux, this is faster
--- a/src/main/java/htsjdk/samtools/util/BufferedLineReader.java
+++ b/src/main/java/htsjdk/samtools/util/BufferedLineReader.java
@@ -42,7 +42,7 @@ import java.nio.charset.CharsetEncoder;
  */
 public class BufferedLineReader extends LineNumberReader implements LineReader {
     
-    private static float MAX_BYTES_PER_CHAR_UTF8 = Charset.forName("UTF8").newEncoder().maxBytesPerChar();
+    private static final float MAX_BYTES_PER_CHAR_UTF8 = StandardCharsets.UTF_8.newEncoder().maxBytesPerChar();
 
     private static class StringBackedInputStream extends InputStream {
         private int idx = 0;
--- a/src/main/java/htsjdk/samtools/util/BufferedLineReader.java
+++ b/src/main/java/htsjdk/samtools/util/BufferedLineReader.java
@@ -41,6 +41,25 @@ import java.nio.charset.Charset;
  */
 public class BufferedLineReader extends LineNumberReader implements LineReader {
 
+    private static class StringBackedInputStream extends InputStream {
+        private int idx = 0;
+        private final String str;
+        private final int len;
+
+        StringBackedInputStream(String str) {
+            this.str = str;
+            this.len = str.length();
+        }
+
+        @Override
+        public int read() throws IOException {
+            if(idx >= len) {
+                return -1;
+            }
+            return (byte) str.charAt(idx++);
+        }
+    }
+
     public BufferedLineReader(final InputStream is) {
         this(is, Defaults.NON_ZERO_BUFFER_SIZE);
     }
@@ -54,7 +73,14 @@ public class BufferedLineReader extends LineNumberReader implements LineReader {
      * is necessary because the String is in unicode.
      */
     public static BufferedLineReader fromString(final String s) {
-        return new BufferedLineReader(new ByteArrayInputStream(s.getBytes()));
+        final InputStream is;
+        if (s.length() >= Integer.MAX_VALUE - 8) {
+           is = new StringBackedInputStream(s);
+        }
+        else {
+            is = new ByteArrayInputStream(s.getBytes());
+        }
+        return new BufferedLineReader(is);
     }
 
     /**
--- a/src/main/java/htsjdk/samtools/util/BufferedLineReader.java
+++ b/src/main/java/htsjdk/samtools/util/BufferedLineReader.java
@@ -31,7 +31,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.LineNumberReader;
 import java.nio.charset.Charset;
-import java.nio.charset. StandardCharsets; 
+import java.nio.charset.StandardCharsets; 
 
 /**
  * Implementation of LineReader that is a thin wrapper around BufferedReader.  On Linux, this is faster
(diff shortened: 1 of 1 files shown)


src/main/java/htsjdk/samtools/util/BufferedLineReader.java
```diff
@@ -41,6 +41,25 @@
...
+            this.str = str;
+            this.len = str.length();
+        }
+
+        @Override
+        public int read() throws IOException {
+            if(idx >= len) {
+                return -1;
+            }
+            return (byte) str.charAt(idx++);
```
lbergelson, line 59: I might be confused about the results of charAt, but won't this mangle non-ascii characters?  We already do mangle non-ascii in a lot of places, but I would prefer we don't add more instances of it.  SAM headers can contain UTF-8 in a number of places.

lbergelson: @nh13 Any comment on this?

nh13: It should work with UTF-8. 

```scala
scala> val string =  "ﾖ"
string: String = ﾖ

scala> string.charAt(0)
res0: Char = ﾖ

scala> string.length
res1: Int = 1
```

lbergelson: I meant that casting from char -> byte would cause problems with utf-8 since we'd be losing the second byte of information.

lbergelson: I was unclear was unclear in my writing.

nh13: True.  Any suggestions on how to make it work.  PRs welcome as I'm about to go on vacation.

```scala
scala> 'ﾖ'.isValidByte
res0: Boolean = false

scala> 'ﾖ'.isValidInt
res1: Boolean = true
```

nh13, line 59:
```suggestion
            return (int) str.charAt(idx++);
```

nh13: Changing `byte` to `int`

src/main/java/htsjdk/samtools/util/BufferedLineReader.java
```diff
@@ -41,6 +41,25 @@
...
+            this.str = str;
+            this.len = str.length();
+        }
+
+        @Override
+        public int read() throws IOException {
+            if(idx >= len) {
+                return -1;
+            }
+            return (int) str.charAt(idx++);
```
nh13, line 62: @lbergelson I think solves the previous issue, where we cast as `byte` instead of `int`

src/main/java/htsjdk/samtools/util/BufferedLineReader.java
```diff
@@ -54,7 +73,14 @@ public BufferedLineReader(final InputStream is, final int bufferSize) {
      * is necessary because the String is in unicode.
      */
     public static BufferedLineReader fromString(final String s) {
-        return new BufferedLineReader(new ByteArrayInputStream(s.getBytes()));
+        final InputStream is;
+        if (s.length() >= Integer.MAX_VALUE - 8) {
```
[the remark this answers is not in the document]
nh13: @pshapiro4broad I think @jacarey is correct, so I have updated based on his comments.  Can you take another ook?

src/main/java/htsjdk/samtools/util/BufferedLineReader.java
```diff
@@ -54,7 +76,18 @@ public BufferedLineReader(final InputStream is, final int bufferSize) {
      * is necessary because the String is in unicode.
      */
     public static BufferedLineReader fromString(final String s) {
-        return new BufferedLineReader(new ByteArrayInputStream(s.getBytes()));
+        final InputStream is;
+        // Developer Note: if the string is long enough such that the # of bytes needed exceeds the
+        // maximum array size, we need to use a custom string-backed input stream, versus a byte array
+        // backed input stream.  The # of bytes needed is the length of the string times the number
+        // bytes per character to store in UTF8.
+        if (s.length() * MAX_BYTES_PER_CHAR_UTF8 >= Integer.MAX_VALUE) {
```
nh13, line 84: @pshapiro4broad I think this solves the issue, and I added comments to describe what we're doing here. Can you look one more time?

nh13:
@lbergelson and @pshapiro4broad my apologies for this dropping off, but I've updated based on folks' review, can you take another look?

src/main/java/htsjdk/samtools/util/BufferedLineReader.java
```diff
@@ -40,6 +41,27 @@
  * @author alecw@broadinstitute.org
  */
 public class BufferedLineReader extends LineNumberReader implements LineReader {
+    
+    private static float MAX_BYTES_PER_CHAR_UTF8 = Charset.forName("UTF8").newEncoder().maxBytesPerChar();
```
pshapiro4broad, line 45: Add `final`, use `StandardCharsets.UTF_8`:
```suggestion
    private static final float MAX_BYTES_PER_CHAR_UTF8 = StandardCharsets.UTF_8.newEncoder().maxBytesPerChar();
```

pshapiro4broad (approved):
This looks good to me. It gives you a bit more room than before, but it seems like we could run into a case where we can't hold it in memory. At that point the code that uses this will need to stream the data instead of holding it in memory.

src/main/java/htsjdk/samtools/util/BufferedLineReader.java
```diff
@@ -31,6 +31,7 @@
 import java.io.InputStreamReader;
 import java.io.LineNumberReader;
 import java.nio.charset.Charset;
+import java.nio.charset.CharsetEncoder;
```
nh13, line 34:
```suggestion
import java.nio.charset. StandardCharsets; 
```

src/main/java/htsjdk/samtools/util/BufferedLineReader.java
```diff
@@ -31,6 +31,7 @@
 import java.io.InputStreamReader;
 import java.io.LineNumberReader;
 import java.nio.charset.Charset;
+import java.nio.charset. StandardCharsets;
```
nh13, line 34:
```suggestion
import java.nio.charset.StandardCharsets; 
```

nh13:
@lbergelson can you merge?

nh13:
@pshapiro4broad or @lbergelson I think this is ready for merging.  Any objection to merging?

PR #1505, opened 2020-09-01
Change try-finally to try-with-resources in MetricsFile.write

lbergelson:
### Description

There's a weird failure state in CrosscheckFingerPrints where it ends without an exception but writes no metrics.  One possible explanation is that the order of exception suppression is suppressing the causal exception and then we're discarding the new exception.

This cleans up the code and changes the order of suppression to match what we want.


lbergelson:
@pshapiro4broad I'm also kind of skeptical, but it's the only plausible scenario we've come up with so far.

Issue #1503, opened 2020-08-23
Bug in .gitignore?


dariober:
The `.gitignore` in commit e803eea contains the line:

```
build/
```

I think this makes the directories `src/test/java/htsjdk/samtools/cram/build/` and `src/main/java/htsjdk/samtools/cram/build/` to be erroneously ignored. Shouldn't it be something like `build/*` instead?

----

Steps to reproduce:

```
git clone https://github.com/samtools/htsjdk
cd htsjdk
```

New file `src/main/java/htsjdk/samtools/cram/build/foo.txt` is ignored:

```
touch src/main/java/htsjdk/samtools/cram/build/foo.txt

git status
On branch master
Your branch is up-to-date with 'origin/master'.

nothing to commit, working tree clean
```

Edit `.gitignore` and try again, now `foo.txt` is picked up:

```
sed -i 's|build/|build/*|' .gitignore 

git status

On branch master
Your branch is up-to-date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   .gitignore

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	src/main/java/htsjdk/samtools/cram/build/foo.txt

no changes added to commit (use "git add" and/or "git commit -a")
```

This is with git 2.17.1

PR #1509, opened 2020-09-02
Updating .gitignore

lbergelson:
* Limit ignored directories to only the ones in the root folder
* This fixes https://github.com/samtools/htsjdk/issues/1503
* Adding additional ignores for files jenerated by jenv and japi-compliance-checker

@dariober This should fix your issue.
@jmarshall I used your suggested fix, want to review?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@ htsjdk.iws
 .command_tmp
 atlassian-ide-plugin.xml
 /htsjdk.version.properties
-/test-output
+/test-output/
 
 #intellij
 .idea/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -2,19 +2,22 @@ htsjdk.iws
...
 src/htsjdk.iml
 *.iml
 *.ipr
 *.iws
-out/
-
+/out/
 
+#gradle stuff
+/.gradle/
```
lbergelson, line 16: Maybe we want to ignore .gradle folders if they appear anywhere in the subtree?  It's unclear...

jmarshall (approved):
LGTM in terms of adding the leading `/` to directories at the top-level. I have no wisdom to share as to where _.gradle_ directories might appear.

Issue #1510, opened 2020-09-14
VCF output with invalid QUAL values based on Locale


dennishendriksen:
Consider htsjdk.variant.vcf.VCFEncoder:

```java
private static final String QUAL_FORMAT_STRING = "%.2f";
private static final String QUAL_FORMAT_EXTENSION_TO_TRIM = ".00";

private static String formatQualValue(final double qual) {
        String s = String.format(QUAL_FORMAT_STRING, qual);
        if (s.endsWith(QUAL_FORMAT_EXTENSION_TO_TRIM)) {
            s = s.substring(0, s.length() - QUAL_FORMAT_EXTENSION_TO_TRIM.length());
        }
        return s;
}
```
This writes invalid QUAL values to VCF for some locales such as 'en_NL':
```java
  public static void main(String[] args) {
    System.out.println(String.format("%.2f", 29.00));
  }
```

```
en_NL
29,00
```

```
en_US
29.00
```

An easy fix would be to check for suffix ',00' as well as '.00'. A proper fix should not do any string manipulation.

lbergelson:
Ah, lazy local handling strikes again.  I think if we hardcode it to use Locale.US it should be good, but there are probably issues like this all over the place.  I might recommend.  having your code explicitly set the local at the start of processing in order to work around this.   (GATK does `Locale.setDefault(Locale.US);` at the very start of it's main method to prevent these problems when running).  That may obviously be impractical if your other code relies on the local locale though.  

I'll patch this instance and try to scan for other places this happens.

tfenne:
Fixed by #1514. VCFEncoder now formats QUAL (and other machine-readable numeric output) with Locale.US, so non-US locales no longer produce invalid QUAL values. Closing as resolved.

PR #1514, opened 2020-09-28
Adding Locales to String.format calls

lbergelson:
*Setting the local to US in String.format calls that are used to generate computer readable outputs.  Error messages are still written with the default local. 
*Fix for https://github.com/samtools/htsjdk/issues/1510


pshapiro4broad (approved):
- How did you find all the cases? I'm guessing manual inspection but I was wondering if there was a better way.

- Can you add a test? It looks like the JVM lets you set the default locale, so in theory you could change the default locale inside a test to force an error case.

lbergelson:
You're right, this needs a test.  I just checked manually.  It's probably not perfect but it's better than nothing. I could automatically insert locals in ALL calls but it seemed like extra verbosity in most cases with no benefit.

PR #1517, opened 2020-10-16
Fix race condition in SamRecordDuplicateComparator

fleharty:
### Description

The getLibraryId function of SAMRecordDuplicateComparator.java can create race conditions when run multithreaded.

- [x] Check your code style.

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
+++ b/src/main/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
@@ -26,6 +26,8 @@ package htsjdk.samtools;
 import htsjdk.samtools.DuplicateScoringStrategy.ScoringStrategy;
 
 import java.io.Serializable;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.SortedSet;
@@ -63,24 +65,24 @@ public class SAMRecordDuplicateComparator implements SAMRecordComparator, Serial
 
     @Deprecated // This results in sort order depending on the order in which the headers are listed.  Will be removed in future version.
     public SAMRecordDuplicateComparator(final List<SAMFileHeader> headers) {
-        for (final SAMFileHeader header : headers) {
-            populateLibraryIds(header);
-        }
+        populateLibraryIds(headers);
     }
 
     public SAMRecordDuplicateComparator(final SAMFileHeader header) {
-        populateLibraryIds(header);
+        populateLibraryIds(Collections.singletonList(header));
     }
 
-    private void populateLibraryIds(final SAMFileHeader header) {
+    private void populateLibraryIds(final List<SAMFileHeader> headers) {
         final SortedSet<String> libraryNameOrder = new TreeSet<>();
         libraryNameOrder.add(UNKNOWN_LIBRARY_STRING);
 
         // Determine order of library names
-        for (final SAMReadGroupRecord readGroup : header.getReadGroups()) {
-            final String libraryName = readGroup.getLibrary();
-            if (null != libraryName) {
-                libraryNameOrder.add(libraryName);
+        for (final SAMFileHeader header : headers) {
+            for (final SAMReadGroupRecord readGroup : header.getReadGroups()) {
+                final String libraryName = readGroup.getLibrary();
+                if (null != libraryName) {
+                    libraryNameOrder.add(libraryName);
+                }
             }
         }
 
@@ -131,7 +133,7 @@ public class SAMRecordDuplicateComparator implements SAMRecordComparator, Serial
     /** Get the library ID for the given SAM record. */
     private short getLibraryId(final SAMRecord rec) {
         if (this.libraryIds.size() == 0) {
-            populateLibraryIds(rec.getHeader());
+            populateLibraryIds(Arrays.asList(rec.getHeader()));
         }
         final String library = getLibraryName(rec);
         return updateLibraryId(library);
--- a/src/main/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
+++ b/src/main/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
@@ -99,7 +99,7 @@ public class SAMRecordDuplicateComparator implements SAMRecordComparator, Serial
     /**
      * Populates the set of transient attributes on SAMRecords if they are not already there.
      */
-    private void populateTransientAttributes(final SAMRecord... recs) {
+    synchronized private void populateTransientAttributes(final SAMRecord... recs) {
         for (final SAMRecord rec : recs) {
             if (rec.getTransientAttribute(Attr.LibraryId) != null) continue;
             rec.setTransientAttribute(Attr.LibraryId, getLibraryId(rec));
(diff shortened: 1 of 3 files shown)


fleharty:
@lbergelson This resolves the race condition thing we discussed the other day.  Would you be able to review this?

src/main/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
```diff
@@ -57,27 +60,44 @@
     
     public SAMRecordDuplicateComparator() {}
 
+    @Deprecated // not thread safe. will be removed in future version
```
yfarjoun, line 63: The problem isn't that it's not thread-safe (I think), it's that the eventual order depends on the order of the headers in the list.....right?

src/main/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
```diff
@@ -57,27 +60,44 @@
...
     public SAMRecordDuplicateComparator(final List<SAMFileHeader> headers) {
-        // pre-populate the library names
-        for (final SAMFileHeader header : headers) {
-            for (final SAMReadGroupRecord readGroup : header.getReadGroups()) {
-                final String libraryName = readGroup.getLibrary();
-                if (null != libraryName) {
-                    final short libraryId = this.nextLibraryId++;
-                    this.libraryIds.put(libraryName, libraryId);
-                }
+        for (SAMFileHeader header : headers) {
```
yfarjoun, line 65: final

yfarjoun: I'm worried now about the interaction between the UNKNOWN_LIBRARY and this for loop, since at each iteration the UNKNOWN_LIBRARY will be added to the libraryIds Map. see idea below.

src/main/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
```diff
@@ -57,27 +60,44 @@
...
+            final String libraryName = readGroup.getLibrary();
+            if (null != libraryName) {
+                libraryNameOrder.add(libraryName);
             }
         }
+
+        // Pre-populate library names
+        for (final String name : libraryNameOrder) {
+            final short libraryId = this.nextLibraryId++;
+            this.libraryIds.put(name, libraryId);
```
yfarjoun, line 89: can this be done using `computeIfAbsent`?

src/test/java/htsjdk/samtools/SAMRecordDuplicateComparatorTest.java
```diff
@@ -24,14 +24,22 @@
 package htsjdk.samtools;
 
 import htsjdk.HtsjdkTest;
+import htsjdk.samtools.util.CloserUtil;
+import htsjdk.samtools.util.Histogram;
+import htsjdk.samtools.util.ProgressLogger;
```
yfarjoun, line 29: do you need all these imports here?

src/main/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
```diff
@@ -105,19 +125,27 @@ private static String getLibraryName(final SAMRecord rec) {
...
+        return updateLibraryId(library);
+    }
+
+    /** Update library ID in thread-safe manner. */
+    private synchronized short updateLibraryId(final String library) {
+
         Short libraryId = this.libraryIds.get(library);
 
         if (libraryId == null) {
             libraryId = this.nextLibraryId++;
```
yfarjoun, line 146: here too, can this be done with `computeIfAbsent`?

yfarjoun (changes requested):
this is indeed nasty race condition. I think that there's still some funny business going on with the deprecated constructor, and also, I suspect that you can use `computeIfAbsent` and switch to a `ConcurrentHashMap` and remove the `synchronized` tag from the methods`.

src/main/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
fleharty: I'm not sure I entirely understand what you are looking for here.  The libraryNameOrder is a set, so if it gets added multiple times, it doesn't have any impact on the set.

yfarjoun (approved):
LGTM @lbergelson want to take a glance to see if I may have missed something?

src/main/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
```diff
@@ -105,20 +125,20 @@ private static String getLibraryName(final SAMRecord rec) {
             }
         }
 
-        return "Unknown Library";
+        return UNKNOWN_LIBRARY_STRING;
     }
 
     /** Get the library ID for the given SAM record. */
     private short getLibraryId(final SAMRecord rec) {
```
lbergelson, line 132: I think there are still race conditions here in some (potentially rare) conditions.  

If the records have different sam headers ( and you're not pre-initializing them with the now deprecated constructor)  than the id is dependent on the order that getLibraryID is called in which is a race condition.  Maybe we don't care about that case since we don't expect to have different headers given that you've deprecated that constructor.  

What happened to syncronizing populateTransientAttributes?  

Using a ConcurrentHashMap.calculateIfMissing is a good idea.

lbergelson (commented):
@fleharty Some questions...  One other option would be to get rid of libraryId entirely and just use the library name in comparisons directly.  That would avoid all of the nasty state around the libraryId.  It might have a performance impact given that it's changing from a short comparison to a string comparison, but there's a good chance the impact is negligible.

fleharty:
@lbergelson 
I'm unsure about populateTransientAttributes.  I can't get it to fail, so as far as I know it is okay.  Horror of horrors would be that we do have a sneaky race condition that occurs with very low probability.  I'm more than willing to synchronize this method if you think it is necessary, but I have no way of testing that change helps anything.

We depricated SAMRecordDuplicateComparator(final List<SAMFileHeader> headers) because it isn't used anywhere internally to HTSJDK, and I'm unaware of any external tools using this constructor.

I don't have permissions to merge this PR.  Could you do that for me?

fleharty:
@lbergelson Back to you.

lbergelson (approved):

src/main/java/htsjdk/samtools/SAMRecordDuplicateComparator.java
```diff
@@ -50,26 +53,43 @@
...
-    private final Map<String, Short> libraryIds = new HashMap<String, Short>(); // from library string to library id
+    // We use a ConcurrentHashMap to ensure thread safety
+    private final Map<String, Short> libraryIds = new ConcurrentHashMap<>(); // from library string to library id
     private short nextLibraryId = 1;
     
     private ScoringStrategy scoringStrategy = ScoringStrategy.TOTAL_MAPPED_REFERENCE_LENGTH;
     
     public SAMRecordDuplicateComparator() {}
 
+    @Deprecated // This results in sort order depending on the order in which the headers are listed.  Will be removed in future version.
```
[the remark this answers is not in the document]
lbergelson: I just approved, but then I realized this comment is out of date.  Why not undeprecate it?

PR #1520, opened 2020-12-11
Check base qualities before accumulating inserted bases in SamLocusIterator

mollysacks:
### Description

Updated `accumulateIndels` in `SamLocusIterator` to check if bases in an insertion have base quality >= `qualityScoreCutoff` before adding them to the accumulator. The motivation for this change is that insertions should be filtered and accumulated consistently with matching bases.

- [ ] Check your code style.

### DIFF
--- a/src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
+++ b/src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
@@ -301,38 +301,33 @@ public class SamLocusIteratorTest extends AbstractLocusIteratorTestTemplate {
             builder.addFrag("record" + i, 0, startPosition, true, false, "10M3I17M3I3M", qualityString, 10);
         }
         final int insStart1 = 174;
-        final int insStart2 = 194;
-        // test both for include indels and do not include indels
-        for (final boolean incIndels : new boolean[] {false, true}) {
-            final SamLocusIterator sli = createSamLocusIterator(builder);
-            sli.setIncludeIndels(incIndels);
-            sli.setQualityScoreCutoff(10);
-            // make sure we accumulated depth for each position
-            int pos = startPosition;
-            for (final SamLocusIterator.LocusInfo li : sli) {
-                Assert.assertEquals(li.getPosition(), pos++);
-                // make sure we are accumulating normal coverage
-                Assert.assertEquals(li.getRecordAndOffsets().size(), coverage);
-                // Check the correct assignment of the alignment type
-                for (final SamLocusIterator.RecordAndOffset rao : li.getRecordAndOffsets()) {
-                    Assert.assertEquals(rao.getAlignmentType(), SamLocusIterator.RecordAndOffset.AlignmentType.Match);
-                }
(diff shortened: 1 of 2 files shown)


src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -287,6 +287,57 @@ public void testSimpleInsertion() {
...
+                Assert.assertEquals(li.getDeletedInRecord().size(), 0);
+                if (incIndels && li.getPosition() == insStart1) {
+                    Assert.assertEquals(li.getInsertedInRecord().size(), coverage);
+                    // Check the correct assignment of the alignment type
+                    for (final SamLocusIterator.RecordAndOffset rao : li.getInsertedInRecord()) {
+                        Assert.assertEquals(rao.getAlignmentType(), SamLocusIterator.RecordAndOffset.AlignmentType.Insertion);
+                    }
+                } else if (incIndels && li.getPosition() == insStart2){
+                    // Second insertion should not be included because base quality is less than qualityScoreCutoff
+                    Assert.assertEquals(li.getInsertedInRecord().size(), 0);
```
lbergelson, line 332: It looks like the final two branches do the same thing.  Were they meant to assert different things?

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -287,6 +287,57 @@ public void testSimpleInsertion() {
...
+        final int startPosition = 165;
+        String qualityString = "++++++++++,,,+++++++++++++++++***+++";
+        for (int i = 0; i < coverage; i++) {
+            // add a negative-strand fragment mapped on chrM with base quality of 10
+            builder.addFrag("record" + i, 0, startPosition, true, false, "10M3I17M3I3M", qualityString, 10);
+        }
+        final int insStart1 = 174;
+        final int insStart2 = 194;
+        // test both for include indels and do not include indels
+        for (final boolean incIndels : new boolean[] {false, true}) {
```
lbergelson, line 306: If you have two different test cases it's usually better to make a data provider instead of iterating over them within the test.

lbergelson (changes requested):
@mollysacks Looks good.  Minor comments about the tests.

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
mollysacks: Oops, you're right! The expected behavior is the same for both branches so I will get rid of the else if

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
mollysacks: I think I'm going to just not include incIndels = false, because there is another test that tests this case and my change doesn't affect the behavior of the code when incIndels = false.

lbergelson (approved):
Looks good to me.

lbergelson:
@mollysacks Thank you.

PR #1483, opened 2020-05-31
Extend SamLocusAndReferenceIterator to work with CRAMFileReader

jkaczynski1:
### Description
Extend the SamLocusAndReferenceIterator class to work with the CRAMFileReader and other file readers. Currently the SamLocusAndReferenceIterator class depends on the legacy SamReader class. The purpose of the enhancement is to add support for newer file reader classes such as the CRAMFileReader.

### Things to think about before submitting:
- [yes] Make sure your changes compile and new tests pass locally.
- [yes] Add new tests or update existing ones:
  - A bug fix should include a test that previously would have failed and passes now.
  - New features should come with new tests that exercise and validate the new functionality.
- [n/a] Extended the README / documentation, if necessary
- [yes] Check your code style.
- [yes] Write a clear commit title and message
  - The commit message should describe what changed and is targeted at htsjdk developers
  - Breaking changes should be mentioned in the commit message.

PR #1521, opened 2020-12-17
- Expand SamLocusAndReferenceIterator tests to cover BAMs and CRAMs.

yfarjoun:
(Sorry about the premature push into master earlier)

There seems to be a lack of testing of SamLocusAndReferenceIterator on cram and bam files, which lead one of our contributors to submit #1483 which seems to be solving a problem that doesn't exist. I hope that this PR will provide an example of how to properly create a SamLocusAndReferenceIterator in a way that is agnostic of the type of file (sam/bam/cram) at hand.


src/main/java/htsjdk/variant/vcf/SamFileUtils.java
```diff
@@ -0,0 +1,63 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package htsjdk.variant.vcf;
```
lbergelson, line 25: Looks like this is the wrong package for something called "SamFileUtils"

src/main/java/htsjdk/variant/vcf/SamFileUtils.java
```diff
@@ -0,0 +1,63 @@
...
+     * @param prefix - The prefix string to be used in generating the file's name; must be at least three characters long
+     * @param suffix - The suffix string to be used in generating the file's name; may be null, in which case the suffix ".tmp" will be used
+     * @return A File object referencing the newly created temporary file
+     * @throws IOException - if a file could not be created.
+     */
+    public static File createTemporaryIndexedFile(final String prefix, final String suffix) throws IOException {
+        final File out = File.createTempFile(prefix, suffix);
+        out.deleteOnExit();
+        String indexFileExtension = null;
+        if (suffix.endsWith("vcf.gz")) {
```
lbergelson, line 47: these should probably reference the constants in FileExtensions

lbergelson: This would probably be better placed in something like htsjdk.samtools.util.TestUtil instead of a new class

src/main/java/htsjdk/variant/vcf/SamFileUtils.java
lbergelson: It would probably be better to put this method in htsjdk.samtools.util.TestUtil instead of a new class.

src/test/java/htsjdk/samtools/reference/SamLocusAndReferenceIteratorTest.java
```diff
@@ -1,26 +1,87 @@
...
+    @DataProvider
+    Object[][] testSamLocusAndReferenceIteratorData() {
+        return new Object[][]{
+                {simpleSmallFileSam, false},
+                {simpleSmallFileBam, false},
+                {simpleSmallFileCram, true}
+        };
+    }
+
+    @BeforeSuite
```
lbergelson, line 46: Could we change this from `@BeforeSuite` to `@BeforeClass` ?  That will limit the chance of confusing things happening when this setup method is unexpectedly run before unrelated tests. (https://testng.org/doc/documentation-main.html#annotations)

lbergelson (changes requested):
@yfarjoun Thanks for doing this.  I have a few comments then you're good to merge.

src/main/java/htsjdk/samtools/util/FileExtensions.java
```diff
@@ -53,7 +53,12 @@
     /** extensions for alignment files SAM, BAM, CRAM. */
     public static final String SAM = ".sam";
     public static final String BAM = ".bam";
-    public static final String BAI_INDEX = ".bai";
+    public static final String BAM_INDEX = ".bai";
```
lbergelson, line 56: I think this was deliberately called the BAI index in able to differentiate it from BAM CSI index.

src/main/java/htsjdk/samtools/util/TestUtil.java
```diff
@@ -89,6 +90,35 @@ public static File getTempDirecory(final String prefix, final String suffix) {
...
+        } else if (suffix.endsWith(FileExtensions.VCF)) {
+            indexFileExtension = FileExtensions.VCF_INDEX;
+        } else if (suffix.endsWith(FileExtensions.BAM)) {
+            indexFileExtension = FileExtensions.BAM_INDEX;
+        } else if (suffix.endsWith(FileExtensions.CRAM)) {
+            indexFileExtension = FileExtensions.CRAM_INDEX;
+        }
+
+        if (indexFileExtension != null) {
+            final File indexOut = new File(out.getAbsolutePath() + indexFileExtension);
```
lbergelson, line 116: Oh, I didn't think about this, but don't we need to be careful about extension vs replacement here?  Don't some of these follow replacement rules instead of additive rules.  I.e. bam.bai is ok, but it's just .crai instead of .cram.crai?

src/main/java/htsjdk/samtools/util/FileExtensions.java
yfarjoun: hmmm. good point. I guess the name should be `BAM_BAI_INDEX` and `BAM_CSI_INDEX`

src/main/java/htsjdk/samtools/util/TestUtil.java
yfarjoun: I guess we need objects that generate index_names from file names...

lbergelson: We have them, they're just not centralized.

yfarjoun: where?

yfarjoun: they are impossible to find!

yfarjoun:
back to you @lbergelson

tfenne:
This PR combines some nice to have marginal test coverage, with breaking changes in files that have changed very significantly since this PR was first drafted.  I do not think it is worth the effort to rebase and clean up, so closing.

PR #1524, opened 2020-12-22
Feature/async pool

sstadick:
### Description

This PR adds a Pooled Async Writer aimed at the use case where there are significantly more writers than there are threads. 

- [X] Check your code style.


src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+    @Override
+    public void close() throws IOException {
+        this.poolClosed.set(true);
+        for (PooledWriter<A> writer : this.writers) {
+            writer.close();
+        }
+        this.executor.shutdown();
+    }
+
+    public static class PooledWriter<B> implements WrappedWriter<B> {
```
tfenne, line 33: Calling the interface `WrapperWriter` is super confusing to me.  I think it should just be `Writer` since many implementation will not wrap anything.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+        private final WrappedWriter<B> writer;
+        private final BlockingQueue<B> queue;
+        private final AtomicBoolean writing;
+        private final AtomicBoolean isClosed;
+        public final AsyncWriterPool<B> outer;
```
tfenne, line 38: You shouldn't need this.  Instead make the class be `public class PooledWriter` without the `static` keyword.  Then you can access the outer with `AsyncWriterPool.this`.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+
+        private final int buffsize;
```
tfenne, line 40: You can't afford an `er`?  `bufferSize` please.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+        private Future<Integer> currentTask;
```
tfenne, line 41: This likely needs a bit more explanation.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+            this.writer = writer;
+            this.queue = queue;
+            this.buffsize = buffSize;
+            this.outer = outer;
+            this.writing = new AtomicBoolean(false);
+            this.isClosed = new AtomicBoolean(false);
+            this.currentTask = null;
+
+            // Add self to writer
+            this.outer.writers.add(this);
```
tfenne, line 55: I would do this in the outer class directly, it feels odd here.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        /**
+         * Will throw any error thrown in task
+         */
+        private void checkAndRethrow() {
+            if (!(this.currentTask == null)) {
```
tfenne, line 62:
```suggestion
            if (this.currentTask != null) {
```

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+                try {
+                    // NB: ignoring output of number of records previously written
+                    this.currentTask.get();
+                } catch (CancellationException | InterruptedException | ExecutionException e) {
+                    this.isClosed.set(true);
```
tfenne, line 67: Why do this here?  We don't know, but maybe it's recoverable?

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+                    throw new RuntimeException(e);
```
tfenne, line 68: It's generally nice to add some context when re-throwing an exception.  E.g. `exception while writing records asynchronously`

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+                }
+            }
+        }
+
+        @Override
+        public void write(final B item) {
+            if (this.isClosed.get()) throw new RuntimeIOException("Attempt to add record to closed writer.");
```
tfenne, line 75: Does `isClosed` need to be an atomic boolean?  I think not.  I.e. if we make it clear that `PooledWriter`s are _not_ inherently thread safe and you should synchronize if multiple threads want to call `write()` or `close()` then this can be a simple `boolean` and will perform better.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+
+            // Put new item in queue
+            try {
+                this.queue.put(item);
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
```
tfenne, line 81: Message on chaining.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+            }
+
+            // If queue size is large enough, and not currently writing items, send to executor
+            if (queue.size() >= this.buffsize && !this.writing.getAndSet(true)) {
+
+                checkAndRethrow();
```
tfenne, line 87: Why is this here?  Likely it should be the first line in `write()`.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+
+            // Put new item in queue
+            try {
+                this.queue.put(item);
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+
+            // If queue size is large enough, and not currently writing items, send to executor
+            if (queue.size() >= this.buffsize && !this.writing.getAndSet(true)) {
```
tfenne, line 85: Are we sure that `queue.size()` is constant time for the implementation(s) we're using?  If so great, if not we should likely keep our own counter.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+
+                checkAndRethrow();
+                this.drain(this.buffsize);
```
tfenne, line 88: Why have it only drain `buffSize`?  Why not have it drain the queue until it's empty?

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        private void drain(int toDrain) {
+            // check the result of the previous task
+            this.currentTask = outer.executor.submit(() -> {
+                int counter = 0;
+
+                try {
+                    // BlockingQueue is threadsafe and uses a different lock for both reading and writing
+                    // ends of the queue, so this is safe and avoids an extra coffee.
```
tfenne, line 99: coffee?

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+                    while (counter < toDrain && !this.queue.isEmpty()) {
+                        this.writer.write(this.queue.poll());
+                        counter++;
```
tfenne, line 102:
```suggestion
                    B item = null
                    while ((item = this.queue.poll()) != null) {
                        this.writer.write(item);
                        counter++;
```

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+                    }
+                } finally {
+                    this.writing.set(false);
+                }
+                return counter;
```
tfenne, line 107: Is there any need to keep count of how many things are written?

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
tfenne: Alternatively I _think_ you could make this more efficient if you did the following (but honestly I'm not 100% sure it would be faster, it all depends on how much the lock acquisition for poll() is):
1. Within each writer define a `private List<B> buffer = new ArrayList<B>(bufferSize)`
2. In the drain method you can then use `this.queue.drainTo(this.buffer, this.bufferSize)` to grab a whole group of items in one operation
3. Then you can write those items to the output and clear the buffer

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+            });
+        }
+
+
+        @Override
+        public void close() throws IOException {
+            // NB: We don't need to check on writing here since checkAndRethrow will block till writing is done
```
tfenne, line 114: Yes, basically any method that is public should have `checkAndRethrow` as it's first line.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,124 @@
...
+            this.checkAndRethrow();
+            if (!this.isClosed.getAndSet(true)) {
+                this.drain(this.queue.size());
+                this.checkAndRethrow(); // Wait for last write to finish
+                this.writer.close();
+            }
+        }
+    }
+
```
tfenne, line 123: Bonus newline.

src/main/java/htsjdk/samtools/util/WrappedWriter.java
```diff
@@ -0,0 +1,26 @@
+package htsjdk.samtools.util;
+
+import java.io.Closeable;
+
+
+public interface WrappedWriter<A> extends Closeable {
```
tfenne, line 6: Can we just call this `Writer`?

src/main/java/htsjdk/samtools/util/WrappedWriter.java
```diff
@@ -0,0 +1,26 @@
...
+    default WrappedWriter<A> append(A item) {
+        this.write(item);
+        return this;
+    }
+
+    /** Writes out one or more items in order and returns a reference to the writer. */
+    default WrappedWriter<A> append(Iterable<A> items) {
+       this.write(items);
+       return this;
+    }
```
tfenne, line 25: I don't think I would include these.  It just raises the question of how are they different than the `write()` methods.

tfenne (changes requested):
DIdn't look at the tests yet, but wanted to get you feedback on the core code first.

lbergelson:
@tfenne I'm glad to see you taking an interest in this!  You had proposed something similar before so I feel like you have thought more about this than I have...

tfenne:
Thanks @lbergelson - in full disclosure @sstadick works at Fulcrum, so I'm not exactly an independent reviewer.  That said, we want this in HTSJDK for a number of reasons, not the least of which is so that we can use it in an upcoming PR to Picard to speed up `IlluminaBasecallsToFastq`.  I'm happy to shephard this through, but let me know if you'd also like to review once it's in a more complete state.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
sstadick: In order for AsyncPoolWriter to not be type based on the writer, the writer has to be instantiated in the inner class. Or at least I can't figure out how to do it any other way.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
sstadick: Preliminary attempts with a reused buffer and `drainTo` don't show much speedup.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
sstadick: Yes! First pass at this and it speeds it up quite a lot.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,126 @@
...
+    private final AtomicBoolean poolClosed;
+    private final List<PooledWriter<?>> writers;
+
+    public AsyncWriterPool(final int threads) {
+        assert threads >= 1; // TODO: throw error
+
+        this.poolClosed = new AtomicBoolean(false);
+        this.writers = new ArrayList<>();
+
+        executor = Executors.newWorkStealingPool(threads);
```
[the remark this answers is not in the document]
nh13: Also, is `newFixedThreadPool` better since the threads are presumably long lived?  Perhaps try it out?

nh13 (commented):
I added a few things you could try to see if performance can be further improved.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,126 @@
...
+            }
+        }
+
+        @Override
+        public void write(final A item) {
+            if (this.isClosed.get()) throw new RuntimeIOException("Attempt to add record to closed writer.");
+
+            // Put new item in queue
+            try {
+                this.queue.put(item);
```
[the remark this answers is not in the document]
sstadick: That is up to the user. That put is on whatever thread a user is using and not on a thread created by us.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,126 @@
...
+
+        private void drain(int toDrain) {
+            // check the result of the previous task
+            this.currentTask = AsyncWriterPool.this.executor.submit(() -> {
+                int counter = 0;
+
+                try {
+                    // BlockingQueue is threadsafe and uses a different lock for both reading and writing
+                    // ends of the queue, so this is safe and avoids an extra copy.
+                    while (counter < toDrain) {
```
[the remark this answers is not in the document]
sstadick: I've tried this a few different ways and using any variation of drainTo both inside our outside of the writing thread is not faster than just polling, at least using the unbounded LinkedBlockingQueue that we've been testing with so far, which has separate locks for pushing and popping. 

I'll keep an eye on this when trying an Array backed queue. The fastest implementation so far uses your suggestion to drain till empty, which is another strike against this.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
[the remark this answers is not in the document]
sstadick: I gave this a shot. The work stealing executor is a little faster, and I think should scale up better given that it doesn't create contention on one queue that all threads are polling on, but each thread gets its own queue.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,126 @@
...
+            if (queue.size() >= this.bufferSize && !this.writing.getAndSet(true)) {
+
+                checkAndRethrow();
+                this.drain(this.queue.size());
+            }
+        }
+
+        private void drain(int toDrain) {
+            // check the result of the previous task
+            this.currentTask = AsyncWriterPool.this.executor.submit(() -> {
```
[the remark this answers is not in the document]
sstadick: So far, running with an unbounded queue, this has not been the case, and since the caller can supply the queue type I'm not too worried about it.

It is possible, but given that I'm now draining queues till empty, even if there is a pathological input dataset that is sending to just one writer, we become a pass-through writer on a different thread, writing as fast as it's adding data with just a bit of extra overhead for copying to the intermediate queue. So I think it should be fine.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,126 @@
...
+                    this.currentTask.get();
+                } catch (CancellationException | InterruptedException | ExecutionException e) {
+                    this.isClosed.set(true);
+                    throw new RuntimeException("Exception while writing records asynchronously", e);
+                }
+            }
+        }
+
+        @Override
+        public void write(final A item) {
```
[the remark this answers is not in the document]
nh13: It's probably more efficient to add multiple items into the underlying queues once, as well as draining from the underlying queue (versus offering once).

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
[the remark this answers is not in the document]
nh13: Suppose the call to `put` blocks since the queue is full.  Now a separate thread is draining the queue and encounters an error writing to the underling writer.  What happens?  Does the exception from the latter get appropriately propogated?

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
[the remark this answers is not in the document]
nh13: Perhaps a note to here is appropriate to say this could theoretically happen but hasn't been observed practically.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,245 @@
...
+        /**
+         * Close the enclosed writer, first blocking and waiting for any ongoing writing to complete, draining any
+         * remaining elements from the queue, checking the result for exceptions / blocking till it is complete, and
+         * finally closing the writer itself.
+         *
+         * @throws IOException if the enclosed writer's close raises an exception
+         */
+        @Override
+        public void close() throws IOException {
+            // NB: We don't need to check on writing here since checkAndRethrow will block till writing is done
```
tfenne, line 235: It does, but I think that's bad.  I think what should be done here is that if writing is currently in progress is to generate a `Future` that first caches `this.currentTask` then calls `get()` on it, and then does it's own business, effectively chaining the futures.

tfenne: And to clarify, by _bad_ I mean that if the first writer in the list is currently writing and all the other writers have 0 < n < bufferSize items in queue, all the other writers will wait for the first one to finish draining before they get their close() methods called, which is inefficient.

src/main/java/htsjdk/samtools/util/Writer.java
```diff
@@ -0,0 +1,17 @@
+package htsjdk.samtools.util;
+
+import java.io.Closeable;
+import java.io.Flushable;
+
+
+public interface Writer<A> extends Closeable, Flushable {
```
tfenne, line 7: I'm curious why extend `Flushable`?  Is it really necessary above to call `flush()` after writing a batch out?  What makes us think we know better when to flush() than the underlying writer?

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,245 @@
...
+        private void checkAndRethrow() {
+            if (this.currentTask != null) {
+                try {
+                    // NB: ignoring output of number of records previously written
+                    this.currentTask.get();
+                } catch (CancellationException | InterruptedException | ExecutionException e) {
+                    this.isClosed = true;
+                    throw new RuntimeException("Exception while writing records asynchronously", e);
+                }
+            }
```
tfenne, line 152: I think there should probably be a `finally { this.currentTask = null }` in here so that we're not continuously checking the same task - we only need to check it once per `Future`.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,245 @@
...
+         * @throws IOException if the enclosed writer's close raises an exception
+         */
+        @Override
+        public void close() throws IOException {
+            // NB: We don't need to check on writing here since checkAndRethrow will block till writing is done
+            this.checkAndRethrow();
+            if (!this.isClosed) {
+                this.isClosed = true;
+                this.drain();
+                this.checkAndRethrow(); // Wait for last write to finish
```
tfenne, line 240: Again, don't want to block here because this then causes all writers to close() serially which is undesirable.

tfenne (changes requested):
All the small changes to formatting throughout various files make it much harder to review this PR because it's a mix of formatting changes that are largely irrelevant, and functional changes that need review.  In future please either refrain from re-formatting, or isolate all reformatting to a single commit so it can be reviewed separately.

I'm 99% sure there are race conditions in this code.  Because all the exception propogation is done via `Future`s in `this.currentTask` we have to be 100% sure that `this.currentTask.get()` is invoked at least once per `Future` instance, and I don't think that is guaranteed.  For example I believe that it is possible for the following to happen:
  a. this.drain() to be invoked and a task enqueued
  b. the draining task to complete it's while loop and empty the queue
  c. the queue to fill back up to `bufferSize - 1`
  d. a thread calls write() and gets past the `if (!this.writing.get()) checkAndRethrow();` line
  e. the draining task then flips `this.writing` back to false
  f. the thread in `write()` pushed the queue over `bufferSize` and calls `this.drain()` without ever having invoked `checkAndRethrow()`

I think:
1. checkAndRethrow() should not block
2. When `this.currentTask.get()` is called the `this.currentTask` should be nulled out to avoid repeated checking
3. `checkAndRethrow()` must be called after setting `this.writing` to `true` (i.e. locking for write) and _before_ over-writing `this.currentTask`
4. `close()` should not block either, but instead if writing is in progress it should add a call to `this.currentTask.get()` (caching the appropriate reference) before doing it's own work
5. AsyncWriterPool.close() will then need to call `checkAndRethrow` on all writers after the pool is shutdown.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,126 @@
...
+        private final AtomicBoolean isClosed;
+
+        private final int bufferSize;
+
+        // Holds the Future of the last task submitted to the AsyncWriterPools executor, or null if nothing has been
+        // submitted yet. Inside the Future is the number of items written and any exceptions that may have been thrown.
+        private Future<Void> currentTask;
+
+
+        public PooledWriter(final Writer<A> writer, final BlockingQueue<A> queue, final int bufferSize) {
```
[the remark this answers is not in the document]
nh13: @tfenne Ah I see now what `bufferSize` is for .  Given that I would:
1. check that `bufferSize` is not larger than the queue's capacity.  If it is, then we'll never reach `bufferSize`.
2. As @tfenne suggests, get's rid of the other constructors (sorry I asked earlier) so that setting `bufferSize` is _intentional_.
3. Rephrase the docs.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,245 @@
...
+        // submitted yet. Inside the Future is the number of items written and any exceptions that may have been thrown.
+        private Future<Void> currentTask;
+
+
+        /**
+         * Exchange a class implementing {@link Writer} for a {@code PooledWriter}.
+         *
+         * @param writer     a class implementing {@link Writer}
+         * @param queue      a queue to use for this writer, bounded or unbounded.
+         * @param bufferSize the minimum number of items to write needed before sending to a thread for writing.
```
nh13, line 86: This sentence doesnt make size "to write needed"
```suggestion
         * @param bufferSize the minimum number of items needed before scheduling a thread for writing.
```

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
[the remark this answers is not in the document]
nh13: Agreed, thank @tfenne interpreting my comment with a solution.  Can we move the if/break into the while condition?
```java
while (!this.isClosed && !this.queue.offer(item, 5, TimeUnit.SECONDS)) {
}
```

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,245 @@
...
+
+        /**
+         * Update the minimum number of items required before sending items to a thread for writing.
+         *
+         * @param newBufferSize the new size to use for {@code bufferSize}
+         * @throws IllegalArgumentException if {@code newBufferSize} <= 0
+         */
+        public void setBufferSize(final int newBufferSize) {
+            if (newBufferSize <= 0) throw new IllegalArgumentException("bufferSize must be >= 1");
+            this.bufferSize = newBufferSize;
```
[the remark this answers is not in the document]
sstadick: I've removed this entirely.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,245 @@
...
+                    A item = this.queue.poll();
+                    while (item != null) {
+                        this.writer.write(item);
+                        item = this.queue.poll();
+                    }
+                    this.writer.flush();
+                } finally {
+                    this.writing.set(false);
+                }
+                return null;
```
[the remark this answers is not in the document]
sstadick: Unfortunately yes, in order to use the "nice" Callable task submission instead of going down the runnable route.

src/main/java/htsjdk/samtools/util/Writer.java
sstadick: Fixed, this ended up being an issue in how I had integrated into Picard, and not the Writer. So nothing implements Flushable anymore.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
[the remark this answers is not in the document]
tfenne: Silly java.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,253 @@
...
+     *
+     * @throws IOException if any writer raises an exception
+     */
+    @Override
+    public void close() throws IOException {
+        this.poolClosed.set(true);
+        List<CompletableFuture<Void>> closers = new ArrayList<>();
+        for (final PooledWriter<?> writer : this.writers) {
+            closers.add(writer.asyncClose());
+        }
```
tfenne, line 58: I think this can be replaced with something like:
```java
final List<CompletableFuture<Void>> closers = this.writers.stream().map(w -> w.asyncClose()).toArray(CompletableFuture[]::new);
``` 

Streams can be a little cumbersome in Java, but since you want an array out the far end anyway, this is more concise.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,253 @@
...
+        // Convert List to array and wait on all futures to complete
+        CompletableFuture.allOf(closers.toArray(new CompletableFuture[0])).join();
```
tfenne, line 60: Will this handle things gracefully if one or more writers have already been `closed()` directly?

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,253 @@
...
+     * Any class that implements {@link Writer} can be exchanged for a {@code PooledWriter}. The PooledWriter provides
+     * the same API as {@link Writer}, but will manage buffering of writes and sending to the {@link AsyncWriterPool} it
+     * is attached to.
+     * <p>
+     * Note, {@code PooledWriter} is not intrinsically threadsafe, and any writing / closing done across multiple
+     * threads should be synchronized.
+     *
+     * @param <A> the type of the item the writer can write
+     */
+    public class PooledWriter<A> implements Writer<A> {
```
tfenne, line 90: Can this be `private` now?

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,253 @@
...
+    /**
+     * Exchange a class implementing {@link Writer} for a {@link PooledWriter}.
+     *
+     * @param writer         a class implementing {@link Writer}
+     * @param queue          a queue to use for this writer, bounded or unbounded.
+     * @param writeThreshold the minimum number of items needed before scheduling a thread for writing.
+     * @param <A>            the type of the items the {@link Writer} can write.
+     * @return a {@link PooledWriter}
+     */
+    public <A> PooledWriter<A> pool(final Writer<A> writer, final BlockingQueue<A> queue, final int writeThreshold) {
```
tfenne, line 73: I would leave the type here as `Writer<A>` for the return type.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,253 @@
...
+        private final AtomicBoolean isClosed = new AtomicBoolean(false);
+        private final BlockingQueue<A> queue;
+        private final Writer<A> writer;
+        private final int writeThreshold;
+
+        // Holds the Future of the last task submitted to the AsyncWriterPools until it is checked, then it is null again.
+        private final AtomicReference<Future<Void>> currentTask = new AtomicReference<>(null);
+
+        /**
+         * Exchange a class implementing {@link Writer} for a {@code PooledWriter}.
```
tfenne, line 100: Javadoc needs updating.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,253 @@
...
+     * @param <A> the type of the item the writer can write
+     */
+    public class PooledWriter<A> implements Writer<A> {
+        private final AtomicBoolean isClosed = new AtomicBoolean(false);
+        private final BlockingQueue<A> queue;
+        private final Writer<A> writer;
+        private final int writeThreshold;
+
+        // Holds the Future of the last task submitted to the AsyncWriterPools until it is checked, then it is null again.
+        private final AtomicReference<Future<Void>> currentTask = new AtomicReference<>(null);
```
tfenne, line 97: I'm 99% sure that this doesn't need to be wrapped in an AtomicReference.  Since `PooledWriter` isn't supposed to be thread safe, users of it should either a) be invoking methods from a single thread or b) syncrhonizing externally.  And I _think_ the `currentTask` is only updated in methods invoked by the user, not from within the executor threads.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,253 @@
...
+            this.writeThreshold = writeThreshold;
+        }
+
+        /**
+         * Non-blocking check of the previously completed task. This will check the future if it is not null and if it
+         * isDone. If the task generated any exception, these will be re-thrown.
+         *
+         * @throws RuntimeException if any exception was raised during the writing on the receiving thread
+         */
+        private void asyncCheckAndRethrow() {
```
tfenne, line 123: I would call these `nonBlocking` and `blocking` as your javadoc suggests :)

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,253 @@
...
+         * variants that go by the same name.
+         *
+         * @throws RuntimeException if any exception was raised during the writing on the receiving thread
+         */
+        private void checkAndRethrow() {
+            try {
+                this.currentTask.get().get();
+            } catch (CancellationException | InterruptedException | ExecutionException e) {
+                this.isClosed.set(true);
+                throw new RuntimeException("Exception while writing records asynchronously", e);
```
tfenne, line 155: In the case of an ExecutionException we might want to unwrap it and rethrow the exception that caused the ExecutionException.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
```diff
@@ -0,0 +1,253 @@
...
+     * the same API as {@link Writer}, but will manage buffering of writes and sending to the {@link AsyncWriterPool} it
+     * is attached to.
+     * <p>
+     * Note, {@code PooledWriter} is not intrinsically threadsafe, and any writing / closing done across multiple
+     * threads should be synchronized.
+     *
+     * @param <A> the type of the item the writer can write
+     */
+    public class PooledWriter<A> implements Writer<A> {
+        private final AtomicBoolean isClosed = new AtomicBoolean(false);
```
tfenne, line 91: I think this also could be a regular `boolean` now too, right?

tfenne (commented):
Looking great @sstadick!  A few more minor comments, but I think we're 99% there.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
sstadick: No, I think as it is here, there is a chance that `AsyncWriterPool.close` could launch the `nonBlockingClose` and then someone could call `close` on the writer leading to two threads modifying things and two threads writing at the same time. I've changed it so that now:

1. if isClosed is true, and either `nonBlockingClose` or `close` are called it will result in an exception. 
2. `nonBlockingClose` now immediatly sets `isClosed` to true so that two "closing" pathways can't be running at the same time.
3. I've followed the same pattern as `checkAndRethrow` so that there is now a `blocking` and `nonBlocking` close method.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
sstadick: `currentTask` is updated by calls to `drain` and calls to `checkAndRethrow`. The only time those methods can be called from another thread is during the `nonBlockingClose`. I believe that with the newest changes to `nonBlockingClose` we remain in a scenario where only one thread can access `currentTask`.

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
sstadick: see comment about `currentTask`

src/main/java/htsjdk/samtools/util/AsyncWriterPool.java
sstadick: Is there a way to do that without making `ExecutionException` a checked exception?

src/main/java/htsjdk/io/AsyncWriterPool.java
```diff
@@ -0,0 +1,270 @@
...
+     * threads should be synchronized.
+     *
+     * @param <A> the type of the item the writer can write
+     */
+    private class PooledWriter<A> implements Writer<A> {
+        private final BlockingQueue<A> queue;
+        private final Writer<A> writer;
+        private final int writeThreshold;
+
+        private Boolean isClosed = false;
```
tfenne, line 91: I think `scala` has ruined you for Java.  In Java `boolean` with a `b` is a primitive and `Boolean` is an object that wraps a primitive.  Unless you need it to be nullable, `boolean` is much preferred.

src/main/java/htsjdk/io/AsyncWriterPool.java
```diff
@@ -0,0 +1,270 @@
...
+         *
+         * @param writer         a class implementing {@link Writer}
+         * @param queue          a queue to use for this writer, bounded or unbounded.
+         * @param writeThreshold the minimum number of items needed before scheduling a thread for writing.
+         * @throws IllegalArgumentException if writeThreshold <= 0 or if writeThreshold > queue.remainingCapacity()
+         */
+        private PooledWriter(final Writer<A> writer, final BlockingQueue<A> queue, final int writeThreshold) {
+            if (writeThreshold <= 0) throw new IllegalArgumentException("writeThreshold must be >= 1");
+            if (writeThreshold > queue.remainingCapacity())
+                throw new IllegalArgumentException("writeThreshold can't be larger then queue capacity.");
```
tfenne, line 107:
```suggestion
            if (writeThreshold <= 0) throw new IllegalArgumentException("writeThreshold must be >= 1: " + writeThreshold);
            if (writeThreshold > queue.remainingCapacity())
                throw new IllegalArgumentException("writeThreshold (" + writeThreshold + ") can't be larger then queue capacity (" + queue.remainingCapacity() + ").");
```

src/main/java/htsjdk/io/AsyncWriterPool.java
```diff
@@ -0,0 +1,270 @@
...
+         * {@code nonBlocking} variants that go by the same name.
+         *
+         * @throws RuntimeException if any exception was raised during the writing on the receiving thread
+         */
+        private void checkAndRethrow() {
+            try {
+                this.currentTask.get();
+            } catch (CancellationException | InterruptedException | ExecutionException e) {
+                this.isClosed = true;
+                throw new RuntimeException("Exception while writing records asynchronously", e);
```
tfenne, line 152:
```suggestion
            } catch (CancellationException | InterruptedException | ExecutionException e) {
                this.isClosed = true;
                final Throwable t = e instanceof ExecutionException ? e.getCause() : e;
                throw new RuntimeException("Exception while writing records asynchronously", t);
```

src/main/java/htsjdk/io/AsyncWriterPool.java
```diff
@@ -0,0 +1,270 @@
...
+        /**
+         * Non-blocking check of the previously completed task. This will check the future if it is not null and if it
+         * isDone. If the task generated any exception, these will be re-thrown.
+         *
+         * @throws RuntimeException if any exception was raised during the writing on the receiving thread
+         */
+        private void nonBlockingCheckAndRethrow() {
+            if (this.currentTask != null && this.currentTask.isDone()) {
+                this.checkAndRethrow();
+            }
```
tfenne, line 123:
```suggestion
            if (this.currentTask != null && this.currentTask.isDone()) checkAndRethrow();
```

src/main/java/htsjdk/io/AsyncWriterPool.java
```diff
@@ -0,0 +1,270 @@
...
+        /**
+         * Blocking check of the previously completed task. This will wait for the future to complete if it is not null.
+         * If the task generated any exception, these will be re-thrown.
+         *
+         * @throws RuntimeException if any exception was raised during the writing on the receiving thread
+         */
+        private void blockingCheckAndRethrow() {
+            if (this.currentTask != null) {
+                this.checkAndRethrow();
+            }
```
tfenne, line 135:
```suggestion
            if (this.currentTask != null) checkAndRethrow();
```

src/main/java/htsjdk/io/AsyncWriterPool.java
```diff
@@ -0,0 +1,270 @@
...
+         * @param item an item to enqueue for writing
+         * @throws RuntimeIOException if queue is already closed
+         * @throws RuntimeException   if the previous attempt to write failed, or if the attempt to add to the queue has
+         *                            been interrupted
+         */
+        @Override
+        public void write(final A item) {
+            if (this.isClosed) throw new RuntimeIOException("Attempt to add record to closed writer.");
+
+            this.nonBlockingCheckAndRethrow();
```
tfenne, line 173:
```suggestion
            nonBlockingCheckAndRethrow();
```

FYI the common practise in Java is to prefix member _variables_ with `this.` and to omit the `this.` for method calls.

src/main/java/htsjdk/io/AsyncWriterPool.java
```diff
@@ -0,0 +1,270 @@
...
+
+            // Put new item in queue
+            try {
+                while (!this.isClosed && !this.queue.offer(item, 5, TimeUnit.SECONDS)) {
+                }
```
tfenne, line 178:
```suggestion
                while (!this.isClosed && !this.queue.offer(item, 5, TimeUnit.SECONDS)) { /* Just wait. */ }
```

src/main/java/htsjdk/io/AsyncWriterPool.java
```diff
@@ -0,0 +1,270 @@
...
+            } catch (InterruptedException e) {
+                throw new RuntimeException("Exception while placing item in queue", e);
+            }
+
+            // If queue size is large enough, and the there is no last task
+            if (queue.size() >= this.writeThreshold && this.currentTask == null) {
```
tfenne, line 184:
```suggestion
            if (this.currentTask == null && queue.size() >= this.writeThreshold) {
```
Just inverting the order of checks since a null-check is fast and I think `size()` on blocking queues is a little more expensive.

src/main/java/htsjdk/io/AsyncWriterPool.java
```diff
@@ -0,0 +1,270 @@
...
+                this.drain();
```
tfenne, line 185:
```suggestion
               drain();
```

src/main/java/htsjdk/io/AsyncWriterPool.java
```diff
@@ -0,0 +1,270 @@
...
+        /**
+         * Launch a task in the {@link AsyncWriterPool} to pull items from this writers queue, and write them, till the
+         * queue is empty.
+         * <p>
+         * If items are being added to the queue faster than they can be written, and the queue is unbounded, the queue
+         * may result in an {@code OutOfMemoryError}.
+         */
+        private void drain() {
+            // check the result of the previous task
+            this.currentTask = AsyncWriterPool.this.executor.submit(() -> {
```
tfenne, line 198:
```suggestion
            if (this.currentTask != null) throw new IllegalStateException("drain() called while currentTask is not null");
            this.currentTask = AsyncWriterPool.this.executor.submit(() -> {
```

src/main/java/htsjdk/io/AsyncWriterPool.java
```diff
@@ -0,0 +1,270 @@
...
+         * Non-blocking close the the enclosed writer. This wraps a call to {@link PooledWriter#blockingClose()} and
+         * sends it to execute on the executor. {@code .isClosed} is immediately set to {@code true} on the current
+         * thread. The new thread will block until any ongoing writes are complete, drain any remaining enqueued items,
+         * check for exceptions, and then call close on the underlying writer.
+         *
+         * @return a future that can be waited on
+         * @throws IllegalStateException if isClosed is already true
+         */
+        private CompletableFuture<Void> nonBlockingClose() {
+            if (this.isClosed) throw new IllegalStateException("Writer is already closed");
```
tfenne, line 221: I tend to think that this should just return - I think it's totally reasonable that I as a user might first close the writers independently (they do implement closeable after all) and _then_ close the pool which is going to call these methods again.

src/main/java/htsjdk/io/AsyncWriterPool.java
```diff
@@ -0,0 +1,270 @@
...
+         * finally closing the writer itself.
+         *
+         * @throws IOException           if the enclosed writer's close raises an exception
+         * @throws IllegalStateException if isClosed is already true
+         */
+        @Override
+        public void close() throws IOException {
+            if (this.isClosed) throw new IllegalStateException("Writer is already closed");
+
+            this.isClosed = true;
```
tfenne, line 266:
```suggestion
```

I would just have this call `blockingClose()` and move any checks in there as I think that makes blockingClose() and nonBlockingClose() more symmetrical.

src/main/java/htsjdk/io/AsyncWriterPool.java
```diff
@@ -0,0 +1,270 @@
...
+         * @throws IOException if the enclosed writer's close raises an exception
+         */
+        private void blockingClose() throws IOException {
+            // Wait for any ongoing writes to finish and check for errors
+            this.blockingCheckAndRethrow();
+            // drain any remaining items in the queue
+            this.drain();
+            // Wait for last write to finish
+            this.blockingCheckAndRethrow();
+            this.writer.close();
```
tfenne, line 250:
```suggestion
            if (this.isClosed) {
                this.blockingCheckAndRethrow();
            }
            else {
                this.isClosed = true;
            
                // Wait for any ongoing writes to finish and check for errors
                this.blockingCheckAndRethrow();
                // drain any remaining items in the queue
                this.drain();
                // Wait for last write to finish
                this.blockingCheckAndRethrow();
                this.writer.close();
            }
```

src/main/java/htsjdk/io/Writer.java
```diff
@@ -0,0 +1,15 @@
+package htsjdk.io;
+
+import java.io.Closeable;
+
+public interface Writer<A> extends Closeable {
+
+    void write(A item);
```
tfenne, line 7: JavaDoc please (I know it's obvious, but still.)

tfenne (approved):
Why are there changes to unrelated classes (e.g. SnappyLoaderUnitTest, IOUtilTest)?  Can you please back those changes out - I don't mind reformatting of classes that are already part of the PR but I don't think it makes sense to include formattng changes to unrelated classes in a PR - it's just going to be confusing looking at history later.

Other than that I have made a bunch of minor suggestions. I think once those are incorporated and tests are still passing, we're good to merge.

src/main/java/htsjdk/samtools/fastq/FastqWriter.java
```diff
@@ -1,14 +1,17 @@
...
+
 import java.io.Closeable;
 
 /**
  * Simple interface for a class that can write out fastq records.
  *
  * @author Tim Fennell
  */
-public interface FastqWriter extends Closeable {
+public interface FastqWriter extends Closeable, Writer<FastqRecord> {
```
sstadick, line 12: This is the only change of substance here.

src/test/java/htsjdk/io/AsyncWriterPoolTest.java
```diff
@@ -0,0 +1,125 @@
...
+                return false;
+            }
+        }
+        return true;
+    }
+
+
+    @Test
+    public void testWritingToFile() throws IOException {
+        // Test with thread pools sized 1-4
```
sstadick, line 66: Note this tries pool sizes of 1-4

tfenne:
@lbergelson I believe this is ready for merging.  Ok to do so or would you like to take a look?  Also, it appears that the tests are passing on 2/4 of the travis builds, but that it's failing to start the test server on the other 2/4.  I don't think that should block this PR but do you know anything about that?

lbergelson:
@tfenne I'd like to take a look before you merge.  Would you mind waiting a day or two?  Just so I can familiarize myself and ask any questions while it's still fresh.

tfenne:
@lbergelson Happy to wait a day or two!

sstadick:
Hi @lbergelson! Just wanted to bump this and see if you had any questions.

src/main/java/htsjdk/io/AsyncWriterPool.java
```diff
@@ -0,0 +1,252 @@
...
+     * Any class that implements {@link Writer} can be exchanged for a {@code PooledWriter}. The PooledWriter provides
+     * the same API as {@link Writer}, but will manage buffering of writes and sending to the {@link AsyncWriterPool} it
+     * is attached to.
+     * <p>
+     * Note, {@code PooledWriter} is not intrinsically threadsafe, and any writing / closing done across multiple
+     * threads should be synchronized.
+     *
+     * @param <A> the type of the item the writer can write
+     */
+    private class PooledWriter<A> implements Writer<A> {
```
lbergelson, line 90: My instinct is exactly the opposite of Tim's here.  I was going to say make this static.

lbergelson (changes requested):
@sstadick I have a few comments/ questions.  The only ones that really need resolving are the comments on the tests.

Did you look into adapting AsyncWriter to make it it usable instead of having to reimplement it as a private class?  It's unfortunate to have AsyncWriterPool which has no relation to AsyncWriter.

src/main/java/htsjdk/io/Writer.java
```diff
@@ -0,0 +1,20 @@
+package htsjdk.io;
+
+import java.io.Closeable;
+
+public interface Writer<A> extends Closeable {
```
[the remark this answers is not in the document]
sstadick: I agree, I just wanted to minimize changes for this PR.

src/main/java/htsjdk/io/AsyncWriterPool.java
```diff
@@ -0,0 +1,252 @@
...
+         */
+        @Override
+        public void write(final A item) {
+            if (this.isClosed) throw new RuntimeIOException("Attempt to add record to closed writer.");
+
+            nonBlockingCheckAndRethrow();
+
+            // Put new item in queue
+            try {
+                while (!this.isClosed && !this.queue.offer(item, 5, TimeUnit.SECONDS)) { /* Just wait. */ }
```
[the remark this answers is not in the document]
sstadick: In the docstrings for both the `PooledWriter` and the `AsyncWriterPool` I note that these classes are not threadsafe and should not be acted no from multiple threads. So, if we take it as given that only one thread will be calling `write` and `close` on the `PooledWriter`, those operations can only happen synchronously which I believe avoids the scenario you are describing. A writer can't be closed while writing.

Regarding the timeout, I don't have a strong opinion on it. @tfenne did you have a particular reason for going with 5 seconds as opposed to 1 second?

Given the constraint that the `PooledWriter` and `AsyncWriterPool` should only ever be accessed from the same thread, and that the drain method first pops an item off the queue before writing, meaning it always makes space for one new item, even if the write fails, allowing the emplacement of the waiting item into the queue and a retrigger of the checking methods, is there is reason we can't go back to using `put`?

src/test/java/htsjdk/io/AsyncWriterPoolTest.java
```diff
@@ -0,0 +1,125 @@
...
+                assert lines.size() == 100;
+            }
+        }
+    }
+
+    @Test
+    public void testNoSelfSuppression() throws IOException {
+
+        AsyncWriterPool pool = new AsyncWriterPool(4);
+        File file = File.createTempFile("AsyncPoolWriterTest", ".tmp");
```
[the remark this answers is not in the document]
sstadick: I added `file.deleteOnExit()` to avoid having to set up a tmp directory and I think does the same thing as `newTempFile` internally

src/main/java/htsjdk/io/AsyncWriterPool.java
sstadick: I think this has to be non-static so that it can access the pool instance of the outer class. I'm not a java expert though and defer to either your @tfenne!

sstadick:
@lbergelson Thanks for taking the time to go through this. Let me know if there's anything I didn't address.

Regarding using AsyncWriter, if we are looking the same thing, AsyncWriter is a thread-per-writer paradigm vs AsyncWriterPool, which is an n-writers to m-threads relationship. I agree that the naming leaves a little to be desired and am open to other names. I address this in the newly added docs for `AsyncWriterPool` as well. Hopefully that clarifies things a bit.

sstadick:
@lbergelson, just wanted to check in on this and see if you've had a chance to give it another look.

src/main/java/htsjdk/io/AsyncWriterPool.java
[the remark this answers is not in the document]
tfenne: I don't have strong feelings about the timeout, and perhaps the right answer is to also make this configurable.  Maybe we can a) make the wait time a member of `AsyncWriterPool` and then have methods to get/set it?  I don't think we need to have constructor variants for it, just a way to set it.

Re: 5 vs. 1 second, I think this is really just trying to make sure that in the case where something goes wrong (e.g. the the underlying writers are all failing to write due to filesystem/network issues, or threads are dying) that the client thread that is pushing records into the writers/pool doesn't hang forever.  That's why we want to us `offer()` instead of `put()` @sstadick.  And I _think_ that because that's really the only reason to use `offer()` that having a high timeout is fine - the timeout is really "how long do you want to wait to return in the case that things have gone horribly wrong", and has minimal to no effect on normal operations/usage.

So perhaps all that's really needed here is a better comment explaining what I'm saying above?  And just to be clear for @lbergelson the `offer()` method will block for the timeout duration, which means that thread gets parked and won't spin or cause any load, and when a space frees up in the queue it will be notified, get un-parked and deposit it's item.

src/main/java/htsjdk/io/Writer.java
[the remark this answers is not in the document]
tfenne: If it's a couple of minutes work, let's maybe add this to the SAM filter writer types and the VariantContextWriter since those are the two other big ones?

src/main/java/htsjdk/io/AsyncWriterPool.java
tfenne: It either has to be non-static, _or_ take in a reference to the pool in it's constructor.  But yes, one way or another it needs a reference to the pool instance.  I don't have strong feelings either way on whether that's by being a non-static inner class, or by taking the pool directly in the constructor.

src/main/java/htsjdk/io/AsyncWriterPool.java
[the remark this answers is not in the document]
lbergelson: I like the fact that it won't hang if something causes things to stop prematurely.  I think we should just clearly throw an error if it hits the case where it hits the case and can't successfully complete the offer.

src/main/java/htsjdk/io/AsyncWriterPool.java
lbergelson: Yeah, my feeling is always static if you can and just pass in through the constructor, but it's not a big deal here either way.  I just thought it was funny that we basically had opposite feelings about the cleanest way to do it.

lbergelson:
@sstadick Hey, sorry for the slow response.  On top of my regularly scheduled pandemic slowness I had a weird week with a death in the family, then the impeachment, then my birthday, so it was a particularly distracted week for me..  What I meant about AsyncWriter, is that it's a very similar class to the PooledWriter, and it seems like maybe AsyncWriter could have been adapted with an additional constructor and some slight changes and used as the pooled writer instead of adding another implementation.  Or the pooled writer implementation might be cleaner and more modern and maybe it could replace AsyncWriter.

src/main/java/htsjdk/io/AsyncWriterPool.java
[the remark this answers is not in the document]
lbergelson: @tfenne I didn't mean to imply that it would waste a lot of cycles, just that you could potentially have a bunch of 5s waits in your program in pathological situations.  You'd have to pick bad queue sizes though to trigger it I think.

sstadick:
@lbergelson @tfenne I'll address outstanding questions below, but I think everything is in a good state as is.

`AsyncWriter` vs `AsyncWriterPool`: I still think that their models of viewing the world are fundamentally different enough that they warrant being separate. `AsyncWriter` being NxN writers to threads and `AsyncWriterPool` being MxN writers to threads where `AsyncWriterPool` is aimed at scenarios where you have way more writers than threads. `AsyncWriterPool` can be NxN, but there is overhead introduced that is avoided by `AsyncWriter`. This difference is addressed in the docs for `AsyncWriterPool`.

Implementing `Writer` on other writer types: I looked into doing this, but the other writers have not standardized around a `write` method and reworking them to do that would introduce non-trival changes that feel out of scope for this PR.

Let me know if I didn't address all outstanding issues.

@lbergelson, 6 days ago feels like a very long time, but I'm sorry to hear that week! That's a lot all at once.

tfenne:
@lbergelson I'm going to go ahead and squash and merge this later today.  Please yell if you object - but it sounds like we're in a good place.  And we can happily address any lingering concerns in follow-up PRs too.

lbergelson:
@tfenne It still seems like there's a gross failure mode where it will think it did a successful write if the thread is closed while it's waiting on an offer.  I'm ok merging and addressing any problems in future PRs though.

lbergelson:
@sstadick Yeah, 6 days is a long time.  Let me know if you want to do some babysitting and I'll respond faster :)

sstadick:
@lbergelson I'm good for now, rate of response is more than adequate 👍

tfenne:
Thanks @lbergelson I'm going to merge this so that @jacarey can start using it in some Picard work he's doing.  That should give it a really good stress test and you can be sure we'll come back and fix any issues that come up.

PR #1528, opened 2021-01-06
Gff3Writer : give a chance to implement a custom way to escape strings

lindenb:
### Description

the Gff3Writer escapes strings as UTF8. While the format is **correct**, it produces strings that are not expected by some gff3 parsers. For example: it would generate a line like:

```
chr1	ucsc	exon	11869	12227	.	+	.	Parent=transcript%3ANR_148357.t1;Name=NR_148357;biotype=misc_RNA;exon_id=NR_148357%3AE0
```

which is not recognized by "bcftools csq" because it doesn't look for "Parent=transcript%3A" but for "Parent=transcript:"

```
[csq.c:729 gff_id_parse] Could not parse the line, "Parent=transcript:" not present: chr1	ucsc	exon	11869	12227	.	+	.	Parent=transcript%3ANR_148357.t1;Name=NR_148357;biotype=misc_RNA;exon_id=NR_148357%3AE0
```

this PR just  adds a new **protected** method `escapeString` which invokes `encodeString`. It can be overriden to implement a custom way to escape the strings.

PS: for now, travis-ci is too slow on my side, I cannot validate the code.

- [X] Check your code style.

### DIFF
--- a/src/main/java/htsjdk/tribble/gff/Gff3Writer.java
+++ b/src/main/java/htsjdk/tribble/gff/Gff3Writer.java
@@ -68,9 +68,9 @@ public class Gff3Writer implements Closeable {
 
     private void writeFirstEightFields(final Gff3Feature feature) throws IOException {
         writeJoinedByDelimiter(Gff3Constants.FIELD_DELIMITER, this::tryToWrite, Arrays.asList(
-                encodeString(feature.getContig()),
-                encodeString(feature.getSource()),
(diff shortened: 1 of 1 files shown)


lbergelson:
@lindenb This seems like a reasonable accommodation for dealing with those tools.

lbergelson (approved):

lbergelson:
@lindenb Thank you.  I hope it does what you need now!

PR #1530, opened 2021-01-11
Fix CRAM BAI creation error message.

cmnbroad:
Fix the not entirely useful CRAM error message that was reported in Picard https://github.com/broadinstitute/picard/issues/1624. Its still pretty cryptic for an end user, since it indicates an internal error or malformed input, but at least its now more precise.


lbergelson (approved):
Looks good.

PR #1531, opened 2021-01-28
Allow multiple VCFSimpleHeaderLine of same type

### LINKED ISSUE #277: VCFHeader class should allow multiple "other metadata" lines with the same key to be added
The proposed VCF 4.3 spec (https://github.com/samtools/hts-specs/pull/88) adds some clarifications regarding uniqueness of header lines. Specifically, "structured" lines that have their value enclosed with "<>" must have an ID attribute that is unique within their type (contig, filter, etc.), however the actual keys for header lines (this is the part before the first "=" sign, eg., "comment" in "##comment=X" or "contig" in "##contig=X") are not required to be unique.

The current `VCFHeader` class in htsjdk is not 100% compliant with the above. Specifically, although it does correctly enforce the requirement of unique IDs within each type of "structured" line, and allows multiple lines with the same key but different IDs (eg., multiple "##contig=" lines with different ID values), for unstructured lines with no ID attribute (which are called "other" lines in the code), it incorrectly forbids adding a new line with a duplicate key. This is a consequence of using a Map to store the "other" header lines:

        private final Map<String, VCFHeaderLine> mOtherMetaData = new LinkedHashMap<String, VCFHeaderLine>();

Calling `addMetaDataLine()` to add one of these "other" header lines with the same key as an existing line will silently refuse to add the new line, since it does a map lookup by key before allowing the line to be added.

The situation is complicated by the fact that in addition to the `mOtherMetaData` map (used for lookup operations on "other" header lines), a separate copy of all header lines is also stored in a set, and this set is used when the header is actually written out:

        private final Set<VCFHeaderLine> mMetaData = new LinkedHashSet<VCFHeaderLine>();

So a `VCFHeader` could have (for example) multiple "##comment=" lines, and these will be correctly read into `mMetaData` and written out again, but lookup operations via `getOtherHeaderLine()` will return only a single "##comment=" line, and add operations via `addMetaDataLine()` will silently fail to add additional "##comment=" lines, since these operations use the map rather than the set.

The class should be patched so that:
1. `addMetaDataLine()` allows "other metadata" header lines with the same key as existing lines to be added (while still enforcing the requirement of unique ID attributes within each type of "structured" header line)
2. `getOtherHeaderLine()` is capable of returning multiple values for a particular key.

And unit tests should be added to confirm the above behavior.

### LINKED ISSUE #500: VCF header lines get dropped

mjhipp:
### Description

Closes #277 
Closes #500 

Currently, it is not possible to add multiple `VCFSimpleHeaderLine`s to a VCFHeader if they are of the same type (see issues linked above). A specific example, is adding `ALT` lines, which are included in [4.2 spec](https://samtools.github.io/hts-specs/VCFv4.2.pdf) (that shows a header with multiple ALT lines on page 11).

To fix this, I added a new map and methods for working with `SimpleVCFHeaderLine`s that are not of the default types. 

I have seen another fix to this (#835), but it was closed due to large scope. I am hoping this change is small and useful enough to be considered. Thanks!

- [x] Check your code style.

### DIFF
--- a/src/main/java/htsjdk/variant/vcf/VCFHeader.java
+++ b/src/main/java/htsjdk/variant/vcf/VCFHeader.java
@@ -373,6 +373,9 @@ public class VCFHeader implements Serializable {
             return addMetaDataLineMapLookupEntry(mFilterMetaData, filterLine.getID(), filterLine);
         } else if ( line instanceof VCFContigHeaderLine ) {
             return addContigMetaDataLineLookupEntry((VCFContigHeaderLine) line);
+        } else if ( line instanceof VCFSimpleHeaderLine ){
+            final VCFSimpleHeaderLine simpleLine = (VCFSimpleHeaderLine) line;
+            return addMetaDataLineMapLookupEntry(mOtherMetaData, simpleLine.getID(), simpleLine);
         } else {
             return addMetaDataLineMapLookupEntry(mOtherMetaData, line.getKey(), line);
         }
--- a/src/main/java/htsjdk/variant/vcf/VCFHeader.java
+++ b/src/main/java/htsjdk/variant/vcf/VCFHeader.java
@@ -42,6 +42,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -75,6 +76,7 @@ public class VCFHeader implements Serializable {
     private final Map<String, VCFInfoHeaderLine> mInfoMetaData = new LinkedHashMap<String, VCFInfoHeaderLine>();
     private final Map<String, VCFFormatHeaderLine> mFormatMetaData = new LinkedHashMap<String, VCFFormatHeaderLine>();
     private final Map<String, VCFFilterHeaderLine> mFilterMetaData = new LinkedHashMap<String, VCFFilterHeaderLine>();
+    private final Map<String, Map<String, VCFSimpleHeaderLine>> mIdMetaData = new LinkedHashMap<String, Map<String, VCFSimpleHeaderLine>>();
     private final Map<String, VCFHeaderLine> mOtherMetaData = new LinkedHashMap<String, VCFHeaderLine>();
     private final Map<String, VCFContigHeaderLine> contigMetaData = new LinkedHashMap<>();
 
@@ -374,8 +376,7 @@ public class VCFHeader implements Serializable {
         } else if ( line instanceof VCFContigHeaderLine ) {
             return addContigMetaDataLineLookupEntry((VCFContigHeaderLine) line);
         } else if ( line instanceof VCFSimpleHeaderLine ){
-            final VCFSimpleHeaderLine simpleLine = (VCFSimpleHeaderLine) line;
-            return addMetaDataLineMapLookupEntry(mOtherMetaData, simpleLine.getKey() + ":" + simpleLine.getID(), simpleLine);
+            return addIdMetaDataLineLookupEntry((VCFSimpleHeaderLine) line);
         } else {
             return addMetaDataLineMapLookupEntry(mOtherMetaData, line.getKey(), line);
         }
@@ -402,6 +403,24 @@ public class VCFHeader implements Serializable {
         }
         contigMetaData.put(line.getID(), line);
         return true;
+    }
+
+    /**
+     * Add a simple header line to the lookup list for ID header lines (mIdMetaData). If there's
+     * already an ID line with the same field type and ID, does not add the line.
+     *
+     * Note: does not add the ID line to the master list of header lines in mMetaData --
+     *       this must be done separately if desired.
+     *
+     * @param line ID header line to add
+     * @return true if line was added to the list of ID lines, otherwise false
+     */
+    private boolean addIdMetaDataLineLookupEntry(final VCFSimpleHeaderLine line) {
+        // Check first layer for key/field type
+        if (!mIdMetaData.containsKey(line.getKey())) {
+            mIdMetaData.put(line.getKey(), new LinkedHashMap<String, VCFSimpleHeaderLine>());
+        }
+        return addMetaDataLineMapLookupEntry(mIdMetaData.get(line.getKey()), line.getID(), line);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


clintval:
Hey @lbergelson is this a PR your team would be willing to review? Many thanks if so!

lbergelson:
@clintval Yeah, we're definitely going to get to it.  Reviews have been really slow lately because working from home in the pandemic is killing my productivity.

cmnbroad:
While it does fix the duplicate line problem, it changes current behavior in a way that we should at least think about. Currently, any ID header line other than INFO/FORMAT (i.e., ALT, META, SAMPLE, PEDIGREE, or any other custom line) can be retrieved via `getOtherHeaderLine(String key)`. I think with this change that will no longer be the case.

The real issue is that `otherMeta` probably shouldn't have **any** ID lines in it. Making the change in this PR and adding the new `getOtherHeaderLine(String key, String id)` kind of codifies that, though, and it feels like that's going in the wrong direction. An alternative fix would be to stop putting the ID lines in there at all, and store them elsewhere, and expose them through the new method. That would have the same backward compatibility issue as this fix, but it would also eliminate the need for the synthetic key names, which might be confusing since they look a lot like the ALT SV ID type hierarchy.

If we're going to break compatibility to fix this, I'd be inclined to favor doing the more complete fix though it's likely to be a bit more work and might impact other methods. Or maybe we should just resurrect the old PR and wait for that. @lbergelson any thoughts ?

mjhipp:
> An alternative fix would be to stop putting the ID lines in there at all, and store them elsewhere, and expose them through the new method.

A nested map is one possibility for this. Where the outer map is from line type ("ALT", "PEDIGREE", etc) to an inner map, and the inner map is ID to metadata line. This would scale for any reasonable amount of line types, without having to hard code them (defining `mAltMetaData` map).

This would require a similar change to methods, but without synthetic key names.

Would add a new map at the top:

```java
private final Map<String, Map<String, VCFSimpleHeaderLine>> mIdMetaData = new LinkedHashMap<String, Map<String, VCFSimpleHeaderLine>>();
```

Adding a new line would involve checking for the `key`, (making a new inner map if necessary), and adding to the inner map at the `id` hash.

which would change the `get` method I added to:

```java
    /**
     * @param key    the header key or field type
     * @param id     the header id
     * @return the meta data line, or null if there is none
     */
    public VCFSimpleHeaderLine getIdHeaderLine(final String key, final String id) {
      return mIdMetaData.get(key).get(id);
    }
```

(probably on multiple lines with a null check in there)

Could possibly also change the existing `getOtherHeaderLine` to add some backwards compatibility. This could involve first checking `mOtherMetaData` for the `key`, and if not found, check `mIdMetaData` for that `key`. If the `key` is found in the outer map of `mIdMetaData`, take the first (or last) value in the inner map and cast to `VCFHeaderLine`.

mjhipp:
I made the changes above. I think now the only backwards compatibility issue is that `getOtherHeaderLines()` will not return `VCFSimpleHeaderLines`, where it had before. To get those, you use `getIdHeaderLines()`

src/main/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -579,11 +601,23 @@ public boolean hasFilterLine(final String id) {
...
+        if (mOtherMetaData.containsKey(key)) {
+            return mOtherMetaData.get(key);
+        } else if (mIdMetaData.containsKey(key)) {
+            // Get the first item in the linked hash map
+            Map<String, VCFSimpleHeaderLine> fieldMetaData = mIdMetaData.get(key);
+            if (fieldMetaData.keySet().size() > 0) {
+                return fieldMetaData.get(fieldMetaData.keySet().iterator().next());
+            }
+        }
+        return null;
```
mjhipp, line 620: This is for backwards compatibility, where previously a `VCFSimpleHeaderLine` could be accessed by this method (only one per field type). It may be better to let this return `null` and force users to use `getIdHeaderLine()` to find `VCFSimpleHeaderLine`s. This would be more consistent with `getOtherHeaderLines`/`getIdHeaderLines`.

cmnbroad: While this method makes sense by itself, it introduces a new inconsistency with `getOtherHeaderLines()`, since it has a different idea of what "other" means (it returns lines that will never be returned by the existing method).

src/main/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -593,6 +627,30 @@ public VCFHeaderLine getOtherHeaderLine(final String key) {
...
+     * Returns the ID HeaderLines, in order by:
+     * - First key/field type to be added
+     * - Within key/field types, in original order
+     */
+    public Collection<VCFSimpleHeaderLine> getIdHeaderLines() {
+        final Collection<VCFSimpleHeaderLine> lines = new LinkedList<>();
+        mIdMetaData.forEach((k, v) -> lines.addAll(v.values()));
+        return lines;
+    }
+
```
cmnbroad, line 653: This class already has a public method `List<VCFIDHeaderLine> getIDHeaderLines()` that returns all ID lines. I think it will be confusing to also have this new method, with a slightly different name and signature that returns only the ID header lines that are not INFO, FORMAT, etc.

cmnbroad (commented):
@mjhipp Again, thanks for attempting to fix this. I added a couple of more inline comments on your most recent commit, but I'm a bit skeptical that we can fix this issue without introducing a different set of new issues (especially given that we don't have good unit tests for the existing methods).

Before we go any further we should get @lbergelson's thoughts on whether we want trade off some new inconsistency/weirdness in order to fix the duplicate line issue. If so, I think we'd need to add a bunch more unit tests to verify the behavior of whatever tradeoffs we make, but I'll hold off on further comment until he chimes in.

cmnbroad:
After further discussing this internally and thinking more about it, I'm inclined to wait for a more complete fix to VCFHeader and address the numerous issues all at once rather than continue trying to make incremental fixes. I've resurrected and rebased my old #835 branch and will resubmit a new draft PR so we can try to find a path forward and get that merged.

PR #1532, opened 2021-01-28
Fix error message typo in CRAMLazyReferenceSource

ldgauthier:
Fix slightly misleading rogue paren following contig name in error message.


lbergelson (approved):
👍 Thank you

PR #1533, opened 2021-02-02
Fix CRAMFileReader stream management

cmnbroad:
`CRAMIterator` (which is a closeable iterator) closes it's underlying stream when it's closed (`BAMIterator` doesn't do this; it relies on the enclosing reader to do it). In the case where the underlying stream was supplied to the `CRAMFileReader` constructor directly, rather than through a `File`, this causes serial index queries to fail after the first query, since the `CRAMFileReader` constructors are typed as `InputStream`, and a `SeekableStream` can't be reconstituted from an `InputStream`. These stream constructors are the ones used by `SamReaderFactory` when an input is a non-local file, ie., when it's provided through a SamPathInputResource.

Changing CRAMIterator to not close the stream on `close()` will leak result in leaking streams that are created via other `CRAMFileReader` constructors. The ultimate issue here is that `CRAMFileReader` doesn't have a constructor that accepts a `SeekableStream` directly for the input source, only the index (??). Since rationalizing the constructors is  a bigger change that will require deprecating existing constructors, and affect lots of code and tests, for now this PR uses a `SeekableStream` wrapper class `DeferredCloseSeekableStream` to that is used to create iterators, and defers closing the underlying stream until the reader is closed.

Fixes an issue reported in GATK https://github.com/broadinstitute/gatk/issues/6475.

### DIFF
--- a/src/main/java/htsjdk/samtools/CRAMFileReader.java
+++ b/src/main/java/htsjdk/samtools/CRAMFileReader.java
@@ -39,6 +39,7 @@ public class CRAMFileReader extends SamReader.ReaderImplementation implements Sa
     private File cramFile;
     private final CRAMReferenceSource referenceSource;
     private InputStream inputStream;
+    private DeferredCloseSeekableStream deferredCloseSeekableStream;
     private CRAMIterator iterator;
     private BAMIndex mIndex;
     private File mIndexFile;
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/CRAMFileReader.java
```diff
@@ -387,20 +388,85 @@ public SAMRecordIterator assertSorted(final SortOrder sortOrder) {
...
+        public boolean eof() throws IOException { return delegateStream.eof(); }
+
+        @Override
+        public String getSource() { return delegateStream.getSource(); }
+    }
+
     @Override
     public void close() {
+        // if at any point we created a deferredCloseSeekableStream, close the underlying delegate now
+        if (deferredCloseSeekableStream != null) {
```
lbergelson, line 463: Isn't delegate.close() handled by closing the input stream below?  Do we need to handle this specially or can we just let the close down below do this?

lbergelson (approved):
@cmnbroad One question. Looks good otherwise.  Does the cram reader code have a lockout mechanism to prevent people from instantiating multiple cram iterators at the same time like the bam code does?  Always a bit weird that we generate multiple iterators on the same underlying stream.

src/main/java/htsjdk/samtools/CRAMFileReader.java
cmnbroad: There are a couple of constructors that take `File` that don't initialize `inputStream`. Its not really clear to whether this is truly necessary or not, but I think it's safest to keep it. Having said that, I think this close should be done through CloserUtil, so I'm going to push a commit with that change. Let me know if you disagree.

cmnbroad:
@lbergelson Ni, I don't think there is any such protection for iterators. I believe there is a ticket for that - one of the many things that need to be cleaned up in this file.

lbergelson (approved):

PR #1536, opened 2021-02-08
Correct NM tag calculation

### LINKED ISSUE #1535: SAM NM tag calculation is off-spec
The NM tag describes the number of mismatches between the read and the reference. According to the spec, this is defined as: 

> Number of differences (mismatches plus inserted and deleted bases) between the sequence and
reference, counting only (case-insensitive) A, C, G and T bases in sequence and reference as potential
matches, with everything else being a mismatch. Note this means that ambiguity codes in both
sequence and reference that match each other, such as ‘N’ in both, or compatible codes such as ‘A’ and
‘R’, are still counted as mismatches

Currently, matching ambiguity codes, such as `N - N`, are counted as matches, which is logically correct, but according to the spec, they should be counted as mismatches. This results in Picard's ValidateSamFile to report false errors. Note that this only happens if the read sequence contains ambiguity codes.

michaelgatzen:
Fixes #1535 

- Adding enum BaseComparisonMode to toggle different modes of comparing bases
    - MatchExact: bases match when they are equal
    - MatchAmbiguity: bases match when they are equal or the read base can be expressed using an ambiguity code in the reference
    - NMTagMode: bases match when they are equal AND from `[AaCcGgTt]`. This is compliant to the SAM spec.
- Modified different versions of base comparison methods accordingly
- Modified unit tests to catch off-spec behavior


src/main/java/htsjdk/samtools/util/SequenceUtil.java
```diff
@@ -435,20 +443,43 @@ public static String makeSoftClipCigar(final int clipLength) {
...
+                    return basesEqual(readBase, refBase);
+                }
+            case MatchAmbiguity:
+                if (bisulfiteSequence) {
+                    return bisulfiteBasesMatchWithAmbiguity(negativeStrand, readBase, refBase);
+                } else {
+                    return readBaseMatchesRefBaseWithAmbiguity(readBase, refBase);
+                }
+            case NMTagMode:
+                // TODO Different treatment for bisulfite?
```
michaelgatzen, line 465: Are bisulfite sequences relevant here?

michaelgatzen:
TODO Modify integration test

lbergelson:
@yfarjoun I have a recollection of you having an opinion about this?

yfarjoun:
me too....mostly it was due to being conservative and not liking change, I think....not really a good reason. Perhaps @tfenne has a more substantial opinion?

michaelgatzen:
Just for reference, the reason why this came up is that DRAGEN apparently produces reads with Ns in them, which is the only time that this becomes relevant. This would lead to ValidateSamFile reporting an error on DRAGEN BAMs.

tfenne:
@yfarjoun My opinion is similar - I worry about change.  But if the default behavior in tools remains the same, and there aren't backwards compatibility breaks, I don't object.

FWIW a lot of UMI consensus tools generate reads with Ns in them (e.g. fgbio), so this does come up quite a bit in those situations.  And there are variant callers that use NM to filter reads, so there are use cases where a chance in this behavior would have a significant change on downstream pipelines.

michaelgatzen:
Ok I see. The methods here are indeed used for tools like MergeBamAlignment, so this **will** result in downstream change (for reads with ambiguity codes in them). This change would make it more spec compliant, but I understand the hesitation about change.

One other option is to keep it the way it was and be more liberal in SamFileValidator, so that we don't report errors for files that are spec-compliant. What do people think?

PR #1538, opened 2021-02-22
Invaliding out of date SAMRecord.mAlignmentBlocks cache

d-cameron:
### Description

SAMRecord.mAlignmentBlocks cache field is not invalidated when changing the alignment start.
This causes processes using this field (such as NM tag calculation) to return an incorrect result.

- [N/A] Extended the README / documentation, if necessary
- [X] Check your code style.


lbergelson (approved):
Wow, that's surprising that that has been broken for so long with no one noticing.  Thank you for the fix!

PR #1539, opened 2021-02-23
Add VCF 4.3 writing support

andersleung:
- Percent encoding of special characters is now supported in VCFWriter
- Default behavior is to write VCFs as latest guaranteed compatible version
(pre v4.3 -> 4.2 and v4.3 -> v4.3), but system property can be set to
automatically attempt conversion from v4.2 -> v4.3 and fall back to v4.2
if conversion is impossible, or alternatively fail with error if
conversion is impossible
- Fix bug where extra fields in INFO and FORMAT header lines are dropped

@lbergelson 

### Description

- [x] Check your code style.


lbergelson:
@cmnbroad This PR is relevant to the headerline refactor.

cmnbroad:
@andersleung I won't have time to review this until next week, but I took a quick look and one suggestion would be to integrate `VCFPercentEncoder` and it's tests into the existing `VCFPercentEncodedTextTransformer` and test classes. The `VCFTextTransformer` interface will have to be extended to have `encode` methods to do this, but since the encoder and decoder are complementary, the code, and tests, should live together. Also then the test cases won't have to be duplicated, and the tests can use composition to verify that encode and decode are inverses, i.e., `decode(encode(a)) == a`.

lbergelson (commented):
@andersleung Here are my comments so far. They may or not be relevant after the rebase.

PR #1540, opened 2021-03-03
Fix IOPath::hasExtension NPE when a URI has no hierarchical part.

cmnbroad:
Discovered by James when inadvertently passing a query interval `chr1:18502956`, which looks like exactly like a URI with no hierarchical part, to GATK as a BAM input.


lbergelson (approved):
@cmnbroad Thank you.

cmnbroad:
Thanks for the review @lbergelson.

PR #1541, opened 2021-03-03
Add QuietTestWrapper to suppress excessive test output that exceeds CI log size threshold.

cmnbroad:
We're close to exceeding the the travis threshold for maximum test log size, since TestNG serialize test method arguments, and they wind up in the travis log. I have branches that add new tests that are failing because of this, so we're pretty close to the max already.

If a data provider generates lots of test data (i.e., `List<SAMRecord>`), the log gets pretty big. This PR adds a new class to use with test methods that have arguments with large serialized representation; adds it to a couple of verbose tests; and updates some existing test that already used a similar strategy. 

Now instead of seeing megabytes of this:

image

You see this:
image


lbergelson (approved):
@cmnbroad Looks sane. 👍

PR #1542, opened 2021-03-11
Exclude unintentional transitive junit dependency

lbergelson:
Exclude unintentional transitive junit dependency

Our new "dependency free" json parser mjson accidentally includes junit as a dependency because of a bug in it's pom file.

This manually excludes it.


lbergelson:
@jrobinso This should fix the problem.

lbergelson:
@cmnbroad Can you take a look at this?  Should be pretty simple...

droazen (approved):
:+1:

PR #1548, opened 2021-04-15
Add VCF 4.3 writing

andersleung:
### Description

* Percent encoding of special characters is now supported in VCFWriter
* Default behavior is to write VCFs as latest guaranteed compatible version (pre v4.3 -> 4.2 and v4.3 -> v4.3), but system property can be set to automatically attempt conversion from v4.2 -> v4.3 and fall back to v4.2if conversion is impossible, or alternatively fail with error if conversion is impossible

@lbergelson @cmnbroad 

- [x] Check your code style.


cmnbroad (changes requested):
Did a high level first pass - I haven't looked at the tests yet, but I'm checkpointing what I have to start.

src/main/java/htsjdk/variant/vcf/VCFHeader.java
```diff
@@ -604,6 +604,10 @@ private VCFHeaderVersion initializeHeaderVersion() {
         return metaDataVersion;
     }
 
+    public Collection<VCFValidationFailure<VCFHeaderLine>> getValidationErrors(final VCFHeaderVersion targetVersion) {
+        return mMetaData.getValidationErrors(targetVersion);
+    }
```
cmnbroad, line 609: If the header upgrade code gets moved right into `VCFHeader` (see my comments elsewhere), then this doesn't have to be exposed at all. But if we keep it, it should get javadoc, and be moved in with the other public methods.

src/test/java/htsjdk/beta/codecs/variants/vcf/HtsVCFCodecTest.java
```diff
@@ -105,15 +105,6 @@ public void testRoundTripVCFThroughStream(final IOPath inputPath, final HtsVersi
...
 
-    @Test(expectedExceptions = IllegalArgumentException.class)
-    public void testRejectWritingV43HeaderAsV42()  {
-        // read vcf v4.3 and try to write it to a vcf v4.2 (header is rejected)
-        final IOPath outputPath = IOUtils.createTempPath("rejectWrite43HeaderVCF", ".vcf");
-        readWriteVCFToPath(new HtsPath(VARIANTS_TEST_DIR + "variant/vcf43/all43Features.vcf"),
-                outputPath,
-                VCFCodecV4_3.VCF_V43_VERSION);
-    }
-
```
cmnbroad, line 116: Now that this case succeeds, the "vcfReadWriteTests" data provider at the top of this file should have a 4.3 file added to it (previously, we couldn't read in a 4.3 file and write it out as the "default" version, since the default was 4.2, but now that the default is 4.3, we can add that test case and remove the comment saying why its not there). The new test case would be:

`                { new HtsPath(VARIANTS_TEST_DIR + "variant/vcf43/all43FeaturesCompressed.vcf.gz"), VCFCodecV4_3.VCF_V43_VERSION },
`

cmnbroad: Additionally, there should be at least one new test that demonstrates that a 4.2 input that fails upgrade gets rejected by the VCF4.3 encoder.

There are also a bunch of other accidental things happening in this file now that we should clean up. Specifically, there are test inputs that are subjected to auto-upgrade now, but fail  (because they contain a PEDIGREE line that can't be upgraded - which nicely demonstrates the fallback case and the error message display, so thats good), but we'll need to revisit these tests in light of the changes on this branch to make sure they're still testing what they're supposed to be testing.

Perhaps I should add fix up this file and submit it for merging into your branch while you carry on with the other changes.

src/test/java/htsjdk/variant/vcf/VCFCodec43FeaturesTest.java
```diff
@@ -166,6 +175,81 @@ public void testSymbolicAlternateAllele(final Path testFile, int ignored) {
...
+
+        final VariantContextWriter writer = new VariantContextWriterBuilder()
+            .setOutputFile(out)
+            .unsetOption(Options.INDEX_ON_THE_FLY)
+            .unsetOption(Options.DO_NOT_WRITE_GENOTYPES)
+            .build();
+
+        writer.writeHeader(readHeader);
+        for (final VariantContext vc : readVCF.b) {
+            writer.add(vc.fullyDecode(readHeader, false));
```
cmnbroad, line 195: Is this call to fullDecode necessary ? What happens if you don't call it ? It seems like this test should succeed without i ?

cmnbroad (changes requested):
Only part way through but checkpointing what I have so far.

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1638,106 +1651,221 @@ public boolean isFullyDecoded() {
...
+            newGc.add(gb.make());
+        }
+        builder.genotypesNoValidation(newGc);
+    }
+
+    private void decodeAttributes(
+        final BiConsumer<String, Object> put,
+        final Function<String, VCFCompoundHeaderLine> metadataGetter,
+        final Map<String, Object> attributes,
+        final boolean lenientDecoding
```
cmnbroad, line 1681: We should try find some way to make this code type safe. Its not obvious from the argument types that the `put` and `attributes` args are related, and need to be type-compatible, and its a bit awkward to require the caller to deconstruct the source objects that own the maps and pass the pieces in pairs like this.

One suggestion to consider would be delegating this to new `getDecodedAttributes` methods on `Genotype` and `VariantContext`, since those classes own the map/put pairs that are passed in here together. Or maybe there is some other alternative.

But if thats awkward, I would at least reorder the args so that `attributes` is first, followed by the related `put`, and document how they're related.

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1638,106 +1651,221 @@ public boolean isFullyDecoded() {
...
-            throw new TribbleException("Could not decode field " + field + " with value " + string + " of declared type " + format.getType());
+            return newVC;
         }
     }
 
-    private final void fullyDecodeGenotypes(final VariantContextBuilder builder, final VCFHeader header) {
-        final GenotypesContext gc = new GenotypesContext();
-        for ( final Genotype g : getGenotypes() ) {
-            gc.add(fullyDecodeGenotypes(g, header));
+    private static List<Double> makeGPValueLinear(final List<Double> gpValues) {
```
cmnbroad, line 1853: Suggested name: phredScaleToLinearScale ?

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1638,106 +1651,221 @@ public boolean isFullyDecoded() {
...
+            // This will ensure that the VariantContext is decoded in a percent-encoding aware way
+            final VariantContext newVC = this.fullyDecode(header, true);
+            newVC.version = headerVersion;
+
+            final Object oldGP = newVC.getAttribute(VCFConstants.GENOTYPE_POSTERIORS_KEY);
+            // We need to special-case GP because there is a discrepancy in the scale used to record
+            // its values between pre-4.3 and 4.3+ VCF. Pre-4.3 GP is phred scale encoded while
+            // 4.3+ GP is a linear probability, bringing it in line with other standard keys that
+            // use the P suffix (c.f. VCF 4.3 spec section 7.2).
+            if (headerVersion.isAtLeastAsRecentAs(VCFHeaderVersion.VCF4_3) && oldGP != null) {
```
cmnbroad, line 1838: Shouldn't this also check the VC version ? We shouldn't need to do this if its already 4.3 ?

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1638,106 +1651,221 @@ public boolean isFullyDecoded() {
...
-        builder.genotypesNoValidation(gc);
-    }
 
-    private final Genotype fullyDecodeGenotypes(final Genotype g, final VCFHeader header) {
-        final Map<String, Object> map = fullyDecodeAttributes(g.getExtendedAttributes(), header, true);
-        return new GenotypeBuilder(g).attributes(map).make();
+        final boolean wasLinearScale = GeneralUtils.compareDoubles(sum, 1, VCFConstants.VCF_ENCODING_EPSILON) == 0;
+        if (!wasLinearScale) {
+            gpValues.replaceAll(GP -> QualityUtil.getErrorProbabilityFromPhredScore((int) Math.round(GP)));
+        }
```
cmnbroad, line 1867: This should log a warning if it finds values that don't appear to be version-conforming (i.e, pre v4.3 and already linear).

cmnbroad (changes requested):
Checkpointing a partial review. And I still haven't really looked at the tests.

src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
```diff
@@ -243,13 +252,18 @@ public void close() {
...
             super.add(context);
             if (this.mHeader == null) {
                 throw new IllegalStateException("Unable to write the VCF: header is missing, " +
                                                    "try to call writeHeader or setHeader first.");
             }
+            // If this context came from a version of VCF different from that of the header which we wrote,
+            // we need to make sure it is encoded in a way which is compatible with the version of the header.
+            if (context.getVersion() != this.mHeader.getVCFHeaderVersion()) {
+                context = context.makeCompatibleWithHeaderVersion(this.mHeader);
+            }
```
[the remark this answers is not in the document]
andersleung: FIxed

src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
```diff
@@ -265,43 +279,12 @@ public void add(final VariantContext context) {
     }
 
     @Override
-    public void setHeader(final VCFHeader header) {
-        rejectVCFV43Headers(header);
-
+    public void setHeader(VCFHeader header) {
```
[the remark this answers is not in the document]
andersleung: Fixed

src/main/java/htsjdk/variant/variantcontext/VariantContextBuilder.java
```diff
@@ -214,6 +224,17 @@ public VariantContextBuilder copy() {
...
+    /**
+     * Tells this builder to create a VariantContext conforming to this version of VCF
+     *
+     * @param version the version of VCF to which the VariantContext produced by this builder conforms
+     * @return this builder
+     */
+    public VariantContextBuilder version(final VCFHeaderVersion version) {
+        this.version = version;
+        return this;
+    }
```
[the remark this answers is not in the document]
andersleung: Added javadoc to the getter, but the class is arranged so getters are all separate from setters already (a bit inconsistently), so we'd have to move the whole class around for consistency

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
```diff
@@ -1638,106 +1651,221 @@ public boolean isFullyDecoded() {
...
+     * in the encoding of the GP key or percent encoding of strings between versions. Returns a new VariantContext
+     * if the version of this VariantContext differs from the version of the header
+     * @param header the header to ensure compatibility with
+     * @return the current header if the version of this VariantContext matches the version of the header, otherwise
+     * a new, fully decoded VariantContext, whose attributes are compatible with the header's version
+     */
+    public VariantContext makeCompatibleWithHeaderVersion(final VCFHeader header) {
+        final VCFHeaderVersion headerVersion = header.getVCFHeaderVersion();
+        if (this.version.equals(headerVersion)) {
+            return this;
```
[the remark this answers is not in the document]
andersleung: Removed the check. I wanted to make these static as well (or maybe to have them in a separate class only responsible for upgrading VCs) but that requires threading the VC through most of the methods as an extra argument, which is fine if you think that's cleaner

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
andersleung: Changed

src/main/java/htsjdk/variant/variantcontext/VariantContext.java
andersleung: The check would've been redundant before because we were checking whether the version was the same, but with the other change I added it in

src/main/java/htsjdk/variant/variantcontext/writer/VCFWriter.java
```diff
@@ -125,6 +123,16 @@ public VCFWriter(final Path location, final OutputStream output, final SAMSequen
         this.writeFullFormatField = writeFullFormatField;
     }
 
+    /**
+     * See {@link VCFVersionUpgradePolicy}. Controls how the writer will handle headers and variant contexts
+     * from versions of VCF before the current version. The writer will call {@link VCFHeader#upgradeVersion} headers
+     * passed in by {@link VCFWriter#writeHeader} and {@link VCFWriter#setHeader} before writing the header out.
+     * @param policy the policy to use
+     */
```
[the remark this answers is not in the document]
andersleung: The javadoc in `VCFVersionUpgradePolicy` and `VariantContextWriterBuilder` only mentions `VCFWriter` but I added a little clarifying note that it has no effect on `BCF2Writer` in `VCFVersionUpgradePolicy`, `VariantContextWriterBuilder` and this method.

It doesn't affect BCF because BCF2Writer calls the static method which writes the header verbatim.

tfenne:
Sorry this sat for so long.  I believe this PR to be superseded by #1596 and am closing it.  Feel free to reopen if you disagree and are still interested.

PR #1547, opened 2021-04-14
Adding PL Tag 'DNBSEQ' as the Platform/Technology for BGI/MGI.

gbggrant:
Adding PL Tag 'DNBSEQ' as the Platform/Technology for BGI/MGI.  Leaving BGI as a deprecated enum.

### Description

According to the SAM file spec, the proper PL tag for read groups generated on BGI/MGI technology is 'DNBSEQ'. 
We have reports of a ValidateSamFile error caused by this issue: 
https://github.com/broadinstitute/picard/issues/1661

### Things to think about before submitting:
- [ ] Make sure your changes compile and new tests pass locally.
- [ ] Add new tests or update existing ones:
  - A bug fix should include a test that previously would have failed and passes now.
  - New features should come with new tests that exercise and validate the new functionality.
- [ ] Extended the README / documentation, if necessary
- [ ] Check your code style.
- [ ] Write a clear commit title and message
  - The commit message should describe what changed and is targeted at htsjdk developers
  - Breaking changes should be mentioned in the commit message.

PR #1552, opened 2021-04-23
Deprecate OTHER as a PL value

lbergelson:
* Deprecate the PlatformValue OTHER since the spec advises omitting the PL field when the technology is not listed in the official ontology.
* A continuation of https://github.com/samtools/htsjdk/pull/1547 which deprecated BGI.
* Alphabetized the list of PL values and added clarifying comments to match the spec.

I kind of hate how I've formatted the new list but I'm not sure what a better way to do it is...


gbggrant (approved):
Yeah, not wild about the formatting.

PR #1554, opened 2021-06-01
Respect genotype filtering / FT tag when calculating AC/AN/AF

RoriCremer:
### Description

A gate has been added to the getCalledChrCount() method to not count filtered out genotypes.
This fixes a bug in which AC/AN/AF were calculated without respecting if the genotypes were filtered out (and therefore shouldn't be used for calculations)
Re Laura: "It’s probably pretty rare that a user had FT fields they ignored, but if so they should unfilter rather then have some janky arg to ignore FT in annotation."

A genotype filtering unit test has been added since getCalledChrCount() changes based on filtering

- [ ] Check your code style.

### DIFF
--- a/src/test/java/htsjdk/variant/variantcontext/VariantContextUnitTest.java
+++ b/src/test/java/htsjdk/variant/variantcontext/VariantContextUnitTest.java
@@ -511,7 +511,6 @@ public class VariantContextUnitTest extends VariantBaseTest {
         List<Allele> alleles = Arrays.asList(Aref, T);
         Genotype g1 = GenotypeBuilder.create("AA", Arrays.asList(Aref, Aref));
         Genotype g2 = GenotypeBuilder.create("AT", Arrays.asList(Aref, T));
-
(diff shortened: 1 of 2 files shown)


ldgauthier (approved):
Thanks for the fix!  My only suggestion would be to add a comment to note explicitly that the first two genotypes in the test are PASS since this behavior is a little bit different for VCs versus Genotypes.

src/test/java/htsjdk/variant/variantcontext/VariantContextUnitTest.java
```diff
@@ -538,6 +537,29 @@ public void testFilters() {
         Assert.assertNotNull(vc.getFiltersMaybeNull());
     }
 
+    @Test
+    public void testGenotypeFilters() {
```
lbergelson, line 541: Could we rename this to testGenotypeCountsRespectGenotypeFilters or something along those lines?

lbergelson (changes requested):
@RoriCremer Looks good, 1 comment about clarifying the test name.  Once that's addressed 👍 to merge.

lbergelson (approved):
👍

PR #1526, opened 2020-12-23
convert Runtime exceptions in the CRAM code into CRAMExceptions (also…

yfarjoun:
… runtime)

(sorry again about the push to master)

### Description

Please explain the changes you made here.
Explain the **motivation** for making this change. What existing problem does the pull request solve?

### Things to think about before submitting:
- [ ] Make sure your changes compile and new tests pass locally.
- [ ] Add new tests or update existing ones:
  - A bug fix should include a test that previously would have failed and passes now.
  - New features should come with new tests that exercise and validate the new functionality.
- [ ] Extended the README / documentation, if necessary
- [ ] Check your code style.
- [ ] Write a clear commit title and message
  - The commit message should describe what changed and is targeted at htsjdk developers
  - Breaking changes should be mentioned in the commit message.

PR #1557, opened 2021-07-25
convert Runtime exceptions in the CRAM code into CRAMExceptions (also…

yfarjoun:
… runtime)

In the cram code the code throws an IllegalArgumentException when asked about the bases when a reference was not provided. This PR changes it to a CRAMException to be more informative and consistent.


tfenne:
Closing as stale, alongside its near-duplicate #1526. This predates the major rewrite of the CRAM code, and the CRAMReferenceRegion constructor and message it converts no longer exist on current master, so the change no longer applies. Thanks for the contribution!

PR #1559, opened 2021-08-06
Silence AsciiLineReader warning when creating a FASTA sequence index

clintval:
### Description

The `AsciiLineReader.from()` method requires either a block-compressed or positional buffered input stream. If the input stream is neither, then a warning is emitted and the stream is wrapped in a positional buffered stream:

https://github.com/samtools/htsjdk/blob/7719274fe370a51a24e6067de21bbe7e18c160a9/src/main/java/htsjdk/tribble/readers/AsciiLineReader.java#L75-L95

I would like to silence the warning produced when creating a FASTA sequence index by pre-wrapping the stream in a positional buffered stream. This will make my logs that use this method cleaner and I can avoid these warnings:

```
WARNING 2021-08-05 19:54:23     AsciiLineReader Creating an indexable source for an AsciiFeatureCodec using a stream that is neither a PositionalBufferedStream nor a BlockCompressedInputStream
```

- [ ] Check your code style.

---

I will rely on CI testing as it looks like `htsjdk` might have tests that I cannot reproduce on my local (specifically htsget tests that ping servers).

### DIFF
--- a/src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
+++ b/src/main/java/htsjdk/samtools/reference/FastaSequenceIndexCreator.java
@@ -29,10 +29,13 @@ import htsjdk.samtools.SAMSequenceRecord;
 import htsjdk.samtools.util.GZIIndex;
 import htsjdk.samtools.util.IOUtil;
 import htsjdk.tribble.readers.AsciiLineReader;
+import htsjdk.tribble.readers.PositionalBufferedStream;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.zip.GZIPInputStream;
 
 /**
  * Static methods to create an {@link FastaSequenceIndex}.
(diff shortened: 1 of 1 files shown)


clintval:
Any interest in merging this PR by the maintainers? It should be a small quality of life improvement o help me clean up my logs!

cmnbroad (commented):
One nit inline, but otherwise this seems like a reasonable change to me. The warnings are annoying, but I'd hate to just remove them since the code path that produces them can result in indexing bugs if the wrong type of stream is used. This suppresses them in a case where they're not necessary.

cmnbroad:
@lberegelson This seems reasonable to me.

clintval:
@cmnbroad all set!

clintval:
@lberegelson @cmnbroad I'm about to release some tools and would like to bundle this if possible! Could a quick last review and merge be arranged? Thank you in advance!

lbergelson (approved):
This looks fine.  I'd looked at it a while ago and had thought there was a subtle bug with BlockCompressed input streams, but I got interrupted before I could figure out if that was true and then never got back to it.  On closer inspection I don't see any problem with it and I think I was probably wrong.  Sorry for the wait.

PR #1564, opened 2021-08-10
Some long reads tests using PacBio data.

cmnbroad:
The tests are written using the new plugin API (during which discovered that we need to add a utility method that resolves an index file name for a file that doesn't yet exist, given the file's name, but that can go into a separate PR). I have some ONT data as well but its currently too large to put in the repo.


src/test/java/htsjdk/samtools/LongReadsTest.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    @Test
+    final public void testQueryOverlappingForEachRead() throws IOException {
+        final IOPath copiedBAMPath = createBAMCopyWithIndex(PAC_BIO_READS);
+
+        // for each mapped read in the original source BAM, query both the original and the copy for
+        // all reads that overlap that read's coordinates, and compare the results
+        try (final ReadsDecoder sourceDecoder =
+                     HtsDefaultRegistry.getReadsResolver().getReadsDecoder(ReadsBundle.resolveIndex(PAC_BIO_READS));
```
lbergelson, line 69: This is an aside, but ReadsBundle.resolveIndex() probably needs a bit of renaming.  Seeing it in practice makes me think it should be ReadsBundle.withResolvedIndex() or something along those lines to make it clear that it's retuning a new reads bundle.

src/test/java/htsjdk/samtools/LongReadsTest.java
```diff
@@ -0,0 +1,227 @@
...
+             final ReadsDecoder originalDecoder =
+                     HtsDefaultRegistry.getReadsResolver().getReadsDecoder(ReadsBundle.resolveIndex(PAC_BIO_READS));
+             final ReadsDecoder copyDecoder =
+                     HtsDefaultRegistry.getReadsResolver().getReadsDecoder(ReadsBundle.resolveIndex(tmpBAMPath))) {
+            for (final SAMRecord samRecord : sourceDecoder) {
+                //skip unmapped records since we're going to use the read's coordinates as a query
+                if (!samRecord.getReadUnmappedFlag()) {
+                    try (final CloseableIterator<SAMRecord> originalIt = originalDecoder.queryContained(
```
lbergelson, line 103: We should probably add an overload of queryContained that takes a Locatable so you can just pass in the SamRecord.  Is it possible to make Locatable a subinterface of HtsInterval which returns smaller primitives?

src/test/java/htsjdk/samtools/LongReadsTest.java
```diff
@@ -0,0 +1,227 @@
...
+                new ReadsEncoderOptions().setPreSorted(false))) {
+            bamEncoder.setHeader(samFileHeader);
+            for (final SAMRecord samRecord : recordsInReverseOrder) {
+                bamEncoder.write(samRecord);
+            }
+        }
+
+        // now read back in both the original and the newly sorted reads that we wrote out, and compare the order
+        try (final ReadsDecoder originalDecoder =
+                     HtsDefaultRegistry.getReadsResolver().getReadsDecoder(
```
lbergelson, line 173: I wonder if we should add methods to reduce the boiler plate.  I wish we had default values in java...

src/test/java/htsjdk/samtools/LongReadsTest.java
```diff
@@ -0,0 +1,227 @@
...
+                Assert.assertEquals(sortedIt.next().getAlignmentStart(), originalIt.next().getAlignmentStart());
+            }
+            Assert.assertFalse(sortedIt.hasNext());
+        }
+
+    }
+
+    private IOPath createBAMCopyWithIndex(final IOPath originalBAM) {
+        final IOPath tmpBAMPath = IOPathUtils.createTempPath("longReadsRoundTrip", ".bam");
+        // we need a better way to resolve the sibling - we can't use SamFiles.findIndex() because the file has
```
lbergelson, line 196: That's a good point...

lbergelson (approved):
👍 Looks good to me.  I have some comments unrelated to the tests.

src/test/java/htsjdk/samtools/LongReadsTest.java
cmnbroad: Yes, makes sense.

src/test/java/htsjdk/samtools/LongReadsTest.java
cmnbroad: The Locatable overload might be a good idea. Though we'd probably want to add parallel ones for query and queryContained. I think you'd get incompatible return types if you tried to make Locatable to extend HtsInterval though.

src/test/java/htsjdk/samtools/LongReadsTest.java
cmnbroad: Yeah we should come up with something to make this easier.

src/test/java/htsjdk/samtools/LongReadsTest.java
```diff
@@ -0,0 +1,227 @@
...
+            for (final SAMRecord samRecord : bamDecoder) {
+                bamEncoder.write(samRecord);
+            }
+        }
+        // mark the index for deletion too once its been created
+        tempIndexPath.toPath().toFile().deleteOnExit();
+        return tmpBAMPath;
+    }
+
+    private void compareQueryResults(final CloseableIterator<SAMRecord> originalIt, final CloseableIterator<SAMRecord> copyIt) {
```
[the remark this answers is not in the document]
cmnbroad: Oh, yeah....

Issue #1565, opened 2021-08-28
Unclear what the exception is referring to when htsjdk is used to validate a NIST .vcf file


namra1:
### Description of the issue:
htsjdk  errors when trying to validate a NIST .vcf file. The .vcf file is generated by NIST so I would think it has a valid format. The error message is unclear to me.

### Your environment:
* version of htsjdk
**  HTSJDK Version: 2.24.1
* version of java
**  v1.8.0_242-8u242-b08-0ubuntu3~18.04-b08
* which OS 
**  v1.8.0_242-8u242-b08-0ubuntu3~18.04-b08

### Steps to reproduce
If you're reporting a bug, tell us how to reproduce this issue. If possible, include a short **code snippet** or attach **test data** to demonstrate the problem.

gatk ValidateVariants -R hg19.fa -V NIST_NA12878_calls_in_PLDv2.vcf 

I also get an error attempting to index the .vcf file  in igv-tools in version 2.10.3.
[NIST_NA12878_calls_in_PLDv2.vcf.gz](https://github.com/samtools/htsjdk/files/7071645/NIST_NA12878_calls_in_PLDv2.vcf.gz)
[NIST_NA12878_calls_in_PLDv2.vcf.gz](https://github.com/samtools/htsjdk/files/7071648/NIST_NA12878_calls_in_PLDv2.vcf.gz)

### Expected behaviour
Tell us what should happen.
It should validate  the vcf file
### Actual behaviour

gatk ValidateVariants -R hg19.fa -V NIST_NA12878_calls_in_PLDv2.vcf 
Using GATK jar /gatk/gatk-package-4.2.2.0-local.jar
Running:
    java -Dsamjdk.use_async_io_read_samtools=false -Dsamjdk.use_async_io_write_samtools=true -Dsamjdk.use_async_io_write_tribble=false -Dsamjdk.compression_level=2 -jar /gatk/gatk-package-4.2.2.0-local.jar ValidateVariants -R hg19.fa -V NIST_NA12878_calls_in_PLDv2.vcf
22:27:12.103 INFO  NativeLibraryLoader - Loading libgkl_compression.so from jar:file:/gatk/gatk-package-4.2.2.0-local.jar!/com/intel/gkl/native/libgkl_compression.so
Aug 28, 2021 10:27:12 PM shaded.cloud_nio.com.google.auth.oauth2.ComputeEngineCredentials runningOnComputeEngine
INFO: Failed to detect whether we are running on Google Compute Engine.
[...]

PR #1566, opened 2021-09-20
Yp add context to exception

yash-puligundla:
### Description

When an invalid header line is present in a VCF file, htsjdk errors without specific details. More details [here](https://github.com/samtools/htsjdk/issues/1565)

## Fix
Adding the line number of the VCF file's invalid line to the error message.

### DIFF
--- a/src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
+++ b/src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
@@ -202,7 +202,7 @@ public abstract class VCFCompoundHeaderLine extends VCFHeaderLine implements VCF
      * @param lineType     the header line type
      *
      */
-    protected VCFCompoundHeaderLine(String line, VCFHeaderVersion version, SupportedHeaderLineType lineType) {
+    protected VCFCompoundHeaderLine(String line, VCFHeaderVersion version, SupportedHeaderLineType lineType, Integer lineNumber) {
         super(lineType.toString(), "");
 
         final ArrayList<String> expectedTags = new ArrayList(Arrays.asList("ID", "Number", "Type", "Description"));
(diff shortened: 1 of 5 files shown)


src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
```diff
@@ -240,7 +240,7 @@ protected VCFCompoundHeaderLine(String line, VCFHeaderVersion version, Supported
             throw new TribbleException(mapping.get("Type") + " is not a valid type in the VCF specification (note that types are case-sensitive)");
         }
         if (type == VCFHeaderLineType.Flag && !allowFlagValues())
-            throw new IllegalArgumentException("Flag is an unsupported type for this kind of field");
+            throw new IllegalArgumentException("Flag is an unsupported type for this kind of field at line number "+ lineNumber);
```
cmnbroad, line 243: I like the idea of reporting line numbers, but ideally we wouldn't have to propagate them to the line constructors like this. For this issue it's probably sufficient to just include the text of the actual line (which is available in `line`) in the error message.

cmnbroad (changes requested):
I suggested a somewhat simpler change (see inline comments). Perhaps at some point after Anders and I are done with the VCFHeader refactoring (this is the same code that we're refactoring on a separate branch), we might want to make a more general change to have the parser decorate this kind of exception with the current line number.

src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java
yash-puligundla: Thank you for the feedback. I made the suggested changes. Please let me know if you find anything else.

cmnbroad (approved):
Looks good Yash! Thanks.

cmnbroad:
Fixes https://github.com/samtools/htsjdk/issues/1565. @lbergelson Can you merge this one (I don't seem to be able to merge at all anymore).

Issue #1512, opened 2020-09-23
Improve error message "An index is required, but none found"


juhawilppu:
I was trying to run RenameSampleInVcf with files called
- file.vcf
- file.vcf.tbi (index)

and it was failing with error message "An index is required, but none found".

After looking at the source code, it turns out the index file needs to be named as "file.vcf.idx". I'm not sure if this was obvious to others, but the error message could be improved and mention the specific file ending.

### Your environment:
* Picard version: 2.16.0-SNAPSHOT

### Steps to reproduce
```
java -Xms3000m -jar /usr/gitc/picard.jar RenameSampleInVcf \
    INPUT=file.vcf \
    OUTPUT=output.vcf \
    NEW_SAMPLE_NAME=test
```

### Expected behaviour
Command fails with message "An index is required, but none found with file ending .idx".

### Actual behaviour
Command fails with message "An index is required, but none found".

yfarjoun:
tbi is an index for the block-gzipped version of vcf. you'll need to
generate an index off of the vcf itself using IndexFeatureFile (GATK) or
igvtools

lbergelson:
It does seem reasonable to state which index it's looking for though to prevent tbi / idx confusion.

PR #1567, opened 2021-09-21
add file format to index is required error message

yash-puligundla:
### Description

When index file is absent, htsjdk errors with "An index is required, but none found" message
More details [here](https://github.com/samtools/htsjdk/issues/1512)

## Fix
Error message specifies the expected index file format "An index is required, but none found with file ending .idx".


lbergelson (approved):
@yash-puligundla Looks good.  Congratulations on your first contribution to htsjdk!

yash-puligundla:
@lbergelson I believe I should have used the constant "TRIBBLE_INDEX" instead of hard coding ".idx" to the message. Let me know if I should do that.

PR #1568, opened 2021-09-22
remove hardcoded .idx

yash-puligundla:
@lbergelson I realized I had hardcoded the ".idx" file format in the exception message after the [Pull Request](https://github.com/samtools/htsjdk/pull/1567) was merged. 

**Description:**
When index file is absent, htsjdk errors with "An index is required, but none found" message
[here](https://github.com/samtools/htsjdk/issues/1512)

**Fix:**
Error message specifies the expected index file format "An index is required, but none found with file ending " + (index File Format).


lbergelson (approved):
@yash-puligundla Good catch, I'm a bit embarrassed that I didn't catch it...

PR #1569, opened 2021-09-27
handle utf-8 encoding

yash-puligundla:
### Description

If UTF-8 encoded characters are present in a SAM file, it is being corrupted while writing the file. This is because AsciiWriter downcasts the input `char` to a `byte`.

More details [here](https://github.com/samtools/htsjdk/issues/1202)
[SAM specification](https://samtools.github.io/hts-specs/SAMv1.pdf) specifies different field where UTF-8 encoding is allowed 

## Fix
cast from String to bytes using `str.getBytes(StandardCharsets.UTF_8)`

### DIFF
--- a/src/main/java/htsjdk/samtools/util/AsciiWriter.java
+++ b/src/main/java/htsjdk/samtools/util/AsciiWriter.java
@@ -73,7 +73,7 @@ public class AsciiWriter extends Writer {
     @Override
     public void write(final char[] chars, int off, int len) throws IOException {
         String str = new String(chars,off,len);
-        byte[] b = str.getBytes(StandardCharsets.UTF_8);//
+        byte[] b = str.getBytes(StandardCharsets.UTF_8);
         int bufferLength = b.length;
         os.write(b, 0, bufferLength);
     }
--- a/src/main/java/htsjdk/samtools/util/AsciiWriter.java
+++ b/src/main/java/htsjdk/samtools/util/AsciiWriter.java
@@ -73,7 +73,7 @@ public class AsciiWriter extends Writer {
     @Override
     public void write(final char[] chars, int off, int len) throws IOException {
         String str = new String(chars,off,len);
-        byte[] b = str.getBytes(StandardCharsets.UTF_8);
+        byte[] b = str.getBytes(StandardCharsets.UTF_8);//
         int bufferLength = b.length;
         os.write(b, 0, bufferLength);
     }
--- a/src/main/java/htsjdk/samtools/util/AsciiWriter.java
+++ b/src/main/java/htsjdk/samtools/util/AsciiWriter.java
@@ -28,6 +28,7 @@ import htsjdk.samtools.Defaults;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.Writer;
+import java.nio.charset.*;
 
 /**
  * Fast (I hope) buffered Writer that converts char to byte merely by casting, rather than charset conversion.
@@ -70,17 +71,10 @@ public class AsciiWriter extends Writer {
      * All other Writer methods vector through this, so this is the only one that must be overridden.
      */
     @Override
-    public void write(final char[] chars, int offset, int length) throws IOException {
-        while (length > 0) {
-            final int charsToConvert = Math.min(length, buffer.length - numBytes);
-            StringUtil.charsToBytes(chars, offset, charsToConvert, buffer, numBytes);
-            numBytes += charsToConvert;
-            offset += charsToConvert;
-            length -= charsToConvert;
-            if (numBytes == buffer.length) {
-                os.write(buffer, 0, numBytes);
-                numBytes = 0;
-            }
-        }
+    public void write(final char[] chars, int off, int len) throws IOException {
+        String str = new String(chars,off,len);
+        byte[] b = str.getBytes(StandardCharsets.UTF_8);
+        int bufferLength = b.length;
+        os.write(b, 0, bufferLength);
     }
 }
--- a/src/test/java/htsjdk/samtools/SAMFileRoundTripTest.java
+++ b/src/test/java/htsjdk/samtools/SAMFileRoundTripTest.java
@@ -1,0 +1,94 @@
+package htsjdk.samtools;
+
+import htsjdk.*;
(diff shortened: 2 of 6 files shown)


yash-puligundla:
pushed more commits to re-trigger Travis build. But, that didn't work
image

src/main/java/htsjdk/samtools/util/AsciiWriter.java
```diff
@@ -28,6 +28,7 @@
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.Writer;
+import java.nio.charset.*;
```
cmnbroad, line 31: By convention we try to avoid import *, unless not doing so would require a really large list. In this case, it looks like you're only using `StandardCharsets`, so you can just import that.

src/main/java/htsjdk/samtools/util/AsciiWriter.java
```diff
@@ -70,17 +71,10 @@ public void flush() throws IOException {
...
-            offset += charsToConvert;
-            length -= charsToConvert;
-            if (numBytes == buffer.length) {
-                os.write(buffer, 0, numBytes);
-                numBytes = 0;
-            }
-        }
+    public void write(final char[] chars, int off, int len) throws IOException {
+        String str = new String(chars,off,len);
+        byte[] b = str.getBytes(StandardCharsets.UTF_8);
```
cmnbroad, line 76: This is going to write all SAM content as UTF-8, but I think the spec only permits UTF-8 in a few places, and requires 7 bit ascii elsewhere. So we may need a different approach..

src/test/java/htsjdk/samtools/SAMFileRoundTripTest.java
```diff
@@ -0,0 +1,94 @@
...
+        final SAMFileWriterFactory factory = new SAMFileWriterFactory();
+        try (SamReader reader = SamReaderFactory.makeDefault().open(input);
+             SAMFileWriter writer = factory.makeSAMWriter(reader.getFileHeader(), false, new FileOutputStream(outputFile))) {
+            for (SAMRecord rec : reader) {
+                writer.addAlignment(rec);
+            }
+        }
+        catch (final Exception ex) {
+            Assert.assertTrue(ex.getMessage().contains(exceptionString));
+            throw ex;
```
cmnbroad, line 91: Its a bit hard to tell if this test is throwing on read or on write. We should replace this test with separate tests that show that we fail in both cases. Since we will presumably fail when we read a misplaced multibyte character, the test that shows that we fail on write will have to create the output programmatically, rather then via roundtrip.

src/test/java/htsjdk/samtools/SAMFileRoundTripTest.java
```diff
@@ -0,0 +1,94 @@
...
+        outputFile.delete();
+        outputFile.deleteOnExit();
+        final SAMFileWriterFactory factory = new SAMFileWriterFactory();
+        try (SamReader reader = SamReaderFactory.makeDefault().open(input);
+             SAMFileWriter writer = factory.makeSAMWriter(reader.getFileHeader(), false, new FileOutputStream(outputFile))) {
+            for (SAMRecord rec : reader) {
+                writer.addAlignment(rec);
+            }
+        }
+        catch (final Exception ex) {
```
cmnbroad, line 89: The expectedExceptions clause indicates that you expect `IllegalArgumentException` or `SAMFormatException`. So it would be preferable to narrow the scope of the catch clause to specify the narrower types that you actually expect (you can use multi-catch clause to match more than one type). If some other exception got thrown, like an IO exception, this clause will catch it and the assert will fail, even though the real problem would be completely unrelated.

lbergelson:
So looking at this more closely, it's very unclear to me why we have this class in the first place.  It seems to be entirely some sort of performance optimization that's designed to avoid the cost of converting string formats.  Is it possible that we could just remove this and replace it with a normal java Writer set to use UTF-8?

src/main/java/htsjdk/samtools/util/AsciiWriter.java
lbergelson: It seems like we're actually just doing a costly reimplementation of a standard Java Writer here now.  It seems like validation of the character ranges should be done in the SamRecord and not here.   Maybe this class has outlived it's usefulness and we can just use a standard Writer in place of it.

src/test/java/htsjdk/samtools/SAMFileRoundTripTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+        Assert.assertEquals(writtenSam, originalsam);
+    }
+
+    @DataProvider(name = "Utf8NegativeTestCases")
+    public Object[][] Utf8NegativeTestCases() {
+        return new Object[][]{
+                {"roundtrip_with_utf8_bad_1.sam", "Invalid character in read bases"},
+                {"roundtrip_with_utf8_bad_2.sam", "Non-numeric value in POS column"},
+                {"roundtrip_with_utf8_bad_2.sam", "Non-numeric value in POS column"}
```
cmnbroad, line 72: I think these tests may be failing somewhat accidentally, because theyre testing multi byte chars in fields that are already constrained to specific sets of characters (numeric for POS, or valid bases for read bases). I'm not certain that its demonstrating that the chars are rejected because they're multi-byte; but rather that its just that they don't conform to the constraints. I suspect they would still fail if you used single byte ASCII characters that didn't conform. It may be find to leave these, but I think we some need additional cases that use multi byte chars in places where there is no constraint other than that they be ASCII.

yash-puligundla:
> So looking at this more closely, it's very unclear to me why we have this class in the first place. It seems to be entirely some sort of performance optimization that's designed to avoid the cost of converting string formats. Is it possible that we could just remove this and replace it with a normal java Writer set to use UTF-8?

Just to clarify, Do you mean the class would be replaced with a normal java Writer set to use UTF-8 for certain fields and Ascii for the rest of the fields that do not permit UTF-8?

cmnbroad (changes requested):
@yash-puligundla I did a quick pass with some comments to start, let me know if you have any questions, or if anything doesn't make sense.

PR #1575, opened 2021-10-21
add predicate to GFF3Codec to give a chance to filter out some unused attributes

lindenb:
### Description

When using a GFF3Reader with DecodeDepth==DEEP, it may use a large amount of memory with attributes that will never be used ("version" ,"tag", etc...). This PR gives the GFF3Codec a chance to set a `Predicate<String>` to only keep a defined set of attributes.

the private attribute of ID_ATTRIBUTE_KEY and NAME_ATTRIBUTE_KEY Gff3BaseData was removed to check if the predicate does not remove them.

a new method `setFilterOutAttribute` was added to GFF3Codec

the `static` attribute of GFF3Codec.parseLine was removed

I added a test codecFilterOutFieldsTest

- [X] Check your code style.

### DIFF
--- a/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
+++ b/src/main/java/htsjdk/tribble/gff/Gff3BaseData.java
@@ -9,9 +9,6 @@ import java.util.List;
 import java.util.Map;
 
 public class Gff3BaseData {
-    static final String ID_ATTRIBUTE_KEY = "ID";
-    static final String NAME_ATTRIBUTE_KEY = "Name";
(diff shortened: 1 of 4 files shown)


lbergelson (changes requested):
@lindenb Thank you, this seems like a good idea.  I have a few nitpicky comments, but look good overall.

lindenb:
@lbergelson thank you for your review. I moved the `static final String` from Gff3BaseData to Gff3Constants, and replaced `setFilterOutAttribute`  by a new constructor.

PR #1578, opened 2021-11-01
ignore TC, TN on CRAM read

yash-puligundla:
### Description

As per CRAMv3, 

> TC and TN are legacy data series from CRAM 1.0. They have no function in CRAM 3.0 and should not
> be present. However, some implementations do output them and decoders must silently skip these fields.
> It is illegal for TC and TN to contain any data values, although there may be empty blocks associated
> with them.
> TC, TN data series should not be written by cram writer. However, if they are present, they should be ignored

More details [here](https://github.com/samtools/hts-specs/pull/599)

## Fix
* Remove TC, TN from CompressionHeaderEncodingMap.java
* Add TC, TN to the list of data series being ignored
* there are pre-existing tests that check that the data series that should be ignored are being ignored and are always null.


cmnbroad (changes requested):
Looks good so far, a couple of other requests:

- We really should be logging a warning whenever we see a DataSeries that we're going to silently ignore.
- Lets also add a read/write roundtrip test using the test file on which this issue was originally reported (in the specs repo) to make sure that we can still consume them. I'd guess that we already have at least some test files that contain these, but having that test would remove any doubt.

yash-puligundla:
- I have added a test to inspect the CompressionHeaderEncodingMap and assert that obsolete data series, TC and TN are not present.
- made the recommended changes to the comments and added links to the related hts-specs issues.
- DATASERIES_NOT_WRITTEN_BY_HTSJDK and DATASERIES_NOT_READ_BY_HTSJDK are defined only in StructureTestUtils.

cmnbroad:
@yash-puligundla Also, you might try wrapping both of the two data series sets (`DATASERIES_NOT_WRITTEN_BY_HTSJDK` and `DATASERIES_NOT_READ_BY_HTSJDK`) with Collections.unmodifiableSet - I think that will fix the spotBugs issue. Not sure why it only complains about one of them.

yash-puligundla:
> @yash-puligundla Also, you might try wrapping both of the two data series sets (`DATASERIES_NOT_WRITTEN_BY_HTSJDK` and `DATASERIES_NOT_READ_BY_HTSJDK`) with Collections.unmodifiableSet - I think that will fix the spotBugs issue. Not sure why it only complains about one of them.

I will do that. Thanks!

yash-puligundla:
- Added a roundtrip test
- made other suggested changes

**NOTE:**
We can't truly roundtrip and test for the absence of TC, TN as we have added code to ignore TC, TN on CRAM read. However, following @cmnbroad 's suggestion, I ran the tests -

- by removing the code changes made both for ignoring TC, TN on read and not writing TC, TN. In this case, the test would fail as TC and TN are not null. 

- by removing just the code change that ignores TC, TN and inspecting the encoding map. The encoding map of the roundtripped file does not have TC, TN in it.

src/test/java/htsjdk/samtools/CRAMComplianceTest.java
```diff
@@ -362,6 +371,81 @@ public void testBAMThroughCRAMRoundTripViaPath() throws IOException {
...
+
+                // get the CompressionHeaderEncodingMap
+                CompressionHeaderEncodingMap encodingMap = compressionHeader.getEncodingMap();
+
+                //make sure obsolete DataSeries TC, TN are not present in the CompressionHeaderEncodingMap
+                for (final DataSeries dataseries : CompressionHeaderEncodingMap.DATASERIES_NOT_READ_BY_HTSJDK) {
+                    Assert.assertNull(encodingMap.getEncodingDescriptorForDataSeries(dataseries),
+                            "Unexpected encoding key found: " + dataseries.getCanonicalName());
+                }
+            } while(containerIterator.hasNext());
```
cmnbroad, line 415: You can iterate through the containers directly by using the container iterator that you already have (and skip the counting input stream, the byte offsets, and the cram header) by simplifying this loop to:

```suggestion
        try (final InputStream cramInputStream = new BufferedInputStream(new FileInputStream(tempOutCRAM));
            final CramContainerIterator containerIterator = new CramContainerIterator(cramInputStream)) {
            while (containerIterator.hasNext()) {
                // read the container and its CompressionHeader
                final Container container = containerIterator.next();
                final CompressionHeader compressionHeader = container.getCompressionHeader();

                // get the CompressionHeaderEncodingMap
                final CompressionHeaderEncodingMap encodingMap = compressionHeader.getEncodingMap();

                //make sure obsolete DataSeries TC, TN are not present in the CompressionHeaderEncodingMap
                for (final DataSeries dataseries : CompressionHeaderEncodingMap.DATASERIES_NOT_READ_BY_HTSJDK) {
                    Assert.assertNull(encodingMap.getEncodingDescriptorForDataSeries(dataseries),
                            "Unexpected encoding key found: " + dataseries.getCanonicalName());
                }
            }
        }```

File: src/test/java/htsjdk/samtools/CRAMComplianceTest.java [hunk 2]
```diff
@@ -362,6 +371,81 @@ public void testBAMThroughCRAMRoundTripViaPath() throws IOException {
...
+        };
+    }
+
+    @Test(dataProvider = "obsoleteDataseriesCRAM", description = "round trip a CRAM file with obsolete dataseries TC and TN. TC and TN should not be present in the output CRAM file")
+    public void testRoundtripObsoleteDataseries(final File sourceFile, final File referenceFile) throws IOException {
+        final CRAMEncodingStrategy testStrategy = new CRAMEncodingStrategy();
+        final File tempOutCRAM = File.createTempFile("testRoundTrip", ".cram");
+        tempOutCRAM.deleteOnExit();
+        CRAMTestUtils.writeToCRAMWithEncodingStrategy(testStrategy, sourceFile, tempOutCRAM, referenceFile);
+        assertRoundTripFidelity(sourceFile, tempOutCRAM, referenceFile, false);
```
cmnbroad, line 395: Similar to the previous comment, this test is really just validating that we don't emit obsolete data series. Testing round trip fidelity is unrelated and can be removed.

src/test/java/htsjdk/samtools/CRAMComplianceTest.java
```diff
@@ -362,6 +371,81 @@ public void testBAMThroughCRAMRoundTripViaPath() throws IOException {
...
+                //make sure obsolete DataSeries TC, TN are not present in the CompressionHeaderEncodingMap
+                for (final DataSeries dataseries : CompressionHeaderEncodingMap.DATASERIES_NOT_READ_BY_HTSJDK) {
+                    Assert.assertNull(encodingMap.getEncodingDescriptorForDataSeries(dataseries),
+                            "Unexpected encoding key found: " + dataseries.getCanonicalName());
+                }
+            } while(containerIterator.hasNext());
+        }
+    }
+
+    public void assertRoundTripFidelity(
```
cmnbroad, line 419: This can be removed (see my other comment about keeping the test focused on the specific issue of obsolete data series).

src/test/java/htsjdk/samtools/cram/structure/CompressionHeaderEncodingMapTest.java
```diff
@@ -53,11 +58,51 @@ public void testAllCRAMEncodingsRoundTripThroughStream() throws IOException {
             final byte[] roundTrippedBytes = baos.toByteArray();
             try (final InputStream is = new ByteArrayInputStream(roundTrippedBytes)) {
                 final CompressionHeaderEncodingMap roundTippedEncodingMap = new CompressionHeaderEncodingMap(is);
-                assertExpectedHTSJDKGeneratedEncodings(roundTippedEncodingMap, Collections.EMPTY_SET);
+                assertExpectedHTSJDKGeneratedEncodings(roundTippedEncodingMap, CompressionHeaderEncodingMap.DATASERIES_NOT_READ_BY_HTSJDK);
             }
         }
     }
 
+    @Test(dataProvider = "readCramTestFiles", description = "make sure that obsolete dataseries, TC and TN are ignored on CRAM read")
```
cmnbroad, line 66: Maybe name this data provider the more specific `obsoleteDataSeriesCRAM` like you did for the other test. In fact, we usually don't create a data provider at all for a one-off test case like this. Either way is fine, but if we keep it, lets rename it.

src/test/java/htsjdk/samtools/cram/structure/CompressionHeaderEncodingMapTest.java
```diff
@@ -53,11 +58,51 @@ public void testAllCRAMEncodingsRoundTripThroughStream() throws IOException {
...
+    public void testIgnoreObsoleteDataseriesOnRead(final File sourceFile) throws IOException{
+
+        // iterate through the CRAM file
+        try (InputStream cramInputStream = new BufferedInputStream(new FileInputStream(sourceFile));
+             CountingInputStream countingInputStream = new CountingInputStream(cramInputStream);
+             CramContainerIterator containerIterator = new CramContainerIterator(countingInputStream)) {
+            CramHeader cramHeader = containerIterator.getCramHeader(); //actual cram header
+            final long containerByteOffset = countingInputStream.getCount();
+            do {
```
cmnbroad, line 75: Same comments here as for the other loop (simplify the loop to use the container iterator), and also add finals in the try clause and inside the block.

src/test/java/htsjdk/samtools/cram/structure/CompressionHeaderEncodingMapTest.java
```diff
@@ -53,11 +58,51 @@ public void testAllCRAMEncodingsRoundTripThroughStream() throws IOException {
...
+                for (final DataSeries dataseries : CompressionHeaderEncodingMap.DATASERIES_NOT_READ_BY_HTSJDK) {
+                    Assert.assertNull(encodingMap.getEncodingDescriptorForDataSeries(dataseries),
+                            "Unexpected encoding key found: " + dataseries.getCanonicalName());
+                }
+            } while(containerIterator.hasNext());
+        }
+    }
+
+    @DataProvider(name="readCramTestFiles")
+    public Object[][] readCramTestFiles() {
```
cmnbroad, line 94: Similar comment for these names -> "obsoleteDataSeriesCRAM` or something similar.

cmnbroad (changes requested):
Code looks good now, but still a few test comments.

src/test/java/htsjdk/samtools/CRAMComplianceTest.java
```diff
@@ -362,6 +370,41 @@ public void testBAMThroughCRAMRoundTripViaPath() throws IOException {
...
 
+    @Test(description = "TC and TN should not be present in the output CRAM file")
+    public void testObsoleteDataSeriesNotWritten() throws IOException {
+
+        // 1301_slice_aux.cram has legacy dataseries TC, TN present in it.
+        // TC, TN are now removed and are no longer written by htsjdk
+        // This file is used to test if CRAM reader can ignore these dataseries.
+        // 1301_slice_aux.cram is taken from hts-specs repo. It uses reference files, ce.fa and ce.fa.fai.
+        // 1301_slice_aux.cram file: https://github.com/samtools/hts-specs/blob/master/test/cram/3.0/passed/1301_slice_aux.cram
+        // Reference files: https://github.com/samtools/hts-specs/tree/master/test/cram
```
cmnbroad, line 381: I think the first, third, fourth, and last lines of this comment are not really relevant/applicable to this test, since this is testing writing (so the structure of the input file is irrelevant - any CRAM, BAM or SAM input file would have worked just as well), and the reference is embedded right in the repo now, so this can be reduced to:

```suggestion
        // TC, TN are now removed and are no longer written by htsjdk
        // 1301_slice_aux.cram is taken from hts-specs repo. It uses reference files, ce.fa and ce.fa.fai.
        // 1301_slice_aux.cram file: https://github.com/samtools/hts-specs/blob/master/test/cram/3.0/passed/1301_slice_aux.cram
```

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
```diff
@@ -148,10 +150,8 @@ public final CRAMEncodingStrategy createEncodingStrategyForParams(
         final List<DataSeries> seriesToUse = new ArrayList<>();
         // skip the ones this implementation doesn't use
         for (final DataSeries ds : DataSeries.values()) {
-            if (ds != DataSeries.TM_TestMark
-                    && ds != DataSeries.TV_TestMark
-                    && ds != DataSeries.QQ_scores
-                    && ds != DataSeries.BB_Bases) {
+            if (!StructureTestUtils.DATASERIES_NOT_WRITTEN_BY_HTSJDK.contains(ds))
+             {
```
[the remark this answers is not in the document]
cmnbroad: Open brace goes on the line with the if (or at least aligned with the if -;).

cmnbroad (changes requested):
Two very minor issues left, then we should be good to merge.

cmnbroad (approved):
Looks good now @yash-puligundla. Thanks!

cmnbroad:
@lbergelson Can you merge this one ? thanks.

lbergelson:
@yash-puligundla @cmnbroad Thank you!

Issue #1582, opened 2021-11-23
Switching to Python3


pgrt:
Hello,

I am one of the Debian maintainers of htsjdk.

You might be interested in switching to Python3 for the Python scripts in htsjdk. If so, please find the enclosed patch which can help to achieve it.

Best,

Pierre
[80-2to3.txt](https://github.com/samtools/htsjdk/files/7589586/80-2to3.txt)

lbergelson:
@pgrt Thank you, this is a good idea.

PR #1585, opened 2021-12-08
Update explain samflags script to python3

lbergelson:
Updating the scripts/explain_sam_flags.py to use python3

This patch was contributed by @pgrt in https://github.com/samtools/htsjdk/issues/1582


lbergelson:
@pgrt I've applied your patch, thank you.

PR #1592, opened 2022-02-01
Fix decoding of CRAM Scores read feature during normalization.

cmnbroad:
Fix for another issue discovered by Yash's spec compliance tests.


lbergelson (approved):
@cmnbroad This looks good to me.

PR #1599, opened 2022-03-23
Add an option to SAMFileWriter to disable checking of ordering or rec…

tfenne:
…ords when presorted=true.

### Description

This change adds a method to SAMFileWriter (not the factory) to enable/disable the code that checks that records are added in order when specifying a sort order and that the data is pre-sorted.  The reasons for doing this are two-fold:

1. When processing a queryname sorted file produced by `samtools sort -n` the difference of opinion in what constitutes `queryname` sort order causes HTSJDK to throw an exception when trying to write the file out while retaining the sort order in the header.  This is a non-trivial pain point.
2. It will increase performance (perhaps only a little)

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMFileWriter.java
+++ b/src/main/java/htsjdk/samtools/SAMFileWriter.java
@@ -43,6 +43,11 @@ public interface SAMFileWriter extends Closeable {
 	 */
 	void setProgressLogger(final ProgressLoggerInterface progress);
 
+	/** If true writers that are writing pre-sorted records should check the order during writing. */
+	default void setSortOrderChecking(final boolean check) {
+		// no-op
+	}
+
     /**
      * Must be called to flush or file will likely be defective. 
      */
--- a/src/main/java/htsjdk/samtools/SAMFileWriterImpl.java
+++ b/src/main/java/htsjdk/samtools/SAMFileWriterImpl.java
@@ -28,7 +28,7 @@ import htsjdk.samtools.util.SortingCollection;
 
 import java.io.File;
 import java.io.StringWriter;
-import java.util.function.Supplier;
+import static htsjdk.samtools.SAMFileHeader.SortOrder;
 
 /**
  * Base class for implementing SAM writer with any underlying format.
@@ -92,6 +92,22 @@ public abstract class SAMFileWriterImpl implements SAMFileWriter
         }
         this.sortOrder = sortOrder;
         this.presorted = presorted;
+        setSortOrderChecking(this.sortOrderChecker != null);
+    }
+
+    @Override
+    public void setSortOrderChecking(boolean check) {
+        final boolean doCheck = check &&
+                this.presorted &&
+                this.sortOrder != SAMFileHeader.SortOrder.unsorted &&
+                this.sortOrder != SortOrder.unknown;
+
+        if (doCheck) {
+            this.sortOrderChecker = new SAMSortOrderChecker(this.sortOrder);
+        }
+        else {
+            this.sortOrderChecker = null;
+        }
     }
 
     /**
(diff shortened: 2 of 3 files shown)


src/main/java/htsjdk/samtools/SAMFileWriterImpl.java
```diff
@@ -189,12 +204,14 @@ public void addAlignment(final SAMRecord alignment)
...
-                    + sortOrderChecker.getSortKey(prev) + "] and ["
-                    + sortOrderChecker.getSortKey(alignment) + "]");
+        if (this.sortOrderChecker != null) {
+            if (!sortOrderChecker.isSorted(alignment)) {
+                final SAMRecord prev = sortOrderChecker.getPreviousRecord();
+                throw new IllegalArgumentException("Alignments added out of order in SAMFileWriterImpl.addAlignment for " +
+                        getFilename() + ". Sort order is " + this.sortOrder + ". Offending records are at ["
+                        + sortOrderChecker.getSortKey(prev) + "] and ["
+                        + sortOrderChecker.getSortKey(alignment) + "]");
+            }
```
nh13, line 214:
```suggestion
        if (this.sortOrderChecker != null && !sortOrderChecker.isSorted(alignment)) {
            final SAMRecord prev = sortOrderChecker.getPreviousRecord();
            throw new IllegalArgumentException("Alignments added out of order in SAMFileWriterImpl.addAlignment for " +
                    getFilename() + ". Sort order is " + this.sortOrder + ". Offending records are at ["
                    + sortOrderChecker.getSortKey(prev) + "] and ["
                    + sortOrderChecker.getSortKey(alignment) + "]");
```

nh13 (approved):
LGTM, might be nice to add a little test, but lgtm

tfenne:
@nh13 Happy to add a test so long as @lbergelson and/or others are on board with this PR in general.  I looked briefly at the tests and adding one didn't seem trivially easy, so I figured I'd wait and make sure there aren't conceptual objections before spending that time.

tfenne:
Added a test.  Intent is to merge this once tests pass on CI.

src/main/java/htsjdk/samtools/SAMFileWriter.java
```diff
@@ -43,6 +43,11 @@
 	 */
 	void setProgressLogger(final ProgressLoggerInterface progress);
 
+	/** If true writers that are writing pre-sorted records should check the order during writing. */
```
lbergelson, line 46: Should this throw unsupported instead by default?

src/main/java/htsjdk/samtools/SAMFileWriterImpl.java
```diff
@@ -28,7 +28,7 @@
 
 import java.io.File;
 import java.io.StringWriter;
-import java.util.function.Supplier;
+import static htsjdk.samtools.SAMFileHeader.SortOrder;
```
lbergelson, line 31: Did you mean to statically import this?

lbergelson (commented):
This seems useful for a lot of cases, maybe it should be threaded through to the factory as an option?

src/main/java/htsjdk/samtools/SAMFileWriter.java
tfenne: Done.

tfenne:
@lbergelson I don't object to making this option available through the factory, I just found it to be rather cumbersome.  There are so many methods in there and I didn't want to take the time to figure out which of those need amending vs. not.  Is this something you feel strongly about?

lbergelson:
No, I don't feel strongly.  It would be nice but it is definitely a huge hassle.

tfenne:
Going to merge this as is then, and we can circle back and add it to the factory at some point.

Issue #766, opened 2016-12-02
Htsjdk defines queryname sorting differently than samtools which can cause failures


lbergelson:
samtools queryname sort (`samtools sort -n`) sorts readnames in a numerically aware fashion
htsjdk expects queryname sorting to be done by lexicographical comparison of readnames

This causes  reads like HXXX:999 and HXXX:1000 to sort in different ways when using picard SortSam vs samtools to sort.

This creates problems when reading and then writing files sorted with samtools.  If a vcf writer is set to assume sorted, than it will fail validation checks when outputting a file that was sorted in samtools order.

BAM spec doesn't define which order is correct, so it would be good if we could accept either.  We could potentially broaden how we interpret queryname order in order to allow either sorting.

droazen:
A big :+1: to being able to accept either order -- I think this has been a pain point for many years. @vdauwera care to comment?

nh13:
@lbergelson @droazen make sure to do a thorough search of the samtools mailing list since this has been raised many, many times ([example](https://sourceforge.net/p/samtools/mailman/samtools-devel/thread/509CF4E2-2AB8-450C-9B6A-5487B2E76B15%40sanger.ac.uk/#msg26674128)).  I would even suggest bringing back up the topic to the samtools-devel list, as folks may not see this issue.

Also, why would the read names from a SAM or BAM make an impact on a VCF writer?

vdauwera:
+1 for accepting either

droazen:
@nh13 We're aware that this has been discussed in the past -- but nothing has come of those discussions. We think this is a bug in htsjdk. If both definitions of queryname ordering are compliant with the BAM spec, then it seems to us that htsjdk should accept both when doing validation.

Note that we're not proposing standardizing on a single universal order (which is what the thread you linked to contemplates), we just want to relax the validation that is done in htsjdk to be able to accept both conventions.

nh13:
Do folks rely on a specific sort order implementation?  If so, then validating that the input follows htsjdk's qname sort order implementation sounds useful to have in the API.

droazen:
Perhaps, but by default htsjdk should not be blowing up on input that is compliant with the spec. Ie., when you tell htsjdk "please check whether the input is not actually queryname-sorted", it should not explode on input that is in fact queryname-sorted according to the spec as it does now.

lbergelson:
@nh13 vcf writer -> bam writer.  Mental failure due to sleep deprivation.

[...]

PR #1600, opened 2022-03-30
Implementation of a SAMRecord comparator that matches samtools' queryname sort order.

tfenne:
@nh13, @lbergelson I put this together mostly for fun - it's a comparator that will sort read names the same way samtools does.  I'm not really sure what the best way to integrate this is in HTSJDK though.  Thoughts?

## Summary by CodeRabbit

* **New Features**
  * Added natural sorting for SAM record read names.
  * Numeric portions of read names are compared by numeric value, including very large numbers.
  * Leading zeros do not affect numeric ordering.
  * Existing tie-breaking behavior is preserved when read names match.

* **Tests**
  * Added comprehensive coverage for mixed names, numeric ordering, prefixes, empty names, large values, and record-level comparisons.

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMRecordQueryNameNaturalComparator.java
+++ b/src/main/java/htsjdk/samtools/SAMRecordQueryNameNaturalComparator.java
@@ -1,0 +1,92 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2009 The Broad Institute
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package htsjdk.samtools;
+
+import java.io.Serializable;
+
+/**
+ * Comparator for natural "queryname" (i.e. SO:queryname SS:natural) ordering of SAMRecords.
+ */
+public class SAMRecordQueryNameNaturalComparator extends SAMRecordQueryNameComparator implements Serializable {
+    private static final long serialVersionUID = 1L;
+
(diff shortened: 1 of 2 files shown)


nh13:
Always best to be I second place: https://github.com/samtools/htsjdk/pull/1601

nh13:
Do you need to fall back on the first and second of pair flags as samtools does? https://github.com/samtools/samtools/blob/bdc5bb81c11f2ab25ea97351213cb87b33857c4d/bam_sort.c#L1798

tfenne:
@nh13 I guess it depends on whether you want to match samtools 100% or implement the "natural" sort order which doesn't prescribe how ties are broken.  I don't think I have a strong opinion so long as e.g. first of pair sorts before second of pair, and secondary/supplementary records sort after primary

tfenne:
I think the ideal way to make this work, but which would be a PITA to retrofit would be something akin to fgbio's [SamOrder](https://github.com/fulcrumgenomics/fgbio/blob/main/src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala), which defines more than the `SO` tag, and has capabilities for setting `SO`, `SS` etc. and for detecting the ordering based on more than the `SO` tag.  But we had the benefit of writing that many years later, and after the sub-sort stuff landed in the spec.

lbergelson:
Ah, the eternal problem https://github.com/samtools/htsjdk/issues/766.  I was assuming that awkwardly working around this was the impetus for the last pr?

I think we have a few options none of which are trivial.  
I like what you describe.  The subsort field was invented to fix this problem so it makes sense that we would use it.  Do you know if samtools is outputting the appropriate SS field in new versions?  We never started doing so, but maybe they're more responsible...  That would solve the problem going forward, and we could allow users to pick which variant order they want to use when writing new files.  It wouldn't solve the problem of reading older sam files.

And alternative would be to try to make some sort of meta queryname sorter which tries to figure out the subsort when it's reading a file.  That shouldn't be hard I would think, just try both sort orders and if one passes it's valid.  I can't remember if there's any issue with stateful comparators but I think it wouldn't be a problem to have one that figures out which it is and then remembers it.  

We should write the SS field in either case...

tfenne:
@lbergelson I don't think samtools writes the `SS` field either for queryname sorting.  Though we might be able to get @jmarshall or @nh13 to help with that.

I think the easiest first step would be to have HTSJDK accept either as being queryname sorted - HTSJDK already has that `SAMSortOrderChecker` which is stateful.  I don't think it would be hard to extend that to allow for either/or checking for queryname sorting.  And that way we could at least get to the point where HTSJDK (and picard etc.) would accept either samtools or HTSJDK queryname sorting (or, "natural" and "lexicographic") as valid.  It's possible that would just fix `ValidateSamFile` in Picard, or it might require a little more tweaking.

Then the question becomes ... if it can accept either as valid, is there any reason not to switch to the samtools style as the default output for `queryname`?

lbergelson:
The main reason not to switch the default immediately is that a lot of people used mixed pipeline where some of the tools are on older versions.  Switching would break them until they update all parts of the pipeline.  

Does samtools have trouble reading htsjdk queryname bams?  I'm not sure the problem is symmetrical.

tfenne:
I don't believe there are issues with samtools reading picard/htsjdk queryname sorted files, but it's not something I do regularly.  I think samtools tends more towards trusting/assuming the user provided acceptable input and not validating.

tfenne:
Revisited the comparator and made a couple of changes to align it more precisely with samtools:

- Replaced the `long`-based numeric comparison (which overflowed on very long numeric runs) with a digit-by-digit port of samtools' current `strnum_cmp`, so it matches for numeric runs of any length. The read-name comparison is also exposed as a `public static compareNatural(String, String)`.
- Aligned leading-zero handling with current samtools: equal-value numeric runs now compare equal regardless of leading zeros (`8` == `08` == `008`) rather than ordering by leading-zero count. (The SAM spec defines no formal natural-order algorithm, so samtools' `strnum_cmp` is the reference.)
- Expanded the test coverage accordingly.

Fuller integration with the SAM `SS` (sub-sort) tag — modeling `SS` in the header and automatically selecting this comparator for `SS:queryname:natural` — is out of scope for this PR and can be tackled separately.

PR #1601, opened 2022-03-30
Change queryname sort to match samtools

nh13:
This PR is meant to start a discussion to standardize the queryname sort order between samtools and htsjdk.  It is a constant source of confusion, and sometimes problems, especially when trying to use the output of multi-threaded `samtools sort -n` going into picard/gatk/htsjdk.


lbergelson:
I'm in favor of fixing this embarrassing and long standing problem.   This would be a compatibility nightmare though as I'm sure you're aware... Should we consolidate this discussion in Tim's PR?

nh13:
Closing in favor of https://github.com/samtools/htsjdk/pull/1600

PR #1603, opened 2022-04-11
Remove validateArg call with a high invocation count when writing CRAM.

cmnbroad:
The containing method gets called frequently, and the validateArg call shows up as a hotspot (3-4%) when profiling writing CRAM .


lbergelson (approved):
@cmnbroad I'm all for eliminating hotspots, but does this open us up to problems?   If we inline the check instead of calling validateArg would that keep the safety but reduce the performance problem?

cmnbroad:
We're integrating this as part of Yash's codecs PRs.

PR #1609, opened 2022-05-03
Add support for re-serializing (sharding and reassembling) CRAM containers.

### LINKED ISSUE #1606: Add support for reserializing CRAM containers
There are cases where we want to be able to re-serialize CRAM containers into a new stream without fully decoding and then re-encoding them (and without the need for the reference). For example, if we want to stitch together CRAM shards that were created in parallel (i.e, as Disq does), or if want to split a CRAM up into smaller CRAMs preserving container boundaries. The main thing required is a means to update the stream-relative values that are in the container and slice headers, specifically the global record counter.

cmnbroad:
Fixes https://github.com/samtools/htsjdk/issues/1606.

### DIFF
--- a/src/main/java/htsjdk/samtools/CRAMContainerStreamRewriter.java
+++ b/src/main/java/htsjdk/samtools/CRAMContainerStreamRewriter.java
@@ -1,0 +1,92 @@
+package htsjdk.samtools;
+
+import htsjdk.samtools.cram.build.CramIO;
+import htsjdk.samtools.cram.structure.Container;
+import htsjdk.samtools.cram.structure.CramHeader;
+import htsjdk.samtools.util.RuntimeIOException;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+/**
+ * Rewrite a series of containers to a new stream. The CRAM header and SAMFileHeader containers are automatically
+ * written to the stream when this class is instantiated. An EOF container is automatically written when
+ * {@link #finish()} is called.
+ */
+public class CRAMContainerStreamRewriter {
+    private final OutputStream outputStream;
+    private final String outputStreamIdentifier;
+    private final CramHeader cramHeader;
+    private final SAMFileHeader samFileHeader;
+    private final CRAMIndexer cramIndexer;
+
+    private long streamOffset = 0L;
+    private long recordCounter = 0L;
+
+    /**
+     * Create a CRAMContainerStreamRewriter for writing a series of CRAM containers into an output
+     * stream, with an optional output index.
+     *
+     * @param outputStream where to write the CRAM stream.
+     * @param samFileHeader {@link SAMFileHeader} to be used. Sort order is determined by the sortOrder property of this arg.
+     * @param outputStreamIdentifier used for display in error message display
+     * @param indexer CRAM indexer. Can be null if no index is required.
+     */
+    public CRAMContainerStreamRewriter(
+            final OutputStream outputStream,
+            final CramHeader cramHeader,
+            final SAMFileHeader samFileHeader,
+            final String outputStreamIdentifier,
+            final CRAMIndexer indexer) {
+        this.outputStream = outputStream;
+        this.cramHeader = cramHeader;
+        this.samFileHeader = samFileHeader;
+        this.outputStreamIdentifier = outputStreamIdentifier;
+        this.cramIndexer = indexer;
+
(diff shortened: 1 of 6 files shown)


src/main/java/htsjdk/samtools/CRAMContainerStreamRewriter.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Rewrite a series of containers to a new stream. The CRAM header and SAMFileHeader containers are automatically
+ * written to the stream when this class is instantiated. An EOF container is automatically written when
+ * {@link #finish()} is called.
+ */
+public class CRAMContainerStreamRewriter {
+    private final OutputStream outputStream;
+    private final String outputStreamIdentifier;
+    private final CramHeader cramHeader;
+    private final SAMFileHeader samFileHeader;
```
lbergelson, line 20: This field is assigned but never accessed.

src/main/java/htsjdk/samtools/CRAMContainerStreamRewriter.java
```diff
@@ -0,0 +1,92 @@
...
+     */
+    public void finish() {
+        try {
+            CramIO.writeCramEOF(cramHeader.getCRAMVersion(), outputStream);
+            outputStream.flush();
+            if (cramIndexer != null) {
+                cramIndexer.finish();
+            }
+        } catch (final IOException e) {
+            throw new RuntimeIOException(String.format("IOException closing stream for %s", outputStreamIdentifier));
```
lbergelson, line 88: Is the output stream actually closed here?  It seems like it's just flushed.

src/main/java/htsjdk/samtools/cram/structure/Container.java
```diff
@@ -335,6 +336,28 @@ public List<SAMRecord> getSAMRecords(
...
+     * Update the stream-relative values (global record counter and stream byte offset) for this
+     * container. For use when re-serializing a container that has been read from an existing stream
+     * into a new stream. This method mutates the container and it's slices - the container is no
+     * longer valid in the context of it's original stream.
+     *
+     * @param containerRecordCounter the new global record counter for this container
+     * @param streamByteOffset the new stream byte offset counter for this container
+     * @return the updated global record counter
+     */
+    public long relocateContainer(final long containerRecordCounter, final long streamByteOffset) {
```
lbergelson, line 350: Is it necessary to mutate the container? It was previously effectively immutable wasn't it?  It seems like it would be nicer to copy it since I assume the container / slice objects are fairly lightweight and you could just pass through the reference to the heavy data?

Not a blocker if you think this is necessary for efficiency reasons / copying a container being very annoying.

src/main/java/htsjdk/samtools/cram/structure/ContainerHeader.java
```diff
@@ -249,4 +249,8 @@ public boolean isEOF() {
         return v3 || v2;
     }
 
+    void setGlobalRecordCounter(final long recordCounter) {
+        this.globalRecordCounter = recordCount;
```
lbergelson, line 253: again, unfortunate to make this mutable.

src/main/java/htsjdk/samtools/cram/structure/Container.java
```diff
@@ -190,6 +190,7 @@ public int write(final CRAMVersion cramVersion, final OutputStream outputStream)
                 // landmark 0 = byte length of the compression header
                 // landmarks after 0 = byte length of the compression header plus all slices before this one
                 landmarks.add(tempOutputStream.size());
+                slice.byteOffsetOfContainer = containerByteOffset;
```
lbergelson, line 193: Why is this updated again here when it's already set by relocate?

src/main/java/htsjdk/samtools/cram/structure/Slice.java
```diff
@@ -78,7 +78,7 @@ public class Slice {
 
     private final CompressionHeader compressionHeader;
     private final SliceBlocks sliceBlocks;
-    private final long byteOffsetOfContainer;
+    public long byteOffsetOfContainer;
```
lbergelson, line 81: Why make this public?  It seems like it's already updated through relocate.

lbergelson (changes requested):
@cmnbroad Questions / comments about mutability.  If this is the only sane way to do this I can accept that it adds some more state.

PR #1612, opened 2022-06-01
Require sequence dictionary MD5s when writing CRAM.

### LINKED ISSUE #506: SAMFileHeader in CRAM files is missing attributes required by the CRAM spec

### LINKED ISSUE #5195: CRAMs generated by GATK don't always have M5/UR attributes in the sequence dictionary
See https://github.com/samtools/htslib/issues/769#issuecomment-422041279. These are requried by the CRAM spec, but htsjdk-generated files will only have them if they appear in the input/source dictionary from which the CRAM was generated. The fix should be in htsjdk see https://github.com/samtools/htsjdk/issues/506.

cmnbroad:
Fixes https://github.com/samtools/htsjdk/issues/506 (and by extension will also fix https://github.com/broadinstitute/gatk/issues/5195). If the SAM header used to create a CRAM doesn't have a sequence dictionary with MD5s, the code will attempt to use consult the reference's dictionary to inject them if possible. Otherwise, the write operation will fail with an error message.


src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -148,6 +148,23 @@ public SAMSequenceRecord setMd5(final String value) {
         return this;
     }
 
+    /**
+     * Set the M5 attribute for this sequence using the provided sequenceBases.
```
lbergelson, line 152: typo: M5 -> MD5

src/main/java/htsjdk/samtools/SAMSequenceRecord.java
```diff
@@ -148,6 +148,23 @@ public SAMSequenceRecord setMd5(final String value) {
...
+     * @param sequenceBases sequence bases to use to compute the MD5 for this sequence
+     * @return the update sequence record
+     */
+    public SAMSequenceRecord setComputedMd5(final byte[] sequenceBases) {
+        try {
+            final MessageDigest md5Digester = MessageDigest.getInstance("MD5");
+            if (md5Digester != null) {
```
lbergelson, line 159: This method seems broken.  It doesn't actually digest the bases. Also, the bases should be normalized.

Why not use SequenceUtil.calculateMD5String()?

src/main/java/htsjdk/samtools/SAMSequenceDictionary.java
```diff
@@ -294,6 +301,99 @@ public String md5() {
...
+        final List<SAMSequenceRecord> missingMD5s = getSequencesWithMissingMD5s();
+        if (!missingMD5s.isEmpty()) {
+            throw new RuntimeException(SAMSequenceDictionary.createFormattedMD5Message(contextMessage, missingMD5s));
+        }
+    }
+
+    /**
+     * @return all sequences in this dictionary that do not have an MD5 attribute
+     */
+    List<SAMSequenceRecord> getSequencesWithMissingMD5s() {
```
lbergelson, line 365: These seem like very specific methods.  It's a bit funny to have them here as package private.  I wonder if it makes more sense to move them to the callsite.

lbergelson (changes requested):
@cmnbroad I wasn't sure if you re-marked it as draft by accident or on purpose.  It looks like it needs some touching up though.  There's a bunch of stuff that I think should be changed though.  Sorry if you're already aware / working on that.

Issue #1615, opened 2022-06-27
EdgeReadIterator (used in Picard's CollectWgsMetrics) doesn't handle intervals correctly


michaelgatzen:
When running Picard's CollectWgsMetrics with an IntervalList, only the start and end coordinates of the first interval in that list are considered. This can be traced back to a bug in EdgeReadIterator and AbstractLocusInfo.

michaelgatzen:
I can't assign this to myself because I'm not a contributor, but I'm working on this issue.

lbergelson:
closed by #1616

PR #1616, opened 2022-06-30
Fix EdgeReadIterator

michaelgatzen:
- Fixing #1615 
  - EdgeReadIterator was calling AbstractLocusIterator's `getCurrentInterval()` to determine the read overlaps. However, the `lastInterval` that it uses internally was never updated. This PR removes the method `getCurrentInterval()` and instead implements this functionality in `EdgeReadIterator`, which is the only place this has ever been called from (at least in Picard and GATK). **This is arguably a breaking API change**, although I would argue that this method has never worked and should not be placed in AbstractLocusIterator, but instead in an iterator implementation that actually does keep track of the intervals.
  - Keeping track of the intervals has now been moved to AbstractLocusIterator
  - Added a new function that determines if a read is fully contained by the current interval, in that case we don't need to determine the overlaps
  - Renamed a few variables to better describe their use
- Added tests to catch previously faulty behavior
  - Also extended existing tests with the possibility to define custom headers
- Fixed another bug in AbstractLocusIterator and IntervalListReferenceSequenceMask, where the documentation says that it would sort an unsorted input IntervalList, but it never does

### DIFF
--- a/src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
+++ b/src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
@@ -142,11 +142,6 @@ public abstract class AbstractLocusIterator<T extends AbstractRecordAndOffset, K
 
     private final LocusComparator<Locus> locusComparator = new LocusComparator<>();
 
-    /**
-     * Last processed interval, relevant only if list of intervals is defined.
-     */
-    private int lastInterval = 0;
-
 
 
     public SAMFileHeader getHeader() {
@@ -159,7 +154,7 @@ public abstract class AbstractLocusIterator<T extends AbstractRecordAndOffset, K
      *
      * @param samReader    must be coordinate sorted
      * @param intervalList Either the list of desired intervals, or null.  Note that if an intervalList is
-     *                     passed in that is not coordinate sorted, it will eventually be coordinated sorted by this class.
+     *                     passed in that is not coordinate sorted, it will eventually be coordinated sorted here.
      * @param useIndex     If true, do indexed lookup to improve performance.  Not relevant if intervalList == null.
      *                     It is no longer the case the useIndex==true can make performance worse.  It should always perform at least
      *                     as well as useIndex==false, and generally will be much faster.
@@ -176,8 +171,14 @@ public abstract class AbstractLocusIterator<T extends AbstractRecordAndOffset, K
         this.samReader = samReader;
         this.useIndex = useIndex;
         if (intervalList != null) {
-            intervals = intervalList.uniqued().getIntervals();
-            this.referenceSequenceMask = new IntervalListReferenceSequenceMask(intervalList);
+            if (!intervalList.getHeader().getSequenceDictionary().isSameDictionary(getHeader().getSequenceDictionary())) {
+                throw new SAMException("The sequence dictionary of the interval list file differs from the sequence dictionary of the input SAM file.");
+            }
+
+            final boolean isSorted = intervalList.getHeader().getSortOrder() == SAMFileHeader.SortOrder.coordinate;
+            final IntervalList sortedIntervalList = isSorted ? intervalList : intervalList.sorted();
+            intervals = sortedIntervalList.uniqued().getIntervals();
+            this.referenceSequenceMask = new IntervalListReferenceSequenceMask(sortedIntervalList);
         } else {
             intervals = null;
             this.referenceSequenceMask = new WholeGenomeReferenceSequenceMask(samReader.getFileHeader());
@@ -574,11 +575,6 @@ public abstract class AbstractLocusIterator<T extends AbstractRecordAndOffset, K
 
     protected List<Interval> getIntervals() {
         return intervals;
-    }
-
-    protected Interval getCurrentInterval() {
-        if (intervals == null) return null;
-        return intervals.get(lastInterval);
     }
 
     public boolean isIncludeIndels() {
--- a/src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
+++ b/src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
@@ -175,8 +175,8 @@ public abstract class AbstractLocusIterator<T extends AbstractRecordAndOffset, K
                 throw new SAMException("The sequence dictionary of the interval list file differs from the sequence dictionary of the input SAM file.");
             }
 
-            final boolean isSorted = intervalList.getHeader().getSortOrder() == SAMFileHeader.SortOrder.coordinate;
-            final IntervalList sortedIntervalList = isSorted ? intervalList : intervalList.sorted();
+            final boolean intervalListIsSorted = intervalList.getHeader().getSortOrder() == SAMFileHeader.SortOrder.coordinate;
+            final IntervalList sortedIntervalList = intervalListIsSorted ? intervalList : intervalList.sorted();
             intervals = sortedIntervalList.uniqued().getIntervals();
             this.referenceSequenceMask = new IntervalListReferenceSequenceMask(sortedIntervalList);
         } else {
@@ -265,6 +265,9 @@ public abstract class AbstractLocusIterator<T extends AbstractRecordAndOffset, K
      */
     @Override
     public K next() {
+        if (this.samIterator == null) {
+            iterator();
+        }
         // if we don't have any completed entries to return, try and make some!
         while (complete.isEmpty() && samHasMore()) {
             final SAMRecord rec = samIterator.peek();
--- a/src/main/java/htsjdk/samtools/util/EdgeReadIterator.java
+++ b/src/main/java/htsjdk/samtools/util/EdgeReadIterator.java
@@ -24,10 +24,9 @@
 
 package htsjdk.samtools.util;
 
-import htsjdk.samtools.AlignmentBlock;
-import htsjdk.samtools.SAMRecord;
-import htsjdk.samtools.SAMSequenceRecord;
-import htsjdk.samtools.SamReader;
+import htsjdk.samtools.*;
+
+import java.util.Iterator;
 
 /**
  * Iterator that traverses a SAM File, accumulating information on a per-locus basis.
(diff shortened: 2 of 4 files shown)


src/main/java/htsjdk/samtools/util/EdgeReadIterator.java
```diff
@@ -88,73 +139,86 @@ public EdgeReadIterator(final SamReader samReader, final IntervalList intervalLi
...
+            // Ensure there are consecutive AbstractLocusInfos up to and including the end of the AlignmentBlock
+            for (int locusPos = accumulatorNextPosition; locusPos <= referencePositionStartOfAlignmentBlock + alignmentBlock.getLength(); ++locusPos) {
                 accumulator.add(createLocusInfo(getReferenceSequence(rec.getReferenceIndex()), locusPos));
             }
 
-            /* Let's assume an alignment block starts in some locus. 
-             * We put two records to the accumulator. The first one has the "begin" type which corresponds to the locus 
-             * where the block starts. The second one has the "end" type which corresponds to the other locus where the block ends. 
-            */
+            // Let's assume an alignment block starts in some locus.
```
lbergelson, line 173: This is all pretty confusing.  It could really use a nice ASCII art if you're feeling creative.

src/main/java/htsjdk/samtools/util/IntervalListReferenceSequenceMask.java
```diff
@@ -46,10 +46,9 @@ public class IntervalListReferenceSequenceMask implements ReferenceSequenceMask
 
     public IntervalListReferenceSequenceMask(final IntervalList intervalList) {
         this.header = intervalList.getHeader();
-        if (intervalList.getHeader().getSortOrder() != SAMFileHeader.SortOrder.coordinate) {
-            intervalList.sorted();
-        }
-        final List<Interval> uniqueIntervals = intervalList.uniqued().getIntervals();
+        final boolean isSorted = intervalList.getHeader().getSortOrder() == SAMFileHeader.SortOrder.coordinate;
```
lbergelson, line 49: good catch here

src/test/java/htsjdk/samtools/util/EdgeReadIteratorTest.java
```diff
@@ -398,14 +402,124 @@ public void testIntersectingIntervalWithComplicatedCigar() {
...
+                new Interval("chr2", 20, 24), // startOutsideChr2: first alignment block
+                new Interval("chr2", 35, 44), // startOutsideChr2: second alignment block
+                new Interval("chr2", 65, 70), // endOutsideChr2: first alignment block (second isn't covered at all)
+                new Interval("chr4", 20, 30), // spanningThreeIntervalsChr4: first interval
+                new Interval("chr4", 40, 50), // spanningThreeIntervalsChr4: second interval
+                new Interval("chr4", 60, 70), // spanningThreeIntervalsChr4: third interval
+        };
+
+        EdgeReadIterator iterator = new EdgeReadIterator(builder.getSamReader(), intervals);
+        iterator.iterator(); // We need this call to initialize the samReader. Usually a call to hasNext() would do that, which we don't need here.
```
lbergelson, line 444: This is a bug here... next() should be valid without calling iterator() or hasNext().  Maybe we should fix that in AbstractLocusIterator.  I think there's a different bug where calling next() returns null instead of throwing NoSuchElementException.

lbergelson (changes requested):
@michaelgatzen Thanks for taking on this unenviable task.  This set of of classes is quite confusing.  I think I found a subtle (and in practice probably not important) bug with using the IntervalComparator if the intervals happen to have names or strands defined.  I have a few stylistic comments too.  I agree the breaking change is fine since it's so broken to begin with.

src/main/java/htsjdk/samtools/util/AbstractLocusIterator.java
```diff
@@ -176,8 +171,14 @@ public AbstractLocusIterator(final SamReader samReader, final IntervalList inter
         this.samReader = samReader;
         this.useIndex = useIndex;
         if (intervalList != null) {
-            intervals = intervalList.uniqued().getIntervals();
-            this.referenceSequenceMask = new IntervalListReferenceSequenceMask(intervalList);
+            if (!intervalList.getHeader().getSequenceDictionary().isSameDictionary(getHeader().getSequenceDictionary())) {
+                throw new SAMException("The sequence dictionary of the interval list file differs from the sequence dictionary of the input SAM file.");
+            }
+
+            final boolean isSorted = intervalList.getHeader().getSortOrder() == SAMFileHeader.SortOrder.coordinate;
```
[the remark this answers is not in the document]
michaelgatzen: Yes good question. First of all, yes, we need both the SAM file and the interval list file to be coordinate sorted. AbstractLocusIterator logs a warning if the sort order is unspecified, and throws an exception if it is specified but not coordinate. What I'm doing here is checking if the interval list is coordinate sorted, because if it is, we might as well skip the sorting. Or do you think we should sort either way?

src/main/java/htsjdk/samtools/util/EdgeReadIterator.java
```diff
@@ -69,13 +75,58 @@ public EdgeReadIterator(final SamReader samReader, final IntervalList intervalLi
...
+     */
+     protected boolean intervalCompletelyContainsRead(final SAMRecord rec) {
+         // currentInterval should never be null when calling this method, but we have to check it just to make sure,
+         // so that we don't get a NullPointerException in the return statement.
+         if (currentInterval == null) {
+             return false;
+         }
+        // Here we need to update the currentInterval. We have to do this using an
+        // IntervalCoordinateComparator to take factor in the order in the sequence dictionary.
+        while (nextInterval != null && intervalCoordinateComparator.compare(new Interval(rec), nextInterval) > 0) {
```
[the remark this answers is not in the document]
michaelgatzen: Another good point (and good catch!). I'm using the IntervalCoordinateComparator mostly for convenience, because it takes into account the sort order of the contigs in the sequence dictionary, because they might not be alphabetically sorted. And you're right, it could happen that having a strand or name specified could affect the comparison, but that happens only after looking at start and end coordinates. So this would only be an issue if the current read happens to have the same start and end coordinates as the interval (which would be very rare), and in such cases I think we can afford that we get this wrong, because worst case we will process a few reads with the slower implementation.

src/test/java/htsjdk/samtools/util/EdgeReadIteratorTest.java
michaelgatzen: Fixed in AbstractLocusIterator

src/main/java/htsjdk/samtools/util/EdgeReadIterator.java
michaelgatzen: ```\o/```

src/main/java/htsjdk/samtools/util/EdgeReadIterator.java
```diff
@@ -69,13 +75,58 @@ public EdgeReadIterator(final SamReader samReader, final IntervalList intervalLi
...
+        }
+    }
+
+    /**
+     * This function updates currentInterval and nextInterval according to the position of the record that it is
+     * presented and determines if the current read is fully contained in the currentInterval.
+     * @param rec The record we want to consider
+     * @return True, if rec is fully contained in the current interval, otherwise false
+     */
+     protected boolean intervalCompletelyContainsRead(final SAMRecord rec) {
```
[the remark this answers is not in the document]
michaelgatzen: I can't decide what's better. Clearly for readability it'd be better to advance the iterator and then check if it contains the read, but if I have it in one function, the calling line is quite elegant:
```
final boolean needToConsiderIntervals = intervals != null && !advanceCurrentIntervalAndCheckIfIntervalContainsRead(rec);
```

Here, the function is only called if `intervals != null` and the assignment can be done to a final variable. Or can you think of another way to accomplish this, without unnecessary function calls (for when `intervals == null`) or not having a final variable?

src/main/java/htsjdk/samtools/util/EdgeReadIterator.java
```diff
@@ -69,13 +75,58 @@ public EdgeReadIterator(final SamReader samReader, final IntervalList intervalLi
...
-     *                     passed in that is not coordinate sorted, it will eventually be coordinated sorted by this class.
+     *                     passed in that is not coordinate sorted, it will be coordinated sorted here.
      * @param useIndex     If true, do indexed lookup to improve performance.  Not relevant if intervalList == null.
      *                     It is no longer the case the useIndex==true can make performance worse.  It should always perform at least
      *                     as well as useIndex==false, and generally will be much faster.
      */
     public EdgeReadIterator(final SamReader samReader, final IntervalList intervalList, final boolean useIndex) {
-        super(samReader, intervalList, useIndex);
+        super(samReader,
+                intervalList == null || intervalList.getHeader().getSortOrder() == SAMFileHeader.SortOrder.coordinate ? intervalList : intervalList.sorted(),
```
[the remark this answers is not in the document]
michaelgatzen: Done

src/main/java/htsjdk/samtools/util/EdgeReadIterator.java
```diff
@@ -69,13 +75,58 @@ public EdgeReadIterator(final SamReader samReader, final IntervalList intervalLi
...
+                useIndex);
+
+        if (getIntervals() == null) {
+            intervalListIterator = null;
+            intervalCoordinateComparator = null;
+            overlapDetector = null;
+        } else {
+            // For the easy case of a read being fully contained in an interval, we use an iterator to keep track
+            // of the current interval.
+            intervalListIterator = getIntervals().iterator();
```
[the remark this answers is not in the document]
michaelgatzen: Done

michaelgatzen:
@lbergelson Thanks for the review, and all the deep digging you did for that! I addressed most of the comments, feel free to comment or resolve the threads!

src/main/java/htsjdk/samtools/util/EdgeReadIterator.java
```diff
@@ -142,10 +133,33 @@ void accumulateSamRecord(SAMRecord rec) {
...
+        final boolean needToConsiderIntervals = intervals != null && !advanceCurrentIntervalAndCheckIfIntervalContainsRead(rec);
 
         // interpret the CIGAR string and add the base info
         for (final AlignmentBlock alignmentBlock : rec.getAlignmentBlocks()) {
+            // General concept
+            //
+            // Example: Read (or more accurately, AlignmentBlock) from position 101 to 108
+            //
+            // accumulator (ArrayList<LocusInfo>)     30  31  32  33  34  35  36  37  38 (has one LocusInfo at each position, corresponding to the genomic position)
+            // LocusInfo objects (with genomic pos.) 100 101 102 103 104 105 106 107 108 (the LocusInfo objects can contain EdgingRecordAndOffset objects, if a record starts or ends there)
```
lbergelson, line 145: Thanks for this.   I think it's helpful for future viewers.

lbergelson (approved):
@michaelgatzen I think this is fine to merge.  If we encounter problems with too strict sequence dictionary checking we can revisit it.

PR #1619, opened 2022-09-14
Added ULTIMA and ELEMENT as valid value for RG-PL according to SAM spec.

rickymagner:
### Description

The [SAM Spec](https://samtools.github.io/hts-specs/SAMv1.pdf) allows for two additional values for PL inside Read Groups than currently implemented. I added both (`ELEMENT` and `ULTIMA`), in particular so the Ultima pipelines can be updated to use this for their read group values. I also cleaned up the formatting a bit to help readability of the code / comments.


lbergelson (approved):
👍

PR #1624, opened 2022-10-05
Minor improvements to AbstractLocusIterator

lbergelson:
* Use SequenceUtil to compare dictionaries
* Remove redundant sorting

@michaelgatzen Could you take a look at this? It turns out that IntervalList.uniqued() also sorts so I removed the other sorting, no good way to avoid it entirely.
I also switched the dictionary check to a more standard method that I hadn't known about, but it's now consistent with how other dictionary matching in CollectWgsMetrics is done.


michaelgatzen (approved):
If the tests pass then sure, that works

PR #1593, opened 2022-02-14
Use allele info in VariantContext comparisons for stable sorts

clintval:
TLDR: This PR will make coordinate VCF sorting in HTSJDK/Picard similar to bcftools.

I tried to use Picard's [SortVcf](https://gatk.broadinstitute.org/hc/en-us/articles/360036858331-SortVcf-Picard) for a VCF comparison task, but I had difficulty because it's output was non-stable based on the full primary information of a variant context (contig, start, and alleles). I finally chose the [`bcftools`](https://samtools.github.io/bcftools/bcftools.html#sort) implementation because it [has a slightly more-stable sorting method](https://github.com/samtools/bcftools/blob/dd937aa89b3998aa4c6fd15375206e19c22d3f20/vcfsort.c#L95-L117). However, I want to ensure future users of Picard don't hit the same issue I did, hence this PR.

Before I write unit tests, is this something you are willing to accept?

PR #1629, opened 2022-10-28
Refactoring VariantContextComparator

lbergelson:
* Responding to my own pr comments on #1593
* doing some refactoring to use the Comparator.comparing API
* adding a few more tests

@clintval I might have gone a bit overboard on refactoring.  I really hate how normal comparators are written and like to rewrite them as the newer Comparator.comparing() chains so I did that. 

It's great that you added tests, I ended up condensing them into a single test case with a dataprovider and a few additional use cases.  I never would have bothered having any if you hadn't mad them in the first place though!

### DIFF
--- a/src/main/java/htsjdk/variant/variantcontext/VariantContextComparator.java
+++ b/src/main/java/htsjdk/variant/variantcontext/VariantContextComparator.java
@@ -11,19 +11,22 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeSet;
 
 /**
  * A Comparator that orders VariantContexts by the ordering of the contigs/chromosomes in the List
- * provided at construction time, then by start position with each contig/chromosome.
+ * provided at construction time, then by start position with each contig/chromosome, then by Ref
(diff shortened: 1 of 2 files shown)


lbergelson:
@cmnbroad Could you take a look at this, it ended up being more new code than I meant it to be so it probably needs a second pair of eyes.

lbergelson:
blocked because of https://github.com/samtools/htsjdk/issues/1637

src/main/java/htsjdk/variant/variantcontext/VariantContextComparator.java
```diff
@@ -83,20 +97,24 @@ public int compare(final VariantContext firstVariantContext, final VariantContex
...
-				if (contigCompare != 0) return contigCompare;
+		return internalComparator.compare(firstVariantContext, secondVariantContext);
+	}
+
+	//Sorts and uniques the inputs and then compares the values pairwise in their natural order
+	//until there is a mismatch or 1 of the inputs runs out.
+	private static class SortedUniquedElementsComparator<T extends Comparable<T>> implements Comparator<Collection<T>>{
+		@Override
+		public int compare(final Collection<T> o1, final Collection<T> o2) {
+			final Iterator<T> left = new TreeSet<>(o1).iterator();
```
vruano, line 108: Perhaps is worth to add trivial cases to avoid TreeSet constructons since quite often alt-alleles list are just 1 in length?

PR #1634, opened 2022-11-16
Allow the BAM index to be up to 5 seconds older than the BAM before e…

tfenne:
…mitting the "WARNING: BAM index file ... is older than BAM " message.

### Description

HTSJDK itself when writing a BAM and indexing it will frequently leave an mtime on the index that is slightly older than the BAM itself, and this causes a lot of spurious warnings.  This is a simple attempt to fix those.

- [x] Check your code style.


lbergelson (approved):
Good idea.  This should help some of the cloud instability issues.

PR #1636, opened 2022-11-16
Make AbstractVCFCodec thread-safe, allowing for multi-threaded VariantContext genotype decoding

### LINKED ISSUE #1026: Strange bug in VCFEncoder.addGenotypeData  when reading a list of Variant and then using the list in a multithreaded env.
### Subject of the issue

Genotypes in variantcontext look mutable (??)

## Minimal example

I've posted an example here: https://gist.github.com/lindenb/7ee76e30dfb662048fff3cf6125b980d

## Description

This is a strange bug, may be it's not about htsjdk but I would be happy to know the cause of this.

I'm reading a list of variants from a VCFFileReader:

```
final CloseableIterator<VariantContext> iter2=vcfFileReader.iterator();
final List<VariantContext> inMemoryVariants =  Collections.unmodifiableList(iter2.stream().
		limit(3).
		collect(Collectors.toList())
		);
iter2.close();
vcfFileReader.close();
```
this **read-only** List is then used by some java **Threads** , I wait for the end of the thread and I write some variants of a new VCF file:

```
final VariantContextWriter w = (...)
final VCFHeader header2= (...)
w.writeHeader(header2);
w.add(bestResult.ctx1);
w.add(bestResult.ctx2);
w.close();
```

but here, I get an exception form **VCFEncoder.writeAllele** called from **addGenotypeData** : 

```
java.lang.RuntimeException: Allele C* is not an allele in the variant context
	at htsjdk.variant.vcf.VCFEncoder.writeAllele

```

at the beginning of **VCFEncoder.addGenotypeData** if I add the  statement below, I see my message on stderr: ` boummm3 {C=1, .=., T*=0} [C*, C*] class java.util.HashMap`

```
final List<Allele> saveAllele = Collections.unmodifiableList(g.getAlleles());
if(!alleleMap.containsKey(g.getAllele(0))) {
                    System.err.println(
                            " boummm3 "+alleleMap+" "+saveAllele+" "+alleleMap.getClass());
                  
                }

```

Adding  the following statement (accessing the alleles) before using the list inMemoryVariants fixes the problem.

```
inMemoryVariants.forEach(C->{long i=C.getGenotypes().stream().flatMap(G->G.getAlleles().stream()).count();});`
```

I'm puzlled, do you have any idea ?

### Your environment
* version of htsjdk 2.13.0
* version of java  1_8_152
* which OS linux 32

rcoleb:
### Description

**Problem:** Processing multiple `VariantContext` objects concurrently results in shared data due to the stateful nature of `AbstractVCFCodec`. Internal data structures (`parts`, `genotypeParts`, `alleleMap`, `lineNo`) are re-used to parse the `LazyGenotypesContext` contained within `VariantContext` objects; this re-use makes `AbstractVCFCodec` not thread-safe - calling `LazyGenotypesContext.decode()` in concurrent threads results in the genotypes of one `VariantContext` leaking over to other `VariantContext` objects and overwriting the genotypes there.

**Solution:** I have wrapped the internal state objects of `AbstractVCFCodec` in `ThreadLocal` objects. This will give each parsing thread its own copies of the internal state, removing cross-thread information leakage. This solution should not appreciably affect existing workflows, as single threaded applications will still maintain the resource re-use optimization. This solution also removes the need for expensive and rate-limiting synchronization.

I believe this fixes #1026, as well as the issue [reported here](https://gatk.broadinstitute.org/hc/en-us/community/posts/360077666612-Multithreaded-Processing-on-VariantContexts-with-VCFFileReader). 

This also resolves the `// todo: make this thread safe?` comment on line 69 of `AbstractVCFCodec`. 

    - Some tests did not run / pass due to working in Windows / due to path issues. However, on reviewing these tests they do not appear related to this functionality (as best I can tell).
- [ ] Add new tests or update existing ones: [*the test suite is explicitly not multi-threaded, and the comments in the grade build file suggest that making it multi-threaded would break existing tests and make them nondeterministic. However I have confirmed locally that this PR solves the problem reported above.*]
- [X] Check your code style.


rcoleb:
I have pushed a commit that fixes the issue that caused the automatic tests to fail. No error was shown in the IDE due to String.format accepting Object arguments rather than more specifically-typed values. I have checked for other usages of the variable in question and this appears to be the only remaining usage.

lbergelson:
@rcoleb Thank you for doing this.  This seems like a good idea, it's very natural to want to decode these things in parallel and confusing to users that it doesn't work right.  

I'll have to take a close look to make sure there aren't any wierd knock on effects, it's vaguely possible there are wierd expectations about String identity equality or something along those lines that might be disrupted by using multiple caches.  

Do you know if there is any noticeable performance impact of using threadlocal access vs direct access? I assume it's non 0 but likely minimal compared to the extremely costly and wasteful vcf parsing.

We have a two very large outstanding pr's (#1581 and #1596) which are likely to conflict with this.  I will probably hold off on trying to merge this until we get in #1581 (and possibly 1596)  in order to reduce any rebase pain.  I'm currently in the process of reviewing the first one so that should hopefully move along soon.

rcoleb:
There will certainly be some performance cost, but it is likely minimal by my estimation: usage of ThreadLocals will inject a couple of additional method calls into the variable usage but because internally they're implemented as a map, not much more than that. And if the genotypes are even being decoded, it seems likely that the addition of method calls is minimal compared to whatever operations will be performed on the genotypes themselves. 

As an aside, there is a possible solution for adding a "threadsafe" flag to VCFFileReader constructors, which could direct the AbstractVCFCodec to use thread-safe variables if requested, and direct-access otherwise, but that might be more pollution of the constructor that you'd prefer. Overall, I believe (and our internal testing supports) the performance impacts from ThreadLocal usage are much less than the performance gains from concurrent processing. 

Also worth noting, I also don't know how HTSJDK performs with files that don't use AbstractVCFCodec (e.g. BCF files) and whether those would also benefit from thread-safe operations.  We don't use those much, so I don't have the internal pressure to optimize a workflow around them.

Sounds good! Please of course take this in when it is appropriate (if at all) - I am just happy for the potential opportunity to contribute to a library that we use heavily in our lab. In the meantime we are synchronizing the genotype decoding, as even the performance impact of synchronization doesn't compare to the gains we are getting from parallel processing.

tfenne:
Superseded by #1812 which fixes this more correctly and completely.

PR #1640, opened 2022-12-02
Removing scala test infrastructure

lbergelson:
* We tried adding support for scala based tests but it never really caught on.
  It did complicate the testing infrastructure and reporting considerably and
  caused complication in the build.
* Removing scala and scala test and the associated gradle plugins
* Updating gradle 5.2.1 -> 7.5.1
* Converting from the java plugin to the java library plugin

This leaves in place the now defunct scala tests:
  * FastqReaderWriterTest
  * AbstractProgressLoggerTest
  * StringUtilTest
  * BamFileIoUtilsTest

These will need to be back ported to java.  This allows us to move to a new version of gradle and eventually to new jvm versions in a much simplified way.  It also looks like it may speed up the tests.


gradle/spotbugs-exclude.xml
```diff
@@ -71,5 +75,12 @@
         <Class name="htsjdk.samtools.util.BlockCompressedInputStreamTest" />
         <Bug pattern="ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD" />
     </Match>
+    <Match>
+        <!--This seems to be a false positive due to a bug in how spotify handles breaks with gotos -->
```
cmnbroad, line 79: haha spotify breaks us really ?

build.gradle
```diff
@@ -177,7 +178,15 @@ spotbugs {
...
-tasks.withType(com.github.spotbugs.SpotBugsTask) {
+
+spotbugsMain {
+    reports {
+        xml.enabled = false
+        html.enabled = true
+    }
+}
+
+spotbugsTest {
```
cmnbroad, line 189: Kind of a nit, but this seems like a step backwards.

build.gradle
```diff
@@ -92,69 +88,74 @@ tasks.withType(Test) { task ->
     task.maxHeapSize = "2G"
 
     task.jvmArgs '-Djava.awt.headless=true'  //this prevents awt from displaying a java icon while the tests are running
-}
 
-task findScalaAndJavaTypes(type: Exec) {
-    description = "Check that Scala files only exist in the scala test dir and that java files do not reside in the scala test dir."
-    commandLine './scripts/checkScalaAndJavaFiles.sh'
```
cmnbroad, line 99: This script can be deleted now ?

build.gradle
```diff
@@ -31,23 +29,21 @@ jacocoTestReport {
...
     }
 
-    testCompile "org.scala-lang:scala-library:2.12.8"
-    testCompile "org.scalatest:scalatest_2.12:3.0.5"
-    testRuntime 'org.pegdown:pegdown:1.6.0' // Necessary for generating HTML reports with ScalaTest
-    testCompile "org.testng:testng:6.14.3"
-    testCompile "com.google.jimfs:jimfs:1.1"
-    testCompile "com.google.guava:guava:26.0-jre"
-    testCompile "org.apache.commons:commons-lang3:3.7"
+    api "gov.nih.nlm.ncbi:ngs-java:2.9.0"
```
cmnbroad, line 40: What is ngs-java for - SRA ?

cmnbroad (approved):
A few minor questions but mostly looks good.

build.gradle
lbergelson: Yeah, it's for NRA.

src/test/java/htsjdk/HtsjdkTest.java
```diff
@@ -1,10 +1,8 @@
 package htsjdk;
 
-import org.scalatest.testng.TestNGSuite;
-
 /**
  * Base class for all Java tests in HTSJDK.
  */
-public class HtsjdkTest extends TestNGSuite {
+public class HtsjdkTest{
```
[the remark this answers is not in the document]
lbergelson: Yeah, it is pointless now, I wasn't sure about deleting it or just letting it hang out in case we decide to move some functionality there.

lbergelson:
@cmnbroad I'm hoping we don't.... I don't believe we export any of the newly implementationed ones in our public API but I don't know how to check that automatically.  It might cause issues for people who are depending on ours for their own cod without realizing it, but that should be an easy fix for them.  If know how to check the API in a sane way that would be very useful, it's definitely something a tool could exist for.

cmnbroad:
@yash-puligundla This PR might fix the "no test events" issues you were previously seeing (I seem to remember thinking that scala-test was implicated for that). You might rebase on this once it goes in.

Issue #1614, opened 2022-06-14
LittleEndianInputStream EOF test is incorrect


cmnbroad:
As reported by @jrobinso: The test for EOF in `LittleEndianInputStream.readString` is wrong because in.read() is cast to a byte.   The contract for read() returns an "int", and the signal for EOF is the return value is -1.    However this is no longer a valid test after casting to a byte.  The correct thing to do here, I think, is type "b" as an int,  the cast to byte only when written to "bis".
```
    byte b;
    while ((b = (byte) in.read()) != 0) {
        if(b < 0) {
            throw new EOFException();
        }
        bis.write(b);
```

lbergelson:
This is more complicated than originally described because the file encoding isn't really described and java bytes include negative values.

Currently EOFException seems to be triggered correctly when the end of file is reached prematurely, but it's also triggered incorrectly on some non-ASCII characters which truncate to negative byte values.  

It's unclear if this is meant to only read ASCII, ISO 8859-1, or UTF-8 but nothing outside of the ascii space works correctly.   How to fix it depends on what file encoding we're intending to support which means digging into what this thing actually is used for.

jrobinso:
@lbergelson File encoding should be irrelevant, this is a raw byte stream.  Regardless of other issues casting in.read() to a byte before that test is incorrect,  I think.   A positive int can become a negative byte.

jrobinso:
InputStream.read() is supposed to return an integer between 0 and 255, or -1 if EOF,   there is no encoding involved.   So 1/2 of the values it can legally return will be negative when cast to a byte.

jrobinso:
It's a weird convention, but its been like that since the beginning.   I suppose the contract is for an "int" specifically so they can have a return value (-1) which indicates EOF.  If the contract was for a byte there would be no available values to indicate EOF.

lbergelson:
@jrobinso I agree that it's currently wrong and the fix you proposed is an improvement.  What I meant to say is that we have to decide what the file encoding should be when we do the conversion to String because we will get incorrect/inconsistent answers if there's a mismatch and the file contains non-ascii values even once your fix goes in.

jrobinso:
@lbergelson I still don't see what encoding has to do with this class, which is a stream reader for raw bytes.   You can decode these bytes anyway you please, but that is downstream.   In many cases the bytes will not be decoded to strings at all, but again what is done with the bytes is no concern here.

lbergelson:
[...]

PR #1641, opened 2022-12-02
Test case helper and fix EOF detection

lbergelson:
This started as me fixing the minor bug in https://github.com/samtools/htsjdk/issues/1614.  I got annoyed with something that I am regularly annoyed with, needing different data providers for the severals sets of test cases I want to run even though they run the same code.

I often want to right the following test cases for the same function:
 * The function matches a given value
 * The function mismatches the value
 *  The function throws a specific exception.

I added a helper that lets you capture those cases in one data provider.
```
Expected.match(42)
Expected.mismatch("bleh")
Expected.exception(UserException)
```
It's used as an element in the data provider and then tested by calling it's test function:
```.       
  @DataProvider
    public Object[][] testCases() {
       ...                         
     {utf8File, StandardCharsets.UTF_8, Expected.match("🐋 UTF8 is Great 🐋")},
     {utf8File, StandardCharsets.ISO_8859_1, Expected.mismatch("🐋 UTF8 is Great 🐋")}
     {missingTerminator, StandardCharsets.UTF_8, Expected.exception(EOFException.class)},
     ...
   }

@Test(dataProvider = "testCases")
public void testAllCases(String filename, Charset charset, Expected<String> expected) {
     expected.test(() -> {
         try(final LittleEndianInputStream in = new LittleEndianInputStream(new BufferedInputStream(Files.newInputStream(Paths.get(filename))))){
             return in.readString(charset);
         }
     });            
} 
```

It seems like a useful thing to me, but I'm wanted to get feedback on the idea.

### DIFF
--- a/src/test/java/htsjdk/testutil/Expected.java
+++ b/src/test/java/htsjdk/testutil/Expected.java
@@ -1,0 +1,48 @@
+package htsjdk.testutil;
+
+import org.testng.Assert;
+
+public interface Expected<T> {
+    void test(ThrowingSupplier<T> functionToTest);
+
+
+    interface ThrowingConsumer<T> {
+        void test(T a) throws Exception;
+    }
+
+    static <T> Expected<T> match(final T expected) {
+        return new ComparisonExpected<>((T actual) -> Assert.assertEquals(actual, expected));
+    }
+
+    static <T> Expected<T> mismatch(final T expected) {
+        return new ComparisonExpected<>((T actual) -> Assert.assertNotEquals(actual, expected));
+    }
+
+    static <T> Expected<T> exception(final Class<? extends Exception> exceptionClass) {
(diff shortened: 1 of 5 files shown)


src/test/java/htsjdk/testutil/Expected.java
```diff
@@ -0,0 +1,48 @@
...
+    static <T> Expected<T> exception(final Class<? extends Exception> exceptionClass) {
+        return functionToTest -> Assert.assertThrows(exceptionClass, functionToTest::produce);
+    }
+
+    interface ThrowingSupplier<T> {
+        T produce() throws Exception;
+    }
+}
+
+final class ComparisonExpected<T> implements Expected<T> {
```
lbergelson, line 30: This is weirdly scoped because you're not yet allowed to put private classes in an interface.

PR #1645, opened 2023-01-11
Removed 'backed' last cache hit hard references; it seems that that i…

### LINKED ISSUE #1643: Race conditions accessing CRAM's reference data through ReferenceSource
### Description of the issue:
Problem spotted in GATK due to smelly use and no use of synchronized in CRAM ReferenceSource:
 https://github.com/broadinstitute/gatk/issues/8139

.../cram/.../ReferenceSource.java has a mixture of synchronized and non-synchronzed access to its cache member fields that results in a race condition when accessing it in parallel by at least one tool in GATK. 

Although one may argue that GATK should not use it in parallel, this class already contain synchronized modifier which implies that is supposed to support it.

### Your environment:
* version of htsjdk
   Current master:9f0e80f5bf
* version of java
   Not relevant
* which OS
   Not relevant

### Steps to reproduce

See original post. No concrete test input data provided but the problem becomes clear when inspecting the code.

### Expected behaviour

Either
(a) this class is made multithread safe or 
(b) it clearly stated that is not multi-thread safe and the multithread elements (e.g. synchronized modifiers) are removed. 

### Actual behaviour
At least there is one way to cause race conditions in multi-thread use of this class. There could be more, so code should be review for other possible issues.

vruano:
### Description

Removed 'backed' last cache hit hard references; it seems that that it might be trying to solve an non-existent issue.
Changed WeakReference to SoftReference to make cache more persistent in memory. Addresses GATK issue #1643.

- [X, but not new tests] Make sure your changes compile and new tests pass locally.
- [NA] Extended the README / documentation, if necessary
- [X] Check your code style.

### DIFF
--- a/src/main/java/htsjdk/samtools/cram/ref/ReferenceSource.java
+++ b/src/main/java/htsjdk/samtools/cram/ref/ReferenceSource.java
@@ -33,7 +33,7 @@ import htsjdk.utils.ValidationUtils;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.lang.ref.WeakReference;
+import java.lang.ref.SoftReference;
 import java.net.URL;
 import java.nio.file.Path;
 import java.util.ArrayList;
@@ -58,13 +58,7 @@ public class ReferenceSource implements CRAMReferenceSource {
     private final ReferenceSequenceFile rsFile;
     private int downloadTriesBeforeFailing = 2;
 
-    private final Map<String, WeakReference<byte[]>> cacheW = new HashMap<>();
-
-    // Optimize for locality of reference by maintaining the backing reference bases for the most recent request.
-    // Failure to do this will result in the bases being aggressively GC'd when the only other reference to them
-    // is the weak reference hash map above, resulting in much thrashing.
-    private byte[] backingReferenceBases;
-    private int backingContigIndex;
+    private final Map<String, SoftReference<byte[]>> cache = new HashMap<>();
 
     public ReferenceSource(final File file) {
         this(IOUtil.toPath(file));
@@ -116,7 +110,7 @@ public class ReferenceSource implements CRAMReferenceSource {
     }
 
     private byte[] findInCache(final String name) {
-        final WeakReference<byte[]> weakReference = cacheW.get(name);
+        final SoftReference<byte[]> weakReference = cache.get(name);
         if (weakReference != null) {
             final byte[] bytes = weakReference.get();
             if (bytes != null)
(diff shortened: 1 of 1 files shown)


lbergelson:
@cmnbroad Can you take a look at this?  It seems like weak -> soft reference is the right change, and hopefully removing the unsychronized method helps, but it would be good if you could take a look.

src/main/java/htsjdk/samtools/cram/ref/ReferenceSource.java
```diff
@@ -116,7 +110,7 @@ else if (Defaults.USE_CRAM_REF_DOWNLOAD) {
     }
 
     private byte[] findInCache(final String name) {
-        final WeakReference<byte[]> weakReference = cacheW.get(name);
+        final SoftReference<byte[]> weakReference = cache.get(name);
         if (weakReference != null) {
```
droazen, line 114: Can you explicitly check for an empty array here as well, and throw an informative error if you encounter one?

vruano: Under what circumstances the bases array would be an empty array as supposed to a null (claimed by the GC)?

cmnbroad:
I don't have any issue with changing `WeakReference` to `SoftReference`, if it somehow actually fixes a problem. Is there any evidence that it does ? The gatk code isn't (and shouldn't be) sharing these across threads as far as I know, since that came up in the original code review.

We can't remove the backing reference though, since that was added to address a real performance issue (described in the comments). As I understand it, `SoftReference` is not required to have any different behavior than `WeakReference`. If thats correct, removing the backing reference would reintroduce the perf issue for at least some Java implementations.

I agree that the class here contract isn't clear; though, the synchronized does make it look like it's thread safe, but it isn't. I'm not sure why that was originally in there.

vruano:
Thanks @cmnbroad. So far there is no evidence that Soft for Weak fixes the issue, may well in most cases but as you mentioned is not guarantee in all implementations. 

I guess that there is two ways to proceed, either we actually make it thread safe adding back the last-requested hard reference or we simply scrap all the thread safety away   and enforce the originally intended non-thread safe contract. 

I think the most pragmatical would be the former; two make it truly thread-safe as is an improvement on the existing code. But perhaps you see this as an opportunity to "make it right" although with the risk of breaking additional third party dependent code.

lbergelson:
@cmnbroad I think the Weak -> Soft reference switch makes sense, it definitely seems like the intent is that WeakReferences are collected eagerly and Soft are kept until there is memory pressure.  It's possible that a jvm will implement it badly but I would guess that it's going to do basically what we want on any normal jvm we're likely to run on.  I think it's very possible that the switch will fix the performance issue even if the explicit reference is removed.  

With the Weak references the cache is effectively doing nothing, which explains why holding the last one specifically was necessary.  If there is bad memory pressure we could still see thrashing here, but the alternative might be that we hold the reference and get OOM?  Probably safest to keep the backing bases if you're worried about it, but I think in any case if we want the cache to work we should swap Weak -> Soft.

vruano:
If we reinstate the 'backed' strong references would it make sense that these are updated and use regardless whether the invoking code calls get getReferenceSequence or getReferenceSequenceByRegion? The original code would only use and update the backed references in the latter is called. Is the a reason against doing so for both?

PR #1648, opened 2023-01-27
Expose the ability to encode a Genotoype into a GT field

lbergelson:
* Expose two public methods in VCFEncoder writeGtField and encodeGtField
* Supports https://github.com/broadinstitute/gatk/issues/8160 but seems like a useful thing to be able to do in general
* minor breaking change in VCFEncoder, made methods formatVCFField and buildAlleleStrings static
  It is unlikely anyone overrides either of these methods so it should not be a problem.

### DIFF
--- a/src/test/java/htsjdk/variant/vcf/VCFEncoderTest.java
+++ b/src/test/java/htsjdk/variant/vcf/VCFEncoderTest.java
@@ -3,6 +3,7 @@ package htsjdk.variant.vcf;
 import htsjdk.HtsjdkTest;
 import htsjdk.tribble.util.ParsingUtils;
 import htsjdk.variant.variantcontext.Allele;
+import htsjdk.variant.variantcontext.Genotype;
 import htsjdk.variant.variantcontext.GenotypeBuilder;
 import htsjdk.variant.variantcontext.VariantContext;
 import htsjdk.variant.variantcontext.VariantContextBuilder;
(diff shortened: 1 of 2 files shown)


src/test/java/htsjdk/variant/vcf/VCFEncoderTest.java
```diff
@@ -135,6 +136,20 @@ public Object[][] makeMissingFormatTestData() {
...
+                "chr?", 100, 100,
+                Arrays.asList(Allele.REF_A, Allele.ALT_T, Allele.create("TC")));
+        final Genotype g1 = new GenotypeBuilder("s1", Arrays.asList(Allele.REF_A, Allele.REF_A)).make();
+        final Genotype g2 = new GenotypeBuilder("s2", Arrays.asList(Allele.ALT_T, Allele.create("TC"))).make();
+        vcb.genotypes(g1, g2);
+        final VariantContext vc = vcb.make();
+
+        Assert.assertEquals(VCFEncoder.encodeGtField(vc, g1), "0/0");
+        Assert.assertEquals(VCFEncoder.encodeGtField(vc, g2), "1/2");
+    }
```
droazen, line 151: Can you add a unit test for `writeGtField()` as well?

droazen (approved):
One comment to address, otherwise looks good 👍

PR #1655, opened 2023-02-17
Fix an NPE in CRAMRecordReadFeatures.restoreReadBases

lbergelson:
* Fix for https://github.com/igvteam/igv/issues/1286

@cmnbroad This is a bug related to handling embedded references.


src/test/java/htsjdk/samtools/CRAMReferencelessTest.java
```diff
@@ -24,11 +24,37 @@ private void testReadCRAMWithEmbeddedReference() throws IOException {
...
+            final Iterator<SAMRecord> cramIterator = cramReader.iterator();
+            final Iterator<SAMRecord> cramEmbeddedIterator = cramReaderEmbedded.iterator();
+            int count = 0;
+            while (cramIterator.hasNext() && cramEmbeddedIterator.hasNext()) {
+                count++;
+                final SAMRecord cramRecord = cramIterator.next();
+                final SAMRecord cramRecordEmbedded = cramEmbeddedIterator.next();
+                Assert.assertEquals(cramRecordEmbedded, cramRecord);
+            }
+            Assert.assertEquals(count, 2);
```
droazen, line 57: Also assert that both iterators are exhausted after the loop

src/test/java/htsjdk/samtools/CRAMReferencelessTest.java
```diff
@@ -24,11 +24,37 @@ private void testReadCRAMWithEmbeddedReference() throws IOException {
...
             final Iterator<SAMRecord> cramIterator = cramReader.iterator();
             final Iterator<SAMRecord> cramEmbeddedIterator = cramReaderEmbedded.iterator();
+            int count = 0;
             while (cramIterator.hasNext() && cramEmbeddedIterator.hasNext()) {
+                count++;
                 final SAMRecord cramRecord = cramIterator.next();
                 final SAMRecord cramRecordEmbedded = cramEmbeddedIterator.next();
                 Assert.assertEquals(cramRecordEmbedded, cramRecord);
             }
+            Assert.assertTrue( count >0);
```
droazen, line 34: Add text message for when assertion fails

droazen (approved):
👍 merge after addressing minor comments

cmnbroad (approved):
So I ran this in the debugger, and the reason the other embedded ref tests don't catch this issue is they have less interesting/varied read features than the other embedded ref tests (specifically this one has two read features with a gap between them). So I think this change is fine. We could use some better embedded ref tests.

PR #1659, opened 2023-03-10
Remove low-value progress logging message

nh13:
### Description

I think printing read names for non-coordinate sorted data is useful, but the log message that "Seen many non-increasing record positions. Printing Read-names as well" is not that useful confusing to both developers and users, especially in an API.  I'd be fine with it being at the debug level.  I just don't want to see this message in my logs every time I use a progress logger that doesn't feed coordinate-sorted data.


lbergelson (approved):
Seems reasonable.

PR #1664, opened 2023-03-23
removed redundant code that sets the secondary alignment flag for a r…

KleinSamuel:
…eadpair

### Description

Short:
Removed duplicate code fragment that had no effect.

Long:
When adding a new read pair to a SAMRecord using the addPair() method in SAMRecordSetBuilder,
the parameters "record1NonPrimary" as well as "record2NonPrimary" are checked to set the 
secondary alignment flag of the SAM record. This code fragment for both reads was duplicated with 
no effect. The duplicate is removed in this PR.


lbergelson (approved):
Great catch.  Thank you!

PR #1194, opened 2018-10-10
Adding support for 0-length B arrays in SAM files to conform to 1.6 spec

lbergelson:
Making changes to allow 0-length arrays in SAM because this is allowed in SAM 1.6

### Checklist

- [ ] Code compiles correctly
- [ ] New tests covering changes and new functionality
- [ ] All tests passing
- [ ] Extended the README / documentation, if necessary
- [ ] Is not backward compatible (breaks binary or source compatibility)

PR #1674, opened 2023-08-03
Zero len array

gileshall:
PR https://github.com/samtools/htsjdk/pull/1194 removed a number of Array.getLength(value) == 0 checks when setting B-array tagged fields, as such empty tagged field values had recently been made explicitly valid in SAM/BAM/CRAM files.

Unfortunately the existing such check in setUnsignedArrayAttribute() was omitted from that PR and has now caused the issue raised in https://github.com/samtools/hts-specs/issues/728. Whatever additional problems might (but probably don't) exist elsewhere in the HTSJDK code for zero-length ML tags, this check for a generic zero-length array was incorrect and should be removed. A test case for setUnsignedArrayAttribute() has also been added.

In grepping for other candidates, I noticed one other potentially incorrect looking Array.getLength invocation.

This was in Slice::setAttribute(), which appears to set an optional tag in a Slice header. There are currently no such tags defined, so the question is somewhat moot but — as a separate commit for your consideration — I have also removed that check. (As there are no such tags currently defined, there are no test cases for these fields to which a test exercising this change could be added.)

--

Updated [jmarshall](https://github.com/samtools/htsjdk/commits?author=jmarshall) pull request to remove unused Slice::setAttribute() methods.

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMRecord.java
+++ b/src/main/java/htsjdk/samtools/SAMRecord.java
@@ -1541,9 +1541,6 @@ public class SAMRecord implements HtsRecord, Cloneable, Locatable, Serializable 
         if (!value.getClass().isArray()) {
             throw new IllegalArgumentException("Non-array passed to setUnsignedArrayAttribute for tag " + tag);
         }
-        if (Array.getLength(value) == 0) {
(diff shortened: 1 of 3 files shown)


lbergelson (approved):
Thank you @gileshall @jmarshall

PR #1680, opened 2023-08-08
Add IntervalFileFeature for common interface to name in bed or interval_list

rickymagner:
### Description

This short PR adds an `IntervalFileFeature` interface which extends the `Feature` interface to include a `getName` method common to both lines in bed files and interval_list files. This allows you to be able to parse and work with bed and interval_list file lines uniformly when you need to access metadata in the `name` field. This will be used in an upcoming change in GATK.

### DIFF
--- a/src/main/java/htsjdk/tribble/IntervalFileFeature.java
+++ b/src/main/java/htsjdk/tribble/IntervalFileFeature.java
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2013 The Broad Institute
+ * Copyright (c) 2023 The Broad Institute
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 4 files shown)


src/main/java/htsjdk/tribble/IntervalFileFeature.java
```diff
@@ -0,0 +1,33 @@
...
+ * THE SOFTWARE.
+ */
+package htsjdk.tribble;
+
+
+/**
+ * An interface for features provided via an interval file, e.g. bed or interval_list.
+ * Provides a common interface for accessing the name column for both of these file types.
+ */
+public interface IntervalFileFeature extends Feature {
```
lbergelson, line 31: I might call this "NamedFeature"

rickymagner: Good point. I just updated the name to the more descriptive `NamedFeature`.

lbergelson (approved):
:th

lbergelson:
Ack, I merged without fixing the commit name... .bleh.

PR #1683, opened 2023-09-05
VCF 4.4 optimistic read switch.

cmnbroad:
For igv https://github.com/igvteam/igv/issues/1289.


droazen:
@jrobinso When you get a chance, could you have a look at this and let us know if it will satisfy IGV's needs for now?

lbergelson (approved):
Thank you @cmnbroad  @jrobinso

PR #1688, opened 2023-10-13
Fix IOUtil.unrollPaths for http paths with query parameters

lbergelson:
* Query parameters in an http path were breaking extension detection


src/test/resources/htsjdk/samtools/io/FofnWithHttpQueryParams.fofn
```diff
@@ -0,0 +1,3 @@
+src/test/resources/htsjdk/samtools/io/Level1.fofn
+http://www.example.com/blurg.vcf
+http://www.example.com/blurg.vcf?huh=bleep,bloop
```
jamesemery, line 3: bleep, bloop

jamesemery (approved):
Just tested this with my SAS token issue in GATK. 👍

PR #1693, opened 2023-12-11
Re-Ordering priority in SeekableStreamFactory

lbergelson:
* Now when constructing streams it will prefer NIO plugins.
* If no http / ftp plugin exists it will fall back to the htsjdk bulit in
* No longer special casing local files, these will now be handled through NIO as well.
* Deprecating SeekableStreamFactory.isFilePath() since it is no longer used and interacts poorly
  with nio filesystem providers

### DIFF
--- a/src/main/java/htsjdk/io/HtsPath.java
+++ b/src/main/java/htsjdk/io/HtsPath.java
@@ -253,6 +253,10 @@ public class HtsPath implements IOPath, Serializable {
                 tempURI = getCachedPath().toUri();
             }
         } catch (URISyntaxException uriException) {
+            //check that the uri wasn't a badly encoded absolute uri of some sort
+            //if you don't do this it will be treated as a badly formed file:/ url
+            assertNoNonFileScheme(pathString, uriException);
+
             // the input string isn't a valid URI; assume its a local (non-URI) file reference, and
             // use the URI resulting from the corresponding Path
             try {
@@ -275,6 +279,26 @@ public class HtsPath implements IOPath, Serializable {
         }
 
         return tempURI;
+    }
+
+    /**
+     * check that there isn't a non file scheme at the start of the path
+     * @param pathString
+     * @param cause
+     */
+    private static void assertNoNonFileScheme(String pathString, URISyntaxException cause){
+        final String[] split = pathString.split(":");
+        if(split.length > 1){
+            if(split[0] == null || split[0].isEmpty()){
+                throw new IllegalArgumentException("Malformed url " + pathString + " includes an empty scheme." +
+                        "\nCheck that it is fully encoded.", cause);
+            }
+            if(!split[0].equals("file")){
+                throw new IllegalArgumentException("Malformed url " + pathString + " includes a scheme: " + split[0] + ":/ but was an invalid URI." +
+                        "\nCheck that it is fully encoded.", cause);
+            }
+        }
+
     }
     
 }
--- a/src/main/java/htsjdk/io/HtsPath.java
+++ b/src/main/java/htsjdk/io/HtsPath.java
@@ -254,7 +254,7 @@ public class HtsPath implements IOPath, Serializable {
             }
         } catch (URISyntaxException uriException) {
             //check that the uri wasn't a badly encoded absolute uri of some sort
-            //if you don't do this it will be treated as a badly formed file:/ url
+            //if you don't do this it will be treated as a badly formed file:// url
             assertNoNonFileScheme(pathString, uriException);
 
             // the input string isn't a valid URI; assume its a local (non-URI) file reference, and
@@ -294,7 +294,7 @@ public class HtsPath implements IOPath, Serializable {
                         "\nCheck that it is fully encoded.", cause);
             }
             if(!split[0].equals("file")){
-                throw new IllegalArgumentException("Malformed url " + pathString + " includes a scheme: " + split[0] + ":/ but was an invalid URI." +
+                throw new IllegalArgumentException("Malformed url " + pathString + " includes a scheme: " + split[0] + ":// but was an invalid URI." +
                         "\nCheck that it is fully encoded.", cause);
             }
         }
--- a/src/main/java/htsjdk/samtools/seekablestream/SeekableStreamFactory.java
+++ b/src/main/java/htsjdk/samtools/seekablestream/SeekableStreamFactory.java
@@ -91,9 +91,9 @@ public class SeekableStreamFactory{
                 && canBeHandledByLegacyUrlSupport(path); // otherwise we fall back to the special handlers
     }
 
-    //is this onee of the url types that has legacy htsjdk support built in?
+    //is this one of the url types that has legacy htsjdk support built in?
     public static boolean canBeHandledByLegacyUrlSupport(final String path) {
-        return URL_SCHEMES_WITH_LEGACY_SUPPORT.stream().anyMatch(path::startsWith);
+        return URL_SCHEMES_WITH_LEGACY_SUPPORT.stream().anyMatch(scheme-> path.startsWith(scheme +"://"));
     }
 
     private static class DefaultSeekableStreamFactory implements ISeekableStreamFactory {
--- a/src/main/java/htsjdk/samtools/seekablestream/SeekableStreamFactory.java
+++ b/src/main/java/htsjdk/samtools/seekablestream/SeekableStreamFactory.java
@@ -23,12 +23,17 @@
  */
 package htsjdk.samtools.seekablestream;
 
+import htsjdk.io.HtsPath;
+import htsjdk.io.IOPath;
 import htsjdk.samtools.util.IOUtil;
+import htsjdk.tribble.TribbleException;
+
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URL;
 import java.nio.channels.SeekableByteChannel;
+import java.nio.file.Path;
 import java.util.function.Function;
 
 /**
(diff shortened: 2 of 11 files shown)


droazen:
Can you also fix the order in `Parsing.openInputStream()` (used by `TribbleIndexedFeatureReader.readHeader()`)?

src/main/java/htsjdk/samtools/seekablestream/SeekableStreamFactory.java
```diff
@@ -61,9 +66,28 @@ public static ISeekableStreamFactory getInstance(){
...
+    }
+
+    /**
+     * is this path being handled by one of the legacy SeekableStream types (http(s) / ftp)
+     *
+     * @param path a path to check
+     * @return if the path is not being handled by a FileSystemProvider and it can be read by legacy streams
+     */
+    public static boolean isSpecialCase(final String path){
+        return !new HtsPath(path).isPath()  //if we have a provider for it that's what we'll use
```
droazen, line 84: Maybe break out out a `hasNIOPlugin()` (or similar) method for the `new HtsPath(path).isPath()` everywhere, for clarity?

src/main/java/htsjdk/samtools/seekablestream/SeekableStreamFactory.java
```diff
@@ -61,9 +66,28 @@ public static ISeekableStreamFactory getInstance(){
...
+                && isSpecialCaseUrlType(path); // otherwise we fall back to the special handlers
+    }
+
+    //is this onee of the url types that has legacy htsjdk support built in?
+    private static boolean isSpecialCaseUrlType(final String path) {
```
droazen, line 89: `isSupportedByLegacyCloudReaders()`?

src/main/java/htsjdk/samtools/seekablestream/SeekableStreamFactory.java
droazen: Or should you be using `hasFileSystemProvider()` instead?

src/main/java/htsjdk/samtools/seekablestream/SeekableStreamFactory.java
```diff
@@ -89,26 +113,30 @@ public SeekableStream getStreamFor(final String path) throws IOException {
...
+         * This has a fall back to htsjdk's built in http and ftp providers if no FileSystemProvder is available for them
+         *
+         * @param path    an IOPath to be opened
+         * @param wrapper a wrapper to apply to the stream allowing direct transformations on the byte stream to be applied
+         * @throws IOException
+         */
+        public static SeekableStream getStreamFor(final IOPath path, Function<SeekableByteChannel, SeekableByteChannel> wrapper) throws IOException {
+            if(path.isPath()) {
+                return path.getScheme().equals("file")
+                        ? new SeekablePathStream(path.toPath()) //don't apply the wrapper to local files
```
droazen, line 132: Check usages of `SeekablePathStream.getSource()` to make sure the switch from File -> Path for local files is safe (concerned about file:// appearing unexpectedly downstream). Or to be super-safe, could stick with `SeekableFileStream` for now.

src/main/java/htsjdk/samtools/seekablestream/SeekableStreamFactory.java
```diff
@@ -89,26 +113,30 @@ public SeekableStream getStreamFor(final String path) throws IOException {
...
+        /**
+         * The wrapper will only be applied to the stream if the stream is treated as a non file:// {@link Path}
+         *
+         * This has a fall back to htsjdk's built in http and ftp providers if no FileSystemProvder is available for them
+         *
+         * @param path    an IOPath to be opened
+         * @param wrapper a wrapper to apply to the stream allowing direct transformations on the byte stream to be applied
+         * @throws IOException
+         */
+        public static SeekableStream getStreamFor(final IOPath path, Function<SeekableByteChannel, SeekableByteChannel> wrapper) throws IOException {
```
droazen, line 129: If there aren't already tests covering `file://` vs. local file paths (no protocol), we might want one

droazen (approved):
Looks good, modulo a few requests

src/main/java/htsjdk/samtools/seekablestream/SeekableStreamFactory.java
lbergelson: I switched back to SeekableFileStream although it makes me sad.

src/main/java/htsjdk/io/HtsPath.java
```diff
@@ -253,6 +253,10 @@ private URI getURIForString(final String pathString) {
                 tempURI = getCachedPath().toUri();
             }
         } catch (URISyntaxException uriException) {
+            //check that the uri wasn't a badly encoded absolute uri of some sort
+            //if you don't do this it will be treated as a badly formed file:/ url
```
droazen, line 257: file://

src/main/java/htsjdk/io/HtsPath.java
```diff
@@ -276,5 +280,25 @@ private URI getURIForString(final String pathString) {
...
+     */
+    private static void assertNoNonFileScheme(String pathString, URISyntaxException cause){
+        final String[] split = pathString.split(":");
+        if(split.length > 1){
+            if(split[0] == null || split[0].isEmpty()){
+                throw new IllegalArgumentException("Malformed url " + pathString + " includes an empty scheme." +
+                        "\nCheck that it is fully encoded.", cause);
+            }
+            if(!split[0].equals("file")){
+                throw new IllegalArgumentException("Malformed url " + pathString + " includes a scheme: " + split[0] + ":/ but was an invalid URI." +
```
droazen, line 297: :/ -> ://

src/main/java/htsjdk/samtools/seekablestream/SeekableStreamFactory.java
```diff
@@ -61,9 +72,28 @@ public static ISeekableStreamFactory getInstance(){
...
+     * @return if the path is not being handled by a FileSystemProvider and it can be read by legacy streams
+     */
+    public static boolean isBeingHandledByLegacyUrlSupport(final String path){
+        return !new HtsPath(path).hasFileSystemProvider()  //if we have a provider for it that's what we'll use
+                && canBeHandledByLegacyUrlSupport(path); // otherwise we fall back to the special handlers
+    }
+
+    //is this onee of the url types that has legacy htsjdk support built in?
+    public static boolean canBeHandledByLegacyUrlSupport(final String path) {
+        return URL_SCHEMES_WITH_LEGACY_SUPPORT.stream().anyMatch(path::startsWith);
```
droazen, line 96: Shouldn't you be checking that the path starts with "http://" (for example) instead of "http"? What if there's a file named "http"?

src/main/java/htsjdk/samtools/seekablestream/SeekableStreamFactory.java
```diff
@@ -40,6 +43,14 @@
 public class SeekableStreamFactory{
 
     private static final ISeekableStreamFactory DEFAULT_FACTORY;
+    private static final String HTTP = "http";
+    private static final String HTTPS = "https";
+    private static final String FTP = "ftp";
```
droazen, line 48: See below -- recommend including "://" in these constants to guard against files named after a protocol

src/test/java/htsjdk/samtools/seekablestream/SeekableStreamFactoryTest.java
```diff
@@ -10,20 +12,45 @@
...
+
+    // this test isn't particularly useful since we're not testing the meaninful case of having the http-nio provider
+    // installed
+    @Test(dataProvider = "getSpecialCasePaths")
+    public void testIsBeingHandledByLegacyUrlSupport(String path, boolean notExpected) {
+        Assert.assertEquals(SeekableStreamFactory.isBeingHandledByLegacyUrlSupport(path), !notExpected);
+    }
+
+    @Test(dataProvider = "getSpecialCasePaths")
+    public void testCanBeHandledByLegacyUrlSuppoort(String path, boolean notExpected){
```
droazen, line 52: Suppoort....this test is identical to the one just before it. Did you mean to test `canBeHandledByLegacyUrlSupport()` here instead?

src/test/java/htsjdk/samtools/seekablestream/SeekableStreamFactoryTest.java
```diff
@@ -69,7 +97,17 @@ public void testPathWithEmbeddedSpace() throws IOException {
...
+    public void testGetSeekableStreamWorksOnJimfs() throws IOException {
+        try(final FileSystem fs = Jimfs.newFileSystem(Configuration.unix())) {
+            final Path file = fs.getPath("/file");
+            Files.writeString(file,"hello");
+            try(final SeekableStream stream = SeekableStreamFactory.getInstance().getStreamFor(file.toUri().toString())){
+                final byte[] bytes = stream.readAllBytes();
+                Assert.assertEquals(new String(bytes, StandardCharsets.UTF_8), "hello");
+            }
+        }
+    }
```
droazen, line 112: Is there any easy way to get test coverage for the case where `isBeingHandledByLegacyUrlSupport()` detects that an NIO plugin exists for the legacy-supported URL? I'm guessing not, but thought I'd ask.

droazen (approved):
Looks good, just some minor comments. Merge once tests pass and you've addressed them. Have you verified that with this patch we are using http-nio in tribble in GATK? Did you look at the SAM/BAM/CRAM path?

src/test/java/htsjdk/samtools/seekablestream/SeekableStreamFactoryTest.java
lbergelson: not an easy way, we can add an additional test suite in gradle but it's going to be a hassle.

src/test/java/htsjdk/samtools/seekablestream/SeekableStreamFactoryTest.java
lbergelson: yes

src/test/java/htsjdk/samtools/seekablestream/SeekableStreamFactoryTest.java
droazen: forget it then

src/main/java/htsjdk/samtools/seekablestream/SeekableStreamFactory.java
lbergelson: instead of adding them to the constants which are also used directly compared against getScheme the isHandledBy method now appends '://' to them.

src/test/java/htsjdk/samtools/seekablestream/SeekableStreamFactoryTest.java
lbergelson: I did a manual test where I added the http-nio library to the test dependencies and isBeingHandledByLegacySupport for http / https is changes as expected.

PR #1694, opened 2023-12-13
Improving the somewhat slipshod change I made to HtsPath

lbergelson:
@droazen Can you take a quick look at this?  It fixes some cases that I should have handled before.

### DIFF
--- a/src/main/java/htsjdk/io/HtsPath.java
+++ b/src/main/java/htsjdk/io/HtsPath.java
@@ -27,7 +27,7 @@ import java.nio.file.spi.FileSystemProvider;
  * encoding/escaping.
  *
  * For example, a URI that contains a scheme and has an embedded "#" in the path will be treated as a URI
- * having a fragment delimiter. If the URI contains an scheme, the "#" will be escaped and the encoded "#"
+ * having a fragment delimiter. If the URI contains no scheme, the "#" will be escaped and the encoded "#"
  * will be interpreted as part of the URI path.
  *
  * There are 3 succeeding levels of input validation/conversion:
@@ -69,6 +69,7 @@ import java.nio.file.spi.FileSystemProvider;
  */
 public class HtsPath implements IOPath, Serializable {
     private static final long serialVersionUID = 1L;
+    private static final String HIERARCHICAL_SCHEME_SEPARATOR = "://";
 
     private final String    rawInputString;     // raw input string provided by th user; may or may not have a scheme
     private final URI       uri;                // working URI; always has a scheme ("file" if not otherwise specified)
@@ -255,7 +256,7 @@ public class HtsPath implements IOPath, Serializable {
         } catch (URISyntaxException uriException) {
             //check that the uri wasn't a badly encoded absolute uri of some sort
             //if you don't do this it will be treated as a badly formed file:// url
-            assertNoNonFileScheme(pathString, uriException);
+            assertNoProblematicScheme(pathString, uriException);
 
             // the input string isn't a valid URI; assume its a local (non-URI) file reference, and
             // use the URI resulting from the corresponding Path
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/io/HtsPath.java
```diff
@@ -282,19 +283,36 @@ private URI getURIForString(final String pathString) {
     }
 
     /**
-     * check that there isn't a non file scheme at the start of the path
-     * @param pathString
-     * @param cause
+     * Check for problems associated with the presence of a heirachical scheme.
```
droazen, line 286: hierarchical

src/main/java/htsjdk/io/HtsPath.java
```diff
@@ -282,19 +283,36 @@ private URI getURIForString(final String pathString) {
...
      */
-    private static void assertNoNonFileScheme(String pathString, URISyntaxException cause){
-        final String[] split = pathString.split(":");
+     static void assertNoProblematicScheme(String pathString, URISyntaxException cause){
+        if(pathString.equals(HIERARCHICAL_SCHEME_SEPARATOR)){
+            throw new IllegalArgumentException(HIERARCHICAL_SCHEME_SEPARATOR + " is not a valid path.", cause);
+        }
+
+        if(pathString.endsWith(HIERARCHICAL_SCHEME_SEPARATOR)){
+            throw new IllegalArgumentException("A path consisting of only a scheme is not allowed: " + pathString, cause);
```
droazen, line 304: Hmm...the string `http://host/path://` also ends with the HIERARCHICAL_SCHEME_SEPARATOR, but has a scheme?

src/main/java/htsjdk/io/HtsPath.java
```diff
@@ -282,19 +283,36 @@ private URI getURIForString(final String pathString) {
...
+        }
+
+        final String[] split = pathString.split(HIERARCHICAL_SCHEME_SEPARATOR);
```
droazen, line 307: split(HIERARCHICAL_SCHEME_SEPARATOR, -1) to preserve trailing empty tokens

src/test/java/htsjdk/io/HtsPathUnitTest.java
```diff
@@ -25,6 +27,8 @@
 import java.util.Map;
 import java.util.Optional;
 
+import static htsjdk.io.HtsPath.assertNoProblematicScheme;
```
droazen, line 30: remove static import

src/test/java/htsjdk/io/HtsPathUnitTest.java
```diff
@@ -169,7 +173,15 @@ public Object[][] invalidHtsPath() {
...
                 // filenames and URIs, so use it to test HtsPath constructor failure on all platforms
                 {"\0"},
-                {"ftp://broad.org/file with space"} // this has a non-file scheme but isn't encoded properly
+
+                // this has a non-file scheme but isn't encoded properly, best to reject these with
+                // a helpful error message than to continue on and treat it as a file:// path
+                {"ftp://broad.org/file with space"},
+
+                // if you have a scheme you need something with it
+                {"file://"},
```
droazen, line 182: `file://` should be rejected, yes, but do we have a test that proves we accept `file:///` (with 3 slashes, indicating the root directory)?

droazen: Ah never mind, I see that you do below

src/test/java/htsjdk/io/HtsPathUnitTest.java
```diff
@@ -649,4 +664,52 @@ private String getLocalFileAsURIPathString(final Path localPath) {
...
+        assertNoProblematicScheme(path, null);
+    }
+
+    @DataProvider
+    public Object[][] getProblematic(){
+        return new Object[][]{
+
+                //This is the primary use case, to detect unencoded uris that were intended to be encoded
+                {"http://forgot.com/to encode"},
+                {"ftp://server.com/file name.txt"},
```
droazen, line 694: So if I understand this correctly, `assertNoProblematicScheme()` only rejects these because they have a non-file scheme, and doesn't actually check the encoding at all, and the assumption is that we've already triggered a URI parsing exception? Might want to add a comment here explaining that

droazen (approved):
Looks good, just a couple of small comments / questions

src/test/java/htsjdk/io/HtsPathUnitTest.java
lbergelson: yeah, it's stuck into the catch block so it's already guarenteed to have hit the exception, good idea to comment.  

honestly, it should probably be restructured so it's in the catch block but I wanted to change as little as possible

src/test/java/htsjdk/io/HtsPathUnitTest.java
lbergelson: file:/// should never actually reach this method in the first place, but if it did it should also not be rejected.

src/test/java/htsjdk/io/HtsPathUnitTest.java
lbergelson: thank you intellij

src/main/java/htsjdk/io/HtsPath.java
lbergelson: thats.... a good point

src/main/java/htsjdk/io/HtsPath.java
lbergelson: hieirearchicahcl

lbergelson: I wish intellij was more consistent in its spell checking, it just ignores some stuff and I don't know why

PR #1698, opened 2023-12-18
Update for to use Java 21

lbergelson:
* update tests to run on 21
* remove deprecated uses of finalize, make DiskBackQueue and associated Autocloseable
  so that it's easier for downstream code to handle them without finalize
  Note: if this proves to be insufficient we should look into implementing a Cleaner
* remove deprecated boxing constructors new Byte() -> Byte.valueof() etc

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [ 17 ]
+        java: [ 21 ]
         experimental: [false]
       fail-fast: false
     continue-on-error: ${{ matrix.experimental }}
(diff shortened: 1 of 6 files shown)


tfenne:
Superseded by #1786, which adopts the Java-17-safe cleanups from this PR (removing `finalize()`, dropping `System.runFinalization()`, and replacing deprecated `Class.newInstance()`) rebased onto current `master`, while keeping htsjdk on Java 17. The Java 21 bump is dropped, and the SpotBugs changes here are already superseded on `master` (plugin is now 6.4.8). Original authorship is preserved via a co-author trailer. Proposing we close this in favor of #1786.

tfenne:
Closing as superseded by #1786.

PR #1707, opened 2024-03-25
Ignore version validation for BCF in VCFHeaderReader  because we only need the header

lindenb:
### Description

For now HTSJDK does not support **BCF2.2** but sometimes you just want to only get the VCF header in a **BCF** file while you don't care about the variants . e.g: 

> search all the vcfs in your filesystem with given dictionary and for given sample(s)).

So this PR, overrides `BCF2Codec.validateVersionCompatibility` in **VCFheaderReader** so the validation is just ignored.

VCFheaderReader is then used in **SamSequenceDictionaryExtractor** instead of VCFReader so we can extract a dictionary from a BCF2.2 file

I added tests.

- [X] Check your code style.


tfenne:
Superseded by #1827, which reads BCF 2.1 and 2.2 (raw and BGZF) without a version gate, so `VCFHeaderReader` no longer needs to skip validation for BCF. Thanks, @lindenb.

PR #1708, opened 2024-04-09
Fix CRAMReferenceRegion updating.

cmnbroad:
Fix for https://github.com/broadinstitute/gatk/issues/8768.


src/main/java/htsjdk/samtools/cram/build/CRAMReferenceRegion.java
```diff
@@ -113,9 +113,14 @@ public void fetchReferenceBases(final int referenceIndex) {
 
         // Re-resolve the reference bases if we don't have a current region or if the region we have
         // doesn't span the *entire* contig requested.
+        final SAMSequenceRecord newSequenceRecord = sequenceDictionary.getSequence(referenceIndex);
+        if (newSequenceRecord == null) {
+            throw new IllegalArgumentException(
+                    String.format("Requested reference sequence index %d not found", referenceIndex));
+        }
```
droazen, line 120: Can you call the existing `getSAMSequenceRecord()` method here? It does exactly the same thing.

src/main/java/htsjdk/samtools/cram/build/CRAMReferenceRegion.java
```diff
@@ -113,9 +113,14 @@ public void fetchReferenceBases(final int referenceIndex) {
...
         if ((referenceIndex != this.referenceIndex) ||
                 regionStart != 0 ||
-                (regionLength < referenceBases.length)) {
+                (regionLength != newSequenceRecord.getSequenceLength())) {
             setCurrentSequence(referenceIndex);
```
droazen, line 124: Add a comment that the `setCurrentSequence()` call needs to happen first in this `if` block (in particular, before the assignment to `regionLength` below). Or alternatively, you could have `regionLength = newSequenceRecord.getSequenceLength();` below to eliminate the dependency on order of operations.

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
```diff
@@ -21,32 +21,72 @@ public class CRAMAllEncodingStrategiesTest extends HtsjdkTest {
...
+    public Object[][] defaultStrategyRoundTripTestFiles() {
         return new Object[][] {
                 { new File(TEST_DATA_DIR, "NA12878.20.21.1-100.100-SeqsPerSlice.500-unMapped.cram"),
-                        new File(TEST_DATA_DIR, "human_g1k_v37.20.21.1-100.fasta") },
+                        new File(TEST_DATA_DIR, "human_g1k_v37.20.21.1-100.fasta"),
+                        false, false },
+                { new File(TEST_DATA_DIR, "CEUTrio.HiSeq.WGS.b37.NA12878.20.21.10m-10m100.cram"),
+                        new File("src/test/resources/htsjdk/samtools/reference/human_g1k_v37.20.21.fasta.gz"),
+                        false, false },
+                { new File(TEST_DATA_DIR, "NA12878.unmapped.cram"),
```
droazen, line 33: Can you add one-line comments explaining the provenance of these three crams, and what they are testing?

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
```diff
@@ -21,32 +21,72 @@ public class CRAMAllEncodingStrategiesTest extends HtsjdkTest {
...
+                        new File(TEST_DATA_DIR, "human_g1k_v37.20.21.1-100.fasta"),
+                        false, false },
+                // generated with samtools 1.19 from the gatk bam file CEUTrio.HiSeq.WGS.b37.NA12878.20.21.bam
+                { new File(TEST_DATA_DIR, "CEUTrio.HiSeq.WGS.b37.NA12878.20.21.v3.0.samtools.cram"),
+                        new File("src/test/resources/htsjdk/samtools/reference/human_g1k_v37.20.21.fasta.gz"),
+                        true, false },
+                // these tests use lenient equality to only validate read names, bases and qual scores
+                { new File(TEST_DATA_DIR, "mitoAlignmentStartTestGATKGen.cram"),
+                        new File(TEST_DATA_DIR, "mitoAlignmentStartTest.fa"), true, false },
+                { new File(TEST_DATA_DIR, "mitoAlignmentStartTest.cram"),
```
droazen, line 43: Add comments explaining what these files are as well

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
```diff
@@ -21,32 +21,72 @@ public class CRAMAllEncodingStrategiesTest extends HtsjdkTest {
...
+                        new File(TEST_DATA_DIR, "mitoAlignmentStartTest.fa"), true, false },
+                // files created by rewriting the htsjdk test file src/test/resources/htsjdk/samtools/cram/mitoAlignmentStartTest.cram
+                // using code that replicates the first read (which is aligned to position 1 of the mito contig) either
+                // 10,000 or 20,000 times, to create a file with 2 or 3 containers, respectively, that have reads aligned to
```
droazen, line 47: Did you confirm via direct inspection of the file that it did in fact create 2 or 3 containers with reads mapped to position 1?

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
```diff
@@ -21,32 +21,72 @@ public class CRAMAllEncodingStrategiesTest extends HtsjdkTest {
     private static final File TEST_DATA_DIR = new File("src/test/resources/htsjdk/samtools/cram");
     private final CompressorCache compressorCache = new CompressorCache();
 
-    @DataProvider(name="roundTripTestFiles")
-    public Object[][] roundTripTestFiles() {
+    @DataProvider(name="defaultStrategyRoundTripTestFiles")
+    public Object[][] defaultStrategyRoundTripTestFiles() {
```
droazen, line 25: Can you explain why some of these test cases have to use lenient equality, while others don't?

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
```diff
@@ -91,7 +132,13 @@ public void assertRoundTripFidelity(
             final SAMRecordIterator sourceIterator = sourceReader.iterator();
             final SAMRecordIterator targetIterator = copyReader.getIterator();
             while (sourceIterator.hasNext() && targetIterator.hasNext()) {
-                if (emitDetail) {
+                if (lenientEquality) {
+                    final SAMRecord sourceRec = sourceIterator.next();
+                    final SAMRecord targetRec = targetIterator.next();
+                    Assert.assertEquals(targetRec.getReadName(), sourceRec.getReadName());
```
droazen, line 138: Can lenient mode also check the start/end positions of the read?

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
```diff
@@ -108,13 +155,17 @@ public void assertRoundTripFidelity(
         }
     }
 
-    private void assertRoundtripFidelityWithSamtools(final File sourceCRAM, final File referenceFile) throws IOException {
+    private void assertRoundtripFidelityWithSamtools(
+            final File sourceCRAM,
+            final File referenceFile,
+            final boolean lenientEquality,
+            final boolean emitDetail) throws IOException {
         if (SamtoolsTestUtils.isSamtoolsAvailable()) {
```
droazen, line 163: If samtools is not available, the test probably shouldn't just silently do nothing and succeed. Maybe throw a TestNG `SkipException` in that case?

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
```diff
@@ -21,32 +21,72 @@ public class CRAMAllEncodingStrategiesTest extends HtsjdkTest {
...
+            final boolean lenientEquality,
+            final boolean emitDetail) throws IOException {
         for (final Tuple<String, CRAMEncodingStrategy> testStrategy : getAllEncodingStrategies()) {
             final File tempOutCRAM = File.createTempFile("allEncodingStrategyCombinations", ".cram");
             tempOutCRAM.deleteOnExit();
             CRAMTestUtils.writeToCRAMWithEncodingStrategy(testStrategy.b, cramSourceFile, tempOutCRAM, referenceFile);
-            assertRoundTripFidelity(cramSourceFile, tempOutCRAM, referenceFile, false);
-            assertRoundtripFidelityWithSamtools(tempOutCRAM, referenceFile);
+            assertRoundTripFidelity(cramSourceFile, tempOutCRAM, referenceFile, lenientEquality, emitDetail);
+            assertRoundtripFidelityWithSamtools(tempOutCRAM, referenceFile, lenientEquality, emitDetail);
```
droazen, line 89: Add a comment explaining why we also do a test with samtools

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
```diff
@@ -21,32 +21,72 @@ public class CRAMAllEncodingStrategiesTest extends HtsjdkTest {
...
+                        new File(TEST_DATA_DIR, "mitoAlignmentStartTest.fa"), true, false },
+                { new File(TEST_DATA_DIR, "mitoAlignmentStartTest_3_containers_aligned_to_pos_1.cram"),
+                        new File(TEST_DATA_DIR, "mitoAlignmentStartTest.fa"), true, false }
         };
     }
 
-    @Test(dataProvider = "roundTripTestFiles")
-    public final void testRoundTripDefaultEncodingStrategy(final File sourceFile, final File referenceFile) throws IOException {
+    @Test(dataProvider = "defaultStrategyRoundTripTestFiles")
+    public final void testRoundTripDefaultEncodingStrategy(
```
droazen, line 57: Add a comment before this test method describing the purpose of the test (since we're not just testing "encoding strategies" here....)

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
```diff
@@ -21,32 +21,72 @@ public class CRAMAllEncodingStrategiesTest extends HtsjdkTest {
...
+    public Object[][] encodingStrategiesTestFiles() {
+        return new Object[][] {
+                { new File(TEST_DATA_DIR, "NA12878.20.21.1-100.100-SeqsPerSlice.500-unMapped.cram"),
+                        new File(TEST_DATA_DIR, "human_g1k_v37.20.21.1-100.fasta"), false, false },
+        };
     }
 
-    @Test(dataProvider = "roundTripTestFiles")
-    public final void testAllEncodingStrategyCombinations(final File cramSourceFile, final File referenceFile) throws IOException {
+    @Test(dataProvider = "encodingStrategiesTestFiles")
```
droazen, line 78: Would there be value in having this `testAllEncodingStrategyCombinations ` test use the `defaultStrategyRoundTripTestFiles` DataProvider as well, with its many additional test cases?

src/test/java/htsjdk/samtools/cram/ref/CRAMReferenceRegionTest.java
```diff
@@ -167,6 +168,39 @@ public void testGetReferenceBasesByRegionPositive(
         Assert.assertEquals(bases, Arrays.copyOfRange(fullContigBases, requestedOffset, requestedOffset + requestedLength));
     }
 
+    // simulate the state transitions that occur when writing a CRAM file
+    @Test
+    public void testSerialStateTransitions() {
```
droazen, line 173: Add a comment here explaining and referencing the bug that prompted us to write this test.

src/test/java/htsjdk/samtools/cram/ref/CRAMReferenceRegionTest.java
```diff
@@ -167,6 +168,39 @@ public void testGetReferenceBasesByRegionPositive(
...
+                new AlignmentContext(
+                        new ReferenceContext(CRAMStructureTestHelper.REFERENCE_SEQUENCE_ZERO),
+                        1,
+                        SHORT_FRAGMENT_LENGTH));
+        Assert.assertEquals(cramReferenceRegion.getRegionLength(), SHORT_FRAGMENT_LENGTH);
+
+        // now transition back to the full sequence
+        cramReferenceRegion.fetchReferenceBases(CRAMStructureTestHelper.REFERENCE_SEQUENCE_ZERO);
+        Assert.assertEquals(cramReferenceRegion.getRegionLength(), fullRegionFragmentLength);
+    }
```
droazen, line 202: Is `CRAMReferenceRegion` ever used across multiple threads? Give how the fetch methods update the internal state of the object on every call, it's clearly not safe to use in a parallelized manner. Do we need to make these methods `synchronized`?

droazen (changes requested):
@cmnbroad Back to you with my review comments

src/main/java/htsjdk/samtools/cram/build/CRAMReferenceRegion.java
cmnbroad: Oh yeah, done.

src/main/java/htsjdk/samtools/cram/build/CRAMReferenceRegion.java
cmnbroad: Latter is done.

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
cmnbroad: CRAMs don't always roundtrip with complete fidelity, especially when roundtripping through samtools, which resurrects NM/MD tags whether they were present in the original or not, unless they are specifically excluded. So in some cases when roundtripping, you can't use full SAMRecord comparisons. In those cases we use lenient equality. I've added this as a comment.

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
cmnbroad: Yes. You can also see this in the GATK PR, which uses files that were created using the same code that created these files, to test the detector on files with multiple bad containers.

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
cmnbroad: Its testing the default encoding strategy. Comment added.

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
cmnbroad: Perhaps, but there are currently 81 encoding strategies, so using all of those files would make this test take a really long time, and I would guess, dominate the CI test time, especially now that we have the large CRAM test file in that list. So instead I separated them this way.

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
cmnbroad: Its basically testing interoperability with samtools for all of the different encodings, to make sure both tool sets agree on the data. Added a comment saying its for interop testing.

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
cmnbroad: Yes, added.

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
cmnbroad: Makes sense. Done.

src/test/java/htsjdk/samtools/cram/ref/CRAMReferenceRegionTest.java
cmnbroad: It is currently not used anywhere across threads. As you noted, it's not thread-safe. I guess I could add `synchronized`, but that still wouldn't make it thread-safe, since the class is is stateful, and the usage pattern is `fetch...` followed by one or more `getCurrent..` methods. So synchronized won't fix the statefulness.

cmnbroad: Added a comment saying it's not thread-safe.

src/test/java/htsjdk/samtools/cram/ref/CRAMReferenceRegionTest.java
cmnbroad: Done.

src/test/java/htsjdk/samtools/CRAMAllEncodingStrategiesTest.java
cmnbroad: Done with as much specificity as I have about these, since they've been in the repo for a while.

src/main/java/htsjdk/samtools/cram/build/CRAMReferenceRegion.java
```diff
@@ -113,17 +115,18 @@ public void fetchReferenceBases(final int referenceIndex) {
 
         // Re-resolve the reference bases if we don't have a current region or if the region we have
         // doesn't span the *entire* contig requested.
+        final SAMSequenceRecord newSequenceRecord = getSAMSequenceRecord(referenceIndex);
         if ((referenceIndex != this.referenceIndex) ||
                 regionStart != 0 ||
-                (regionLength < referenceBases.length)) {
+                (regionLength != newSequenceRecord.getSequenceLength())) {
             setCurrentSequence(referenceIndex);
             referenceBases = referenceSource.getReferenceBases(sequenceRecord, true);
```
droazen, line 123: There is still a potential order-of-operations issue here: since the call to `getReferenceBases()` uses the instance variable `sequenceRecord`, the call to `setCurrentSequence()` must happen first. If the order of the statements were reversed in some future refactoring, we might have another terrible bug. Either add a comment explaining that `setCurrentSequence()` must happen first, or eliminate the order dependency by using `newSequenceRecord` in the `getReferenceBases()` call.

droazen (changes requested):
@cmnbroad Back to you with one final comment

PR #1709, opened 2024-05-15
SamUtils PairOrientation is confusingly ambiguous about paired reads needing to map to the same contig

yfarjoun:
- Improved documentation to clarify that PairOrientation (in SamUtils) pertains only to reads on the same contig.
- Now getPairOrientation will throw an exception if thrown when the two reads are mapped to different contigs
- Added tests to verify that the exception is thrown in the 4 cases specified.

### Description

- [x] Check your code style.

### DIFF
--- a/src/main/java/htsjdk/samtools/SamPairUtil.java
+++ b/src/main/java/htsjdk/samtools/SamPairUtil.java
@@ -62,7 +62,7 @@ public class SamPairUtil {
      * @param record the record for which the pair orientation is requested
      * @return PairOrientation of the given SAMRecord.
      * @throws IllegalArgumentException If the record is not a paired read, or
-     * one or both reads are unmapped, or is the two records do are not mapped to the
+     * one or both reads are unmapped, or if the two records are not mapped to the
      * same reference.
      *
      * NOTA BENE: This is NOT the orientation byte as used in Picard's MarkDuplicates. For that please look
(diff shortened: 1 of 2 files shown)


src/main/java/htsjdk/samtools/SamPairUtil.java
```diff
@@ -57,31 +59,41 @@ public static enum PairOrientation
 
     /**
      * Computes the pair orientation of the given SAMRecord.
-     * @param r
+     * @param record the record for which the pair orientation is requested
      * @return PairOrientation of the given SAMRecord.
      * @throws IllegalArgumentException If the record is not a paired read, or
-     * one or both reads are unmapped.
+     * one or both reads are unmapped, or is the two records do are not mapped to the
```
lbergelson, line 65: typo here: "or is the two records do are not"

lbergelson (commented):
@yfarjoun There's a typo.  This seems like a good improvement otherwise though.  Does this require corresponding changes in Picard to avoid hitting the new error case? 

An alternate solution would be to add a new state that captures cross chromosome pairs.

src/main/java/htsjdk/samtools/SamPairUtil.java
yfarjoun: thanks. fixed.

yfarjoun:
@lbergelson I don't think that it should be called with reads from different chromosomes. if is it, there's no meaning to RF,FR or TANDEM and so an exception seems (to me) to be the fastest way to find any problems. keeping it in a fourth meaningless state will either 
1. be a silent error (if that state isnt' checked)
2. explode in a different place due to failing expectations...

it would be good to see if picard builds (and tests pass) with this version of htsjdk.... want to hlep me with that?

yfarjoun:
🎉  thanks @lbergelson

PR #1711, opened 2024-06-25
reducing memory required in IntervalMergerIterator when not concatting names

meganshand:
When using `IntervalMergerIterator` to merge many small intervals over a whole contig it can take quite a bit of memory to store each interval. This can happen when trying to convert a GVCF with many reference blocks that are only a few bases in length into an interval list. The intervals are stored only to concat the names of each interval, which is not useful when merging so many intervals. This change keeps memory usage lower by not adding each interval that is iterated over to `toBeMerged` when `concatenateNames` is false.


meganshand:
@lbergelson can you please take a look when you have a moment?

kockan (approved):
LGTM

lbergelson (approved):
Looks good, thanks @meganshand

PR #1715, opened 2024-09-17
Make Snappy technically an optional dependency

lbergelson:
* Separate SnappyLoader into SnappyLoader and SnappyLoaderInternal
  This allows SnappyLoader to function (and correctly respond that snappy is not available)
  in the case where the Snappy library is completely missing.  Previously it identified and
  worked around cases where the native code was missing or incompatible, but the java code was required.

* Snappy is still marked as a dependency in gradle so downstream projects have to actively opt out of it.

* This is needed to work around notarization issues in the IGV mac builds which strongly discourage native code.

### DIFF
--- a/src/main/java/htsjdk/samtools/util/SnappyLoader.java
+++ b/src/main/java/htsjdk/samtools/util/SnappyLoader.java
@@ -25,24 +25,21 @@ package htsjdk.samtools.util;
 
 import htsjdk.samtools.Defaults;
 import htsjdk.samtools.SAMException;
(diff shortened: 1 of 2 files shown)


droazen (approved):
This looks fine to me @lbergelson , provided you've tested both the snappy and no-snappy cases.

lbergelson:
I tested a good combination of options.  IGV with Snappy, snappy excluded but not disabled, and snappy excluded and disabled.  

Thank you.

PR #1719, opened 2024-10-04
Replace toPath().toString() with getURIString()

lbergelson:
@cmnbroad This is a fix for an issue that @kshakir discovered.  It dropping the scheme because it's using toPath().toString() instead of either toPath().toUri().toString() or getURIString().  

I added a JIMFS test which fails without the changes.  In a second commit I refactored the code with the optionals a bit.  I think it's clearer.


cmnbroad (approved):
Looks good - thx for catching this.

PR #1721, opened 2024-10-08
Carrot setup

lbergelson:
PR to test setting up carrot.

@KevinCLydon I took your branch and added some stuff on it

1.  A dockerfile that just includes the htsjdk source so we can run tests 
2. A super basic test and eval wdl pair that doesn't actually use htsjdk at all
3. I registered them all in carrot.  

Now we need to:
1. do the magic binding between carrot/docker/wdl
2. make a real actual test that we can run in carrot


lbergelson:
@cmnbroad once this is wired up we need to talk with you about creating the specific cram test.  I'm not sure the best way to do it.  I think we could either write a normal htsjdk test which has a path that gets filled in by the carrot test somehow, or we could write a little executable that runs the test given an argument.   Either way wouldn't be too hard.

lbergelson:
#carrot

lbergelson:
@KevinCLydon I'm missing something, it's not using the result of the test workflow in the eval workflow when I invoke it manually.  Is this not the right magic?  test_output:test_workflow.result

lbergelson:
Wait, nevermind, I was confused about which test I ran.

cmnbroad:
Closing since I've cherry-picked everything to my working branch.

PR #1723, opened 2024-10-11
Adding additional java versions to our test matrix and toolchains

lbergelson:
* Add 17, 21, 22, 23 to our toolchains list
* Github actions tests on  17, 21, 23 (experimental)
* Set compiler release target to 17 so that libaries built with newer javas still target 17
* Update github actions to use actions v4 instead of deprecated v3 variants


lbergelson:
apparently toolchains in gradle don't work this way.

PR #1727, opened 2024-10-29
.vcf.bgz as a valid vcf extension in FileExtensions

lindenb:
### Description

VCF suppliers like gnomad, provide VCF with the vcf.bgz extension (eg https://storage.googleapis.com/gcp-public-data--gnomad/release/4.1/vcf/exomes/gnomad.exomes.v4.1.sites.chr1.vcf.bgz ) . This PR adds vcf.bgz as a valid extension, defined as  COMPRESSED_VCF_BGZ

- [X] Check your code style.


lbergelson (approved):
👍 Looks like an improvement to me.  My only comment is that we might want to make it a method "hasCompressedVcfExtension()" or something like that.   We can always revisit that in the future though.

PR #1743, opened 2025-04-14
Restore ReferenceSequenceFileFactory deleted in previous commit.

cmnbroad:


src/main/java/htsjdk/samtools/reference/ReferenceSequenceFileFactory.java
```diff
@@ -127,6 +127,29 @@ public static ReferenceSequenceFile getReferenceSequenceFile(final Path path, fi
...
+            try {
+                return IOUtil.isBlockCompressed(path, true) ? new BlockCompressedIndexedFastaSequenceFile(path) : new IndexedFastaSequenceFile(path);
+            } catch (final IOException e) {
+                throw new SAMException("Error opening FASTA: " + path, e);
+            }
+        } else {
+            return new FastaSequenceFile(path, truncateNamesAtWhitespace);
+        }
+    }
+
```
cmnbroad, line 152: Maybe this can just delegate to the new method.

cmnbroad: Fixed.

PR #1745, opened 2025-05-09
feat: Add sanity check to indexed FASTA file

yfarjoun:
The last position in the fasta file (according ot the index) must be close to the size of the fasta file itself. A mismatch could indicate a corrupt fasta file, or an in correct index.

(tests included)

### Description

- [ ] Check your code style.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved subsequence byte-range calculations by using the index’s computed offsets directly.
  - Strengthened indexed FASTA opening with additional consistency validation, rejecting mismatched or truncated FASTA/index pairs while still allowing valid trailing whitespace and files without a final newline.
- **Tests**
  - Added FASTA fixtures and expanded scenarios covering extra whitespace, header whitespace, and CRLF variants.
  - Added test coverage for index renaming behavior plus sanity-check cases for mismatched indexes, extra whitespace acceptance, and truncated-file rejection.


src/main/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFile.java
```diff
@@ -146,6 +147,41 @@ protected static void sanityCheckDictionaryAgainstIndex(final String fastaFile,
...
+     */
+    public static void sanityCheckFastaAgainstIndex(final String fastaFile,
+                                                    final FastaSequenceIndex FastaSequenceIndex) {
+
+
+        final Iterator<FastaSequenceIndexEntry> iterator = FastaSequenceIndex.iterator();
+        FastaSequenceIndexEntry fastaSequenceIndex = null;
+        while (iterator.hasNext()) {
+            fastaSequenceIndex = iterator.next();
+        }
```
tfenne, line 166: It kinda sucks to have to do this iteration.  I wonder if it would make sense to expand the PR a little bit, and modify `FastaSequenceIndex` to either store index entries in a pair of (Array/ArrayList, HashMap (instead of LinkedHashMap))?  Or just to hold an extra pointer to the final entry in the index?

src/main/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFile.java
```diff
@@ -146,6 +147,41 @@ protected static void sanityCheckDictionaryAgainstIndex(final String fastaFile,
...
+        assert fastaSequenceIndex != null;
```
tfenne, line 167: Asserts can be disabled at runtime - are you ok with that, or do you want to always check this?

src/main/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFile.java
```diff
@@ -146,6 +147,41 @@ protected static void sanityCheckDictionaryAgainstIndex(final String fastaFile,
...
 
+    /** Do some basic checking to make sure the fasta and the index match.
+     * <p>
+     * checks that the length of the fasta file is at least as long as the index proclaims
+     * and that beyond the last position references in the index there is only one line followed by whitespaces
+     *
+     * @param fastaFile Used for error reporting only.
+     * @param index index file to check against the dictionary.
+     */
+    public static void sanityCheckFastaAgainstIndex(final String fastaFile,
```
tfenne, line 158: Why String and not Path or File?

src/main/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFile.java
```diff
@@ -146,6 +147,41 @@ protected static void sanityCheckDictionaryAgainstIndex(final String fastaFile,
...
+        final long lastSequenceEnd = lastSequenceStart + fastaSequenceIndex.getOffset(lastSequenceLength);
+
+        final long fastaLength = new File(fastaFile).length();
+
+        //Question: should we worry about files with lots of whitespace in their end?
+        if (lastSequenceEnd > fastaLength) {
+            throw new IllegalArgumentException("The fasta file is shorter (%d) than its index claims (%d). Please reindex the fasta.".formatted(fastaLength, lastSequenceEnd));
+        }
+        // not sure why need to add 1 here.
+        if (lastSequenceEnd + fastaSequenceIndex.getTerminatorLength() + 1 < fastaLength) {
```
tfenne, line 179: Honestly I wonder if we should allow for more than 1?  What if there are a handful of blank lines at the end?  Maybe we could try a nested solution, that _if_ the file is longer than you have here ... then we read the content after the last base and ensure it's all whitespace?

src/main/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFile.java
```diff
@@ -146,6 +147,41 @@ protected static void sanityCheckDictionaryAgainstIndex(final String fastaFile,
...
+        FastaSequenceIndexEntry fastaSequenceIndex = null;
+        while (iterator.hasNext()) {
+            fastaSequenceIndex = iterator.next();
+        }
+        assert fastaSequenceIndex != null;
+        final long lastSequenceLength = fastaSequenceIndex.getSize();
+        final long lastSequenceStart = fastaSequenceIndex.getLocation();
+        final long lastSequenceEnd = lastSequenceStart + fastaSequenceIndex.getOffset(lastSequenceLength);
+
+        final long fastaLength = new File(fastaFile).length();
```
tfenne, line 172: Is this safe to do?  Does `AbstractIndexedFastaSequenceFile` support or have sub-classes that support URL based access (like http:// or ftp://)?  If so ... is there a method to call to get the file/object length?

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexEntry.java
```diff
@@ -112,6 +112,26 @@ public int getSequenceIndex() {
...
+    /** Return the offset to pos as determined by the number of bases and bytes per line
+     *
+     * @param pos the (1-based) position in the contig that is requested
+     * @return the offset (0-based) from 'location' where pos is located in the file.
+     */
+    public long getOffset(long pos) {
+        final int basesPerLine = this.getBasesPerLine();
+        final int bytesPerLine = this.getBytesPerLine();
+
+        return ((pos - 1) / basesPerLine) * bytesPerLine + (pos - 1) % basesPerLine;
```
tfenne, line 124:
```suggestion
        return ((pos - 1) / basesPerLine) * bytesPerLine + ((pos - 1) % basesPerLine);
```

Just for clarity.

src/main/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFile.java
yfarjoun: There are several operations that explicitly take advantage of the linked-list aspect...come to think of it, I my sanity-check only works on a "freshly make" object, since operations such as "rename" remove and add entries, which moved the renamed entry to the "end" of the linked-list.... I only _need_ it to work upon initialization, but I need to add some protections. upshot is that a list would be overkill, I'll go with the reference to the last element.

yfarjoun:
Thanks @tfenne. I responded to all your comments (I think!) please re-review.

BTW, the test failures were due to the FTP site not responding.

src/main/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFile.java
```diff
@@ -146,6 +147,55 @@ protected static void sanityCheckDictionaryAgainstIndex(final String fastaFile,
...
+                                             final FastaSequenceIndex fastaSequenceIndexes) throws IOException {
+
+        final FastaSequenceIndexEntry lastSequenceIndex = fastaSequenceIndexes.getIndexEntry(fastaSequenceIndexes.getLastSequence());
+
+        final long lastSequenceLength = lastSequenceIndex.getSize();
+        final long lastSequenceStart = lastSequenceIndex.getLocation();
+        final long lastSequenceEnd = lastSequenceStart + lastSequenceIndex.getOffset(lastSequenceLength);
+
+        final long fastaLength = Files.size(fastaFile);
+        //Question: should we worry about files with lots of whitespace in their end?
```
tfenne: I wouldn't worry about "lots of whitespace" at the end.  I think that is still technically valid fasta.

src/main/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFile.java
```diff
@@ -146,6 +147,55 @@ protected static void sanityCheckDictionaryAgainstIndex(final String fastaFile,
...
+
+            while (posOfInterest < fastaLength) {
+                channelBuffer.clear();
+                readFromPosition(channelBuffer, posOfInterest);
+                for (int i = 0; i < channelBuffer.position(); i++) {
+                    byte b = channelBuffer.get(i);
+                    if (!Character.isWhitespace((char) b)) {
+                        throw new IllegalArgumentException(
+                                ("The fasta file %s is too long (relative to the index). In particular has a non-whitespace " +
+                                        "character (%c) as a position too great (%d) given the claims of its index (%d)." +
```
tfenne:
```suggestion
                                        "character (%c) at a position (%d) beyond the last base according to its index (%d)." +
```

src/main/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFile.java
```diff
@@ -146,6 +147,55 @@ protected static void sanityCheckDictionaryAgainstIndex(final String fastaFile,
...
+                                        " Please reindex the fasta.")
+                                        .formatted(fastaFile.getFileName(), (char) b, posOfInterest + i, lastSequenceEnd));
```
tfenne: I would emit the absolute path here

tfenne (approved):
LGTM - should anyone else review?

yfarjoun:
@lbergelson is @tfenne 's review enough here? who else needs to review?

src/main/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFile.java
```diff
@@ -146,6 +147,55 @@ protected static void sanityCheckDictionaryAgainstIndex(final String fastaFile,
...
+        final FastaSequenceIndexEntry lastSequenceIndex = fastaSequenceIndexes.getIndexEntry(fastaSequenceIndexes.getLastSequence());
+
+        final long lastSequenceLength = lastSequenceIndex.getSize();
+        final long lastSequenceStart = lastSequenceIndex.getLocation();
+        final long lastSequenceEnd = lastSequenceStart + lastSequenceIndex.getOffset(lastSequenceLength);
+
+        final long fastaLength = Files.size(fastaFile);
+        
+        if (lastSequenceEnd > fastaLength) {
+            throw new IllegalArgumentException("The fasta file (%s) is shorter (%d) than its index (%s) claims (%d). Please reindex the fasta.".formatted(fastaFile.toUri().toString(),fastaLength, index.toString(), lastSequenceEnd));
```
lbergelson: IllegalArgument seems like the wrong exception here.   I might declare a new `IncompatibleIndexException` or something like that, it would make it easier for calling code to deal with this.  If for example someone wanted to automatically reindex on a failure.

src/main/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFile.java
```diff
@@ -146,6 +147,55 @@ protected static void sanityCheckDictionaryAgainstIndex(final String fastaFile,
...
+                    byte b = channelBuffer.get(i);
+                    if (!Character.isWhitespace((char) b)) {
+                        throw new IllegalArgumentException(
+                                ("The fasta file %s is too long (relative to the index). In particular has a non-whitespace " +
+                                        "character (%c) at a position (%d) beyond the last base (%d), according to its index (%s)." +
+                                        " Please reindex the fasta.")
+                                        .formatted(fastaFile.toUri().toString(), (char) b, posOfInterest + i, lastSequenceEnd,index.toString()));
+                    }
+                }
+                posOfInterest += channelBuffer.limit();
```
lbergelson: I think there might be minor bug here.  When you clear the buffer it should set limit = capacity, but I don't think readFromPosition guarantees that the as many bytes as possible are read, it's possible for it to return fewer than requested.  I think this might have the effect of skipping parts of the fasta file on incomplete reads. 

Setting it with position or capturing the return value from readFromPosition would both do the right thing I think.  

Unlikely to be a real issue since the bug would be to sometimes fail to detect a mismatch if there are characters embedded partway through a block of white space.

src/main/java/htsjdk/samtools/reference/AbstractIndexedFastaSequenceFile.java
```diff
@@ -146,6 +147,55 @@ protected static void sanityCheckDictionaryAgainstIndex(final String fastaFile,
...
 
+    /** Do some basic checking to make sure the fasta and the index match.
+     * <p>
+     * checks that the length of the fasta file is at least as long as the index proclaims
+     * and that beyond the last position references in the index there is only one line followed by whitespaces
+     *
+     * @param fastaFile Path to fasta file
+     * @param fastaSequenceIndexes index file to check against the fasta file.
+     *
+     * @throws IOException in case of io-error when reading fastaFile
```
lbergelson: This should say what it throws if it detects and error.

src/main/java/htsjdk/samtools/reference/FastaSequenceIndexEntry.java
```diff
@@ -112,6 +112,26 @@ public int getSequenceIndex() {
         return sequenceIndex;
     }
 
+    /** Return the offset to pos as determined by the number of bases and bytes per line
+     *
+     * @param pos the (1-based) position in the contig that is requested
+     * @return the offset (0-based) from 'location' where pos is located in the file.
+     */
+    public long getOffset(long pos) {
```
lbergelson, line 116: 👍

src/main/java/htsjdk/samtools/reference/FastaSequenceIndex.java
```diff
@@ -233,4 +241,13 @@ public Iterator<FastaSequenceIndexEntry> iterator() {
     public int size() {
         return sequenceEntries.size();
     }
+
+
+    /**
+     * @return The name of the last entry that was added when parsing the index file. Only guarranteed to be correct just
+     * after initialization. Protected for access from AbstractIndexedFastaSequenceFile.
+     */
+    protected String getLastSequence() {
```
lbergelson: I don't love this with the caveats but it's fine if it's clearly labelled.

src/main/java/htsjdk/samtools/reference/FastaSequenceIndex.java
```diff
@@ -80,13 +81,15 @@ public FastaSequenceIndex( Path indexFile ) {
...
-        parseIndexFile(in);
+        lastSequence=parseIndexFile(in);
     }
 
     /**
-     * Empty, protected constructor for unit testing.
+     * Empty, protected constructor for unit testing. Use with care, lastSequence will be incorrect.
      */
-    protected FastaSequenceIndex() {}
+    protected FastaSequenceIndex() {
```
lbergelson: I wonder if this is really only used for unit testing.

lbergelson (commented):
This looks sane to me I think.  I need to convince myself that it works for fasta.gz though.  If it doesn't it *should* be failing tests here so I'm assuming it works out somehow.  I have a few comments.  I started it running against GATK's tests to see if it hits any weirrd problems I didn't think of.   https://github.com/broadinstitute/gatk/pull/9179

lbergelson:
Hmn, I see a test failure that's not the broken FTP ones.  

```

Gradle suite > Gradle test > htsjdk.samtools.util.SequenceUtilTest > testCalculateNmTag FAILED
    java.lang.IllegalArgumentException: The fasta file file:///home/runner/work/htsjdk/htsjdk/src/test/resources/htsjdk/samtools/SequenceUtil/reference_with_lower_and_uppercase.fasta is too long (relative to the index). In particular has a non-whitespace character (a) at a position (44) beyond the last base (44), according to its index (htsjdk.samtools.reference.FastaSequenceIndex@1f). Please reindex the fasta.
        at htsjdk.samtools.reference.AbstractIndexedFastaSequenceFile.sanityCheckFastaAgainstIndex(AbstractIndexedFastaSequenceFile.java:190)
        at htsjdk.samtools.reference.IndexedFastaSequenceFile.sanityCheckFastaAgainstIndex(IndexedFastaSequenceFile.java:47)
        at htsjdk.samtools.reference.IndexedFastaSequenceFile.<init>(IndexedFastaSequenceFile.java:84)
        at htsjdk.samtools.reference.IndexedFastaSequenceFile.<init>(IndexedFastaSequenceFile.java:116)
        at htsjdk.samtools.reference.ReferenceSequenceFileFactory.getReferenceSequenceFile(ReferenceSequenceFileFactory.java:144)
        at htsjdk.samtools.reference.ReferenceSequenceFileFactory.getReferenceSequenceFile(ReferenceSequenceFileFactory.java:175)
        at htsjdk.samtools.reference.ReferenceSequenceFileFactory.getReferenceSequenceFile(ReferenceSequenceFileFactory.java:105)
        at htsjdk.samtools.reference.ReferenceSequenceFileFactory.getReferenceSequenceFile(ReferenceSequenceFileFactory.java:93)
        at htsjdk.samtools.reference.ReferenceSequenceFileFactory.getReferenceSequenceFile(ReferenceSequenceFileFactory.java:82)
        at htsjdk.samtools.util.SequenceUtilTest.testCalculateNmTag(SequenceUtilTest.java:496)
```

lbergelson:
It looks like `FastaSequenceIndexCreator` uses the no args constructor so now it causes explosions.  [Example](https://storage.googleapis.com/hellbender-test-logs/build_reports/9179/merge_15120705234.0/tests/test/classes/org.broadinstitute.hellbender.BwaMemIntegrationTest.html#testChimericUnpairedMapping)

I think the lastSequence cache needs to be kept up to date or just dynamically checked.  I assume there was a performance reason for caching it?

tfenne:
@coderabbitai review

PR #1758, opened 2026-03-24
Fix SamLocusIterator so that read position is not incorrectly offset …

tfenne:
…after transiting insertions that fail the base quality check.

### Description

This fixes a bug in SamLocusIterator where during accumulation of indels, the position in the read (`readBase`) was incorrectly failling to count indels that failed a base quality check.  The result of this is that indels later in the read were then accumulated with incorrect read offsets.

This affects any tool that uses the pileup engine with a base quality filter _and_ cares about indels!

The added test fails without the accompanying change to SamLocusIterator.

### DIFF
--- a/src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
+++ b/src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
@@ -813,4 +813,43 @@ public class SamLocusIteratorTest extends AbstractLocusIteratorTestTemplate {
         }
     }
 
+    /**
+     * Test that when an insertion fails the base quality check, readBase is still
+     * advanced so that subsequent insertions in the same read are reported at the
+     * correct offset. Previously, failing the BQ check left readBase un-advanced,
+     * causing all subsequent CIGAR positions to be misaligned.
+     *
+     * CIGAR: 10M1I1M1I10M (read length 23)
+     * Quality string assigns BQ=5 to the first inserted base (pos 10) and BQ=30
+     * to the second inserted base (pos 12). With a cutoff of 20, only the second
+     * insertion should be reported, and it should be at read offset 12 (not 11).
+     */
+    @Test
+    public void testInsertionBqFailureDoesNotShiftSubsequentOffsets() {
+        final SAMRecordSetBuilder builder = getRecordBuilder();
+        final int startPosition = 165;
+
+        // Build quality string: 23 bases, all BQ=30 except position 10 (first insertion) = BQ=5
+        final char[] quals = new char[23];
+        java.util.Arrays.fill(quals, '?');  // '?' = ASCII 63 - 33 = BQ 30
+        quals[10] = '&';                     // '&' = ASCII 38 - 33 = BQ 5
+
+        builder.addFrag("record0", 0, startPosition, false, false,
+                "10M1I1M1I10M", new String(quals), 30);
+
+        final SamLocusIterator sli = createSamLocusIterator(builder);
+        sli.setIncludeIndels(true);
+        sli.setQualityScoreCutoff(20);
+
+        for (final SamLocusIterator.LocusInfo li : sli) {
+            for (final SamLocusIterator.RecordAndOffset rao : li.getInsertedInRecord()) {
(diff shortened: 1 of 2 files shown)


src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -813,4 +813,43 @@ public void testOverlappingGappedAlignmentsWithIndels() {
         }
     }
 
+    /**
+     * Test that when an insertion fails the base quality check, readBase is still
+     * advanced so that subsequent insertions in the same read are reported at the
+     * correct offset. Previously, failing the BQ check left readBase un-advanced,
+     * causing all subsequent CIGAR positions to be misaligned.
```
yfarjoun:
```suggestion
     * correct offset. 
```

yfarjoun: "previously" only makes sense in a commit or PR, not in a docstring for a test.

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -813,4 +813,43 @@ public void testOverlappingGappedAlignmentsWithIndels() {
...
+     *
+     * CIGAR: 10M1I1M1I10M (read length 23)
+     * Quality string assigns BQ=5 to the first inserted base (pos 10) and BQ=30
+     * to the second inserted base (pos 12). With a cutoff of 20, only the second
+     * insertion should be reported, and it should be at read offset 12 (not 11).
```
yfarjoun:
```suggestion
     * insertion should be reported, and it should be at read offset 12.
```

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -813,4 +813,43 @@ public void testOverlappingGappedAlignmentsWithIndels() {
...
+                "10M1I1M1I10M", new String(quals), 30);
+
+        final SamLocusIterator sli = createSamLocusIterator(builder);
+        sli.setIncludeIndels(true);
+        sli.setQualityScoreCutoff(20);
+
+        for (final SamLocusIterator.LocusInfo li : sli) {
+            for (final SamLocusIterator.RecordAndOffset rao : li.getInsertedInRecord()) {
+                // The second insertion's first base is at read offset 12.
+                // Before the fix, the drifted readBase would report offset 11.
```
yfarjoun:
```suggestion

```

yfarjoun: [suggestion](https://pullpo.io/cc?l=suggestion) 
add a comment that the first insertion will not be caught here because it's low quality.

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -813,4 +813,43 @@ public void testOverlappingGappedAlignmentsWithIndels() {
...
+                Assert.assertEquals(rao.getOffset(), 12,
```
yfarjoun: please assert that we got to here (i.e. that the loops were not all trivially empty)

yfarjoun (changes requested):
[chore](https://pullpo.io/cc?l=chore) 
please add an assert that "proves" that the code evaluates the assertion on L848.

Also a few nits to make the documentation less time-dependent.

src/test/java/htsjdk/samtools/util/SamLocusIteratorTest.java
```diff
@@ -813,4 +815,45 @@ public void testOverlappingGappedAlignmentsWithIndels() {
         }
     }
 
+    /**
+     * Test that when an insertion fails the base quality check, readBase is still
+     * advanced so that subsequent insertions in the same read are reported at the
+     * correct offset.
```
nh13: **nitpick (non-blocking):** 
There's one, set for stun:
```suggestion
     * correct offset.
```

nh13 (approved):

PR #1759, opened 2026-03-27
Fix up unit tests so that gradle correctly reports results (no more erroneous skips)

tfenne:
### Description

I got tired of gradle reporting incorrect test results, so this PR is intended to fix that, and enable reasonable parallelization of the tests.  The crux is that parallelization at the method level with TestNG and gradle's test infrastructure seem to be incompatible, and this is known issue that is still not resolved in newer versions of gradle and/or TestNG.

The general solution is to switch to "class" level parallelization, which does work (though i took a detour through forked VMs to get here).  Upon fixing the reporting it was clear there were other issues being masked by this, which I've attempted to fix here too.

Issues fixed include:
- Flagging tests that require a local htsget server and disabling them by default
- Auto-detecting samtools for tests that need it, with the ENV var becoming an override to auto-detection, not the only mechanism
- Fixed several tests that had concurrency bugs

- [x] Check your code style.


src/test/java/htsjdk/samtools/cram/structure/CRAMRecordTestHelper.java
```diff
@@ -14,7 +14,7 @@ public class CRAMRecordTestHelper {
 
     final SAMRecordSetBuilder recordSetBuilder = new SAMRecordSetBuilder();
 
-    @BeforeTest
+    @BeforeClass
```
tfenne: All of these changes are because after the removal of the testng.xml, TestNg was trying to run these `BeforeTest` before *every* test in the whole test suite, not just this class.

tfenne:
FYI I also removed the CI task to test FTP as the FTP site no longer exists, and fixed up the testExternalAPIs CI workflow that was also running FTP instead of the other external API tests.

src/test/java/htsjdk/samtools/filter/IntervalKeepPairFilterTest.java
```diff
@@ -16,7 +16,7 @@ public class IntervalKeepPairFilterTest extends HtsjdkTest {
     private static final int READ_LENGTH = 151;
     private final SAMRecordSetBuilder builder = new SAMRecordSetBuilder();
 
-    @BeforeTest
+    @BeforeClass
```
yfarjoun: but don't we want to make sure that every test starts with the same builder? I worry that one of the tests (now or in the future) will modify the builder which will affect the tests that come after it (in the same class). Is that a valid concern?

src/test/java/htsjdk/samtools/util/BlockCompressedInputStreamTest.java
```diff
@@ -160,51 +160,30 @@ public Object[][] customInflateInput() throws IOException {
...
     public void testCustomInflater(final CheckedExceptionInputStreamSupplier bcisSupplier,
                                    final QuietTestWrapper<List<String>> expectedOutputSupplier,
-                                   final int expectedInflateCalls,
-                                   final InflaterFactory customDefaultInflaterFactory) throws Exception
+                                   final int expectedInflateCalls) throws Exception
     {
-        // clear inflate call counter in CountingInflater
         CountingInflater.inflateCalls = 0;
 
-        // If requested, set the global default InflaterFactory to a custom factory. Otherwise, set it to the default.
```
yfarjoun, line 185: no more testing of custom inflater factory?

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
```diff
@@ -245,7 +245,7 @@ public void samRecordFactoryNullHeaderBAMTest() {
     public Object[][] composeAllPermutationsOfSamInputResource() {
         final List<SamInputResource> sources = new ArrayList<>();
         for (final InputResource.Type dataType : InputResource.Type.values()) {
-            if (dataType.equals(InputResource.Type.SRA_ACCESSION))
+            if (dataType.equals(InputResource.Type.SRA_ACCESSION) || dataType.equals(InputResource.Type.HTSGET))
```
yfarjoun, line 248: how did this pass before?

src/test/java/htsjdk/utils/TestNGUtils.java
```diff
@@ -22,13 +23,28 @@ public class TestNGUtils {
...
+     */
+    public static Iterator<Object[]> getDataProviders(final String packageName, final Set<String> excludedGroups) {
         List<Object[]> data = new ArrayList<>();
         final ClassFinder classFinder = new ClassFinder();
         classFinder.find(packageName, Object.class);
 
         for (final Class<?> testClass : classFinder.getClasses()) {
             if (Modifier.isAbstract(testClass.getModifiers()) || Modifier.isInterface(testClass.getModifiers()))
                 continue;
+            if (hasExcludedGroup(testClass, excludedGroups))
```
yfarjoun, line 46: I"m not sure I'm following the logic, but what if one data provider is used in an excluded group but _also_ in a non-excluded group? seems like the dataprovider will be ... not provided?

yfarjoun (changes requested):
[praise](https://pullpo.io/cc?l=praise) 
thanks for doing this... just a few questions.

src/test/java/htsjdk/samtools/util/BlockCompressedInputStreamTest.java
tfenne: The custom inflator factories are still tested (see lines 163-166).  All this does is move change things so that the custom factory is set at construction vs. via a global static "default" factory, which isn't threadsafe in the tests.

src/test/java/htsjdk/samtools/SamReaderFactoryTest.java
tfenne: It seem like on CI, there's a process to fire up a local htsget server, which does not happen when you run tests locally.  This was failing for me, masked by the whole testng/gradle skip issue.

src/test/java/htsjdk/utils/TestNGUtils.java
tfenne: This is used exclusively by `TestDataProviders`, not directly by any tests (TestNG hooks those up).  The goal here is to avoid testing DataProviders that are in test classes where the whole class is annotated with a test group that is excluded in the current test execution.  I.e. if we're not testing `sra` then don't test data providers in tests annotated as `sra` (because they might depend on things that won't work and that's why that group's disabled).

src/test/java/htsjdk/samtools/filter/IntervalKeepPairFilterTest.java
tfenne: Yeah - most of the tests that used these are read only.  And hooh-boy I'd forgotten how confusing TestNG's annotations are.  `@BeforeTest` is a footgun most of the time, and I think `@BeforeMethod` is probably what I would think of as "run this before every test method".  I've switched to that, and fixed where repeated running of the method causes problems (i.e. adding to builders without re-initializing the builder first).

src/test/java/htsjdk/samtools/HtsgetBAMFileReaderTest.java
```diff
@@ -45,7 +46,7 @@ public void init() throws IOException {
         Assert.assertFalse(bamFileReaderHtsgetAsync.isUsingPOST());
     }
 
-    @AfterTest
+    @AfterClass
```
yfarjoun: since the readers are opened with `@BeforeMethod` shouldn't be closed with `@AfterMethod` rather than `@AfterTest`?

src/test/java/htsjdk/utils/TestNGUtils.java
yfarjoun: :lol: I think that I wrote that test.... 👍

yfarjoun (commented):
[question](https://pullpo.io/cc?l=question) 
just one point about closing readers.

yfarjoun (approved):

PR #1762, opened 2026-03-30
Changed SAMRecord.toString() to emit the SAM format string with all fields

tfenne:
### Description

SAMRecord has historically output an abbreviated format in toString().  I think more often than not this is confusing to users, since a SAMRecord has a well defined String format.  Moreover the toString() method is used extensively by TestNG to display when two SAMRecord instances are not equal - and the vast majority of the time the toString() output did not contain the differences.

- [x] Check your code style.


nh13 (commented):
1. can you verify there's no newline, and if there is, strip it (he says knowingly)
2. are there tools (picard, igv, gatk, fgbio, others) that rely on the `toString` returned format (I know I know), but it is a public so a breaking change...
3. Should you add at least one regression test?
4. `toString` was previously lock free, but looking at getSAMString we have some `synchronized` methods...
5. the previous method was lighter-weight, is this a concern?
6. are there tests that (shudders) compared outputs using `toString`?
7. is it time to remove the `format` deprecated method?

tfenne:
Thanks for the review @nh13.  I have done several things based upon it:
- Removed the newline from the toString() / getSAMString() path and fixed up anywhere in htsjdk that was using those methods
- Added tests for the new behavior to SAMRecord.toString()
- Removed the deprecated format() and friends
- Removed the locking by allocating a SAMTextWriter each time in getSAMString - it turns out to be very lightweight so it should be just fine

PR #1739, opened 2025-03-24
Wire up CRAM 3.1 codecs for writing and a naive write profile.

cmnbroad:
Depends on https://github.com/samtools/htsjdk/pull/1736. Draft for now.

PR #1763, opened 2026-03-30
Full CRAM 3.1 support & performance tuning of CRAM read/write

tfenne:
### Description

Ok, so this is a monster PR.  I'm not even sure it should be a single PR, but I wanted to put it up here to get some feedback on how best to proceed.  I've created it as a _draft_ for that reason. It builds upon the work by @cmnbroad in #1739 while implementing the necessary additions to write CRAM 3.1 at comparable compression ratios to htslib/samtools.

This was done with the assistance of Claude Code, though it also represents a significant investment of my own personal time, prompting, reviewing, fixing, redirecting. 

The PR breaks down roughly as follows:
- 8e93d8b6019311a04e99b4cb409a027d6186fcd5..a174cb7f3d61dee957ea7c172ed4f709a3c610a6: initial work by Chris
- 243059c46300e73af31a506fb28fefd7beffd747..673ab5ff76bfb3c0458f6aea0da81a6a66c5fa56 mostly rounding out the CRAM 3.1 functional implementation
- 140ac3d9ddb16a0251f31db21c7566e44e1bd611..bf21087b549cf8ddc83719724509501ac8a1c44f improvements focused on compression gains and making it easier to compare with htslib
- e2b07b416c0e82193bacdecca9419d22e2f6b2fc..f5dc3c42ae35488b287fa3c44e17d08000d4895d runtime performance tuning

Many of the additions and changes are strongly inspired by work in htslib!

Currently with the `normal` profile selected, htsjdk emits CRAMs approximately similar in size to htslib.  With some 1KG low-pass data I've been testing with the results are actually slightly smaller.  However runtime to generate the CRAMs is ~2x slower than htslib/samtools.  I think that gap could be shrunk further (it started out around 5-6x slower), I think there will always be a substantial gap due to a) not having easy access to SIMD instructions and b) not having a good `libdeflate` wrapper on the JVM.

The read path for CRAM 3.1 predates this PR, but the changes here provide substantial performance improvements.  Again, testing on low-pass WGS data I'm seeing a 40-50% speedup.  

Assuming this PR/changeset has legs, I will do some broader benchmarking and report on both compression and runtime performance.

@jkbonfield I wouldn't expect you to review this PR, but if you have thoughts on how to validate the _results_ of this work beyond the broad set of CRAM unit tests in the htsjdk repo, I would be very grateful for your input.

- [x] Check your code style.


jkbonfield:
For testing, hts-specs has a corpus of test files designed to validate CRAM decoder implementations.

https://github.com/samtools/hts-specs/tree/master/test/cram

The 3.0/README.md covers the basics of the CRAM structure, adding in feature at a time.  So blank file with header, then basic one-read bits without reference based encoding, etc.  This is mostly testing structures, features, encodings, and so on.  By the end of it it's also testing compression codecs.

The 3.1 directory has 4 CRAM files with progressive levels of new compression codecs, basically htslib's profiles I think.  They're whole file real world examples, as beyond the new codecs there are no format changes.

More useful for 3.1 is the codecs directory which have raw codec compressed byte streams for you to decode and validate.  These are shrunk down copies of the data in https://github.com/jkbonfield/htscodecs-corpus/tree/master/dat, although I think that may be a little incomplete now.  Basically it's designed to test things like rans, fqz, arith and name tokeniser codecs in isolation, for unit testing purposes.

This is mostly just integration testing.  The nature of CRAM is that even with the same codec we could produce subtly different encoded byte streams (eg it's up to the implementation how it approximates the probabilities tables in rANS and handles rounding errors when the frequencies don't divide up perfectly, but the data stream contains the necessary lookup tables to decode so that's what matters).

So what we need for testing are:

- Can we round-trip our own data without losing information
- Can we round-trip with other implementations;  java enc -> htscodecs dec, or htscodecs enc -> java dec.  The test data here is attempting to validate that bit.

jkbonfield:
Some comments regarding CRAM 4.0.    This is long, but it may explain why I think CRAM 4 is now something to ignore.  (Mainly speaking to you as someone looking at adding it for Noodles).

Personally, I think I made an error when I added it to htslib, but the history here is important to understand how we got to where we did.

Initially I had two aims to improve on CRAM 3.0:
- Fix some glaring issues in the CRAM *format*.
- Improve CRAM *compression* and if possible also speed.
- A third bonus was to simplify where possible (eg cull the kitchen sink of CORE bit-wise encodings).

Fixing the CRAM format included things like replacing ITF8/LTF8 with a true universal variable length encoder that doesn't need to know the expected bit-size (somewhat of an anathema to proper variable length encoding).  It also separated signed from unsigned values, so we no longer take 5 bytes to store -1 (or worse for LTF8).  It also gained better ways of doing constant values.  I also fixed some deficiencies such as how to discover whether the original data had MD/NM (by adding MD:* and NM:* tags with the "*" being auto-generated on decode, so the presence and location are known).  Further improvements would be around =/X vs M in cigar, and so on.  There are quite a few little improvements to make it work better as a round-trip tool.  I added a [CRAM v4 idea tracker](https://github.com/samtools/hts-specs/issues/144) on hts-specs not just because I was hoping for community ideas, but primarily as a way to date stamp the ideas I had and prevent patent squatters like MPEG from patenting obvious future improvements!

Looking at the compression side, I regret not adding in zstd, but it was a bit too new then.  Zstd should totally replace Gzip (deflate) and LZMA.  Deflate is just hampered by a maximum window size of 32kb, which makes deduplication - the core part of LZ - fail on long reads and similar data types.    At high levels zstd approaches lzma for ratio.  It's not quite as good, but close, and frankly lzma is almost never worth the CPU cost so I doubt it's used much.   Similarly bzip2 should be ditched in favour of libbsc, which betters it on pretty much every aspect.   Combined we'd be reducing the number of dependencies while also using more modern high performant libraries.  There's even a [branch of io_lib](https://github.com/jkbonfield/io_lib/tree/cram_modules) with libbsc and zstd.  This was [demonstrated to MPEG at a conference](https://drive.google.com/file/d/0B0fHiozHiJe8Vzg5VjBfLWd4UGs/view?usp=sharing&resourcekey=0-rT9e0vxhy5wQfoIILvM_xg) in their first call-for-proposals on a new genomics format.  Something I now strongly regret taking part in!  (Ultimately they only took the name tokeniser, but have now regretted going with their own entropy encoder as it's simply way too slow.)

The other aspect was improving ratios via custom codecs.  The old rANS wins here because it's so much faster as an entropy encoder, and sometimes entropy encoding (or minimal statistical encoding)  is the best you can get, so the complex LZ searching methods aren't worth it.  (Before CRAM 3.0 we even had zlib being tried with Z_HUFFMAN_ONLY as an option, because it sometimes beat normal zlib while being vastly faster.)  The ideas from there eventually ended up in CRAM 3.1.

One aspect of that was the notion of data transforms.  This is an idea taken out of PNG, and later used by myself for [ZTR](https://academic.oup.com/bioinformatics/article/18/1/3/243465).  Data starts off as raw, but can have a transformation applied to it.   Decoding is like peeling the layers off an onion.  If it's not raw format yet, you pass it through the transform indicated in the data-stream and repeat, until you get to the original raw data.  These could be delta if it's a series of integer values; size squashing - eg an array of 16bit values becoming 8-bit with an escape mechanism for out of bound data; run-length encoding; bit-packing (eg ACGT strings taking up 2bits per value instead of 8); or even bigger transformations such as BWT and MTF.    These can be implemented in CRAM as layered encodings, much like we already have BYTE_ARRAY_LEN which contains two sub-encodings for the literal and length fields.  RLE would be a direct comparison to that in data layout.  I initially implementing PACK and RLE here.

CRAM 4 was an experimental melting pot of compression improvements along side the format changes.

However it rapidly became apparent that getting any traction on these notions was going to be slow in htsjdk, so I took the decision to attempt to produce a cut down CRAM 4, renamed to CRAM 3.1.  The theory was if we only had compression codec changes and we had a high performance separate library to implement them then the code changes needed in other CRAM implementations were very minimal; just updating the list of legal codec numbers and attaching them to callouts to a library API.  This isn't much different to how existing implementations may already be handling lzma, bzip2 and deflate formats.  (Although I'm aware Java has some native implementations, I was incorrectly assuming the faster C implementations were being utilised mostly.)

A consequence of this is the separate onion-layer transform approach had to go, as that was an CRAM encoding change rather than a trivial codec API (ie compress-buffer and decompress-buffer functions).  They got consumed into the rANS-Nx16 codec instead, although minus delta which was a mistake as it would greatly help certain tag types such as ONT signals).

It wasn't such a bad idea, but ultimately it didn't help at all as there was no willingness to simply link against an external library yet we suffered splitting CRAM updates into two new versions instead of 1.

There are still improvements that could be made with CRAM, but I wouldn't make them now I think so ultimately I think CRAM 4.0 is dead in the water.  I'm considering removing it from htslib, if only to reduce the potential attack vector for supplying malicious files.

If I was to pick this up again, it'd probably be to start again from scratch and build on something like the recent [OpenZL](https://github.com/facebook/openzl).  It wouldn't have the same compression ratios as archive mode CRAM, unless we extended it, but I like the notion of a standard structure that off the shelf tooling can decode.  It's like BGZF is basically just decodable with zcat, but with openzl it could include all those useful chunking strategies and data transformations.   I still have plans for a [BGZF2](https://github.com/jkbonfield/htslib/blob/bgzf2/BGZF2.md), but based on OpenZL now instead of zstd.  I just got sucked into other projects and my time evaporated to work on this, but one day!

My rough prototype [showed enormous gains](https://docs.google.com/presentation/d/1Ifksky9LHMfVtSv5VpmXpJ8V3En3frMhgf-lehuq1Uw/edit?usp=sharing) for multi-sample VCF and long-read SAM files.  It's not up to CRAM performance, but closer to CRAM than BAM for Revio and with a rewrite to use OpenZL it ought to be far closer still.

tfenne:
@jkbonfield Thank you so much for both your pointers on testdata in htsspecs, and the history around CRAM 4.  That's a lot to digest, and I think probably I'll reach out via email re: future formats (or finally make it to a GA4GH call).  But the short version is that between AI-assisted coding and Project Panama in Java 21/22+, I think it might be much easier to spin up Java wrappers around native libraries and build Java-enabled implementations of future compression formats much faster.

tfenne:
@yfarjoun and @nh13 I've marked this as ready for review.  I've been back over the code with multiple AIs doing code review, and have fixed up everything they found that I agreed with.

src/main/java/htsjdk/samtools/util/GzipCodec.java
```diff
@@ -0,0 +1,356 @@
...
+     * Decompress GZIP or BGZF data and return a new ByteBuffer containing the result.
+     * Reads the ISIZE field from the GZIP trailer to determine the output size.
+     *
+     * @param input compressed data (from position to limit; position is advanced)
+     * @return a new ByteBuffer containing the decompressed data, positioned at 0 with limit at the end
+     */
+    public ByteBuffer decompress(final ByteBuffer input) {
+        // Read ISIZE from the last 4 bytes of the GZIP block to size the output
+        final int isizeOffset = input.limit() - 4;
+        final int isize = input.duplicate().order(ByteOrder.LITTLE_ENDIAN).position(isizeOffset).getInt();
```
nh13, line 371: **issue:** 
ISIZE is mod 2^32 so you use `Integer.toUnsignedLong` for the 2GB-4GB case, check if isize is zero, but no idea how to handle if the actual size was > 4GB.  I think it doesn't matter for CRAM, but this is a public function...

src/main/java/htsjdk/samtools/cram/io/CRAMByteWriter.java
```diff
@@ -0,0 +1,104 @@
...
+        return pos;
+    }
+
+    /** Reset the writer to empty, reusing the existing buffer. */
+    public void reset() {
+        pos = 0;
+    }
+
+    private void grow(final int minCapacity) {
+        int newCapacity = buf.length << 1;
```
nh13: **nitpick:** could overflow if buf.length > 1GB

CHANGELOG.md
```diff
@@ -10,6 +10,47 @@ early infrastructure for a plugin-based codec framework and resource bundles.
...
+## 5.0.0
+
+Adds **CRAM 3.1 write support** to htsjdk.  This is the culmination of the read-side codec work
+in 4.2.0 and the reader wiring in 4.3.0: htsjdk can now produce CRAM 3.1 files that are
+interoperable with samtools/htslib.
+
+### CRAM 3.1 Write Support
+
+- Enable CRAM 3.1 writing with all spec codecs: rANS Nx16, adaptive arithmetic Range coder, FQZComp, Name Tokenisation, and STRIPE
+- Add configurable compression profiles (FAST, NORMAL, BEST, ARCHIVE) with trial compression for automatic codec selection
```
nh13: **suggestion:** 
```suggestion
- Add configurable compression profiles (FAST, NORMAL, SMALL, ARCHIVE) with trial compression for automatic codec selection
```

README.md
```diff
@@ -11,6 +11,8 @@ manipulating HTS data.
 
 > **NOTE: _HTSJDK has only partial support for the latest Variant Call Format Specification.  VCFv4.3 can be read but not written, VCFv4.4 can be read in lenient mode only, and there is no support for BCFv2.2._**
 
+> **NOTE: _HTSJDK now supports both reading and writing CRAM 3.1 files.  CRAM 3.1 write support includes all codecs defined in the specification (rANS Nx16, adaptive arithmetic Range coder, FQZComp, Name Tokenisation, and STRIPE), configurable compression profiles (FAST, NORMAL, BEST, ARCHIVE), and trial compression for automatic codec selection.  Files produced by htsjdk are interoperable with samtools/htslib._**
```
nh13: **suggestion:** 
```suggestion
> **NOTE: _HTSJDK now supports both reading and writing CRAM 3.1 files.  CRAM 3.1 write support includes all codecs defined in the specification (rANS Nx16, adaptive arithmetic Range coder, FQZComp, Name Tokenisation, and STRIPE), configurable compression profiles (FAST, NORMAL, SMALL, ARCHIVE), and trial compression for automatic codec selection.  Files produced by htsjdk are interoperable with samtools/htslib._**
```

src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
```diff
@@ -107,7 +109,7 @@ public void writeAlignment(final SAMRecord alignment) {
      */
     // TODO: retained for backward compatibility for disq in order to run GATK tests (remove before merging this branch)
```
nh13: **suggestion:** can we remove this now?

src/main/java/htsjdk/samtools/cram/compression/range/RangeEncode.java
nh13: **suggestion:** 
In practice this path is only hit when `rangeParams.isExternalCompression()` is true (bzip2 fallback), but found anyways during review
```suggestion
            inputBuffer.get(rawBytes, 0, inputBuffer.remaining()); 
```

src/main/java/htsjdk/samtools/cram/compression/nametokenisation/NameTokenisationEncode.java
nh13, line 76: **suggestion:** 
found when reviewing:
> If the first block had read names with 8 tokens and the second block only has 5, the encoder still iterates over 8 positions. distributeTokensForPosition for positions 5-7 would get empty/missing token lists, which could produce empty streams or throw, depending on how it handles missing positions.
```suggestion
        maxPositions = 0;
        maxStringValueLength = 0;
```

nh13: This makes me think we need a code review of existing CRAM code...

src/main/java/htsjdk/samtools/cram/encoding/external/ByteArrayStopCodec.java
```diff
@@ -1,42 +1,43 @@
...
-        while ((b = inputStream.read()) != -1 && b != stop) {
-            readingBAOS.write(b);
+        // Scan directly in the underlying byte[] for the stop byte instead of
+        // reading one byte at a time into a ByteArrayOutputStream.
+        final byte[] buf = inputReader.getBuffer();
+        final int startPos = inputReader.getPosition();
+        int scanPos = startPos;
+        while (scanPos < buf.length && (buf[scanPos] & 0xFF) != stop) {
+            scanPos++;
         }
```
nh13, line 37: **suggestion:** 
check that we actually found the stop byte

```suggestion
        }
        if (scanPos == buf.length) {                                                                                                                            
            throw new CRAMException(                
               "Stop byte 0x" + Integer.toHexString(stop) +
               " not found in external block (scanned " + (scanPos - startPos) + " bytes)");                                                               
     }  
```

src/main/java/htsjdk/samtools/cram/structure/CRAMCompressionProfile.java
```diff
@@ -0,0 +1,315 @@
...
+     */
+    SMALL(CramVersions.CRAM_v3_1, 6, 25_000),
+
+    /**
+     * Maximum compression profile. Uses rANS Nx16 for entropy-rich data, FQZComp for quality scores,
+     * and Name Tokeniser for read names at higher compression settings. Writes CRAM 3.1.
+     *
+     * <p>This profile uses trial compression: multiple codecs are tried per block and the smallest
+     * result wins. Additional candidates include BZIP2, the Range (arithmetic) coder, and GZIP.
+     */
```
nh13, line 54: Add a comment like
```
// Note: large slices increase tag dictionary size; CompressionHeader.internalWrite() uses a fixed 100KB buffer
```

src/main/java/htsjdk/samtools/cram/compression/TrialCompressor.java
```diff
@@ -0,0 +1,162 @@
+package htsjdk.samtools.cram.compression;
+
+import htsjdk.samtools.cram.structure.CRAMCodecModelContext;
+import htsjdk.samtools.cram.structure.block.BlockCompressionMethod;
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
```
nh13, line 7: **suggestion:** 
add `<p>Not thread-safe. Instances must not be shared across threads` for `TrialCompressor` and `ExternalCompressor` sub-classes?

src/main/java/htsjdk/samtools/cram/compression/fqzcomp/FQZParam.java
nh13: **nitpick:** 
probably can remove this while we're here

src/main/java/htsjdk/samtools/cram/compression/nametokenisation/NameTokenisationEncode.java
```diff
@@ -19,16 +19,47 @@
...
+            /* TOKEN_DUP     (0x05) */ {RangeParams.PACK_FLAG_MASK | RangeParams.RLE_FLAG_MASK | RangeParams.STRIPE_FLAG_MASK},
+            /* TOKEN_DIFF    (0x06) */ {RangeParams.PACK_FLAG_MASK | RangeParams.STRIPE_FLAG_MASK},
+            /* TOKEN_DIGITS  (0x07) */ {RangeParams.PACK_FLAG_MASK | RangeParams.RLE_FLAG_MASK | RangeParams.STRIPE_FLAG_MASK},
+            /* TOKEN_DELTA   (0x08) */ {0},
+            /* TOKEN_DELTA0  (0x09) */ {RangeParams.PACK_FLAG_MASK},
+            /* TOKEN_MATCH   (0x0A) */ {0},
+            /* TOKEN_NOP     (0x0B) */ {0},
+            /* TOKEN_END     (0x0C) */ {0},
+    };
     private final static String READ_NAME_TOK_REGEX = "([a-zA-Z0-9]{1,9})|([^a-zA-Z0-9]+)";
```
nh13: **nitpick:** these are now dead code

src/main/java/htsjdk/samtools/cram/structure/CRAMEncodingStrategy.java
```diff
@@ -25,57 +25,99 @@
...
     public static final int DEFAULT_MINIMUM_SINGLE_REFERENCE_SLICE_THRESHOLD = 1000;
-
-    // This number must be >= DEFAULT_MINIMUM_SINGLE_REFERENCE_SLICE_THRESHOLD (required by ContainerFactory).
     public static final int DEFAULT_READS_PER_SLICE = 10000;
 
-    // encoding strategies
-    private CompressionHeaderEncodingMap customCompressionHeaderEncodingMap;
-
-    //Note: should this have separate values for tags (separate from CRAMRecord data) ?
+    private CRAMVersion cramVersion = CramVersions.CRAM_v3_1;
```
nh13, line 57: **nitpick:** add to the changelog that all new CRAMs will be 3.1

src/main/java/htsjdk/samtools/SAMFileWriterFactory.java
```diff
@@ -27,6 +27,7 @@
 import htsjdk.samtools.cram.ref.ReferenceSource;
 import htsjdk.samtools.cram.structure.CRAMEncodingStrategy;
 import htsjdk.samtools.util.BlockCompressedOutputStream;
+import htsjdk.samtools.cram.structure.CRAMEncodingStrategy;
```
nh13: **nitpick:** 
importred above
```suggestion
```

src/main/java/htsjdk/samtools/cram/compression/ExternalCompressor.java
```diff
@@ -51,12 +87,12 @@ public boolean equals(Object o) {
 
         ExternalCompressor that = (ExternalCompressor) o;
 
-        return getMethod() == that.getMethod();
+        return method == that.method;
```
nh13, line 90: **question:** should we also compare compression level too?

src/test/java/htsjdk/samtools/cram/cram31/CRAM31FidelityTestBase.java
```diff
@@ -0,0 +1,142 @@
...
+
+        // compare original 3.0 vs samtools 3.1
+        compareCRAMFiles(testInput.toPath(), cramSamtools31.toPath(), testReference.toPath());
+
+        // HTSJDK: write with this profile
+        final IOPath cramHtsjdk31 = IOUtils.createTempPath("cram31Htsjdk", ".cram");
+        writeWithHTSJDK(testInput, cramHtsjdk31, testReference, profile);
+
+        // compare original 3.0 vs HTSJDK output
+        // compare original 3.0 vs HTSJDK output
```
nh13, line 77: **nitpick:** 
```suggestion
```

src/main/java/htsjdk/samtools/cram/compression/fqzcomp/FQZCompEncode.java
```diff
@@ -1,18 +1,528 @@
...
+        // Encoding loop
+        int last = 0;
+        int qctx = 0;
+        int delta = 0;
+        int prevQ = 0;
+        int recIdx = 0;
+        int basesRemaining = 0;
+        boolean firstLen = true;
+        int lastLen = 0;
+        int qualOffset = 0;
```
nh13: **nitpick:** can remove

src/main/java/htsjdk/samtools/cram/compression/nametokenisation/NameTokenisationEncode.java
nh13: **nitpick:** 
```suggestion
        tokenStreamBuffer.put(val.getBytes(StandardCharsets.UTF_8));
```

src/main/java/htsjdk/samtools/cram/structure/Slice.java
```diff
@@ -201,15 +200,16 @@ public Slice(
         this.compressionHeader = compressionHeader;
         this.byteOffsetOfContainer = containerByteOffset;
 
-        final ContentDigests hasher = ContentDigests.create(ContentDigests.ALL);
+        // htslib does not write content digest tags (BD/SD/B5/S5/B1/S1) into slice headers.
```
nh13, line 203: **suggestion:** add to breaking changes in the CHANGELOG?

src/main/java/htsjdk/samtools/cram/CramConverter.java
```diff
@@ -0,0 +1,195 @@
+package htsjdk.samtools.cram;
+
+import htsjdk.samtools.*;
+import htsjdk.samtools.cram.structure.CRAMCompressionProfile;
+import htsjdk.samtools.cram.structure.CRAMEncodingStrategy;
+import htsjdk.samtools.util.CloserUtil;
```
nh13: **suggestion:** 
```suggestion
```

src/main/java/htsjdk/samtools/cram/io/LTF8.java
```diff
@@ -104,6 +104,47 @@ public static long readUnsignedLTF8(final InputStream inputStream) {
...
     }
 
+    /**
+     * Reads an unsigned LTF8 long from a {@link CRAMByteReader}. Equivalent to the InputStream version
+     * but avoids synchronized method call overhead.
+     *
+     * @param reader the reader to read from
+     * @return the decoded value
+     */
+    public static long readUnsignedLTF8(final CRAMByteReader reader) {
```
nh13, line 114: **suggestion:** these are super dense, ITF8.java and the rest of this code I think is more readable.

nh13 (approved):
This needs some battle-testing; it's too big for a static code review :/

tfenne:
Tests pass locally, and are still running into OOM issues in CI.  I'm going to merge this to dev and address there if it's still an issue once some other PRs land.

tfenne:
I should note for the record that this PR adds a metric ton of unit tests for the CRAM code, and has been battle-tested by transcoding a broad variety of real-world data from BAM to CRAM and back again with each compression profile, with samtools and htsjdk via the included validation pipeline.  The validation pipeline uncovered a small number of subtle bugs that have now been fixed!

PR #1764, opened 2026-03-31
A few targeted optimizations to the BAM decoding path that yield 6-7% improvement in performance

tfenne:
### Description

Fixing some low-hanging fruit that gives about a 6-7% reduction in the time taken to read BAM files (in both lazy and eager modes).

The first commit is pulled over from #1763 and optimized decoding of the 4-bit bases in BAM into full bytes.  If/when merged I'll prune it from the other PR.

The second commit eliminates a silly chain of calls where resolving the refId and mateRefId ended up running in circles resovling int->String->int 2-3 times each - this was probably the biggest win(!)

The final commit makes parsing the aux tags slightly more efficient; it's impact will likely scale with the number of String valued tags.

### DIFF
--- a/src/main/java/htsjdk/samtools/BAMRecordCodec.java
+++ b/src/main/java/htsjdk/samtools/BAMRecordCodec.java
@@ -284,10 +284,6 @@ public class BAMRecordCodec implements SortingCollection.Codec<SAMRecord> {
                 header, referenceID, coordinate, readNameLength, mappingQuality,
                 bin, cigarLen, flags, readLen, mateReferenceID, mateCoordinate, insertSize, restOfRecord);
 
-        if (null != header) {
-            // don't reset a null header as this will clobber the reference and mate reference indices
-            ret.setHeader(header);
-        }
         return ret;
     }
 }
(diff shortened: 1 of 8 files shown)


src/main/java/htsjdk/samtools/BAMRecordCodec.java
```diff
@@ -284,10 +284,6 @@ public SAMRecord decode() {
                 header, referenceID, coordinate, readNameLength, mappingQuality,
                 bin, cigarLen, flags, readLen, mateReferenceID, mateCoordinate, insertSize, restOfRecord);
 
-        if (null != header) {
```
tfenne, line 287: The header is passed to the BAMRecord constructor through the factory call 4 lines above - this does nothing other than cause a whole second round of refId/mateRefId resolution.

src/main/java/htsjdk/samtools/BinaryTagCodec.java
```diff
@@ -407,19 +407,12 @@ private static TagValueAndUnsignedArrayFlag readArray(final ByteBuffer byteBuffe
...
-        byteBuffer.reset();
-        byteBuffer.get(buf);
-        // Skip over the null terminator
-        byteBuffer.get();
-        return StringUtil.bytesToString(buf);
+        // Scan the backing array directly to avoid the double-pass of mark/reset/re-read
+        final byte[] array = byteBuffer.array();
+        final int start = byteBuffer.arrayOffset() + byteBuffer.position();
+        int end = start;
+        while (array[end] != 0) { end++; }
```
nh13: **suggestion:** if the array doesn't have a null terminator, should we throw a SAMFormatException (truncated file)?

nh13 (approved):
one nice to have, lgtm

PR #1765, opened 2026-03-31
Fix LTF8 9-byte write bug: wrong bit shift (>> 28 instead of >> 24)

tfenne:
The writeUnsignedLTF8(long, OutputStream) method incorrectly shifts by 28 bits instead of 24 bits when encoding byte 5 of a 9-byte LTF8 value. This causes data corruption for values requiring the full 9-byte encoding (values >= 2^56 with certain bit patterns).

The 9-byte encoding should write bytes at shifts:
  56, 48, 40, 32, 24, 16, 8, 0
but the code wrote:
  56, 48, 40, 32, 28, 16, 8, 0

The corresponding read path (readUnsignedLTF8) correctly uses << 24, so the reader and writer disagreed for affected values.

The correct behavior is confirmed by htslib's reference implementation:
  https://github.com/samtools/htslib/blob/master/cram/cram_io.c#L364-L373

Existing tests did not catch this because their 9-byte test values (-4757, Long.MAX_VALUE, -1) have identical nibbles at bit positions 28 and 24, masking the bug. A targeted test with value 0x0123456789ABCDEF is added to exercise the difference.

### Description

- [x] Check your code style.

### DIFF
--- a/src/test/java/htsjdk/samtools/cram/io/LTF8Test.java
+++ b/src/test/java/htsjdk/samtools/cram/io/LTF8Test.java
@@ -72,32 +72,49 @@ public class LTF8Test extends HtsjdkTest {
     }
 
     /**
-     * Tests the 9-byte (full 64-bit) LTF8 encoding with a value where bits [35:32] differ from
+     * Tests the 9-byte (full 64-bit) LTF8 encoding with values where bits [35:32] differ from
      * bits [27:24], which exposes the >> 28 vs >> 24 bug in writeUnsignedLTF8.
-     *
-     * For 0x0123456789ABCDEFL:
-     *   (value >> 28) & 0xFF == 0x78  (wrong, old behavior)
-     *   (value >> 24) & 0xFF == 0x89  (correct)
      *
      * The 9-byte encoding writes a 0xFF prefix followed by bytes at shifts 56,48,40,32,24,16,8,0.
      */
-    @Test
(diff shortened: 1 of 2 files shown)


src/test/java/htsjdk/samtools/cram/io/LTF8Test.java
```diff
@@ -71,6 +71,47 @@ public void testPredefined (long value, byte[] ltf8) throws IOException {
...
+     * The 9-byte encoding writes a 0xFF prefix followed by bytes at shifts 56,48,40,32,24,16,8,0.
+     */
+    @Test
+    public void testNineByteEncodingWithDistinctBitsAt24And28() throws IOException {
+        final long value = 0x0123456789ABCDEFL;
+
+        // Expected encoding: 0xFF prefix + 8 bytes at shifts 56,48,40,32,24,16,8,0
+        final byte[] expectedBytes = {
+            (byte) 0xFF,
+            (byte) 0x01, // >> 56
```
yfarjoun: let's encode something that _doesn't_ start with zero since the spec is a little ambiguous about the presence or lackthereof of a trailing zero after the prefixed 1's. (you can add the same test but backwards:
`0xFEDCBA987654321L`)

yfarjoun (changes requested):
asking for a test without a leading zero

PR #1768, opened 2026-04-02
Integrate jlibdeflate for faster DEFLATE compression and decompression

tfenne:
Add optional jlibdeflate (libdeflate JNI bindings) support that is automatically used when the native library is available on the classpath, with transparent fallback to the JDK Inflater/Deflater.

It should be noted that `jlibdeflate` (https://github.com/fulcrumgenomics/jlibdeflate) is brand new, and is developed specifically to be used here.  On pure BAM roundtrip on mac/aarch64 it is more than 2x faster than the built in jdk/zlib implementation!!  Once we reach consensus on this PR I'll create a first release of jlibdeflate and publish to sonatype, and amend this PR.

- DeflaterFactory/InflaterFactory auto-detect jlibdeflate availability at first use, caching the result for the lifetime of the JVM
- LibdeflateDeflater/LibdeflateInflater extend JDK Deflater/Inflater and correctly handle both raw DEFLATE (nowrap=true) and zlib format (nowrap=false)
- New Defaults.USE_LIBDEFLATE (samjdk.use_libdeflate) defaults to true; set to false to force JDK-only compression
- jlibdeflate 0.1.0-SNAPSHOT added as a dependency from Sonatype Central snapshots

- [x] Check your code style.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,7 +34,7 @@ jacocoTestReport {
 }
 
 dependencies {
-    implementation 'com.fulcrumgenomics:jlibdeflate:0.1.0-SNAPSHOT'
+    implementation 'com.fulcrumgenomics:jlibdeflate:0.1.0'
     implementation 'commons-logging:commons-logging:1.3.0'
     implementation "org.xerial.snappy:snappy-java:1.1.10.5"
     implementation 'org.apache.commons:commons-compress:1.26.0'
@@ -97,7 +97,7 @@ tasks.withType(Test).configureEach { task ->
     task.maxHeapSize = "6G"
 
     task.jvmArgs '-Djava.awt.headless=true'  //this prevents awt from displaying a java icon while the tests are running
-C
+
     int count = 0
     // listen to events in the test execution lifecycle
 
--- a/build.gradle
+++ b/build.gradle
@@ -16,6 +16,11 @@ plugins {
 
 repositories {
     mavenCentral()
+    maven {
+        name = 'sonatypeSnapshots'
+        url = 'https://central.sonatype.com/repository/maven-snapshots/'
+        mavenContent { snapshotsOnly() }
+    }
 }
 
 jacocoTestReport {
@@ -29,6 +34,7 @@ jacocoTestReport {
 }
 
 dependencies {
+    implementation 'com.fulcrumgenomics:jlibdeflate:0.1.0-SNAPSHOT'
     implementation 'commons-logging:commons-logging:1.3.0'
     implementation "org.xerial.snappy:snappy-java:1.1.10.5"
     implementation 'org.apache.commons:commons-compress:1.26.0'
@@ -91,7 +97,7 @@ tasks.withType(Test).configureEach { task ->
     task.maxHeapSize = "6G"
 
     task.jvmArgs '-Djava.awt.headless=true'  //this prevents awt from displaying a java icon while the tests are running
-
+C
     int count = 0
     // listen to events in the test execution lifecycle
 
--- a/src/main/java/htsjdk/samtools/util/zip/DeflaterFactory.java
+++ b/src/main/java/htsjdk/samtools/util/zip/DeflaterFactory.java
@@ -23,14 +23,25 @@
  */
 package htsjdk.samtools.util.zip;
 
+import htsjdk.samtools.Defaults;
 import htsjdk.samtools.util.BlockCompressedOutputStream;
+import htsjdk.samtools.util.Log;
+
 import java.util.zip.Deflater;
 
 /**
  * Factory for {@link Deflater} objects used by {@link BlockCompressedOutputStream}.
  * This class may be extended to provide alternative deflaters (e.g., for improved performance).
+ *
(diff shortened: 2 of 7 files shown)


build.gradle
```diff
@@ -91,7 +97,7 @@ tasks.withType(Test).configureEach { task ->
     task.maxHeapSize = "6G"
 
     task.jvmArgs '-Djava.awt.headless=true'  //this prevents awt from displaying a java icon while the tests are running
-
+C
```
nh13: **issue:** 
```suggestion
```

src/main/java/htsjdk/samtools/util/zip/LibdeflateDeflater.java
```diff
@@ -0,0 +1,116 @@
...
+            return 0;
+        }
+
+        final int compressed = nowrap
+                ? compressor.deflateCompress(inputBuf, inputOff, inputLen, output, off, len)
+                : compressor.zlibCompress(inputBuf, inputOff, inputLen, output, off, len);
+        if (compressed == -1) {
+            // Output buffer too small — caller will handle this (e.g. fall back to no-compression)
+            done = false;
+            return len; // fill the buffer to signal it didn't fit; finished() returns false
```
nh13: **question:** 
shouldn't this return 0, since `len` bytes were not written?

src/main/java/htsjdk/samtools/util/zip/LibdeflateDeflater.java
```diff
@@ -0,0 +1,116 @@
...
+        return done;
+    }
+
+    @Override
+    public void reset() {
+        inputBuf = null;
+        inputOff = 0;
+        inputLen = 0;
+        finishing = false;
+        done = false;
```
nh13: **nitpick:** 
```suggestion
        inputBuf  = null;
        inputOff  = 0;
        inputLen  = 0;
        finishing = false;
        done      = false;
```

build.gradle
```diff
@@ -29,6 +34,7 @@ jacocoTestReport {
 }
 
 dependencies {
+    implementation 'com.fulcrumgenomics:jlibdeflate:0.1.0-SNAPSHOT'
```
nh13: **todo:** 
update before merge

src/main/java/htsjdk/samtools/util/zip/LibdeflateDeflater.java
```diff
@@ -0,0 +1,116 @@
...
+    public void setInput(final byte[] input, final int off, final int len) {
+        this.inputBuf = input;
+        this.inputOff = off;
+        this.inputLen = len;
+        this.done = false;
+    }
+
+    @Override
+    public void finish() {
+        this.finishing = true;
```
nh13, line 75: **question:** where is `this.finishing` ever read?

src/test/java/htsjdk/samtools/util/zip/LibdeflateTest.java
```diff
@@ -0,0 +1,121 @@
+package htsjdk.samtools.util.zip;
```
nh13, line 24: **nitpick:** all other files have licenses in their header, should we follow that same convention

src/main/java/htsjdk/samtools/util/zip/DeflaterFactory.java
```diff
@@ -43,6 +54,39 @@ public DeflaterFactory() {
...
      */
     public Deflater makeDeflater(final int compressionLevel, final boolean gzipCompatible) {
+        if (Defaults.USE_LIBDEFLATE && isLibdeflateAvailable()) {
+            return new LibdeflateDeflater(compressionLevel, gzipCompatible);
+        }
         return new Deflater(compressionLevel, gzipCompatible);
     }
+
+    /** Returns true if the libdeflate native library can be loaded. */
+    static boolean isLibdeflateAvailable() {
```
nh13, line 64: **nitpick:** it seems odd that `InflaterFactory` calls a method on `DeflatorFactory`, but putting it into it's own separate class seems odd too.  🤷

nh13 (changes requested):
Looking good, just a few questions and one extra `C`

src/main/java/htsjdk/samtools/util/zip/DeflaterFactory.java
tfenne: Agreed.  Feels like it should just be `CompressionFactory`, but consolidating the two would be a bigger breaking change now.

src/main/java/htsjdk/samtools/util/zip/LibdeflateDeflater.java
tfenne: Yes, you're right. It works because BlockCompressedOutputStream does per-block compression and checks `done` but not the return value of this function.  Fixed to do the right thing here.

src/main/java/htsjdk/samtools/util/zip/LibdeflateDeflater.java
tfenne: good catch.  It should be checked in `done()`.

PR #1769, opened 2026-04-02
Update maven central publishing to the new central portal and ...

tfenne:
…document release publication steps in CONTRIBUTING.md


build.gradle
```diff
@@ -228,25 +229,34 @@ publishing {
...
     }
 
-    repositories {
-        maven {
-            credentials {
-                username = isRelease ? project.findProperty("sonatypeUsername") : System.env.ARTIFACTORY_USERNAME
-                password = isRelease ? project.findProperty("sonatypePassword") : System.env.ARTIFACTORY_PASSWORD
-            }
-            def release = "https://oss.sonatype.org/service/local/staging/deploy/maven2/"
-            def snapshot = "https://broadinstitute.jfrog.io/broadinstitute/libs-snapshot-local/"
```
nh13, line 252: 1. will folks using `broadinstitute.jfrog.io` for snapshots no longer get updates?  I this is ok, but good to consider the implications
2. I recall having to "turn on" snapshots in the central portal, did you do that?

build.gradle
```diff
@@ -228,25 +229,34 @@ publishing {
...
-        }
-    }
 }
 
 /**
  * Sign non-snapshot releases with our secret key.  This should never need to be invoked directly.
+ * Uses the system gpg command so that gpg-agent can prompt for the passphrase.
  */
 signing {
-    required { isRelease && gradle.taskGraph.hasTask("publishHtsjdkPublicationToMavenRepository") }
```
nh13, line 262: **question:** this make signing always required whenever `-Drelease=true` is used, for any task. Worth considering if you want to revert it to just when publishing a release.

build.gradle
```diff
@@ -228,25 +229,34 @@ publishing {
...
     sign publishing.publications.htsjdk
+    required { isRelease }
+    tasks.withType(Sign).configureEach { onlyIf { isRelease } }
+}
+
+/**
+ * Publish to Maven Central via Sonatype Central Portal.
+ * Credentials: generate user tokens at https://central.sonatype.com (Account > User Token)
+ * and store as sonatypeUsername/sonatypePassword in ~/.gradle/gradle.properties
+ * or as ORG_GRADLE_PROJECT_sonatypeUsername/ORG_GRADLE_PROJECT_sonatypePassword env vars.
```
nh13, line 270: **question:** I assume this is still manual and not via CI?  Both for snapshots and releases.  Can you confirm?

CONTRIBUTING.md
```diff
@@ -0,0 +1,221 @@
...
+
+```bash
+./gradlew jar
+```
+
+To run tests:
+
+```bash
+./gradlew test
+```
```
nh13, line 15: **suggestion (if-minor):** 

Could also mention `./gradlew install`

build.gradle
nh13: **nitpick:** 

Had some help finding this ;)
```suggestion
final gitVersion = gitVersion().replaceAll("\\.dirty", "")
```

nh13 (commented):
Can we update to the latest gradle while we're here (eg. Gradle 9.4.1 which is two years newer)?

build.gradle
tfenne: 1. Yes - though I believe the Broad's jfrog has been down for some time, and even if it weren't, I don't think anyone has credentials for it or a path to getting them.  
2. Thanks for the reminder, yes.

build.gradle
tfenne: Correct - for now at least this just fixes up the process and makes it possible manually.  We should probably configure automatic snapshots soon though.

build.gradle
tfenne: Fixed by scoping back to whether the new publish task is in the graph.

PR #1771, opened 2026-04-17
Fix asymmetric SamPairUtil.getPairOrientation on dovetail pairs

nh13:
## Summary
- `SamPairUtil.getPairOrientation` returned different orientations for the two ends of the same pair when the reads' 5' positions coincided (e.g., soft-clipped dovetail pairs).
- Root cause: the reverse-strand branch uses `getAlignmentEnd()` (CIGAR-derived, exact); the forward-strand branch approximates via `getAlignmentStart() + getInferredInsertSize()`. At the equality boundary the two disagree by one base, flipping `FR` vs `RF`.
- Fix: prefer `SAMUtils.getMateAlignmentEnd(record)` (from the `MC` tag) in the forward-strand branch; fall back to the TLEN-based formula with a documented caveat when `MC` is absent.
- Added targeted tests covering both a simple dovetail 5' tie and a realistic soft-clipped dovetail.

## The bug — where it lives

```
Normal FR innie (fine — no bug):
  pos:  1                    500
        F: 5'===>              <===5' :R
            ↑                      ↑
          F 5' = 1              R 5' = 500
        1 < 500  →  FR  (both branches agree, plenty of margin)


Dovetail 5' tie (the bug):
  pos:  1                   100                   199
        R: <=================5'                              (1..100, 100M reverse)
                             F: 5'=================>         (100..199, 100M forward)
                             ↑
                   F 5' = R 5' = 100   ← they coincide
```

**What each branch computes at the tie:**

```
Evaluated on R (reverse-strand record):
  pos_forward_5' = record.getMateAlignmentStart()   = 100
  pos_reverse_5' = record.getAlignmentEnd()         = 100   ← CIGAR, exact
  100 < 100 ? no  →  RF

Evaluated on F (forward-strand record):
  pos_forward_5' = record.getAlignmentStart()                      = 100
  pos_reverse_5' = record.getAlignmentStart() + TLEN = 100 + 1     = 101   ← TLEN, off-by-one
  100 < 101 ? yes →  FR   ← disagrees with R!
```

**Why TLEN lies here.** `computeInsertSize` signs TLEN as `(R5' − F5') + 1` when `R5' ≥ F5'`. With 5' ends equal, `TLEN = +1` — so `F.alignStart + TLEN` lands *one base past* where R actually is:

```
pos:       100   101
            │     │
            │     └─ where the forward branch thinks R's 5' lives  (alignStart + TLEN)
            └─────── where R's 5' actually is                       (R.alignEnd)
```

That one-base phantom shift is what flips the `<` comparison and produces `FR` on F but `RF` on R for the same pair.

## The fix

Use the mate CIGAR (`MC`) tag to compute the mate's alignment end exactly — the same quantity the reverse-strand branch reads from its own CIGAR:

```
F has MC tag = R's CIGAR = "100M"
SAMUtils.getMateAlignmentEnd(F) = R.mateAlignmentStart + refLen(MC) − 1
                                 = 1 + 100 − 1
                                 = 100          ← matches R's own alignEnd, exact

Forward branch now:
  100 < 100 ? no  →  RF    ← agrees with R
```

When `MC` is absent, the old TLEN-based formula is retained as a fallback. The Javadoc documents that symmetry cannot be guaranteed in the fallback path, since TLEN from an arbitrary upstream writer may not encode the mate's 5' position exactly.

## Test plan
- [x] Added `"dovetail 5' tie"` row to `SamPairUtilTest.testGetPairOrientation` — verifies symmetry on a simple 100M dovetail tie.
- [x] Added `testGetPairOrientationSymmetryForSoftClippedDovetail` — verifies symmetry on a 90S10M / 10M90S dovetail.
- [x] All 32 existing `SamPairUtilTest` cases continue to pass.
- [ ] Reviewer gut-check: confirm that **"5' tie → RF"** (matching the reverse-strand branch's existing strict-`<` semantics) is the preferred canonicalization. The Javadoc phrasing *"FR means the read mapped to the forward strand comes before the read mapped to the reverse strand"* is silent on ties.

## Commits
1. `Add failing tests for SamPairUtil.getPairOrientation asymmetry` — tests first, verified to fail for the right reason.
2. `Fix SamPairUtil.getPairOrientation asymmetry using mate CIGAR` — implementation + Javadoc update.

### DIFF
--- a/src/main/java/htsjdk/samtools/SamPairUtil.java
+++ b/src/main/java/htsjdk/samtools/SamPairUtil.java
@@ -110,23 +110,29 @@ public class SamPairUtil {
         if (readIsOnReverseStrand) {
             // read's 5' position  ( <---x ) - CIGAR-derived, exact
             negativeStrandFivePrimePos = record.getAlignmentEnd();
-        } else if (SAMUtils.getMateCigarString(record) != null) {
-            // mate's 5' position  ( <---x ) derived from the mate CIGAR (MC)
-            // tag when available.  This matches what the reverse-strand
-            // branch computes for the same pair and keeps getPairOrientation
-            // symmetric across the two ends.
-            negativeStrandFivePrimePos = SAMUtils.getMateAlignmentEnd(record);
         } else {
-            // Fallback: derive the mate's 5' position from TLEN under
-            // htsjdk's convention (see computeInsertSize), where TLEN
-            // encodes an inclusive 5'-to-5' interval length with a ±1
-            // adjustment.  CoordMath.getEnd converts that length back into
-            // the mate's 5' reference position.  This fallback may still
-            // produce asymmetric results when TLEN was written by a tool
-            // that uses SAM v1 §1.4's leftmost-to-rightmost convention
-            // (e.g., bwa on a dovetail pair); callers that need symmetric
-            // results in that situation must provide the MC tag.
-            negativeStrandFivePrimePos = CoordMath.getEnd(record.getAlignmentStart(), record.getInferredInsertSize());
+            // Prefer the mate CIGAR (MC) tag when available so the mate's
+            // 5' position  ( <---x ) matches what the reverse-strand branch
+            // computes for the same pair and keeps getPairOrientation
+            // symmetric across the two ends.  Decode the mate CIGAR once
+            // rather than looking up the MC attribute twice via
+            // SAMUtils.getMateAlignmentEnd.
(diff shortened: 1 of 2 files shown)


tfenne (approved):

tfenne:
I think this makes good sense - at the very least the result should be symmetrical (where possible).  Fixing the underlying off-by-one on the computation involving TLEN should also fix this in a lot of situations when the MC isn't available.

PR #1773, opened 2026-04-25
Apply Palantir Java Format to the entire codebase + enforce on every build

### LINKED ISSUE #1761: Any objections to integrating a code formatter for htsjdk?
Having spent quite a bit of time in the CRAM section of HTSJDK over the last few days, I think I would really like to apply a code-formatter to HTSJDK.

I think I would lean towards the [Palantir Java Format](https://github.com/palantir/palantir-java-format).  It doesn't look hard to add as a gradle plugin, and then could be enforced in CI.

While applying it the first time would obvoiusly cause a *ton* of changes, it sounds like we could add a `.git-blame-ignore-revs` file that would cause the specific commit that applies the formatting to be ignored when computing git-blame annotations(!)

tfenne:
## Summary

Closes #1761.

This is a doozy on purpose: it brings every `.java` file in the repo under a single mechanical code style ([Palantir Java Format](https://github.com/palantir/palantir-java-format)) and wires up the build so the codebase stays that way going forward.

The PR is structured as **two commits, no squash**, so both flow into `dev`:

1. **`Apply Palantir Java Format to entire codebase`** — 1155 files reformatted. No behavior changes. The single non-mechanical edit is the removal of one stray `// the imports for unit testing.` comment in `VariantContextUnitTest.java` that was preventing the formatter from treating an import block as contiguous.
2. **`Add Spotless + Palantir Java Format and auto-format on every build`** — adds the [Spotless](https://github.com/diffplug/spotless) Gradle plugin configured with `palantirJavaFormat()`; hooks `spotlessJavaApply` into `compileJava` so local builds **auto-rewrite** unformatted source in place; adds a CI `formatCheck` job that runs `spotlessCheck` (verify-only) as the actual enforcement boundary; adds a `Code Style` section to `CONTRIBUTING.md`; adds `.git-blame-ignore-revs` pointing at commit 1; deletes the stale `java-style-eclipse.xml` and `java-style-intellij.xml`.

### Local DX

Local builds *fix* formatting rather than fail on it -- contributors don't need to remember to run a formatter command. CI separately verifies (no mutation), so unformatted code still can't slip into `dev`.

`spotlessJavaApply` is sub-second warm thanks to Gradle's up-to-date checking, so the cost on every compile is negligible.

### Git blame

`.git-blame-ignore-revs` points at the bulk-format commit. GitHub honors it automatically in the web UI. Locally, run once per clone:

```bash
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

This is documented in the new `CONTRIBUTING.md` section.

## Test plan

- [x] `./gradlew compileJava compileTestJava` -- passes
- [x] `./gradlew test` -- 21,936 / 21,936 pass
- [x] `./gradlew spotbugsMain spotbugsTest` -- clean
- [x] Verified `compileJava` rewrites unformatted source in place (smoke-tested on a deliberately-malformed file)
- [x] Verified `git blame` with `--ignore-revs-file .git-blame-ignore-revs` credits the original authors, not the reformat commit
- [x] CI: confirm `formatCheck` job is green
- [x] CI: confirm existing `test` / `testExternalApis` / `spotBugs` jobs still green


tfenne:
Fixes #1761

PR #1782, opened 2026-05-20
Optimize SAM text parsing (~2.2x throughput)

tfenne:
## Summary

Speeds up SAM text reading from **645k → 1.44M records/sec (~2.22x)** on a 2M-record exome benchmark, with consistent gains on a 51M-record gzipped WGS file. Achieved through a series of small, measured changes, each independently benchmarked. All 22,112 tests pass; the branch adds direct unit tests for every new method and every bug surfaced during review.

## Changes (10 commits)

| Commit | Headline change | Cumulative speedup |
|--------|-----------------|-------------------|
| `f5ae59b` | Fused `fastqToPhred` / `readStringToNormalizedBases`, tag-parsing rework | 1.46x |
| `7075b77` | Cache `SeekableFileStream.length` / track position locally; skip redundant validation in SILENT mode; fast int parse for `'i'` tag values | 1.68x |
| `65f4aeb` | New `SamLineReader` -- byte-buffer-based, skips UTF-8 `StreamDecoder` for alignment lines (which are ASCII per spec); falls back to UTF-8 for header lines | 1.86x |
| `130567e` | Resolve RNAME/RNEXT via a direct sequence-dictionary lookup; bypass `SAMRecord.setReferenceName`'s redundant lookup/intern dance | 1.91x |
| `1a42d33` | Drop the redundant `setHeader` call in `parseLine` (constructor already set it; the second call did two HashMap lookups that got immediately overwritten) | 1.98x |
| `84f4f2c` | Parse alignment lines from raw bytes; SWAR-scan tabs at 8 bytes per iteration (matches HotSpot's intrinsified `String.indexOf` without materializing a `String`) | 2.15x |
| `8238881` | Fast-path FLAG parsing; skip the empty-field validation loop in SILENT mode | 2.21x |
| `fa85d04` | Honor configured `InflaterFactory` for BGZF-compressed SAM input (consistency with BAM path) | (consistency) |
| `1ecddd0` | Cleanup: drop dead code, tighten docs/visibility, expand test coverage | (cleanup) |
| `d91bc92` | Address findings from a 5-agent code review pass: 5 real bugs (incl. stale error line, 4-byte-tag crash, `skip()` return value, FLAG silent overflow, peek/buffer aliasing), 4 important issues, and 4 nice-to-haves | (~2.22x) |

## Benchmark methodology

* **Exome SAM**: 673 MB, 1.997M records, BAM-derived. Coordinate-sorted. Used `SamReaderFactory.makeDefault().validationStringency(SILENT).open(file)` and iterated all records, consuming bases/quals/CIGAR/tags to defeat dead-code elimination.
* **WGS gzipped SAM**: 5.1 GB BGZF-compressed (`.sam.gz`), 51.7M records, queryname-sorted. Same harness.
* All benchmarks were run on the same hardware (Apple M1 series, JDK 17), with 2 warmup + 5 measure iterations on the exome file. Each optimization was benchmarked individually before and after, with the rule "stop and revert if a change regresses by more than 1-2%."
* CPU profiles were captured with async-profiler (libasyncProfiler 4.4) and the parsing-path hot spots were inspected after every cluster of changes.

## What's in the diff

**Source (9 files changed)**:
- `SAMLineParser.java` -- main rewrite. New `parseLineFromBytes(byte[], off, len, lineNumber)` that scans tabs into parallel offset/length `int[]`s using a SWAR scanner (8 bytes per iteration via `VarHandle.byteArrayViewVarHandle`), parses the 5 numeric fields directly from bytes, and materializes Strings only for fields that ultimately need to be stored as Strings (QNAME, RNAME/RNEXT, CIGAR, Z-type tag values). The legacy `parseLine(String)` is preserved as a thin wrapper.
- `SamLineReader.java` (new) -- replaces `BufferedLineReader` in `SAMTextReader`. Reads raw bytes from an `InputStream` and exposes them via `readNextLine()` + `getLineBuffer()` / `getLineOffset()` / `getLineLength()` without allocating a `String`. The `readLine()` method is preserved as a wrapper that decodes header lines as UTF-8 (spec-required) and alignment lines as ISO-8859-1 (bulk arraycopy on compact-string JVMs).
- `SAMUtils.java` -- byte-range overloads of `readStringToNormalizedBases` and `fastqToPhred` so SEQ/QUAL skip the `String` round-trip; `fastqToPhred(String)` hardened against non-ASCII inputs.
- `TextTagCodec.java` -- byte-range `decodeValue(byte[], off, len)` and byte-range integer-tag parser; `decode(String)` reuses a shared `MutableTagEntry` (documented).
- `SAMTextReader.java` -- switched over to the byte-based API.
- `SAMRecord.java` -- one-line tidy: `setReadString` now uses the consolidated `readStringToNormalizedBases`.
- `SamReaderFactory.java` -- pass `this.inflaterFactory` to `BlockCompressedInputStream` for BGZF-SAM consistency with the BAM path; updated the `inflaterFactory(...)` doc.
- `SeekableFileStream.java` -- cache `length` and track `pos` locally, eliminating per-call stat/lseek syscalls in the read hot path; `skip()` clamps and reports the real byte count.

**Tests**:
- New `SamLineReaderTest` (~40 tests) covers the byte-based API, line terminators (`\n`, `\r`, `\r\n` incl. across buffer boundaries), header UTF-8 vs alignment Latin-1, EOF / peek / line-number behavior, idempotent close.
- `SAMUtilsTest` adds tests for the new byte[] overloads of `fastqToPhred` and `readStringToNormalizedBases`, boundary phred values, and the non-ASCII regression.
- `TextTagCodecTest` adds tests for the byte-range `decodeValue` including the 4-byte-tag empty-value boundary that previously crashed and the unsigned-32-bit `'i'` value boundary.
- `SeekableFileStreamTest` grows from 1 test (just `seek`) to 9, covering `available`/`eof`/`position`/`skip` (incl. clamping at EOF and rejecting negative skip) and all three `read()` overloads.
- `SAMTextReaderTest` pins the FLAG > 2^31 rejection and the cross-API `currentLine` reset.

## Spec compliance

The new line reader's UTF-8 vs ISO-8859-1 split is grounded in the SAM v1.6 specification, section 1: *"SAM file contents are 7-bit US-ASCII, except for certain field values as individually specified which may contain other Unicode characters encoded in UTF-8."* The only fields that allow UTF-8 are a handful of header description fields (`@SQ DS`, `@RG DS`, `@PG DS`, `@PG CL`, `@CO`); every alignment record field is restricted to printable ASCII. Decoding alignment lines as Latin-1 is therefore lossless, and on modern JVMs with compact strings turns the `String` construction into a bulk `arraycopy`.

## Test plan

- [x] `./gradlew test` -- all 22,112 tests pass
- [x] Exome SAM benchmark (645k → 1.44M rec/s, 2.22x)
- [x] WGS gzipped SAM benchmark (consistent ~11% improvement on the SAM parsing portion after subtracting decompression)
- [x] Targeted regression tests for every bug found during the review pass
- [x] CPU profiles before / after each major change to verify no introduced hotspots

## Compatibility notes

- Public method signatures are preserved. `SAMLineParser.parseLine(String)` is unchanged (now a thin wrapper); `TextTagCodec.decode(String)` is unchanged but its returned `Map.Entry` is now shared/reused -- this is documented on the method now. No external callers in this codebase retain the entry across calls.
- `SamReaderFactory.inflaterFactory(...)` now applies to BGZF-SAM as well as BAM. Previously the BGZF-SAM path silently ignored it, so any caller that explicitly configured a custom factory will now observe it being honored.
- `SeekableFileStream.length()`, `eof()`, `position()`, and `available()` no longer issue syscalls per call; the length is captured at construction. Callers expecting to observe a file growing under them (an unusual SAM/BAM/CRAM use case) should be aware. Documented on the class.

## Summary by CodeRabbit

* **New Features**
  * Added a low-allocation SAM line reader and a byte-oriented parsing entrypoint for faster streaming parsing.

* **Performance**
  * Substantially reduced per-line memory allocations and sped up numeric/base/tag decoding.

* **Bug Fixes**
  * Rejects out-of-range FLAG values, fixes mate-reference handling, and ensures error messages report the current malformed line.

* **Documentation**
  * Clarified inflater factory behavior for compressed SAM/BAM handling.

* **Tests**
  * Expanded regression and unit tests for parsing, tag decoding, normalization, and line-reading.

### DIFF
--- a/src/main/java/htsjdk/samtools/SAMLineParser.java
+++ b/src/main/java/htsjdk/samtools/SAMLineParser.java
@@ -93,7 +93,6 @@ public class SAMLineParser {
     private final TextTagCodec tagCodec = new TextTagCodec();
 
     private int currentLineNumber;
-    private String currentLine;
     private byte[] currentLineBuf;
     private int currentLineOff;
     private int currentLineLen;
@@ -238,8 +237,6 @@ public class SAMLineParser {
         // SAM alignment lines are restricted to printable ASCII by the spec, so ISO-8859-1 is
         // lossless and turns getBytes() into a bulk arraycopy on compact-string JVMs.
         final byte[] bytes = line.getBytes(StandardCharsets.ISO_8859_1);
-        // No need to set this.currentLine here: parseLineFromBytes clears it to null up front and
-        // reconstructs the error-message line lazily from the byte buffer when needed.
         return parseLineFromBytes(bytes, 0, bytes.length, lineNumber);
     }
 
@@ -262,9 +259,6 @@ public class SAMLineParser {
         this.currentLineBuf = buf;
         this.currentLineOff = off;
         this.currentLineLen = len;
-        // Clear any String set by an earlier parseLine(String) call so makeErrorString reads from
-        // the current byte buffer rather than the stale prior line.
-        this.currentLine = null;
 
         final int numFields = splitFieldsByTab(buf, off, len, mFieldOffsets, mFieldLengths);
         if (numFields < NUM_REQUIRED_FIELDS) {
@@ -724,16 +718,11 @@ public class SAMLineParser {
         if (mFile != null) {
             fileMessage = "File " + mFile + "; ";
         }
-        final String lineText;
-        if (this.currentLine != null) {
-            lineText = this.currentLine;
-        } else if (this.currentLineBuf != null) {
-            // Byte-based parseLine path: materialize the line lazily so we don't pay the
-            // String allocation on the (much more common) success path.
-            lineText = new String(currentLineBuf, currentLineOff, currentLineLen, StandardCharsets.ISO_8859_1);
-        } else {
-            lineText = "";
-        }
+        // Materialize the offending line lazily from the byte buffer so we don't pay the String
+        // allocation on the (much more common) success path.
+        final String lineText = this.currentLineBuf != null
+                ? new String(currentLineBuf, currentLineOff, currentLineLen, StandardCharsets.ISO_8859_1)
+                : "";
         return "Error parsing text SAM file. "
                 + reason + "; " + fileMessage + "Line "
                 + (this.currentLineNumber <= 0 ? "unknown" : this.currentLineNumber)
--- a/src/main/java/htsjdk/samtools/SAMLineParser.java
+++ b/src/main/java/htsjdk/samtools/SAMLineParser.java
@@ -26,7 +26,6 @@ package htsjdk.samtools;
 import htsjdk.samtools.util.StringUtil;
 import java.io.File;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 
 /**
@@ -318,7 +317,11 @@ public class SAMLineParser {
         samRecord.setMateAlignmentStart(matePos);
         samRecord.setInferredInsertSize(isize);
         if (!mFields[SEQ_COL].equals("*")) {
-            validateReadBases(mFields[SEQ_COL]);
+            // In SILENT mode reportErrorParsingLine is a no-op so the per-base scan is pure waste;
+            // skipping it preserves observable behavior.
+            if (this.validationStringency != ValidationStringency.SILENT) {
+                validateReadBases(mFields[SEQ_COL]);
+            }
             samRecord.setReadString(mFields[SEQ_COL]);
         } else {
             samRecord.setReadBases(SAMRecord.NULL_SEQUENCE);
@@ -327,7 +330,9 @@ public class SAMLineParser {
             if (samRecord.getReadBases() == SAMRecord.NULL_SEQUENCE) {
                 reportErrorParsingLine("QUAL should not be specified if SEQ is not specified");
             }
-            if (samRecord.getReadString().length() != mFields[QUAL_COL].length()) {
+            // Use the byte[] length directly rather than getReadString().length(), which would
+            // allocate a fresh String from the bases byte[] just to discard it.
+            if (samRecord.getReadLength() != mFields[QUAL_COL].length()) {
                 reportErrorParsingLine("length(QUAL) != length(SEQ)");
             }
             samRecord.setBaseQualityString(mFields[QUAL_COL]);
@@ -409,23 +414,22 @@ public class SAMLineParser {
     }
 
     private void parseTag(final SAMRecord samRecord, final String tag) {
-        Map.Entry<String, Object> entry = null;
+        final Object value;
         try {
-            entry = tagCodec.decode(tag);
+            tagCodec.decodeValue(tag);
+            value = tagCodec.getLastValue();
         } catch (SAMFormatException e) {
             reportErrorParsingLine(e);
+            return;
         }
-        if (entry != null) {
-            if (entry.getValue() instanceof TagValueAndUnsignedArrayFlag) {
-                final TagValueAndUnsignedArrayFlag valueAndFlag = (TagValueAndUnsignedArrayFlag) entry.getValue();
-                if (valueAndFlag.isUnsignedArray) {
-                    samRecord.setUnsignedArrayAttribute(entry.getKey(), valueAndFlag.value);
-                } else {
-                    samRecord.setAttribute(entry.getKey(), valueAndFlag.value);
-                }
-            } else {
-                samRecord.setAttribute(entry.getKey(), entry.getValue());
-            }
+        // The tag layout is fixed at KEY[2]:T:VALUE so we can compute the binary tag here without
+        // allocating a 2-character substring just to throw it away inside SAMTag.makeBinaryTag.
+        final short binaryTag = (short) (tag.charAt(1) << 8 | tag.charAt(0));
+        if (value instanceof TagValueAndUnsignedArrayFlag) {
+            final TagValueAndUnsignedArrayFlag valueAndFlag = (TagValueAndUnsignedArrayFlag) value;
+            samRecord.setAttribute(binaryTag, valueAndFlag.value, valueAndFlag.isUnsignedArray);
+        } else {
+            samRecord.setAttribute(binaryTag, value);
         }
     }
 
--- a/src/main/java/htsjdk/samtools/SAMLineParser.java
+++ b/src/main/java/htsjdk/samtools/SAMLineParser.java
@@ -270,9 +270,13 @@ public class SAMLineParser {
         if (numFields == mFieldOffsets.length) {
             reportErrorParsingLine("Too many fields in SAM text record.");
         }
-        for (int i = 0; i < numFields; ++i) {
-            if (mFieldLengths[i] == 0) {
-                reportErrorParsingLine("Empty field at position " + i + " (zero-based)");
+        // In SILENT mode reportErrorParsingLine is a no-op so iterating every field just to find
+        // empty ones is pure waste; preserves observable behavior since no error would be raised.
+        if (this.validationStringency != ValidationStringency.SILENT) {
+            for (int i = 0; i < numFields; ++i) {
+                if (mFieldLengths[i] == 0) {
+                    reportErrorParsingLine("Empty field at position " + i + " (zero-based)");
+                }
             }
         }
         // createSAMRecord(header) already assigns mHeader in the SAMRecord constructor.
@@ -284,7 +288,7 @@ public class SAMLineParser {
         if (mParentReader != null) samRecord.setFileSource(new SAMFileSource(mParentReader, null));
         samRecord.setReadName(decodeField(buf, mFieldOffsets[QNAME_COL], mFieldLengths[QNAME_COL]));
 
-        final int flags = parseFlag(decodeField(buf, mFieldOffsets[FLAG_COL], mFieldLengths[FLAG_COL]), "FLAG");
+        final int flags = parseFlagFromBytes(buf, mFieldOffsets[FLAG_COL], mFieldLengths[FLAG_COL]);
         samRecord.setFlags(flags);
 
         final int rnameOff = mFieldOffsets[RNAME_COL];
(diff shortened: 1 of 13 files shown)


src/main/java/htsjdk/samtools/SAMLineParser.java
```diff
@@ -270,39 +333,45 @@ public SAMRecord parseLine(final String line, final int lineNumber) {
...
-            }
-            validateReferenceName(mateRName, "MRNM");
-            if (mateRName.equals("=")) {
-                if (samRecord.getReferenceName() == null) {
+            if (mrnmLen == 1 && buf[mrnmOff] == '=') {
+                if (resolvedRname == null) {
                     reportErrorParsingLine("MRNM is '=', but RNAME is not set");
                 }
-                samRecord.setMateReferenceName(samRecord.getReferenceName());
+                samRecord.setMateReferenceNameAndIndex(resolvedRname, resolvedRnameIndex);
```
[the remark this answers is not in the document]
tfenne: Good catch -- this is a real bug. When RNAME is specified but not present in the sequence dictionary, my code was passing null to `setMateReferenceNameAndIndex`, which would NPE downstream.

Applied a slight variation of your suggestion in 779d9d8: track whether RNAME was specified via a local boolean during the RNAME branch, so the MRNM='=' fast path can do `if (!rnameSpecified)` instead of a String comparison against `NO_ALIGNMENT_REFERENCE_NAME` (saves a getter + equals per record on the success path). The fallback when `resolvedRname == null` now routes through `setMateReferenceName(samRecord.getReferenceName())` exactly as you suggested. Added a regression test `testMrnmEqualsWithRnameNotInDictionaryDoesNotNpe`.

src/main/java/htsjdk/samtools/SAMLineParser.java
```diff
@@ -353,79 +431,247 @@ public SAMRecord parseLine(final String line, final int lineNumber) {
...
+                        fast = false;
+                        break;
+                    }
+                }
+                if (fast) {
+                    return (int) acc;
+                }
+            }
+        }
+        return parseFlag(decodeField(buf, off, len), "FLAG");
```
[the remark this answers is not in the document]
tfenne: Not adopting this one, with a small caveat: the slow path here (`SamFlagField.parseDefault` -> `SamFlagField.DECIMAL.parse` -> `Integer.parseInt`) does *not* enforce the SAM 16-bit FLAG range either. Tightening the fast-path bound to `0xFFFF` would only re-route values in (65535, Integer.MAX_VALUE] through the slow path, where they would still be accepted with the same numeric value -- no observable behavior change.

In other words, `SAMRecord.isValid()` not re-checking the upper bound is a pre-existing issue independent of this PR's perf goal. Fixing it properly would require enforcing the cap in both paths (and would be a behavior change). Happy to file a follow-up issue, but I'd rather not couple it to this perf PR.

src/main/java/htsjdk/samtools/TextTagCodec.java
```diff
@@ -50,6 +51,33 @@ public class TextTagCodec {
...
+        @Override
+        public Object getValue() {
+            return value;
+        }
+
+        @Override
+        public Object setValue(final Object o) {
+            throw new UnsupportedOperationException();
+        }
+    }
```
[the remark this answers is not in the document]
tfenne: Agreed -- the API contract change wasn't worth it. `decode(String)` isn't called from the SAM-parsing hot path (the parser uses `decodeValue` + `getLastValue` directly), so the shared-entry trick saved nothing at the cost of a quiet breaking change for downstream callers.

Reverted in 779d9d8: `decode(String)` now returns a fresh `AbstractMap.SimpleImmutableEntry` per call, and the now-unused `MutableTagEntry` helper is removed. Benchmark unchanged.

src/main/java/htsjdk/samtools/SAMLineParser.java
```diff
@@ -219,48 +235,96 @@ public SAMRecord parseLine(final String line) {
      * @return a new SAMRecord object
      */
     public SAMRecord parseLine(final String line, final int lineNumber) {
-
-        this.currentLineNumber = lineNumber;
+        // SAM alignment lines are restricted to printable ASCII by the spec, so ISO-8859-1 is
+        // lossless and turns getBytes() into a bulk arraycopy on compact-string JVMs.
+        final byte[] bytes = line.getBytes(StandardCharsets.ISO_8859_1);
         this.currentLine = line;
```
yfarjoun: [nitpick](https://pullpo.io/cc?l=nitpick) 
parseLineFromBytes will nullify `this.currentLine` so you can drop this line

src/main/java/htsjdk/samtools/seekablestream/SeekableFileStream.java
```diff
@@ -41,37 +48,71 @@ public class SeekableFileStream extends SeekableStream {
...
         fis = new RandomAccessFile(file, "r");
+        this.length = file.length();
+        this.pos = 0;
         allInstances.add(this);
     }
 
     @Override
     public long length() {
-        return file.length();
+        return length;
```
yfarjoun: [question](https://pullpo.io/cc?l=question) 
this affects all files that are opened with SeekableFileStream. is it possible that a file is modified after it was opened? would we _want_ length to update or remain stale in that case?

src/main/java/htsjdk/samtools/seekablestream/SeekableFileStream.java
```diff
@@ -41,37 +48,71 @@ public class SeekableFileStream extends SeekableStream {
...
+        return pos;
     }
 
     @Override
-    public long skip(long n) throws IOException {
-        long initPos = position();
-        fis.getChannel().position(initPos + n);
-        return position() - initPos;
+    public long skip(final long n) throws IOException {
+        if (n <= 0) {
```
yfarjoun: [suggestion](https://pullpo.io/cc?l=suggestion) 
Since this is a change in contract (it was possible to skip backwards before), would it be better to throw an exception in the negative n case? Otherwise the caller might assume that it worked and that we are at the start of the file or something like that.

Also, this change in API should be called out loudly.

yfarjoun: [thought](https://pullpo.io/cc?l=thought) 
Looking more carefully into this is seems that any benefits that were obtained by the changes to SeekableFileStream may have been made irrelevant by replacement of BufferedLineReader with SamLineReader. 

Perhaps the commit order can be modified (moving Seekable.. to the bottom) to see if it still results in a speedup, and if not, drop it?

yfarjoun (changes requested):
Thanks @tfenne for the work to speed this up!

"I" noticed a few things that may warrant your attention. 

otherwise LGTM!

src/main/java/htsjdk/samtools/SAMLineParser.java
tfenne: Good catch — dropped in 418a42d. `parseLineFromBytes` nulls `currentLine` on entry (and rebuilds the error-message line lazily from the byte buffer), so this assignment was overwritten before any use.

src/main/java/htsjdk/samtools/seekablestream/SeekableFileStream.java
tfenne: Good question. I ended up removing the length caching entirely (see the thread below), so `length()` is back to master's per-call `file.length()` and this is moot. To answer it directly though: yes — with caching, a file growing under us (e.g. a concurrent writer) would not be observed. That stale-read behavior change applies to every BAM/CRAM/index consumer of `SeekableFileStream`, not just the SAM path, which is exactly why I dropped it rather than caching.

src/main/java/htsjdk/samtools/seekablestream/SeekableFileStream.java
tfenne: Agreed — this was a silent contract change (backwards skip no longer worked, and returning 0 instead is a footgun). Rather than throw + document loudly, I reverted `skip()` to master entirely; see the thread reply below — the whole `SeekableFileStream` optimization turned out not to be worth keeping.

tfenne: You're right, and thanks for pushing on this. I benchmarked the plain-SAM read path (10M-record SAM, `SILENT`, fully page-cached, 2 warmup + 5 measure) with and without this commit's `SeekableFileStream` changes: 1,089k vs 1,083k rec/s — a 0.5% delta, well inside run-to-run noise. `SamLineReader` reads raw bytes in 64KB blocks and never calls `available()`/`getFilePointer()` per record, so the syscall elimination no longer fires once it replaced `BufferedLineReader` earlier in the PR. Reverted `SeekableFileStream` and its tests to master in 418a42d.

yfarjoun:
[suggestion](https://pullpo.io/cc?l=suggestion) 
Another similar change-then-overwrite found:

The byte rewrite stranded the String-based tag decoders. decodeValue(String) / convertStringToObject(char, String) / parseIntegerTagValue(String) have zero internal callers: they are reachable only via the public decode(String). However they're a second, untested copy of the byte-path decode logic.

Perhaps decodeValue(String) could delegate to the byte overload (tag.getBytes(ISO_8859_1) -> decodeValue(byte[], 0, len)), using the same pattern parseLine(String) already uses, and delete the two String-only helpers?

tfenne:
Thanks @yfarjoun, fixed in aac1bf7c5 — `decode(String)` now delegates to the byte path and the duplicate String-based decoders are gone. A deeper sweep for the same change-then-supersede pattern caught two related strays as well (a dead `fields`/`NUM_TAG_FIELDS` here, and a now-unreachable `currentLine` field in `SAMLineParser`), cleaned up in the same push.

PR #1789, opened 2026-07-25
Let CRAMContainerStreamWriter callers keep ownership of their stream

### LINKED ISSUE #1092: Closure of OutputStream in class CRAMContainerStreamWriter
The class `CRAMContainerStreamWriter` receives an `OutputStream` as a parameter (which is thus created outside of the class), but it does close it in its `finish` method, by running `outputStream.close()`.

This can create problems when using the `CRAMContainerStreamWriter` class within classes which already take care of the initizialization and closure of the `OutputStream`, e.g.
```
OutputStream os = CreateSomeStream()
// Use CRAMContainerStreamWriter(os, ...)
os.close() // Error: already called by CRAMContainerStreamWriter
```
I'd suggest to have some `finishNoClose` method which takes care of flushing the caches and writing the EOF, but without closing the `OutputStream`. I'd recommend this to be default behaviour, if not for compatibility issues, since initialization and closure of the `OutputStream` are better done at the same level of code.

tfenne:
Closes #1092.

`CRAMContainerStreamWriter` takes an `OutputStream` created by the caller but unconditionally closed it in `finish()`, so code that manages the stream lifecycle itself had no way to finish writing without having its stream closed underneath it — and a subsequent `close()` by the owner would then fail.

Adds `finish(writeEOFContainer, closeStream)` so callers can opt out of the close. The existing `finish(writeEOFContainer)` overload delegates with `closeStream=true`, so behaviour is unchanged for every current caller, including `CRAMFileWriter`, which relies on it to close the file stream it owns.

The stream is flushed either way, so opting out never loses data.

Tests cover both overloads, that the single-argument form still closes (backwards compatibility), and that a caller can keep using its own stream afterwards.

## Summary by CodeRabbit

* **New Features**
  * Added an overload to CRAM writer finalization that lets callers choose whether the underlying output stream is closed.
  * The original single-argument finalization continues to close the stream by default.

* **Bug Fixes**
  * Finalization now conditionally closes the stream only when requested, allowing callers to keep using the same stream after finalization.
  * Flushing, optional EOF container emission, indexing, and existing error-handling behavior are preserved.

### DIFF
--- a/src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
+++ b/src/main/java/htsjdk/samtools/CRAMContainerStreamWriter.java
@@ -123,11 +123,30 @@ public class CRAMContainerStreamWriter {
     }
 
     /**
-     * Finish writing to the stream. Flushes the record cache and optionally emits an EOF container.
+     * Finish writing to the stream. Flushes the record cache, optionally emits an EOF container, and then
+     * closes the underlying output stream.
+     *
+     * <p>Note that this closes a stream that was supplied by the caller. Callers that own the lifecycle of
+     * the stream themselves should use {@link #finish(boolean, boolean)} with {@code closeStream} set to
+     * false, and close the stream on their own terms.
+     *
      * @param writeEOFContainer true if an EOF container should be written. Only use false if writing a CRAM file
      *                          fragment which will later be aggregated into a complete CRAM file.
      */
     public void finish(final boolean writeEOFContainer) {
+        finish(writeEOFContainer, true);
+    }
+
+    /**
+     * Finish writing to the stream. Flushes the record cache and optionally emits an EOF container.
+     *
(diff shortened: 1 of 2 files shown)


tfenne:
Good catch, thanks — fixed in the follow-up commit.

You were right that the assertion proved nothing about flushing: `ByteArrayOutputStream` makes writes visible immediately and inherits a no-op `flush()`, so `size() > 0` only showed that bytes had been written. It would have passed just as happily if `finish(..., false)` had stopped flushing altogether.

`CloseRecordingStream` now tracks `flush()` alongside `close()`, and the test asserts the flag instead of the stream size.

I confirmed the new assertion actually bites by temporarily removing the `outputStream.flush()` call from `finish()`:

```
AssertionError: declining the close must still flush the stream, or data can be lost
  expected [true] but found [false]
```

The old `size() > 0` assertion passed under that same regression.

PR #1797, opened 2026-08-24
Report the correct preceding record when alignments are out of order

AbhinavMir:
SAMSortOrderChecker.isSorted() records the alignment it checks as the new
previous record. SAMFileWriterImpl.assertPresorted() called
getPreviousRecord() after isSorted(), so the "out of order" message named
the offending record twice instead of naming the record before it.

Capture the previous record before the check, as SamFileValidator and
SamReader.AssertingIterator already do.


tfenne:
Thanks for this @AbhinavMir.  I've updated the PR description based on your original commit's message, and approved workflows to run.  I'll merge once they pass.

tfenne (approved):

PR #1799, opened 2026-09-06
SAMUtils: clip partial = and X elements like M

dnncha:
## Problem and change

`getNumOverlappingAlignedBasesToClip()` correctly handles a mate start inside an `M` element, but handles `=` and `X` through the branch that counts the entire element. With a 150-base alignment starting at 1001 and mate starting at 1051, `150M` returns 100 but `150=` and `150X` return 150.

Route all three aligned operators through the existing partial-element calculation. Leave insertion, deletion, skip, clipping and padding handling unchanged.

Tests cover boundary/interior mate starts, mixed `=`/`X` CIGARs, insertions, deletions, skips and clipping. They also assert the resulting CIGAR from the public clipping API, read-length preservation, and non-mutation when `noSideEffects=true`.

## Validation

Pinned upstream commit: `78296e9adab558f053241720426102d64ee603ff`.

The added tests were first run against unmodified production code: **32 tests/checks, 21 failures**. After the source fix, **63 tests/checks passed with no failures**, and the focused upstream `SAMUtils*Test` suite and `spotlessCheck` passed. HTSJDK validation is limited to the focused SAMUtils tests, not its entire repository test suite.

[Native build logs and test-report artifacts](https://github.com/dnncha/turbo-picard/actions/runs/34027560803). These are executions of the actual upstream code, not an independent algorithm model.

Implementation and regression tests were prepared with AI assistance.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected overlap clipping for reads using extended CIGAR operators (`=` and `X`), ensuring aligned bases are counted and soft-clipped correctly.

- **Tests**
  - Added coverage for overlap counting and clipping across extended CIGAR operations, including combinations with insertions, deletions, skipped regions, and existing clipping.


src/main/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -1129,7 +1129,9 @@ else if (rec.getMateAlignmentStart() == rec.getAlignmentStart() && rec.getFirstO
             final CigarOperator operator = el.getOperator();
             final int refBasesLength = operator.consumesReferenceBases() ? el.getLength() : 0;
             if (refStartPos <= refPos + refBasesLength - 1) { // add to clipped bases
-                if (operator == CigarOperator.MATCH_OR_MISMATCH) { // M
+                if (operator == CigarOperator.MATCH_OR_MISMATCH
+                        || operator == CigarOperator.EQ
+                        || operator == CigarOperator.X) { // M, =, X
```
tfenne, line 1134: `CigarOperator.isAlignment()` already means exactly M, X or EQ, so it can replace the three-way `||`.

```suggestion
                if (operator.isAlignment()) { // M, =, X
```

src/test/java/htsjdk/samtools/SAMUtilsExtendedCigarOverlapTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+import htsjdk.HtsjdkTest;
+import java.util.ArrayList;
+import java.util.List;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class SAMUtilsExtendedCigarOverlapTest extends HtsjdkTest {
+    private SAMRecord record(final String cigar, final int mateStart) {
```
tfenne, line 34:
```suggestion
    /**
     * Builds a read aligned at 1001 on chr1 with the given cigar and a mate starting at {@code mateStart}.
     * The read is second of pair so that it is the one clipped when both mates start at the same position.
     */
    private SAMRecord record(final String cigar, final int mateStart) {
```

src/test/java/htsjdk/samtools/SAMUtilsExtendedCigarOverlapTest.java
```diff
@@ -0,0 +1,101 @@
...
+            cases.add(new Object[] {"150" + operator, 1051, 100});
+            cases.add(new Object[] {"150" + operator, 1150, 1});
+            cases.add(new Object[] {"150" + operator, 1151, 0});
+        }
+        cases.add(new Object[] {"40=20X90=", 1051, 100});
+        cases.add(new Object[] {"60=5I90=", 1051, 105});
+        cases.add(new Object[] {"60=5D90=", 1051, 100});
+        cases.add(new Object[] {"40=5I110=", 1051, 100});
+        cases.add(new Object[] {"40=5D110=", 1051, 105});
+        cases.add(new Object[] {"40=10N110=", 1046, 110});
```
tfenne, line 65:
```suggestion
        // A single 150-base element with the mate starting at each boundary and an interior position.
        for (final String operator : new String[] {"M", "=", "X"}) {
            cases.add(new Object[] {"150" + operator, 1001, 150}); // same start: clip the whole read
            cases.add(new Object[] {"150" + operator, 1002, 149}); // one base in
            cases.add(new Object[] {"150" + operator, 1051, 100}); // in the middle
            cases.add(new Object[] {"150" + operator, 1150, 1}); // last aligned base
            cases.add(new Object[] {"150" + operator, 1151, 0}); // just past the read: no overlap
        }
        // The mate start falls inside an element other than the first one.
        cases.add(new Object[] {"40=20X90=", 1051, 100}); // 10 of the X plus all 90 of the trailing =
        // Indels inside the overlap: insertions are clipped in full, deletions contribute nothing.
        cases.add(new Object[] {"60=5I90=", 1051, 105}); // 10 of the = plus 5I plus 90
        cases.add(new Object[] {"60=5D90=", 1051, 100}); // 10 of the = plus 90
        // Indels before the overlap: the deletion shifts the last element right, the insertion does not.
        cases.add(new Object[] {"40=5I110=", 1051, 100}); // 110= covers 1041-1150
        cases.add(new Object[] {"40=5D110=", 1051, 105}); // 110= covers 1046-1155
        // A skipped region containing the mate start contributes nothing; the element after it is clipped in full.
        cases.add(new Object[] {"40=10N110=", 1046, 110});
        // Soft and hard clips on either end are ignored when counting.
```

src/test/java/htsjdk/samtools/SAMUtilsExtendedCigarOverlapTest.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Test(dataProvider = "overlapCounts")
+    public void testOverlapCount(final String cigar, final int mateStart, final int expected) {
+        Assert.assertEquals(SAMUtils.getNumOverlappingAlignedBasesToClip(record(cigar, mateStart)), expected);
+    }
+
+    @DataProvider(name = "clippedCigars")
+    public Object[][] clippedCigars() {
+        return new Object[][] {
```
tfenne, line 77:
```suggestion
        // The mate always starts at 1051, so the last 100 aligned bases are clipped in every case.
        // 5H10S150=10S5H is deliberately absent: a hard clip after a trailing soft clip is not handled by
        // clipOverlappingAlignedBases (see the FIXME there).
        return new Object[][] {
```

src/test/java/htsjdk/samtools/SAMUtilsExtendedCigarOverlapTest.java
```diff
@@ -0,0 +1,101 @@
...
+            {"150M", "50M100S"},
+            {"150=", "50=100S"},
+            {"150X", "50X100S"},
+            {"40=20X90=", "40=10X100S"},
+            {"60=5I90=", "50=105S"},
+            {"60=5D90=", "50=100S"},
+            {"40=5I110=", "40=5I10=100S"},
+            {"40=5D110=", "40=5D5=105S"},
+            {"10S150=10S", "10S50=110S"},
```
tfenne, line 86:
```suggestion
            {"60=5I90=", "50=105S"}, // insertion inside the overlap is clipped with it
            {"60=5D90=", "50=100S"}, // deletion inside the overlap disappears
            {"40=5I110=", "40=5I10=100S"}, // insertion before the overlap survives
            {"40=5D110=", "40=5D5=105S"}, // deletion before the overlap survives
            {"10S150=10S", "10S50=110S"}, // existing trailing soft clip merges into the new one
```

src/test/java/htsjdk/samtools/SAMUtilsExtendedCigarOverlapTest.java
```diff
@@ -0,0 +1,101 @@
...
+            {"5H150=", "5H50=100S"}
+        };
+    }
+
+    @Test(dataProvider = "clippedCigars")
+    public void testClippedCigar(final String cigar, final String expected) {
+        final SAMRecord original = record(cigar, 1051);
+        final SAMRecord clipped = SAMUtils.clipOverlappingAlignedBases(original, true);
+        Assert.assertEquals(clipped.getCigarString(), expected);
+        Assert.assertEquals(original.getCigarString(), cigar);
```
tfenne, line 96:
```suggestion
        Assert.assertEquals(original.getCigarString(), cigar); // noSideEffects=true must leave the input alone
```

src/test/java/htsjdk/samtools/SAMUtilsExtendedCigarOverlapTest.java
```diff
@@ -0,0 +1,101 @@
...
+package htsjdk.samtools;
+
+import htsjdk.HtsjdkTest;
+import java.util.ArrayList;
+import java.util.List;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class SAMUtilsExtendedCigarOverlapTest extends HtsjdkTest {
```
tfenne, line 33: Now that #1801 has landed there is a `SAMUtilsOverlapClippingTest` holding the existing overlap-clipping tests. Please rebase and add these tests there rather than creating a new class, so all the tests for this function live in one place. The suggestions below apply either way.

tfenne (changes requested):
@dnncha Thanks for this and your other PR.  I've left a pile of suggestions, most of which are just adding notes on exactly what each test is doing.

The one substantive request is that I didn't love how the clipping tests were getting spread over three files.  So I extracted the existing ones into a separate class in #1801 and would just ask you to relocate the new tests into that file please.

Once you update I'll be happy to merge.  Thanks again!

tfenne:
Merged as part of #1802

PR #1800, opened 2026-09-06
SAMUtils: do not overlap-clip mates on different references

dnncha:
## Problem and change

The overlap helper compares numeric alignment positions without first checking that the mates are on the same reference sequence. A read on chr1 at 1001 with `150M` and a mate on chr2 at 1051 therefore returns 100 bases to clip instead of zero.

Return zero for different reference names before comparing coordinates. Comparing names also supports headerless records without introducing reference-index resolution or boxed-integer identity assumptions. The public convenience clipping method then leaves cross-reference records unchanged.

Tests include swapped references, equal positions, the first/second-of-pair tie rule, non-overlapping coordinates, reference indices above the Integer cache range, and headerless records. Same-reference clipping remains covered.

Related downstream report: broadinstitute/picard#2039. This is a library fix, not a claim that already-released Picard versions pick it up automatically.

## Validation

Pinned upstream commit: `78296e9adab558f053241720426102d64ee603ff`.

The added tests were first run against unmodified production code: **13 tests/checks, 5 failures**. After the source fix, **44 tests/checks passed with no failures**, and the focused upstream `SAMUtils*Test` suite and `spotlessCheck` passed. HTSJDK validation is limited to the focused SAMUtils tests, not its entire repository test suite.

[Native build logs and test-report artifacts](https://github.com/dnncha/turbo-picard/actions/runs/34027560803). These are executions of the actual upstream code, not an independent algorithm model.

Implementation and regression tests were prepared with AI assistance.

## Summary by CodeRabbit

* **Bug Fixes**
  * Overlap clipping now returns zero when a paired read and its mate are aligned to different reference sequences.
  * Records with mates on different references remain unchanged during clipping.

* **Tests**
  * Added coverage for same- and different-reference paired reads, various alignments, read-pair orientations, and headerless records.


src/main/java/htsjdk/samtools/SAMUtils.java
```diff
@@ -1114,6 +1115,9 @@ public static int getNumOverlappingAlignedBasesToClip(final SAMRecord rec) {
 
         if (!rec.getReadPairedFlag() || rec.getReadUnmappedFlag() || rec.getMateUnmappedFlag()) return 0;
 
+        // Alignment coordinates are only comparable on the same reference sequence.
```
tfenne, line 1118: Worth recording *why* names are compared rather than indices.

```suggestion
        // Mates on different references cannot overlap. Compare names rather than indices: the name is always
        // populated, whereas resolving an index requires a header.
```

src/test/java/htsjdk/samtools/SAMUtilsReferenceOverlapTest.java
```diff
@@ -0,0 +1,97 @@
...
+            {0, 1, 1051, false, 0},
+            {1, 0, 1051, false, 0},
+            {128, 128, 1051, false, 100},
+            {128, 129, 1051, false, 0},
+            {0, 0, 1001, false, 150},
+            {0, 1, 1001, false, 0},
+            {0, 0, 1001, true, 0},
+            {0, 1, 1001, true, 0},
+            {0, 1, 1151, false, 0},
+            {0, 1, 951, false, 0}
```
tfenne, line 45: Two changes here.

1. Rows that return zero on the *same* reference too (mate at 951 or 1151, and first-of-pair with equal starts) cannot distinguish the old code from the new, so I have dropped them. The remaining rows are the ones that actually failed before the fix, plus three same-reference controls.
2. The 128/129 rows exist to push indices past the `Integer` cache and guard a boxed-`==` bug. The name comparison cannot have that bug, so the test would be encoding an implementation hazard that isn't there. Dropping them also lets the header shrink to two sequences.

Each remaining row is labelled with what it exercises.

```suggestion
        // Columns: reference index, mate reference index, mate start, first of pair, expected bases to clip.
        // The read is 150M at 1001 throughout.
        return new Object[][] {
            // Same reference: existing behaviour, as a control.
            {0, 0, 1051, false, 100},
            {0, 0, 1001, false, 150},
            {0, 0, 1001, true, 0}, // same start: the first of pair is the one left unclipped
            // Different references: these would overlap by coordinate alone and must not be clipped.
            {0, 1, 1051, false, 0},
            {1, 0, 1051, false, 0},
            {0, 1, 1001, false, 0},
```

src/test/java/htsjdk/samtools/SAMUtilsReferenceOverlapTest.java
```diff
@@ -0,0 +1,97 @@
...
+        record.setMateAlignmentStart(mateStart);
+        record.setCigarString("150M");
+        record.setReadString("A".repeat(150));
+        Assert.assertEquals(SAMUtils.getNumOverlappingAlignedBasesToClip(record), expected);
+        if (reference != mateReference) {
+            final SAMRecord clipped = SAMUtils.clipOverlappingAlignedBases(record, true);
+            Assert.assertSame(clipped, record);
+            Assert.assertEquals(clipped.getCigarString(), "150M");
+            Assert.assertFalse(clipped.getReadUnmappedFlag());
+        }
```
tfenne, line 77: The record construction is pulled into a helper, and the `if (reference != mateReference)` block becomes its own named test. That way the count test asserts one thing per row, and the "clipping leaves a cross-reference record untouched" property is a test someone can find by name. With the 128/129 rows gone the header only needs two sequences.

```suggestion
    /** Builds a 150M read at 1001 on the given reference with a mate on {@code mateReference} at {@code mateStart}. */
    private static SAMRecord pairedRead(
            final int reference, final int mateReference, final int mateStart, final boolean firstOfPair) {
        final SAMFileHeader header = new SAMFileHeader();
        header.addSequence(new SAMSequenceRecord("chr1", 10000));
        header.addSequence(new SAMSequenceRecord("chr2", 10000));
        final SAMRecord record = new SAMRecord(header);
        record.setReadName("reference-identity");
        record.setReadPairedFlag(true);
        record.setFirstOfPairFlag(firstOfPair);
        record.setSecondOfPairFlag(!firstOfPair);
        record.setReferenceIndex(reference);
        record.setMateReferenceIndex(mateReference);
        record.setAlignmentStart(1001);
        record.setMateAlignmentStart(mateStart);
        record.setCigarString("150M");
        record.setReadString("A".repeat(150));
        return record;
    }

    @Test(dataProvider = "references")
    public void testOverlapCountAcrossReferences(
            final int reference,
            final int mateReference,
            final int mateStart,
            final boolean firstOfPair,
            final int expected) {
        final SAMRecord record = pairedRead(reference, mateReference, mateStart, firstOfPair);
        Assert.assertEquals(SAMUtils.getNumOverlappingAlignedBasesToClip(record), expected);
    }

    @Test
    public void testClippingLeavesCrossReferenceRecordUntouched() {
        final SAMRecord record = pairedRead(0, 1, 1051, false);
        final SAMRecord clipped = SAMUtils.clipOverlappingAlignedBases(record, true);
        Assert.assertSame(clipped, record);
        Assert.assertEquals(clipped.getCigarString(), "150M");
        Assert.assertFalse(clipped.getReadUnmappedFlag());
```

src/test/java/htsjdk/samtools/SAMUtilsReferenceOverlapTest.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @DataProvider(name = "headerlessReferences")
+    public Object[][] headerlessReferences() {
+        return new Object[][] {{"chr1", 100}, {"chr2", 0}};
+    }
+
+    @Test(dataProvider = "headerlessReferences")
```
tfenne, line 85: Since this test is the one that would fail if indices were used instead of reference name (and that's not immediately obvious), document it.

```suggestion
    /**
     * A record with no header cannot resolve reference indices at all, so the comparison has to be by name.
     * This is the case an index-based comparison could not handle.
     */
    @Test(dataProvider = "headerlessReferences")
```

src/test/java/htsjdk/samtools/SAMUtilsReferenceOverlapTest.java
```diff
@@ -0,0 +1,97 @@
...
+ * THE SOFTWARE.
+ */
+package htsjdk.samtools;
+
+import htsjdk.HtsjdkTest;
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class SAMUtilsReferenceOverlapTest extends HtsjdkTest {
```
tfenne, line 31: Now that #1801 has landed there is a `SAMUtilsOverlapClippingTest` holding the existing overlap-clipping tests. Please rebase and add these tests there rather than creating a new class, so all the tests for this function live in one place. The suggestions below apply either way.

tfenne (changes requested):
Thanks for this.  Same story as #1799.  A few suggestions you can hopefully just apply (pushback if you disagree), and then the same request to relocate tests before merge please!

tfenne:
Merged as part of #1802

PR #1802, opened 2026-09-17
SAMUtils overlap clipping: handle = and X elements, and skip mates on different references

tfenne:
### Description

This PR combines and supersedes #1799 and #1800, both by @dnncha. Thank you @dnncha for finding and fixing these two bugs, and for the thorough tests that came with them. The fixes and tests here are theirs; this PR exists only to get them merged. It merges both branches (preserving the original commits and authorship) and adds one commit on top applying the review feedback I left on those PRs, so the fixes can land without another round trip.

**The two fixes, in `SAMUtils.getNumOverlappingAlignedBasesToClip`:**

- From #1799: a partially overlapping `=` or `X` cigar element is now clipped the same way as `M`. Previously only `M` was pro-rated by the mate's start position, so reads with extended cigars had the entire element clipped (e.g. `150=` with a mate starting 50 bases in became `150S` instead of `50=100S`).
- From #1800: mates on different reference sequences are no longer treated as overlapping. Previously only the coordinates were compared, so a pair on `chr1:1001` / `chr2:1051` was clipped as if the mates overlapped.

**The review fixups (final commit):**

- Use `CigarOperator.isAlignment()` in place of the explicit `M`/`=`/`X` comparison.
- Document why mate references are compared by name rather than by index: the name is always populated, whereas resolving an index requires a header.
- Move the new tests into `SAMUtilsOverlapClippingTest` (introduced in #1801) so that all tests for these two methods live in one class, replacing `SAMUtilsExtendedCigarOverlapTest` and `SAMUtilsReferenceOverlapTest`.
- Annotate each data provider row with what it exercises, and note why `5H10S150=10S5H` is absent from the clipped-cigar cases (the existing FIXME in `clipOverlappingAlignedBases`).
- Trim the cross-reference cases to those that distinguish the old behaviour from the new plus same-reference controls, dropping the rows that return zero on the same reference too and the 128/129 rows guarding a boxed-`==` hazard that the name comparison cannot have.
- Extract a `pairedRead` helper, and make "clipping leaves a cross-reference record untouched" its own named test.

**Testing:** `SAMUtilsOverlapClippingTest` passes (46 tests). Run against `SAMUtils.java` from master, 26 of those 46 fail, covering both fixes, including every cross-reference row and the headerless `chr2` case.

Once this is merged, #1799 and #1800 can be closed.

- [x] Check your code style.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected overlap clipping for reads using all alignment operators, including exact matches and mismatches.
  * Prevented overlap clipping between records mapped to different references.
  * Improved handling of records without reference headers by comparing reference names directly.
  * Preserved expected behavior for insertions, deletions, skipped regions, and soft or hard clipping.


dnncha:
Great, thanks @tfenne

PR #1824, opened 2026-09-19
Read and write VCF, GFF3 and the Tribble line readers as UTF-8

tfenne:
Every VCF and GFF3 text read path and the VCF writer now use UTF-8. The shared Tribble line readers (`AsciiLineReader`, `SynchronousLineReader`, `BlockCompressedInputStream.readLine`, `TabixReader.readLine`) buffer raw bytes and decode each line as UTF-8; malformed bytes (e.g. bare Latin-1 in old files) become U+FFFD and are lossy on rewrite. `VCFEncoder.VCF_CHARSET` is UTF-8, so the `VCFWriter`'s `OutputStreamWriter` encodes UTF-8, and `BCF2Writer` encodes the embedded VCF header text with the same charset. `Gff3Writer` is explicitly UTF-8, which the GFF3 specification mandates. `TabixReader.readLine` also no longer drops a final line that lacks a trailing newline, matching htslib's `bgzf_getline`.

Performance is neutral on the VCF read and write benchmarks: all six `VariantCodecBenchmark` comparisons have |t| < 1.6 over 8 alternating forks a side. A new `LineReaderBenchmark` measures the changed readers directly. `AsciiLineReader` is 14% faster on short lines and 8% faster on 27 KB lines after keeping the line buffer in a local variable, which lets the JIT hoist the field and length loads out of the per-byte loop. `TabixReader.readLine` is 22% faster: the old code appended each byte to a synchronized `StringBuffer` and copied the line out again with `toString()`; the new one stores bytes into a plain array and builds the `String` once.

Follows #1823 in the VCF 4.4/4.5 series.

## Summary by CodeRabbit

* **New Features**
  * VCF, GFF3, BCF, and related text formats now consistently read and write UTF-8, including non-ASCII content.
* **Bug Fixes**
  * Tabix-indexed files now retain a final line even when it lacks a trailing newline.
  * Line readers correctly handle UTF-8 text, long lines, and multibyte characters.
* **Tests**
  * Added coverage for UTF-8 round trips, compressed files, GFF3 comments, BCF headers, and unterminated final lines.
* **Chores**
  * Added benchmarks for line-reading performance across supported reader implementations.


tfenne:
@coderabbitai review

tfenne:
@coderabbitai pause

